Issue #58, opened 2015-03-03
management.NewClientFromPublishSettings is undefined


waitingkuo:
According to README.md
We can use 

    client, err := management.NewClientFromPublishSettings(SUBSCRIPTION_ID, SUBSCRIPTION_CERTIFICATE)
        if err != nil {
            fmt.Println(err)
            os.Exit(1)
        }

to create a client.

But NewClientFromPublishSettings is undefined.

hopetobelievein:
Updated readme.
Now we can use:
<code>subscriptionCert, err := ioutil.ReadFile(SUBSCRIPTION_CERTIFICATE_PATH)
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }

    client, err := management.NewClient(SUBSCRIPTION_ID, subscriptionCert)
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }
</code>

waitingkuo:
Thanks, it works!
Is it possible to use the publishsettings file to create the client?
#57

anweiss:
@ahmetalpbalkan Since we made cert handling the responsibility to the caller, how do we want to provide for .publishsettings handling?

jen20:
@anweiss In my opinion this library should handle the `.publishsettings` files as it's Azure specific.

ahmetalpbalkan:
@jen20 +1. We need the wa

waitingkuo:
According to the current architecture, i've made a patch
3c9cf2fecccf6bbbedbc117ac48c12f381536e02

ahmetalpbalkan:
@waitingkuo can you please send a pull request? We should probably add another constructor that takes Config as well and one should call another (just like what we have today)

paulmey:
fixed in #65 /  ea0d561

Issue #57, opened 2015-03-03
Import PublishSettings from File


waitingkuo:
It seems that there's a private function for importing the publishsetting file
[https://github.com/MSOpenTech/azure-sdk-for-go/blob/master/management/publishSettings.go#L31]

But I can't find another function that use the private function. Will it be public or have a NewClientFromPublishSettingsFile constructor?

hopetobelievein:
It was removed here: https://github.com/MSOpenTech/azure-sdk-for-go/commit/63430b193b5e705cfe3e00054c4e6a3fcf5bd3b8
@anweiss can you please address this issue?

anweiss:
so not sure on this yet. we left the private functions in there just in case. I'll let @ahmetalpbalkan chime in.

waitingkuo:
It seems that we have to create a client then import the publishsettings later.
3c9cf2fecccf6bbbedbc117ac48c12f381536e02

paulmey:
fixed in #65 /  ea0d561

PR #61, opened 2015-03-06
add NewClientFromPublishSettingsFile

waitingkuo:
#57 #58 

Add a function NewClientFromPublishSettingsFile() to create a client from the publishsettings file

Signed-off-by: Wei-Ting Kuo <waitingkuo0527@gmail.com>

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -54,6 +54,24 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
 	return makeClient(subscriptionID, managementCert, config.ManagementURL)
 }
 
+// NewClientFromPublishSettingsFile creates a new Client using a publishsettings file
+func NewClientFromPublishSettingsFile(filePath string) (Client, error) {
+	client := Client{
+		managementURL: defaultAzureManagementURL,
+	}
+	err := client.importPublishSettingsFile(filePath)
+	return client, err
+}
+
+// NewClientFromPublishSettingsFileAndConfig creates a new Client using a publishsettings file and a given Client Config
+func NewClientFromPublishSettingsFileAndConfig(filePath string, config ClientConfig) (Client, error) {
+	client := Client{
+		managementURL: config.ManagementURL,
+	}
+	err := client.importPublishSettingsFile(filePath)
+	return client, err
+}
+
 func makeClient(subscriptionID string, managementCert []byte, managementURL string) (Client, error) {
 	var client Client
 	if subscriptionID == "" {


management/client.go
```diff
@@ -54,6 +54,21 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
 	return makeClient(subscriptionID, managementCert, config.ManagementURL)
 }
 
+// NewClientFromPublishSettingsFile creates a new Client using a publishsettings file
+func NewClientFromPublishSettingsFile(filePath string) (Client, error) {
```
ahmetalpbalkan: We need another constructor that take publish settings file and config (like NewClientFromConfig)

How about these two:
- NewClientFromPublishSettings
- NewClientFromPublishSettingsAndConfig

see the existing constructors above.

management/client.go
```diff
@@ -54,6 +54,21 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
...
+
+	client := Client{
+		managementURL: defaultAzureManagementURL,
+	}
+
+	err := client.importPublishSettingsFile(filePath)
+	if err != nil {
```
ahmetalpbalkan: no need for this err check, just `return client, err` below.

waitingkuo:
Updated

ahmetalpbalkan:
@waitingkuo please update the other comment as well. also please squash your commits into one before pushing. Thanks!

ahmetalpbalkan:
@waitingkuo ping, can you update? otherwise #65 also addresses this problem.

waitingkuo:
Hi @ahmetalpbalkan 

`publishSettings` is private
https://github.com/MSOpenTech/azure-sdk-for-go/blob/master/management/publishSettings.go#L112-L117

Users cannot construct the client by `publishSettings`. Need I make it public?

ahmetalpbalkan:
@waitingkuo the only thing I'm asking for: currently what you have is:

    func NewClientFromPublishSettingsFile(filePath string) (Client, error)

and I'm saying let's have these two instead:

    func NewClientFromPublishSettingsFile(publishSettings string) (Client, error)
    func NewClientFromPublishSettingsFileAndConfig(publishSettings string, configConfig) (Client, error)

You don't need to accept `publishSettings` type as parameter. It's for the parsed publishSettings file and should not be used as input.

waitingkuo:
Sorry for the misunderstanding, just updated

management/client.go
```diff
@@ -54,6 +54,30 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
...
+
+	err := client.importPublishSettingsFile(filePath)
+
+	return client, err
+}
+
+// NewClientFromPublishSettingsFileAndConfig creates a new Client using a publishsettings file and a given Client Config
+func NewClientFromPublishSettingsFileAndConfig(filePath string, config ClientConfig) (Client, error) {
+
+	client := Client{
```
ahmetalpbalkan: please use `NewClientFromConfig` to construct the client here. you're duplicating the logic.

management/client.go
```diff
@@ -54,6 +54,30 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
...
+	}
+
+	err := client.importPublishSettingsFile(filePath)
+
+	return client, err
+}
+
+// NewClientFromPublishSettingsFileAndConfig creates a new Client using a publishsettings file and a given Client Config
+func NewClientFromPublishSettingsFileAndConfig(filePath string, config ClientConfig) (Client, error) {
+
```
ahmetalpbalkan: redundant empty line @waitingkuo

management/client.go
```diff
@@ -54,6 +54,30 @@ func NewClientFromConfig(subscriptionID string, managementCert []byte, config Cl
...
+	client := Client{
+		managementURL: config.ManagementURL,
+	}
+
+	err := client.importPublishSettingsFile(filePath)
+
```
ahmetalpbalkan: redundant empty line

ahmetalpbalkan:
I am mistaken. `client.importPublishSettingsFil`e has a side effect (i.e. setting a global variable), this method should be removed, instead it should return a piece of information (e.g. cert/subs-id) which be embedded in client. 

Somebody must change that non-idempotent behavior, otherwise we're far from an ideal solution. I know it's a bit hard but current design of the library is just wrong. cc:@paulmey

ahmetalpbalkan:
@waitingkuo we addressed this in #65. let us know if this solves your problem.

PR #62, opened 2015-03-09
Replace openssl dependency with pure go implementation for converting

paulmey:
I implemented PKCS12 from the RFC's to read PFX files. With this, we can drop the implicit openssl dependency.

cc: @ahmetalpbalkan

### DIFF
--- a/management/publishSettings.go
+++ b/management/publishSettings.go
@@ -2,10 +2,13 @@ package management
 
 import (
 	"encoding/base64"
+	"encoding/pem"
 	"encoding/xml"
 	"errors"
 	"fmt"
 	"io/ioutil"
+
+	"github.com/Azure/go-pkcs12"
 )
 
 func (client *Client) importPublishSettings(id string, certPath string) error {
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
oh man definitely not "Add my pkcs12 library" :smile: why don't we import from your github repo?

paulmey:
@ahmetalpbalkan: you want the history? Sure, give me a minute...

ahmetalpbalkan:
@paulmey I'm saying we should either make use of your lib by directly `import "github.com/paulmey/pkcs12"` or you need to come up with some godep setup to vendor the dependency. There's no point of duplicating this code here --unless you want to delete your repo and make everybody else use this package from `import "github.com/MSOpenTech/azure-sdk-for-go/utils/pkcs12"`.

paulmey:
The latter is my goal, but let me circle back with the folks here on the other possibilities...

management/http.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const (
 	msVersionHeader           = "x-ms-version"
-	msVersionHeaderValue      = "2014-05-01"
+	msVersionHeaderValue      = "2014-10-01"
```
ahmetalpbalkan: is there a reason to update this? if not let's leave it as is. in fact, this value should come from the `Config` object we added very recently.

management/hostedservice/client.go
```diff
@@ -132,3 +135,7 @@ func (self HostedServiceClient) createHostedServiceDeploymentConfig(dnsName, loc
 	}
 	return deployment
 }
+
+func (self HostedServiceClient) AddServiceCertificate(dnsName string, certificate []byte, format, password string) {
```
ahmetalpbalkan: leftover?

paulmey:
Referenced github.com/Azure/go-pkcs12

ahmetalpbalkan:
LGTM. :+1:

PR #64, opened 2015-03-14
Update virtual machine entities to latest spec

paulmey:
Also cleaned up the code a bit. Preparing another PR for moving the utility methods to a different package.

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -242,7 +242,7 @@ func (self VirtualMachineClient) SetAzureDockerVMExtension(azureVMConfiguration 
 	return azureVMConfiguration, nil
 }
 
-func (self VirtualMachineClient) GetVMDeployment(cloudserviceName, deploymentName string) (*Deployment, error) {
+func (self VirtualMachineClient) GetVMDeployment(cloudserviceName, deploymentName string) (*DeploymentResponse, error) {
 	if cloudserviceName == "" {
 		return nil, fmt.Errorf(errParamNotSpecified, "cloudserviceName")
 	}
@@ -250,7 +250,7 @@ func (self VirtualMachineClient) GetVMDeployment(cloudserviceName, deploymentNam
 		return nil, fmt.Errorf(errParamNotSpecified, "deploymentName")
 	}
 
-	deployment := new(Deployment)
+	deployment := new(DeploymentResponse)
 
 	requestURL := fmt.Sprintf(azureDeploymentURL, cloudserviceName, deploymentName)
 	response, azureErr := self.client.SendAzureGetRequest(requestURL)
@@ -503,8 +503,8 @@ func (self VirtualMachineClient) addDockerPort(configurationSets []Configuration
 	return nil
 }
 
-func (self VirtualMachineClient) createVMDeploymentConfig(role *Role) Deployment {
-	return Deployment{
+func (self VirtualMachineClient) createVMDeploymentConfig(role *Role) DeploymentRequest {
+	return DeploymentRequest{
 		Name:           role.RoleName,
 		DeploymentSlot: "Production",
 		Label:          role.RoleName,
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -1,3 +1,4 @@
+// Package virtualmachine implements operations on Azure virtual machines using the Service Management REST API
 package virtualmachine
 
 import (
@@ -23,7 +24,6 @@ import (
 )
 
 const (
-	azureXmlns                        = "http://schemas.microsoft.com/windowsazure"
 	azureDeploymentListURL            = "services/hostedservices/%s/deployments"
 	azureHostedServiceListURL         = "services/hostedservices"
 	deleteAzureHostedServiceURL       = "services/hostedservices/%s?comp=media"
@@ -147,22 +147,17 @@ func (self VirtualMachineClient) AddAzureLinuxProvisioningConfig(azureVMConfigur
 		return nil, fmt.Errorf(errParamNotSpecified, "userName")
 	}
 
-	configurationSets := ConfigurationSets{}
 	provisioningConfig, err := self.createLinuxProvisioningConfig(azureVMConfiguration.RoleName, userName, password, certPath)
 	if err != nil {
 		return nil, err
 	}
-
-	configurationSets.ConfigurationSet = append(configurationSets.ConfigurationSet, provisioningConfig)
 
 	networkConfig, networkErr := self.createNetworkConfig(osLinux, sshPort)
 	if networkErr != nil {
 		return nil, err
 	}
 
-	configurationSets.ConfigurationSet = append(configurationSets.ConfigurationSet, networkConfig)
-
-	azureVMConfiguration.ConfigurationSets = configurationSets
+	azureVMConfiguration.ConfigurationSets = &[]ConfigurationSet{provisioningConfig, networkConfig}
 
 	if len(certPath) > 0 {
 		azureVMConfiguration.UseCertAuth = true
@@ -189,32 +184,35 @@ func (self VirtualMachineClient) SetAzureVMExtension(azureVMConfiguration *Role,
 		return nil, fmt.Errorf(errParamNotSpecified, "referenceName")
 	}
 
-	extension := ResourceExtensionReference{}
-	extension.Name = name
-	extension.Publisher = publisher
-	extension.Version = version
-	extension.ReferenceName = referenceName
-	extension.State = state
-
-	if len(privateConfigurationValue) > 0 {
-		privateConfig := ResourceExtensionParameter{}
-		privateConfig.Key = "ignored"
-		privateConfig.Value = base64.StdEncoding.EncodeToString([]byte(privateConfigurationValue))
-		privateConfig.Type = "Private"
-
-		extension.ResourceExtensionParameterValues.ResourceExtensionParameterValue = append(extension.ResourceExtensionParameterValues.ResourceExtensionParameterValue, privateConfig)
+	extension := ResourceExtensionReference{
+		Name:          name,
+		Publisher:     publisher,
+		Version:       version,
+		ReferenceName: referenceName,
+		State:         state,
 	}
 
-	if len(publicConfigurationValue) > 0 {
-		publicConfig := ResourceExtensionParameter{}
-		publicConfig.Key = "ignored"
-		publicConfig.Value = base64.StdEncoding.EncodeToString([]byte(publicConfigurationValue))
-		publicConfig.Type = "Public"
-
-		extension.ResourceExtensionParameterValues.ResourceExtensionParameterValue = append(extension.ResourceExtensionParameterValues.ResourceExtensionParameterValue, publicConfig)
+	if privateConfigurationValue != "" {
+		extension.ParameterValues = append(extension.ParameterValues, ResourceExtensionParameter{
+			Key:   "ignored",
+			Value: base64.StdEncoding.EncodeToString([]byte(privateConfigurationValue)),
+			Type:  "Private",
+		})
 	}
 
-	azureVMConfiguration.ResourceExtensionReferences.ResourceExtensionReference = append(azureVMConfiguration.ResourceExtensionReferences.ResourceExtensionReference, extension)
+	if publicConfigurationValue != "" {
+		extension.ParameterValues = append(extension.ParameterValues, ResourceExtensionParameter{
+			Key:   "ignored",
+			Value: base64.StdEncoding.EncodeToString([]byte(publicConfigurationValue)),
+			Type:  "Public",
+		})
+	}
+
+	if azureVMConfiguration.ResourceExtensionReferences == nil {
+		azureVMConfiguration.ResourceExtensionReferences = &[]ResourceExtensionReference{extension}
+	} else {
+		*azureVMConfiguration.ResourceExtensionReferences = append(*azureVMConfiguration.ResourceExtensionReferences, extension)
+	}
 
 	return azureVMConfiguration, nil
 }
@@ -228,7 +226,7 @@ func (self VirtualMachineClient) SetAzureDockerVMExtension(azureVMConfiguration 
 		version = "0.3"
 	}
 
-	err := self.addDockerPort(azureVMConfiguration.ConfigurationSets.ConfigurationSet, dockerPort)
+	err := self.addDockerPort(*azureVMConfiguration.ConfigurationSets, dockerPort)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


management/virtualmachineimage/entities.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/xml"
 
 	"github.com/MSOpenTech/azure-sdk-for-go/management"
+	disk "github.com/MSOpenTech/azure-sdk-for-go/management/virtualmachinedisk"
```
ahmetalpbalkan: It might be better to call it vmdisk

management/virtualmachinedisk/entities.go
```diff
@@ -6,3 +6,12 @@ import "github.com/MSOpenTech/azure-sdk-for-go/management"
 type DiskClient struct {
 	client management.Client
 }
+
+type HostCachingType string // see "HostCachingType*"" constants
```
ahmetalpbalkan: Comments not really necessary as it's obvious :-) and thats not really a godoc comment either

management/virtualmachine/entities.go
```diff
@@ -178,14 +287,24 @@ type RoleSize struct {
...
+type LoadBalancer struct {
+	Name                          string        // Specifies the name of the internal load balancer.
+	Type                          IPAddressType `xml:"FrontendIpConfiguration>Type"`                                    // Specifies the type of virtual IP address that is provided by the load balancer. The only allowable value is Private.
+	SubnetName                    string        `xml:"FrontendIpConfiguration>SubnetName,omitempty"`                    // Required if the deployment exists in a virtual network and a StaticVirtualNetworkIPAddress is assigned. Specifies the subnet of the virtual network that the load balancer uses. The virtual IP address that is managed by the load balancer is contained in this subnet.
+	StaticVirtualNetworkIPAddress string        `xml:"FrontendIpConfiguration>StaticVirtualNetworkIPAddress,omitempty"` // Specifies a specific virtual IP address that the load balancer uses from the subnet in the virtual network.
 }
 
+type IPAddressType string // see "IPAddressType*"" constants
+
+const IPAddressTypePrivate = IPAddressType("Private") // Only allowed value (currently) for IPAddressType
```
ahmetalpbalkan: Use const(....) block and "const ( xxx IPAddrType = "foo" ) should be ok, i.e. no need cast type

management/virtualmachine/entities.go
```diff
@@ -11,132 +12,235 @@ type VirtualMachineClient struct {
...
+	Port      int                   // Specifies the external port to use for the endpoint.
+	Protocol  InputEndpointProtocol //Specifies the transport protocol for the endpoint.
+	Vip       string                `xml:",omitempty"`
+}
+
+type InputEndpointProtocol string // see "InputEndpointProtocol*"" constants
+
+const (
+	// Enum values for InputEndpointProtocol
+	InputEndpointProtocolTcp = InputEndpointProtocol("TCP")
```
ahmetalpbalkan: Move the type next to identifier, type conversion will be inferred automatically

paulmey:
Fixed all comments

management/virtualmachine/client.go
```diff
@@ -648,8 +648,8 @@ func (self VirtualMachineClient) createServiceCertDeploymentConf(certPath string
 	return certConfig, nil
 }
 
-func (self VirtualMachineClient) createSshConfig(certPath, userName string) (SSH, error) {
-	sshConfig := SSH{}
+func (self VirtualMachineClient) createSshConfig(certPath, userName string) (*SSH, error) {
```
ahmetalpbalkan: why not return SSH and reference with `&` at the caller?

management/virtualmachine/client.go
```diff
@@ -665,7 +665,7 @@ func (self VirtualMachineClient) createSshConfig(certPath, userName string) (SSH
 	publicKey.Fingerprint = fingerprint
 	publicKey.Path = "/home/" + userName + "/.ssh/authorized_keys"
```
ahmetalpbalkan: do you think we can just accept certPath as input to this method? hardcoded home path wouldn't work on windows or mac.

management/virtualmachine/client.go
paulmey: That would need an extra variable at the call site [1]. This code is being deleted in #65 , so I don't really care. LMK if you still want me to make the change.

[1] https://github.com/paulmey/azure-sdk-for-go/blob/virtualmachine-cleanup/management/virtualmachine/client.go#L607-607

management/virtualmachine/client.go
paulmey: This is the path where the authorized_keys file is on the vm being deployed. Not the certificate. That should already exist in the cloud service...

Sent from my Windows Phone

management/virtualmachine/client.go
ahmetalpbalkan: no big deal. having all these pointers is just annoying. I can understand if we re using them for omitting things in XML marshaling but otherwise we shouldnt be doing any pointers, ideally. right?

ahmetalpbalkan: just looked at the link. please do introduce the extra var. :)

management/virtualmachine/client.go
ahmetalpbalkan: ah ok

management/virtualmachine/client.go
paulmey: Totally agree and yes, they are only here to make xml.Marshall work properly... :-(

Sent from my Windows Phone

paulmey: Ok

Sent from my Windows Phone

paulmey:
removed pointer to SSH in return from method

management/virtualmachine/entities.go
```diff
@@ -11,132 +12,235 @@ type VirtualMachineClient struct {
...
+	RoleSize                    string                        `xml:",omitempty"`                                         // Specifies the size of the Virtual Machine. The default size is Small.
+	ProvisionGuestAgent         bool                          `xml:",omitempty"`                                         // Indicates whether the VM Agent is installed on the Virtual Machine. To run a resource extension in a Virtual Machine, this service must be installed.
+	VMImageInput                *VMImageInput                 `xml:",omitempty"`                                         // When a VM Image is used to create a new PersistentVMRole, the DiskConfigurations in the VM Image are used to create new Disks for the new VM. This parameter can be used to resize the newly created Disks to a larger size than the underlying DiskConfigurations in the VM Image.
 
-type RoleInstanceList struct {
-	RoleInstance []*RoleInstance
+	UseCertAuth bool   `xml:"-"`
+	CertPath    string `xml:"-"`
```
ahmetalpbalkan: I think these stuff should not be here unless they're going to be serialized. If these are fields that are used in utility methods, looks like there's a separation of concerns problem here.

management/virtualmachine/entities.go
```diff
@@ -11,132 +12,235 @@ type VirtualMachineClient struct {
...
+	WinRMListeners            *WinRMListener        `xml:"WinRM>Listeners>Listener,omitempty"`                           // Optional. Contains configuration settings for the Windows Remote Management service on the Virtual Machine. This enables remote Windows PowerShell.
+	AdditionalUnattendContent string                `xml:",omitempty"`                                                   // Specifies additional base-64 encoded XML formatted information that can be included in the Unattend.xml file, which is used by Windows Setup.
+
+	// Linux provisioning:
+	HostName     string `xml:",omitempty"` // Required. Specifies the host name for the Virtual Machine. Host names must be 1 to 64 characters long.
+	UserName     string `xml:",omitempty"` // Required. Specifies the name of a user account to be created in the sudoer group of the Virtual Machine. User account names must be 1 to 32 characters long.
+	UserPassword string `xml:",omitempty"` // Required. Specifies the password for the user account. Passwords must be 6 to 72 characters long.
+	// issue-paulmey: can't make this bool-omitempty, because default is true
```
ahmetalpbalkan: safe to remove this comment

management/virtualmachine/entities.go
```diff
@@ -11,132 +12,235 @@ type VirtualMachineClient struct {
...
 }
 
-type ResourceExtensionParameter struct {
-	Key   string
-	Value string
-	Type  string
+// Specifies the credentials to use to join the Virtual Machine to the domain.
+type Credentials struct {
+	Domain string // Specifies the name of the domain used to authenticate an account. The value is a fully qualified DNS domain.
+	// If the domains name is not specified, Username must specify the user principal name (UPN) format (user@fully-qualified-DNS-domain) or the fully-qualified-DNS-domain\username format.
```
ahmetalpbalkan: belongs to which field? not legit godoc

management/virtualmachine/entities.go
```diff
@@ -11,132 +12,235 @@ type VirtualMachineClient struct {
...
+	KeyPairs   *[]KeyPair   `xml:"KeyPairs>KeyPair"`
 }
 
-type PublicKeyList struct {
-	PublicKey []PublicKey
+// Specifies a public SSH key.
+type PublicKey struct {
+	Fingerprint string // Specifies the SHA1 fingerprint of an X509 certificate associated with the cloud service and includes the SSH public key.
+	Path        string // Usually /home/username/.ssh/authorized_keys
+	// Specifies the full path of a file, on the Virtual Machine, where the SSH public key is stored. If
```
ahmetalpbalkan: try running godoc and see if this comment indeed goes to `Path`

management/virtualmachine/entities.go
paulmey: fixed in #65

management/virtualmachine/entities.go
paulmey: I'll move it to the type

management/virtualmachine/entities.go
paulmey: moved multiline comments before fields.

ahmetalpbalkan:
next time let's limit comments to max 150 lines of diff. it's hard to read like this.

management/virtualmachine/entities.go
ahmetalpbalkan: @paulmey umm but am I supposed to merge like this or not? :) that's why small and atomic commits are always good.

paulmey:
Sure, next time I'll break it up into smaller PR's. Thanks

ahmetalpbalkan:
@paulmey doesn't have to be separate PR. multi-commit PRs are easy to review since if you'rer refactoring N things not related to each other with different purposes, it makes sense to split it into smaller commits with messages describing the purpose.

management/virtualmachine/client.go
```diff
@@ -718,7 +719,7 @@ func (self VirtualMachineClient) createNetworkConfig(os string, sshPort int) (Co
 func (self VirtualMachineClient) createEndpoint(name string, protocol string, extertalPort int, internalPort int) InputEndpoint {
 	endpoint := InputEndpoint{}
 	endpoint.Name = name
-	endpoint.Protocol = protocol
+	endpoint.Protocol = InputEndpointProtocol(protocol)
```
ahmetalpbalkan: change the signature instead?

management/virtualmachine/entities.go
paulmey: Yes, please merge like this. This is not new code, it's the inability of git to match this diff properly. See https://github.com/MSOpenTech/azure-sdk-for-go/blob/master/management/virtualmachine/entities.go#L65-L66 .

I'm not convinced that this PR would have made sense in smaller pieces, but I do agree that small commits are easier to review. Would you have preferred to review this PR in smaller commits and then have it squashed and rebased before merging?

management/virtualmachine/client.go
paulmey: ah yes, good suggestion

paulmey:
Ah ok, I was under the impression that single commit PR's are preferred by most projects. :+1: Thanks for that info, I'll apply it in the future... :-)

management/virtualmachine/client.go
```diff
@@ -505,19 +504,18 @@ func (self VirtualMachineClient) addDockerPort(configurationSets []Configuration
...
-func (self VirtualMachineClient) createVMDeploymentConfig(role *Role) VMDeployment {
-	deployment := VMDeployment{}
+func (self VirtualMachineClient) createVMDeploymentConfig(role *Role) Deployment {
+	deployment := Deployment{}
 	deployment.Name = role.RoleName
-	deployment.Xmlns = azureXmlns
 	deployment.DeploymentSlot = "Production"
 	deployment.Label = role.RoleName
-	deployment.RoleList.Role = append(deployment.RoleList.Role, role)
+	deployment.RoleList = append(deployment.RoleList, role)
```
ahmetalpbalkan: doesn't this append look a bit weird to you?

also we should probably use struct initializer here.

paulmey: changed to struct initializer + some others as well

ahmetalpbalkan:
@paulmey compilation fails `management/virtualmachine/client.go:537: undefined: oSVirtualHardDisk
`

paulmey:
@ahmetalpbalkan ai...sorry, that should not have happened...

management/virtualmachine/client.go
```diff
@@ -461,27 +460,24 @@ func (self VirtualMachineClient) ResolveRoleSize(roleSizeName string) error {
...
 
 func (self VirtualMachineClient) createShutdowRoleOperation() ShutdownRoleOperation {
-	shutdownRoleOperation := ShutdownRoleOperation{}
-	shutdownRoleOperation.OperationType = "ShutdownRoleOperation"
-	shutdownRoleOperation.Xmlns = azureXmlns
-
-	return shutdownRoleOperation
+	return ShutdownRoleOperation{
+		OperationType: "ShutdownRoleOperation",
+		Xmlns:         azureXmlns,
```
ahmetalpbalkan: is this XmlNs thing really necessary to correctly make the request?

paulmey: I don't think so. I'll remove all of them...

ahmetalpbalkan: are we testing this manually by any chance? Am I safe to merge it? :-) 


On Monday, Mar 16, 2015 at 3:49 PM, Paul Meyer <notifications@github.com>, wrote:

In management/virtualmachine/client.go:

>  }
>  
>  func (self VirtualMachineClient) createShutdowRoleOperation() ShutdownRoleOperation {
> -	shutdownRoleOperation := ShutdownRoleOperation{}
> -	shutdownRoleOperation.OperationType = "ShutdownRoleOperation"
> -	shutdownRoleOperation.Xmlns = azureXmlns
> -
> -	return shutdownRoleOperation
> +	return ShutdownRoleOperation{
> +		OperationType: "ShutdownRoleOperation",
> +		Xmlns:         azureXmlns,

I don't think so. I'll remove all of them...

paulmey:
okay, wait with the merge, I found some bugs...

paulmey:
Fixed some namespace errors. Added the namespace to the XMLName property, which works as documented.
Also added another commit to pull apart the Deployment type for creating and looking at a deployment as they contain overlapping but distinct sets of properties.

paulmey:
Should be good to merge now. #65 builds on this and adds some integration test coverage.

ahmetalpbalkan:
@paulmey merged this, please rebase #65 so that it has newer commits on top of the merge commit.

PR #66, opened 2015-03-18
Fix 'Create storage service'

paulmey:
AccountType property replaces GeoReplicationEnabled and SecondaryReadEnabled

### DIFF
--- a/management/storageservice/client.go
+++ b/management/storageservice/client.go
@@ -1,7 +1,6 @@
 package storageservice
 
 import (
-	"encoding/base64"
 	"encoding/xml"
 	"errors"
 	"fmt"
@@ -84,21 +83,18 @@ func (self StorageServiceClient) GetStorageServiceByLocation(location string) (*
 	return nil, nil
 }
 
-func (self StorageServiceClient) CreateStorageService(name, location string) (*StorageService, error) {
-	if name == "" {
-		return nil, fmt.Errorf(errParamNotSpecified, "name")
-	}
-	if location == "" {
-		return nil, fmt.Errorf(errParamNotSpecified, "location")
-	}
-
-	storageDeploymentConfig := self.createStorageServiceDeploymentConf(name, location)
-	deploymentBytes, err := xml.Marshal(storageDeploymentConfig)
+func (self StorageServiceClient) CreateAsync(parameters StorageAccountCreateParameters) (string, error) {
+	data, err := xml.Marshal(CreateStorageServiceInput{
+		StorageAccountCreateParameters: parameters})
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
-	requestId, err := self.client.SendAzurePostRequest(azureStorageServiceListURL, deploymentBytes)
+	return self.client.SendAzurePostRequest(azureStorageServiceListURL, data)
+}
+
+func (self StorageServiceClient) Create(parameters StorageAccountCreateParameters) (*StorageService, error) {
+	requestId, err := self.CreateAsync(parameters)
 	if err != nil {
 		return nil, err
 	}
@@ -108,7 +104,7 @@ func (self StorageServiceClient) CreateStorageService(name, location string) (*S
 		return nil, err
 	}
 
-	storageService, err := self.GetStorageServiceByName(storageDeploymentConfig.ServiceName)
+	storageService, err := self.GetStorageServiceByName(parameters.ServiceName)
 	if err != nil {
 		return nil, err
 	}
@@ -126,18 +122,6 @@ func (self StorageServiceClient) GetBlobEndpoint(storageService *StorageService)
 	}
 
 	return "", errors.New(fmt.Sprintf(errBlobEndpointNotFound, storageService.ServiceName))
-}
-
-func (self *StorageServiceClient) createStorageServiceDeploymentConf(name, location string) StorageServiceDeployment {
-	storageServiceDeployment := StorageServiceDeployment{}
-
-	storageServiceDeployment.ServiceName = name
-	label := base64.StdEncoding.EncodeToString([]byte(name))
-	storageServiceDeployment.Label = label
-	storageServiceDeployment.Location = location
-	storageServiceDeployment.Xmlns = azureXmlns
-
-	return storageServiceDeployment
 }
 
 // The Check Storage Account Name Availability operation checks to see if the specified storage account name is available, or if it has already been taken.
--- a/management/storageservice/entities.go
+++ b/management/storageservice/entities.go
@@ -33,18 +33,30 @@ type StorageServiceProperties struct {
 	GeoPrimaryRegion      string
 }
 
-type StorageServiceDeployment struct {
-	XMLName               xml.Name `xml:"CreateStorageServiceInput"`
-	Xmlns                 string   `xml:"xmlns,attr"`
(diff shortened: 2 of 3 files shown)


management/storageservice/entities.go
```diff
@@ -34,18 +34,26 @@ type StorageServiceProperties struct {
...
+	AffinityGroup      string `xml:",omitempty"`
+	Location           string `xml:",omitempty"`
+	ExtendedProperties ExtendedPropertyList
+	AccountType        AccountType
 }
 
+type AccountType string
+
+const (
+	AccountTypeStandard_LRS   AccountType = "Standard_LRS"
```
ahmetalpbalkan: why underscores?

management/storageservice/client.go
```diff
@@ -92,7 +92,7 @@ func (self StorageServiceClient) CreateStorageService(name, location string) (*S
 		return nil, fmt.Errorf(errParamNotSpecified, "location")
 	}
 
-	storageDeploymentConfig := self.createStorageServiceDeploymentConf(name, location)
+	storageDeploymentConfig := self.createStorageServiceDeploymentConf(name, location, AccountTypeStandard_LRS)
```
ahmetalpbalkan: I'm -1 for this. 

CreateStorageService should accept this as an output. or we should just accept a `CreateStorageServiceParameters` struct as an input to this method. This approach doesn't scale and will keep breaking.

management/storageservice/client.go
```diff
@@ -128,14 +128,14 @@ func (self StorageServiceClient) GetBlobEndpoint(storageService *StorageService)
 	return "", errors.New(fmt.Sprintf(errBlobEndpointNotFound, storageService.ServiceName))
 }
 
-func (self *StorageServiceClient) createStorageServiceDeploymentConf(name, location string) StorageServiceDeployment {
+func (self *StorageServiceClient) createStorageServiceDeploymentConf(name, location string, accountType AccountType) StorageServiceDeployment {
```
ahmetalpbalkan: -1 for this helper method's existence as well. Ideally user should provide this or we should have one helper like `func NewCreateStorageAccountParameters(name, location string) CreateStorageAccountParameters` and pass that to `CreateStorageService()`

management/storageservice/entities.go
paulmey: Copy/paste from the string value. I'll remove them from the const names.

management/storageservice/client.go
paulmey: The code that was there never touched the properties that were deprecated, serialized the 'false' values anyway. (API returns 400)
The StorageServiceDeployment struct is simple enough, shall I just make that the input of CreateStorageService?

management/storageservice/client.go
paulmey: +1 for ripping this method out completely. See comment above.

ahmetalpbalkan: I guess I can merge like this and address these in a separate PR. General approach for 1:1 mapping between methods and REST API endpoints should be taking all parameters as a struct argument and all URL path components as arguments... 

e.g. create hosted service would definitely need name, so you'd have `name` in method signature but all other details would go into a `parameters CreateHostedServiceParameters` struct. Check C# SDK for examples. I think it is a great model and you can add new things without breaking the back compat on method signature.

paulmey:
Modelled after https://github.com/Azure/azure-sdk-for-net/blob/9f92179070f8f1429920c25bc82a21767deb95de/src/StorageManagement/Generated/IStorageAccountOperations.cs#L97

management/storageservice/client.go
```diff
@@ -84,21 +84,31 @@ func (self StorageServiceClient) GetStorageServiceByLocation(location string) (*
 	return nil, nil
 }
 
-func (self StorageServiceClient) CreateStorageService(name, location string) (*StorageService, error) {
-	if name == "" {
-		return nil, fmt.Errorf(errParamNotSpecified, "name")
+func (self StorageServiceClient) CreateAsync(parameters StorageAccountCreateParameters) (string, error) {
+	if parameters.ServiceName == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "parameters.ServiceName")
```
ahmetalpbalkan: hehe what if I told you all these stuff are not necessary anymore. :) same is true for the entire library though. azure will just tell you "ServiceName" is missing if you leave required ones empty.

by having those guys here, we're actually duplicating the knowledge of what's required and what's not in both API service code and Library. I think this guy should be stupid and shouldn't know what's required and what's not. We expect users to get these required errors from the API and fix their mistakes. What if tomorrow Label becomes optional in a newer API version? Now you need to update the library.

Let's start with this little method here and remove all those checks.

management/storageservice/entities.go
```diff
@@ -33,19 +33,41 @@ type StorageServiceProperties struct {
...
-	ServiceName           string
-	Description           string
-	Label                 string
-	AffinityGroup         string `xml:",omitempty"`
-	Location              string `xml:",omitempty"`
-	GeoReplicationEnabled bool
-	ExtendedProperties    ExtendedPropertyList
-	SecondaryReadEnabled  bool
+type CreateStorageServiceInput struct {
+	XMLName xml.Name `xml:"http://schemas.microsoft.com/windowsazure CreateStorageServiceInput"`
```
ahmetalpbalkan: I'm not sure if this annotation notation is correct.

management/storageservice/entities.go
```diff
@@ -33,19 +33,41 @@ type StorageServiceProperties struct {
...
+	// 100 characters in length. The label can be used identify the storage account for your tracking purposes.
+	Label string
+	// Required if Location is not specified. The name of an existing affinity group in the specified subscription.
+	AffinityGroup string `xml:",omitempty"`
+	// Required if Location is not specified. The name of an existing affinity group in the specified subscription.
+	Location           string `xml:",omitempty"`
+	ExtendedProperties ExtendedPropertyList
+	// Specifies whether the account supports locally-redundant storage, geo-redundant storage, zone-redundant storage,
+	// or read access geo-redundant storage.
+	AccountType AccountType
```
ahmetalpbalkan: this shouldn't appear at the end. ExtendedProperties should.

management/storageservice/entities.go
```diff
@@ -33,19 +33,41 @@ type StorageServiceProperties struct {
...
+
+type StorageAccountCreateParameters struct {
+	// Required. A name for the storage account that is unique within Azure. Storage account names must be
+	// between 3 and 24 characters in length and use numbers and lower-case letters only.
+	// This name is the DNS prefix name and can be used to access blobs, queues, and tables in the storage account.
+	ServiceName string
+	// Optional. A description for the storage account. The description may be up to 1024 characters in length.
+	Description string `xml:",omitempty"`
+	// Required. A label for the storage account specified as a base64-encoded string. The label may be up to
+	// 100 characters in length. The label can be used identify the storage account for your tracking purposes.
```
ahmetalpbalkan: I think comments are unnecessary. You're duplicating the RDFE contract here. Unless you auto-generate the code from the spec, if the requirement tomorrow becomes 90 chars max, your comment is out of sync and you can't just manually maintain that. Let's clear these commands.

ahmetalpbalkan:
LGTM thanks. build got stuck but it compiles. so merging.

PR #67, opened 2015-03-18
Take HTTP proxy from environment

paulmey:
Allows usage of tools like Fiddler for debugging. This is the standard
way of determining the proxy.


ahmetalpbalkan:
LGTM

PR #69, opened 2015-03-19
Use container build infra on Travis CI

paulmey:
See http://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details. Will make builds start faster. Also, Travis seems to have more capacity for container based builds than for VM based (capped at 500?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 language: go
 go: tip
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
- sends 10 char change
- adds a paragraph explanation

:stuck_out_tongue: 

LGTM

PR #70, opened 2015-03-20
Fix a bug in the putBlockBlob function

plutoshe:
In the putBlockBlob function of storage/blob.go, the original code could not achieve to one block state

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -535,6 +535,14 @@ func (b BlobStorageClient) putBlockBlob(container, name string, blob io.Reader, 
 	}
 
 	chunk := make([]byte, chunkSize)
+	preChunk := make([]byte, chunkSize)
+	pren, err := blob.Read(preChunk)
+	if err != nil && err != io.EOF {
+		return err
+	}
+	if err == io.EOF {
+		return fmt.Errorf("Need blob content")
+	}
 	n, err := blob.Read(chunk)
 	if err != nil && err != io.EOF {
 		return err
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@plutoshe can you please explain what was the bug? Maybe write a test case to show it was broken? I am not sure what have you fixed here.

plutoshe:
To fit the logic of 
        if err == io.EOF {
		// Fits into one block
		return b.putSingleBlockBlob(container, name, chunk[:n])
	} else {
		// Does not fit into one block. Upload block by block then commit the block list"
        }

when encounter with one block, we should use putSingleBlockBlob function.  But it will commit the block list when only one block because it couldn't get io.EOF if the data isn't empty.

ahmetalpbalkan:
@plutoshe ok here's a better idea: Let's remove `putBlockBlob` methods entirely. We will only support PutBlock, PutBlockList and let the caller design the algorithm they want. Because we generate some block IDs here and for example you cannot call `PutBlockBlob()` twice on the same blob because it commits block list and the next call reuses the same block IDs again.

I will remove these methods and come up with a package called `blobhelper` to have such convenience methods. Thanks for the suggestion but I'm going to close this PR in favor of removing those methods entirely. Please use `PutBlock` and `PutBlockList`/`GetBlockList` to unblock yourself.

plutoshe:
I also have a problem about append Uncommitted Block to Blob.
The code like follow
cli.PutBlockBlob(cnt, blob, New.string(aData))
cli.PutBlock(cnt,blob,PutBlock, cnt, blob, bBlockID, bData)
blockList, err := cli.GetBlockList(cnt,blob, BlockListTypeAll)
var newBlockList []Block
newBlockList = <<<combine the Committed Block and Uncommitted Block>>>
cli.PutBlockList(cnt,blob,newBlockList)

I want to append UncommittedBlocks. But I encounter a error "storage: remote server returned error. StatusCode=400, ErrorCode=InvalidBlockList, ErrorMessage=The specified block list is invalid."

However, I change the cli.PutBlockBlob function to cli.PutBlock and cli.PutBlockList, It works.

I confused about how to fix it. And I ensure the length of the BlockID is same.

ahmetalpbalkan:
@plutoshe yes it's probably because `PutBlockBlob` method creates a block ID of size 11 bytes (in ASCII) (a different length when converted to base64) and the block ID you're using is of a different length and API requires all block IDs to be of the same length.

If you're 100% sure block ID length is the same, I'd say, double check. Most likely either the block IDs are not of the same length (in bytes) or you're using wrong block IDs or statuses (committed/latest/uncommitted). 

For such reasons `PutBlockBlob` is evil, it has special logic and it not really playing well with other methods you have. Please always use PutBlock/PutBlockList, that way your problem is likely to go away. That's why I am closing this issue.

plutoshe:
Thanks a lot for your help, I will check my code.

PR #75, opened 2015-03-24
Fix: NewVmConfiguration never returns error

paulmey:
Might as well simplify...

### DIFF
--- a/management/vmutils/vmutils.go
+++ b/management/vmutils/vmutils.go
@@ -12,13 +12,13 @@ const (
 )
 
 // Creates configuration for a new virtual machine Role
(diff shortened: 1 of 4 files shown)


management/vmutils/vmutils.go
```diff
@@ -12,13 +12,13 @@ const (
 )
 
 // Creates configuration for a new virtual machine Role
-func NewVmConfiguration(name string, rolesize string) (Role, error) {
+func NewVmConfiguration(name string, rolesize string) Role {
```
ahmetalpbalkan: rolesize not properly camelCased

paulmey: :+1:

PR #76, opened 2015-03-24
Make deleting media optional for hostedservice.DeleteHostedService

paulmey:
Also give async control to caller


management/hostedservice/client.go
```diff
@@ -84,18 +84,16 @@ func (self HostedServiceClient) CheckHostedServiceNameAvailability(dnsName strin
 	return availabilityResponse.Result, availabilityResponse.Reason, nil
 }
 
-func (self HostedServiceClient) DeleteHostedService(dnsName string) error {
+func (self HostedServiceClient) DeleteHostedService(dnsName string, deleteDisksAndBlobsToo bool) (string, error) {
```
ahmetalpbalkan: `Too` part is weird but I'll keep it :)

PR #79, opened 2015-03-24
Move GetTestClient from vmutils to testutils

paulmey:
so it can be used in other SDK package tests. (preparing some tests in the virtualmachine package)


ahmetalpbalkan:
LGTM

PR #80, opened 2015-03-24
Add vm.GetResourceExtensions

paulmey:


management/virtualmachine/resourceextensions_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"flag"
+	"testing"
+
+	"github.com/MSOpenTech/azure-sdk-for-go/management/testutils"
+)
+
+func init() {
+	flag.BoolVar(&extraVerbose, "resourceextensions.verbose", false, "List all resource extensions")
```
ahmetalpbalkan: This is not good practice for test pkgs. test logs don't get printed unless -test.v exists. Make use of that if u can

management/virtualmachine/resourceextensions.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"encoding/xml"
+)
+
+const (
+	azureResourceExtensionsURL = "services/resourceextensions"
+)
+
+func (self VirtualMachineClient) GetResourceExtensions() (extensions []ResourceExtension, err error) {
+	// from https://msdn.microsoft.com/en-us/library/azure/dn495441.aspx
```
ahmetalpbalkan: Move to godoc (see "storage")

management/virtualmachine/resourceextensions_test.go
paulmey: I'd like to have an extra level of verbosity. Don't want to pollute the output with all resource extensions, but it is very handy to look things up. I can remove all of that code and people can put the `t.Logf("%+v", extension)` if they want it...

ahmetalpbalkan: @paulmey usually tests are ran without `-test.v`, and when it fails, all t.Log output is also printed (even without `-test.v`). That's why nobody probably needs an extra level of verbosity.

paulmey: No need for a lecture, I know all that. Likewise, if there was an extra level of verbosity, people would use it. I had fun figuring out how to add a flag, so I'm good. :-) Code is gone and this diff is outdated.

ahmetalpbalkan: It's just weird to have such undocumented params, especially for tests. That was mainly my comment.

paulmey: But it is very documented... that's exactly what the fourth parameter of [BoolVar](http://golang.org/pkg/flag/#BoolVar) does package does. Now merge it already if you're happy with 4440be0! :smile:

ahmetalpbalkan: @paulmey should I be running `go test . --help` to discover that flag? asking just out of curiosity.

paulmey: No, I haven't been able to get `go test` to list all flags for the package under test. You have to build the test binary with `go test -c .` and then execute the test binary with --help (`virtualmachine.test.exe --help`). I think this makes the ability for tests to declare extra flags very hard to discover and in most cases not the best practice (as you pointed out).

ahmetalpbalkan:
LGTM :shipit:

PR #81, opened 2015-03-25
Add storageservice.GetStorageServiceKeys

paulmey:


management/storageservice/client.go
```diff
@@ -83,6 +83,26 @@ func (self StorageServiceClient) GetStorageServiceByLocation(location string) (*
 	return nil, nil
 }
 
+func (self StorageServiceClient) GetStorageServiceKeys(serviceName string) (primary, secondary string, err error) {
```
ahmetalpbalkan: I would favor an implementation like:

```go
type GetStorageServiceKeysResponse struct { PrimaryKey, SecondaryKey string } 
```

and returning that form here.

paulmey: :+1:

management/storageservice/client.go
```diff
@@ -83,6 +83,26 @@ func (self StorageServiceClient) GetStorageServiceByLocation(location string) (*
...
+
+	requestURL := fmt.Sprintf(azureStorageServiceURL, serviceName)
+	data, err := self.client.SendAzureGetRequest(requestURL)
+	if err != nil {
+		return GetStorageServiceKeysResponse{}, err
+	}
+
+	response := GetStorageServiceKeysResponse{}
+	err = xml.Unmarshal(data, response)
+	if err != nil {
```
ahmetalpbalkan: if you look closely this `if` is unnecessary

ahmetalpbalkan:
LGTM

PR #85, opened 2015-03-26
Introduce Azure OperationId type

paulmey:
@ahmetalpbalkan Is this what you meant before in a PR? (can't find the exact comment)

### DIFF
--- a/management/http.go
+++ b/management/http.go
@@ -34,7 +34,7 @@ func (client *Client) SendAzureGetRequest(url string) ([]byte, error) {
 
 //sendAzurePostRequest sends a request to the management API using the HTTP POST method
 //and returns the request ID or an error.
-func (client *Client) SendAzurePostRequest(url string, data []byte) (string, error) {
+func (client *Client) SendAzurePostRequest(url string, data []byte) (OperationId, error) {
 	if url == "" {
 		return "", fmt.Errorf(errParamNotSpecified, "url")
 	}
@@ -44,14 +44,13 @@ func (client *Client) SendAzurePostRequest(url string, data []byte) (string, err
 		return "", err
 	}
 
-	requestId := response.Header[requestIdHeader]
-	return requestId[0], nil
+	return getOperationId(response)
 }
 
 //sendAzurePutRequest sends a request to the management API using the HTTP PUT method
 //and returns the request ID or an error. The content type can be specified, however
 //if an empty string is passed, the default of "application/xml" will be used.
-func (client *Client) SendAzurePutRequest(url string, contentType string, data []byte) (string, error) {
+func (client *Client) SendAzurePutRequest(url string, contentType string, data []byte) (OperationId, error) {
 	if url == "" {
 		return "", fmt.Errorf(errParamNotSpecified, contentType, "url")
 	}
@@ -61,13 +60,12 @@ func (client *Client) SendAzurePutRequest(url string, contentType string, data [
 		return "", err
 	}
 
-	requestId := response.Header[requestIdHeader]
-	return requestId[0], nil
+	return getOperationId(response)
 }
 
 //sendAzureDeleteRequest sends a request to the management API using the HTTP DELETE method
 //and returns the request ID or an error.
-func (client *Client) SendAzureDeleteRequest(url string) (string, error) {
+func (client *Client) SendAzureDeleteRequest(url string) (OperationId, error) {
 	if url == "" {
 		return "", fmt.Errorf(errParamNotSpecified, "url")
 	}
@@ -77,8 +75,15 @@ func (client *Client) SendAzureDeleteRequest(url string) (string, error) {
 		return "", err
 	}
 
+	return getOperationId(response)
+}
+
+func getOperationId(response *http.Response) (OperationId, error) {
 	requestId := response.Header[requestIdHeader]
-	return requestId[0], nil
+	if len(requestId) == 0 {
+		return "", fmt.Errorf("Could not retrieve operation id from %q header", requestIdHeader)
+	}
+	return OperationId(requestId[0]), nil
 }
 
 //sendAzureRequest constructs an HTTP client for the request, sends it to the
--- a/management/operations.go
+++ b/management/operations.go
@@ -2,28 +2,36 @@ package management
 
 import (
 	"encoding/xml"
-	"errors"
 	"fmt"
 	"time"
 )
 
-//operation represents an in-flight operation. Use client.getOperationStatus()
-//to get the operation given the operation ID, or use
-//client.waitAsyncOperation() to block until the operation has completed.
-type operation struct {
-	XMLName        xml.Name `xml:"Operation"`
+//  GetOperationStatusResponse  represents an in-flight operation. Use client.GetOperationStatus()
+// to get the operation given the operation ID, or use
+// client.WaitAsyncOperation() to block until the operation has completed.
+// See https://msdn.microsoft.com/en-us/library/azure/ee460783.aspx
+type GetOperationStatusResponse struct {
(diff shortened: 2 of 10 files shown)


ahmetalpbalkan:
Yes exactly. Except it should be called OperationId. 

We can provide WaitForOperationStatus helper method which supports cancellation with signaling channels, reports progress with chans and polls until status != InProgress. 

Actually even implementing polling itself is hard because GetOperationStatus fails intermittently too but we do not have retries anywhere so we can live with that. Of course all those are in scope of some other PR. 

Main theme of this PR should be replacing all string occurences in method signatures with this management.OperationId to make it explicit they're not just strings and all mean the same thing. 

I'll review once I am on desktop. Thanks.

management/http.go
```diff
@@ -62,12 +62,12 @@ func (client *Client) SendAzurePutRequest(url string, contentType string, data [
 	}
 
 	requestId := response.Header[requestIdHeader]
-	return requestId[0], nil
+	return RequestId(requestId[0]), nil
```
ahmetalpbalkan: actually this could panic if [0] doesn't exist. can we do a len check here and return just err if even header isn't present?

management/operations.go
```diff
@@ -18,14 +18,17 @@ type operation struct {
...
 }
 
+// RequestId is assigned by Azure API and can be used to look up the status of an operation
+type RequestId string
+
 //getOperationStatus gets an operation given the operation ID.
-func (client *Client) getOperationStatus(operationId string) (*operation, error) {
-	if operationId == "" {
-		return nil, fmt.Errorf(errParamNotSpecified, "operationId")
+func (client *Client) getOperationStatus(requestId RequestId) (*operation, error) {
```
ahmetalpbalkan: in some next PR we should consider exposing this method to user since it's a REST call.

management/operations.go
```diff
@@ -39,12 +42,12 @@ func (client *Client) getOperationStatus(operationId string) (*operation, error)
...
 
-//waitAsyncOperation blocks until the operation with the given operationId is
+//waitAsyncOperation blocks until the operation with the given requestId is
 //no longer in the InProgress state. If the operation was successful, nothing is
 //returned, otherwise an error is returned.
-func (client *Client) WaitAsyncOperation(operationId string) error {
-	if operationId == "" {
-		return fmt.Errorf(errParamNotSpecified, "operationId")
+func (client *Client) WaitAsyncOperation(requestId RequestId) error {
+	if requestId == "" {
```
ahmetalpbalkan: check redundant. getoperationstatus does the same

management/operations.go
```diff
@@ -52,7 +55,7 @@ func (client *Client) WaitAsyncOperation(operationId string) error {
 	err := errors.New("")
```
ahmetalpbalkan: var err error, although it can be declared below. so this line is not necessary

management/operations.go
ahmetalpbalkan: beyond scope of this PR but can be fixed here,too: this shouldn't return pointer.

management/operations.go
```diff
@@ -39,12 +42,12 @@ func (client *Client) getOperationStatus(operationId string) (*operation, error)
...
 }
 
-//waitAsyncOperation blocks until the operation with the given operationId is
+//waitAsyncOperation blocks until the operation with the given requestId is
 //no longer in the InProgress state. If the operation was successful, nothing is
 //returned, otherwise an error is returned.
-func (client *Client) WaitAsyncOperation(operationId string) error {
-	if operationId == "" {
-		return fmt.Errorf(errParamNotSpecified, "operationId")
+func (client *Client) WaitAsyncOperation(requestId RequestId) error {
```
ahmetalpbalkan: this param should be called `operationId`

management/operations.go
```diff
@@ -39,12 +42,12 @@ func (client *Client) getOperationStatus(operationId string) (*operation, error)
...
+	if requestId == "" {
+		return fmt.Errorf(errParamNotSpecified, "requestId")
 	}
 
 	status := "InProgress"
```
ahmetalpbalkan: damn this is a nasty hack.

ahmetalpbalkan:
Design LGTM, it'd be great if we improve code around WaitAsyncOperation/getOperationStatus. Thanks @paulmey. :clap: :pray:

paulmey:
Renamed to OperationId and also exposed/refactored GetOperationStatus

management/operations.go
```diff
@@ -2,67 +2,69 @@ package management
...
+	OperationStatusInProgress OperationStatus = "InProgress"
+	OperationStatusSucceeded  OperationStatus = "Succeeded"
+	OperationStatusFailed     OperationStatus = "Failed"
+)
+
+// OperationId is assigned by Azure API and can be used to look up the status of an operation
+type OperationId string
+
+// getOperationStatus gets an operation given the operation ID.
+func (client *Client) GetOperationStatus(operationId OperationId) (Operation, error) {
```
ahmetalpbalkan: this is rather `OperationResponse` but we already have a mix so that's fine.

management/operations.go
```diff
@@ -2,67 +2,69 @@ package management
...
-	}
-
-	if status == "Failed" {
-		return fmt.Errorf("Azure operation %s failed. Code: %s, Message: %s", operationId, operation.Error.Code, operation.Error.Message)
+		if operation.Status == OperationStatusFailed {
+			return fmt.Errorf("Azure operation %s failed. Code: %s, Message: %s", operationId, operation.Error.Code, operation.Error.Message)
+		} else if operation.Status != OperationStatusInProgress {
+			return nil
+		}
+		time.Sleep(2000 * time.Millisecond)
```
ahmetalpbalkan: you delay every successful response 2 seconds anyway. also this is too high. most SDKs use default 30s.

paulmey: Setting default to 30s and opened issue #89 to give the caller more control.

management/operations.go
```diff
@@ -2,67 +2,76 @@ package management
...
-	}
-
-	if status == "Failed" {
-		return fmt.Errorf("Azure operation %s failed. Code: %s, Message: %s", operationId, operation.Error.Code, operation.Error.Message)
+		switch operation.Status {
+		case OperationStatusSucceeded:
+			return nil
+		case OperationStatusInProgress:
+			time.Sleep(pollInterval)
+			break
```
ahmetalpbalkan: are you sure this is `break`?

management/operations.go
```diff
@@ -2,67 +2,76 @@ package management
...
-		time.Sleep(2000 * time.Millisecond)
-		operation, err = client.getOperationStatus(operationId)
+// WaitAsyncOperation blocks until the operation with the given operationId is
+// no longer in the InProgress state. If the operation was successful, nothing is
+// returned, otherwise an error is returned.
+func (client *Client) WaitAsyncOperation(operationId OperationId) error {
+	for {
+		operation, err := client.GetOperationStatus(operationId)
 		if err != nil {
 			return err
```
ahmetalpbalkan: this would be better if we say: `fmt.Errorf("Failed to get operation status '%s': %v", operationId, err)`

management/operations.go
```diff
@@ -2,67 +2,78 @@ package management
...
-	}
-
-	if status == "Failed" {
-		return fmt.Errorf("Azure operation %s failed. Code: %s, Message: %s", operationId, operation.Error.Code, operation.Error.Message)
+		switch operation.Status {
+		case OperationStatusSucceeded:
+			return nil
+		case OperationStatusFailed:
+			if operation.Error != nil {
+				return operation.Error
```
paulmey: This will allow us to do retries of async ops without doing string parsing to see what the Code was

ahmetalpbalkan: :+1:

ahmetalpbalkan:
LGTM

PR #86, opened 2015-03-26
Fix test bug and adjust CI

paulmey:
This bug was not caught by CI, because CI was not building (or executing) tests in ./management.

CI now builds everything, tests storage and management. management integration tests are set up to skip if no subscription is provided.


.travis.yml
```diff
@@ -3,5 +3,6 @@ sudo: false
 language: go
 go: tip
 script:
+    - go build -v ./...
     - go test -v ./storage/...
-    - go build -v ./management/...
+    - go test -v ./management/...
```
ahmetalpbalkan: doesn't this now mean `go test -test.v ./...` ?

ahmetalpbalkan:
actually it's better to skip integration tests with `-test.short` flag but not providing subscription apparently gives the same behavior here. 

LGTM.

.travis.yml
paulmey: the ./core dir also has some tests... not all of them succeed... didn't look why they fail.

ahmetalpbalkan: ah ok

PR #91, opened 2015-03-26
Fix integration_test

paulmey:
Role operations were changed to async, so now have to waited on.

### DIFF
--- a/management/vmutils/integration_test.go
+++ b/management/vmutils/integration_test.go
@@ -74,7 +74,15 @@ func TestDeployPlatformCaptureRedeploy(t *testing.T) {
 	}
 
 	t.Logf("Shutting down VM: %s", vmname)
-	vmc.ShutdownRole(vmname, vmname, vmname)
+	if err := Await(client, func() (management.OperationId, error) {
+		return vmc.ShutdownRole(vmname, vmname, vmname)
+	}); err != nil {
+		t.Error(err)
+	}
+
+	if err := WaitForDeploymentPowerState(client, vmname, vmname, vm.PowerStateStopped); err != nil {
+		t.Fatal(err)
+	}
 
 	imagename := GenerateName()
 	t.Logf("Capturing VMImage: %s", imagename)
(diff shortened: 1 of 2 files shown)


management/vmutils/integration_test.go
```diff
@@ -74,7 +74,22 @@ func TestDeployPlatformCaptureRedeploy(t *testing.T) {
...
+
+	for {
+		deployment, err := vmc.GetDeployment(vmname, vmname)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if deployment.RoleInstanceList[0].PowerState == vm.PowerStateStopped {
+			break
+		}
+		time.Sleep(2 * time.Second)
```
ahmetalpbalkan: Create helper WaitForRoleState. 

I used to have it another project checked in some internal sd repo and it helped a great deal. :-) actually you can make it per deployment and get all role states and compare against goal state:

WaitForRolesState(deploymentId, roleState).

paulmey: Done, but for power state, since that's what I need. We can add a method for RoleStatus as well? (in a separate PR?)

management/vmutils/rolestate.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"time"
+
+	"github.com/MSOpenTech/azure-sdk-for-go/management"
+	vm "github.com/MSOpenTech/azure-sdk-for-go/management/virtualmachine"
+)
+
+// Blocks until all role instances in deployment to reach desired power state
+func WaitForDeploymentPowerState(client management.Client, cloudServiceName, deploymentName string, desiredPowerstate vm.PowerState) error {
```
ahmetalpbalkan: @paulmey this does not timeout or break at all, it's too dangerous for a user to call it. can you add a `timeout time.Duration`

also can you make the polling interval 30s?

ahmetalpbalkan: why does anyone care about `PowerState`? I thought people care more about Role State.

paulmey: RoleStatus can be 'StoppedVM' with PowerState still 'Started' until the VM is finally turned off. I have no idea why there are 2 statuses, but experimentally, I've come to trust PowerState. Since there seem to be no other users, I serve myself... :smile:

ahmetalpbalkan: @paulmey I see. Yeah PowerState is not the thing users go for usually. It'd be great to actually watch for RoleState.

paulmey: I need PowerState for my purpose. I can create another PR for DeploymentStatus and role InstanceStatus (there is nothing called RoleState in the SDK?)

ahmetalpbalkan: Sure we're fine here then.

PR #95, opened 2015-04-01
Fix DeleteHostedService

paulmey:
comp=media is added by func, depending on parameter


ahmetalpbalkan:
Nice catch.

paulmey:
Yeah, the awesome thing is that RDFE says "yay, success!", but in fact nothing happened...

PR #96, opened 2015-04-01
Add ListHostedServices

paulmey:


management/hostedservice/client.go
```diff
@@ -117,6 +117,19 @@ func (self HostedServiceClient) GetHostedService(name string) (HostedService, er
...
+	var response listHostedServiceResponse
+
+	data, err := self.client.SendAzureGetRequest(azureHostedServiceListURL)
+	if err != nil {
+		return response.HostedServices, err
+	}
+
+	err = xml.Unmarshal(data, &response)
+
+	return response.HostedServices, err
```
ahmetalpbalkan: @paulmey ideally we want to return `ListHostedServiceResponse` which could be a typedef for `[]HostedService`. 

What they do in c# in, `class ListHostedServiceResponse : IList<HostedService>` however the interface is implemented through a field on `response.HostedServices` like you have in here.

I'm trying to see if we can avoid such internal fields + if tomorrow ListHostedServices returns a response along with the list (I don't know but not hard to imagine this happening) then the signature will need to change and you'll break people.

That's why we probably want to return `ListHostedServiceResponse` type here (and in many other places) :smile_cat:

paulmey: So when you do this:
```go
ss, _ = hsclient.ListHostedServices()
for _, s := range ss {
  // ... do something interesting with s
}
```
Won't that break if you change it form `type ListHostedServiceResponse []HostedService` to `type ListHostedServiceResponse struct`...

So then we'd want `ListHostedServiceResponse ` to be a `struct` and have the caller always dig into the struct, right?

ahmetalpbalkan: @paulmey yep.

paulmey:
Updated to return the struct instead of the inner slice.

management/hostedservice/client.go
```diff
@@ -117,6 +117,19 @@ func (self HostedServiceClient) GetHostedService(name string) (HostedService, er
...
+func (self HostedServiceClient) ListHostedServices() (ListHostedServiceResponse, error) {
+	var response ListHostedServiceResponse
+
+	data, err := self.client.SendAzureGetRequest(azureHostedServiceListURL)
+	if err != nil {
+		return response, err
+	}
+
+	err = xml.Unmarshal(data, &response)
+
```
ahmetalpbalkan: I find these spaces usually redundant but no big deal.

ahmetalpbalkan:
LGTM

PR #98, opened 2015-04-06
Fix omit empty in hosted service entities

paulmey:
These fields were _named_ omitempty instead of being omitted when empty.


ahmetalpbalkan:
LGTM.

ahmetalpbalkan:
LGTM

PR #101, opened 2015-04-08
Enable update role scenario for resource extensions

paulmey:
To only update the resource extensions for a role, the update role operation needs to post the Role object with just the resource extensions tag filled out. Presence of an empty role name tag is interpreted as trying to change the role name and hence will lead to an error.


ahmetalpbalkan:
LGTM.

PR #106, opened 2015-04-29
Remove blob helper methods from storage package

### LINKED ISSUE #71: storage/blob: Remove PutBlockBlob
`putBlockBlob`/`PutBlockBlob` methods are convenience methods with special logic and we shouldn't have those in the same package (maybe move to `blobhelper`).

ahmetalpbalkan:
Removing various blob upload convenience methods that maintain some
internal logic that makes it hard to operate on these blobs afterwards.
I plan to refactor and move such methods to a `blobutil` package later
on. Fixes #71. 

review: @paulmey

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -2,7 +2,6 @@ package storage
 
 import (
 	"bytes"
-	"encoding/base64"
 	"encoding/xml"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


paulmey:
LGTM. When do you plan to create this blobutil package?

ahmetalpbalkan:
@paulmey sometime lol.

nsilberman:
@ahmetalpbalkan @paulmey what would be great with this blobutil package would be : 
- divide a file in multiple chunks (exactly what you remove from the previous commits)
- upload them to azure blob with an internal queue in order to avoid the "too many open files" limits on some systems (sometimes you cannot update the max open file descriptors parameters) : you will send the blob for example 10 by 10 (or up to 10 at the same time)
it should also avoid this issue : https://code.google.com/p/go/issues/detail?id=3575

PR #107, opened 2015-04-29
Extend the deployments so you can add custom options

svanharmelen:
This extension is fully backwards compatible so doesn’t change the existing API, but extends it with an additional call so you can create a deployment with some more options.

A use case example is that without this code it isn’t possible to create a deployment for a custom virtual network/subnet…

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -1,4 +1,7 @@
-// Package virtualmachine implements operations on Azure virtual machines using the Service Management REST API
+// Package virtualmachine implements operations for managing virtual machines
+// using the Service Management REST API
+//
+// https://msdn.microsoft.com/en-us/library/azure/jj157206.aspx
 package virtualmachine
 
 import (
@@ -19,23 +22,51 @@ const (
 	errParamNotSpecified = "Parameter %s is not specified."
 )
 
-//NewClient is used to instantiate a new VmClient from an Azure client
+//NewClient is used to instantiate a new VirtualMachineClient from an Azure client
 func NewClient(client management.Client) VirtualMachineClient {
 	return VirtualMachineClient{client: client}
 }
 
-func (self VirtualMachineClient) CreateDeployment(role Role, cloudServiceName string) (management.OperationId, error) {
-	data, err := xml.Marshal(DeploymentRequest{
+// CreateDeploymentOptions can be used to create a customized deployement request
+type CreateDeploymentOptions struct {
+	DnsServers         *[]DnsServer
+	LoadBalancers      *[]LoadBalancer
+	ReservedIPName     string
+	Subnet             string
+	VirtualNetworkName string
+}
+
+// CreateDeployment creates a deployment and then creates a virtual machine
+// in the deployment based on the specified configuration.
+//
+// https://msdn.microsoft.com/en-us/library/azure/jj157194.aspx
+func (vm VirtualMachineClient) CreateDeployment(
+	role Role,
+	cloudServiceName string,
+	options *CreateDeploymentOptions) (management.OperationId, error) {
+
+	req := DeploymentRequest{
 		Name:           role.RoleName,
 		DeploymentSlot: "Production",
 		Label:          role.RoleName,
-		RoleList:       []Role{role}})
+		RoleList:       []Role{role},
+	}
+
+	if options != nil {
+		req.DnsServers = options.DnsServers
+		req.LoadBalancers = options.LoadBalancers
+		req.ReservedIPName = options.ReservedIPName
+		req.Subnet = options.Subnet
+		req.VirtualNetworkName = options.VirtualNetworkName
+	}
+
+	data, err := xml.Marshal(req)
 	if err != nil {
 		return "", err
 	}
 
 	requestURL := fmt.Sprintf(azureDeploymentListURL, cloudServiceName)
-	return self.client.SendAzurePostRequest(requestURL, data)
+	return vm.client.SendAzurePostRequest(requestURL, data)
 }
 
 func (self VirtualMachineClient) GetDeployment(cloudServiceName, deploymentName string) (DeploymentResponse, error) {
--- a/management/virtualmachine/entities.go
+++ b/management/virtualmachine/entities.go
@@ -22,6 +22,7 @@ type DeploymentRequest struct {
 	Label          string ``            // Specifies an identifier for the deployment. The label can be up to 100 characters long. The label can be used for tracking purposes.
 	RoleList       []Role `xml:">Role"` // Contains information about the Virtual Machines that are to be deployed.
 	// Optional parameters:
+	Subnet             string          `xml:",omitempty"`                         // Specifies the name of an existing subnet to which the deployment will belong.
 	VirtualNetworkName string          `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
 	DnsServers         *[]DnsServer    `xml:"Dns>DnsServers>DnsServer,omitempty"` // Contains a list of DNS servers to associate with the Virtual Machine.
 	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
--- a/management/virtualmachine/entities.go
+++ b/management/virtualmachine/entities.go
@@ -222,8 +222,8 @@ type VirtualIP struct {
 
 // Contains the configuration sets that are used to create virtual machines.
 type Role struct {
-	RoleName                    string                        // Specifies the name for the Virtual Machine.
-	RoleType                    string                        // Specifies the type of role to use. For Virtual Machines, this must be PersistentVMRole.
+	RoleName                    string                        `xml:",omitempty"` // Specifies the name for the Virtual Machine.
+	RoleType                    string                        `xml:",omitempty"` // Specifies the type of role to use. For Virtual Machines, this must be PersistentVMRole.
 	ConfigurationSets           *[]ConfigurationSet           `xml:"ConfigurationSets>ConfigurationSet",omitempty"`
 	ResourceExtensionReferences *[]ResourceExtensionReference `xml:"ResourceExtensionReferences>ResourceExtensionReference,omitempty"`
 	VMImageName                 string                        `xml:",omitempty"`                                         // Specifies the name of the VM Image that is to be used to create the Virtual Machine. If this element is used, the ConfigurationSets element is not used.
--- a/management/publishSettings.go
+++ b/management/publishSettings.go
@@ -47,7 +47,7 @@ func ClientFromPublishSettingsFileWithConfig(filePath, subscriptionId string, co
 					return client, err
 				}
 
-				pems, err := pkcs12.ConvertToPEM(pfxData, "")
+				pems, err := pkcs12.ConvertToPEM(pfxData, nil)
 
 				cert := []byte{}
 				for _, b := range pems {
--- a/management/storageservice/client.go
+++ b/management/storageservice/client.go
@@ -12,6 +12,7 @@ import (
 const (
 	azureStorageServiceListURL         = "services/storageservices"
 	azureStorageServiceURL             = "services/storageservices/%s"
+	azureStorageServiceKeysURL         = "services/storageservices/%s/keys"
 	azureStorageAccountAvailabilityURL = "services/storageservices/operations/isavailable/%s"
 
 	azureXmlns = "http://schemas.microsoft.com/windowsazure"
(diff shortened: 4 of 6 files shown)


ahmetalpbalkan:
@svanharmelen thanks for this pull request, looks good overall.

We're at a phase which we don't need to be backwards-compatible. Feel free to just refactor `CreateDeployment` to look like `CreateDeploymentWithOptions` as it makes the most sense. We prefer to have 1:1 mapping between REST calls and package methods.

svanharmelen:
@ahmetalpbalkan that remark actually worries me... So what does that mean for me as a user of this package?

I do see and understand that this package isn't up to spec yet and needs work in both styling and functionality, but if you currently do not care about backwards compatibility then I would at least expect a big fat warning in the README to NOT use this package for anything that matters to you, or to be very sure you have vendorred the code into your own repo/source?

Otherwise I can imagine a lot of people will not be happy users of your package. Especially since your versions are actually in the range of v1.0, v1.1 which normally suggests some form of release management and a stable API (since it's a v1+ release).

But good that I know this now :smiley: Then I'll just point my imports to my fork instead of this repo to guard myself from unforeseen API changes for the time being. So is there a roadmap or are there any thoughts on how to get to a stable API?

Last but not least... Will refactor this PR in a few, this time without worrying about the backwards compatibility :wink:

ahmetalpbalkan:
@svanharmelen A few months ago we started taking over this package from MSOpenTech. Historically this is developed to be used in projects like docker-machine, docker-registry etc. Initially it was very hacky when we took over and we had to do lots of refactoring (all of which was breaking). (The package basically had a global static state, these methods were not even defined on client classes like VirtualMachineClient).

We're not done with that refactoring "yet" to put a label on it. That's why we're continuing to break. I believe all the dependencies we know we have are either using an earlier (tagged) version or simply doing Godeps to locally vendor the source code of this package. When we are done and feel confident that the library looks in a good shape, we will be putting a version tag and following semantic versioning for version numbers. That'll probably be called 2.0.

So you're safe to break everything for now. :-)

svanharmelen:
@ahmetalpbalkan check!

So should I also update any styling issues on the parts that I touch? Meaning that currently the code looks a little more like C# than like Go and sometimes has really long lines. If I run tools like `golint` or `go vet` (which my editor does automatically for me) I get a whole bunch of complaints which I would expect need to be fixed before releasing a v2.0 right?

So I can update what I touch, but that would mean introducing a different style than you currently have in the package. For example the updated `CreateDeployment` method would now look like this:

```
// CreateDeployment creates a deployment and then creates a virtual machine
// in the deployment based on the specified configuration.
func (vm VirtualMachineClient) CreateDeployment(
	role Role,
	cloudServiceName string,
	options *DeploymentOptions) (management.OperationId, error) {

	req := DeploymentRequest{
		Name:           role.RoleName,
		DeploymentSlot: "Production",
		Label:          role.RoleName,
		RoleList:       []Role{role},
	}

	if options != nil {
		req.DnsServers = options.DnsServers
		req.LoadBalancers = options.LoadBalancers
		req.ReservedIPName = options.ReservedIPName
		req.Subnet = options.Subnet
		req.VirtualNetworkName = options.VirtualNetworkName
	}

	data, err := xml.Marshal(req)
	if err != nil {
		return "", err
	}

	requestURL := fmt.Sprintf(azureDeploymentListURL, cloudServiceName)
	return vm.client.SendAzurePostRequest(requestURL, data)
}
```

Instead of the current:

```
func (self VirtualMachineClient) CreateDeployment(role Role, cloudServiceName string) (management.OperationId, error) {
	data, err := xml.Marshal(DeploymentRequest{
		Name:           role.RoleName,
		DeploymentSlot: "Production",
		Label:          role.RoleName,
		RoleList:       []Role{role}})
	if err != nil {
		return "", err
	}

	requestURL := fmt.Sprintf(azureDeploymentListURL, cloudServiceName)
	return self.client.SendAzurePostRequest(requestURL, data)
}
``` 

So is that OK, or should I just update the functionality instead so you can do the whole styling in one go?

svanharmelen:
@ahmetalpbalkan just updated the PR according to our earlier discussion. Please let me know if it's OK like this, or that I should leave the styling as is for the time being.

As for some background, I'm currently using this package to build a Terraform provider. I've build the CloudStack provider a few months ago and now have a use case (and some requests from the community) for Azure support as well.

management/virtualmachine/entities.go
```diff
@@ -22,10 +22,20 @@ type DeploymentRequest struct {
...
 	DnsServers         *[]DnsServer    `xml:"Dns>DnsServers>DnsServer,omitempty"` // Contains a list of DNS servers to associate with the Virtual Machine.
-	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
 	LoadBalancers      *[]LoadBalancer `xml:">LoadBalancer,omitempty"`            //Contains a list of internal load balancers that can be assigned to input endpoints.
+	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
+	Subnet             string          `xml:",omitempty"`                         // Specifies the name of an existing subnet to which the deployment will belong.
+	VirtualNetworkName string          `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
+}
+
+// DeploymentOptions can be used to create a customized deployement request
+type DeploymentOptions struct {
```
ahmetalpbalkan: Should be named `CreateDeploymentOptions`

management/virtualmachine/entities.go
```diff
@@ -22,10 +22,20 @@ type DeploymentRequest struct {
...
+	DnsServers         *[]DnsServer
+	LoadBalancers      *[]LoadBalancer
+	ReservedIPName     string
```
ahmetalpbalkan: `xml:",omitempty"`  on strings please (I guess pointers on slices were used b/c we don't wanna transmit those empty arrays.

management/virtualmachine/client.go
```diff
@@ -19,23 +22,42 @@ const (
...
-func (self VirtualMachineClient) CreateDeployment(role Role, cloudServiceName string) (management.OperationId, error) {
-	data, err := xml.Marshal(DeploymentRequest{
+// CreateDeployment creates a deployment and then creates a virtual machine
+// in the deployment based on the specified configuration.
+//
+// https://msdn.microsoft.com/en-us/library/azure/jj157194.aspx
+func (vm VirtualMachineClient) CreateDeployment(
+	role Role,
+	cloudServiceName string,
+	options *DeploymentOptions) (management.OperationId, error) {
```
ahmetalpbalkan: I don't think this should be accepting a pointer (and thus allow "nil" instead of zero value). We can allow fields inside `DeploymentOptions` to be pointers (which we do) and zero value will just work and we'll get rid of nil-check below.

management/virtualmachine/entities.go
svanharmelen: Check, will update that one...

ahmetalpbalkan:
@paulmey please review. (esp the signature change on CreateDeployment)

management/virtualmachine/entities.go
svanharmelen: I understand, but these values will never be marshalled to XML as their values are used to fill the bigger `DeploymentRequest` struct which does have the correct tags. So are you sure you want these tags here as well?

management/virtualmachine/client.go
svanharmelen: My reasoning is that it's very user unfriendly (at least IMO) that if you don't need the extra options, you still need to create the object. Which is this case also has a pretty long name `virtualmachine.CreateDeploymentOptions{}`, so means some extra typing even through you don't use it.

management/virtualmachine/entities.go
ahmetalpbalkan: @svanharmelen ah my bad, ignore that.

management/virtualmachine/client.go
ahmetalpbalkan: @svanharmelen working a lot with .NET SDK that somewhat sounds normal. :smile: As I said before, I think methods here should 1:1 match the REST API and have high-fidelity mapping and flexibility to use all features of the REST APIs. For the convenience methods we have `vmutils` package. But I'm not saying this is a religious choice or anything. That's one of the things we need to decide and unify across the codebase.

svanharmelen:
@ahmetalpbalkan updated the type to `CreateDeploymentOptions`...

One additional question is about all the pointer to slices. Is there a specific reason to have all these? It makes some uses of the package a little nasty (having to do this `(*virtualmachine.DeploymentResponse.RoleList)[0]` instead of just this `virtualmachine.DeploymentResponse.RoleList[0]`) and using pointers to slices don't really add any value as slices are referenced types anyway. So in terms of memory usage there is no benefit and the zero value of a slice is `nil` just as it is for a pointer to a slice.

So just wondering why these are used so much in the package... Thx!

NOTE: I not trying to be a smart-ass or to be annoy anyone! I'm just trying to help shape up the package so I can use it in a Go like manner in any projects that I work on that would need it... Just wanted to express that :smiley:

ahmetalpbalkan:
@svanharmelen they're probably inherited from old code or missed in our code reviews. Feel free to submit a new PR or just open a brief issue and we'll get to it.

Your help is always appreciated, we really need more eyes on the code and the design to shape it up. (The design itself is totally non-trivial.) In reply to your earlier long comment, golint/go vet fixes probably should go into a separate PR and we're open to styling and UX fixes, feel free to bring them on as proposal issues instead of PRs if you don't want to refactor the whole repo. :)

management/virtualmachine/entities.go
```diff
@@ -22,10 +22,11 @@ type DeploymentRequest struct {
...
 	// Optional parameters:
-	VirtualNetworkName string          `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
 	DnsServers         *[]DnsServer    `xml:"Dns>DnsServers>DnsServer,omitempty"` // Contains a list of DNS servers to associate with the Virtual Machine.
-	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
 	LoadBalancers      *[]LoadBalancer `xml:">LoadBalancer,omitempty"`            //Contains a list of internal load balancers that can be assigned to input endpoints.
+	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
+	Subnet             string          `xml:",omitempty"`                         // Specifies the name of an existing subnet to which the deployment will belong.
+	VirtualNetworkName string          `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
```
paulmey: I see that you changed the order of the properties in this request. I know that the API is picky about the order and https://msdn.microsoft.com/en-us/library/azure/jj157194.aspx lists VirtualNetworkName as coming before Dns.
Did you test this against the API?

(yeah, I also know the documentation is not as it should be, so I would be surprised if it's wrong in this case, be we should test it)

paulmey:
See note above, LGTM otherwise...

ahmetalpbalkan:
@paulmey I think generally this pattern is ok. The method signature for a REST API call should take parameters of the underlying REST request URL as first parameters (e.g. cloudServiceName, deploymentId) and the rest should go into a type-safe property bag (i.e. xxxParameters struct, in this case CreateDeploymentParameters). This is extensible and likely to sustain in a lifecycle where we maintain back-compat. I don't know what's the state of consistency about this but I remember we followed this in storageserviceclient recently?

paulmey:
@ahmetalpbalkan I mostly agree with the pattern, we indeed used that on the 'create storage account' method. My only reservation is with parameters that require some structure, eg. it would probably be a bad idea to have a complete ConfigurationSet as one of the additions parameters, but I think that's obvious and easy enough to see... :-)

management/virtualmachine/entities.go
svanharmelen: @paulmey yes, I certainly tested this :smiley: It's not only reordered but I also added the `Subnet` field and did more then a few tests today with creating virtual networks and subnets and then adding virtual machine to them. So added a few, updated a few existing and also deleted a few. So I would say I touched all operations...

svanharmelen:
As we're talking about patterns, I did some work on a `networksecuritygroup` package this evening and followed the current package approach a little, but at the same time wrote it with my Go hat on... Maybe that is a good one to look at and to discuss while we're talking about this stuff...

Will create a PR for that one in a few minutes.

management/virtualmachine/entities.go
svanharmelen: @paulmey just reverted the ordering as there isn't any good argument for the reordering other than personal preference. So if you have any experience with the order being picky, let's just keep that in sync with the API docs...

ahmetalpbalkan:
@svanharmelen can you please squash all commits into one? (we prefer ease of revert-ability and easily readable commit history)

svanharmelen:
@ahmetalpbalkan done...

svanharmelen:
@ahmetalpbalkan hmm... something went wrong here... I'll just open up a new PR so we have a clean merge...

ahmetalpbalkan:
yep squash is never easy if you're using it for the first time :)) I must write a blog post about git squash/fixup stuff.

PR #108, opened 2015-04-29
Fixing a small bug in the tags...

svanharmelen:
Without this fix you aren’t able to get a virtual network if you have registered DNS servers…


ahmetalpbalkan:
LGTM. Thanks.

PR #109, opened 2015-04-29
Add support for network security groups

svanharmelen:
This needs just a little more testing, but first tests look good.

Of course this one is a little different as most other packages in this SDK, but personally I really like this approach which also use a more Go like style…

### DIFF
--- a/management/networksecuritygroup/client.go
+++ b/management/networksecuritygroup/client.go
@@ -1,0 +1,254 @@
+// Package networksecuritygroup implements operations for managing network security groups
+// using the Service Management REST API
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913824.aspx
+package networksecuritygroup
+
+import (
+	"encoding/xml"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/management"
+)
+
+const (
+	createSecurityGroupURL           = "services/networking/networksecuritygroups"
+	deleteSecurityGroupURL           = "services/networking/networksecuritygroups/%s"
+	getSecurityGroupURL              = "services/networking/networksecuritygroups/%s?detaillevel=full"
+	listSecurityGroupsURL            = "services/networking/networksecuritygroups"
+	addSecurityGroupToSubnetURL      = "services/virtualnetwork/%s/subnets/%s/networksecuritygroups"
+	getSecurityGroupForSubnetURL     = "services/virtualnetwork/%s/subnets/%s/networksecuritygroups"
+	removeSecurityGroupFromSubnetURL = "services/virtualnetwork/%s/subnets/%s/networksecuritygroups/%s"
+	setSecurityGroupRuleURL          = "services/networking/networksecuritygroups/%s/rules/%s"
+	deleteSecurityGroupRuleURL       = "services/networking/networksecuritygroups/%s/rules/%s"
+
+	errParamNotSpecified = "Parameter %s is not specified."
+)
+
+// NewClient is used to instantiate a new SecurityGroupClient from an Azure client
+func NewClient(client management.Client) SecurityGroupClient {
+	return SecurityGroupClient{client: client}
+}
+
+// CreateNetworkSecurityGroup creates a new network security group within
+// the context of the specified subscription
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913818.aspx
+func (sg SecurityGroupClient) CreateNetworkSecurityGroup(
+	name string,
+	label string,
+	location string) (management.OperationId, error) {
+	if name == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "name")
+	}
+	if location == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "location")
+	}
+
+	data, err := xml.Marshal(SecurityGroupRequest{
+		Name:     name,
+		Label:    label,
+		Location: location,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	requestURL := fmt.Sprintf(createSecurityGroupURL)
+	return sg.client.SendAzurePostRequest(requestURL, data)
+}
+
+// DeleteNetworkSecurityGroup deletes the specified network security group from the subscription
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913825.aspx
+func (sg SecurityGroupClient) DeleteNetworkSecurityGroup(
+	name string) (management.OperationId, error) {
+	if name == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "name")
+	}
+
+	requestURL := fmt.Sprintf(deleteSecurityGroupURL, name)
+	return sg.client.SendAzureDeleteRequest(requestURL)
+}
+
+// GetNetworkSecurityGroup returns information about the specified network security group
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913821.aspx
+func (sg SecurityGroupClient) GetNetworkSecurityGroup(name string) (SecurityGroupResponse, error) {
+	if name == "" {
+		return SecurityGroupResponse{}, fmt.Errorf(errParamNotSpecified, "name")
+	}
+
+	requestURL := fmt.Sprintf(getSecurityGroupURL, name)
+	response, err := sg.client.SendAzureGetRequest(requestURL)
+	if err != nil {
+		return SecurityGroupResponse{}, err
+	}
+
+	var securityGroup SecurityGroupResponse
+	if err := xml.Unmarshal(response, &securityGroup); err != nil {
+		return SecurityGroupResponse{}, err
+	}
+
+	return securityGroup, nil
+}
+
+// ListNetworkSecurityGroups returns a list of the network security groups
+// in the specified subscription
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913815.aspx
+func (sg SecurityGroupClient) ListNetworkSecurityGroups() (SecurityGroupList, error) {
+	response, err := sg.client.SendAzureGetRequest(listSecurityGroupsURL)
+	if err != nil {
+		return nil, err
+	}
+
+	var securityGroups SecurityGroupList
+	if err = xml.Unmarshal(response, &securityGroups); err != nil {
+		return nil, err
+	}
+
+	return securityGroups, nil
+}
+
+// AddNetworkSecurityToSubnet associates the network security group with
+// specified subnet in a virtual network
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913822.aspx
+func (sg SecurityGroupClient) AddNetworkSecurityToSubnet(
+	name string,
+	subnet string,
+	virtualNetwork string) (management.OperationId, error) {
+	if name == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "name")
+	}
+	if subnet == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "subnet")
+	}
+	if virtualNetwork == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "virtualNetwork")
+	}
+
+	data, err := xml.Marshal(SecurityGroupRequest{Name: name})
+	if err != nil {
+		return "", err
+	}
+
+	requestURL := fmt.Sprintf(addSecurityGroupToSubnetURL, virtualNetwork, subnet)
+	return sg.client.SendAzurePostRequest(requestURL, data)
+}
+
+// GetNetworkSecurityGroupForSubnet returns information about the network
+// security group associated with a subnet
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913817.aspx
+func (sg SecurityGroupClient) GetNetworkSecurityGroupForSubnet(
+	subnet string,
+	virtualNetwork string) (SecurityGroupResponse, error) {
+	if subnet == "" {
+		return SecurityGroupResponse{}, fmt.Errorf(errParamNotSpecified, "subnet")
+	}
+	if virtualNetwork == "" {
+		return SecurityGroupResponse{}, fmt.Errorf(errParamNotSpecified, "virtualNetwork")
+	}
+
+	requestURL := fmt.Sprintf(getSecurityGroupForSubnetURL, virtualNetwork, subnet)
+	response, err := sg.client.SendAzureGetRequest(requestURL)
+	if err != nil {
+		return SecurityGroupResponse{}, err
+	}
+
+	var securityGroup SecurityGroupResponse
+	if err := xml.Unmarshal(response, &securityGroup); err != nil {
+		return SecurityGroupResponse{}, err
+	}
+
+	return securityGroup, nil
+}
+
+// RemoveNetworkSecurityGroupFromSubnet removes the association of the
+// specified network security group from the specified subnet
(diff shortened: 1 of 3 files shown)


svanharmelen:
@paulmey @ahmetalpbalkan this is the PR I was just referring to in PR #107

It needs another round of testing which I will do tomorrow (CET time here, so getting a little late), but first tests look good and I can create, update and delete groups just fine. And if we're wanting to discuss patterns, I guess this is a good one to discuss things over :wink:

paulmey:
code LGTM, if you can add some integration tests, that would be awesome. the vmutils package has a pattern where it picks up the cert and subscription id from environment variables.

ahmetalpbalkan:
@svanharmelen can you please rebase this on top of master (MSOpenTech&rarr;Azure changes in the imports –and thus tests).

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+	getSecurityGroupForSubnetURL     = "services/virtualnetwork/%s/subnets/%s/networksecuritygroups"
+	removeSecurityGroupFromSubnetURL = "services/virtualnetwork/%s/subnets/%s/networksecuritygroups/%s"
+	setSecurityGroupRuleURL          = "services/networksecuritygroups/%s/rules/%s"
+	deleteSecurityGroupRuleURL       = "services/networksecuritygroups/%s/rules/%s"
+
+	errParamNotSpecified = "Parameter %s is not specified."
+)
+
+// NewClient is used to instantiate a new SecurityGroupClient from an Azure client
+func NewClient(client management.Client) *SecurityGroupClient {
```
ahmetalpbalkan: Is there a specific reason you return pointers? We don't care much about extra struct allocation, however nillable return values raise more risks like panics. If none of the methods defined on the type are modifying the struct, can we not return pointers?

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+	if location == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "location")
+	}
+
+	req := SecurityGroup{
+		Name:     name,
+		Location: location,
+	}
+
+	if label != "" {
```
ahmetalpbalkan: check seems necessary. you're already assigning to string, not to *string.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+func NewClient(client management.Client) *SecurityGroupClient {
+	return &SecurityGroupClient{client: client}
+}
+
+// CreateNetworkSecurityGroup creates a new network security group within
+// the context of the specified subscription
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913818.aspx
+func (sg *SecurityGroupClient) CreateNetworkSecurityGroup(
+	name string,
```
ahmetalpbalkan: it looks like we need a `struct CreateNetworkSecurityGroupParameters` here to follow consistency.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	requestURL := fmt.Sprintf(deleteSecurityGroupURL, name)
+	return sg.client.SendAzureDeleteRequest(requestURL)
+}
+
+// GetNetworkSecurityGroup returns information about the specified network security group
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913821.aspx
+func (sg *SecurityGroupClient) GetNetworkSecurityGroup(name string) (*SecurityGroup, error) {
```
ahmetalpbalkan: I believe there is no need to return pointer here.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var securityGroup SecurityGroup
+	if err := xml.Unmarshal(response, &securityGroup); err != nil {
+		return nil, err
+	}
+
+	return &securityGroup, nil
```
ahmetalpbalkan: you can rewrite this as:

```go
err = xml.Unmarshal(response, &securityGroup);
return securityGroup, err 
```

gives the same impact really once you change this method to not to return pointer.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// ListNetworkSecurityGroup returns a list of the network security groups
+// in the specified subscription
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913815.aspx
+func (sg *SecurityGroupClient) ListNetworkSecurityGroup() ([]*SecurityGroup, error) {
```
ahmetalpbalkan: 2 suggestions:

1. Name should be ListNetworkSecurityGroup**s**
2. No need to return pointer array here. `[]SecurityGroup` is just fine.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+
+	requestURL := fmt.Sprintf(addSecurityGroupToSubnetURL, virtualNetwork, subnet)
+	return sg.client.SendAzurePostRequest(requestURL, data)
+}
+
+// GetNetworkSecurityGroupForSubnet returns information about the network
+// security group associated with a subnet
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913817.aspx
+func (sg *SecurityGroupClient) GetNetworkSecurityGroupForSubnet(
```
ahmetalpbalkan: There's no need to return pointer here either.

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,235 @@
...
+	return sg.client.SendAzureDeleteRequest(requestURL)
+}
+
+// SetNetworkSecurityGroupRule adds or updates a network security rule that
+// is associated with the specified network security group
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913819.aspx
+func (sg *SecurityGroupClient) SetNetworkSecurityGroupRule(
+	securityGroup string,
+	rule *Rule) (management.OperationId, error) {
```
ahmetalpbalkan: I don't think `rule` should be a pointer here (nil is not really a valid input, so let language's type system handle that for you).

management/networksecuritygroup/entities.go
```diff
@@ -0,0 +1,45 @@
...
+
+// SecurityGroupClient is used to manage operations on network security groups
+type SecurityGroupClient struct {
+	client management.Client
+}
+
+// SecurityGroup represents a network security group
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn913821.aspx
+type SecurityGroup struct {
```
ahmetalpbalkan: Although `struct SecurityGroup` looks the same in the "Create" request and "List" or "Get" response, they probably should be separate. In order to achieve that you can do:

```go
type SecurityGroup struct { ... }
type SecurityGroupCreateParameters SecurityGroup
type SecurityGroupResponse SecurityGroup
type SecurityGroupListResponse { SecurityGroups []SecurityGroupResponse }
```

I prefer to follow this convention (esp with SecurityGroupListResponse) because sometimes our  ListXxx API responses return more info than just an array of things (like they include a bit more info) and it's impossible to return those as well without breaking others.

ahmetalpbalkan:
@svanharmelen this is exciting thanks! Please bear with us and we are going to deliver a nice package!

management/networksecuritygroup/client.go
svanharmelen: 1. You're right, will rename that one in the next commit!

2. Seeing all the other pointer related comments, let's have a discussion on that topic. Will create an issue so we have a place to discuss.

management/networksecuritygroup/client.go
svanharmelen: Hmm.. Sharp eye :wink: The zero value initialised in the struct will also be an empty string, so indeed it doesn't matter if I explicitly set it to an empty string. It's all the same inside the struct... Will remove the additional check.

management/networksecuritygroup/client.go
svanharmelen: I'll open up a separate issue to discuss the use of pointers...

management/networksecuritygroup/client.go
ahmetalpbalkan: @svanharmelen in a nutshell, why are you returning pointers in general?

management/networksecuritygroup/client.go
svanharmelen: Why would you want/need that here? The arguments here are all used to create a new group and it are only 3 strings, so the signature is still pretty clean and very easy to use.

Or are you saying that you would like to switch the default approach for all functions to use parameter structs? So what would be the guideline/general rule for that? Only have one argument being a parameter struct? Of having a few parameters for the mandatory fields and all optional ones into a parameter struct? But what if you have 7 or 8 mandatory fields and the signature gets really long?

A lot of questions/options so some more details are welcome.

management/networksecuritygroup/client.go
svanharmelen: I'll open up a separate issue to discuss the use of pointers...

management/networksecuritygroup/client.go
svanharmelen: I'll open up a separate issue to discuss the use of pointers...

management/networksecuritygroup/client.go
svanharmelen: I'll open up a separate issue to discuss the use of pointers...

management/networksecuritygroup/client.go
svanharmelen: Yes I know, but this approach/style seems more idiomatic in Go. Have a look through some of the standard packages (a very nice example is [this one](https://github.com/tonnerre/golang/blob/master/src/pkg/json/encode.go#L78-L83)) and you'll notice a that every now and then they use this idiom.

Besides that I personally find it very clear for anyone reading this code what to expect if there was an error and what to expect when everything runs without an error.

management/networksecuritygroup/entities.go
svanharmelen: Yeah, I must admit I was still looking at the structs/entities so will tweak them a little... This PR is mainly here because I wanted to open the discussion on styles and approaches, so will do some refactoring on these before coming to a final (squashed) PR.

svanharmelen:
@paulmey yes, please consider this PR work in progress for now... When done with tweaking and testing I'll surely add some tests!

management/networksecuritygroup/client.go
ahmetalpbalkan: @svanharmelen let's start this discussion before merging this b/c this highly inconsistent with the rest of the project.

svanharmelen: Sure thing... I'm offline today (phone only) but will pick up tomorrow. Thx!

svanharmelen:
So... My week turned out a little different than I expected, but I have some time again now so let's try to get through the open PR's :smiley: 

I believe the main discussion point is about the use of pointers. While thinking a bit about it and discussing with a few colleagues I think it comes down to two things here:

1. My personal preference based on working with other Go packages
2. Technical arguments for one or the other solution

It can very well be that this is mostly a matter of personal preference. I'm using quite a lot of API clients written in Go, and almost all of them seem to use pointers to structs over just structs. Have a quick look at some examples of packages I've used lately that demonstrate this:

1. https://github.com/awslabs/aws-sdk-go/blob/master/aws/request.go
2. https://github.com/google/google-api-go-client/blob/master/compute/v1/compute-gen.go
3. https://github.com/google/go-github/blob/master/github/repos_comments.go

_Note: These are just some pointers to specific files, but please browse around the packages to check out other files as well._

As for the technical arguments it turns out that it's very debatable and it mostly comes down to performance/efficiency. Of course now-a-days memory is not a primary concern for structs with sizes like we see in this package (even through some could actually get pretty large depending on what you have deployed within your subscription). But then again you're writing an API client for other developers so you cannot possibly know how they are going to use your package. So trying to make the package as efficient as possible sounds like a good strategy to me.

So even through I personally really favour using pointers to struct and that I also feel like this is more common and idiomatic in Go, it's certainly not an official best practice or guideline that Go requires you to be followed. So if you really feel like the current approach in this package is the way to go, then I will adjust this PR's accordingly so we can move on in getting them merged and with that adding the new functionalities...

ahmetalpbalkan:
@svanharmelen thanks for this. It looks to me like the libraries you mentioned (1) uses pointers obviously because most methods mutate the struct and in (2)(3) it looks like the libraries do not want copies of the returned values to be passed along and try to be memory-efficient.

> As for the technical arguments it turns out that it's very debatable and it mostly comes down to performance/efficiency. Of course now-a-days memory is not a primary concern for structs with sizes like we see in this package (even through some could actually get pretty large depending on what you have deployed within your subscription).

That's still not entirely correct. Even though we return a pointer, we at least have one copy of it in memory. And:

> you cannot possibly know how they are going to use your package

If the developers want to be efficient, they can take reference of the returned value and pass pointers around and still be efficient.

> I also feel like this is more common and idiomatic in Go

I can hardly say this is the case in the Go standard library packages.

On the other hand, using pointers can easily lead to misbehaving code that tries to mutate state (of the client structs or responses) that is harder to trace. Moreover, switching between those (pointers vs values as return values) is an easy choice to make and will probably take only a few minutes.

I'm not saying it's a wrong thing, it's still an option that might even make our code cleaner, however for now I would prefer being consistent with the rest of the library and then we'll probably clean all that stuff up.

It'd be great if you could just modify this PR to not to operate on pointer types, that probably should  be a separate PR that we'd be happy to review if you volunteer to refactor the entire library. Let's keep this PR about "Add support for network security groups". :)

svanharmelen:
@ahmetalpbalkan not agreeing on all fronts :wink: but totally fine with the conclusion. I'll update the PR tomorrow (both this one and #109) so they are consistent with the rest of the package. Will give you a ping when I'm done so you can have another look at it...

svanharmelen:
@ahmetalpbalkan could you have another look at this one? I think it's a lot more to your liking now, as it follows the current styling a lot more matching the rest of the package quite nicely (IMO)...

Let me know any comments or concerns we need to look at. Thx...

management/networksecuritygroup/client.go
```diff
@@ -0,0 +1,254 @@
...
+
+	requestURL := fmt.Sprintf(getSecurityGroupURL, name)
+	response, err := sg.client.SendAzureGetRequest(requestURL)
+	if err != nil {
+		return SecurityGroupResponse{}, err
+	}
+
+	var securityGroup SecurityGroupResponse
+	if err := xml.Unmarshal(response, &securityGroup); err != nil {
+		return SecurityGroupResponse{}, err
```
ahmetalpbalkan: you can return `securityGroup` here as well. (that brings me to my earlier comment, you can just do the following):

    err := xml.Unmarshal
    return obj, err

and it'll be just fine. :) I remember you had a readability argument but it seems clearer to me. Up to you if you like to change, not a big deal.

ahmetalpbalkan:
LGTM overall.

svanharmelen:
That would be great as I'm depend on both this PR and PR #113 for the work I've prepared for Terraform. So once this is merged I can move on and add Azure support in Terraform :smiley:

svanharmelen:
Thx!

ahmetalpbalkan:
@svanharmelen thanks for the PR. We appreciate when you bring more as well. :smile:

PR #116, opened 2015-05-11
Added IsResourceNotFoundError recognition function with tests.

aznashwan:


ahmetalpbalkan:
Thanks for the PR @aznashwan. Reviewing now.

management/errors_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/management"
+)
+
+// isResourceNotFoundTestCases is a set of structs comprising of the error
+// IsResourceNotFoundError should test and the expected result.
+var isResourceNotFoundTestCases = []struct {
```
ahmetalpbalkan: No need for making this a pkg level var. can you move this into the test case?

management/errors_test.go
```diff
@@ -0,0 +1,42 @@
...
+	err      error
+	expected bool
+}{
+	{
+		&management.AzureError{
```
ahmetalpbalkan: You might initialize these anonymous structs in a single line, would be easier to read. Not a problem though.

management/errors.go
```diff
@@ -0,0 +1,29 @@
...
+
+//Error implements the error interface for the AzureError type.
+func (e *AzureError) Error() string {
+	return fmt.Sprintf("Error response from Azure. Code: %s, Message: %s", e.Code, e.Message)
+}
+
+// IsResourceNotFoundError returns true if the provided error is an AzureError
+// reporting that a given resource has not been found.
+func IsResourceNotFoundError(err error) bool {
+	if err, ok := err.(*AzureError); ok && err.Code == "ResourceNotFound" {
```
ahmetalpbalkan: No need for if block. You can just 

    return ok && err.Code="..."

aznashwan:
All points addressed. Thanks for the review :smile:

management/errors.go
```diff
@@ -0,0 +1,26 @@
...
+// AzureError represents an error returned by the management API. It has an error
+// code (for example, ResourceNotFound) and a descriptive message.
+type AzureError struct {
+	XMLName xml.Name `xml:"Error"`
+	Code    string
+	Message string
+}
+
+//Error implements the error interface for the AzureError type.
+func (e *AzureError) Error() string {
```
ahmetalpbalkan: Hhmm I don't know why this was initially defined on a pointer to comply to the `error` interface historically. This sounds like a bad idea however.

While we're at it can you please define this on `AzureError` (without pointer). Once you do that I guess you'll need to change one or two places in the code plus the test cases below (to remove `&`) but we will end up in a nicer world IMO. :hatching_chick:

aznashwan:
Done.
After a quick look through the code I can't imagine any other problems this might cause...

ahmetalpbalkan:
LGTM thanks.

management/client.go
```diff
@@ -13,19 +9,6 @@ const (
 	errParamNotSpecified                  = "Parameter %s is not specified."
```
paulmey: is this still used? (since fmt is removed from the imports?)

ahmetalpbalkan: @paulmey looks like this is merged, feel free to submit a PR.

PR #118, opened 2015-05-12
Added support to associate a Network Security Group with a VM

svanharmelen:
Also removed `subnet` from the deployment options as it’s not used there. The subnet config is part of a `ConfigurationSet` of type `Network` and so is configured by the function `ConfigureWithSubnet`

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -32,7 +32,6 @@ type CreateDeploymentOptions struct {
 	DnsServers         *[]DnsServer
 	LoadBalancers      *[]LoadBalancer
 	ReservedIPName     string
-	Subnet             string
 	VirtualNetworkName string
 }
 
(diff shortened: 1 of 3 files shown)


svanharmelen:
@ahmetalpbalkan I have another PR lined up to refactor all the pointers to slices, but in my branch is build on top of this one. So if you have a minute to check this one and if OK merge it, I can add the other one right after that.

Of course if any comments, just let me know...

ahmetalpbalkan:
LGTM overall. Normally let's stick to the field orders in `DeploymentRequest` and not change them because REST API complains if some fields are not in the order as they documented in certain APIs. However if you tested it and it works, it's fine.

svanharmelen:
@ahmetalpbalkan thanks... And thanks for the tip about commit messages, will take it to heart :smiley:

About the ordering of fields, I know from the last PR so this one actually fixes the ordering to be inline with the docs :wink:

PR #119, opened 2015-05-12
Remove client-side logic from CreateHostedService

### LINKED ISSUE #38: Remove Logic from Hosted Service Client
When creating a hosted service, there is a call made to check the name availability. My opinion is that we should remove this call and make it a user responsibility, since they have to handle the `ConflictError` which can result from the race between checking the availability and creating the hosted service. 

However, the check being present makes it very difficult to find out what actual error gets returned if you hit the race condition!

ahmetalpbalkan:
Fixes #38


paulmey:
LGTM

PR #120, opened 2015-05-12
Refactor all slices to no longer use a pointer...

svanharmelen:
I've tested with this quite a lot to be sure everything still works as expected and I can confirm it does (added and removed multiple VM's, VNET's, subnets, endpoints and NSGs).

Next to it seeming a bit odd (unless very specific reason to do so) to have a pointer to a referenced type, removing the pointers for slices makes working with the package a lot nicer and more intuitive.

### DIFF
--- a/core/http/request_test.go
+++ b/core/http/request_test.go
@@ -398,7 +398,7 @@ func TestParseHTTPVersion(t *testing.T) {
 
 type logWrites struct {
 	t   *testing.T
-	dst []string
+	dst *[]string
 }
 
 func (l logWrites) WriteByte(c byte) error {
--- a/core/http/request_test.go
+++ b/core/http/request_test.go
@@ -398,7 +398,7 @@ func TestParseHTTPVersion(t *testing.T) {
 
 type logWrites struct {
 	t   *testing.T
-	dst *[]string
+	dst []string
 }
 
 func (l logWrites) WriteByte(c byte) error {
--- a/management/virtualmachine/entities.go
+++ b/management/virtualmachine/entities.go
@@ -53,7 +53,7 @@ type DeploymentResponse struct {
 	LoadBalancers          []LoadBalancer     `xml:">LoadBalancer"`
 	ExtendedProperties     []ExtendedProperty `xml:">ExtendedProperty"`
 	PersistentVMDowntime   PersistentVMDowntime
-	VirtualIPs             []VirtualIP `xml:">VirtualIP`
+	VirtualIPs             []VirtualIP `xml:">VirtualIP"`
 	ExtensionConfiguration string      // cloud service extensions not fully implemented
 	ReservedIPName         string
 	InternalDnsSuffix      string
@@ -224,16 +224,16 @@ type VirtualIP struct {
 type Role struct {
 	RoleName                    string                       `xml:",omitempty"` // Specifies the name for the Virtual Machine.
 	RoleType                    string                       `xml:",omitempty"` // Specifies the type of role to use. For Virtual Machines, this must be PersistentVMRole.
-	ConfigurationSets           []ConfigurationSet           `xml:"ConfigurationSets>ConfigurationSet",omitempty"`
-	ResourceExtensionReferences []ResourceExtensionReference `xml:"ResourceExtensionReferences>ResourceExtensionReference,omitempty"`
-	VMImageName                 string                       `xml:",omitempty"`                                         // Specifies the name of the VM Image that is to be used to create the Virtual Machine. If this element is used, the ConfigurationSets element is not used.
-	MediaLocation               string                       `xml:",omitempty"`                                         // Required if the Virtual Machine is being created from a published VM Image. Specifies the location of the VHD file that is created when VMImageName specifies a published VM Image.
-	AvailabilitySetName         string                       `xml:",omitempty"`                                         // Specifies the name of a collection of Virtual Machines. Virtual Machines specified in the same availability set are allocated to different nodes to maximize availability.
-	DataVirtualHardDisks        []DataVirtualHardDisk        `xml:"DataVirtualHardDisks>DataVirtualHardDisk,omitempty"` // Contains the parameters that are used to add a data disk to a Virtual Machine. If you are creating a Virtual Machine by using a VM Image, this element is not used.
-	OSVirtualHardDisk           *OSVirtualHardDisk           `xml:",omitempty"`                                         // Contains the parameters that are used to create the operating system disk for a Virtual Machine. If you are creating a Virtual Machine by using a VM Image, this element is not used.
-	RoleSize                    string                       `xml:",omitempty"`                                         // Specifies the size of the Virtual Machine. The default size is Small.
-	ProvisionGuestAgent         bool                         `xml:",omitempty"`                                         // Indicates whether the VM Agent is installed on the Virtual Machine. To run a resource extension in a Virtual Machine, this service must be installed.
-	VMImageInput                *VMImageInput                `xml:",omitempty"`                                         // When a VM Image is used to create a new PersistentVMRole, the DiskConfigurations in the VM Image are used to create new Disks for the new VM. This parameter can be used to resize the newly created Disks to a larger size than the underlying DiskConfigurations in the VM Image.
+	ConfigurationSets           []ConfigurationSet           `xml:">ConfigurationSet,omitempty"`
+	ResourceExtensionReferences []ResourceExtensionReference `xml:">ResourceExtensionReference,omitempty"`
+	VMImageName                 string                       `xml:",omitempty"`                     // Specifies the name of the VM Image that is to be used to create the Virtual Machine. If this element is used, the ConfigurationSets element is not used.
+	MediaLocation               string                       `xml:",omitempty"`                     // Required if the Virtual Machine is being created from a published VM Image. Specifies the location of the VHD file that is created when VMImageName specifies a published VM Image.
+	AvailabilitySetName         string                       `xml:",omitempty"`                     // Specifies the name of a collection of Virtual Machines. Virtual Machines specified in the same availability set are allocated to different nodes to maximize availability.
+	DataVirtualHardDisks        []DataVirtualHardDisk        `xml:">DataVirtualHardDisk,omitempty"` // Contains the parameters that are used to add a data disk to a Virtual Machine. If you are creating a Virtual Machine by using a VM Image, this element is not used.
+	OSVirtualHardDisk           *OSVirtualHardDisk           `xml:",omitempty"`                     // Contains the parameters that are used to create the operating system disk for a Virtual Machine. If you are creating a Virtual Machine by using a VM Image, this element is not used.
+	RoleSize                    string                       `xml:",omitempty"`                     // Specifies the size of the Virtual Machine. The default size is Small.
+	ProvisionGuestAgent         bool                         `xml:",omitempty"`                     // Indicates whether the VM Agent is installed on the Virtual Machine. To run a resource extension in a Virtual Machine, this service must be installed.
+	VMImageInput                *VMImageInput                `xml:",omitempty"`                     // When a VM Image is used to create a new PersistentVMRole, the DiskConfigurations in the VM Image are used to create new Disks for the new VM. This parameter can be used to resize the newly created Disks to a larger size than the underlying DiskConfigurations in the VM Image.
 
 	UseCertAuth bool   `xml:"-"`
 	CertPath    string `xml:"-"`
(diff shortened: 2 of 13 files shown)


core/http/request_test.go
```diff
@@ -398,7 +398,7 @@ func TestParseHTTPVersion(t *testing.T) {
 
 type logWrites struct {
 	t   *testing.T
-	dst *[]string
+	dst []string
```
ahmetalpbalkan: Hey don't change this. This is golang stdlib :-)

ahmetalpbalkan:
Thanks for this @scanharmelen looks nice. My main question is does the Azure REST API work fine with those empty elements in the xml requests? But since you've tested and they work ok we can have this. Main reason we had pointers was simply because we didn't want to include those in the request when they're empty/nilslice. I don't know if there's a choice that predates us but I guess manual verification is good enough in this case until we have a full test suite for this.

core/http/request_test.go
svanharmelen: Whoeps... Search and replace error :wink: Will revert that one.

svanharmelen:
@ahmetalpbalkan yeah, that was the only reason I could think of as well, but indeed I tested it quite a lot today and to be honest it would be a very poor API if it would fail over those empty values. It's of course still very valid XML, even according to the docs... So I think we're good.

management/vmutils/network.go
```diff
@@ -42,15 +42,15 @@ func ConfigureWithExternalPort(role *Role, name string, localport, externalport
 	role.ConfigurationSets = updateOrAddConfig(role.ConfigurationSets, ConfigurationSetTypeNetwork,
 		func(config *ConfigurationSet) {
 			if config.InputEndpoints == nil {
-				config.InputEndpoints = &[]InputEndpoint{}
+				config.InputEndpoints = []InputEndpoint{}
```
ahmetalpbalkan: is this needed? I think it will work exactly the same without initialization http://play.golang.org/p/upFJkl5T5S

management/vmutils/configurationset.go
```diff
@@ -4,15 +4,15 @@ import (
 	. "github.com/Azure/azure-sdk-for-go/management/virtualmachine"
 )
 
-func updateOrAddConfig(configs *[]ConfigurationSet, configType ConfigurationSetType, update func(*ConfigurationSet)) *[]ConfigurationSet {
+func updateOrAddConfig(configs []ConfigurationSet, configType ConfigurationSetType, update func(*ConfigurationSet)) []ConfigurationSet {
 	if configs == nil {
-		configs = &[]ConfigurationSet{}
+		configs = []ConfigurationSet{}
```
ahmetalpbalkan: same comment about the initialization... not needed anymore?

management/vmutils/network.go
```diff
@@ -78,10 +78,10 @@ func ConfigureWithSubnet(role *Role, subnet string) error {
 	role.ConfigurationSets = updateOrAddConfig(role.ConfigurationSets, ConfigurationSetTypeNetwork,
 		func(config *ConfigurationSet) {
 			if config.SubnetNames == nil {
```
ahmetalpbalkan: same comment about the initialization... not needed anymore?

management/vmutils/vmutils.go
```diff
@@ -44,7 +44,7 @@ func ConfigureForLinux(role *Role, hostname, user, password string, sshPubkeyCer
 				if config.SSH.PublicKeys == nil {
 					keys = []PublicKey{}
```
ahmetalpbalkan: same comment about the initialization... not needed anymore?

management/vmutils/extensions.go
```diff
@@ -43,9 +43,9 @@ func AddAzureVMExtensionConfiguration(role *Role, name, publisher, version, refe
 	}
 
 	if role.ResourceExtensionReferences == nil {
```
ahmetalpbalkan: same comment about the initialization... not needed anymore?

svanharmelen:
@ahmetalpbalkan good catch! Your absolutely right that's not needed anymore, so we can now make that a lot cleaner. Let me refactor it right away...

ahmetalpbalkan:
@svanharmelen thanks. There might be other places that are not really obvious in the diff we're seeing here, I appreciate if you can take a peek at the rest of the nil checks as well.

svanharmelen:
@ahmetalpbalkan yes, will take a closer look at what code could/should be improved after changing the pointers to slices to just slices. Didn't actually look at changing the logic as the zero value of a slice is also `nil`, so all nil checks are still valid. But your absolutely right that while changing this we should also change/improve the related code peaces where applicable as well.

paulmey:
@svanharmelen did you test the 'update role' method with this new XML? e.g.:
- create a VM with an extension installed
- update the role for some other reason
- verify that the extension is still there

I'm worried that not all languages consider 'empty' and 'null' the same thing. Specifying an empty tag v.s. not having that tag at all is treated differently by some serializers in the .NET space (which is what the Azure API's use on the server side) (even though the XML spec says that they should be equivalent...). I'm not confident at all that having empty tags will work fine in all cases.

svanharmelen:
@paulmey yes I did. I tested with adding and removing endpoints and changing network security groups. Both test cases worked just fine. Any specific test case in mind you want me to try?

ahmetalpbalkan:
So @paulmey discovered the reason for this initially was this: Normal arrays (`*[]T` vs `[]T) are OK . XML element is not created for empty-slices or nil-slices (even without `'xml:",omitempty"'`). But the problem comes when you use nesting in the `’xml:...’` clause.

(example: http://play.golang.org/p/KVduOc9bw3 ) In this case

```go
    type T struct {
		Field []T `xml:"Fields>Field,omitempty"`
	}
```

and if we use `[]T` instead of  `*[]T`, the parent element `<Fields>` gets created anyway even though there won't be any `<Field>` (assuming it's by design in encoding/xml pkg).

We might be still fine if it's tested thoroughly and works. @paulmey @brendandixon wdyt?

ahmetalpbalkan:
... (In addition to my comment above) If we do

    field []T `xml:"Fields,omitempty>Field,omitempty"`

and have `,omitempty` in all steps, it just omits all those. @svanharmelen can you update accordingly?

svanharmelen:
Check... Let me see if I can update the code without updating the tests with that suggestion. Would indeed be the best of both if we can make it so. Will be tomorrow through as it's getting late over here (CET).

management/virtualmachine/entities.go
```diff
@@ -22,10 +22,10 @@ type DeploymentRequest struct {
...
 	RoleList       []Role `xml:">Role"` // Contains information about the Virtual Machines that are to be deployed.
 	// Optional parameters:
-	VirtualNetworkName string          `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
-	DnsServers         *[]DnsServer    `xml:"Dns>DnsServers>DnsServer,omitempty"` // Contains a list of DNS servers to associate with the Virtual Machine.
-	LoadBalancers      *[]LoadBalancer `xml:">LoadBalancer,omitempty"`            // Contains a list of internal load balancers that can be assigned to input endpoints.
-	ReservedIPName     string          `xml:",omitempty"`                         // Specifies the name of a reserved IP address that is to be assigned to the deployment.
+	VirtualNetworkName string         `xml:",omitempty"`                         // Specifies the name of an existing virtual network to which the deployment will belong.
+	DnsServers         []DnsServer    `xml:"Dns>DnsServers>DnsServer,omitempty"` // Contains a list of DNS servers to associate with the Virtual Machine.
```
paulmey: @ahmetalpbalkan showed me how to fix this... Apparently this should be `xml:"Dns,omitempty>DnsServers,omitempty>DnsServer,omitempty"` to omit all of the levels...
When you do that, all of the 'levels' are omitted when empty and my worries evaporate... :-)

management/virtualmachine/entities.go
```diff
@@ -222,18 +222,18 @@ type VirtualIP struct {
...
-	ProvisionGuestAgent         bool                          `xml:",omitempty"`                                         // Indicates whether the VM Agent is installed on the Virtual Machine. To run a resource extension in a Virtual Machine, this service must be installed.
-	VMImageInput                *VMImageInput                 `xml:",omitempty"`                                         // When a VM Image is used to create a new PersistentVMRole, the DiskConfigurations in the VM Image are used to create new Disks for the new VM. This parameter can be used to resize the newly created Disks to a larger size than the underlying DiskConfigurations in the VM Image.
+	RoleName                    string                       `xml:",omitempty"` // Specifies the name for the Virtual Machine.
+	RoleType                    string                       `xml:",omitempty"` // Specifies the type of role to use. For Virtual Machines, this must be PersistentVMRole.
+	ConfigurationSets           []ConfigurationSet           `xml:"ConfigurationSets>ConfigurationSet",omitempty"`
```
ahmetalpbalkan: `xml:"ConfigurationSets>ConfigurationSet",omitempty"`

there's an extra `"`  after ConfigurationSet

ahmetalpbalkan:
@svanharmelen totally unrelated, but at `entities.go:56`:

     	VirtualIPs             []VirtualIP `xml:">VirtualIP`

this is buggy, can you please change it to 

     	VirtualIPs             []VirtualIP `xml:"VirtualIPs>VirtualIP`

TIA!

svanharmelen:
@paulmey @ahmetalpbalkan unfortunately the suggested fix (doing something like `Dns,omitempty>DnsServers,omitempty>DnsServer,omitempty`) actually turned out not to work as expected.

If you look [here](https://github.com/golang/go/blob/master/src/encoding/xml/typeinfo.go#L124-L128) you'll see that the XML tag is split on comma's and then the first part is assigned back as the tag and the rest of the parts are processed as flags. So in effect the suggested fix results in an effective tag like `Dns,omitempty`.

The following example also shows that this not works as now with some data in the struct only the top level field (`Fields` is this case) is used and the child `Field` is left out: http://play.golang.org/p/8aHDIpm5HO

There are some sounds in the community to introduce an optional `omitparents` tag which would give us the behaviour we would want/need here, but that will not be in any time soon as it seems nobody actually make a PR for that (maybe I will if time permits).

So the only other solution is making a custom XML marshal function that filters out these kind of empty parents. I'm working on a solution using that approach which will give us clean marshalled XML the API would accept, while still using slices without pointers. This would possibly be a temp function as it isn't needed anymore as soon as the `omitparents` tag would be added to the language.

Little busy today so will have a closer look tomorrow again...

brendandixon:
If the Xml tags are put on the nested structures, it works as we would like. See http://play.golang.org/p/3tmTN8hRWd.

svanharmelen:
@brendandixon thanks for the tip, but looking at the current structs and their tags this will not work here. The tags are pointing to other structs and not just to a single field of another struct. So more like this: http://play.golang.org/p/jbFMh1q8Wx

So unfortunately that will not work unless we would create a whole lot of additional struct would become quite a mess if you ask me...

paulmey:
@svanharmelen you had a typo in your play (the first two tags were missing a "). Then it seems to work as expected (http://play.golang.org/p/0s_MCbV77w). IMHO it's a bug that the '>' naming shorthand leaves some empty tags around...
Extra structs is a bit messy, I agree, but then again the API is messy and if we want to stay close to the API's with this SDK, some of that mess is going to shine through...

svanharmelen:
@paulmey your right about the typo, but even with that fixed it still doesn't work as we want/need. Our use case is having and empty child node which leaves the parent node, were we don't want to retain the parent node if it's empty.

So take a look at the same corrected example you send me, but then with an empty struct: http://play.golang.org/p/gxthZrhNjU As you can see this still shows the parent `Fields`, so I'm afraid it really doesn't fix our problem...

A custom marshal function seems to do the trick (have a prototype I just tested with) until a new flag is added... Which I will make a change request for, hoping it will make Go 1.5.

ahmetalpbalkan:
@svanharmelen it does?: http://play.golang.org/p/EnyU72pFei if you add another omitempty to the "Fields".

svanharmelen:
@ahmetalpbalkan nope, it really doesn't :wink: It now *seems* to work with an empty struct, but if you put data in it, you'll notice you are missing the `Field` field in the output: http://play.golang.org/p/nKM9NPo5pm

Did you read my comment from about 9 hours ago? In that one I explain that it doesn't work and why, including a link to the relevant XML marshal source code. In there you can see/read that something like `Fields,omitempty>Field,omitempty` will just not work...

ahmetalpbalkan:
Oh sorry I was looking at the comments from mobile, didn't click all of them. 

I guess we're fine in sending empty array elements as long as the REST API is fine with it? @paulmey

svanharmelen:
@ahmetalpbalkan no worries :smiley:

If we're fine with that, that is of course by far the easiest and best solution (taken that the API is fine with it). But if we're not fine with that, a custom marshal func would also do the trick.

ahmetalpbalkan:
I'm fine with having empty arrays as long as API is also fine with them.

ahmetalpbalkan:
@svanharmelen yeah I'm totally fine. I have a few outstanding comments about nil checks, once you fix that we can take a last look and merge.

svanharmelen:
@ahmetalpbalkan updated the PR following your comments... Please review at will to see if I still missed anything.

ahmetalpbalkan:
LGTM

ahmetalpbalkan:
@svanharmelen I found one more nil check on slices after merging: `management/vmutils/configurationset.go` feel free to address anytime.

svanharmelen:
@ahmetalpbalkan if I'm reading the code correctly, this is all needed... It tries to retrieve a certain type of `ConfigurationSet` and if not found, it will create and add a `ConfigurationSet` of that type.

In this case it firsts needs to be checked if there already is one we can edit, before adding a new empty one.

paulmey:
Thanks guys!

ahmetalpbalkan:
@svanharmelen ah you're right, I read it as nil check for slice; turns out it isn't. tahnks.

PR #126, opened 2015-05-15
management/storage: Remove helpers and cleanup

### LINKED ISSUE #68: GetStorageServiceByLocation should be removed

ahmetalpbalkan:
Fixes #68. Removed some 'special logic' containing helper methods (including `Create` in favor of `CreateAsync`).
Did some code cleanup and stopped returning pointers.

cc:@paulmey

### DIFF
--- a/management/storageservice/entities.go
+++ b/management/storageservice/entities.go
@@ -12,11 +12,10 @@ type StorageServiceClient struct {
 }
 
 type StorageServiceList struct {
-	XMLName         xml.Name         `xml:"StorageServices"`
-	StorageServices []StorageService `xml:"StorageService"`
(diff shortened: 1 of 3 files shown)


management/storageservice/entities.go
```diff
@@ -12,11 +12,11 @@ type StorageServiceClient struct {
 }
 
 type StorageServiceList struct {
-	XMLName         xml.Name         `xml:"StorageServices"`
-	StorageServices []StorageService `xml:"StorageService"`
+	XMLName         xml.Name                 `xml:"StorageServices"`
```
paulmey: nit: I think this XMLName might not even be necessary, since we're only Unmarshalling this type...

paulmey:
LGTM

Issue #88, opened 2015-03-26
Allow specify x-ms-api-version in client configuration


ahmetalpbalkan:
`NewClient*WithConfig` methods should accept `ApiVersion` string to override the default hard-coded version.

ahmetalpbalkan:
Resolved via #127

PR #127, opened 2015-05-15
management: Custom API versions and User Agent Header

### LINKED ISSUE #55: Add User-Agent header
We must add User-Agent header while making queries to management APIs. It should be configurable but should have a default. We can get the UA convention from other language sdks we have.

ahmetalpbalkan:
Fixes #55, #88.

Also exposed default values used to build basic client and
a `DefaultConfig` method that allows user to get a copy of
the config object (that will make it easy to construct and
override only a part of it).

Signed-off-by: Ahmet Alp Balkan
cc: @paulmey

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -6,8 +6,11 @@ import (
 )
 
 const (
-	defaultAzureManagementURL             = "https://management.core.windows.net"
-	defaultOperationPollInterval          = time.Second * 30
+	DefaultAzureManagementURL    = "https://management.core.windows.net"
+	DefaultOperationPollInterval = time.Second * 30
+	DefaultAPIVersion            = "2014-10-01"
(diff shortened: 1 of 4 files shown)


management/client.go
```diff
@@ -6,8 +6,11 @@ import (
 )
 
 const (
-	defaultAzureManagementURL             = "https://management.core.windows.net"
-	defaultOperationPollInterval          = time.Second * 30
+	DefaultAzureManagementURL    = "https://management.core.windows.net"
+	DefaultOperationPollInterval = time.Second * 30
+	DefaultAPIVersion            = "2014-10-01"
+	DefaultUserAgent             = "azure-sdk-for-go"
```
paulmey: cool!

paulmey:
LGTM

PR #130, opened 2015-05-15
management/hostedservice: Refactoring various parts

ahmetalpbalkan:
1. `CreateHostedService` func now takes a struct input
   to stabilize the function signature. (base64 encoding of `Label` left to the
   user, just like in the rest API. implicitly handling that confuses user more.)
2. `CreateHostedService` is actually not a long running operation,
   therefore just returning `error` value from that func now.
3. Changed `CheckHostedServiceNameAvailabilityMethod` to return
   the correct response struct to the user.
4. Updated example codes and tests to accomodate changes.
   (Please run the integration tests after merging this.)

cc: @paulmey please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,56 +20,55 @@ Download publish settings file from https://manage.windowsazure.com/publishsetti
 package main
 
 import (
-    "fmt"
+	"encoding/base64"
+	"fmt"
 
-    "github.com/Azure/azure-sdk-for-go/management"
-    "github.com/Azure/azure-sdk-for-go/management/hostedservice"
-    "github.com/Azure/azure-sdk-for-go/management/virtualmachine"
-    "github.com/Azure/azure-sdk-for-go/management/vmutils"
+	"github.com/Azure/azure-sdk-for-go/management"
+	"github.com/Azure/azure-sdk-for-go/management/hostedservice"
+	"github.com/Azure/azure-sdk-for-go/management/virtualmachine"
+	"github.com/Azure/azure-sdk-for-go/management/vmutils"
 )
 
 func main() {
-    dnsName := "test-vm-from-go"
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -20,56 +20,54 @@ Download publish settings file from https://manage.windowsazure.com/publishsetti
...
 
 import (
-    "fmt"
+	"encoding/base64"
+	"fmt"
 
-    "github.com/Azure/azure-sdk-for-go/management"
-    "github.com/Azure/azure-sdk-for-go/management/hostedservice"
-    "github.com/Azure/azure-sdk-for-go/management/virtualmachine"
-    "github.com/Azure/azure-sdk-for-go/management/vmutils"
```
paulmey: I don't see the changes for this?

ahmetalpbalkan: I copy pasted the example fromm example method below. Its tabs vs spaces.

paulmey: but the example method itself is part of the vmutils package, I guess... so these references need to be put back manually...?

ahmetalpbalkan: I'm not sure what u mean. I believe Example() method is the copy of the readme. So I updated both to have working examples.

paulmey: Sorry for being unclear...
What I'm saying is that this code does not work if you copy it into a .go file and `go run` it.
The funcs ConfigureDeploymentFromPlatformImage, ConfigureForLinux and ConfigureWithPublicSSH are in the vmutils package. The new readme code doesn't import vmutils (because you copy-pasted code *from* the vmutils package).

BTW, I think you pointed me at the same thing last time I updated this example in the README... :smile:

ahmetalpbalkan: @paulmey oooh I see, I'll fix.

ahmetalpbalkan: fixed

PR #133, opened 2015-05-15
Cleanup in AzureError struct

ahmetalpbalkan:
cc:@paulmey

### DIFF
--- a/management/errors.go
+++ b/management/errors.go
@@ -8,7 +8,6 @@ import (
 // AzureError represents an error returned by the management API. It has an error
 // code (for example, ResourceNotFound) and a descriptive message.
 type AzureError struct {
-	XMLName xml.Name `xml:"Error"`
 	Code    string
 	Message string
 }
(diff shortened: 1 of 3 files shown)


management/errors.go
```diff
@@ -24,3 +23,14 @@ func IsResourceNotFoundError(err error) bool {
...
+
+// getAzureError converts an error response body into an AzureError instance.
+func getAzureError(responseBody []byte) error {
+	var azErr AzureError
+	err := xml.Unmarshal(responseBody, &azErr)
+	if err != nil {
+		return err
+	}
+
+	return fmt.Errorf("Failed parsing contents to AzureError format: %v", err)
```
paulmey: this will always return err? Comparing with the old code, I think you want to wrap the error on line 32 and return azErr here (currently, azErr is not used at all)...

ahmetalpbalkan: Ah I put this to the wrong place.

ahmetalpbalkan:
fixed

ahmetalpbalkan:
thx

PR #140, opened 2015-05-16
management/location: Refactoring

ahmetalpbalkan:
- Removed client side logic from `management/location`
- Renamed return structs to reflect operations better
- Moved helpers to where they are needed

cc: @paulmey

### DIFF
--- a/management/location/entities.go
+++ b/management/location/entities.go
@@ -14,7 +14,7 @@ type LocationClient struct {
 	client management.Client
 }
 
-type LocationList struct {
+type ListLocationsResponse struct {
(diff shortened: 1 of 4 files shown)


management/location/entities.go
```diff
@@ -14,7 +14,7 @@ type LocationClient struct {
 	client management.Client
 }
 
-type LocationList struct {
+type ListLocationsResponse struct {
 	XMLName   xml.Name   `xml:"Locations"`
```
paulmey: can probably be removed

paulmey:
LGTM

PR #146, opened 2015-05-18
management/storageservice: Fix inconsistencies

ahmetalpbalkan:
Fixed method signature and struct name inconsistencies in `management/storageservice`
package.


ahmetalpbalkan:
merging based on similar other patches got in

PR #148, opened 2015-05-19
Fixed IsResourceNotFoundError.

aznashwan:
Fixed the IsResourceNotFoundError function.
The type to cast to had to be changed due to the recent changes in the definition of the getAzureError function.

### DIFF
--- a/management/errors.go
+++ b/management/errors.go
@@ -20,7 +20,7 @@ func (e AzureError) Error() string {
 // IsResourceNotFoundError returns true if the provided error is an AzureError
 // reporting that a given resource has not been found.
 func IsResourceNotFoundError(err error) bool {
-	azureErr, ok := err.(AzureError)
+	azureErr, ok := err.(*AzureError)
 	return ok && azureErr.Code == "ResourceNotFound"
 }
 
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
I'm not sure why this is the case, we used to return pointers, now we don't. So the old code sounds right to me @aznashwan?

aznashwan:
I have trouble getting why this happens as well.
I strongly suspect that it's because casting an interface auto-implies some pointer vodoo.
The point is that after the change to getAzureError if you try to cast to the struct it fails; saying that the value it got was of type *AzureError which is uncastable to AzureError...

ahmetalpbalkan:
@aznashwan removing your `pkg/` dir, `go get -u` and rebuilding may help. Apparently you still have the old code somehow. From what I can tell, this shouldn't be happening. Please let us know if it still happens after building again.

aznashwan:
@ahmetalpbalkan after some internal code review I realised I had written not one; but two if statements which were complete utter bogus :disappointed: 

Thank your for your patience; and my apologies for the misunderstanding.

PR #149, opened 2015-05-19
Extract interface for management.Client

paulmey:
Allows tests to stub out at the HTTP client level.

cc: @aznashwan @ahmetalpbalkan

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -18,11 +18,43 @@ const (
 	errParamNotSpecified                  = "Parameter %s is not specified."
 )
 
-// Client is the base Azure Service Management API client instance that
-// can be used to construct client instances for various services.
-type Client struct {
+type client struct {
 	publishSettings publishSettings
 	config          ClientConfig
+}
+
+// Client is the base Azure Service Management API client instance that
+// can be used to construct client instances for various services.
+type Client interface {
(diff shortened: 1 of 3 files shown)


management/client.go
```diff
@@ -25,6 +25,13 @@ type Client struct {
 	config          ClientConfig
 }
 
+type AzureHttpClient interface {
+	SendAzureGetRequest(url string) ([]byte, error)
```
ahmetalpbalkan: Call this Client. And rename Client to client.

paulmey: yep, good suggestion

paulmey:
how about this @ahmetalpbalkan

paulmey:
It's actually quite convenient for making decorators this way: https://github.com/paulmey/azure-sdk-for-go/commit/0e7de0d2cc56ac52396bd94e283b9b481f1fda62

ahmetalpbalkan:
Would like to see @aznashwan's input before merging.

aznashwan:
@ahmetalpbalkan looks great.

This makes the client itself patchable, which is great.
I am, however, a bit concerned about the logistical aspects of up-and-coming testing; for which I will follow-up with an email when I get to that hurdle.

Thanks very much for the work guys! :smiley:

ahmetalpbalkan:
Thanks for your input! Merging.

PR #151, opened 2015-05-20
Fix a few things that got broken while refactoring

svanharmelen:
Biggest issue was that `doAzureOperation` was now always sending empty values as `contentType` and `data` parameters, breaking a lot of API calls of course.

Next to that some minor fixes/tweaks to make things play nice and little more robust.


ahmetalpbalkan:
LGTM thanks @svanharmelen for noticing.

Issue #153, opened 2015-05-20
Error trying to add a disk to a deployment...


svanharmelen:
The XML looks just fine, but still I receive the following error: `The specified disk size value of 0 GB is invalid. Disk size must be between 1 GB and 1023 GB.`

Strange enough the call itself looks just fine, so doubt if it's related to the SDK but wouldn't know what else it could be:

```xml
REQUEST: POST /<xxx>/services/hostedservices/sander-linux1/deployments HTTP/1.1
Host: management.core.windows.net
Content-Type: application/xml
User-Agent: azure-sdk-for-go
X-Ms-Version: 2014-10-01

<Deployment
    xmlns="http://schemas.microsoft.com/windowsazure">
    <Name>sander-linux1</Name>
    <DeploymentSlot>Production</DeploymentSlot>
    <Label>sander-linux1</Label>
    <RoleList>
        <Role>
            <RoleName>sander-linux1</RoleName>
            <RoleType>PersistentVMRole</RoleType>
            <ConfigurationSets>
                <ConfigurationSet>
                    <ConfigurationSetType>LinuxProvisioningConfiguration</ConfigurationSetType>
                    <HostName>sander-linux1</HostName>
                    <UserName>sander</UserName>
                    <UserPassword>xxx</UserPassword>
                    <DisableSshPasswordAuthentication>false</DisableSshPasswordAuthentication>
                </ConfigurationSet>
                <ConfigurationSet>
                    <InputEndpoints>
                        <InputEndpoint>
                            <LocalPort>22</LocalPort>
                            <Name>SSH</Name>
                            <Port>22</Port>
                            <Protocol>TCP</Protocol>
                        </InputEndpoint>
                    </InputEndpoints>
                    <SubnetNames>
                        <SubnetName>test</SubnetName>
                    </SubnetNames>
                    <NetworkSecurityGroup>sandergroep2</NetworkSecurityGroup>
                </ConfigurationSet>
            </ConfigurationSets>
            <DataVirtualHardDisks>
                <DataVirtualHardDisk>
                    <DiskLabel>disk1</DiskLabel>
                    <HostCaching>None</HostCaching>
                    <MediaLink>xxx.blob.core.windows.net/disks/disk1.vhd</MediaLink>
                    <Lun>0</Lun>
                    <LogicalDiskSizeInGB>12</LogicalDiskSizeInGB>
                </DataVirtualHardDisk>
            </DataVirtualHardDisks>
[...]

PR #154, opened 2015-05-21
Fix struct ordering to satify the API

svanharmelen:
Fixes issue #153

### DIFF
--- a/management/virtualmachine/entities.go
+++ b/management/virtualmachine/entities.go
@@ -253,7 +253,7 @@ type VMImageInput struct {
 // OSDiskConfiguration is used to resize the OS disk of a new VM created from a
 // previously saved VM image.
 type OSDiskConfiguration struct {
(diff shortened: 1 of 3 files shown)


ahmetalpbalkan:
@svanharmelen can you please squash into one commit and fix the test breaks?

svanharmelen:
Sure... Just don't see what the actual problem is. It looks like the command fails and directly after that is looks to pass:

image

So any pointers what the actual problem is here?

ahmetalpbalkan:
so the gofmt has edited these two files because they were not correctly formatted.

svanharmelen:
@ahmetalpbalkan check, thanks... Only thing I can see is an additional newline on the last line of both files. Removing those fixed it. Will squash again...

ahmetalpbalkan:
lgtm

PR #157, opened 2015-05-26
Despite the docs, these are the correct URLs

svanharmelen:


ahmetalpbalkan:
@svanharmelen interesting catch. LGTM.

But I'd like to double check the misdocumented endpoints with our folks:

* https://msdn.microsoft.com/en-us/library/azure/dn913822.aspx
* https://msdn.microsoft.com/en-us/library/azure/dn913817.aspx
* https://msdn.microsoft.com/en-us/library/azure/dn913817.aspx

Can't tell what's going on, docs seem like they're recently updated (apr 2015) but it could be an API version thing. 

It seems like they're added on API version `2014-10-01` and but they are recently updated in `2015-04-01` https://msdn.microsoft.com/en-us/library/azure/gg592580.aspx:

> image

They don't seem related to "subnets" in particular but I think this has something to do with that. 

Default API version we use in this client is `2014-10-01` and not `2015-04-01`... So can you please try bumping up the default version in `client.go` and try things without this change? If it works, we can just bump up the default version...

svanharmelen:
@ahmetalpbalkan just tested, but also when using the API version set to `2015-04-01` I get the exact same error...

I recall I've this before with another resource as well and also tested that against the latest API version without success.

Looked into my history and found them again:

https://msdn.microsoft.com/en-us/library/azure/dn913819.aspx and https://msdn.microsoft.com/en-us/library/azure/dn913816.aspx both have:

```
services/networksecuritygroups/<network-security-group-name>/rules/<rule-name>
```

 but it should be:
```
services/networking/networksecuritygroups/<network-security-group-name>/rules/<rule-name>
```

ahmetalpbalkan:
@svanharmelen that's totally interesting. I'll follow up internally.

ahmetalpbalkan:
Confirmed, docs are wrong. They'll be fixed soon. Thanks for reporting @svanharmelen.

PR #158, opened 2015-05-27
Updating bad struct field name

svanharmelen:


svanharmelen:
@paulmey @ahmetalpbalkan would be great if this one could be merged as I was just about to push all my tests (also using this one) into a Terraform PR... But noticed just in time I didn't push this one back to you guys already...

Thx!

ahmetalpbalkan:
Lgtm thx

svanharmelen:
Thx!

paulmey:
hmmm wait a minute... is https://msdn.microsoft.com/en-us/library/azure/jj157205.aspx wrong? Do we use this struct in a different context?

svanharmelen:
@paulmey yes, that doc is wrong (again :wink:)

paulmey:
Ok... sorry about that...

paulmey:
:disappointed:

ahmetalpbalkan:
Oh man. @Svanharmelen I should get those docs fixed, just tell me. :-) how do you find the correct names though?

svanharmelen:
@paulmey no worries... 

@ahmetalpbalkan Doing some reverse engineering based on the errors and a lot of debugging and then when I think I know what is wrong scrolling through the source code of the Powershell and/or .NET SDK to see how it's done in there...

ahmetalpbalkan:
thanks reporting this.

ahmetalpbalkan:
@svanharmelen hmm product team claims the `Name` was actually correct. (I also see `Name` [on the .NET SDK source](https://github.com/Azure/azure-sdk-for-net/blob/2eda9e5e9f43decd392d560b6b259168a3d94622/src/ServiceManagement/Compute/ComputeManagement/Generated/Models/VirtualMachineDiskUpdateParameters.cs#L76)). Have you tried if `DiskName` does the job? Where exactly did you see `DiskName` used?

svanharmelen:
@ahmetalpbalkan sure thing I tested and tried this! Let me revert it back and test again as I needed to fix more stuff to get everything working as expected. So let me (try) to confirm this was essential in getting it to work.

But even if it works with just `Name`, I'm 100% sure it works with `DiskName` as well as that's what I'm currently successfully use to get disks resized...

svanharmelen:
Nope, I can confirm that if I change `DiskName` back to just `Name`, it breaks again... The error is a little cryptic, but it points me into the XML parser having issues reading the fields it expects:

```
Error response from Azure. Code: BadRequest, Message: The value for parameter 'Label' is null or empty.
```

The XML I send looks like this (the working XML, the one breaking is 100% the same except for `DiskName` now being just `Name`):

```xml
<Disk xmlns="http://schemas.microsoft.com/windowsazure">
    <DiskName>terraform-test1-terraform-test1-1-201505272023270700</DiskName>
    <Label>terraform-test1-1</Label>
    <ResizedSizeInGB>20</ResizedSizeInGB>
</Disk>
```

ahmetalpbalkan:
Thanks a lot for your input! And when you update disk like that, is the new name your provide visible on the Get endpoint?

paulmey:
@svanharmelen that error could also indicate 'wrong' order of XML elements... In my experience, the serializer allows for unrecognized elements (DiskName?), but not for recognized elements in the wrong order. So maybe 
```xml
<Disk xmlns="http://schemas.microsoft.com/windowsazure">
    <Label>terraform-test1-1</Label>
    <Name>terraform-test1-terraform-test1-1-201505272023270700</Name>
    <ResizedSizeInGB>20</ResizedSizeInGB>
</Disk>
```
will work? (yes, I agree that the error messages are not very accurate...)

The way I've reverse engineered these things in the past is running them through Fiddler and looking at the actual XML going over the wire.

ahmetalpbalkan:
Yeah @paulmey could be right. If the changes you make using "DiskName" are actually effective in when you list/get disks (when xml elements are in the documented order) then the docs are wrong.

svanharmelen:
So the docs are wrong either way, the question is which way :wink:

Will check if it's the ordering instead of the naming in just a few...

paulmey:
Yeah, there's no excuse... :disappointed:

svanharmelen:
@paulmey you were right! It's indeed the order, not the name... So will put in a small PR to fix this...

PR #165, opened 2015-06-05
Fix integration tests

paulmey:
ResourceExtensionReferences needs to be a pointer to be able to
completely omit the element from the XML, which is necessary for
deploying a VM from a VM image (TestDeployFromVmImage).

cc: @svanharmelen @ahmetalpbalkan

### DIFF
--- a/management/vmutils/extensions_test.go
+++ b/management/vmutils/extensions_test.go
@@ -11,7 +11,7 @@ func Test_AddAzureVMExtensionConfiguration(t *testing.T) {
 
 	role := vm.Role{}
 	AddAzureVMExtensionConfiguration(&role,
-		"nameOfExtension", "nameOfPublisher", "versionOfExtension", "nameOfReference", "state", []byte{}, []byte{})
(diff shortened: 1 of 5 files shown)


management/vmutils/extensions_test.go
```diff
@@ -28,12 +28,7 @@ func Test_AddAzureVMExtensionConfiguration(t *testing.T) {
       <ResourceExtensionParameterValues>
         <ResourceExtensionParameterValue>
           <Key>ignored</Key>
-          <Value></Value>
-          <Type>Private</Type>
-        </ResourceExtensionParameterValue>
-        <ResourceExtensionParameterValue>
-          <Key>ignored</Key>
-          <Value></Value>
+          <Value>AQID</Value>
```
ahmetalpbalkan: I wonder where this string is coming from.

ahmetalpbalkan:
LGTM overall.

paulmey:
Thanks

PR #170, opened 2015-06-11
storage: Close underlying http conns before return

ahmetalpbalkan:
This was causing memory leaks due to not closed response bodies
and thus HTTP connections (especially in PutBlock method).

https://github.com/docker/distribution/issues/620


paulmey:
LGTM

PR #171, opened 2015-06-15
Added database server and service support.

aznashwan:
Added support for database server and service CR_D.

This pull depends on: https://github.com/Azure/azure-sdk-for-go/pull/173

### DIFF
--- a/management/sql/client.go
+++ b/management/sql/client.go
@@ -1,0 +1,160 @@
+package sql
+
+import (
+	"encoding/xml"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/management"
+)
+
+// Definitions of numerous constants representing API endpoints.
+const (
+	azureCreateDatabaseServerURL = "services/sqlservers/servers"
+	azureListDatabaseServersURL  = "/services/sqlservers/servers"
+	azureDeleteDatabaseServerURL = "/services/sqlservers/servers/%s"
+
+	azureCreateDatabaseURL = "services/sqlservers/servers/%s/databases"
+	azureGetDatabaseURL    = "/services/sqlservers/servers/%s/databases/%s"
+	azureListDatabasesURL  = "/services/sqlservers/servers/%s/databases?contentview=generic"
+	azureDeleteDatabaseURL = "/services/sqlservers/servers/%s/databases/%s"
+
+	errParamNotSpecified = "Parameter %s was not specified."
+)
+
+// SqlDatabaseClient defines various database CRUD operations.
+// It contains a management.Client for making the actual http calls.
+type SqlDatabaseClient struct {
+	mgmtClient management.Client
+}
+
+// NewClient returns a new SqlDatabaseClient struct with the provided
+// management.Client as the underlying client.
+func NewClient(mgmtClient management.Client) SqlDatabaseClient {
+	return SqlDatabaseClient{mgmtClient}
+}
+
+// CreateServer creates a new Azure SQL Database server and return its name.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+func (c SqlDatabaseClient) CreateServer(params DatabaseServerCreateParams) (string, error) {
+	req, err := xml.Marshal(params)
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := c.mgmtClient.SendAzurePostRequestWithReturnedResponse(azureCreateDatabaseServerURL, req)
+	if err != nil {
+		return "", err
+	}
+
+	var name string
+	err = xml.Unmarshal(resp, &name)
+
+	return name, err
+}
+
+// ListServers retrieves the Azure SQL Database servers for this subscription.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505702.aspx
+func (c SqlDatabaseClient) ListServers() (ListServersResponse, error) {
+	var resp ListServersResponse
+
+	data, err := c.mgmtClient.SendAzureGetRequest(azureListDatabaseServersURL)
+	if err != nil {
+		return resp, err
+	}
+
+	err = xml.Unmarshal(data, &resp)
+	return resp, err
+}
+
+// DeleteServer deletes an Azure SQL Database server (including all its databases).
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505695.aspx
+func (c SqlDatabaseClient) DeleteServer(name string) error {
+	if name == "" {
+		return fmt.Errorf(errParamNotSpecified, "name")
+	}
+
+	url := fmt.Sprintf(azureDeleteDatabaseServerURL, name)
+	_, err := c.mgmtClient.SendAzureDeleteRequest(url)
+	return err
+}
+
+// CreateDatabase creates a new Microsoft Azure SQL Database on the given database server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+func (c SqlDatabaseClient) CreateDatabase(server string, params DatabaseCreateParams) error {
+	if server == "" {
+		return fmt.Errorf(errParamNotSpecified, "server")
+	}
+
+	req, err := xml.Marshal(params)
+	if err != nil {
+		return err
+	}
+
+	target := fmt.Sprintf(azureCreateDatabaseURL, server)
+	_, err = c.mgmtClient.SendAzurePostRequest(target, req)
+	return err
+}
+
+// GetDatabase gets the details for an Azure SQL Database.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505708.aspx
+func (c SqlDatabaseClient) GetDatabase(server, database string) (ServiceResource, error) {
+	var db ServiceResource
+
+	if database == "" {
+		return db, fmt.Errorf(errParamNotSpecified, "database name")
+	}
+	if server == "" {
+		return db, fmt.Errorf(errParamNotSpecified, "server name")
+	}
+
+	url := fmt.Sprintf(azureGetDatabaseURL, server, database)
+	resp, err := c.mgmtClient.SendAzureGetRequest(url)
+	if err != nil {
+		return db, err
+	}
+
+	err = xml.Unmarshal(resp, &db)
+	return db, err
+}
+
+// ListDatabases returns a list of Azure SQL Databases on the given server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505711.aspx
+func (c SqlDatabaseClient) ListDatabases(server string) (ListDatabasesResponse, error) {
+	var databases ListDatabasesResponse
+	if server == "" {
+		return databases, fmt.Errorf(errParamNotSpecified, "server name")
+	}
+
+	url := fmt.Sprintf(azureListDatabasesURL, server)
+	resp, err := c.mgmtClient.SendAzureGetRequest(url)
+	if err != nil {
+		return databases, err
+	}
+
+	err = xml.Unmarshal(resp, &databases)
+	return databases, err
+}
+
+// DeleteDatabase deletes the Azure SQL Database off the given server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505705.aspx
+func (c SqlDatabaseClient) DeleteDatabase(server, database string) error {
+	if database == "" {
+		return fmt.Errorf(errParamNotSpecified, "database name")
+	}
+	if server == "" {
+		return fmt.Errorf(errParamNotSpecified, "server name")
+	}
+
+	url := fmt.Sprintf(azureDeleteDatabaseURL, server, database)
+
+	_, err := c.mgmtClient.SendAzureDeleteRequest(url)
+
+	return err
+}
--- a/management/sql/entities.go
+++ b/management/sql/entities.go
@@ -1,0 +1,76 @@
+package sql
+
+import (
+	"encoding/xml"
+)
+
+// DatabaseServerCreateParams represents the set of possible parameters
+// when issuing a database server creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+type DatabaseServerCreateParams struct {
+	XMLName                    xml.Name `xml:"http://schemas.microsoft.com/sqlazure/2010/12/ Server"`
+	AdministratorLogin         string
+	AdministratorLoginPassword string
+	Location                   string
+	Version                    string
+}
+
+// DatabaseServerCreateResponse represents the response following the creation of
+// a database server on Azure.
+type DatabaseServerCreateResponse struct {
+	ServerName string
+}
+
+const (
+	DatabaseServerVersion11 = "2.0"
+	DatabaseServerVersion12 = "12.0"
+)
+
+// DatabaseServer represents the set of data recieved from
+// a database server list operation.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505702.aspx
+type DatabaseServer struct {
+	Name                     string
+	AdministratorLogin       string
+	Location                 string
+	FullyQualifiedDomainName string
+	Version                  string
+	State                    string
+}
+
+type ListServersResponse struct {
+	DatabaseServers []DatabaseServer `xml:"Server"`
+}
+
+// DatabaseCreateParams represents the set of possible parameters when issuing
+// a database creation to Azure, and reading a list response from Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+type DatabaseCreateParams struct {
+	XMLName            xml.Name `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
+	Name               string
+	Edition            string `xml:",omitempty"`
+	CollationName      string `xml:",omitempty"`
+	MaxSizeBytes       int64  `xml:",omitempty"`
+	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
+}
+
+// ServiceResource represents the set of parameters obtained from a database
+// get or list call.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505708.aspx
+type ServiceResource struct {
+	Name               string
+	State              string
+	SelfLink           string
+	Edition            string
+	CollationName      string
+	MaxSizeBytes       int64
+	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
+}
+
+type ListDatabasesResponse struct {
+	ServiceResources []ServiceResource `xml:"ServiceResource"`
+}


ahmetalpbalkan:
Feel free to update the SendAzurePostRequest method to return the response []byte for that. :-)

management/database/client.go
```diff
@@ -0,0 +1,154 @@
+package database
```
ahmetalpbalkan: can we call this `sql` to be consistent with the nuget naming scheme as well https://www.nuget.org/packages/Microsoft.WindowsAzure.Management.Sql/ ? We have a product called DocumentDB as well and that's `database`, too.

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+// management.Client as the underlying client.
+func NewClient(mgmtClient management.Client) DatabaseClient {
+	return DatabaseClient{mgmtClient}
+}
+
+// CreateDatabaseServer issues a database server creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+func (c DatabaseClient) CreateDatabaseServer(params DatabaseServerCreateParams) error {
+	if params.Location == "" {
```
ahmetalpbalkan: no need to make these parameter checks unless they're required in constructing the URL, let the API fail with an error.

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+	azureGetDatabaseUrl    = "/services/sqlservers/servers/%s/databases/%s"
+	azureListDatabasesUrl  = "/services/sqlservers/servers/%s/databases?contentview=generic"
+	azureDeleteDatabaseUrl = "/services/sqlservers/servers/%s/databases/%s"
+
+	errParamNotSpecified = "Parameter %s is not specified."
+)
+
+// DatabaseClient defines various database CRUD operations.
+// It contains a management.Client for making the actual http calls.
+type DatabaseClient struct {
```
ahmetalpbalkan: SqlDatabaseClient would be better.

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return err
+	}
+
+	// NOTE: would be nice to be able to read the response here:
+	_, err = c.mgmtClient.SendAzurePostRequest(azureCreateDatabaseServerUrl, req)
+	return err
+}
+
+// ListDatabaseServers issues a request for a list of all database servers off Azure.
```
ahmetalpbalkan: let's be consistent with the docs link below

> ListDatabaseServers retrieves the Azure SQL Database servers for a given subscription.

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	State    string
+}
+
+type ListDatabaseServersResponse struct {
+	XMLName xml.Name         `xml:"Servers"`
+	Servers []DatabaseServer `xml:"Server"`
+}
+
+// Database represents the set of possible parameters when issuing
+// a database creation to Azure, and reading a list response from Azure.
```
ahmetalpbalkan: please do not reuse the same type `Database` for both serializing to requests and deserializing from responses. you can alternatively embed this type (make it unexported) into two exported types with xxxRequest and yyyResponse.

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"encoding/xml"
+)
+
+// DatabaseServerCreateParams represents the set of possible parameters
+// when issuing a database server creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+type DatabaseServerCreateParams struct {
+	XMLName  xml.Name `xml:"http://schemas.microsoft.com/sqlazure/2010/12/ Server"`
```
ahmetalpbalkan: is space before `Server` intentional?

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	DatabaseServerVersion12 = "12.0"
+)
+
+// DatabaseServerResponse represents the set of data recieved from
+// a database server list operation.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505702.aspx
+type DatabaseServer struct {
+	Name     string
+	UserName string `xml:"AdministratorLogin"`
```
ahmetalpbalkan: use the same name from XML response

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	Location string
+	DNSName  string `xml:"FullyQualifiedDomainName"`
```
ahmetalpbalkan: use the same name from XML response

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	Version  string
+	State    string
+}
+
+type ListDatabaseServersResponse struct {
+	XMLName xml.Name         `xml:"Servers"`
```
ahmetalpbalkan: I believe you don't need XMLName here. The root element of XML, in this case `<Servers>` will be figured out by the xml pkg automatically.

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505702.aspx
+func (c DatabaseClient) ListDatabaseServers() ([]DatabaseServer, error) {
```
ahmetalpbalkan: ListServers, let's be consistent with the API naming

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	err = xml.Unmarshal(data, &resp)
+	return resp.Servers, err
+}
+
+// DeleteDatabaseServer issues an Azure database server deletion request.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505695.aspx
+func (c DatabaseClient) DeleteDatabaseServer(name string) error {
```
ahmetalpbalkan: DeleteServer

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+// NewClient returns a new DatabaseClient struct with the provided
+// management.Client as the underlying client.
+func NewClient(mgmtClient management.Client) DatabaseClient {
+	return DatabaseClient{mgmtClient}
+}
+
+// CreateDatabaseServer issues a database server creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+func (c DatabaseClient) CreateDatabaseServer(params DatabaseServerCreateParams) error {
```
ahmetalpbalkan: CreateServer

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+func (c DatabaseClient) ListDatabaseServers() ([]DatabaseServer, error) {
+	var resp ListDatabaseServersResponse
+
+	data, err := c.mgmtClient.SendAzureGetRequest(azureListDatabaseServersUrl)
+	if err != nil {
+		return nil, err
+	}
+
+	err = xml.Unmarshal(data, &resp)
+	return resp.Servers, err
```
ahmetalpbalkan: Do not return the internal field, just return the response object `ListDatabaseServersResponse` and let the user figure out.

If some other field gets added right next to `.Servers` field, we'll be introducing breaking change to return that as well, we follow returning xxxResponse objects as a convention in the codebase.

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	// DatabaseServer should contain the name of the
+	// server the database is/should be hosted on.
+	DatabaseServer     string `xml:"-"`
+	Edition            string `xml:",omitempty"`
+	CollationName      string `xml:",omitempty"`
+	MaxSize            int64  `xml:"MaxSizeBytes,omitempty"`
+	ServiceObjectiveId string `xml:",omitempty"`
+}
+
+type ListDatabsesResponse struct {
```
ahmetalpbalkan: typo

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	XMLName   xml.Name   `xml:"http://schemas.microsoft.com/windowsazure ServiceResources"`
+	Databases []Database `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
```
ahmetalpbalkan: Use the same name with the API: `ServiceResource`. no need for introducing new names/concepts.

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+type Database struct {
+	XMLName xml.Name `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
+	Name    string
+	// DatabaseServer should contain the name of the
+	// server the database is/should be hosted on.
+	DatabaseServer     string `xml:"-"`
+	Edition            string `xml:",omitempty"`
+	CollationName      string `xml:",omitempty"`
+	MaxSize            int64  `xml:"MaxSizeBytes,omitempty"`
```
ahmetalpbalkan: use the same name as the API.

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+// Database represents the set of possible parameters when issuing
+// a database creation to Azure, and reading a list response from Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+type Database struct {
+	XMLName xml.Name `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
+	Name    string
+	// DatabaseServer should contain the name of the
+	// server the database is/should be hosted on.
+	DatabaseServer     string `xml:"-"`
```
ahmetalpbalkan: is this `xml:"-"` because this is used as a common type as xxxRequest and yyyResponse objects?

management/database/entities.go
```diff
@@ -0,0 +1,67 @@
...
+	"encoding/xml"
+)
+
+// DatabaseServerCreateParams represents the set of possible parameters
+// when issuing a database server creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505699.aspx
+type DatabaseServerCreateParams struct {
+	XMLName  xml.Name `xml:"http://schemas.microsoft.com/sqlazure/2010/12/ Server"`
+	UserName string   `xml:"AdministratorLogin"`
```
ahmetalpbalkan: use the same name as the API (same below)

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// CreateDatabase issues a database creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+func (c DatabaseClient) CreateDatabase(params Database) error {
+	if params.Name == "" {
+		return fmt.Errorf(errParamNotSpecified, "Name")
+	}
+	if params.DatabaseServer == "" {
```
ahmetalpbalkan: `Server` definitely should not go into `Database` (which should be `DatabaseCreateParameters`). This signature should read as:

    func (c DatabaseClient) CreateDatabase(server string, params DatabaseCreateParameters)

management/database/client.go
```diff
@@ -0,0 +1,154 @@
...
+	url := fmt.Sprintf(azureDeleteDatabaseServerUrl, name)
+	_, err := c.mgmtClient.SendAzureDeleteRequest(url)
+	return err
+}
+
+// CreateDatabase issues a database creation request to Azure.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+func (c DatabaseClient) CreateDatabase(params Database) error {
+	if params.Name == "" {
```
ahmetalpbalkan: do not have logic for checking which request-body parameters are required. let the API do it.

ahmetalpbalkan:
@aznashwan apparently the `SendAzurePostRequest` does not return a response body. you can change it to return `([]byte, management.OperationID, error)` if you like.

In the rest of the Azure REST APIs, the POST requests usually take some time and run in the background and expect user to poll the API with `management.OperationID` for operation status, however apparently all SQL Azure APIs are effective immediately when server responds with a result to the REST call...

aznashwan:
@ahmetalpbalkan : thanks for the review.

I've addressed all your points and awaiting any further input you guys might have :smile: 

PS: as a side-note; I was in a conflict between your preferring the names used by the API and what golint suggests; I decided to go ahead and rename [this](https://github.com/Azure/azure-sdk-for-go/pull/171/files#diff-fdf8224c399357a019e9146a15741be0R57) field so as to not curse the project with a red CI build forever.

management/sql/client.go
```diff
@@ -0,0 +1,163 @@
...
+
+	url := fmt.Sprintf(azureDeleteDatabaseServerURL, name)
+	_, err := c.mgmtClient.SendAzureDeleteRequest(url)
+	return err
+}
+
+// CreateDatabase creates a new Microsoft Azure SQL Database on the given database server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+func (c SqlDatabaseClient) CreateDatabase(params DatabaseCreateParams, server string) error {
```
ahmetalpbalkan: server should come first, params next

management/sql/client.go
```diff
@@ -0,0 +1,163 @@
...
+	if params.Name == "" {
```
ahmetalpbalkan: do not check for `params.name`, it's not required to make the request, it's rather required to do the operation successfully and that's what API server checks in the input as well. so leave the check to the API server when you can.

management/sql/client.go
```diff
@@ -0,0 +1,163 @@
...
+
+	target := fmt.Sprintf(azureCreateDatabaseURL, server)
+	_, err = c.mgmtClient.SendAzurePostRequest(target, req)
+	return err
+}
+
+// GetDatabase gets the details for an Azure SQL Database.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505708.aspx
+func (c SqlDatabaseClient) GetDatabase(database, server string) (ServiceResource, error) {
```
ahmetalpbalkan: `server ` first.

management/sql/client.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	err = xml.Unmarshal(resp, &databases)
+	return databases, err
+}
+
+// DeleteDatabase deletes the Azure SQL Database off the given server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505705.aspx
+func (c SqlDatabaseClient) DeleteDatabase(database, server string) error {
```
ahmetalpbalkan: `server` first.

management/sql/client.go
```diff
@@ -0,0 +1,163 @@
...
+	var name string
+	err = xml.Unmarshal(resp, &name)
+
+	return name, err
+}
+
+// ListServers retrieves the Azure SQL Database servers for this subscription.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505702.aspx
+func (c SqlDatabaseClient) ListServers() (ListDatabaseServersResponse, error) {
```
ahmetalpbalkan: should be `ListServersResponse`

management/sql/entities.go
```diff
@@ -0,0 +1,78 @@
...
+	Name                     string
+	AdministratorLogin       string
+	Location                 string
+	FullyQualifiedDomainName string
+	Version                  string
+	State                    string
+}
+
+type ListDatabaseServersResponse struct {
+	DatabaseServers []DatabaseServer `xml:"Server"`
```
ahmetalpbalkan: "Server" should be "ServiceResource" otherwise it will not work:

http://play.golang.org/p/4_tpfVF6dK it doesn't match the response.

ahmetalpbalkan:
@aznashwan thanks a lot for the updates, I just left a few more comments. The Id→ID conversion is fine, we do it in other places as well. :-)

management/sql/entities.go
ahmetalpbalkan: Hmm actually many of the fields in DatabaseServer does not fit what's returned from the API?

aznashwan: I don't use ?countview=generic as it offers a lot more information that really needed (in particular stuff like Type and the Links).
Without that param; the response looks like the second example from that page; which is the one I'm targeting.

aznashwan:
@ahmetalpbalkan : thanks a lot for the time!
I've done the necessary changes for both this and the POST response pull and and looking forward to seeing the code in and usable :smile:

management/sql/entities.go
```diff
@@ -0,0 +1,78 @@
...
+	XMLName            xml.Name `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
+	Name               string
+	Edition            string `xml:",omitempty"`
+	CollationName      string `xml:",omitempty"`
+	MaxSizeBytes       int64  `xml:",omitempty"`
+	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
+}
+
+// ServiceResource represents the set of parameters obtained from a database
+// get or lost call.
```
ahmetalpbalkan: `list`

management/sql/entities.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
```
ahmetalpbalkan: Link points to `Create Database`. :)

management/sql/entities.go
```diff
@@ -0,0 +1,78 @@
...
+	State              string
+	SelfLink           string
+	Edition            string
+	CollationName      string
+	MaxSizeBytes       int64
+	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
+}
+
+type ListDatabasesResponse struct {
+	XMLName          xml.Name          `xml:"http://schemas.microsoft.com/windowsazure ServiceResources"`
```
ahmetalpbalkan: All these schema URIs and xml.Name(s) are redundant. We prefer to not to return XMLName to the user where possible:  http://play.golang.org/p/Za2dGTHf3G

This struct probably should just be like:

```go
type ListDatabasesResponse struct {
	ServiceResources []ServiceResource `xml:"ServiceResource"`
}
```

management/sql/entities.go
aznashwan: I'm sorry; but I don't fully get what you mean here.

management/sql/entities.go
```diff
@@ -0,0 +1,78 @@
...
+	MaxSizeBytes       int64  `xml:",omitempty"`
+	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
+}
+
+// ServiceResource represents the set of parameters obtained from a database
+// get or lost call.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505701.aspx
+type ServiceResource struct {
+	XMLName            xml.Name `xml:"http://schemas.microsoft.com/windowsazure ServiceResource"`
```
ahmetalpbalkan: XMLName field redundant.

management/sql/entities.go
aznashwan: sorry about that; I noticed the type now

management/sql/entities.go
aznashwan: Thanks for the example. Fixed it now.

ahmetalpbalkan:
Looks great @aznashwan. Did you have a chance to test each of these calls? I'm just verbally asking because sometimes API doesn't work as it's documented (e.g. Ordering of request fields sometimes matter). Does Terraform use this to verify functionality of this package?

aznashwan:
@ahmetalpbalkan: I will get to actually writing the resources for Terraform which use this portion of the SDK in a moment.
I have tested each method "manually" (I can share the code snippets if you like), even after all the changes, and I can vouch they work.

ahmetalpbalkan:
Awesome, thanks.

PR #173, opened 2015-06-16
Added POST method which also returns the response to management.Client.

aznashwan:
Required for https://github.com/Azure/azure-sdk-for-go/pull/171

### DIFF
--- a/management/http.go
+++ b/management/http.go
@@ -28,6 +28,15 @@ func (client client) SendAzurePostRequest(url string, data []byte) (OperationID,
 	return client.doAzureOperation("POST", url, "", data)
 }
 
+func (client client) SendAzurePostRequestWithReturnedResponse(url string, data []byte) ([]byte, error) {
+	resp, err := client.sendAzureRequest("POST", url, "", data)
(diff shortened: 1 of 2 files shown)


management/http.go
```diff
@@ -28,6 +28,21 @@ func (client client) SendAzurePostRequest(url string, data []byte) (OperationID,
 	return client.doAzureOperation("POST", url, "", data)
 }
 
+func (client client) SendAzurePostRequestWithReturnedResponse(url string, data []byte) (OperationID, []byte, error) {
+	resp, err := client.sendAzureRequest("POST", url, "", data)
+	if err != nil {
+		return "", nil, err
+	}
+
+	opID, err := getOperationID(resp)
```
ahmetalpbalkan: hmm if you're going to create a new method that's only going to be used by SQL Azure, no need for returning `OperationID`, because operation is not asynchronous and we return the operation ID in the error if that happens to be the case.

aznashwan: Fixed.

ahmetalpbalkan:
LGTM thanks.

PR #174, opened 2015-06-16
Added Database Update operation and waiting features.

aznashwan:
Added database update operation.

Also added some waiting features on more time-consuming operations, as well as a wait for database creation function.

### DIFF
--- a/management/sql/client.go
+++ b/management/sql/client.go
@@ -3,6 +3,7 @@ package sql
 import (
 	"encoding/xml"
 	"fmt"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/management"
 )
@@ -16,9 +17,12 @@ const (
 	azureCreateDatabaseURL = "services/sqlservers/servers/%s/databases"
 	azureGetDatabaseURL    = "/services/sqlservers/servers/%s/databases/%s"
 	azureListDatabasesURL  = "/services/sqlservers/servers/%s/databases?contentview=generic"
+	azureUpdateDatabaseURL = "/services/sqlservers/servers/%s/databases/%s"
 	azureDeleteDatabaseURL = "/services/sqlservers/servers/%s/databases/%s"
 
 	errParamNotSpecified = "Parameter %s was not specified."
+
+	DatabaseStateCreating = "Creating"
 )
 
 // SqlDatabaseClient defines various database CRUD operations.
@@ -99,6 +103,28 @@ func (c SqlDatabaseClient) CreateDatabase(server string, params DatabaseCreatePa
 	return err
 }
 
+// WaitForDatabaseCreation is a helper method which waits
+// for the creation of the database on the given server.
+func (c SqlDatabaseClient) WaitForDatabaseCreation(
+	server, database string,
+	cancel chan struct{}) error {
+	for {
+		stat, err := c.GetDatabase(server, database)
+		if err != nil {
+			return err
+		}
+		if stat.State != DatabaseStateCreating {
+			return nil
+		}
+
+		select {
+		case <-time.After(management.DefaultOperationPollInterval):
+		case <-cancel:
+			return management.ErrOperationCancelled
+		}
+	}
+}
+
 // GetDatabase gets the details for an Azure SQL Database.
 //
 // https://msdn.microsoft.com/en-us/library/azure/dn505708.aspx
@@ -106,10 +132,10 @@ func (c SqlDatabaseClient) GetDatabase(server, database string) (ServiceResource
 	var db ServiceResource
 
 	if database == "" {
-		return db, fmt.Errorf(errParamNotSpecified, "database name")
+		return db, fmt.Errorf(errParamNotSpecified, "database")
 	}
 	if server == "" {
-		return db, fmt.Errorf(errParamNotSpecified, "server name")
+		return db, fmt.Errorf(errParamNotSpecified, "server")
 	}
 
 	url := fmt.Sprintf(azureGetDatabaseURL, server, database)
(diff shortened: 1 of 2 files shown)


management/sql/client.go
```diff
@@ -99,6 +105,28 @@ func (c SqlDatabaseClient) CreateDatabase(server string, params DatabaseCreatePa
...
+		default:
+			stat, err := c.GetDatabase(server, database)
+			if err != nil {
+				return err
+			}
+			if stat.State != DatabaseStateCreating {
+				return nil
+			}
+		}
+		time.Sleep(OperationPollInterval)
```
ahmetalpbalkan: use `<- time.After(interval)`

management/sql/client.go
```diff
@@ -16,9 +17,14 @@ const (
...
 	azureGetDatabaseURL    = "/services/sqlservers/servers/%s/databases/%s"
 	azureListDatabasesURL  = "/services/sqlservers/servers/%s/databases?contentview=generic"
+	azureUpdateDatabaseURL = "/services/sqlservers/servers/%s/databases/%s"
 	azureDeleteDatabaseURL = "/services/sqlservers/servers/%s/databases/%s"
 
 	errParamNotSpecified = "Parameter %s was not specified."
+
+	DatabaseStateCreating = "Creating"
+
+	OperationPollInterval = 1 * time.Second
```
ahmetalpbalkan: 1s is way too frequent. Our API throttles requests (I believe 1000 per subscription per hour). So consider making it 30s (if available/accessible, use the default in the ../management pkg)

management/sql/client.go
```diff
@@ -141,6 +169,28 @@ func (c SqlDatabaseClient) ListDatabases(server string) (ListDatabasesResponse,
...
 }
 
+// Updates details of the given SQL Database off the given server.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505718.aspx
+func (c SqlDatabaseClient) UpdateDatabase(
+	server, database string,
+	params ServiceResourceUpdateParams) (management.OperationID, error) {
+	if database == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "database name")
```
ahmetalpbalkan: `database name` --> `database` (they should match to argument names)

management/sql/client.go
```diff
@@ -141,6 +169,28 @@ func (c SqlDatabaseClient) ListDatabases(server string) (ListDatabasesResponse,
...
+	}
+	if server == "" {
+		return "", fmt.Errorf(errParamNotSpecified, "server name")
```
ahmetalpbalkan: `server name` --> `server` (they should match to argument names)

management/sql/client.go
```diff
@@ -141,6 +169,28 @@ func (c SqlDatabaseClient) ListDatabases(server string) (ListDatabasesResponse,
...
+	}
+
+	url := fmt.Sprintf(azureUpdateDatabaseURL, server, database)
+	req, err := xml.Marshal(params)
+	if err != nil {
+		return "", err
+	}
+
+	return c.mgmtClient.SendAzurePutRequest(url, "", req)
```
ahmetalpbalkan: I can't see it in the docs, but I think the Content-Type probably should be text/xml.

management/sql/entities.go
```diff
@@ -71,6 +71,18 @@ type ServiceResource struct {
 	ServiceObjectiveID string `xml:"ServiceObjectiveId,omitempty"`
 }
 
+// ServiceResourceUpdateParams represents the set of parameters available
+// for a database service update operation.
+//
+// https://msdn.microsoft.com/en-us/library/azure/dn505718.aspx
+type ServiceResourceUpdateParams struct {
```
ahmetalpbalkan: move this below or up because currently ServiceResource and ListDatabasesResponse are related and they probably better sit next to each other.

management/sql/client.go
aznashwan: Do you mean something like:
```go
timeout := time.After(interval)
for {
    select {
    case <- cancel: ...
    case <- timeout:
        GetDatabase and check...
}
```?

**Comment by ahmetalpbalkan** [comment]
My comment was, can you make the polling utility method look like this with `<-time.After(d)`


image

Sorry had to take screenshot I was on mobile.

**Comment by aznashwan** [comment]
@ahmetalpbalkan: good point. Done and tested.

File: management/sql/client.go [hunk 7]
```diff
@@ -99,6 +103,27 @@ func (c SqlDatabaseClient) CreateDatabase(server string, params DatabaseCreatePa
...
+// WaitForDatabaseCreation is a helper method which waits
+// for the creation of the database on the given server.
+func (c SqlDatabaseClient) WaitForDatabaseCreation(
+	server, database string,
+	cancel chan struct{}) error {
+	for {
+		select {
+		case <-cancel:
+			return management.ErrOperationCancelled
+		case <-time.After(management.DefaultOperationPollInterval):
```
ahmetalpbalkan: Now you introduced a delay (30sec) even though maybe the operation was already completed ok the first try :-) Please try to stick to the reference implementation I sent in this case.

ahmetalpbalkan:
@aznashwan can you please limit PRs to only 1 commit? (Can you squash 'em?)

management/sql/client.go
aznashwan: Good point. Fixed.

aznashwan:
@ahmetalpbalkan sorry for all the randomness; it's a little late around these parts.

Tested and squashed once and for all, I hope.

ahmetalpbalkan:
Thanks. LGTM. Waiting CI to finish.

PR #175, opened 2015-06-17
Fixed Database Service update request struct.

aznashwan:
I realized just now that I got the fields of the create and update request structs mixed up a little.
 
On the bright side; the rest of everything got a much more thorough testthrough as Terraform provisions all the SQL's like a champ now :smile:


ahmetalpbalkan:
LGTM.

PR #176, opened 2015-06-18
Added affinity group management support.

aznashwan:
This PR depends on the SQL client naming fix: https://github.com/Azure/azure-sdk-for-go/pull/177

### DIFF
--- a/management/affinitygroup/client.go
+++ b/management/affinitygroup/client.go
@@ -1,0 +1,131 @@
+package affinitygroup
+
+import (
+	"encoding/base64"
+	"encoding/xml"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/management"
+)
+
+const (
+	azureCreateAffinityGroupURL = "/affinitygroups"
+	azureGetAffinityGroupURL    = "/affinitygroups/%s"
+	azureListAffinityGroupsURL  = "/affinitygroups"
+	azureUpdateAffinityGroupURL = "/affinitygroups/%s"
+	azureDeleteAffinityGroupURL = "/affinitygroups/%s"
+
+	errParameterNotSpecified = "Parameter %s not specified."
+)
+
+// AffinityGroupClient simply contains a management.Client and has
+// methods for doing all affinity group-related API calls to Azure.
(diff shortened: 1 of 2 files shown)


management/affinitygroups/client.go
```diff
@@ -0,0 +1,130 @@
+package affinitygroups
```
ahmetalpbalkan: package name must be singular.

ahmetalpbalkan:
I'm not sure how it passed the build earlier, but looks like:

    management/sql/client.go:30:6: type SqlDatabaseClient should be SQLDatabaseClient

can you fix this in this PR as well, otherwise this will keep failing too.

management/affinitygroups/client.go
```diff
@@ -0,0 +1,130 @@
...
+
+// NewClient returns an AffinityGroupClient with the given management.Client.
+func NewClient(mgmtClient management.Client) AffinityGroupClient {
+	return AffinityGroupClient{mgmtClient}
+}
+
+// CreateAffinityGroup creates a new affinity group.
+//
+// https://msdn.microsoft.com/en-us/library/azure/gg715317.aspx
+func (c AffinityGroupClient) CreateAffinityGroup(params CreateAffinityGroupParams) (management.OperationID, error) {
```
ahmetalpbalkan: I believe affinity groups are created immediately, we return operationIDs when we kick off a long-running operation, like Create Deployment that returns HTTP 202 Accepted and then expect users to poll on the operation ID for completion.

I don't think that's the case here.

aznashwan:
@ahmetalpbalkan: fixed

aznashwan:
@ahmetalpbalkan: I split up the SQL client naming fix: https://github.com/Azure/azure-sdk-for-go/pull/177

ahmetalpbalkan:
@aznashwan there's merge conflicts now due to commit `167d321`, can you get rid of that one in this PR and re-push to your repo?

aznashwan:
@ahmetalpbalkan: Should be good now. Thanks for your patience :smile:

ahmetalpbalkan:
Thanks LGTM.

PR #177, opened 2015-06-19
Fixed SQL Client naming scheme.

aznashwan:


management/sql/client.go
```diff
@@ -25,22 +25,22 @@ const (
 	DatabaseStateCreating = "Creating"
 )
 
-// SqlDatabaseClient defines various database CRUD operations.
+// Client defines various database CRUD operations.
```
ahmetalpbalkan: This should be SQLClient just to be consistent with other packages.

ahmetalpbalkan: Or SQLDatabaseClient

aznashwan: Will do.
However; I must warn you that [on my setup, at least], go lint cries that using it as sql.SQLDatabaseClient in other packages is needless repetition; that's why I made it into just Client in the first place...

aznashwan:
@ahmetalpbalkan fixed; and the CI seems happy now :smile:

PR #179, opened 2015-06-26
Pre-alpha ARM Storage package

brendandixon:
This is the first cut of the ARM Storage work. At present, some calls fail because the Swagger used to generate the code has errors. The client.go and models.go files are generated (as noted in their front-matter). common.go and token.go are helper files.


ahmetalpbalkan:
@brendandixon this is still experimental / incomplete. Can't merge to here, also I can't do the preliminary review here as the repo have far too many watchers and they will be getting emails about all those. Please send the PR to your own repository and cc us there.

PR #181, opened 2015-07-31
Add support for creating Clients from a previously read settings file

catsby:
This adds support for initializing a `client` from the contents of a settings file that's stored in memory. At present, `client`s are created by supplying a file path to a settings file, but for my use case I have the contents of the file stored elsewhere, either encrypted in a database or something like [HashiCorp's Vault][1]

I originally tried modifying `ClientFromPublishSettingsFileWithConfig` to check the `filePath` parameter supplied, to see if it's valid XML and not a file path, but settled on duplicating and modifying the `ClientFromPublishSettingsFile` and `ClientFromPublishSettingsFileWithConfig` methods into `ClientFromPublishSettingsData` and `ClientFromPublishSettingsDataWithConfig`, instead. There is duplication, but I felt it was simpler to reason about. 

Example usage, borrowing from the snippet in the README:

```go
package main

import (
	"encoding/base64"
	"fmt"
	"io/ioutil"
	"log"

	"github.com/Azure/azure-sdk-for-go/management"
	"github.com/Azure/azure-sdk-for-go/management/hostedservice"
	"github.com/Azure/azure-sdk-for-go/management/virtualmachine"
	"github.com/Azure/azure-sdk-for-go/management/vmutils"
)

func main() {
	dnsName := "cts-settings-data-ex"
	storageAccount := "terraformtesting"
	location := "West US"
	vmSize := "Small"
	vmImage := "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-14_04-LTS-amd64-server-20140724-en-us-30GB"
	userName := "testuser"
	userPassword := "Test123"

	publishSettingsContent, err := ioutil.ReadFile("downloaded.publishsettings")
	if err != nil {
		panic(err)
	}

	client, err := management.ClientFromPublishSettingsData(publishSettingsContent, "")
	if err != nil {
		panic(err)
	}

	log.Println("Client created...")

	// create hosted service
	if err := hostedservice.NewClient(client).CreateHostedService(hostedservice.CreateHostedServiceParameters{
		ServiceName: dnsName,
		Location:    location,
		Label:       base64.StdEncoding.EncodeToString([]byte(dnsName))}); err != nil {
		panic(err)
	}

	log.Println("Service created...")

	// create virtual machine
	role := vmutils.NewVMConfiguration(dnsName, vmSize)
	vmutils.ConfigureDeploymentFromPlatformImage(
		&role,
		vmImage,
		fmt.Sprintf("http://%s.blob.core.windows.net/sdktest/%s.vhd", storageAccount, dnsName),
		"")
	vmutils.ConfigureForLinux(&role, dnsName, userName, userPassword)
	vmutils.ConfigureWithPublicSSH(&role)

	log.Println("Creating ...")
	operationID, err := virtualmachine.NewClient(client).
		CreateDeployment(role, dnsName, virtualmachine.CreateDeploymentOptions{})
	if err != nil {
		panic(err)
	}
	if err := client.WaitForOperation(operationID, nil); err != nil {
		panic(err)
	}
	log.Println("Done!")
}
```

Suggestions on naming and testing are welcome. 
Thanks!

[1]: https://vaultproject.io

### DIFF
--- a/management/publishSettings.go
+++ b/management/publishSettings.go
@@ -21,12 +21,7 @@ func ClientFromPublishSettingsData(settingsData []byte, subscriptionID string) (
 // from https://manage.windowsazure.com/publishsettings.
 // If subscriptionID is left empty, the first subscription in the string is used.
 func ClientFromPublishSettingsDataWithConfig(settingsData []byte, subscriptionID string, config ClientConfig) (client Client, err error) {
-	publishData := publishData{}
-	if err = xml.Unmarshal(settingsData, &publishData); err != nil {
-		return client, err
-	}
-
-	return clientFromPublishData(publishData, subscriptionID, config)
+	return clientFromPublishData(settingsData, subscriptionID, config)
 }
 
 // ClientFromPublishSettingsFile reads a publish settings file downloaded from https://manage.windowsazure.com/publishsettings.
@@ -47,15 +42,15 @@ func ClientFromPublishSettingsFileWithConfig(filePath, subscriptionID string, co
 		return client, err
 	}
 
+	return clientFromPublishData(publishSettingsContent, subscriptionID, config)
+}
+
+func clientFromPublishData(data []byte, subscriptionID string, config ClientConfig) (client Client, err error) {
 	publishData := publishData{}
-	if err = xml.Unmarshal(publishSettingsContent, &publishData); err != nil {
+	if err = xml.Unmarshal(data, &publishData); err != nil {
 		return client, err
 	}
 
-	return clientFromPublishData(publishData, subscriptionID, config)
-}
-
-func clientFromPublishData(publishData publishData, subscriptionID string, config ClientConfig) (client Client, err error) {
 	for _, profile := range publishData.PublishProfiles {
 		for _, sub := range profile.Subscriptions {
 			if sub.ID == subscriptionID || subscriptionID == "" {
--- a/management/publishSettings.go
+++ b/management/publishSettings.go
@@ -26,6 +26,36 @@ func ClientFromPublishSettingsDataWithConfig(settingsData []byte, subscriptionID
 		return client, err
 	}
 
+	return clientFromPublishData(publishData, subscriptionID, config)
+}
+
+// ClientFromPublishSettingsFile reads a publish settings file downloaded from https://manage.windowsazure.com/publishsettings.
+// If subscriptionID is left empty, the first subscription in the file is used.
+func ClientFromPublishSettingsFile(filePath, subscriptionID string) (client Client, err error) {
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@catsby thanks for the pull request, looks like we need the CLA to be signed.

My only comment is you can totally call `ClientFromPublishSettingsData` from `ClientFromPublishSettingsFile` by first reading the file with `ioutil.ReadFile` and then passing the contents.

Having duplicated logic is something we usually really want to avoid as people usually forget maintaining both at the same time.

catsby:
Hey @ahmetalpbalkan thanks for the reply 

> Having duplicated logic is something we usually really want to avoid as people usually forget maintaining both at the same time.

I've refactored some to consolidate and remove duplication, as well as preserve backwards compatibility. I think... can you take a look? Thanks!

ahmetalpbalkan:
@catsby I think we can even remove duplication in `xml.Unmarshal`. For now, by creating another method we have two paths that go like this:

```
func A+-----------+   
                  |   
                  v   
                func c
                  ^   
                  |   
func B+-----------+   
```

However if you could call `ClientFromPublishSettingsData` from `ClientFromPublishSettingsFile` without creating another method (as in `func c` above), we'll have:

```
  +--------------------+           
  |      caller        |           
  +--+------------+----+           
     |            |                
     |            |                
     |            |                
     |            |                
     |            |                
     v            v                
func A+----->func B+------>func c()

```

which really enforces a single path. Both are equally OK, however I find latter easier to read. :-) If you could also fix that, the rest is LGTM.

catsby:
Hey @ahmetalpbalkan , thanks for the continued help here. I hate to be difficult but I'm not sure I follow what you're suggesting, though I did just push 902e8bb to remove some duplication of `xml.Unmarshall`. 

> However if you could call ClientFromPublishSettingsData from ClientFromPublishSettingsFile without creating another method (as in func c above), we'll have

By this, do you mean for `ClientFromPublishSettingsFile` to determine if the `filePath` input is actually a file path, or contents of a file, and then choose `ClientFromPublishSettingsFileWithConfig` or `ClientFromPublishSettingsData` accordingly? I avoided this mostly because I didn't want to change the signature/purpose of any existing methods with this addition, specifically any exported methods. 

With 902e8bb , both `ClientFromPublishSettingsFile` and `ClientFromPublishSettingsData` pass on a `[]byte` slice to `clientFromPublishData`, which does the unmarshalling.

ahmetalpbalkan:
LGTM. Thanks!

catsby:
Thanks so much @ahmetalpbalkan !

PR #182, opened 2015-08-03
Reconcile publishSettings methods

ahmetalpbalkan:
@catsby this is the change I was proposing in the #181. :) Implementation of `ClientFromPublishSettingsDataWithConfig` is already 100% the same with `clientFromPublishData` method and makes it obsolete.


catsby:
Ah, wonderful, thanks for clarifying and notifying me :smile:

PR #185, opened 2015-08-13
Alpha of ARM packages

brendandixon:
Alpha of generated ARM packages


brendandixon:
@ahmetalpbalkan @paulmey @devigned Ok folks. Here's an alpha of the ARM packages.

arm/storage/client.go
```diff
@@ -0,0 +1,721 @@
...
+	SubscriptionId string
+}
+
+func New(subscriptionId string) *StorageManagementClient {
+	return NewWithBaseUri(DefaultBaseUri, subscriptionId)
+}
+
+func NewWithBaseUri(baseUri string, subscriptionId string) *StorageManagementClient {
+	client := &StorageManagementClient{BaseUri: baseUri, SubscriptionId: subscriptionId}
+	client.PollingMode = autorest.PollUntilDuration
```
ahmetalpbalkan: area for little idiomatic go improvement: these can probably go into the struct initializer above.

ahmetalpbalkan:
Here's my first main comment: I see this pattern in pretty much a lot of packages.

* `DefaultBaseUri         = "https://management.azure.com"`
* `NewXXXClientWithBaseUri(subscriptionId string)`
* `NewXXXClientWithBaseUri(baseUri string, subscriptionId string)`

in pretty much all packages. I think this is

0. creates duplication across packages.
0. limits us from changing the method signatures. (will describe below)
0. limits the amount of things user can configure. for instance, `baseUri` is configurable by the user, but polling duration isn't. How about we add a retry logic in the future? How about if the client wants to configure an User-Agent etc?

If I'm calling into 5 different packages, I'm initializing 5 different clients by passing my subscription ID 5 times. This doesn't make sense to me:

```go
sc := storage.New(MY_SUBS_ID)
// sc.doThis();
net := network.New(MY_SUBS_ID)
// net.doThat();
co := compute.New(MY_SUBS_ID)
// co.doThat();
```

I think a better model (and what we follow in the `management/...` package) would be:

```go
azClient := azure.New(MY_SUBS_ID, AND_BUNCH_OF_OTHER_THINGS
                         LIKE_RETRY_POLICY,USER-AGENT,POLLING_INTERVAL_WHATEVER)

net := network.New(azClient)
stg := storage.New(azClient)
cmp := compute.New(azClient)
```

Here's what I just achieved with this:

* All clients initialized exactly the same.
* I can freely more config like `baseUrl`, `pollingInterval`, `userAgent`, `retryPolicy` whatever to the `azClient` and I don't even change individual packages == I don't break a lot of places when I need to change something.
* It's more readable.
* I don't pass subscription ID to 5 different places.
* I don't duplicate fields in `type FooClient struct { autorest.Client , SubscriptionId string, BaseUri string }` etc. in every single client.
* I don't duplicate `DefaultBaseUri`, `DefaultPollingDuration` etc in every single client package.

It would be great if we can fix it this way. @brendandixon if you also agree, I'd like to take this fix because it will be breaking change.

ahmetalpbalkan:
[followup to the comment above] 

We have this model already in `management/...` package.  Please take a look:
* https://github.com/Azure/azure-sdk-for-go/blob/99b5c364c7be7088d138aef0afce0b5a70855b03/management/client.go#L70-L104
* https://github.com/Azure/azure-sdk-for-go/blob/99b5c364c7be7088d138aef0afce0b5a70855b03/management/storageservice/client.go#L24-L26
* https://github.com/Azure/azure-sdk-for-go/blob/99b5c364c7be7088d138aef0afce0b5a70855b03/management/storageservice/entities.go#L10-L12

You can see how tiny the individual `FooClient` types in each package are.

arm/storage/client.go
brendandixon: Not sure I follow. These routines are called by other routines, some of which may provide new values.

brendandixon:
@ahmetalpbalkan We should chat. First, the generator must handle more than just ARM clients. It does not know that these packages are related and, hence, cannot divine the need for a common "base" class.  Second, the go-autorest package provides a "base" class used by all generated classes (autorest.Client) which provides public access to everything and more (e.g. client.PollingDuration, client.PollingMode). It also provides easy-to-use "hooks" by which to provide a modified version of http.Client and more. So, the user has full access to everything. The rest is somewhat imposed on the code by the generator model.

brendandixon:
@ahmetalpbalkan By the way, in my experience, it is generally a poor practice to expose, as individual arguments, things a user may want to configure. It tends to be much too fragile. Better models, and that which I've tried to provide through the go-autorest package, allows users to make modifications w/o the core code having to know all things.

ahmetalpbalkan:
@brendandixon fair enough. If the code generation model enforces this model, we can live with it. I guess it would be nice to have all the client configurations centralized in one place because pretty much nobody just initializes only one client, it's nearly always a mix of those.

I didn't realize all those fields were public, so it's all good as long as it is configurable.

brendandixon:
@ahmetalpbalkan It actually supports configuration well-beyond those few values. Users may provide instances of http.Client that do whatever magic they desire (e.g., handle proxies, implement circuit breakers, log requests / responses). Further the model supports fan-in / out through channels w/o the developer having to writes lots of code.

I could add an initializer / constructor that accepts, along with other values, an instantiated autorest.Client. Further reduction of repetition, given how the generator must work, is best handled via convention (e.g., using a local, well-defined Azure properties file -- which we really need to have and do not (yet) have). Such an approach will never break signatures and reduces the risk of checking in bad things (e.g., keys) to central repositories.

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+}
+
+func NewWithBaseUri(baseUri string, subscriptionId string) *AuthorizationClient {
+	client := &AuthorizationClient{BaseUri: baseUri, SubscriptionId: subscriptionId}
+	client.PollingMode = autorest.PollUntilDuration
+	client.PollingDuration = DefaultPollingDuration
+	return client
+}
+
+////////////////////////////////////////////////////////////////////////////////
```
ahmetalpbalkan: Currently this block looks like a godoc (because it's just above the type declaration). Maybe have a new line after this comment block?

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+	return &ManagementLocksClient{*NewWithBaseUri(baseUri, subscriptionId)}
+}
+
+// CreateOrUpdateAtResourceGroupLevel create or update a management lock at
+// the resource group level.;
+//
+// resourceGroupName is the resource group name. lockName is the lock name.
+// parameters is the management lock parameters.
+func (client *ManagementLocksClient) CreateOrUpdateAtResourceGroupLevel(resourceGroupName string, lockName string, parameters ManagementLock) (result ManagementLock, err error) {
+
```
ahmetalpbalkan: nit: looks like REST call methods have this empty line at the beginning. bug or feature?

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+		"lockName":                  url.QueryEscape(lockName),
+		"parentResourcePath":        parentResourcePath,
+		"resourceGroupName":         url.QueryEscape(resourceGroupName),
+		"resourceName":              url.QueryEscape(resourceName),
+		"resourceProviderNamespace": url.QueryEscape(resourceProviderNamespace),
+		"resourceType":              resourceType,
+		"subscriptionId":            url.QueryEscape(client.SubscriptionId),
+	}
+
+	queryParameters := map[string]interface{}{
```
ahmetalpbalkan: hmm, if you want to save 2 lines on each method you can do something like

    queryParameters := baseQueryParameters()

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+		autorest.WithErrorUnlessOK(),
+		autorest.ByClosing())
+
+	return
+}
+
+func (client *AvailabilitySetsClient) NewDeleteRequest(resourceGroupName string, availabilitySetName string) (*http.Request, error) {
+
+	pathParameters := map[string]interface{}{
+		"availabilitySetName": url.QueryEscape(availabilitySetName),
```
ahmetalpbalkan: does having `url.QueryEscape` inside `WithPathParameters` sound better to you? Looks like all query parameters or path parameters should be escaped anyway, so why do it here?

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+
+// Delete the operation to delete the availability set.;
+//
+// resourceGroupName is the name of the resource group. availabilitySetName is
+// the name of the availability set.
+func (client *AvailabilitySetsClient) Delete(resourceGroupName string, availabilitySetName string) (err error) {
+
+	req, err := client.NewDeleteRequest(resourceGroupName, availabilitySetName)
+	if err != nil {
+		return fmt.Errorf("compute: Failure preparing AvailabilitySetsClient.Delete request (%v)", err)
```
ahmetalpbalkan: you mean "failure creating"?

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+	}
+
+	req, err = autorest.Prepare(
+		req,
+		client.WithAuthorization(),
+		client.WithInspection())
+	if err != nil {
+		return fmt.Errorf("compute: Failure sending AvailabilitySetsClient.Delete request (%v)", err)
```
ahmetalpbalkan: you mean "Failure preparing"?

arm/authorization/client.go
brendandixon: Artifact. Do we prefer no empty line? I can change. I tend to vary on the size of the method.

arm/authorization/client.go
brendandixon: @ahmetalpbalkan I'm sorry, but which line? The type declaration occurs before the two methods. Are the comment-less methods a problem? I tended to direct the comments to the block before the type.

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Cannot be done. The generator does not have sufficient insight. The arguments are declared locally. The "base" lacks knowledge of their presence. Further, in some cases, specific query arguments are necessary which would lead to more complex code in the generator.

arm/compute/client.go
brendandixon: @ahmetalpbalkan The "switch" in the Swagger (i.e., x-ms-skip-url-encoding) applies per argument and, so, cannot be moved into WithPathParameters. It must be applied, at runtime, per parameter.

arm/compute/client.go
brendandixon: @ahmetalpbalkan Not precisely. In the model, an http.Request is *prepared* for use. Creating one never fails.

arm/compute/client.go
brendandixon: @ahmetalpbalkan Yup. I'll fix. Thank you.

arm/authorization/client.go
ahmetalpbalkan: Here's what godoc you're currently generating here looks like :

> image

Either have an empty line after this code block or have proper Godoc for the `type`. :)

arm/authorization/client.go
ahmetalpbalkan: I've never seen it anywhere else. :-) Is this a thing?

arm/compute/client.go
ahmetalpbalkan: ok then if you fix the issue below you'll have `Failure preparing ...` message  thrown from two different places in the code. How can I tell which call failed in that case?

arm/compute/client.go
ahmetalpbalkan: That's interesting. Per HTTP spec you can encode every single query parameter value or path character (excluding `/`, `?`, `&`, `=`, `[`, `]` which are path separators or querystring separators). What's the use case for `x-ms-skip-url-encoding`?

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Got it. Thank you. I failed to see that. The files are long and need some kind of divider. I failed to see that these collide with good. I'll fix.

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Uh, yes. See https://azure.microsoft.com/en-us/documentation/articles/resource-group-lock-resources/.

arm/compute/client.go
brendandixon: @ahmetalpbalkan Doh. Read the comment I added below. :)

arm/compute/client.go
brendandixon: @ahmetalpbalkan URL encoding is *not* idempotent. That is, if a parameter is already encoded, encoding again does "bad things." It is a common problem. The x-ms-skip-url-encoding switch gives the Swagger author the ability to control. No harm, no foul.

jeffwilcox: `%20cool%20`

arm/authorization/client.go
ahmetalpbalkan: oh I was talking about having empty lines after the method signature :)

arm/authorization/client.go
ahmetalpbalkan: @brendandixon if they're indeed different clients and you're having trouble skimming long files why not split different client types into different source files such as:

- compute/availabilitysets.go
- compute/availabilitysets.go
- compute/virtualmachineimages.go
- compute/virtualmachineimages_entities.go

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Ok. I can add empty lines if it matters. It felt like boilerplate whitespace...

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Definitely an option and easy to do.

arm/authorization/client.go
ahmetalpbalkan: I'm saying it would be nice to remove these empty lines; not add:

> image

> 
image

brendandixon: @ahmetalpbalkan Oh! Got it. Easy. Is that more idiomatic?

arm/storage/client.go
ahmetalpbalkan: I mean you can replace: 

```
client := &StorageManagementClient{BaseUri: baseUri, SubscriptionId: subscriptionId}
client.PollingMode = autorest.PollUntilDuration
```

with the following:

```
return &StorageManagementClient{
    BaseUri: baseUri,
    SubscriptionId: subscriptionId,
    Client: autorest.Client{
        PollingMode: autorest.PollUntilDuration,
    }}
```

so you don't mix “assignments” with struct “initializers” and you can just `return` this single statement..

arm/authorization/client.go
ahmetalpbalkan: @brendandixon that would be really nice to have IMO.

brendandixon: @ahmetalpbalkan Will do. I will keep all types in a single, models.go, file, however.

arm/storage/client.go
brendandixon: @ahmetalpbalkan Got it. The pronoun above w/o a referent left the question in my mind.

arm/authorization/client.go
ahmetalpbalkan: I don't know, it's the first time I'm seeing empty lines after method declarations in my life, just seemed unusual.

arm/authorization/client.go
ahmetalpbalkan: fair enough. it's already quite readable as is.

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+
+func (client *ManagementLocksClient) CreateOrUpdateAtSubscriptionLevelRequestPreparer() autorest.Preparer {
+	return autorest.CreatePreparer(
+		autorest.AsJSON(),
+		autorest.AsPut(),
+		autorest.WithBaseURL(client.BaseUri),
+		autorest.WithPath("/subscriptions/{subscriptionId}/providers/Microsoft.Authorization/locks/{lockName}"))
+}
+
+// DeleteAtSubscriptionLevel deletes the management lock of a subscription.;
```
ahmetalpbalkan: Looks like all REST call methods have a trailing `;` in the description.

brendandixon: @ahmetalpbalkan It's a problem in the Swagger. Again, this is *generated* code.

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+		resp,
+		client.ByInspecting(),
+		autorest.WithErrorUnlessOK(),
+		autorest.ByUnmarshallingJSON(&result),
+		autorest.ByClosing())
+
+	return result, err
+}
+
+func (client *ManagementLocksClient) NewCreateOrUpdateAtSubscriptionLevelRequest(lockName string, parameters ManagementLock) (*http.Request, error) {
```
ahmetalpbalkan: So the helper equivalent of this (`CreateOrUpdateAtSubscriptionLevel` method) has some godoc above, and the parameters are all the same. Maybe add godoc for these methods, too? It's obvious, but it'll help if we add golint checks to the repo. (currently "method is missing documentation" checks are suppressed).

brendandixon: @ahmetalpbalkan Yah, I'd thought about that. It would be somewhat redundant, but probably useful.

arm/authorization/client.go
ahmetalpbalkan: @brendandixon I have no way of knowing where this `;` comes from as someone who's just reviewing the code. This might as well just can be some mistake in the generation code or somewhere else, I'm just asking.

arm/authorization/client.go
```diff
@@ -0,0 +1,759 @@
...
+}
+
+// DeleteAtResourceLevel deletes the management lock of a resource or any
+// level below resource.;
+//
+// resourceGroupName is the name of the resource group.
+// resourceProviderNamespace is resource identity. parentResourcePath is
+// resource identity. resourceType is resource identity. resourceName is
+// resource identity. lockName is the name of lock.
+func (client *ManagementLocksClient) DeleteAtResourceLevel(resourceGroupName string, resourceProviderNamespace string, parentResourcePath string, resourceType string, resourceName string, lockName string) (err error) {
```
ahmetalpbalkan: This comment is going to be about named return values, in this case `(err error)`.

Go stdlib uses named returns for documentation purposes, for instance when you return `(bool,bool,string)` it's totally ambiguous, that's why people document it as `(found bool, allowed bool, value string)` ...

I don't think named returns make sense here in this case where `error` is a well known obvious return type. Is it possible to change `(err error)` to `error` in this case.

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+type VirtualMachineImagesClient struct {
+	ComputeManagementClient
+}
+
+func NewVirtualMachineImagesClient(subscriptionId string) *VirtualMachineImagesClient {
+	return NewVirtualMachineImagesClientWithBaseUri(DefaultBaseUri, subscriptionId)
+}
+
+func NewVirtualMachineImagesClientWithBaseUri(baseUri string, subscriptionId string) *VirtualMachineImagesClient {
+	return &VirtualMachineImagesClient{*NewWithBaseUri(baseUri, subscriptionId)}
```
ahmetalpbalkan: Ok we discussed this vaguely before. I don't think returning `*FooClient` has any advantage over returning `FooClient` here... And you're getting into C-style deferencing etc business like this line:

    &VirtualMachineImagesClient{*NewWithBaseUri}

this just looks nasty to me. I think we should.

* either do not use pointers at all (my preference)
* (assuming you got a reason to use pointers) change `type VirtualMachineImagesClient struct` to embed `*ComputeManagementClient`; not `ComputeManagementClient` (otherwise you'll do this deferencing here and it just looks ugly.)

arm/authorization/client.go
brendandixon: @ahmetalpbalkan Yes. Will change.

arm/authorization/client.go
ahmetalpbalkan: I think even something like:

```
// NewCreateOrUpdateAtSubscriptionLevelRequest returns the HTTP request used to
// make the CreateOrUpdateAtSubscriptionLevel.
```

is very useful

arm/compute/client.go
brendandixon: @ahmetalpbalkan The type must be a pointer. I will change the embedded type.

arm/authorization/client.go
brendandixon: @ahmetalpbalkan That's along the lines I was considering.

arm/compute/client.go
ahmetalpbalkan: @brendandixon It would be nice to know to reason behind the choice for future reference. Why do we go with pointers?

brendandixon: @ahmetalpbalkan Magic. :)

Actually, it is how the autorest.Client both provides defaults and allows easy overrides. If the overridden fields is not set, it sets it when first accessed. An alternative approach, which I considered, would return the default and not set it. That is, often, a dangerous pattern. Pointers are not bad when used correctly.

Let me think about this a point and reconsider what I believe to be best and safe.

ahmetalpbalkan: @brendandixon it's a common practice to define everything on pointer types and return pointers as long as you're not returning `nil`, I see no problem with that, it's just the gen'd code gets confusing but I think when you change the embedded type it will be ok.

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+func (client *VirtualMachineImagesClient) ListOffersRequestPreparer() autorest.Preparer {
+	return autorest.CreatePreparer(
+		autorest.AsJSON(),
+		autorest.AsGet(),
+		autorest.WithBaseURL(client.BaseUri),
+		autorest.WithPath("/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/publishers/{publisherName}/artifacttypes/vmimage/offers"))
+}
+
+// ListPublishers gets a list of virtual machine image publishers.;
+//
```
ahmetalpbalkan: no biggie: Looks like when the method has no parameters to document, the godoc is followed by an extra empty comment like:

```
// MethodName description lorem ipsum.
//
func (....
```

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+	return autorest.CreatePreparer(
+		autorest.AsJSON(),
+		autorest.AsGet(),
+		autorest.WithBaseURL(client.BaseUri),
+		autorest.WithPath("/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/publishers"))
+}
+
+// ListSkus gets a list of virtual machine image skus.;
+//
+func (client *VirtualMachineImagesClient) ListSkus(location string, publisherName string, offer string) (result []VirtualMachineImageResource, err error) {
```
ahmetalpbalkan: I think `Sku` string comes from swagger definitions. Can you please convey the feedback they should probably use `SKU` over `Sku`? :)

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+		autorest.AsJSON(),
+		autorest.AsGet(),
+		autorest.WithBaseURL(client.BaseUri),
+		autorest.WithPath("/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/publishers/{publisherName}/artifacttypes/vmimage/offers/{offer}/skus"))
+}
+
+// List gets a list of virtual machine images.;
+//
+// filter is the filter to apply on the operation.
+func (client *VirtualMachineImagesClient) List(location string, publisherName string, offer string, skus string, filter string, top int, orderby string) (result []VirtualMachineImageResource, err error) {
```
ahmetalpbalkan: How does user do pagination on this `List` call? Assuming they don't know the total count, they're still forced to provide a `top int` argument, so I'm not sure how they can possibly discover if there is a second page or not.

arm/authorization/models.go
```diff
@@ -0,0 +1,44 @@
...
+
+type LockLevel string
+
+const (
+	CanNotDelete LockLevel = "CanNotDelete"
+	NotSpecified LockLevel = "NotSpecified"
+	ReadOnly     LockLevel = "ReadOnly"
+)
+
+// Management lock information.
```
ahmetalpbalkan: This comment should definitely start like per godoc.

    // ManagementLock ...

arm/compute/client.go
brendandixon: @ahmetalpbalkan Maybe. I need to check the Swagger name, but the core generator (code outside my code) *tends* to screw around with names to Pascal / Camel case them. To not do so in this case requires a dictionary of "names that should be capitalized" or decoration in the Swagger on how to transform / not transform the name. Make sense?

arm/compute/client.go
brendandixon: @ahmetalpbalkan Great question. Unimplemented feature at the moment. :) I have the model in my head and it should be natural:

  l, _ := vmic.ListOffers(xxxx)
  l, _ := l.NextPage()

Or something like this.

arm/compute/client.go
ahmetalpbalkan: @brendandixon golint usually complains if you use for instance `Url` instead of `URL` or `SomeId` instead of `SomeID`. It has a dictionary I believe and the reason it's not failing on this PR is because `.travis.yml` does not include this package. (if you update .travis.yml, you'll probably encounter many errors).

To make the public facing type/field names Go-idiomatic, I went through the entire codebase and fixed all such  mixed-case abbreviations manually... Otherwise golint will just complain.

arm/authorization/models.go
brendandixon: @ahmetalpbalkan Hmm, not sure I agree. The comment as is does not affect godoc. I will examine and consider.

arm/compute/client.go
```diff
@@ -0,0 +1,2031 @@
...
+	if err != nil {
+		return result, fmt.Errorf("compute: Failure sending VirtualMachineExtensionImagesClient.ListVersions request (%v)", err)
+	}
+
+	resp, err := autorest.SendWithSender(client, req)
+	if err != nil {
+		return result, fmt.Errorf("compute: Failure sending VirtualMachineExtensionImagesClient.ListVersions request (%v)", err)
+	}
+
+	result = []VirtualMachineImageResource{}
```
ahmetalpbalkan: I think there's a bit of difference between 

    result = []VirtualMachineImageResource{}

and 

    var result []VirtualMachineImageResource

. The first one actually initializes something that takes up space in memory (i.e. zero-length slice) whereas the second one is just `nil`-slice. It will not cause any problems here but just to be idiomatic, I'd do `var foo []T`.

arm/compute/client.go
brendandixon: @ahmetalpbalkan So much the worse for golint. :) The problem remains that, without a dictionary or switch, the generator cannot a priori (that is, discern beforehand) what is best. Perhaps I could add a dictionary from golint?

arm/subscriptions/client.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/azure/go-autorest/autorest"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+const (
+	ApiVersion             = "2014-04-01-preview"
+	DefaultBaseUri         = "https://management.azure.com"
+	DefaultPollingDuration = 10 * time.Minute
```
ahmetalpbalkan: do we actually poll `10 minutes` between requests? sounds a bit too high. I believe many other clients default to 30s.

arm/compute/client.go
brendandixon: @ahmetalpbalkan The code is correct as it is. json.Unmarshal *requires* a pointer to an existing struct.

arm/subscriptions/client.go
```diff
@@ -0,0 +1,254 @@
...
+	client.PollingDuration = DefaultPollingDuration
+	return client
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//
+// Subscriptions Client
+//
+////////////////////////////////////////////////////////////////////////////////
+type SubscriptionsClient struct {
```
ahmetalpbalkan: heh this type declaration is super confusing:

    type SubscriptionsClient struct { SubscriptionClient }

whereas this totally makes sense:

    type AvailabilitySetsClient struct {	ComputeManagementClient }

maybe call SubscriptionClient something else? Normally I wouldn't care about this but we're exporting both types and it impairs discoverability from the GoDoc interface.

arm/subscriptions/client.go
brendandixon: @ahmetalpbalkan It is something users can tune as needed. In my experiments with storage.Create, polling for five minutes, at times, did not return with 200 OK. Note, polling here is applied (by default) only for 202 Accepted. Further, the initial interval given by ARM providers is a minimum of 25s with a maximum of 900s (15 minutes).

arm/subscriptions/client.go
brendandixon: @ahmetalpbalkan It is an artifact of generation. The generator created both from the names provided by the Swagger. The outer name is derived from the service (in this, case Subscriptions) and the inner name comes from the Swagger (see https://github.com/Azure/azure-rest-api-specs/blob/master/arm-subscriptions/2014-04-01-preview/swagger/subscriptions.json).

arm/subscriptions/client.go
ahmetalpbalkan: @brendandixon please talk to ADXSDK folks about what they use as defaults. I agree that storage account creation is not so fast but this `10m` is far too defensive and definitely shouldn't be the default in all packages...

In my experiments, storage account creation takes less than 5 minutes nearly always, so I don't think this also definitely shouldn't be 10 minutes.

arm/compute/client.go
ahmetalpbalkan: @brendandixon I'm not sure what you're talking about. Both expressions I listed above are of the same types. Check this example out: http://play.golang.org/p/P0yLMrZYfN

arm/subscriptions/client.go
brendandixon: @ahmetalpbalkan Perhaps the best approach is to use the Retry-After interval as a hint and use a multiplier of it. A sensibly written backend will provide a longer retry interval when under load...hence, leading to a longer polling time? This *only* affects the convenience methods. Developers have full control (again without having to write much code).

ahmetalpbalkan: @brendandixon Makes sense to use if there's a `Retry-After` header. However if there's not, as a user 90% of the time I don't want to adjust the polling interval at all... So we should have good defaults here.

arm/subscriptions/client.go
ahmetalpbalkan: Does the  user directly initialize or call methods of `SubscriptionClient` or `ComputeManagementClient` (based on the examples above)? If not maybe we can un-export these types and it would cause less confusion.

arm/subscriptions/client.go
brendandixon: @ahmetalpbalkan There is no such thing as "a good default." Only lucky values. In "the real world" it all varies too much. Ask Netflix. :) I will give this some more thought.

ahmetalpbalkan: @brendandixon anything is better than 10 minute in this case. I use tons of CLIs every day and most of them have good defaults. :-) It's not hard to come up with something. We've been in this SDK business for a few years now and pretty sure the SDK team has some know-how (that might work better than 10 minutes) which served them well in the past.

arm/compute/client.go
brendandixon: @ahmetalpbalkan And they're the same size. See http://play.golang.org/p/Rho7YDbGDE.

Leaving it as it is leaves the generator less complicated (otherwise we have to add a case to handle this...).

brendandixon: @ahmetalpbalkan I was actually surprised that json.Unmarshal accepted []Foo form. Now that I see both are the same size, it is apparent that []Foo is equivalent to []Foo{}.

arm/subscriptions/client.go
brendandixon: @ahmetalpbalkan I actually had a version where I attempted such. And I'm forgetting why, or if enough has changed to merit another attempt, but it had problems...

ahmetalpbalkan: No big deal for now. If the user does not use these types directly, we can most likely un-export them later on. :white_check_mark:

arm/subscriptions/client.go
ahmetalpbalkan: @brendandixon I'm recalling all my comments in this thread. For some reason I read `PollingDuration` as `PollingInterval` and 10 minutes sounded too much. Our long tail in resources like storage account or VMs can sometimes take up to 25-30 minutes and yet would succeed.

So my new comment now is: 10 minutes is too short for most VMs to be created. :smile: 

p.s. sorry for the misunderstanding here.

arm/compute/client.go
ahmetalpbalkan: @brendandixon umm, equivalent, not really. :) If you actually compare the two forms using `==` you'll get a `false`. One does an allocation in stack, the other one doesn't. 

Totally off topic but here's a benchmark: http://play.golang.org/p/XBLXdWv9_8 the one that doesn't initialize the empty slice is ~20% faster. Again, doesn't really matter in this case. It's more of a styling argument here.

:white_check_mark:

ahmetalpbalkan:
@brendandixon generally LGTM, feel free to merge at your own discretion.

PR #188, opened 2015-08-19
Promoted GetOSImage from tests to osimage/client

davemcdermid:
Added GetOSImage as a public method in osimage/client. Allows clients to find an image based on specific criteria (e.g. ImageFamily) instead of using an image name. Promoted method from integration tests.

### DIFF
--- a/management/osimage/client.go
+++ b/management/osimage/client.go
@@ -4,7 +4,7 @@ package osimage
 import (
 	"encoding/xml"
 	"errors"
-	
+
 	"github.com/Azure/azure-sdk-for-go/management"
 )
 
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
Hi Dave, thanks for the PR, it looks like this method is a helper involving client-side logic. We deliberately removed such methods from the managament/... package a while ago and we expect users to have their own helpers. Feel free to copy the method to your project and have it there, I will be closing this in sake of consistency. 

Thanks again.

davemcdermid:
Ok, fair enough. Thanks for the reply.

PR #191, opened 2015-08-21
Made examples a package

brendandixon:
@ahmetalpbalkan This should fix the CI build.


arm/main.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/azure/azure-sdk-for-go/arm/examples"
+)
+
+func help() {
+	fmt.Println(`Select an example to run:
+		Client  -- Basic storage account example
```
ahmetalpbalkan: Never seen a subcommand that starts with an uppercase letter. :-)

arm/examples/client.go
```diff
@@ -32,17 +32,17 @@ func (i Inspectors) ByInspecting() autorest.RespondDecorator {
 	}
 }
 
-func main() {
-	if len(os.Args) < 2 {
+func Client(args []string) {
+	if len(args) < 1 {
 		fmt.Println("Please provide a name to check")
-		os.Exit(1)
+		return
```
ahmetalpbalkan: Use log.Fatalf instead of fmt.Print+return

ahmetalpbalkan:
Lgtm. Feel free to merge once you fix the comment above if it makes sense. Log.fatal is mode idiomatic although it doesnt give the same execution flow.

arm/examples/client.go
```diff
@@ -32,17 +32,17 @@ func (i Inspectors) ByInspecting() autorest.RespondDecorator {
...
 }
 
-func main() {
-	if len(os.Args) < 2 {
-		fmt.Println("Please provide a name to check")
-		os.Exit(1)
+func Client(args []string) {
+	if len(args) < 1 {
+		log.Fatalf("Please provide a name to check")
+		return
```
ahmetalpbalkan: You don't need return anymore it already will exit

arm/examples/client.go
```diff
@@ -50,7 +49,7 @@ func main() {
 	spt, err := helpers.NewServicePrincipalTokenFromCredentials(c, azure.AzureResourceManagerScope)
 	if err != nil {
 		fmt.Println(err)
```
ahmetalpbalkan: Ideally you should never call fmt.println in this pkg for errors as it's a very short program and exits on every error. Just log.fatal

arm/examples/storage.go
```diff
@@ -78,13 +74,13 @@ func main() {
 	if err != nil {
 		if sa.Response.StatusCode != 202 {
```
ahmetalpbalkan: Use http.StatusWhatever. I don't know what 202 means :-)

ahmetalpbalkan:
Also spotted some struct initializations in storage example where you could have used struct initializer instead of assignment. Can't comment at that point as github mobile sucks.

arm/examples/storage.go
```diff
@@ -78,13 +74,13 @@ func main() {
...
 			fmt.Printf("Creation of %s.%s failed with err -- %v\n", resourceGroup, name, err)
-			os.Exit(1)
+			return
 		} else {
 			fmt.Printf("Create initiated for %s.%s -- poll %s to check status\n",
```
ahmetalpbalkan: %s/%s is more accurate than %s.%s but fwiw.

brendandixon: @ahmetalpbalkan I'll change it later...

brendandixon:
@ahmetalpbalkan Regarding struct initialization. Point me to specifics and I'll fix later.

ahmetalpbalkan:
Just search for {} in storage.go

brendandixon:
@ahmetalpbalkan Yah, right. Searching for {} returns all those interfaces as well. :) I'll find / fix later.

ahmetalpbalkan:
You have interfaces in arm/examples?

brendandixon:
@ahmetalpbalkan Oh, I thought you meant in arm generally. No worries. I'll fix them.

PR #194, opened 2015-08-27
Corrected HTTP handling

### LINKED ISSUE #193: Confusing error message upon 400 Bad Request
I am using the StorageAccountsClient to create a new Storage Account. It's working pretty well in most cases. However, in case wrong values are passed e.g. wrong location (uswest instead of westus), the SDK throws an error message of the type:

> `autorest.Client:ShouldPoll Response to https://management.azure.com/subscriptions/ddc30f29-****-****-****-********/resourceGroups/vivekrg1/providers/Microsoft.Storage/storageAccounts/vivekst?api-version=2015-05-01-preview requires polling but polling is disabled`

It would be nice for the SDK to bail out with the actual error message e.g. 400 Bad Request. Invalid Location.

Here is the code snippet being used:

	cp := storage.StorageAccountCreateParameters{}
	cp.Location = location
	cp.Properties.AccountType = t

	smc := p.StorageAccountsClient(subID)
	smc.PollingMode = autorest.PollUntilDuration
	smc.PollingDuration = 15 * time.Minute

	sa, err := smc.Create(resGroupName, storageAccountName, cp)
	if err != nil {
		return nil, err
	}
	return &sa, nil

Btw: Thanks for adding the ARM support. Cool stuff.

brendandixon:
Fixed #193.

### DIFF
--- a/arm/authorization/managementlocks.go
+++ b/arm/authorization/managementlocks.go
@@ -56,7 +56,10 @@ func (client ManagementLocksClient) CreateOrUpdateAtResourceGroupLevel(resourceG
 		return result, autorest.NewErrorWithError(err, "authorization.ManagementLocksClient", "CreateOrUpdateAtResourceGroupLevel", "Failure preparing request")
 	}
 
-	resp, err := autorest.SendWithSender(client, req)
+	resp, err := autorest.SendWithSender(
+		client,
+		req,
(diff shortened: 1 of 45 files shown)


brendandixon:
Closing. Travis is very confused. Will submit a new PR.

PR #198, opened 2015-08-31
Exposing UnexpectedStatusCodeError.Got to support (fixes #197)

### LINKED ISSUE #197: Dealing with some benign errors
When a call returns UnexpectedStatusCodeError, there is no programatic way to tell what status code was actually return to let the user code decide how to proceed.

solution proposed is to expose .Got() (returning .got)

An example scenario: when creating a queue that already exist, user code might decide this is fine, but have no way to be sure of that. The Queue service doc say in this case the response is 204, the following code is possible with the suggestion solution:

	qc := az.GetQueueService()

	err = qc.CreateQueue(queueName)
	if storageError, ok := err.(*storage.UnexpectedStatusCodeError); ok {
		if storageError.Got() != http.StatusNoContent {
			panic(err)
		}
	}

	qc.GetMessages(queueName, storage.GetMessagesParameters{})

kenegozi:
see #197 for the reason behind this

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -79,6 +79,11 @@ func (e UnexpectedStatusCodeError) Error() string {
 	return fmt.Sprintf("storage: status code from service response is %s; was expecting %s", got, strings.Join(expected, " or "))
 }
 
+// Got is the actual status code returned by Azure.
+func (e UnexpectedStatusCodeError) Got() int {
+	return e.got
+}
+
 // NewBasicClient constructs a Client with given storage service name and
 // key.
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@kenegozi we can just make `Got` field exported. That'd be easier.

kenegozi:
@ahmetalpbalkan I did not want to make the error object immutable. This feels more future-ready to when it becomes a fully supported GA :)

ahmetalpbalkan:
Thanks for the PR @kenegozi

PR #201, opened 2015-09-02
Queues - get approximate message count (fixes #199)

kenegozi:

PR #202, opened 2015-09-02
Queue metadata with userdefined

kenegozi:
this one include #201 and add support for SetMetadata and to deal with user defined fields.
It is separate as implementation is less obvious so if this is not the way the team wants to go with user-defined data, it would be easier to only accept the other, straight-forward patch.

### DIFF
--- a/storage/queue.go
+++ b/storage/queue.go
@@ -6,6 +6,13 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
+)
+
+const (
+	// casing is per Golang's http.Header canonicalizing the header names.
+	approximateMessagesCountHeader  = "X-Ms-Approximate-Messages-Count"
+	userDefinedMetadataHeaderPrefix = "X-Ms-Meta-"
 )
 
 // QueueServiceClient contains operations for Microsoft Azure Queue Storage
@@ -109,6 +116,77 @@ type PeekMessageResponse struct {
 	ExpirationTime string `xml:"ExpirationTime"`
 	DequeueCount   int    `xml:"DequeueCount"`
 	MessageText    string `xml:"MessageText"`
+}
+
+// QueueMetadataResponse represents user defined metadata and queue
+// properties on a specific queue.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
+type QueueMetadataResponse struct {
+	ApproximateMessageCount int
+	UserDefinedMetadata     map[string]string
+}
+
+// SetMetadata operation sets user-defined metadata on the specified queue.
+// Metadata is associated with the queue as name-value pairs.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179348.aspx
+func (c QueueServiceClient) SetMetadata(name string, metadata map[string]string) error {
+	uri := c.client.getEndpoint(queueServiceName, pathForQueue(name), url.Values{"comp": []string{"metadata"}})
+	headers := c.client.getStandardHeaders()
+	headers["Content-Length"] = "0"
+	for k, v := range metadata {
+		headers[userDefinedMetadataHeaderPrefix+k] = v
+	}
+
+	resp, err := c.client.exec("PUT", uri, headers, nil)
+	if err != nil {
+		return err
+	}
+	defer resp.body.Close()
+
+	return checkRespCode(resp.statusCode, []int{http.StatusNoContent})
+}
+
+// GetMetadata operation retrieves user-defined metadata and queue
+// properties on the specified queue. Metadata is associated with
+// the queue as name-values pairs.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
+//
+// Because the way Golang's http client (and http.Header in particular)
+// canonicalize header names, the returned metadata names would always
+// be all lower case.
+func (c QueueServiceClient) GetMetadata(name string) (QueueMetadataResponse, error) {
+	qm := QueueMetadataResponse{}
+	qm.UserDefinedMetadata = make(map[string]string)
+	uri := c.client.getEndpoint(queueServiceName, pathForQueue(name), url.Values{"comp": []string{"metadata"}})
+	headers := c.client.getStandardHeaders()
+	resp, err := c.client.exec("GET", uri, headers, nil)
+	if err != nil {
+		return qm, err
+	}
+	defer resp.body.Close()
+
+	for k, v := range resp.headers {
+		if len(v) != 1 {
+			return qm, fmt.Errorf("Unexpected number of values (%d) in response header '%s'", len(v), k)
+		}
+
+		value := v[0]
+
+		if k == approximateMessagesCountHeader {
+			qm.ApproximateMessageCount, err = strconv.Atoi(value)
+			if err != nil {
+				return qm, fmt.Errorf("Unexpected value in response header '%s': '%s' ", k, value)
+			}
+		} else if strings.HasPrefix(k, userDefinedMetadataHeaderPrefix) {
+			name := strings.TrimPrefix(k, userDefinedMetadataHeaderPrefix)
+			qm.UserDefinedMetadata[strings.ToLower(name)] = value
+		}
+	}
+
+	return qm, checkRespCode(resp.statusCode, []int{http.StatusOK})
 }
 
 // CreateQueue operation creates a queue under the given account.
--- a/storage/queue_test.go
+++ b/storage/queue_test.go
@@ -1,6 +1,8 @@
 package storage
 
 import (
+	"time"
+
 	chk "gopkg.in/check.v1"
 )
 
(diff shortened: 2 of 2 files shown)


storage/queue.go
```diff
@@ -111,6 +112,96 @@ type PeekMessageResponse struct {
 	MessageText    string `xml:"MessageText"`
 }
 
+// QueueMetadata represents user defined metadata and queue properties on
+// a specific queue.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
+type QueueMetadata struct {
+	ApproximateMessageCount int
+	userDefinedMetadata     map[string]string
```
ahmetalpbalkan: I think you should just export this map (maybe all of the headers, doesn't matter) to the user and they decide what to do with it. Therefore Get/Set below is obsolete. We don't need that many helpers. We're trying to deliberately remove client-side magic from the SDK.

storage/queue.go
```diff
@@ -111,6 +112,96 @@ type PeekMessageResponse struct {
...
+}
+
+// Get retrieve a value for a case insensitive name.
+func (qm *QueueMetadata) Get(name string) (value string) {
+	qm.ensureInit()
+	value, _ = qm.userDefinedMetadata[strings.ToUpper(name)]
```
ahmetalpbalkan: this method will go away anyway but 

1. do not use named returns (like `value string` here)
2. no need to do `val, _ := mapVar[key]`,... you can just do `return mapVar[key]` it works the same :)

storage/queue.go
```diff
@@ -111,6 +112,96 @@ type PeekMessageResponse struct {
...
+// SetMetadata operation sets user-defined metadata on the specified queue.
+// Metadata is associated with the queue as name-value pairs.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
+func (c QueueServiceClient) SetMetadata(name string, qm QueueMetadata) error {
+	uri := c.client.getEndpoint(queueServiceName, pathForQueue(name), url.Values{"comp": []string{"metadata"}})
+	headers := c.client.getStandardHeaders()
+	headers["Content-Length"] = "0"
+	for k, v := range qm.userDefinedMetadata {
+		headers[userDefinedMetadataHeaderPrefix+k] = v
```
ahmetalpbalkan: as a user it's not clear to me if SDK adds the prefix for me or not... Can you please explore what other clients do? Becuase this is one of those “client-side magics” and I may just provide `x-ms-meta-foo` and you'll save it as `x-ms-meta-x-ms-meta-foo`.

ahmetalpbalkan:
> this one include #201

please close #201 then :)

storage/queue.go
kenegozi: other clients transparently add the prefix when sending to the server and remove the prefix when it gets back, so this is following up on them, following the 'least surprising" principle

ahmetalpbalkan: I see. In this case I'm not entirely happy with `SetMetadata` signature.

1. We have a type `QueueMetadata` that's shared in both request and response calls. That's not ideal, although they look similar, they're not equivalent. 
2. I need to create this object before setting metadata, also I'm not sure why.

Here's my suggestion, let's do following

```go
type GetMetadataResponse { Metadata map[string]string }
func GetMetadata(name string) (map[string]string, error)
func SetMetadata(name string, meta map[string]string) error
```

So also in order to set only one metadata key do I need to have a list of all metadata values? Can't I just send a PUT request with only one `x-ms-meta-foo` and API would "merge" the metadata (by creating/overwriting `-foo`)?

storage/queue.go
kenegozi: no problem, but since Go's http client canonicalize the http header names, the (case-sensitive) map is not exactly what the user might expect. I would document this.

ahmetalpbalkan: @kenegozi we can lowercase the keys before putting to the API. We can document this in the SetMetadata call.

kenegozi: for now I kept it as-is (the way to golang canonical works) and documented it. lmk if its fine or you prefer lowercasing

storage/queue.go
```diff
@@ -6,8 +6,13 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 )
 
+// casing is per Golang's http.Header canonicalizing the header names.
+const approximateMessagesCountHeader = "X-Ms-Approximate-Messages-Count"
```
ahmetalpbalkan: you can make this a 

```
const (
   name = val
   name = val
)

``
block

File: storage/queue.go (hunk 3 above)
Comment (kenegozi) [inline storage/queue.go]: That's about where it currently stands. (the type name is QueueMetadataResponse, since it is namespaced under storage. and is only relevant to Queue)

As for PATCH semantics - this azure API do not support it. the PUT request sets whatever is sent and discards all the rest (which is proper for PUT).

File: storage/queue.go [hunk 5]
```diff
@@ -111,6 +116,75 @@ type PeekMessageResponse struct {
...
+		if len(v) != 1 {
+			return qm, fmt.Errorf("Unexpected number of values (%d) in response header '%s'", len(v), k)
+		}
+
+		if k == approximateMessagesCountHeader {
+			qm.ApproximateMessageCount, err = strconv.Atoi(v[0])
+			if err != nil {
+				return qm, fmt.Errorf("Unexpected value in response header '%s': '%s' ", k, v[0])
+			}
+		} else if strings.Index(k, userDefinedMetadataHeaderPrefix) == 0 {
```
ahmetalpbalkan: use `strings.HasPrefix`

storage/queue.go
kenegozi: good idea, done

storage/queue.go
```diff
@@ -111,6 +116,75 @@ type PeekMessageResponse struct {
...
+			qm.UserDefinedMetadata[k[len(userDefinedMetadataHeaderPrefix):]] = v[0]
```
ahmetalpbalkan: now this makes me think if we should just return `http.Header`s  with the user defined metadata keys and use its `.Get` method etc... This notice is actually fine, and people should be testing their code, but this will be hard to change in the future if need be.

As far as I observed so far, our SDKs return lowercase names (as documented in the REST API Docs) for the headers... Therefore: **I think we should lowercase all of them before putting to the map.** This would be less surprising for the user and they'd notice the pattern on the first use and we can sustain this behavior more easily.

Title-case is just very confusing. English is my second language and for instance I can't say I fully get what you mean by the following:

> Title-cased (first letter in upper case, all the rest lower case)

to me, that sounds like `X-ms-meta-foo`; not `X-Ms-Meta-Foo`. I know it's obvious what title-case is but my preference would be making it all lowercase.

storage/queue.go
ahmetalpbalkan: actually  you're even gonna need `strings.TrimLeft` here :)

kenegozi: done

storage/queue.go
kenegozi: The other SDKs do not expose the raw headers. It would be a distraction. The fact that custom key/value metadata is sent over the wire as specialized headers is an implementation detail and it should not leak to the caller.

Lowercase is definitely easier to understand, so I'll switch it to.

storage/queue.go
```diff
@@ -111,6 +118,74 @@ type PeekMessageResponse struct {
...
+			return qm, fmt.Errorf("Unexpected number of values (%d) in response header '%s'", len(v), k)
+		}
+
+		if k == approximateMessagesCountHeader {
+			qm.ApproximateMessageCount, err = strconv.Atoi(v[0])
+			if err != nil {
+				return qm, fmt.Errorf("Unexpected value in response header '%s': '%s' ", k, v[0])
+			}
+		} else if strings.HasPrefix(k, userDefinedMetadataHeaderPrefix) {
+			qm.UserDefinedMetadata[strings.ToLower(k[len(userDefinedMetadataHeaderPrefix):])] = v[0]
```
ahmetalpbalkan: Please use `strings.TrimLeft`... Also do you know how to squash all the commits in the PR into one commit? That would be great.

storage/queue.go
```diff
@@ -111,6 +118,74 @@ type PeekMessageResponse struct {
...
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
+type QueueMetadataResponse struct {
+	ApproximateMessageCount int
+	UserDefinedMetadata     map[string]string
+}
+
+// SetMetadata operation sets user-defined metadata on the specified queue.
+// Metadata is associated with the queue as name-value pairs.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179384.aspx
```
ahmetalpbalkan: The link should be https://msdn.microsoft.com/en-us/library/azure/dd179348.aspx

ahmetalpbalkan:
LGTM, left 2 comments about small nits. It'd be great if you could squash all commits into one and push -f to your branch. Thanks @kenegozi.

storage/queue.go
kenegozi: yeah I put link to Get on both Get and Set actions. Fixed.

storage/queue.go
kenegozi: TrimPrefix. done

ahmetalpbalkan:
LGTM thanks.

PR #203, opened 2015-09-08
Fix the wrong value of storage.ResolvingDNS

bingosummer:


ahmetalpbalkan:
Hi @bingosummer, thanks for the report. The code in `arm` package is auto-generated from API specs, therefore we cannot merge this pull request. The specs for this API lives at: https://github.com/Azure/azure-rest-api-specs/blob/76e30c33f1480a420592c58ed14d1d4e3427d673/arm-storage/2015-05-01-preview/swagger/storage.json#L495

Can you please open an issue over at that repo and get it fixed? Once that's done, we need to re-generate the code here and submit a separate PR. cc: @devigned

bingosummer:
Got it.

ahmetalpbalkan:
closing this as other one got merged. https://github.com/Azure/azure-rest-api-specs/pull/17/files

PR #215, opened 2015-09-18
Including the machine sizes that were missing

Nepomuceno:
Some of the machine sizes supported on Azure were not on the list of machine sizes.


ahmetalpbalkan:
@Nepomuceno Hi and thanks for your pull request. Unfortunately we won't be able to merge it, the code in `arm/` package is auto-generated from API definitions. Fortunately, they are open and you can contribute. There's already an open issue for that at https://github.com/Azure/azure-rest-api-specs/issues/10 , please submit a Pull Request for that repo and next time we re-generate the code, they'll be picked up.

PR #216, opened 2015-09-22
storage/file: Preliminary support for Azure File Service

ahmetalpbalkan:
Support for CreateShare/DeleteShare calls.

@paulmey PTAL.

### DIFF
--- a/storage/blob_test.go
+++ b/storage/blob_test.go
@@ -168,7 +168,7 @@ func (s *StorageBlobSuite) TestContainerExists(c *chk.C) {
 	c.Assert(ok, chk.Equals, true)
 }
 
-func (s *StorageBlobSuite) TestCreateDeleteContainer(c *chk.C) {
+func (s *StorageBlobSuite) TestCreateContainerDeleteContainer(c *chk.C) {
 	cnt := randContainer()
 	cli := getBlobClient(c)
 	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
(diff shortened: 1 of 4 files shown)


storage/blob_test.go
```diff
@@ -168,7 +168,7 @@ func (s *StorageBlobSuite) TestContainerExists(c *chk.C) {
 	c.Assert(ok, chk.Equals, true)
 }
 
-func (s *StorageBlobSuite) TestCreateDeleteContainer(c *chk.C) {
+func (s *StorageBlobSuite) TestCreateContainerDeleteContainer(c *chk.C) {
```
paulmey: Sneaking in some test fixes eh? Well, ok then, but only for this time...

ahmetalpbalkan: I ctrl-f'ed TestCreateContainer and wondered how come I missed this one, turns out it was just naming thing. Better searchability fix.

storage/file_test.go
```diff
@@ -0,0 +1,62 @@
...
+	defer cli.DeleteShare(name)
+
+	// First create
+	ok, err := cli.CreateShareIfNotExists(name)
+	c.Assert(err, chk.IsNil)
+	c.Assert(ok, chk.Equals, true)
+
+	// Second create, should not give errors
+	ok, err = cli.CreateShareIfNotExists(name)
+	c.Assert(err, chk.IsNil)
```
paulmey: do you also want to check the value of `ok`?

ahmetalpbalkan: good catch

ahmetalpbalkan: fixed

paulmey:
LGTM, but the build is failing. Looks like it may not be calculating the `file` endpoint correctly?

ahmetalpbalkan:
nope it used to pass in the earlier commit... I believe it is a network issue in Travis. Kicked off a rebuild.

PR #217, opened 2015-09-24
Azure table implementation

MindFlavor:
### Azure tables
An implementation of Azure Tables REST API using JSON instead of XML where possible. I strove to be aligned with the existing storage code so the usage is pretty straightforward:

```go
err := cli.CreateTable(tn)
err := cli.DeleteTable(tn)
```
I could have mapped the entities to maps (```map[string]interface{}```) like it's done in the python Azure SDK but that would be awkward in golang: json and xml packages are able to handle custom structs using reflection so I did here too. For insert just pass the storage.TableEntity implementing struct and it will be marshalled automatically:
```go
ce := &CustomEntity{Name: "Luke", Surname: "Skywalker", SomeDate: time.Now(), Number: 60, PKey: "pkey", RKey: "5"}

err := cli.InsertOrReplaceEntity(tn, ce)
```
As for unmarshaling you have to specify the type you are expecting:

```go
entries, contToken, err = cli.QueryTableEntities(tn, nil, reflect.TypeOf(ce), 10, "")
````

Note that these methods store in Azure table rows (without hierarchy) so nested structs should not work by design.

There are more examples in the test files. 
 
### storage.TableEntity interface
```go
type TableEntity interface {
	PartitionKey() string
	RowKey() string
	SetPartitionKey(string) error
	SetRowKey(string) error
}
```
Implement this interface to store your structs in Azure Tables.
This interface requires you to specify the partition and row keys as getter/setter. This way you can are free to represent the Azure Table primary key in your struct as you see fit. 

### table tag
The marshaler allows you to avoid persisting in Azure tables certain attributes. Just append the table:- (table minus) tag and it will be ignored. For example:
```go
type CustomEntity struct {
	Name     string `json:"name"`
	Surname  string `json:"surname"`
	SomeDate time.Time
	Number   int
	PKey     string `json:"pk" table:"-"`
	RKey     string `json:"rk" table:"-"`
}
```
In this case the fields ```PKey``` and ```RKey``` won't be stored. This is useful because Azure Tables always require you to have the ```PartitionKey``` and ```RowKey``` fields so if you were to store ```PKey``` and ```RKey``` too you will end up duplicating the information. 
The code will call your ```TableEntity.SetPartitionKey``` and ```TableEntity.SetRowKey``` methods during unmarshal allowing you to restore the ignored fields correctly.

### Comments
The code is very poorly commented. Most methods lack the proper description. I will add them as soon as I can.

### ToDo
* The biggest todo is to return the ```If-Match``` header in the Entity get methods. Without it you cannot use optimistic concurrency in deletes. The ```DeleteEntity``` method supports is anyway so it's just a matter to return it to the GetEntity callers.
* Better support for OData queries. Right now you can specify only the ```$filter``` param and you have to construct it by hand. Azure Tables also support the ```$select``` parameter for projection but I have not inserted the code here. It should be easy though. What would be harder is to implement the odata protocol. There is a project on that here (https://github.com/amsokol/go-odata) but I haven't looked at it yet.


ahmetalpbalkan:
@MindFlavor this is amazing, thanks. I had something going on at https://github.com/ahmetalpbalkan/azure-sdk-for-go/tree/table-sdk mind taking a look at that, too? I remember I got stuck at some request signing problem.

MindFlavor:
Yes, I wrestled with that too. Azure tables are peculiar in that regard and the documentation is somewhat *lacking* (what to url encode and when, what to sign and so on...) :sweat_smile:. I ended up copying the azure SDK for python code (it's lame, I know :wink:). I am sorry I haven't found your fork before, I will look into it as soon as possible.

ahmetalpbalkan:
@MindFlavor I especially suggest taking a look at `storage/table_entity.go`. I allow maps/structs to be table entities as well and this makes InsertEntity function a bit more clearer. I also had a massive code-reorganizing etc over there as well. If you like the model over there, we can merge that first and add your goodness on top perhaps?

gblmarquez:
@MindFlavor Thanks for this contribution I'll look forward to use it.

MindFlavor:
@ahmetalpbalkan I looked at your code and I liked the odata.type handling as tag. Very elegant. We should definitely use it because it describes the entity better. 

I'm still partial towards the interface entity instead of the map/struct. In your implementation you can marshal pretty much anything provided the structs/maps have the ```PrimaryKey```, ```RowKey``` fields. In my implementation you have to implement the ```storage.TableEntity interface``` first. 
I suggest the interface way for following reasons. For the sake of the examples use this struct:

```go 
type MyType struct {
  Name string
  Surname string
  Age int
}
```

1. structs/maps are required to have the ```PrimaryKey```, ```RowKey``` fields. Failing to do so will cause a runtime error (the Azure REST API call will fail). Using interfaces will raise a compile time error, which is better. Our MyType struct lacks the required fields (and methods): in the struct implementation the compilation will go fine but Azure will reject the insert. In the interface implementation you will get a compile time error: ```type MyType doesn't implement storage.TableEntity interface```.
2. Go's duck typing makes easy to add interfaces to preexisting structs. Just add the methods are you are good to go. This way making preexisting struct serializable to Azure tables won't break you code. With the plain struct you will have to refactor adding ```PrimaryKey```, ```RowKey``` fields. Our MyType struct will have to be refactored in order to work:

```go 
type MyType struct {
  Name string
  Surname string
  Age int
  PrimaryKey string // <-- data domain field
  RowKey string // <-- data domain field
}
```
If we use the interface implementation no breaking refactor would be needed, just the addition of the interface receiver's methods. Something like:

```go
func (c *MyType) PartitionKey() string {
	return "primarykey"
}

func (c *MyType) RowKey() string {
	return c.Surname + " " + c.Name // <-- translate from user domain to data domain
}

func (c *MyType) SetPartitionKey(s string) error {
	return nil // nothing to do, PartitionKey is hardcoded and data domain only
}

func (c *MyType) SetRowKey(s string) error {
  f := string.Fields(s) <-- translate from data domain to user domain
  c.Surname = f[0]
  c.Name= f[1]
  return nil
}
```
We have the chance to insulate our code from the data domain.
3. ```PrimaryKey```, ```RowKey``` forces you to store explicitly store the Azure table primary key in struct. This will lead duplication if the pk can be inferred by other fields. With the interface way you can generate the ```PrimaryKey```, ```RowKey``` field at runtime allowing you to store the originating field only. As an example suppose our MyType now has the primary keys in its fields (but with different names):
```go 
type MyType struct {
  Name string
  Surname string
  Age int
  PKey string // <-- user domain field
  RKey string // <-- user domain field
}
```
The resulting struct refactor requires to have:
```go 
type MyType struct {
  Name string
  Surname string
  Age int
  PKey string // <-- user domain field
  RKey string // <-- user domain field
  PrimaryKey string // <- data domain field
  RowKey string // <- data domain field
}
```
While the interface way would be:
```go 
// unchanged, just a tag added
type MyType struct {
  Name string
  Surname string
  Age int
  PKey string 'table:-' // <-- user domain field
  RKey string 'table:-' // <-- user domain field 
}

func (c *MyType) PartitionKey() string {
	return c.PKey // <- from user to data domain
}

func (c *MyType) RowKey() string {
	return c.RKey // <- from user to data domain
}

func (c *MyType) SetPartitionKey(s string) error {
	c.PKey = s // <- from data to user domain
	return nil
}

func (c *MyType) SetRowKey(s string) error {
	c.RKey = s // <- from data to user domain
	return nil
}
```
The difference is in the Azure Table entity. In the first case the entity would be:
```json
{ 
  Name : "",
  Surname : "",
  Age : "",
  PKey : "",
  RKey : "",
  PrimaryKey : "",
  RowKey : ""
}
```
In the second: 
```json
{ 
  Name : "",
  Surname : "",
  Age : "",
  PrimaryKey : "",
  RowKey : ""
}
```
The lack of replication in the second json will lead to less data stored and, perhaps more importantly, less data sent through the wire.

Please let me know what you think.

MindFlavor:
I've changed the return type for []*TableEntity *[]TableEntity to make return values easier to handle.

gblmarquez:
Is this PR will be merged?

ahmetalpbalkan:
@gblmarquez for now we're holding off on merging brand new packages to the `storage` package, mostly because we'll probably have auto-generated code for these packages as well, soon.

If you could go-get this fork, or my fork with the table implementation (or maintain yours), this solution might work for you in the meanwhile. I think this PR is more mature and ready-to-use in that sense.

dstpierre:
What's the state of Table storage, @ahmetalpbalkan will the auto-generated code coming, are you still recommending using @MindFlavor fork?

michaelbironneau:
@MindFlavor I've been trying to use this, which is awesome by the way, but am having authentication issues for `QueryTableEntities()`. Namely, I can make requests using `QueryTables()` fine but any entities query request via `QueryTableEntities()` returns a 403. Is this just me or has something changed wrt request signing for that endpoint?

michaelbironneau:
OK, figured it out. In `QueryTableEntities`, line 76 needs to change from

```
uri += fmt.Sprintf("()?$top=%d", top)
```

to 

```
uri += fmt.Sprintf("?$top=%d", top)
```

Not sure why this is necessary but it resolves the issue I was having earlier.

MindFlavor:
Wow @michaelbironneau that was quick :+1: !! I didn't even got the chance to open github! 

Thank you for the bugfix, I'll push the change ASAP!

ahmetalpbalkan:
It looks like this PR is organically maturing. :-) If it already has users, we can review again and proceed with merging.

storage/table_test.go
```diff
@@ -0,0 +1,298 @@
...
+	const alphanum = "abcdefghijklmnopqrstuvwxyz"
+	var bytes = make([]byte, 32)
+	rand.Read(bytes)
+	for i, b := range bytes {
+		bytes[i] = alphanum[b%byte(len(alphanum))]
+	}
+	return string(bytes)
+}
+
+func testEquality(c *chk.C, c1, c2 *CustomEntity) {
```
ahmetalpbalkan: hi @MindFlavor just started to take a look at this PR. You can use `c.Assert(a, DeepEquals, b)` to eliminate need for this method.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+// TableEntity interface specifies
+// the functions needed to support
+// marshaling and unmarshaling into
+// Azure Tables. The struct must only contain
+// simple types because Azure Tables do not
+// support hierarchy.
+type TableEntity interface {
+	PartitionKey() string
+	RowKey() string
+	SetPartitionKey(string) error
```
ahmetalpbalkan: these `Set`ters probably don't need to return an error. I assume it'll all be in-memory objects?

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+// same function again passing the received ContinuationToken as previousContToken
+// parameter in order to get the following entries. The query parameter
+// is the odata query. To retrieve all the entries pass the empty string.
+// The function returns a pointer to a TableEntity slice, the *ContinuationToken
+// if there are more entries to be returned and an error in case something went
+// wrong.
+//
+// Example:
+// 		entities, cToken, err = tSvc.QueryTableEntities("table", cToken, reflect.TypeOf(entity), 20, "")
+func (c *TableServiceClient) QueryTableEntities(tableName AzureTable, previousContToken *ContinuationToken, retType reflect.Type, top int, query string) (*[]TableEntity, *ContinuationToken, error) {
```
ahmetalpbalkan: any reasons why you return `*[]TableEntity` and not `[]TableEntity`? Even when you get error, you can return "nil" for slices.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+	headers := c.getStandardHeaders()
+
+	headers["Content-Length"] = "0"
+
+	resp, err := c.client.execTable("GET", uri, headers, nil)
+
+	contToken := extractContinuationTokenFromHeaders(resp.headers)
+
+	if err != nil {
+		return nil, contToken, err
```
ahmetalpbalkan: hmm why do you need to return the ContinuationToken when you hit an error? do you expect users to continue iterating?

storage/table.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// TableServiceClient contains operations for Microsoft Azure Table Storage
+// Service.
+type TableServiceClient struct {
+	client Client
+}
+
+// AzureTable is the typedef of the Azure Table name
+type AzureTable string
```
ahmetalpbalkan: most probably you don't this typedef. Just use string for table names. :)

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+	uri := c.client.getEndpoint(tableServiceName, pathForTable(tableName), url.Values{})
+	headers := c.getStandardHeaders()
+
+	buf := new(bytes.Buffer)
+
+	if err := injectPartitionAndRowKeys(entity, buf); err != nil {
+		return err
+	}
+
+	//	log.Printf("request.body == %s", string(buf.Bytes()))
```
ahmetalpbalkan: unused code.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// InsertEntity inserts an entity in the specified table.
+// The function fails if there is an entity with the same
+// PartitionKey and RowKey in the table.
+func (c *TableServiceClient) InsertEntity(tableName AzureTable, entity TableEntity) error {
+	uri := c.client.getEndpoint(tableServiceName, pathForTable(tableName), url.Values{})
+	headers := c.getStandardHeaders()
+
+	buf := new(bytes.Buffer)
```
ahmetalpbalkan: `var b bytes.Buffer` is often more idiomatic, zero value of bytes.Buffer is usable. You just need to refer to it as `&b` when you use it as io.Writer.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+
+	headers := c.getStandardHeaders()
+
+	buf := new(bytes.Buffer)
+
+	if err := injectPartitionAndRowKeys(entity, buf); err != nil {
+		return err
+	}
+
+	//	log.Printf("request.body == %s", string(buf.Bytes()))
```
ahmetalpbalkan: leftover debug comment

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+	return c.insertOrReplaceOrMergeEntity(table, entity, true)
+}
+
+func (c *TableServiceClient) insertOrReplaceOrMergeEntity(table AzureTable, entity TableEntity, mergeInsteadOfReplace bool) error {
+	uri := c.client.getEndpoint(tableServiceName, pathForTable(table), url.Values{})
+	uri += fmt.Sprintf("(PartitionKey='%s',RowKey='%s')", url.QueryEscape(entity.PartitionKey()), url.QueryEscape(entity.RowKey()))
+
+	headers := c.getStandardHeaders()
+
+	buf := new(bytes.Buffer)
```
ahmetalpbalkan: same comment as above, `var b bytes.Buffer` for less allocs.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	//	log.Printf("request.body == %s", string(buf.Bytes()))
+
+	headers["Content-Length"] = fmt.Sprintf("%d", buf.Len())
+
+	var err error
+	var resp *odataResponse
+
+	if mergeInsteadOfReplace {
```
ahmetalpbalkan: looks like you can just have a `var method string` to determine MERGE vs PUT, and just call `c.client.execTable(method, uri, headers, buf)` to reduce repetition.

storage/table_entities.go
```diff
@@ -0,0 +1,379 @@
...
+
+		dec := make(map[string]interface{})
+		if err := json.NewDecoder(buf).Decode(&dec); err != nil {
+			return nil, err
+		}
+
+		var pKey, rKey string
+		// strip pk and rk
+		for key, val := range dec {
+			switch {
```
ahmetalpbalkan: you can do `switch key { ... }`

storage/table_test.go
```diff
@@ -0,0 +1,298 @@
...
+	c.Assert(len(*entries), chk.Equals, 2)
+	c.Assert(contToken, chk.NotNil)
+
+	entries, contToken, err = cli.QueryTableEntities(tn, contToken, reflect.TypeOf(ce), 2, "")
+	c.Assert(err, chk.IsNil)
+	c.Assert(len(*entries), chk.Equals, 1)
+	c.Assert(contToken, chk.IsNil)
+}
+
+func randTable() string {
```
ahmetalpbalkan: hm we have a test prefix for blob containers we create and a test hook that cleans up all test containers before the test begins, we should probably do something similar here. If we ever have a ListTables() function and its tests, the leftovers will impact the tests. This can be fixed later, though.

storage/client.go
```diff
@@ -356,6 +400,66 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
...
+		}
+		// try unmarshal as odata.error json
+		err = json.Unmarshal(respBody, &respToRet.odata)
+		return respToRet, err
+	}
+
+	return respToRet, nil
+}
+
+func (c Client) createSharedKeyLiteTable(url string, headers map[string]string) (string, error) {
```
ahmetalpbalkan: I believe we can just move everything to SharedKeyLite, this ...Table suffix (and `execTable()` method) might not be necessary.

storage/table_test.go
MindFlavor: Sure you are right, I'll remove it.

storage/table_entities.go
MindFlavor: The idea here was to give the implementer the chance of aborting the method if needed (for example if the format is not valid).

Referencing the above example: 

```go
func (c *MyType) PartitionKey() string {
    return "primarykey"
}

func (c *MyType) RowKey() string {
    return c.Surname + " " + c.Name // <-- translate from user domain to data domain
}

func (c *MyType) SetPartitionKey(s string) error {
    return nil // nothing to do, PartitionKey is hardcoded and data domain only
}

func (c *MyType) SetRowKey(s string) error {
  f := string.Fields(s) <-- translate from data domain to user domain
  c.Surname = f[0]
  c.Name= f[1]
  return nil
}
```
It's clear how the ```SetRowKey``` method could fail (perhaps only one field was provided?). In the code there is no check but you could (should, I was lazy :innocent:) prevent the panic and return a nicer error instead.

I agree most implementations won't benefit from it though. At any rate in case is not needed it's just a ```return nil``` more in the code. What do you think?

storage/table_entities.go
MindFlavor: Well that's embarrassing, I don't really remember why I did it :blush:. And there should be no performance difference either as slices are pointers anyway. I'll try and remove it.

storage/table_entities.go
MindFlavor: The documentation ([https://msdn.microsoft.com/en-us/library/azure/dd179421.aspx](https://msdn.microsoft.com/en-us/library/azure/dd179421.aspx)) is not clear on that context so I took the safe path and returned the ContinuationToken if available. If the REST API call does not return it  the ```contToken``` will by simply ```nil``` anyway. 

I've hardened ```extractContinuationTokenFromHeaders``` in the meantime. Now it checks for the presence of both HTTP headers before trying to construct a ```ContinuationToken```:

```go
func extractContinuationTokenFromHeaders(headers map[string][]string) *ContinuationToken {
	if kh, ok := headers[continuationTokenPartitionKeyHeader]; ok { // continuationTokenPartitionKeyHeader must be present
		if rh, ok := headers[continuationTokenRowHeader]; ok { 		// continuationTokenRowHeader must be present too
			if (len(kh) > 0) && (len(rh) > 0) {						// they both must have at least one entry
				return &ContinuationToken{kh[0], rh[0]}
			}
		}
	}
	return nil														// otherwise return nil
}
```

storage/table_entities.go
MindFlavor: Nice! And it's much more readable too!
Changed! :+1:

storage/table_entities.go
MindFlavor: You are absolutely right, there is a lot of duplication in the code. I followed you advice and merged every insert/merge/insertormerge/insertorupdate RESP API call in one ```execTable``` method: 

```go
func (c *TableServiceClient) execTable(table AzureTable, entity TableEntity, operation tableOperation) error {
	uri := c.client.getEndpoint(tableServiceName, pathForTable(table), url.Values{})
	if operation != tableOperationTypeInsert {
		uri += fmt.Sprintf("(PartitionKey='%s',RowKey='%s')", url.QueryEscape(entity.PartitionKey()), url.QueryEscape(entity.RowKey()))
	}

	headers := c.getStandardHeaders()

	var buf bytes.Buffer

	if err := injectPartitionAndRowKeys(entity, &buf); err != nil {
		return err
	}

	headers["Content-Length"] = fmt.Sprintf("%d", buf.Len())

	var err error
	var resp *odataResponse
	var expectedStatusCode int

	switch operation {
	case tableOperationTypeInsert:
		resp, err = c.client.execTable("POST", uri, headers, &buf)
		expectedStatusCode = http.StatusCreated
	case tableOperationTypeUpdate:
		resp, err = c.client.execTable("PUT", uri, headers, &buf)
		expectedStatusCode = http.StatusNoContent
	case tableOperationTypeMerge:
		resp, err = c.client.execTable("MERGE", uri, headers, &buf)
		expectedStatusCode = http.StatusNoContent
	case tableOperationTypeInsertOrReplace:
		resp, err = c.client.execTable("PUT", uri, headers, &buf)
		expectedStatusCode = http.StatusNoContent
	case tableOperationTypeInsertOrMerge:
		resp, err = c.client.execTable("MERGE", uri, headers, &buf)
		expectedStatusCode = http.StatusNoContent
	}

	if resp != nil {
		if resp.body != nil {
			defer resp.body.Close()
		}
	}

	if err != nil {
		return err
	}

	if err := checkRespCode(resp.statusCode, []int{expectedStatusCode}); err != nil {
		return err
	}

	return nil
}
```

It's much cleaner now :)

storage/table_entities.go
MindFlavor: Nice! :+1:

storage/table_entities.go
ahmetalpbalkan: @MindFlavor I think you can move `defer resp.body.Close()` outside nil checks by placing it after `if err != nil`. If err==nil, then resp and reps.Body will be non-nil. this is a common practice in Go http package.

ahmetalpbalkan: @MindFlavor that said I still prefer having `expectedStatusCode` inside each method's body. This  is simply because, the verb doesn't indicate the expected status code, the endpoint does. And the endpoint is defined inside methods (like CreateTable, MergeEntity etc.) and the expected status code is defined within documentation of those methods. So the `expectedStatusCode` probably should reside in each endpoint method.

storage/table_entities.go
ahmetalpbalkan: @MindFlavor hmm that's a lot of unnecessary checks there, first of all `extractContinuationTokenFromHeaders` should take `http.Header` not `map[string][]string`. Because `Header` has a `Get` method https://golang.org/pkg/net/http/#Header.Get that directly returns you a string, so your method would be:

```go
func extractContinuationTokenFromHeaders(h http.Header) {
    if h.Get(key1) != "" && h.Get(key2) != "" { return &CT{h.Get(key1), h.Get(key2) }
    return nil
}
```

it has been a lot clearer, ha? :wink:

storage/table_entities.go
ahmetalpbalkan: @MindFlavor thanks, much appreciated! :heart:

storage/table_entities.go
ahmetalpbalkan: @MindFlavor I just didn't understand this part: we call SetPartitionKey/SetRowKey only when deserializing from Azure Storage response right? There's no way it can be invalid. Users should not store user-domain data in PK/RK as far as I know. Even in the C# SDK, these fields are just `string`s and you're going to be given whatever server gives you, you can't error out in that case? :)

storage/client.go
```diff
@@ -356,6 +400,66 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 		body:       resp.Body}, nil
 }
 
+func (c Client) execInternalJSON(verb, url string, headers map[string]string, body io.Reader) (*odataResponse, error) {
+	req, err := http.NewRequest(verb, url, body)
+	for k, v := range headers {
+		req.Header.Add(k, v)
+	}
+
+	httpClient := http.Client{}
```
ahmetalpbalkan: hey @MindFlavor I just merged #298 which reads this from `Client.HTTPClient` (and defaults to `net/http#DefaultClient`). In this case you'll probably need to rebase your code on top of `master` and refactor this part sometime.

storage/table_entities.go
MindFlavor: Wow indeed it is! 
Here's my proposed implementation:

```go
func extractContinuationTokenFromHeaders(h http.Header) *ContinuationToken {
	ct := ContinuationToken{h.Get(continuationTokenPartitionKeyHeader), h.Get(continuationTokenRowHeader)}

	if ct.NextPartitionKey != "" && ct.NextRowKey != "" {
		return &ct
	}
	return nil
}
```

storage/table_entities.go
MindFlavor: Yes your logic is sound. I've change the code to follow your advice:

```go
func (c *TableServiceClient) execTable(table AzureTable, entity TableEntity, specifyKeysInUrl bool, method string) (int, error) {
	uri := c.client.getEndpoint(tableServiceName, pathForTable(table), url.Values{})
	if specifyKeysInUrl {
		uri += fmt.Sprintf("(PartitionKey='%s',RowKey='%s')", url.QueryEscape(entity.PartitionKey()), url.QueryEscape(entity.RowKey()))
	}

	headers := c.getStandardHeaders()

	var buf bytes.Buffer

	if err := injectPartitionAndRowKeys(entity, &buf); err != nil {
		return 0, err
	}

	headers["Content-Length"] = fmt.Sprintf("%d", buf.Len())

	var err error
	var resp *odataResponse

	resp, err = c.client.execTable(method, uri, headers, &buf)

	if err != nil {
		return 0, err
	}

	defer resp.body.Close()

	return resp.statusCode, nil
}
```

The insert/whaever methods became:

```go
func (c *TableServiceClient) InsertEntity(table AzureTable, entity TableEntity) error {
	if sc, err := c.execTable(table, entity, false, "POST"); err != nil {
		return checkRespCode(sc, []int{http.StatusCreated})
	} else {
		return err
	}
}

func (c *TableServiceClient) UpdateEntity(table AzureTable, entity TableEntity) error {
	if sc, err := c.execTable(table, entity, true, "PUT"); err != nil {
		return checkRespCode(sc, []int{http.StatusNoContent})
	} else {
		return err
	}
}
```

And so on :)

storage/client.go
MindFlavor: Ok renamed according to your suggestion! :+1:

storage/client.go
MindFlavor: Merged.

MindFlavor:
Sorry @ahmetalpbalkan I missed your comment
> @MindFlavor I just didn't understand this part: we call SetPartitionKey/SetRowKey only when deserializing from Azure Storage response right? There's no way it can be invalid. Users should not store user-domain data in PK/RK as far as I know. Even in the C# SDK, these fields are just strings and you're going to be given whatever server gives you, you can't error out in that case? :)

I'll fix it right away removing the error and push again :)

MindFlavor:
I've also noticed something weird in Travis' test suite. As you can see from the last log:

```
table_test.go:263:
    c.Assert(entries[0].(*CustomEntity), chk.DeepEquals, ceList[0])
... obtained *storage.CustomEntity = &storage.CustomEntity{Name:"Test", Surname:"Test2", SomeDate:time.Time{sec:63594497276, nsec:347889129, loc:(*time.Location)(0xb75f40)}, Number:0, PKey:"pkey", RKey:"r0"}
... expected *storage.CustomEntity = &storage.CustomEntity{Name:"Test", Surname:"Test2", SomeDate:time.Time{sec:63594497276, nsec:347889129, loc:(*time.Location)(0xb7a2a0)}, Number:0, PKey:"pkey", RKey:"r0"}
```

The DeepEquals does not work as expected, it does not follow the ```loc:(*time.Location)``` pointer and compares the addresses instead (```loc:(*time.Location)(0xb75f40)``` in one case and ```loc:(*time.Location)(0xb7a2a0)``` in the other). It should compare the Deep internal values instead :).
This is clearly a bug (and on my pc works correctly but... that's always the case of course :blush: )

To work around this I've removed the indirect time field in the tests for the time being. I think we should reinstate it sooner or later though (and that's why I had a separated ```testEquality``` method I think).

storage/client.go
ahmetalpbalkan: it still looks like this part is not addressed. can you just `git fetch origin && git rebase origin/master` and `git push -f`

MindFlavor: Well of course, I was merging from the wrong repo :blush: ... It should be ok now...

ahmetalpbalkan: @MindFlavor  you should still update this part to use c.Client.

MindFlavor: Once again, sorry :-). Let's hope I got it right this time :blush:

storage/client.go
```diff
@@ -68,6 +74,20 @@ type AzureStorageServiceError struct {
 	RequestID                 string
 }
 
+type odataErrorMessageMessage struct {
+	Lang  string `json:"lang"`
+	Value string `json:"value"`
+}
+
+type odataErrorMessageInternal struct {
```
ahmetalpbalkan: @MindFlavor  do you ever use the error message/code in generating a human-readable error? perhaps we should return a typed-error that people can cast into and read these fields?

storage/client.go
```diff
@@ -364,6 +408,70 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
...
+		}
+		// try unmarshal as odata.error json
+		err = json.Unmarshal(respBody, &respToRet.odata)
+		return respToRet, err
+	}
+
+	return respToRet, nil
+}
+
+func (c Client) createSharedKeyLite(url string, headers map[string]string) (string, error) {
```
ahmetalpbalkan: I think I'm going to switch everything to SharedKeyLite after merging. :smile:

cyonix:
Would be nice to see this pull request get merged in soon guys!!

ahmetalpbalkan:
Sorry for the delay here. We are okay with merging this, thanks a lot to @MindFlavor staying on top of this.
@MindFlavor there are still a few comments above, if you feel like addressing them I appreciate follow up requests. I will now squash this and merge.

Issue #208, opened 2015-09-10
storage - implement PutBlockBlob API


kenegozi:
When uploading a smaller blob, it is much simpler in many cases to use the PutBlob API call with body (as described here https://msdn.microsoft.com/en-us/library/azure/dd179451.aspx)  and avoid the need for multiple requests uploading blocks and committing them.

Per the official documentation the upper limit is 64 MB ... A common scenario with much smaller sizes is uploading images and thumbnails to blob storage later to be served to browsers/clients.

ahmetalpbalkan:
@kenegozi this is actually very similar to #129. What a user ideally wants is an API just uploads the blob whatever the size is.

I don't think we should provide two APIs such as `PutSingleBlockBlob` (the one that you described that accepts up to 64 MB) and a `PutBlockBlob` (described over at #129). Ideally there should be only one.

If you'd like to take a stab at it, there's a planned design at #129.

kenegozi:
I appear as if the philosophy of the SDK is to avoid surprised, and map rest API calls to SDK methods. a PutBlockBlob should just call the PutBlob REST API with BlockBlob type and a body.

A helper that potentially magically break things down to multiple calls is super cool but belong in a separate, higher level package (blobutil is a good suggestion, similar to std package's 'io' and 'ioutil').

ahmetalpbalkan:
@kenegozi Agree with your “no surprises” argument. IMO the problem with accepting a 1 GB blob and then transmitting that over the wire and seeing the REST call to fail might be a bit frustrating. On the other hand adding the 64 MB check to the client SDK is also one of the things we want to avoid (duplicating the validation).

But let's go with your idea, storage.PutBlockBlob and blobutil.PutBlockBlob and make it clear in the docs and refer from one to another in the comments?

kpfaulkner:
Is this still desired? Have blobutil.PutBlockBlob that does the chunking and putList for you?

kenegozi:
It's been a while, and I've been away from doing golang/azure work recently, so if you have a good way to provide the usage to users that you are happy with then this can be closed as "no thanks"

Issue #129, opened 2015-05-15
storage/blobutil package to make it easy to upload blobs


ahmetalpbalkan:
We recently removed helper methods from `storage/blob` pkg so that users won't be fooled by methods containing special client-side logic.

Now we must have a good client library for handling blob uploads (typically block blobs) from user's point of view easily.

Main areas for consideration for `PutBlob()` functionality (We must write logic to handle these cases):
1. Block ID generation strategy for multi-block blobs.
2. Uploading blocks (chunks) concurrently in `runtime.NumCPU` goroutines (io efficiency)
3. Support overwrites (to the entire blob)
4. If blob is small enough, we may as well rely on single PutBlob call rather than bunch of PutBlocks; then PutBlockList
5. Handle large blob uploads (limit data stored in-memory, accept `io.ReadCloser`)

cc: @paulmey wdyt?

ahmetalpbalkan:
@paulmey ping again. I'll start after your input on this.

paulmey:
Sounds good indeed, go for it!

boonep:
Wanted to check on PutBlob() status, let me know if I can do anything to help.  Thanks!

ahmetalpbalkan:
@boonep sorry I didn't have time to get back to this. There's pretty much a spec of what I wanted to build above and your help is greatly appreciated and welcomed. If you like to put out a `blobutil` package prototype, I'd be happy to proactively review and help iterate.

jerbob92:
It would be great to have this in, it would make file uploads much easier.

rdhakne:
I see that in C++ Azure SDK there are functions to 'upload from file' (void upload_from_file(const utility::string_t &path)). I dont see the same in golang SDK. 
Am I missing something or they will be coming soon? :)

ahmetalpbalkan:
To anyone interested having this blobutil package: You can now use `AppendBlob` type which is super easy to just keep uploading chunks of data to a blob. It doesn't have the complexity of `BlockBlob`s or anything like that.

mcardosos:
@jhendrixMSFT

marstr:
This is something that should be considered as part of @jhendrixMSFT's refactor as seen in #470.

In an effort to consolidate issues, I'm going to close this one. If folks feel there is further independent conversation to have here, or if #470 doesn't encapsulate this, feel free to re-open.

PR #218, opened 2015-09-25
Add CreateBlockBlobWithData and CreateBlockBlobFromReader

tomclegg:


ahmetalpbalkan:
@tomclegg thanks for the pull request. We used to have a version of this method that was accepting data as you submitted and we ended up deleting that.

This sort of gives the impression that any size of data in `[]byte` or `io.Reader` would work with these calls; whereas, that's not true. This API call has a 16 or 64 MB limit (or something like that) and bigger files should be chunked into separate requests that look like: 

    CreateBlob --> PutBlock(data) --> PutBlock(data) -->  ... --> PutBlockList(blocks)

however the initial call to CreateBlob also accepts data, but this creates a situation where user is not in control of the block ID this data is uploaded to, therefore manipulating the blob later on using PutBlock/PutBlockList calls become harder. That's why I try to avoid sending data in this CreateBlob call...

We have a plan to implement #208  and that would happily accommodate `[]byte`/`io.Reader`. Does it make sense to you as well to implement this contribution for that `PutBlockBlob` in #208 that would work with any size of contents, instead of this initial call?

tomclegg:
My use case involves creating a 64 MiB block atomically, and I don't care what the block IDs are because my blocks are immutable, so PutBlob seems like the right API. PutBlock, on the other hand, means uploading 16x 4MiB blocks: forking the SDK to use the more appropriate API seemed easier and neater.

If I'm following #129 and #208 correctly, the plan is to add "upload small blobs in one go, split bigger blobs into blocks and upload with optional concurrency." That sounds useful too.

It seems to me that even if we have `blobutil.PutBlockBlob` it will still make sense to have a `PutBlob` that just calls the corresponding REST API, which is what I was aiming for here. So I think this solves a different problem than #208 (but might make #208 easier to write).

Re size restriction: PutBlock seems to ignore the size and let the server decide whether it's 413, so I figured PutBlob should use the same approach. Ideally the server would tell us the maximum size so we don't end up imposing different limits in the SDK and the actual API. (Does it?) I'd be happy to add a size check if you like.

ahmetalpbalkan:
@tomclegg giving a second thought your argument makes sense. This is a valid API call that users should be allowed to use. `blobutil.PutBlockBlob` probably will end up using this call with no data to have full control of block IDs (they all have to be the same length).

We shouldn't check any sizes on the client side as you said.

ahmetalpbalkan:
@tomclegg however, I think it'd be useful to document the max sizes on these methods as godoc comments. Do you mind doing that? Also IMHO `..WithData` might be a bit obsolete because `[]byte` can be easily made a reader by `bytes.NewReader(data)`.

tomclegg:
Should I rename CreateBlockBlobWithData to PutBlockBlob or PutBlob? (I named it to match the existing SDK function, but following the REST API name would also seems reasonable. FWIW, when I first tried to find it in the SDK, I looked for PutBlob.)

tomclegg:
@ahmetalpbalkan yes, ...WithData is kind of sugary, I'll remove it.

ahmetalpbalkan:
CreateBlockBlobWithReader is better IMO. PutBlockBlob/PutBlob is the REST API docs terminology but I don't necessarily agree it makes it clear that this is a size-limited endpoint that should be used to create first chunk of the blob.

storage/blob_test.go
```diff
@@ -389,6 +390,41 @@ func (s *StorageBlobSuite) TestGetBlobRange(c *chk.C) {
...
+
+func (s *StorageBlobSuite) TestCreateBlockBlobFromReaderWithShortData(c *chk.C) {
+	cli := getBlobClient(c)
+	cnt := randContainer()
+	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
+	defer cli.deleteContainer(cnt)
+
+	name := randString(20)
+	data := randBytes(8888)
+	err := cli.CreateBlockBlobFromReader(cnt, name, 9999, bytes.NewReader(data))
```
ahmetalpbalkan: this test is now pretty much the same thing as the other test, right? Also you're saying I'm sending 9999 bytes but `data` is 8888 bytes.

tomclegg: Yes, up to this line the test is just like the previous one. But the previous one only tests a "success" case. This one tests whether an upload error ("we said we were uploading a 9999-byte blob, but we provided fewer than 9999 bytes of data") causes `CreateBlock` to return an error (so our application knows something went wrong) and avoids storing the truncated data in the container (so we don't have to wonder every time our application calls `GetBlob` whether we're getting back bogus data due to a previous upload mishap).

Perhaps `...WithTruncatedData` would be a better hint than `...WithShortData` about what's being tested?

ahmetalpbalkan: ah sorry didn't see the `Not(..)` part.

ahmetalpbalkan:
LGTM. Thanks @tomclegg.

PR #219, opened 2015-09-25
Add SetBlobMetadata and GetBlobMetadata

tomclegg:


ahmetalpbalkan:
@tomclegg thanks for the pull request, reviewing now.

storage/blob.go
```diff
@@ -541,6 +541,77 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+		return nil, err
+	}
+	defer resp.body.Close()
+
+	if err := checkRespCode(resp.statusCode, []int{http.StatusOK}); err != nil {
+		return nil, err
+	}
+
+	metadata := map[string]string{}
+	metaPrefix := "x-ms-meta-"
```
ahmetalpbalkan: can you please extract this into a const? (also the usage above)

we have a const in this pkg `userDefinedMetadataHeaderPrefix` that contains this value (however it's Camel-Cased).

storage/blob.go
```diff
@@ -541,6 +541,77 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+		// Can't trust CanonicalHeaderKey() to munge case
+		// reliably. "_" is allowed in identifiers:
+		// https://msdn.microsoft.com/en-us/library/azure/dd179414.aspx
+		// https://msdn.microsoft.com/library/aa664670(VS.71).aspx
+		// http://tools.ietf.org/html/rfc7230#section-3.2
+		// ...but "_" is considered invalid by
+		// CanonicalMIMEHeaderKey in
+		// https://golang.org/src/net/textproto/reader.go?s=14615:14659#L542
+		// so k can be "X-Ms-Meta-Foo" or "x-ms-meta-foo_bar".
+		if strings.ToLower(k[:len(metaPrefix)]) != metaPrefix {
```
ahmetalpbalkan: you can use strings.TrimLeft instead of [:] operator.

storage/blob.go
```diff
@@ -541,6 +541,77 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+			continue
+		}
+		// metadata["foo"] = content of the last X-Ms-Meta-Foo header
+		metadata[strings.ToLower(k[len(metaPrefix):])] = v[len(v)-1]
```
ahmetalpbalkan: strings.TrimLeft would be easier to read.

storage/blob.go
tomclegg: `TrimLeft(s string, cutset string)` ? Don't quite follow... If you mean `HasPrefix`, this is what the above comment is about: we can't use `HasPrefix` because we don't know whether `k` is a `"x-ms-meta-foo_bar"` or a `"X-Ms-Meta-Foobar"`.

storage/blob.go
```diff
@@ -541,6 +545,77 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.body.Close()
+
+	if err := checkRespCode(resp.statusCode, []int{http.StatusOK}); err != nil {
+		return nil, err
+	}
+
+	metadata := map[string]string{}
```
ahmetalpbalkan: can you `make(map...)` this one?

storage/blob_test.go
```diff
@@ -343,6 +343,49 @@ func (s *StorageBlobSuite) TestListBlobsPagination(c *chk.C) {
...
+		"Foo":     "different bar",
+		"bar_BAZ": "different waz qux",
+	}
+
+	err = cli.SetBlobMetadata(cnt, blob, mPut2)
+	c.Assert(err, chk.IsNil)
+
+	m, err = cli.GetBlobMetadata(cnt, blob)
+	c.Assert(err, chk.IsNil)
+	c.Check(m["foo"], chk.Equals, mPut2["Foo"])
```
ahmetalpbalkan: another `expected := map...` then `chk.DeepEquals` check could've been better perhaps?

storage/blob.go
```diff
@@ -541,6 +545,77 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+		// Can't trust CanonicalHeaderKey() to munge case
+		// reliably. "_" is allowed in identifiers:
+		// https://msdn.microsoft.com/en-us/library/azure/dd179414.aspx
+		// https://msdn.microsoft.com/library/aa664670(VS.71).aspx
+		// http://tools.ietf.org/html/rfc7230#section-3.2
+		// ...but "_" is considered invalid by
+		// CanonicalMIMEHeaderKey in
+		// https://golang.org/src/net/textproto/reader.go?s=14615:14659#L542
+		// so k can be "X-Ms-Meta-Foo" or "x-ms-meta-foo_bar".
+		if strings.ToLower(k[:prefixLen]) != userDefinedMetadataHeaderPrefixLcase {
```
ahmetalpbalkan: so my comment was, you can do:

    strings.ToLower(k).HasPrefix(strings.ToLower(userDefinedMetadataHeaderPrefix))

here and get rid of subscript notation and `userDefinedMetadataHeaderPrefixLcase`

ahmetalpbalkan:
generally LGTM, will be happy to merge after CLA signed/approved.

storage/blob.go
tomclegg: Ah. I was mainly avoiding extra string allocs, but yes, this way is clearer.

ahmetalpbalkan:
LGTM. Thanks @tomclegg.

Issue #196, opened 2015-08-29
Enable gofmt/govet/golint for ./arm/...


ahmetalpbalkan:
@brendandixon please feel free to edit .travis.yml accordingly when you feel comfortable enabling those checks for the `arm` package.

brendandixon:
@ahmetalpbalkan It's on the to-do list.

ahmetalpbalkan:
apparently done. :)

@brendandixon GitHub protip: If you add your commit message something like "Fixes #xxx" or "Closes #xxx", when commit is merged, issue(s) will close automatically once the commit is merged. https://help.github.com/articles/closing-issues-via-commit-messages/

brendandixon:
@ahmetalpbalkan Sweet. I did not know that.

PR #222, opened 2015-10-02
Added golint / govet to Travis CI for arm packages

brendandixon:
This change:

- Improves the UserAgent string to disambiguate arm packages from others in the SDK
- Improves setting the http.Response into generated results (reduces likelihood of a nil reference)
- Adds gofmt, golint, and govet to Travis CI for the arm packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,12 @@ before_script:
 
 go: tip
 script:
+  - test -z "$(gofmt -s -l $(find ./arm/* -type d -print) | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w management | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w storage | tee /dev/stderr)"
   - go build -v ./...
+  - test -z "$(go vet $(find ./arm/* -type d -print) | tee /dev/stderr)"
+  - test -z "$(golint ./arm/... | tee /dev/stderr)"
   - go test -v ./storage/... -check.v
   - test -z "$(golint ./storage/... | tee /dev/stderr)"
   - go vet ./storage/...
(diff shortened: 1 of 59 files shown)


brendandixon:
This PR address #196.

arm/subscriptions/version.go
```diff
@@ -25,11 +25,11 @@ import (
 const (
 	major = "0"
 	minor = "1"
-	patch = "0"
+	patch = "1"
```
ahmetalpbalkan: this means you need to update the changelog?

.travis.yml
```diff
@@ -8,9 +8,12 @@ before_script:
 
 go: tip
 script:
+  - test -z "$(gofmt -s -l $(find ./arm/* -type d -print) | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w management | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w storage | tee /dev/stderr)"
   - go build -v ./...
+  - test -z "$(go vet $(find ./arm/* -type d -print) | tee /dev/stderr)"
```
ahmetalpbalkan: Is there a shorter form of `go vet` that doesn't require you to do `$(find ...)`? Maybe like  `go vet -v ./arm` or `go vet -v ./arm/...` ?

brendandixon: @ahmetalpbalkan Not that I could find. I tried various incantations before settling on this. I felt it poor form to list each package individually since it would mean having to update the file any time we added a new package.

arm/subscriptions/version.go
brendandixon: Grrr. Correct.

ahmetalpbalkan: if you haven't tagged you probably shouldn't bump the version up. just undo the change in `patch`.

ahmetalpbalkan:
LGTM

PR #225, opened 2015-10-12
Switch to x/crypto/pkcs12 instead of Azure repo

paulmey:


ahmetalpbalkan:
Can you vendor with godep?

paulmey:
yup, one sec

paulmey:
done

ahmetalpbalkan:
Doesn't seem right. Did you use godep tool to vendor the pkg or just edited the json file? Source files are missing.

paulmey:
I used the tool, but forgot to `git add` the godeps workspace

phinze:
:+1: Just confirmed that this PR fixes the Terraform master build, which was broken when https://github.com/Azure/go-pkcs12/pull/26 landed.

management/publishSettings.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	"github.com/Azure/go-pkcs12"
+	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/golang.org/x/crypto/pkcs12"
```
ahmetalpbalkan: This line is wrong. The import path should be hust golang.org/...

ahmetalpbalkan:
LGTM. Feel free to merge after import path fix.

management/publishSettings.go
paulmey: See https://github.com/Azure/azure-sdk-for-go/blob/master/arm/compute/client.go#L22 for example. This is what godeps does to your import paths. This is not wrong, IMHO.

ahmetalpbalkan: that's what happens when you do `godep save -r` which rewrites your paths IMHO. I don't think we need path rewriting here but let's leave it as is for consistency until somebody complains. :)

ahmetalpbalkan: also keep in mind that if some project imports this package + `golang.org/x/crypto/pkcs12` separately, then it'll end up two copies of it, possibly even duplicate types with the same name in the runtime.

PR #229, opened 2015-10-26
Use the Content-Length header if it is set

paulmey:
The Content-Length header is never used by the Go runtime.
Instead, the ContentLength field on the request struct is used to set the header
when writing the actual request. This field is only set for [certain types of io.Readers](https://golang.org/src/net/http/request.go?s=18140:18370#L536 ).
Azure storage signatures include the Content-Length header, hence, it needs to be
respected, even if the io.Reader is some other type that is not recognized by the Go
runtime.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -12,6 +13,7 @@ import (
 	"net/url"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


storage/client.go
```diff
@@ -295,6 +296,19 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 	}
 
 	req, err := http.NewRequest(verb, url, body)
+	if err != nil {
+		return nil, err
+	}
```
paulmey: Also added this error check that wasn't there at all

ahmetalpbalkan: Can you add a bit detail before the err? It might look awkward upper in the stack and it light be hard to tell where this is coming from.

paulmey: As in `errors.New("storage.Client: error creating request object: " + err.Error())`?

ahmetalpbalkan: yeah except: `azure/storage:` would be better.

paulmey: done

PR #233, opened 2015-11-10
godep: update go-autorest to 1.1.0

colemickens:


Godeps/_workspace/src/github.com/Azure/go-autorest/autorest/azure/token.go
```diff
@@ -1,12 +1,19 @@
...
+	"encoding/base64"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/golang.org/x/crypto/pkcs12"
+	"github.com/dgrijalva/jwt-go"
```
ahmetalpbalkan: looks like this line is not properly vendored (by rewriting the import path with `-r`). @brendandixon ?

colemickens: That was going to be my remark.

We don't seem to be vendoring sub-dependencies.

This is already the case with another dependency: `gopkg.in/check.v1`

I can introduce another commit to vendor both `jwt-go` and `gopkg.in/check.v1` and rewrite paths... if that's what we want to do.

brendandixon: We need to re-vendor each time. That should do the trick. At least, that's how I understood it...

colemickens: I've added a new commit; it vendors sub-dependencies (`pkcs12`, `jwt-go`, and `check.v1`), includes the sub-dependencies' test code (keeping in line with how autorest was originally vendored), and rewrites paths.

Hopefully this resolves any concerns.

LMK and I'll squash it.

ahmetalpbalkan: go-check is a test package that's why that was the case as far as I
remember. When someone vendors our SDK I don't see why they should get go
check. They're not going to test the package.

also godep leaves that pkg out, again because it's referenced only in test
builds; not the normal builds.
On Tue, Nov 10, 2015 at 11:07 AM Cole Mickens <notifications@github.com>
wrote:

> In
> Godeps/_workspace/src/github.com/Azure/go-autorest/autorest/azure/token.go
> <https://github.com/Azure/azure-sdk-for-go/pull/233#discussion_r44449656>:
>
> >  	"net/http"
> >  	"net/url"
> >  	"strconv"
> >  	"time"
> >
> >  	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/github.com/Azure/go-autorest/autorest"
> > +	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/golang.org/x/crypto/pkcs12"
> > +	"github.com/dgrijalva/jwt-go"
>
> I've added a new commit; it vendors sub-dependencies (pkcs12, jwt-go, and
> check.v1), includes the sub-dependencies' test code (keeping in line with
> how autorest was originally vendored), and rewrites paths.
>
> Hopefully this resolves any concerns.
>
> LMK and I'll squash it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Azure/azure-sdk-for-go/pull/233/files#r44449656>.
>
-- 
—Ahmet Alp Balkan

colemickens: I believe that it was not included before because Godep was previously only vendoring two directories. You can see that change in Godeps.json.

I believe this, because I know that `go-autorest` test code was vendored, hence I know that the `-t` flag was used.

I'm fine with not vendoring the test code... but please note that we were already vendoring the `autorest` test code, so I would be undoing that as well, since it's a global knob, unless I'm misunderstanding a finer point of using `godep`.

colemickens: Actually, I take that back.

The test dependencies are included regardless of specifying `-t` to Godep.

The only reason it wasn't vendored before is because we were only using `arm/` and `management/` as sources of things to vendor. Now that we vendor everything in `./...`, it looks at additional files that include gocheck.

I think this is actually the right thing to do. It's a dependency of our repo. If they vendor our SDK, they should still be able to edit the code and run the tests; that's part of the benefit of vendoring. 

Further, the only way to undo adding gocheck is to go back to excluding parts of the project from Godep's control... which I really think is the wrong thing to do.

brendandixon:
@ahmetalpbalkan @colemickens FWIW I agree: Unless there is a clear standard to not vendor (doubtful at this stage in Go's life), let's vendor it all.

colemickens:
Note: This should be held until the PR adding Godep to go-autorest (https://github.com/Azure/go-autorest/pull/7) is pulled in. Then I'll vendor that version and we can proceed here.

ahmetalpbalkan:
> The test dependencies are included regardless of specifying -t to Godep.

@colemickens this sounds interesting. I never observed this and this was why this package was not vendored until now. A few months ago godep was changed to vendor source of the packages that are referenced during the `go build -v` –normally that shouldn't contain test-only packages

colemickens:
You can check the behavior yourself. Clone the repo into the proper place in a GOPATH. Modify `godeps` to include source paths that actually include the test (that's why they weren't ever included before).

As soon as you do `go build ./...`, then `godep save ./...` you'll see it complain about `check.v1` missing. The resolution is to `go get` the package and then `go dep save

brendandixon:
@ahmetalpbalkan @colemickens By the way, should we consider moving to glide (https://github.com/Masterminds/glide). It is the favored option for managing dependencies with Go 1.5.x.

colemickens:
It looks like it is expected that dependencies for your own tests are vendored: https://github.com/tools/godep/pull/276#issuecomment-142124382

So, we should expect to see gocheck being pulled into `azure-sdk-for-go` but when someone else vendors us... they won't have to pull in `check.v1`.

ahmetalpbalkan:
@brendandixon first time I'm hearing about it, so I'd probably need more data points to call it the favored option. FWIW Godep has go15vendorexperiment support as well.

brendandixon:
@ahmetalpbalkan I somewhat misspoke. I had thought I encountered the comment preferring Glide on the Go blog. It was not. It was here: http://engineeredweb.com/blog/2015/go-1.5-vendor-handling/. Still, I find Glide quite interesting. I agree we should research it and consider.

PR #234, opened 2015-11-11
Adding AddRole method so that this bug can be fixed: #3568

kbxkb:
This is a pretty serious bug and limits what we can achieve with production grade deployments on Azure: https://github.com/hashicorp/terraform/issues/3568

As a first step to fix this bug, I propose that we add an "AddRole" method to azure-sdk-for-go for the virtualmachine client. This is a missing API now. Azure already has a REST API for this: https://msdn.microsoft.com/en-us/library/azure/jj157186.aspx

However, azure-sdk-for-go does not have a corresponding method that calls this REST API. Hence, we cannot add a new VM to an existing deployment or cloud service today.

Please accept this straightforward change. It is just the addition of a base method that is missing. Once accepted and merged, I plan to use it from terraform and fix the issue. Thanks again!

I have tested this method thoroughly.

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -62,6 +62,7 @@ func (vm VirtualMachineClient) CreateDeployment(
 	return vm.client.SendAzurePostRequest(requestURL, data)
 }
 
+//https://msdn.microsoft.com/en-us/library/azure/ee460804.aspx
 func (vm VirtualMachineClient) GetDeployment(cloudServiceName, deploymentName string) (DeploymentResponse, error) {
 	var deployment DeploymentResponse
 	if cloudServiceName == "" {
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -11,6 +11,7 @@ import (
 const (
 	azureDeploymentListURL   = "services/hostedservices/%s/deployments"
 	azureDeploymentURL       = "services/hostedservices/%s/deployments/%s"
+	azureListDeploymentsInSlotURL = "services/hostedservices/%s/deploymentslots/Production"
 	deleteAzureDeploymentURL = "services/hostedservices/%s/deployments/%s?comp=media"
 	azureAddRoleURL          = "services/hostedservices/%s/deployments/%s/roles"
 	azureRoleURL             = "services/hostedservices/%s/deployments/%s/roles/%s"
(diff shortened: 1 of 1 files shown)


management/virtualmachine/client.go
```diff
@@ -118,6 +119,25 @@ func (vm VirtualMachineClient) GetRole(cloudServiceName, deploymentName, roleNam
 	return role, nil
 }
 
+// AddRole adds a Virtual Machine to a deployment of Virtual Machines, where role name = VM name
+// See https://msdn.microsoft.com/en-us/library/azure/jj157186.aspx
+func (vm VirtualMachineClient) AddRole(cloudServiceName string, deploymentName string, role Role) (management.OperationID, error) {
+	if cloudServiceName == "" {
```
ahmetalpbalkan: I feel like this method should get PersistentVMRole as input, not just role. We try to keep inputs of methods and underlying requests matching 1:1 as much as we can. But I might be wrong I'm reading docs on mobile.

ahmetalpbalkan: Just looked a look and this seems ok. @paulmey PTAL,  LGTM.

kbxkb: Thanks @ahmetalpbalkan - This method is almost a mirror copy of the existing "UpdateRole" - which takes a Role parameter, then serializes a PersistentVMRole instance made out of that Role parameter. IMHO, we should stick to the same data types for consistency. Please let me know what you feel.

kbxkb: The Travis CI build is failing due to an unrelated error... "The command "go get -u github.com/golang/lint/golint" failed and exited with 1 during ." Any idea how can I kick off the build again? Thanks!

kbxkb:
Hi guys,
The Travis CI build is failing due to an unrelated error... "The command "go get -u github.com/golang/lint/golint" failed and exited with 1 during ." Any idea how can I kick off the build again? Thanks!

management/virtualmachine/client.go
ahmetalpbalkan: That's an ongoing issue with go tool / git combo in Travis started today. https://github.com/golang/go/issues/13206

We can ignore it if it compiles on your machine.

kbxkb: Thanks! It compiles and I have tested it as well. @paulmey can you please review and merge if okay

paulmey:
LGTM

Issue #232, opened 2015-11-09
arm/resources: DeploymentProperties object seems slightly off for template/parameters fields


colemickens:
Take, as an example:

https://msdn.microsoft.com/en-us/library/azure/dn790564.aspx

That page describes Template and Parameters as "complex types". The example shows the fields as being set to the actual JSON content of the template and deployment. 

Hence, where `azure-sdk-for-go` currently defines DeploymentProperties:

```
type DeploymentProperties struct {
	Template       *map[string]*string `json:"template,omitempty"`
	TemplateLink   *TemplateLink       `json:"templateLink,omitempty"`
	Parameters     *map[string]*string `json:"parameters,omitempty"`
	ParametersLink *ParametersLink     `json:"parametersLink,omitempty"`
	Mode           DeploymentMode      `json:"mode,omitempty"`
}
```

I would have expected:
```
	...
	Template       *map[string]interface{} `json:"template,omitempty"`
	...
	Parameters     *map[string]interface{} `json:"parameters,omitempty"`
	...
```

This also applies to DeploymentPropertiesExtended, and possibly others.

Alternatively, my expectations of the API are just wrong... and it's my responsibility to serialize the template/parameters first? But if that's the case, I'm not sure I understand why it's a map, rather than just a string.

brendandixon:
Agreed. I am in the midst of making changes to the Autorest generator for Go anyway. I'll add this. Thanks!

brendandixon:
BTW: This will be a general change. The challenge is that the Swagger defines such fields "objects" (in the JSON sense). The closest analogue in Go is a map[string]xxxx. I do not know why I chose to map to "map[string]string". It should be map[string]interface{}. (The pointer to string within the map was to handle JSON serialization in Go which, due to default values, cannot distinguish between the default case and the null case.)

colemickens:
I can confirm the deployment functionality works as expected. Thanks!

Issue #231, opened 2015-11-03
arm/compute.VirtualMachinesClient.Get fails with 503 Service Unavailable


axw:
Mostly I'm using the List method, but I'd like to optimise a case where we're only interested in querying a single instance. To that end, I've attempted to use the Get method, only to find it doesn't seem to work.

The Get method has a mysterious "expand" parameter, which AFAICT, does not correspond to anything documented against the compute REST API. The only reference I could find for $expand (which is how it's serialised) is to the AD Graph API. Anyway, I've set it to "". If I comment out https://github.com/Azure/azure-sdk-for-go/blob/master/arm/compute/virtualmachines.go#L403, Get works just fine, so $expand does appear to be at fault.

```
autorest:DoErrorUnlessStatusCode GET https://management.azure.com/subscriptions/6f0e3b87-f601-40fd-9727-76d05563fedd/resourceGroups/juju-azure-next-environment-14e690dc-d5c9-4958-8d6a-06004681e7eb/providers/Microsoft.Compute/virtualMachines/machine-0?%24expand=&api-version=2015-06-15 failed with 503 Service Unavailable
```

axw:
I haven't tested with VirtualMachineExtensionsClient.Get, but it also passes $expand.

brendandixon:
@axw $expand is part of the OData support used across the APIs (see http://www.odata.org). However, those values are optional. The SDK has a bug in that it is failing to distinguish between required and optional parameters. I will try to get a build out that addresses this soon.

axw:
@brendandixon Thanks, and sorry, I now see there's a doc-comment in the Go code about it accepting a value of "instanceView". It's a shame I can't use this when listing all VMs, as I'd like to be able to get the power state without having to query VMs individually.

brendandixon:
@axw One of the features missing from the Go SDK is OData support. Many of the APIs accept those parameters. But, sadly, it does not appear that ARM Listing of VMs is one of those. That's odd in my mind. Let me look into this more.

axw:
@brendandixon Thanks. I've pulled the latest azure-sdk-for-go, and the models seemed to have changed a fair bit. Some of the field types don't match up to enum types any more. For example, compute.HardwareProfile.VMSize is now a string, whereas previously it was VirtualMachineSizeTypes. Is this intentional?

brendandixon:
@axw Let me take a closer look.

[...]

Issue #230, opened 2015-11-03
arm/compute.VirtualMachinesClient.Deallocate errors on 200 OK


axw:
When using the VirtualMachinesClient.Deallocate method, a 200 OK response causes an error. The code is only allowing for 202 Accepted.

```
autorest:WithErrorUnlessStatusCode GET https://management.azure.com
/subscriptions/6f0e3b87-f601-40fd-9727-76d05563fedd/providers/Microsoft.Compute/locations/southeastasia/operations/9136bb98-77e2-495c-994b-8aff2b51547a?monitor=true&api-version=2015-06-15 failed with 200 OK
```

brendandixon:
@axw This is certainly a bug. I need to see if the Swagger is incorrect (likely), but even if it is, the code generator can do better (e.g., always include 200 in the suite of accepted return codes). Let me dig a bit and I'll try to get a new build out soon.

axw:
Thanks! This is fixed on master.

Issue #224, opened 2015-10-07
Subnet struct does not support the routeTable property


vgarg:
As per the documentation at [1] the the properties JSON for a subnet looks as follows:

```
{ 
   "properties":{ 
      "addressPrefix":"10.1.0.0/24",
      "networkSecurityGroup":{ 
         "id":"/subscriptions/{guid}/../microsoft.network/networkSecurityGroups/myNSG1"
      },
    "routeTable": { "id": "/subscriptions/{guid}/../Microsoft.Network/routeTables/myRT1" }
   }
}
```

We know this approach works as we have used the same approach within an ARM template.

Now we have a need to create another subnet at runtime and I am using the GO SDK to do so. However there does not appear to be a way to specify the routeTable for the subnet using the SDK.

On our end, we need the fix urgently. For now I guess I could just hack the SDK and add the missing property in the Subnet struct. Does that sound like something that would work?

[1] https://msdn.microsoft.com/en-us/library/azure/mt163621.aspx

brendandixon:
@vgarg I took a quick look at the Swagger that backs the SDK (see https://github.com/Azure/azure-rest-api-specs). I could find no references to "routeTable" in the arm-network Swagger. Could you, please, look it over as well? You are likely more familiar with the Network ARM templates than I am.

@devigned @vgarg's issue appears to be a Swagger problem.

vgarg:
@brendandixon Indeed, I see no mention of routeTable in the swagger JSON. Looks like the swagger JSON and the actual REST API documentation are out of sync. Or perhaps the routeTable property is new, and did not cause a version change?

brendandixon:
@vgarg Unclear. Let me follow-up with @devigned as he oversees the Swagger.

brendandixon:
@vgarg It looks like the ARM Network Swagger was recently updated. It now includes the routeTable. See https://github.com/Azure/azure-rest-api-specs/blob/master/arm-network/2015-05-01-preview/swagger/network.json#L6039. We're working an update to the ARM packages that will include these changes.

devigned:
@brendandixon we need to automate the codegen pipeline... We should talk about this soon.

brendandixon:
@devigned Sounds good to me. :)

ahmetalpbalkan:
Apparently fixed, closing.

PR #236, opened 2015-11-13
Address outstanding issues, upgrade to latest Swagger

brendandixon:
These changes address the issues #232, #231, #230, #224, #184, and #183. In addition, it upgrades the Swagger for various services (e.g., Network) and adds several new services (e.g., Redis, DNS).


arm/compute/models.go
```diff
@@ -195,6 +221,81 @@ const (
...
+	// StandardD12 specifies the standard d12 state for virtual machine size
+	// types.
+	StandardD12 VirtualMachineSizeTypes = "Standard_D12"
+	// StandardD12V2 specifies the standard d12v2 state for virtual machine
+	// size types.
+	StandardD12V2 VirtualMachineSizeTypes = "Standard_D12_v2"
+	// StandardD13 specifies the standard d13 state for virtual machine size
+	// types.
+	StandardD13 VirtualMachineSizeTypes = "Standard_D13"
+	// StandardD13V2 specifies the standard d13v2 state for virtual machine
```
ahmetalpbalkan: not sure why it is lowercase like d13v2 in the comments.

ahmetalpbalkan:
Looks good to me.

ahmetalpbalkan:
Actually I do not see a changelog either. Maybe contribute a release checklist so that we all know what must be done before merging, after merging etc.

I'm thinking of updating UA strings to `-dirty` after releasing new tags possibly, so that when a buggy request comes, we'd know if it's master branch or a tag.

arm/compute/models.go
brendandixon: It comes directly from the Swagger. Without very special checks, we cannot detect and alter such things. We could raise an issue with the Swagger.

brendandixon:
@ahmetalpbalkan I'll update the changelog. But I'm not clear on your second point. These changes will merge to master. I'll create a corresponding tag, but they'll be at head.

I don't believe there is value in altering the UA string. Doing so would likely not be kept up (e.g., removing "dirty" at the right time). As long as the UA has the version / tag, we're good.

paulmey:
LGTM with the changelog update

ahmetalpbalkan:
Looks like there's a CI failure due to Godep path rewriting we did for gocheck.

brendandixon:
@ahmetalpbalkan I know. I have not had time to correct it yet...

PR #237, opened 2015-11-18
DeleteRole REST API needs a query parameter

kbxkb:
DeleteRole REST API (used to destroy a VM) needs a query parameter to also delete the VHD disk from storage while deleting the VM. This parameter is "comp=media". Documentation is here: https://msdn.microsoft.com/en-us/library/azure/jj157184.aspx

Without this parameter, the DeleteRole() method would only delete the VM but not the disk, thereby leaking storage dollars. Added the parameter. Thanks.

Context: I am trying to fix https://github.com/hashicorp/terraform/issues/3568. In other words, I am adding the capability to manage multiple VM-s under the same Cloud Service to Terraform. Once this capability is there, we will have to also be able to destroy one VM at a time. As of now, terraform takes the shortcut of deleting the entire cloud service when deleting a VM. That works today as all terraform supports is one VM per cloud service. Hence DeleteRole() method is not called, DeleteDeployment() is called directly. But as part of the fix, I must start calling DeleteRole() to remove individual VM-s. Thus, DeleteRole() must work correctly. Hence this code change. Thanks

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -14,7 +14,8 @@ const (
 	azureListDeploymentsInSlotURL = "services/hostedservices/%s/deploymentslots/Production"
 	deleteAzureDeploymentURL = "services/hostedservices/%s/deployments/%s?comp=media"
 	azureAddRoleURL          = "services/hostedservices/%s/deployments/%s/roles"
-	azureRoleURL             = "services/hostedservices/%s/deployments/%s/roles/%s"
+	azureRoleURLGetUpdate    = "services/hostedservices/%s/deployments/%s/roles/%s"
+	azureRoleURLDelete       = "services/hostedservices/%s/deployments/%s/roles/%s?comp=media"
 	azureOperationsURL       = "services/hostedservices/%s/deployments/%s/roleinstances/%s/Operations"
 	azureRoleSizeListURL     = "rolesizes"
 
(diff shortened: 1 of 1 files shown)


management/virtualmachine/client.go
```diff
@@ -265,7 +266,7 @@ func (vm VirtualMachineClient) DeleteRole(cloudServiceName, deploymentName, role
 		return "", fmt.Errorf(errParamNotSpecified, "roleName")
 	}
 
-	requestURL := fmt.Sprintf(azureRoleURL, cloudServiceName, deploymentName, roleName)
+	requestURL := fmt.Sprintf(azureRoleURLDelete, cloudServiceName, deploymentName, roleName)
```
ahmetalpbalkan: @kbxkb Thanks for your pull request. At this line, you're always deleting the disks implicitly. I don't think this is very desirable. I prefer adding a `bool deleteDisks` parameter and appending `?comp=media` to the URL here.

kbxkb: Very good suggestion! I will add this. Thanks!

kbxkb: Closing pull request, will raise again after taking care of the comment above. Also will rebase so that auto-merge can work. Thank you @ahmetalpbalkan

ahmetalpbalkan:
Also, please consider not closing the PR next time, you can push -f to your fork. :-)

kbxkb:
Oh thank you! I did not know that. Will do that from next time.

PR #238, opened 2015-11-18
Add query parameter to DeleteRole to delete disk based on boolean

kbxkb:
DeleteRole REST API (used to destroy a VM) needs a query parameter to also delete the VHD disk from storage while deleting the VM. This parameter is "comp=media". Documentation is here: https://msdn.microsoft.com/en-us/library/azure/jj157184.aspx

Without this parameter, the DeleteRole() method would only delete the VM but not the disk, thereby leaking storage dollars. Added the parameter, but made it effective only if the boolean argument is true. Thanks.

Context: I am trying to fix hashicorp/terraform#3568. In other words, I am adding the capability to manage multiple VM-s under the same Cloud Service to Terraform. Once this capability is there, we will have to also be able to destroy one VM at a time. As of now, terraform takes the shortcut of deleting the entire cloud service when deleting a VM. That works today as all terraform supports is one VM per cloud service. Hence DeleteRole() method is not called, DeleteDeployment() is called directly. But as part of the fix, I must start calling DeleteRole() to remove individual VM-s. Thus, DeleteRole() must work correctly. Hence this code change. Thanks

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -14,8 +14,7 @@ const (
 	azureListDeploymentsInSlotURL = "services/hostedservices/%s/deploymentslots/Production"
 	deleteAzureDeploymentURL = "services/hostedservices/%s/deployments/%s?comp=media"
 	azureAddRoleURL          = "services/hostedservices/%s/deployments/%s/roles"
-	azureRoleURLGetUpdate    = "services/hostedservices/%s/deployments/%s/roles/%s"
-	azureRoleURLDelete       = "services/hostedservices/%s/deployments/%s/roles/%s?comp=media"
+	azureRoleURL             = "services/hostedservices/%s/deployments/%s/roles/%s"
 	azureOperationsURL       = "services/hostedservices/%s/deployments/%s/roleinstances/%s/Operations"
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
LGTM. Although this is breaking the API signature, I think we can take it as we've done it many times in the past for `management` pkg and we're not tracking it in version tags. @paulmey PTAL.

kbxkb:
Though Terraform is just one user of azure-sdk-for-go, DeleteRole() is not called from Terraform (yet, before my fixes). There may be other users out there who may get impacted by the API signature modification. Having said that, I do agree with @ahmetalpbalkan that this is a necessary change. Thanks!

ahmetalpbalkan:
@kbxkb we always suggest other projects to vendor a specific tag/commit.@paulmey PTAL and merge as you see fit.

kbxkb:
@paulmey can you please take a look at your convenience, thanks!

paulmey:
LGTM

paulmey:
Sorry for the delay and thanks for contributing!

davem-git:
What version is this fix in?

PR #242, opened 2015-11-25
Automatically detect mime type of block uploads

kjnsn:


ahmetalpbalkan:
@kjnsn looks like a safe solution. There might be clients relying on the existing behavior but I can hardly think this would be an issue. Do you mind adding this for `PutBlockWithLength` method as well?

storage/blob.go
```diff
@@ -629,7 +630,13 @@ func (b BlobStorageClient) CreateBlockBlob(container, name string) error {
 func (b BlobStorageClient) CreateBlockBlobFromReader(container, name string, size uint64, blob io.Reader) error {
 	path := fmt.Sprintf("%s/%s", container, name)
 	uri := b.client.getEndpoint(blobServiceName, path, url.Values{})
+	splitted := strings.Split(name, ".")
+	if len(splitted) == 0 {
+		splitted = []string{""}
```
ahmetalpbalkan: instead of this empty slice hack can you please do:

    if len(splitted) > 1 { headers["Content-Type"] = mime.TypeByExtension("." + splitted[len(splitted)-1])  }

ahmetalpbalkan:
@kjnsn Hello, if you could update your pull request with the comment above we can easily take this in!

kjnsn:
@ahmetalpbalkan changes made as suggested

ahmetalpbalkan:
@kjnsn sorry for the inactivity here. `CreateBlockBlobFromReader()` now supports extra headers as an argument. I am not completely sure we should be detecting and setting the Content-Type based on blob file name extension as it may create undesired consequences for some clients. You can do the MIME type detection in your code and set the Content-Type header in `extraHeaders` argument.

PR #249, opened 2015-12-12
Make DeleteDisk return synchronously

paulmey:
Doc says it returns synchronously. Prevent users from doing an extra "get operation" call.


ahmetalpbalkan:
LGTM

paulmey:
Thanks @ahmetalpbalkan !

PR #256, opened 2015-12-28
Added QueryParameterName and QueryParameterValue to Error func output

alvadb:


ahmetalpbalkan:
@alvadb thanks for your pull request. Can you please gofmt your code?

alvadb:
gofmt has been done.

ahmetalpbalkan:
@alvadb thanks, looks good. Also remember that for complicated "value" strings, instead of %s, %q is better as it does escaping of new lines and non-printable characters and surrounds strings with quotes unlike %s.

Issue #209, opened 2015-09-10
storage - allow setting headers on PutBlob request


kenegozi:
Add an argument to the PutBlockBlob call suggested in issue #208 to allow the user to set custom headers. The most important one is Content-Type, e.g. without this one images that are uploaded to blob storage cannot be treated as images by browser requests.

Since PutBlockBlob is a new API call on the SDK, it should be fine to add the trailing argument to it and not the others. Also the others can utilize the suggested #207 to set these headers, however the whole idea in this one (PutBlockBlob) is to allow a simple flow with a single API call.

ahmetalpbalkan:
@kenegozi we can still break. Reiterating my comment on #208, we will come up with a non-existing method, so we can potentially get a `BlobProperties` parameter in the method.

Also, in case of block uploads and then commits with PutBlockList, the PutBlockList call also takes all those properties (such as x-ms-content-type, as you said) so we should probably add that parameter initially to `PutBlockList` (will be a breaking change, but fine for now) and the `PutBlockBlob` that'll come up can just make use of that change to `PutBlockList`.

How does this sound? Want to contribute?

tomclegg:
This feature would make it possible to supply [conditional headers](https://msdn.microsoft.com/en-us/library/azure/dd179371.aspx), which I currently need to use with `DeleteBlob`. If an `addHeaders map[string]string` argument is going to be the way to do this, eventually we'll need it for every API method. Perhaps it should all get added to all methods at once?

ahmetalpbalkan:
@tomclegg I doubt if this should be a map[string]string. I prefer to have it a type (just like BlobProperties) or perhaps since it's really common we can come up with something separate, like `AccessCondition` that also exists in C# Storage SDK (https://github.com/Azure/azure-storage-net/blob/f525e18b485f65577d5528f20437a79977b084c3/Lib/Common/AccessCondition.cs#L26) and plumb that all the way deep in the call stack to handle it centrally in one place. This will be a breaking change but it is okay.

I'm +1 for idea, my preference would be having a struct for this rather than a map. What do you think?

kenegozi:
three possible solutions:

1. passing map[string]string (or map[string]interface{} and properly
dealing with dates)
    and providing consts for the various header values
pros: simplest solution to implement, minimal breaking changes, no
surprises and magic to users, forward-compatible with new headers, easy to
cover metadata as well
cons: less discoverability (although a set of consts such as
[...]

PR #257, opened 2016-01-05
Accept extra headers

Misterhex:
#209 

Accept extra headers as `map[string]string` on blob-related functions `CreateBlockBlobFromReader`, `PutPageBlob` and `PutBlockWithLength`.

This enable the use case of passing extra headers such as `content-type` with `image/jpeg` to allow azure to serve the blob with correct `content-type`.

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -614,7 +614,7 @@ func (b BlobStorageClient) GetBlobMetadata(container, name string) (map[string]s
 //
 // See https://msdn.microsoft.com/en-us/library/azure/dd179451.aspx
 func (b BlobStorageClient) CreateBlockBlob(container, name string) error {
-	return b.CreateBlockBlobFromReader(container, name, 0, nil)
+	return b.CreateBlockBlobFromReader(container, name, 0, nil, nil)
 }
 
 // CreateBlockBlobFromReader initializes a block blob using data from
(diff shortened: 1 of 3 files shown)


ahmetalpbalkan:
@Misterhex thanks for your PR. There's already an open pr #242 about content-type detection from blob extension. If that solves your problem, please let me know and I'll take it over. Otherwise, can you please add the same extraHeaders argument to PutPageBlob as well?

Misterhex:
@ahmetalpbalkan Sure, i will do that.

storage/blob.go
```diff
@@ -649,7 +649,14 @@ func (b BlobStorageClient) CreateBlockBlobFromReader(container, name string, siz
 //
 // See https://msdn.microsoft.com/en-us/library/azure/dd135726.aspx
 func (b BlobStorageClient) PutBlock(container, name, blockID string, chunk []byte) error {
-	return b.PutBlockWithLength(container, name, blockID, uint64(len(chunk)), bytes.NewReader(chunk))
+	return b.PutBlockWithLength(container, name, blockID, uint64(len(chunk)), bytes.NewReader(chunk), nil)
+}
+
+func (b BlobStorageClient) PutBlockWithContentType(container, name, blockID string, chunk []byte, contentType string) error {
```
ahmetalpbalkan: I think this is pretty trivial to achieve with the method below. Let's remove this method as achieving this is as simple as passing a simple map initializer value to the other method. Also the CI build fails because there's no godoc comment on this.

ahmetalpbalkan:
@Misterhex did you close to push new changes? if so you can reopen and add extra commits.

Misterhex:
@ahmetalpbalkan i don't see how that could fit my use case. for example, if name is "123.jpg", according to pr #242 , content-type would be set to "jpg" where what needed is "image/jpeg".

storage/blob.go
Misterhex: Agree with the having only the method below. 

It is also probably a good idea to merge the extra headers into the default, overriding if key matches.

Thoughts?

Misterhex:
@ahmetalpbalkan have pushed the new changes.

ahmetalpbalkan:
@Misterhex apparently this change is breaking the blob_test.go, can you please fix the tests?

Misterhex:
@ahmetalpbalkan will do

ahmetalpbalkan:
@Misterhex I think the test is still broken:

```
# github.com/Azure/azure-sdk-for-go/storage
storage/blob_test.go:447: not enough arguments in call to cli.CreateBlockBlobFromReader
```

ahmetalpbalkan:
lgtm. thanks @Misterhex

Issue #260, opened 2016-01-13
DNSZone Client


stack72:
According to the [docs](https://msdn.microsoft.com/en-us/library/mt130622.aspx) for creating or updating a zone, the request body should be:

```
{
  "location": "global",
  "tags": {},
  "properties": { }
}
```

The go-sdk has changed this as follows:

```
client := meta.(*ArmClient)
	dnsZonesClient := client.dnsZonesClient

	log.Printf("[INFO] preparing arguments for Azure ARM DNS Zone creation.")

	name := "mytestzone"
	location := "global"
	resGroup := d.Get("resource_group_name").(string)

	zone := dns.Zone{
		Location:   &location,
		Properties: &dns.ZoneProperties{},
	}

	zoneParams := dns.ZoneCreateOrUpdateParameters{
		Zone: &zone,
	}

	resp, err := dnsZonesClient.CreateOrUpdate(resGroup, name, zoneParams, "", "*")
	if err != nil {
		return err
	}
```

This makes the request look as follows:

```
{
	"Zone": {
		"location": "global",
		"properties": {}
	}
}
```

This throws an error as follows:

```
{"error":{"code":"InvalidRequestContent","message":"The request content was invalid and could not be deserialized: 'Could not find member 'Zone' on object of type 'ResourceDefinition'. Path 'Zone', line 1, position 8.'."}}
```

This feels like the `CreateOrUpdate` func is incorrect **OR** the docs are incorrect

Thoughts?

Paul

brendandixon:
The DNS Swagger (see https://github.com/Azure/azure-rest-api-specs/blob/master/arm-dns/2015-05-04-preview/swagger/dns.json#L383 and https://github.com/Azure/azure-rest-api-specs/blob/master/arm-dns/2015-05-04-preview/swagger/dns.json#L894) disagrees with the documentation. Please file a Github issue against the Swagger and I'll poke around internally.

Thank you raising this issue.

stack72:
Hi @brendandixon

Thanks for the response. Even if the docs are wrong, by following the code in the SDK, the request returns an error. What should the actual request struct look like?

Thanks

Paul

brendandixon:
@stack72 My guess is that the documentation is correct and the SDK is wrong (meaning, the underlying Swagger is wrong). I forwarded the issue to someone on the ARM DNS team. They should respond with better guidance.

stack72:
[...]

PR #266, opened 2016-01-21
Updated DNS to latest Swager

brendandixon:
This should address #260.


ahmetalpbalkan:
@brendandixon looks like you dropped gocheck from vendored dependencies. Intentional?

brendandixon:
Nope. I wonder how that happened...

brendandixon:
Actually, it seems I corrected that mistake. It should be there now.

brendandixon:
Ok to merge? This addresses (we hope) #260. It is intended to update *only* the arm/dns client.

ahmetalpbalkan:
lgtm

PR #269, opened 2016-01-28
Switch from Godeps to Glide

colemickens:
Let me know if you want me to hold off on this for sequencing with other changes.

PR #271, opened 2016-01-29
updated for Autorest changes

garimakhulbe:
-	client.Send(req, codes..int)  -> client.Send(req)
-	autorest.WithErrorUnless… → azure.WithErrorUnless..
-       NewErrorWithError(resonse.statusCode) -> autorest.NewErrorWithError(response).


ahmetalpbalkan:
build is failing because it's referring to Godeps/ even though you removed it from import statements. Maybe we should remove Godeps/ here before merging #269.

ahmetalpbalkan:
Changes LGTM. There's zero diff with my locally-patched version of the SDK to get docker-machine unblocking with latest autorest changes etc. Thanks!

ahmetalpbalkan:
This now has merge conflicts. (my guess is github's conflict resolver is not super smart and even though the conflicts with #269 are actually the same lines and same diffs, it is still failing).

brendandixon:
I merged #269 first because it gives us a cleaner starting point. It's trivial for @garimakhulbe to rebuild the packages contained in this PR. I'll work with her to get those built.

colemickens:
This is passing now.

Bug filed on glide: https://github.com/Masterminds/glide/issues/235

brendandixon:
W00t!

brendandixon:
@colemickens @garimakhulbe Please update the CHANGELOG to reflect the version bump and so forth. Then let's merge.

ahmetalpbalkan:
LGTM

brendandixon:
@ahmetalpbalkan Funny you should ask. We should evaluate placing each arm package under date-stamped directories.

ahmetalpbalkan:
@brendandixon not sure I follow. what do you mean?

colemickens:
LGTM.

arm/CHANGELOG.md
```diff
@@ -21,9 +21,9 @@ updated in this release):
 - *arm/notificationhubs 2014-09-01*
 - arm/redis 2015-08-01
 - *arm/resources/authorization 2015-01-01*
-- *arm/resources/features 2014-08-01-preview*
```
ahmetalpbalkan: @garimakhulbe this is wrong. you should not edit existing changelog item. The part you edited was for v0.4.0, which is already released. (but not tagged, @brendandixon any ideas why? again makes me think we should have a checklist for releases)

anyhow, please create a new section (such as 0.5.0 and add the changes for this version there).

garimakhulbe: I think I need to update this (v0.5.0) in version template and regenerate the SDK. Right?

const ( 
26 	major = "0" 
27 	minor = "4" 
28 	patch = "0" 
29 	// Always begin a "tag" with a dash (as per http://semver.org) 
30 	tag             = "-beta" 
31 	semVerFormat    = "%s.%s.%s%s" 
32 	userAgentFormat = "Azure-SDK-for-Go/%s;Package arm/%s;API %s" 
33 )

brendandixon:
It appears all issues have been addressed. Merging.

colemickens:
So is the ARM subpackage versioned differently from the rest of `azure-sdk-for-go`? I ask because it seems we just bumped to 0.5.0 in the code, but I see that we already have git tags for `1.0`, `1.1`, and `1.2`.

If this is the case, is there any tagging that needs to be done as a result of this merge?

ahmetalpbalkan:
@colemickens yeah they're independent. I think we should pin to go-autorest latest commit now (as changelog says) with glide.

brendandixon:
@garimakhulbe @colemickens The ARM packages are in bit of an odd position vis a vis the rest of the SDK. The other packages are well past beta, but the ARM packages must be beta given the state of the Swagger. Having multiple semantic versions feels wrong to me, but I don't have a better solution. I am willing to listen!

colemickens:
@brendandixon I'm fine with the versioning as is, just wanted to make sure we tagged if we needed to.

@ahmetalpbalkan I'm torn on what to do here. We can pin it in the `glide.yaml` file. This would make it marginally safer to run `glide up` and not worry that `go-autorest` is going to be pulled up to an incompatible version. That having been said, we could have breaking changes from any dependency, at any time... and pinning in the `glide.yaml` means that someone has to know to check there first, before running `glide up` to determine if they're going to actually update a dependency or not.

I guess I would use it more like this, though I don't have a very strong opinion about this: leave glide.yaml alone. Run `glide up` with some regularity when you want updated packages. Run unit tests to ensure things are okay. If you find that ANY dependency is being pulled up to too high of a version, **then** pin it in glide.yaml. That way it's a more conscious, temporary thing to do.

PR #273, opened 2016-01-29
Add encrypted storage key for Travis

colemickens:
We are currently leaking our storage key on Travis.

This commit adds the secondary storage key to the .travis.yml file in an encrypted format (encrypted with the public portion of a key pair Travis creates for our repo; they keep the private key).

Travis smartly prevents pull requests from receiving encrypted data. Otherwise Jane Schmane could send a PR that just reads the environment variable and POSTs it to her server.

This means that storage tests will always fail for PRs. Or, it means we disable storage tests in that scenario on pull requests (Travis gives us an env variable for pull requests). That option makes me quite nervous, because it means that Travis is no longer 100% authoritative on whether or not the commit is safe to merge. Green on Travis for a PR would no longer mean "100% test pass", it would mean "100% test pass except storage tests which may or may not pass".

I'm trying to find out if there's a way to tell Travis that it's okay to give the credential out. This is a pattern I've seen on larger repos. PRs aren't automatically built, but rather they wait until a Collaborator comes along and leaves a comment that a bot reads and then submits the CI build. It would be smart of Travis to give us Collaborators a way of saying "Travis, please retest this with the encrypted environment variables now that I've code reviewed this for obvious malicious intent.

I have a support ticket open with Travis to try to understand what our options are.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+env:
+  secure: "FFgOmjPVglBgZPUeDjPXWs0HTjNOSjLDuTjrOS156z49oa/JtY28eSDZTtUwTW/Mw9pTGkVc0BtaMdDRI+MTZh30rOOEtFJrjF1y5HhCkw6+s0L2SAOZEoUviJTD/+wqtE7ZNYGRnMJ9svu437IXQTtsxmQjMcwmVOmQBNwP+0c="
+
 sudo: false
 
 language: go


colemickens:
(PR sent mostly for discussion for now; not asking for this to be merged yet)

ahmetalpbalkan:
Having private (not encrypted) was also causing PRs from other users to not to be built in the past as well, that's the whole reason we made the key intentionally public in the first place. storage tests almost always catch an issue, it's very good to have them.

I think if we were to take this PR you'll still need to disable on travis-ci.org UI to remove the plain text key we inject.

colemickens:
>I think if we were to take this PR you'll still need to disable on travis-ci.org UI to remove the plain text key we inject.

Correct.

>Having private (not encrypted) env was also causing PRs from other users to not to be built in the past as well, that's the whole reason we made the key intentionally public in the first place. storage tests almost always catch an issue, it's very good to have them.

Agreed.

I just get nervous every time I see it in the build output, hence sending this in case anyone saw an approach that avoided issues on both sides. At the very least, it's something we should be conscious of and keep an eye on. (I'd also noticed the VHDs this morning and was a bit surprised but they all had "SDK" in the name so I wasn't worried about malice).

It seems to me that it would be a highly used feature if Travis let maintainers "approve" PRs to be built with the encrypted environment populated. ..... although based on this thread I wouldn't hold our breath for it: https://github.com/travis-ci/travis-ci/issues/1946

ahmetalpbalkan:
I'm ok proceeding with this. Having myself recently tried a similar feature using travis rubygem, it works just fine.

colemickens:
I don't follow. How is the problem with Travis not giving encrypted keys to
PRs resolved?

ahmetalpbalkan:
@colemickens oh totally forgot about that. in that case we can't really solve this one. close?

colemickens:
Yep. I was hoping the Travis folks would get back to me with some sort of workaround but that hasn't happened.

PR #277, opened 2016-02-05
Updated for godocs..

garimakhulbe:
Added comments for godocs.

### DIFF
--- a/arm/compute/client.go
+++ b/arm/compute/client.go
@@ -1,3 +1,6 @@
+// Package compute implements the Azure ARM Compute service API version
+// 2015-06-15.
+// the Compute Management Client.
 package compute
 
 // Copyright (c) Microsoft and contributors.  All rights reserved.
@@ -30,7 +33,7 @@ const (
 	DefaultBaseURI = "https://management.azure.com"
 )
 
-// ManagementClient is the the Compute Management Client.
+// ManagementClient is the base client for Compute.
 type ManagementClient struct {
 	autorest.Client
 	BaseURI        string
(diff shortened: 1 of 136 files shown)


ahmetalpbalkan:
only one file per package should have the godoc comment, not all files in the package. actually the convention is to have a `doc.go` in each package that just has the godoc comment, but no need to do it here as long as it's on a single file.

also pls join the azure linux team at aka.ms/azuregithub.

arm/network/client.go
```diff
@@ -1,3 +1,10 @@
+// Package network implements the Azure ARM Network service API version
+// 2015-06-15.
+//
+// The Microsoft Azure Network management API provides a RESTful set of web
+// services that interact with Microsoft Azure Networks service to manage
+// your network resrources. The API has entities that capture the
+// relationship between an end user and the Microsoft Azure Networks service..
```
ahmetalpbalkan: seems like we have a duplicate `..` problem in some packages at the end.

arm/compute/client.go
```diff
@@ -1,3 +1,7 @@
+// Package compute implements the Azure ARM Compute service API version
+// 2015-06-15.
+//
+// The Compute Management Client..
```
ahmetalpbalkan: this doesn't look very good. :( maybe omit what services describe in swagger and just keep the first line you have? It just seems sufficient to me.

ahmetalpbalkan: it's actually not a big deal, we can merge as the first line is what's going to be shown at https://godoc.org/github.com/Azure/azure-sdk-for-go/arm

brendandixon: Let's merge. The Swagger documentation needs work.

arm/network/client.go
garimakhulbe: I fixed this in latest PR.

Issue #275, opened 2016-02-02
storage - add support for blob type 'Append'


dovreshef:
As detailed here: http://blogs.msdn.com/b/windowsazurestorage/archive/2015/04/13/introducing-azure-storage-append-blob.aspx

preetpalS:
Isn't this issue solved? (https://godoc.org/github.com/Azure/azure-sdk-for-go/storage#BlobType)

ahmetalpbalkan:
@preetpalS Ah, right. This is contributed in #278. Closing.

PR #278, opened 2016-02-09
[storage] Adds support for Append Blobs

BrianBland:
Adds PutAppendBlob and AppendBlock methods

Addresses #275

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -167,8 +167,9 @@ type BlobType string
 
 // Types of page blobs
 const (
-	BlobTypeBlock BlobType = "BlockBlob"
-	BlobTypePage  BlobType = "PageBlob"
+	BlobTypeBlock  BlobType = "BlockBlob"
+	BlobTypePage   BlobType = "PageBlob"
+	BlobTypeAppend BlobType = "AppendBlob"
 )
 
 // PageWriteType defines the type updates that are going to be
@@ -330,7 +331,6 @@ func (b BlobStorageClient) createContainer(name string, access ContainerAccessTy
 	uri := b.client.getEndpoint(blobServiceName, pathForContainer(name), url.Values{"restype": {"container"}})
 
 	headers := b.client.getStandardHeaders()
-	headers["Content-Length"] = "0"
 	if access != "" {
 		headers["x-ms-blob-public-access"] = string(access)
 	}
@@ -556,7 +556,6 @@ func (b BlobStorageClient) SetBlobMetadata(container, name string, metadata map[
 	for k, v := range metadata {
 		headers[userDefinedMetadataHeaderPrefix+k] = v
 	}
-	headers["Content-Length"] = "0"
 
 	resp, err := b.client.exec("PUT", uri, headers, nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ahmetalpbalkan:
@BrianBland thanks for the pull request. LGTM. `extraHeaders` will let users make use of `x-ms-blob-condition-maxsize` and `x-ms-blob-condition-appendpos`, which might be frequently used in the use cases of AppendBlob.

It looks like there's something funny going on about our storage tests. @colemickens have you changed something lately?

BrianBland:
@ahmetalpbalkan it looks like I need to change some things up to support the new signing code as of version 2015-02-21 (see https://msdn.microsoft.com/en-us/library/azure/mt584141.aspx)

ahmetalpbalkan:
@BrianBland oh that explains the problem. We were actually planning to switch on SharedKeyLite signing (which is way simpler, doesn't have Content-Length either). If you don't mind making the change to work with `2015-02-21`, please go ahead.

BrianBland:
Hopefully you don't mind the repeated push trial and error with the CI system. I'm temporarily locked out of my Azure account :disappointed:

BrianBland:
@ahmetalpbalkan if you have any idea why `TestBlobSASURICorrectness` is failing with the new API version, I'm ready to fix it. I can't find any documentation suggesting why this is suddenly invalid.

colemickens:
(Won't be near a computer for a while but I haven't changed anything for
storage tests)

ahmetalpbalkan:
@BrianBland oh that's just comparing a precomputed signature with the existing algorithm. You just need to change the hard-coded computed value with the result of then modified algorithm.

BrianBland:
@ahmetalpbalkan I think you're referring to `TestGetBlobSASURI`. The failing test is getting a 403 status code from the server

BrianBland:
Apparently the canonicalizedResource for the presigned blob urls must begin with "/blob", even though [the documentation](https://msdn.microsoft.com/en-us/library/azure/dn140255.aspx) suggests that it should begin with "blob"

ahmetalpbalkan:
@BrianBland I'll follow up on that. Might be a docs issue indeed.

storage/blob.go
```diff
@@ -801,6 +802,49 @@ func (b BlobStorageClient) GetPageRanges(container, name string) (GetPageRangesR
...
+// PutAppendBlob initializes an empty append blob with specified name. An
+// append blob must be created using this method before appending blocks.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179451.aspx
+func (b BlobStorageClient) PutAppendBlob(container, name string, extraHeaders map[string]string) error {
+	path := fmt.Sprintf("%s/%s", container, name)
+	uri := b.client.getEndpoint(blobServiceName, path, url.Values{})
+	headers := b.client.getStandardHeaders()
+	headers["x-ms-blob-type"] = string(BlobTypeAppend)
+	headers["Content-Length"] = "0"
```
ahmetalpbalkan: @BrianBland Hmm, I don't think you will still need this header now that you updated `buildCanonicalizedString` with

```diff
+	contentLength := headers["Content-Length"]
+	if contentLength == "0" {
+		contentLength = ""
+	}
```

In fact, there should be more occurrences of `headers["Content-Length"] = "0"` in this file, I think we can remove all those if you don't mind doing this in PR as well?

BrianBland: No worries, removed all of those

ahmetalpbalkan:
@BrianBland thanks LGTM!

ahmetalpbalkan:
@BrianBland hi there, just circling back on the docs issue here https://msdn.microsoft.com/en-us/library/azure/dn140255.aspx I see that it says `canonicalizedresource = "/blob/...`, did you see a different value back then?

> image

BrianBland:
I thought so, but at least it looks correct now

PR #279, opened 2016-02-10
[storage] Don't hide 404 status codes in an opaque error

BrianBland:
404 errors from the storage API don't provide message bodies, so don't hide these inside an opaque fmt.Errorf()

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -325,7 +325,7 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 	}
 
 	statusCode := resp.StatusCode
-	if statusCode >= 400 && statusCode <= 505 {
+	if statusCode >= 400 && statusCode <= 505 && statusCode != 404 {
 		var respBody []byte
 		respBody, err = readResponseBody(resp)
 		if err != nil {


ahmetalpbalkan:
@BrianBland hmm I am not sure I could follow the need to do this. I feel like methods such as `GetBlob` would not fail for non-existing blobs (404 responses) whereas they should fail?

BrianBland:
The intent was to be able to call methods such as `GetBlobProperties` and detect if the blob does not exist without a separate call to `BlobExists`, but I could just keep the extra round trip for now

ahmetalpbalkan:
@brianbland, alright feel free to close.

BrianBland:
Sure thing, I'll punt this to a later time, probably after I get the new storage drivers merged from docker/distribution#1438

ahmetalpbalkan:
@BrianBland I'm just realizing this is for the storage driver! I've been advocating for append interfaces for a longtime in registry, glad to see it's finally happening. The ability to make random files on registry blobs made me do some nasty hacks in azure driver in the past. If you're switching to an append-only mode, you could possibly delete a ton of code from azure driver.

BrianBland:
Exactly, sorry about the convoluted random write behavior. Turns out that only makes the system more prone to eventual consistency issues with backends such as S3/Swift. Feel free to check out https://github.com/docker/distribution/issues/1395 to review the new `StorageDriver` interface, or even just the new Azure driver code in the WIP PR.

ahmetalpbalkan:
@BrianBland will do, thanks.

PR #280, opened 2016-02-11
Add support of PostShudownAction for shutdown vm

razrin:
The implementation of Shutdown Role is incomplete: https://msdn.microsoft.com/en-us/library/azure/jj157195.aspx

The parameter PostShudownAction was not implemented

### DIFF
--- a/management/virtualmachine/client.go
+++ b/management/virtualmachine/client.go
@@ -209,7 +209,7 @@ func (vm VirtualMachineClient) StartRole(cloudServiceName, deploymentName, roleN
 	return vm.client.SendAzurePostRequest(requestURL, startRoleOperationBytes)
 }
 
-func (vm VirtualMachineClient) ShutdownRole(cloudServiceName, deploymentName, roleName string) (management.OperationID, error) {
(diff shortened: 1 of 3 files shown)


management/virtualmachine/client.go
```diff
@@ -209,7 +209,7 @@ func (vm VirtualMachineClient) StartRole(cloudServiceName, deploymentName, roleN
 	return vm.client.SendAzurePostRequest(requestURL, startRoleOperationBytes)
 }
 
-func (vm VirtualMachineClient) ShutdownRole(cloudServiceName, deploymentName, roleName string) (management.OperationID, error) {
+func (vm VirtualMachineClient) ShutdownRole(cloudServiceName, deploymentName, roleName string, postactions ...PostShutdownAction) (management.OperationID, error) {
```
ahmetalpbalkan: You don't have to make this argument variadic. We don't have much API breaking concerns for now. Feel free to make it a single argument as you're using only the first element.

ahmetalpbalkan:
@razrin generally LGTM. thanks for your contribution. we can fix once you issue the variadic argument issue.

razrin:
Any news on this pull request ?
Does i need to do anything ?

ahmetalpbalkan:
@razrin looks good. thanks.

PR #281, opened 2016-02-11
Updated for go-autorest v4.0.0 (LRO support)

garimakhulbe:


arm/cdn/customdomains.go
```diff
@@ -100,19 +100,27 @@ func (client CustomDomainsClient) CreatePreparer(customDomainName string, custom
...
-func (client CustomDomainsClient) CreateResponder(resp *http.Response) (result CustomDomain, err error) {
+func (client CustomDomainsClient) CreateResponder(resp *http.Response) (result autorest.Response, err error) {
 	err = autorest.Respond(
 		resp,
 		client.ByInspecting(),
 		azure.WithErrorUnlessStatusCode(http.StatusOK, http.StatusCreated, http.StatusAccepted),
 		autorest.ByUnmarshallingJSON(&result),
 		autorest.ByClosing())
-	result.Response = autorest.Response{Response: resp}
+	result = autorest.Response{Response: resp}
```
colemickens: I'm not understanding how this code works.

`autorest.ByUnmarshallingJSON(&result)` unmarshals into the named out parameter, but then it is clobbered by `result = autorest.Response{Response: resp}` which seems to just include the raw response.

ahmetalpbalkan: Yeah long-running methods shouldn't have ByUnmarshallingJSON , we're really returning the end-result of polling, which doesn't map to a specific entity type we have. It's probably not failing, because json deserialization is not strict, it'll skip the extra or not-found fields.

colemickens: I think maybe it just needs to be reordered.

colemickens: Oh I didn't see your comment, my internet is spotty.

Agreed. I'm more concerned about the fact that we don't have any tests the covered checking the fields being populated.

glide.lock
```diff
@@ -1,21 +1,20 @@
...
-  - /autorest
+  - autorest
   - autorest/azure
   - autorest/date
   - autorest/to
 - name: github.com/dgrijalva/jwt-go
-  version: afef698c326bfd906b11659432544e5aae441d44
-  subpackages:
-  - .
+  version: f2193411bd642f7db03249fd79d5292c9b34916a
```
ahmetalpbalkan: Hmm I think the only changes to this file should be on "autorest" item (and for that only the commit number should be changing). Is glide updating everything altogether? If there's no clear need we probably should not update jwt-go or pkcs12.

arm/cdn/customdomains.go
```diff
@@ -53,7 +53,7 @@ func NewCustomDomainsClientWithBaseURI(baseURI string, subscriptionID string) Cu
 // endpointName is name of the endpoint within the CDN profile profileName is
 // name of the CDN profile within the resource group resourceGroupName is
 // name of the resource group within the Azure subscription
-func (client CustomDomainsClient) Create(customDomainName string, customDomainProperties CustomDomainParameters, endpointName string, profileName string, resourceGroupName string) (result CustomDomain, ae error) {
+func (client CustomDomainsClient) Create(customDomainName string, customDomainProperties CustomDomainParameters, endpointName string, profileName string, resourceGroupName string) (result autorest.Response, ae error) {
```
ahmetalpbalkan: We should add documentation to these long-running operation methods (e.g. `Create` and `CreateSender`), indicating:

1. SDK will do the polling of this long running operation for you.
2. Returned `autorest.Response` contains the final HTTP response from the polling operation.

We _really_ need to make this situation clear as I stated in my response to the offline thread.

glide.lock
colemickens: `glide up` will update all dependencies. I opened [a feature request for `glide up` to accept a specific package to upgrade](https://github.com/Masterminds/glide/issues/252).

The other subtle changes (`/autorest`->`autorest`, removal of `.` subpackage) are related to https://github.com/Masterminds/glide/issues/235

I'm interested to hear opinions or read about updating strategies (and that doesn't have to be here on this PR). I've heard both - "why wouldn't you want to take new features/fixes" and "why would you upgrade something that works". Go's lack of versioning at the package level makes me lean towards being more cautious as well. Although the compiler will complain if API-breaking changes were introduced, one could imagine a breaking change to a library (changes functionality) without changing the API.

garimakhulbe: Yes, glide up command updates everything.

ahmetalpbalkan: I guess we can live with this approach until something breaks. Most of these packages are not very dynamic and less likely to behaviorally break.

arm/dns/zones.go
```diff
@@ -167,8 +167,7 @@ func (client ZonesClient) DeleteResponder(resp *http.Response) (result autorest.
 	err = autorest.Respond(
 		resp,
 		client.ByInspecting(),
-		azure.WithErrorUnlessStatusCode(http.StatusOK),
-		autorest.ByClosing())
```
ahmetalpbalkan: seems unintentional

arm/cdn/customdomains.go
brendandixon: @ahmetalpbalkan @colemickens The code is incorrect. I'm working with @garimakhulbe to make the appropriate changes to the generator. This PR will then be updated. :)

arm/authorization/classicadministrators.go
```diff
@@ -92,8 +92,8 @@ func (client ClassicAdministratorsClient) ListSender(req *http.Request) (*http.R
 
 // ListResponder handles the response to the List request. The method always
 // closes the http.Response Body.
-func (client ClassicAdministratorsClient) ListResponder(resp *http.Response) (result ClassicAdministratorListResult, err error) {
-	err = autorest.Respond(
+func (client ClassicAdministratorsClient) ListResponder(resp *http.Response) (result ClassicAdministratorListResult, ae error) {
+	ae = autorest.Respond(
```
ahmetalpbalkan: let's call this `err` as some linter tools rely on the variable name IIRC.

ahmetalpbalkan:
Arghh, this polling logic is not working as I outlined earlier (my mistake).

Apparently RPs can return `200 OK` when the operation status is still InProgress:

```
GET https://management.azure.com/subscriptions/[...]/providers/Microsoft.Compute/locations/westus/operations/eb27e532-0794-424e-98c9-42c7b57342f8

HTTP/1.1 200 OK
[headers]

{
    "operationId": "eb27e532-0794-424e-98c9-42c7b57342f8",
    "startTime": "2016-02-17T18:59:48.0895519+00:00",
    "status": "InProgress"
}
```

I think we really need to have the xxxOperationStatus objects .NET SDK has and check their `.Status` field to determine whether we should continue polling or not. Apparently HTTP 200 is not really an indicator of anything on “Get Operation Status” endpoints.

ahmetalpbalkan:
@garimakhulbe thanks for the PR. When https://github.com/Azure/go-autorest/pull/40 is merged, let's update glide manifest to that one. This seems to be working fine for me. Also feel free to fix the `go vet` error by adding `| grep -v './arm/resources$` to `find ..` in travis.yml.

garimakhulbe:
Build successful. Finally. :+1: 
Please review.

ahmetalpbalkan:
LGTM. Thanks.

PR #286, opened 2016-03-01
Remove unused packages of patched legacy stdlibs

ahmetalpbalkan:
`rm -rf core/http/{cookiejar cgi fcgi httptest httputil pprof testdata}`

we don't need those packages and testdata as we never encourage others to use the core/http/** or run their tests.


paulmey:
LGTM, is travis borked?

ahmetalpbalkan:
yeah I can't tell what's going on even though I have org owner permissions.

ahmetalpbalkan:
amend/push -f triggered it now.

PR #287, opened 2016-03-04
Ignore little warning in travis build

ahmetalpbalkan:
./arm/resources$ contains no files, it makes error message confusing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ script:
   - test -z "$(gofmt -s -l -w management | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w storage | tee /dev/stderr)"
   - go build -v ./...
-  - test -z "$(go vet $(find ./arm/* -type d -print) | tee /dev/stderr)"
+  - test -z "$(go vet $(find ./arm/* -type d -print | grep -v '^./arm/resources$') | tee /dev/stderr)"
   - test -z "$(golint ./arm/... | tee /dev/stderr)"
   - go test -v ./storage/... -check.v
   - test -z "$(golint ./storage/... | tee /dev/stderr)"


ahmetalpbalkan:
fixes the issue it's supposed to fix.
```
$ go vet ./storage/...
storage/blob_test.go:141: no formatting directive in Fatalf call
exit status 1
The command "go vet ./storage/..." exited with 1.
```

remains as a mystery, I can't locally repro this error on my computer.

colemickens:
```
 eval "$(gimme tip)"
go version devel +80e5b52 Fri Mar 4 23:15:43 2016 +0000 linux/amd64
```

Wait, are we using tip of golang to run this? If so, then this is the reason we've just started seeing that failure in Travis: https://github.com/golang/go/commit/867910ea170cad67ef929ae472f1e19052da7d7c (relevant: https://github.com/golang/go/issues/12294)

ahmetalpbalkan:
@colemickens we've been always using tip. Event though we don't always use tip, that'll fix the issue for `go vet` (because it's part of go.tool) but `go lint` etc will still be pulled from master.

Thanks for pointing to the bug, should we pin to go1.5?

ahmetalpbalkan:
@colemickens wow wait this `tip` actually gives you the development version of go. I'm pinning this to 1.6 momentarily.

colemickens:
It doesn't really matter to me. I generally like being on the bleeding edge and doubt that being on `tip` will cause us to miss anything in regards to the Travis runs.

If anything we should just point out that the change to `tip` seems to regress their format-string-check in some cases -- seeing as that `storage` hit is a false positive, the format string definitely has format placeholders in it.

(I had forgotten that `golint` was a separate tool; I similarly don't have strong feelings about which version is used).

ahmetalpbalkan:
I was assuming `tip` actually means tip of stable feed.

colemickens:
from above:

```
eval "$(gimme tip)"
go version devel +80e5b52 Fri Mar 4 23:15:43 2016 +0000 linux/amd64
```

I don't think that's `stable`.

ahmetalpbalkan:
https://github.com/travis-ci/gimme/issues/39

PR #288, opened 2016-03-04
Pin travis build to go:1.6

ahmetalpbalkan:
Apparently `tip` gives development version of go and not the `tip` of the stable feed. (https://github.com/travis-ci/gimme/issues/39)

We ought to increment this version number as new go stable versions come up.


colemickens:
:+1: LGTM.

PR #289, opened 2016-03-06
add support for extra headers in AppendBlock

dovreshef:
A couple line difference. Implementation copied from other functions in the same file. Tested locally.


dovreshef:
Will resubmit as one commit

PR #290, opened 2016-03-06
add support for extra headers in AppendBlock

dovreshef:
A couple lines difference. Implementation taken from other functions in the file...


ahmetalpbalkan:
Thanks for the contribution, merging now.

PR #298, opened 2016-03-24
Reuse http.Client across requests, and allow caller to provide its own.

tomclegg:
Enables connection sharing with http keepalive, and allows callers to
use Go's other http client/transport features like request timeouts.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -37,6 +37,10 @@ const (
 // Client is the object that needs to be constructed to perform
 // operations on the storage account.
 type Client struct {
+	// HTTPClient is the http.Client used to initiate API
+	// requests.  If it is nil, http.DefaultClient is used.
+	HTTPClient *http.Client
+
 	accountName string
 	accountKey  []byte
 	useHTTPS    bool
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@tomclegg if I'm not mistaken, there should not be any difference between `http.Client{}` and `http.DefaultClient`. Do you have a use case that requires you to change the HTTP client?

tomclegg:
@ahmetalpbalkan relevant differences between `&http.Client{}` and `http.DefaultClient` are
* there is just one `http.DefaultClient`, vs. a new `&http.Client{}` each time you reach that LOC
* a caller can modify `http.DefaultClient` before it gets used

My main motivation is that I want to let Go's http library do connection pooling with http keepalive, which it does automatically if you use the same Client for multiple requests.

This happens to provide access to other stdlib features too, like setting a timeout for http requests. It also lets you stub the Azure service for testing, without resorting to changing the global http.DefaultTransport.

(Is there any reason the caller _shouldn't_ be allowed to provide its own `http.Client`?)

ahmetalpbalkan:
@tomclegg my only concern was we have another pr #217 coming in and it should probably rebase on top of this change. LGTM.

PR #301, opened 2016-03-25
Allow caller to provide headers for DeleteBlob methods.

tomclegg:
Same change as #257 (and other recent changes) ... but for DeleteBlob and DeleteBlobIfExists.

### DIFF
--- a/storage/blob_test.go
+++ b/storage/blob_test.go
@@ -218,7 +218,7 @@ func (s *StorageBlobSuite) TestBlobExists(c *chk.C) {
 	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypeBlob), chk.IsNil)
 	defer cli.DeleteContainer(cnt)
 	c.Assert(cli.putSingleBlockBlob(cnt, blob, []byte("Hello!")), chk.IsNil)
-	defer cli.DeleteBlob(cnt, blob)
+	defer cli.DeleteBlob(cnt, blob, nil)
 
 	ok, err := cli.BlobExists(cnt, blob+".foo")
 	c.Assert(err, chk.IsNil)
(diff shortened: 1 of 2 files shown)


storage/blob_test.go
```diff
@@ -273,13 +273,75 @@ func (s *StorageBlobSuite) TestDeleteBlobIfExists(c *chk.C) {
...
+	_, err = cli.GetBlob(cnt, blob)
+	c.Assert(err, chk.Not(chk.IsNil))
+
+	// Create a non-empty blob and set metadata in the hopes of
+	// triggering a "modified" event that will make an
+	// If-Unmodified-Since request give us a 412.
+	c.Assert(cli.CreateBlockBlobFromReader(cnt, blob, 1, bytes.NewReader([]byte{'-'}), nil), chk.IsNil)
+	c.Assert(cli.SetBlobMetadata(cnt, blob, map[string]string{"foo": "bar"}), chk.IsNil)
+
+	if false {
```
ahmetalpbalkan: dead code block?

ahmetalpbalkan:
this look fantastic thanks @tomclegg. LGTM overall.

storage/blob_test.go
tomclegg: I don't really know what to do with this. It would be a useful test, but (afaict) it fails because of a bug in Azure itself. So... just delete this section and pretend we never noticed? Or do the check, but if Azure incorrectly deletes the block, just re-write the block and log something but don't fail the test?

ahmetalpbalkan: @tomclegg there's probably something funky going on, perhaps an error silently ignored somewhere like time format issue or something. I'd drop this particular code block, I think `If-Match` provides enough coverage as far as the SDK is concerned. The rest is up to the service.

tomclegg:
@ahmetalpbalkan I've removed both `if-*modified-since` tests. (Without the Unmodified test, the Modified test can't really tell whether its header is being ignored, so seems pointless.)

ahmetalpbalkan:
@tomclegg you rock :metal:!

PR #302, opened 2016-03-25
Update for Polling and Asynchronous calls

garimakhulbe:
Uses go-autorest version v7.0.0.

### DIFF
--- a/arm/authorization/classicadministrators.go
+++ b/arm/authorization/classicadministrators.go
@@ -75,7 +75,8 @@ func (client ClassicAdministratorsClient) ListPreparer(apiVersion string) (*http
 		"api-version": APIVersion,
 	}
 
-	return autorest.Prepare(&http.Request{},
+	req := http.Request{}
+	return autorest.Prepare(&req,
 		autorest.AsJSON(),
 		autorest.AsGet(),
 		autorest.WithBaseURL(client.BaseURI),
--- a/arm/cdn/customdomains.go
+++ b/arm/cdn/customdomains.go
@@ -46,15 +46,19 @@ func NewCustomDomainsClientWithBaseURI(baseURI string, subscriptionID string) Cu
 	return CustomDomainsClient{NewWithBaseURI(baseURI, subscriptionID)}
 }
 
-// Create sends the create request.
+// Create sends the create request. This method handles polling itself for
+// long-running operations. User can cancel polling for long-running calls by
+// passing cancel channel as an argument to this method (cancel <-chan
+// struct{}). This channel won't cancel the operation, it will only stop the
+// polling.
 //
 // customDomainName is name of the custom domain within an endpoint
 // customDomainProperties is custom domain properties required for creation
 // endpointName is name of the endpoint within the CDN profile profileName is
 // name of the CDN profile within the resource group resourceGroupName is
 // name of the resource group within the Azure subscription
-func (client CustomDomainsClient) Create(customDomainName string, customDomainProperties CustomDomainParameters, endpointName string, profileName string, resourceGroupName string) (result autorest.Response, err error) {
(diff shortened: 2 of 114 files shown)


ahmetalpbalkan:
@garimakhulbe there's no go-autorest 7.0.0 yet? https://github.com/Azure/go-autorest

arm/authorization/classicadministrators.go
```diff
@@ -75,7 +75,8 @@ func (client ClassicAdministratorsClient) ListPreparer(apiVersion string) (*http
 		"api-version": APIVersion,
 	}
 
-	return autorest.Prepare(&http.Request{},
+	req := http.Request{}
+	return autorest.Prepare(&req,
```
ahmetalpbalkan: do you know why this diff has occurred? do we use `req` anywhere else?

garimakhulbe: http.Request Cancel channel can be used now to ignore polling for long-running.  So, request will be like...

 ```req := http.request{} ``` for non-lro 

```req := http.request{ Cancel:cancel } ``` for lro

File: arm/cdn/customdomains.go [hunk 2]
```diff
@@ -46,15 +46,17 @@ func NewCustomDomainsClientWithBaseURI(baseURI string, subscriptionID string) Cu
 	return CustomDomainsClient{NewWithBaseURI(baseURI, subscriptionID)}
 }
 
-// Create sends the create request.
+// Create sends the create request.This method handles polling itself. User
+// can make polling optional for long-running operations by passing cancel
+// channel in last argument of this method (cancel <-chan struct{}).
```
garimakhulbe: Added polling comment in method. @ahmetalpbalkan please review.

arm/cdn/customdomains.go
```diff
@@ -118,14 +122,18 @@ func (client CustomDomainsClient) CreateResponder(resp *http.Response) (result a
 	return
 }
 
-// DeleteIfExists sends the delete if exists request.
+// DeleteIfExists sends the delete if exists request.This method handles
```
ahmetalpbalkan: @garimakhulbe there should be a space after period `.` 

also the following sentence didn't seem grammatically correct:

    User can make polling optional for long-running operations

I don't understand what it means. I believe `cancel` channel is there for cancellation (not for avoiding LRO entirely) right?

garimakhulbe: Yes, Polling will still be there for lro, it can't be cancelled. It should not be "optional" in comment. I have updated it to "can ignore polling". Is this making sense? 

```
+// DeleteIfExists sends the delete if exists request.This method handles
+// polling itself for long-running operations. User can ignore polling for
 +// long-running calls by passing cancel channel in last argument of this
 +// method (cancel <-chan struct{}). Pass cancel channel if don't want to wait
 +// for polling to complete, else pass nil.
```

garimakhulbe: Correction: User can cancel the polling, but LRO functionality will remain the same. If I call CreateOrUpdate with cancel channel, it will still be the same call. Only difference will be that sdk won't do polling this time.

I will change the comments.

ahmetalpbalkan: @garimakhulbe It still isn't very clear to me, and probably will be confusing to the users as well.

What is the primary use case of this `chan`:

a. disable polling
b. cancel polling

if (a) I believe user already has ways to disable polling by reaching out to autorest.Client and flipping a flag there. if (b), then the documentation still doesn't make it clear that it's primarily for "cancellation" and not "opting out".

ahmetalpbalkan:
we can hash out the godoc issues later, LGTM

garimakhulbe:
Thanks for the comments. Merged.

PR #317, opened 2016-05-01
Add extra headers support to GetBlobRange

dovreshef:


ahmetalpbalkan:
@dovreshef thanks for your PR. generally LGTM. do you mind adding a test case for GetBlobRange that uses an extra header as well?

dovreshef:
@ahmetalpbalkan 

It turns out that to truly make use of the extra headers, I need to be able to inspect the return headers. Should I therefore create a new GetBlobRange that will accept extra headers & return the headers, as well as the body?

I need it for example to check the hash of the returned data.

ahmetalpbalkan:
@dovreshef ok, this change is fairly trivial and we've accepted very similar ones before. Thanks.

PR #319, opened 2016-05-03
List blob metadata

rchippada:
This PR adds Metadata to Blob objects in BlobListResponse.


ahmetalpbalkan:
@rchippada thanks for your patch. This is amazing, reviewing now.

ahmetalpbalkan:
LGTM.

PR #323, opened 2016-05-06
Updating to correct scope in NewServicePrincipalTokenFromCredentials

vishrutshah:
azure.AzureResourceManagerScope is not recognized as it looks like it has evolved to azure.PublicCloud.ResourceManagerEndpoint


ahmetalpbalkan:
LGTM. Thanks.

PR #325, opened 2016-05-07
Expose the public IP address of virtual machines

kklin:
The public IP name and timeout were provided, but there was no way to
get the actual address.


ahmetalpbalkan:
thank you for the pull request @kklin.

Issue #335, opened 2016-05-27
storage.BlobProperties should include Lease status of the Blob


sangeethkumarp:
Particularly in case of a Page Blob, this property is useful in determining if the VHD is being used by a VM or not.

PR #336, opened 2016-05-28
Adding LeaseStatus to storage.BlobProperties

sangeethkumarp:
Particularly in case of a Page Blob, this property is useful in determining if the VHD is being used by a VM or not.

Tracking this enhancement as issue - #335

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -119,6 +119,7 @@ type BlobProperties struct {
 	CopyProgress          string   `xml:"CopyProgress"`
 	CopyCompletionTime    string   `xml:"CopyCompletionTime"`
 	CopyStatusDescription string   `xml:"CopyStatusDescription"`
+	LeaseStatus           string   `xml:"x-ms-lease-status"`
 }
 
 // BlobListResponse contains the response fields from ListBlobs call.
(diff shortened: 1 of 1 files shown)


storage/blob.go
```diff
@@ -119,6 +119,7 @@ type BlobProperties struct {
 	CopyProgress          string   `xml:"CopyProgress"`
 	CopyCompletionTime    string   `xml:"CopyCompletionTime"`
 	CopyStatusDescription string   `xml:"CopyStatusDescription"`
+	LeaseStatus           string   `xml:"x-ms-lease-status"`
```
ahmetalpbalkan: Is this element really in the xml? If not, please omit the field tag.

sangeethkumarp: Please refer to *Response Headers* section in [Get Blob Properties](https://msdn.microsoft.com/en-us/library/azure/dd179394.aspx) for the field : **x-ms-lease-status**.

I'm currently using this property in my app and its working as expected.

colemickens: I think the point is that that field comes from a header, and is not actually part of the XML response, and thus the `xml:` tag is misleading/unnecessary.

sangeethkumarp: Sorry, I misunderstood the question. That's a good catch!

Response headers of GetBlobProperties contains the Lease Status under the header: **x-ms-lease-status**

But, the Xml of ListBlob contains the Lease Status under the tag: **LeaseStatus** as seen at [ListBlobs documentation](https://msdn.microsoft.com/en-us/library/azure/dd135734.aspx). I'm modifying the tag name accordingly.

sangeethkumarp:
I've only tested the GetBlobProperties() in the last commit and as @ahmetalpbalkan pointed out, ListBlobs() havd the LeaseStatus property as empty string, since I was using the wrong Xml tag.

After committing the correct tag name, both GetBlobProperties() and ListBlobs() are resulting with LeaseStatus property as expected.

PR #347, opened 2016-06-27
Update go-autorest version v7.0.5

garimakhulbe:


ahmetalpbalkan:
LGTM

.travis.yml
```diff
@@ -7,6 +7,7 @@ go: 1.6
 install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
+  - go get -u golang.org/x/net/context
   - export GO15VENDOREXPERIMENT=1
```
garimakhulbe: @ahmetalpbalkan Also added this.

ahmetalpbalkan: @garimakhulbe hmm what purpose does it serve? Do any of the libs have any dependencies on this?

garimakhulbe: We are using crypto which has dependency in this and build was failing due to this.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 -----
 
+## `v2.2.0-beta`
+
+- Uses go-autorest version v7.0.5.
+- Update packages jwt-go and crypto in go-autorest and SDK.
```
ahmetalpbalkan: this is a bit cryptic. You should probably say `Updated the versions of packages "a" and "b" in glide.lock.`

.travis.yml
ahmetalpbalkan: @garimakhulbe that's why we probably shouldn't have updated the `x/crypto` in glide.lock in the first place.

garimakhulbe: @ahmetalpbalkan Then we need to manually update the commit number in glide.lock and revert go-autorest lock file. Shall I do that?

garimakhulbe:
@ahmetalpbalkan As per our discussion, I am merging it for now and will remove this new dependency once go 1.7 is out.

PR #349, opened 2016-06-28
storage: friendlier error for malformed key

### LINKED ISSUE #344: storage: user-friendly error message for wrong storage keys
If the storage key is malformed b64 string we return: `illegal base64 data at input byte 8`. it can be more human friendly.

ahmetalpbalkan:
This returns a better error message for malformed base64 storage account keys.

Fixes #344.


boumenot:
LGTM.

PR #355, opened 2016-07-08
Storage emulator (passes go test ./storage)

mcardosos:


storage/client.go
```diff
@@ -114,6 +121,9 @@ func (e UnexpectedStatusCodeError) Got() int {
 // NewBasicClient constructs a Client with given storage service name and
 // key.
 func NewBasicClient(accountName, accountKey string) (Client, error) {
+	if accountName == storageEmulatorAccountName {
+		return NewClient(storageEmulatorAccountName, storageEmulatorAccountKey, DefaultBaseURL, DefaultAPIVersion, !defaultUseHTTPS)
```
ahmetalpbalkan: instead of `!defaultUseHTTPS` might be better to just say `false`

ahmetalpbalkan:
@mcardosos This is amazing thank you. If we ever get Windows Containers on Travis-CI or AppVeyor some day, we can install Azure SDK to get the Storage Emulator and just run our tests on Windows without a need for a storage account! That'd be very amazing.

ahmetalpbalkan:
@mcardosos I see that you work for Azure. You do not need to sign the cla. Just log in at http://aka.ms/azuregithub with your Github & Org accounts for the next time you send PRs to /Azure repos.

ahmetalpbalkan:
@mcardosos why did you choose to export the dev credentials as consts? Do other sdks do that as well?

ahmetalpbalkan:
Also maybe could you add some test cases on client_tests.go to see if the client is getting constructed successfully (and other code paths you added take effect on the client)?

mcardosos:
@ahmetalpbalkan I exported them so users can use Storage Emulator without having to look for / copy and paste the account and key form here https://azure.microsoft.com/en-us/documentation/articles/storage-use-emulator/#authenticating-requests-against-the-storage-emulator
I don't know if all SDKs do this. PHP and Node have them as consts, but it looks like those are used just for tests.

storage/file.go
```diff
@@ -86,6 +89,18 @@ func (f FileServiceClient) DeleteShareIfExists(name string) (bool, error) {
 // deleteShare makes the call to Delete Share operation endpoint and returns
 // the response
 func (f FileServiceClient) deleteShare(name string) (*storageResponse, error) {
+	if err := f.checkForStorageEmulator(); err != nil {
```
ahmetalpbalkan: I'm a bit against adding this check here but I couldn't find a better place either. It appears like `GetFileService()` does not return an error.

ahmetalpbalkan:
@mcardosos sorry for the latency on this pull request. it slipped off my mind, I am merging now.

storage/file.go
mcardosos: Yeah, I know it is not the most elegant thing :(
Still, if a user doesn't know files aren't supported, the error while doing the file operations would be pretty confusing.

PR #361, opened 2016-07-22
Add support for blobs with spaces (and other special URL characters) in the name

bcc:
I've updated every test that operates on a blob to check for the existence of a space, and other special URL characters. I've excluded the + (plus) character from this, as it appears to need additional special handling (as it stands in for 'space' in some parts of the URL component, but not others).  The others work as expected.

### DIFF
--- a/storage/blob_test.go
+++ b/storage/blob_test.go
@@ -74,7 +74,7 @@ func (s *StorageBlobSuite) TestGetBlobSASURI(c *chk.C) {
 func (s *StorageBlobSuite) TestBlobSASURICorrectness(c *chk.C) {
 	cli := getBlobClient(c)
 	cnt := randContainer()
-	blob := randName(5)
+	blob := randNameWithSpecialChars(5)
 	body := []byte(randString(100))
 	expiry := time.Now().UTC().Add(time.Hour)
 	permissions := "r"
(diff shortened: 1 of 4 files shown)


storage/blob_test.go
```diff
@@ -1022,3 +1022,8 @@ func randBytes(n int) []byte {
 	}
 	return data
 }
+
+func randName(n int) string {
+	name := randString(n) + "/" + randString(n) + "-._~:?#[]@!$&'()*,;= " + randString(n)
```
ahmetalpbalkan: we probably don't want to have `"-._~:?#[]@!$&'()*,;= "` in all tests. That'll make debugging very hard. Maybe have a method like `randNameWithSpecialChars` and invoke this from a new test method called `TestPutBlobWithSpecialName`?

storage/client_test.go
```diff
@@ -107,6 +107,7 @@ func (s *StorageClientSuite) Test_buildCanonicalizedResource(c *chk.C) {
 		{"https://foo.blob.core.windows.net/path?a=b&c=d", "/foo/path\na:b\nc:d"},
 		{"https://foo.blob.core.windows.net/?comp=list", "/foo/\ncomp:list"},
 		{"https://foo.blob.core.windows.net/cnt/blob", "/foo/cnt/blob"},
+		{"https://foo.blob.core.windows.net/cnt/bl ob", "/foo/cnt/bl%20ob"},
```
ahmetalpbalkan: perhaps add a few more cases with other special characters (in container name, as well as blob name)?

ahmetalpbalkan:
@bcc thanks for the pull request. This looks good in general except I am a bit against having special char blobs spread over all test cases. Please do not forget to sign the CLA (see the comment above).

ahmetalpbalkan:
@bcc pinging to see if you are still interested in carrying this. I have left a couple of comments and we need CLA to be signed to merge this.

bcc:
@ahmetalpbalkan Yep, had some fun and games getting workplace to sign off on the CLA. Will hopefully have this resolved early next week. Comments are fine, I've not had time to make the changes yet, but seems reasonable.

bcc:
@ahmetalpbalkan Support for + in blob names also now supported. I've made the changes you suggested, but I've left the special chars in TestBlobSASURICorrectness as I think that sense for that particular test.

ahmetalpbalkan:
@bcc thanks for addressing these issues. I guess I should write a separate test to validate the special characters. `randNameWithSpecialChars` might not return special characters all the time, so that will introduce some non-determinism in validating that code path you added. 

Would you like to just add a test for uploading a constant blob name with a special characters in it? That might help address the original issue (blob upload with special chars) better. If not, I can just go ahead and merge and take the fix myself.

bcc:
@ahmetalpbalkan `randNameWithSpecialChars` will always return the same set of special characters, only the prefix and postfix alphabet characters are random. I had the same concerns regarding the repeatability of that test.

ahmetalpbalkan:
@bcc ah okay then! I missed that point.

ahmetalpbalkan:
Thanks for the contribution @bcc!

PR #374, opened 2016-08-05
provide missing SetBlobProperties function and add missing properties to GetBlobProperties

schreibe72:
Add following properties to GetBlobProperties:

* Cache-Control
* Cache-Language

Add function SetBlobProperties to set ContentSettings like:

* Content-MD5
* Content-Language
* Content-Encoding
* Content-Type
* Cache-Control

Add Test for SetBlobProperties.

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -600,6 +601,37 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
 		BlobType:              BlobType(resp.headers.Get("x-ms-blob-type")),
 		LeaseStatus:           resp.headers.Get("x-ms-lease-status"),
 	}, nil
+}
+
+type ContentSetting struct {
+	ContentMD5      string `header:"x-ms-blob-content-md5"`
+	ContentLanguage string `header:"x-ms-blob-content-language"`
+	ContentEncoding string `header:"x-ms-blob-content-encoding"`
+	ContentType     string `header:"x-ms-blob-content-type"`
+	CacheControl    string `header:"x-ms-blob-cache-control"`
+}
+
+func (b BlobStorageClient) SetBlobProperties(container, name string, contentSetting ContentSetting) error {
+	params := url.Values{"comp": {"properties"}}
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), params)
+	headers := b.client.getStandardHeaders()
+
+	value := reflect.ValueOf(contentSetting)
+	for i := 0; i < value.NumField(); i++ {
+		k := value.Type().Field(i).Tag.Get("header")
+		v := value.Field(i).String()
+		if v != "" {
+			headers[k] = v
+		}
+	}
+
+	resp, err := b.client.exec("PUT", uri, headers, nil)
+	if err != nil {
+		return err
+	}
+	defer resp.body.Close()
+
+	return checkRespCode(resp.statusCode, []int{http.StatusOK})
 }
 
 // SetBlobMetadata replaces the metadata for the specified blob.
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -112,6 +112,8 @@ type BlobProperties struct {
 	ContentLength         int64    `xml:"Content-Length"`
 	ContentType           string   `xml:"Content-Type"`
 	ContentEncoding       string   `xml:"Content-Encoding"`
+	CacheControl          string   `xml:"Cache-Control"`
+	ContentLanguage       string   `xml:"Cache-Language"`
 	BlobType              BlobType `xml:"x-ms-blob-blob-type"`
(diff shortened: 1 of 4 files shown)


ahmetalpbalkan:
@schreibe72 thank you for your pull request, please ping me once you signed the CLA and I'll review.

schreibe72:
@ahmetalpbalkan I have signed the CLA and send via DocuSign.

ahmetalpbalkan:
@schreibe72 I guess there is a processing delay in that case, we can come back to that later. I'll leave some comments now.

storage/blob.go
```diff
@@ -122,6 +125,16 @@ type BlobProperties struct {
 	LeaseStatus           string   `xml:"LeaseStatus"`
 }
 
+// ContentSetting contains various properties of a blob and is an entry
+// in SetBlobProperties
+type ContentSetting struct {
```
ahmetalpbalkan: Maybe call this BlobHeaders? (more about this below)

storage/blob.go
```diff
@@ -602,6 +617,37 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
...
+// applications either.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/ee691966.aspx
+func (b BlobStorageClient) SetBlobProperties(container, name string, contentSetting ContentSetting) error {
+	params := url.Values{"comp": {"properties"}}
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), params)
+	headers := b.client.getStandardHeaders()
+
+	value := reflect.ValueOf(contentSetting)
+	for i := 0; i < value.NumField(); i++ {
```
ahmetalpbalkan: I think this logic is very valuable to be buried in `SetBlobProperties`, why don't you create a method called `headersFromStruct` that uses a field tag called "header"; (not "xml") and test it so that we can use it here and so on?

storage/blob_test.go
```diff
@@ -647,6 +647,37 @@ func (s *StorageBlobSuite) TestSetMetadataWithExtraHeaders(c *chk.C) {
 	c.Assert(err, chk.IsNil)
 }
 
+func (s *StorageBlobSuite) TestSetProperties(c *chk.C) {
```
ahmetalpbalkan: TestSetBlobProperties*

storage/blob.go
```diff
@@ -122,6 +125,16 @@ type BlobProperties struct {
...
 }
 
+// ContentSetting contains various properties of a blob and is an entry
+// in SetBlobProperties
+type ContentSetting struct {
+	ContentMD5      string `xml:"x-ms-blob-content-md5"`
+	ContentLanguage string `xml:"x-ms-blob-content-language"`
+	ContentEncoding string `xml:"x-ms-blob-content-encoding"`
+	ContentType     string `xml:"x-ms-blob-content-type"`
+	CacheControl    string `xml:"x-ms-blob-cache-control"`
```
ahmetalpbalkan: I think the field tag is never used to actually serialize into xml, so we should probably use a tag called "header" for this.

ahmetalpbalkan:
It looks like Set Blob Properties and Get Blob Properties (+List Blobs) work a bit weird as in "Set" accepts x-ms-* headers and "Get/List" return XML fields named properly.

Therefore it appears like using the same struct both on response deserialization path from 'Get/List' and as headers in 'Set' is going to be a bit challenging.

ahmetalpbalkan:
@schreibe72 just got this crazy idea. What if we annotate this `ContentSetting` struct (or better: `BlobHeaders` or  struct) with BOTH `"xml"` and `"header"` tags?

This way:

0. we can embed this struct into `type BlobPropertiesResponse` and thanks to `xml` tag, it will deserialize just fine
0. For SetBlobProperties, `func headersFromStruct(v interface{}) (map[..], error)` will do the job and read from `"header"` field tag.
0. For GetBlobProperties we write a new method called `func structFromHeaders(map[string]string, v interface{}) error` which looks at the `"header"` field tag and deserializes the headers in the response of [Get Blob Properties](https://msdn.microsoft.com/en-us/library/azure/dd179394.aspx) into our struct and just return it! This way we don't have to populate its fields manually?

storage/blob.go
```diff
@@ -590,6 +602,8 @@ func (b BlobStorageClient) GetBlobProperties(container, name string) (*BlobPrope
 		ContentLength:         contentLength,
 		ContentEncoding:       resp.headers.Get("Content-Encoding"),
 		ContentType:           resp.headers.Get("Content-Type"),
+		CacheControl:          resp.headers.Get("Cache-Control"),
```
ahmetalpbalkan: @schreibe72 thanks for making the changes. I think we should somehow embed BlobHeaders struct into BlobProperties struct and have some deserializer that does map-->struct. I feel like it might be a lot of work whereas this patch is probably solving your problem. What do you think?

schreibe72:
i would split this in to two pull requests. One is for the new function setBlobProperties, the other for refactoring getBlobProperties. 
The deserializer is a bit more tricky, because there are not only string types. The other thing is, that if we change the struct BlobProperties some depending code might break.

ahmetalpbalkan:
@schreibe72 fair point. Thanks.

PR #375, opened 2016-08-09
go1.7: use TLS renegotiation, rm stdlib forks

### LINKED ISSUE #90: Eliminate stdlib forks living at core/
`core/` currently has `net/http` and `crypto/tls` vendored (with some changes we don't entirely know). We must watch out if we can make feature requests to golang about these and eventually get rid of those (nobody will vendor our entire repo like this) :smile:

ahmetalpbalkan:
The time for removing the weird `./core/` packages has come at last. Go 1.7 introduces TLS renegotiation: https://github.com/golang/go/commit/af125a5193c75dd59307fcf1b26d885010ce8bfd

The actual change is at `management/http.go` (hidden on GitHub UI due to removing so many files, you have to click on it manually, it's all the way at the end.)

Removing our net/http and crypto/tls forks and just using Renegotation field
in the crypto/tls.Config field of the net/http.Client of the ./management/**
packages. (Fixes #90) 

⚠️ ⚠️ ⚠️  Should be merged once go1.7 stable is out and preferably like 1 month later.
If the customers of the ./management/ package are not vendoring the SDK and using
go versions prior to 1.7, they will get a compilation error. CI is expected to fail as it also uses go1.6.

cc: @paulmey


paulmey:
LGTM, are you going to update the travis file in the same PR?

ahmetalpbalkan:
@paulmey probably should update that separately when it's out (planned Aug 15). then rebase/merge this.

paulmey:
@ahmetalpbalkan makes sense, I hope travis has the new go version available as soon as its released...

ahmetalpbalkan:
@paulmey tests pass now.

paulmey:
awesome sauce. ship it!

ahmetalpbalkan:
@paulmey hehee https://golang.org/doc/go1.7

image

masterzen:
FYI, I understand that it is the right way going forward, but this change is breaking my WinRM library (see masterzen/winrm#60), and prevents users to use Go < 1.7.
As winrm is itself a 'library', it's not really possible to vendor its dependencies.

colemickens:
You may not be able to (or want to) vendor but you can still lock to a version, allowing you and your users to get a correct set of dependencies.

We do this in `azure-sdk-for-go`, for example. If you have a project that uses `glide` and you consume `azure-sdk-for-go`, you're ensured to get a compatible version of `go-autorest`.

PR #376, opened 2016-08-09
Make storage client work correctly with read only secondary accounts

dovreshef:
We need to remove the "-secondary" string from the account name when using it to authorize access.

See also here: https://blogs.technet.microsoft.com/ksasaki/2013/12/23/use-ra-grs/.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -235,7 +235,7 @@ func (c Client) GetFileService() FileServiceClient {
 
 func (c Client) createAuthorizationHeader(canonicalizedString string) string {
 	signature := c.computeHmac256(canonicalizedString)
-	return fmt.Sprintf("%s %s:%s", "SharedKey", c.accountName, signature)
+	return fmt.Sprintf("%s %s:%s", "SharedKey", c.getCanonicalizedAccountName(), signature)
 }
 
 func (c Client) getAuthorizationHeader(verb, url string, headers map[string]string) (string, error) {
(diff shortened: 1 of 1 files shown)


storage/client.go
```diff
@@ -255,6 +255,12 @@ func (c Client) getStandardHeaders() map[string]string {
 	}
 }
 
+func (c Client) getCanonicalizedAccountName() string {
+	// since we may be trying to access a secondary storage account, we need to
+	// remove the -secondary part of the storage name
+	return strings.Replace(c.accountName, "-secondary", "", -1)
```
ahmetalpbalkan: could this be `strings.TrimSuffix`?

ahmetalpbalkan:
> See also here: https://blogs.technet.microsoft.com/ksasaki/2013/12/23/use-ra-grs/.

@dovreshef do you know if there is an English version of this article?

dovreshef:
@ahmetalpbalkan that's the only mention of the issue that I found. Had to use Google translate :).

ahmetalpbalkan:
@dovreshef  I'm trying to understand why would you specify storage account name as the `-secondary` endpoint in your code. Why do you need to hit the secondary endpoint instead of the primary one?

dovreshef:
@ahmetalpbalkan We use it to simulate a regional downtime, read all the backups off the storage.

ahmetalpbalkan:
Ah okay, in that case, merging.

dovreshef:
@ahmetalpbalkan thanks!!

PR #381, opened 2016-08-11
vmutils: add ConfigureWithCustomData* helpers

rjeczalik:
Hey Azure-go team!

I happen to need to set custom data for vm creation request, but couldn't find a suitable way to do this. Looks to me like there's currently no way, so I've added helpers to vmutils. My aim is to extend azure_instance resource in Azure provider for Terraform with custom_data and in order to do that, I'd need azure-go library support first.

Please take a look.

### DIFF
--- a/management/vmutils/vmutils.go
+++ b/management/vmutils/vmutils.go
@@ -22,7 +22,7 @@ func NewVMConfiguration(name string, roleSize string) vm.Role {
 	}
 }
 
-// ConfigureForLinux adds configuration for when deploying a generalized Linux
+// ConfigureForLinux adds configuration when deploying a generalized Linux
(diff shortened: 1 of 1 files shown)


rjeczalik:
From travis:

```
$ go get -u github.com/golang/lint/golint
package golang.org/x/tools/go/gcimporter15: unrecognized import path "golang.org/x/tools/go/gcimporter15" (parse https://golang.org/x/tools/go/gcimporter15?go-get=1: no go-import meta tags)
```

The failure is due to godoc.org being down. Will restart build after it's up.

ahmetalpbalkan:
LGTM @paulmey PTAL & merge?

paulmey:
LGTM. Thanks @rjeczalik!

PR #382, opened 2016-08-17
Checking err for nil after execTable

stascode:
extractContinuationTokenFromHeaders(resp.headers) panics if execTable fails, e.g. due to lack of ca-certificates package installed


garimakhulbe:
LGTM. Thanks!

PR #383, opened 2016-08-24
Generate the keyvault management client with validation

cosmincojocar:


garimakhulbe:
We use Azure/AutoRest tool to generate Go SDK. Do you generate this code using AutoRest tool or add it yourself?

ahmetalpbalkan:
@cosmincojocar we're fixing this in PR #385. We don't manually edit the generated code, so we'll just regenerate. Thanks for reporting.

cosmincojocar:
It is generated with the latest version of AutoRest tool which got recently the Go support. Only the go fmt and the version increase was done manually.  

I am planning to implement manually the secrets and the key API part, since there isn't any swagger definition for that. What do you think?

garimakhulbe:
API spec team is updating the key vault swagger and will release it in few days. Internal preview of spec is [here](https://github.com/Azure/azure-rest-api-specs-pr/blob/master/keyvault/2015-06-01/swagger/keyvault.json). I think we should wait till this release.

garimakhulbe:
@cosmincojocar PR #385 is merge and validation code is now in release v4.0.0-beta.

PR #385, opened 2016-08-24
Add validation code in ARM packages

garimakhulbe:
Added validation code in all arm packages.
Logic api changed.


colemickens:
If anyone wants to look at this without their browser crashing... https://reviewable.io/reviews/Azure/azure-sdk-for-go/385

ahmetalpbalkan:
@colemickens lol I was doing

    curl -sL https://github.com/Azure/azure-sdk-for-go/pull/385.patch | less

ahmetalpbalkan:
@garimakhulbe  I have several questions about validation changes:

1. I see you use double-quote strings aroung regexps. If your regexps contain backslashes etc it's more idiomatic to use backtick-quoted strings to avoid `\\`s, which make it harder to read regexps.

2. I see that a lot of String fields are validated as `validation.Constraint{"Name", validation.Null, true, nil}` in the generated code. However I can't see that [`validateString` code](https://github.com/Azure/go-autorest/blob/f0b1c4ee355163629bf0d9d52b7bef1d238f3426/autorest/validation/validation.go#L185-L231) handles the `validation.Null` case? Are you sure this works correctly? Maybe it's a c# vs Go thing that people assumed strings are nullable in C# and they wrote specs accordingly?

3. I noticed some type signature changes in the diff. Make sure you're using a correct version number to indicate as such.

ahmetalpbalkan:
also I noticed bunch of `"Validation error "`s (pay attn to the space at the end). Can we make this error little bit more meaningful for the user? e.g maybe by saying `"Invalid input value"` instead?

@garimakhulbe can you please paste a full error string of a validation error, for say, invalid storage account name so that we can design the error a bit better? I can't fully infer how the error will look like from the code.

garimakhulbe:
@ahmetalpbalkan 
1. About double backslash ```\\s``` - I will update this. (good change)

```{"accountName", validation.Pattern, `^[-\w\._]+$`, nil}}}});```

2.About null check for strings - these are string pointers therefore validating for null. For example, string attribute ```Name``` below..
```
checkNameAvailabilityInput,
              []validation.Constraint{{"Name", validation.Null, true, nil},
```
defined as..
```
type CheckNameAvailabilityInput struct {
	Name *string `json:"name,omitempty"`
	....
}
```
3.signature changes in the diff - will keep this mind.

4.Full error string of a validation error - current error string is not well formatted (mention ```validation error``` twice).
image
```
batch.AccountClient#Create: Validation error : StatusCode=0 -- Original Error: autorest/validation: validation error: Constraint=Pattern Parameter=resourceGroupName Value="dhf$kjsdf" Details=string 'dhf$kjsdf' doesn't match pattern ^[-\w\._]+$

```


I will remove the extra text and space after ```Validation Error ```. How is it now?
image
```
batch.AccountClient#Create: Validation error: StatusCode=0 -- Original Error: autorest/validation:  Constraint=Pattern Parameter=resourceGroupName Value="dhf$kjsdf" Details=string 'dhf$kjsdf' doesn't match pattern ^[-\w\._]+$

```

**Comment by ahmetalpbalkan** [comment]
I'm satisfied with the answers, thanks!

About the error we return:
```
batch.AccountClient#Create: Validation error : StatusCode=0 -- Original Error:
autorest/validation: validation error: Constraint=Pattern Parameter=resourceGroupName Value="dhf$kjsdf" Details=string 'dhf$kjsdf' doesn't match pattern ^[-\w\._]+$
```

Couple of things to note (you can address them in separate PRs over time, but sooner is better than later):

- `Original Error:` seems unnecessary in this context, we already keep wrapping errors about
- Clearly `StatusCode=0` is not very useful in context, I guess we could've architected errors in go-autorest a bit better. Perhaps omit this string if `StatusCode=0`?
- If you put `Parameter=` before `Constraint=`, it would seem a bit better to me.
- Pay attention how twice you had `Validation error :` and `validation error: ` :) This is because you over-wrap. But if you change the first `Validation error: ` to `Invalid input:`, the error will seem clearer. “Input” is what user gives you and what they know; "validation" is a task you do on the input –which is not clear to the user from the method signature or anything. So saying straight out "validation failed" is sort of a bad UX.
- I realized you have single quotes around the string (`'%s'`) please use `%q` instead, which turns a string into a Go string literal with proper escaping and such.
- You are repeating the value `dhf$kjsdf` twice in the error message, if you're sure you always return `Value=` from the validation package, then the innermost error shouldn't have the `'%s'` perhaps?

**Comment by ahmetalpbalkan** [comment]
made a few edits to my comment above ^, make sure you read it on github :)

**Comment by garimakhulbe** [comment]
This PR only includes addition of validation code in arm packages. It does not include name conflict  related code changes(e.g. `Version() and Version`). 

Bumped up version for breaking changes in package `logic`.

**Comment by garimakhulbe** [comment]
validation code from arm/batch/account.go

```Go
if err := validation.Validate([]validation.Validation{
		{resourceGroupName,
			[]validation.Constraint{{"resourceGroupName", validation.Pattern, `^[-\w\._]+$`, nil}}},
		{accountName,
			[]validation.Constraint{{"accountName", validation.MaxLength, 24, nil},
				{"accountName", validation.MinLength, 3, nil},
				{"accountName", validation.Pattern, `^[-\w\._]+$`, nil}}},
		{parameters,
			[]validation.Constraint{{"parameters.Properties", validation.Null, false,
				[]validation.Constraint{{"parameters.Properties.AutoStorage", validation.Null, false,
					[]validation.Constraint{{"parameters.Properties.AutoStorage.StorageAccountID", validation.Null, true, nil}}},
				}}}}}); err != nil {
		return result, validation.NewErrorWithValidationError(err, "batch.AccountClient", "Create")
	}
```

sample errors..

`batch.AccountClient#Create: Invalid input: autorest/validation: validation failed: parameter=parameters.Properties.AutoStorage.StorageAccountID constraint=Null value=(*string)(nil) details: value can not be null; required parameter`

`batch.AccountClient#Create: Invalid input: autorest/validation: validation failed: parameter=resourceGroupName constraint=Pattern value="dhfkj$sdf" details: value doesn't match pattern ^[-\w\._]+$`

`batch.AccountClient#Create: Invalid input: autorest/validation: validation failed: parameter=accountName constraint=MinLength value="xx" details: value length must be greater than 3`

ahmetalpbalkan:
@garimakhulbe LGTM

garimakhulbe:
Thanks!

PR #386, opened 2016-08-24
Add code analysis via GAS for SDL scanning

garimakhulbe:


garimakhulbe:
cc @mbearup

.travis.yml
```diff
@@ -8,10 +8,14 @@ install:
...
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  - gas -fmt=json -out=results.json ./... > /dev/null
+  - cat ./results.json
+  - test $(grep -c "\"severity\":\s*\"HIGH\"" ./results.json) -lt 8
```
ahmetalpbalkan: What's this 8? Can u pls provide some sample output and make code reviewers lives easier perhaps? ;-)

ahmetalpbalkan:
What's this -lt 8? Can u pls provide some sample output and make code reviewers lives easier perhaps? ;-)  will the PR author see the output of gas if it goes above 8?

garimakhulbe:
```test $(grep -c "\"severity\":\s*\"HIGH\"" ./results.json) -lt 8```
This command counts occurrence of ```"severity": "HIGH",``` in GAS report.json. ```-lt 8``` means count of high severity issues should be less than 8.  Added cat for report.

File: .travis.yml [hunk 2]
```diff
@@ -8,10 +8,22 @@ install:
...
   - glide install
 
 script:
+  # GAS command for SDL scanning
+  - (cd arm && gas -fmt=json -out=results.json ./... > /dev/null)
+  - cat arm/results.json
+  - test $(grep -c "\"severity\":\s*\"HIGH\"" arm/results.json) -lt 8
+  - (cd management && gas -fmt=json -out=results.json ./... > /dev/null)
+  - cat management/results.json
+  - test $(grep -c "\"severity\":\s*\"HIGH\"" management/results.json) -lt 8
```
mbearup: When I ran the scan there was only one finding in the management folder and none in the other folders. So you could make this line (management) "-lt 2" and the others "-lt 1"
To answer Ahmet's question, this is comparing the total number of HIGH findings to the current value to make sure it doesn't increase. We have more elegant tools for parsing incident severity/count internally but those aren't public at this time.

.travis.yml
```diff
@@ -8,10 +8,22 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  # GAS commands for SDL scanning
+  - (cd arm && gas -fmt=json -out=results.json ./... > /dev/null)
```
ahmetalpbalkan: How about just  just the following (assuming `gas` is capable of printing stuff to stdout just fine):

    gas -fmt=json ./arm | tee results.json
    test $(grep -c "\"severity\":\s*\"HIGH\"" results.json) -lt 1

you can keep overwriting the results.json, it is fine because if it fails once, it will not proceed. and if it succeeds you don't need the prior files.

But I'm generally -1 for listing each pkg separately and saying `-lt 1` , `-lt 2`. I don't currently know why they are 1 or 2. Why don't we just fix them and make it zero?

.travis.yml
garimakhulbe: @mbearup I have updated the thresholds. ```Test123``` in finding is dummy password so we can ignore that.
@ahmetalpbalkan take a look.

ahmetalpbalkan:
I'm realizing you are actually parsing JSON using regex. Please don't. Use `jq`instead.

ahmetalpbalkan:
@garimakhulbe I just downloaded gas and I believe you can just have these two lines to scan the entire repo and just skip that one example folder:

    gas -skip=examples_test.go ./... | tee gas-scan.txt
    test $(grep -c 'Severity: HIGH' ./gas-scan.txt) -lt 1; echo $?

.travis.yml
mbearup: In some cases folks have dozens of findings and fixing them is time consuming and risky, so as a rule I'm not pushing everyone to solve things today. In this case as I said there's just one finding, so if we want to fix it or mark it as a false positive that's fine. The reason this is broken out per-folder is because the build process pulls in 3rd party code under the 'vendor' folder, and this code has additional findings that we're not triaging at this time.
Regarding pipe to tee, yes you can do this. Gas prints out some additional garbage when scanning some source files (notably packer-azure) but it doesn't do this for the go-sdk, and nothing in the current travis config is filtering this out anyway.

garimakhulbe:
@ahmetalpbalkan We need to modify this a bit. Need to skip vendor too, that's why split this in sub-folders.
```
gas -skip=examples_test.go skip=vendor/* ./... | tee gas-scan.txt
test $(grep -c 'Severity: HIGH' ./gas-scan.txt) -lt 1; echo $?
```
Lets try this. Thanks.

garimakhulbe:
For one issue, we can make password blank in examples_test.go.

.travis.yml
ahmetalpbalkan: > The reason this is broken out per-folder is because the build process pulls in 3rd party code under the 'vendor' folder, and this code has additional findings that we're not triaging at this time.

we can easily fix this by using `gas $(go list ./... | grep -v 'vendor')`.

ahmetalpbalkan:
@garimakhulbe sounds good. I removed `echo $?` (it was my debug thingy, don't include it).

ahmetalpbalkan:
> For one issue, we can make password blank in examples_test.go.

@garimakhulbe actually, can you move `examples_test.go` into `examples` folder perhaps (and rename to `example.go`)? We can skip the folder altogether.

.travis.yml
```diff
@@ -8,10 +8,14 @@ install:
...
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  - gas -skip=*examples* -skip=*vendor* ./... | tee gas-scan.txt
+  - cat gas-scan.txt
+  - test $(grep -c 'Severity:\s*HIGH' gas-scan.txt) -eq 0
```
ahmetalpbalkan: you can now use `test -z $(grep` without `-c`.

.travis.yml
```diff
@@ -8,10 +8,14 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  - gas -skip=*examples* -skip=*vendor* ./... | tee gas-scan.txt
```
ahmetalpbalkan: for some reason it's still not skipping. maybe try `-skip=management/examples/*.go`?

.travis.yml
```diff
@@ -8,15 +8,20 @@ install:
...
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  - gas -skip=management/examples/*.go -skip=*vendor* ./... | tee gas-scan.txt
+  - cat gas-scan.txt
+  - test -z "$((grep 'Severity:\s*HIGH' gas-scan.txt) -eq 0)"
```
ahmetalpbalkan: you don't need `-eq 0` anymore. can you please run these commands locally like `test ...; echo $?` to see what they behave to try out :) I'm getting an email every single time you push a commit.

ahmetalpbalkan: also extra parens not required. `-z` is testing empty string. that's why `-eq 0` is not needed anymore as grep output is supposed to be empty.

garimakhulbe: Last commit. Its over now. :)

.travis.yml
```diff
@@ -8,10 +8,14 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/Masterminds/glide
   - go get -u golang.org/x/net/context
+  - go get -u github.com/HewlettPackard/gas
   - export GO15VENDOREXPERIMENT=1
   - glide install
 
 script:
+  - gas -skip=management/examples/*.go -skip=*vendor* ./... | tee gas-scan.txt
+  - cat gas-scan.txt
```
ahmetalpbalkan: you don't need this `cat`. https://en.wikipedia.org/wiki/Tee_(command)

management/examples/example.go
```diff
@@ -16,7 +17,7 @@ func Example() {
 	vmSize := "Small"
 	vmImage := "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-14_04-LTS-amd64-server-20140724-en-us-30GB"
 	userName := "testuser"
-	userPassword := "Test123"
+	userPassword := ""
```
ahmetalpbalkan: you don't need to remove the password.

garimakhulbe: Don't remember when I changed this. Noticed it in final commit. Keeping this as it is for now. :)

garimakhulbe: Merge?

ahmetalpbalkan: You gotta remove `cat`.

ahmetalpbalkan:
LGTM

garimakhulbe:
Thanks!

PR #391, opened 2016-08-30
management/osimage: implement ShareImage

mischief:


ahmetalpbalkan:
@mischief Hmm these are actually private APIs, right? I'm not actually sure we should be sharing these publicly. The `ShowInGui` field in the other PR you submitted looks like also a part of the private APIs.

@paulmey wdyt, should we merge? We OSS'd azure-extensions-cli, which also contained private APIs because publisher subscriptions have to be whitelisted anyway. I think the same applies here.

paulmey:
I don't think it belongs in this SDK if it's not part of the public API set. Maybe start an SDK for image/extension publishing?

ahmetalpbalkan:
Agreed. @mischief I encourage you to develop your own package by extending the `management` package with the methods/types you're contributing here. To change existing types, I recommend copying them to your fork. 

To see this in practice: https://github.com/Azure/azure-extensions-cli/blob/master/sdk.go

mischief:
this decision is somewhat confusing, considering the same API calls were added to https://github.com/Azure/azure-sdk-for-python/pull/497, but as you wish.

ahmetalpbalkan:
@mischief we can always reevaluate.

IMO, adding these to general-purpose Python SDK was not the right move either. My argument is, the cost of maintenance and it will look confusing for users (they will naturally think they can use a method named `PublishImage`). We even kept the methods I linked above outside the SDK.

Perhaps @huguesv who merged to Python can provide input.

PR #392, opened 2016-08-30
management/osimage: fix field ordering and set omitempty for optional fields

mischief:
with the old ordering, the azure api would reject POST requests from
AddOSImage.


mischief:
forgot to mention i've added `ShowInGui` which is not listed as optional in the docs for the api call.

ahmetalpbalkan:
Closing as Private API. https://github.com/Azure/azure-sdk-for-go/pull/391#issuecomment-243882801

PR #394, opened 2016-09-02
Version v5.0.0-beta

garimakhulbe:
1.This PR includes fix for name conflicts in swagger types and variables/methods defined by us in Go SDK(UserAgent(), Version(), etc..). 
For duplicate types, it will add package name in type name, for example `Version` type from swagger conflicts with function `Version()`  therefore update to..
```Go
type VersionServerMangement struct {
	Major         *int32 `json:"major,omitempty"`
	Minor         *int32 `json:"minor,omitempty"`
        .....
}
```
```Go
// PowerShellSessionResourceProperties is
type PowerShellSessionResourceProperties struct {
       ....
	Version              *VersionServerManageMent  `json:"version,omitempty"`
	.....
}
```
For enum duplicates, enum type is being added to name. 

```Go
const (
	....
	// StateNew specifies the state new state for state.
	StateNew State = "New"  

     // Earlier
   New State = "New"  //conflicting with method New()
	...
```

2.Added check to not trim package name if group name is plural of package name.
We trim package name from method group name to avoid lint error in go.

> Storage.StorageAccounts  (golint error for stuttering name)  ---> Storage.Accounts

There is new swagger for `EventHub` and it has method group `EventHubs`. After trimming package name from group `EventHubs`... 
> EventHubs -> s ( again results in lint error for improper name)

To avoid this, I have added a check for plural before trim.

3.Removed Container Service until it is discussion. 

4.New APIs added: Server management, Datalake Analytics - Account and Job, Eventhub.


garimakhulbe:
@vishrutshah @mcardosos Please review this PR.

arm/compute/containerservice/client.go
```diff
@@ -1,44 +0,0 @@
-// Package compute implements the Azure ARM Compute service API version
-// 2016-03-30.
```
mcardosos: I'm not a big fan of removing everything from container service

mcardosos:
> }

Aren't account.go and client.go too similar?

---
Refers to: arm/datalake-store/account/account.go:31 in 2567707. [](commit_id = 2567707433bedd9c9900c86e61c3c649df575774, deletion_comment = False)

arm/resources/subscriptions/models.go
```diff
@@ -24,6 +24,35 @@ import (
...
+	// limit.
+	CurrentPeriodOff SpendingLimit = "CurrentPeriodOff"
+	// Off specifies the off state for spending limit.
+	Off SpendingLimit = "Off"
+	// On specifies the on state for spending limit.
+	On SpendingLimit = "On"
+)
+
+// tate enumerates the values for tate.
+type tate string
```
mcardosos: >tate [](start = 5, length = 4)

state?

garimakhulbe: Found it. Its a new swagger and enum type name is `SubscriptionState` and package name is `Subsciptions`, therefore when we trim package name, it becomes `tate`. It was string type before so we did not do any stuttering type check earlier.

arm/compute/version.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	major = "4"
-	minor = "0"
+	minor = "1"
```
vishrutshah: Shouldn't removal of entire service be considered as breaking change then we must increase major version.

arm/compute/containerservice/client.go
vishrutshah: Do we have any plan to enable this service after removal?

arm/datalake-analytics/job/client.go
```diff
@@ -0,0 +1,583 @@
+// Package job implements the Azure ARM Job service API version
```
vishrutshah: datalake-analytics Job is not an management service. It should not exists in the ARM namespace.

arm/datalake-store/filesystem/version.go
```diff
@@ -24,7 +24,7 @@ import (
```
vishrutshah: datalake-store catalog and filesystem are not an management service. It should not exists in the ARM namespace.

arm/eventhub/eventhubs.go
```diff
@@ -0,0 +1,764 @@
...
+		result.Response = autorest.Response{Response: resp}
+		return result, autorest.NewErrorWithError(err, "eventhub.EventHubsClient", "ListKeys", resp, "Failure sending request")
+	}
+
+	result, err = client.ListKeysResponder(resp)
+	if err != nil {
+		err = autorest.NewErrorWithError(err, "eventhub.EventHubsClient", "ListKeys", resp, "Failure responding to request")
+	}
+
+	return
```
vishrutshah: May be i am confused. Should this return `results` ?

arm/network/client.go
```diff
@@ -1,5 +1,5 @@
 // Package network implements the Azure ARM Network service API version
-// 2016-06-01.
+// 2016-09-01.
 //
```
vishrutshah: Did you know whether there were any breaking changes between 2016-07-01 & 2016-09-01?

CHANGELOG.md
```diff
@@ -1,11 +1,28 @@
 # CHANGELOG
 
 -----
+## `v4.1.0-beta`
+
+| api                           | version             | note     |
+|:------------------------------|:--------------------|:---------|
+| arm/network                   | 2016-09-01          | updated  |
```
vishrutshah: Are there any reasons for updating version of all the services if no changes detected? Just curious about teams initial reasoning.

vishrutshah:
@garimakhulbe I've left some minor comments. Other changes look good. Thanks!

arm/eventhub/eventhubs.go
mcardosos: It returns result :D (that's why the return types have names)

vishrutshah: Cool :)

CHANGELOG.md
garimakhulbe: I think this is because we don't release individual API packages instead sdk as a whole.

arm/network/client.go
garimakhulbe: I can see some new fields added.

> AcceleratedNetworkingEnabled *bool                       `json:"acceleratedNetworkingEnabled,omitempty"`

arm/compute/containerservice/client.go
garimakhulbe: It depends on when core team updates CompositeSwagger modeler. ACS will be added after that.

arm/resources/subscriptions/models.go
garimakhulbe: I revert API version to 2015-.. because I am not sure how we are going to handle this case. It is not a bug, it is the way we handle stuttering types.

We remove package name from type names to avoid stuttering type golint error.
For example, if there is a swagger for API ABC and it has type ABCAccount, then to use this type in go `ABC.ABCAccount` which gives golint error. To avoid this, we trim the package name in typenames. `ABCAccount` -> `Account`.

In this case, `tate`..
package: `subscriptions`
Enum type: `SubscriptionState` -> after trimming `subscriptions`, it becomes `tate`.

arm/compute/version.go
garimakhulbe: Baumped up to 5.0.0.

garimakhulbe:
@mcardosos @vishrutshah 
Please take a look.

CHANGELOG.md
```diff
@@ -1,11 +1,28 @@
...
+| arm/resources/subscriptions   | 2016-06-01          | updated  |
```
mcardosos: Subscription stayed the same
Also, update arm packages list in README

garimakhulbe:
@mcardosos @vishrutshah 
This PR is complete. Only pending is to decide where we are going to put `datalake-analytics` and `datalake-store` in SDK as you commented they are not part of ARM. IMO it should be consistent with other SDKs but let me know how you want to proceed in this.

arm/compute/containerservice/client.go
vishrutshah: @garimakhulbe Just FYI: please let @mcardosos & @jhendrixMSFT know who is the point of contact from core team (who knows about this) tracking this so that they know whom to talk to.

stack72:
Any reason why we are jumping so fast from 4.0 to 5.0? This feels like a lot of changes in a short period of time - I understand that this is all still classed as beta, but there are still some tools that depend on the SDKs

garimakhulbe:
We removed Container Service from SDK until we get that finalized internally. That might a breaking change for some of users so bumped up the version.

garimakhulbe:
@jhendrixMSFT @mcardosos Please take a look now.

jhendrixMSFT:
Is the commit hash from the API specs repo captured anywhere?  I was discussing this with @mcardosos yesterday, it would be great if we could include this in the commit notes.

garimakhulbe:
I will add it in Changelog.md.

mcardosos:
LGTM, just waiting for travis :D

jhendrixMSFT:
Thanks for adding, this will help with testing changes to the Go generator for a given SDK.  Changes look good to me.

PR #395, opened 2016-09-06
Add more file storage share operations

enieuw:


ahmetalpbalkan:
@enieuw can you please write tests in `file_test.go` for these?

storage/file.go
```diff
@@ -55,6 +170,58 @@ func (f FileServiceClient) createShare(name string) (*storageResponse, error) {
 	return f.client.exec("PUT", uri, headers, nil)
 }
 
+// GetShareProperties provides various information about the specified
+// file. See https://msdn.microsoft.com/en-us/library/azure/dn689099.aspx
+func (f FileServiceClient) GetShareProperties(name string) (*ShareProperties, error) {
+	verb := "HEAD"
```
garimakhulbe: @enieuw Can we have this directly in `.exec` instead of defining it here like we did in other places?

```Go
resp, err := f.client.exec("PUT", uri, headers, nil)
return f.client.exec("DELETE", uri, f.client.getStandardHeaders(), nil)
```

storage/file.go
```diff
@@ -55,6 +170,58 @@ func (f FileServiceClient) createShare(name string) (*storageResponse, error) {
...
+// SetShareProperties replaces the ShareHeaders for the specified file.
+//
+// Some keys may be converted to Camel-Case before sending. All keys
+// are returned in lower case by SetShareProperties. HTTP header names
+// are case-insensitive so case munging should not matter to other
+// applications either.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/mt427368.aspx
+func (f FileServiceClient) SetShareProperties(name string, shareHeaders ShareHeaders) error {
+	params := url.Values{"restype": {"share"}, "comp": {"properties"}}
```
garimakhulbe: @enieuw Can we add these attributes using `values.set()` method instead of hard coding them? It improves readability and correct way of adding `values`.

enieuw:
@ahmetalpbalkan @garimakhulbe I've updated the PR with your suggestions.

One test does seem to be failing because the environment the tests are run against isn't empty. The StorageFileSuite. TestListShares creates a share and then uses ListShares to verify if the share was actually created but in the travis build it detects two more shares which are not created from the unit tests causing it to fail. The tests run fine if I run them on an empty storage account. Can you remove the shares from the storage account used for tests?

garimakhulbe:
@enieuw I have removed the extra shares. Can you start build again?

ahmetalpbalkan:
@garimakhulbe Please don't do such manual tasks. That's not the right way to fix this.

@enieuw Tests should be cleaning up the test shares before the file storage test suite starts. We have code doing this in blob_test.go. You can take a look and copy it from there.

enieuw:
@ahmetalpbalkan Do you want me to remove all existing shares before starting the unit tests? That seems rather odd to me as I assumed every run of the testsuite would get a fresh storage account.  Blindly removing all shares seems rather dangerous to me because if a user that isn't aware of this fact runs the tests on an existing storage account they might nuke data left on shares. 

I've modified the unit tests to make use of the prefix functionality to filter out any shares not created by the tests. This will ensure no user data gets deleted and will clean up any dangling shares that might have been left after an interrupted run.

garimakhulbe:
@enieuw Removing test shares might affect parallel running travis builds else its fine since its a test account. `prefix` option seems fine to me.

garimakhulbe:
Thanks for your contribution.

PR #405, opened 2016-09-27
Update dependent packages versions.

jhendrixMSFT:
To pick up the fix made in https://github.com/Azure/go-autorest/pull/89.


jhendrixMSFT:
@garimakhulbe Here's the version bump.

garimakhulbe:
Thanks!

PR #411, opened 2016-10-05
Generate the KeyVault data plane client

cosmincojocar:
I placed KeyVault data plane client in a different folder than KeyVault management client. This is similar with the structure of storage clients.


garimakhulbe:
@cosmincojocar Thank you for your contribution. We will look into this. 

@Azure/go-admins please check if we have swagger exist for this and if not can we do this using swagger.

mcardosos:
> 	DefaultBaseURI = "http://localhost"

I think this extension needs to be implemented in AutoRest first
[https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/swagger-extensions.md#x-ms-parameterized-host](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/swagger-extensions.md#x-ms-parameterized-host)

---
Refers to: keyvault/client.go:39 in c4392c0. [](commit_id = c4392c0e10e61187e51e6749cc690a9d961554e8, deletion_comment = False)

mcardosos:
I think [x-parameterized-host](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/swagger-extensions.md#x-ms-parameterized-host) needs to be implemented in AutoRest first
This is regarding `DefaultBaseURI` in `client.go`

colemickens:
1. Was this hand-edited after generating? If so, why?

2. Why is this in a separate package? It looks be to be data plane AND control plane... and we already have control plane in `arm/keyvault`. That seems bad.

cosmincojocar:
The code is entirely generated from https://github.com/Azure/azure-rest-api-specs/blob/master/keyvault/2015-06-01/swagger/keyvault.json. Only the version.go was added manually since autorest does not generate this file. 

In addition go fmt was applied since the generated code does seems to be formatted.

@colemickens 
There are two different swagger specs for KeyVault:
-  control plane: https://github.com/Azure/azure-rest-api-specs/blob/master/arm-keyvault/2015-06-01/swagger/keyvault.json
- data plane: https://github.com/Azure/azure-rest-api-specs/blob/master/keyvault/2015-06-01/swagger/keyvault.json

It would be nice to have them unified, because otherwise is not so easy to generated both clients in the same package.

mcardosos:
Hello!
Most of the time we use the [rakefile](https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.Go/rakefile.rb) to regen the SDK `rake SDK_VERSION=3.14.16` :) I sent a [PR](https://github.com/Azure/autorest/pull/1501) to AutoRest so it is easy to regen dataplane too.
The `version.go` file should always be AutoRest generated.

keyvault/version.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	major = "5"
+	minor = "0"
+	patch = "0"
+	// Always begin a "tag" with a dash (as per http://semver.org)
+	tag             = "-beta"
+	semVerFormat    = "%s.%s.%s%s"
+	userAgentFormat = "Azure-SDK-for-Go/%s arm-%s/%s"
```
cosmincojocar: That's right.

cosmincojocar:
@mcardosos What about using the acronyms **adp** - Azure Data Plane in the rakefile for data plane packages. This would be similar with the existing naming conversions arm and asm.

cosmincojocar:
What is the resolution of this PR? I seems that the key-vault data plane is not active in gen.go.

mcardosos:
Right now we are not focusing that much on dataplane, and Autorest is not supporting some extensions yet

PR #414, opened 2016-10-11
Added code gen file (original in github.com/Azure/autorest)

mcardosos:
Original PR here...
https://github.com/Azure/autorest/pull/1501


garimakhulbe:
LGTM.

jhendrixMSFT:
LGTM too, ready to merge?

mcardosos:
Not yet D: I discussed with @vishrutshah about creating a new file that does the same, but written in go. What do you think?

jhendrixMSFT:
Would such a conversion make the build script better/easier to maintain?  I get that it makes the repo homogeneous however it seems like it's easier (more natural) to do "buildy" type things in rakefiles than it would be in Go.

vishrutshah:
@jhendrixMSFT I agree with you on simplicity. 
I just had a thought with two questions: 
1. considering packing of the golang, are we ok / Is it good practice to ship ruby file with our go package? 
2. What options gopher prefers for their build infrastructure?

jhendrixMSFT:
@vishrutshah This is a good point, the fact that we've moved this script to the Go SDK repo means it now ships with the SDK which makes it look out-of-place.  What is gopher?

vishrutshah:
@jhendrixMSFT Go's mascot is a [gopher](https://en.wikipedia.org/wiki/Go_(programming_language)#/media/File:Golang.png). I think people who works on go referred as gopher :)

garimakhulbe:
@jhendrixMSFT @vishrutshah Also why we want to ship this build script with SDK? Script is very autorest specific and no use for our users?

jhendrixMSFT:
@garimakhulbe It's not that we want to ship it with the SDK.  I believe the motivation to move it out of the Autorest repo is to make it easier for the Go SDK folks to iterate on it without having to get the Autorest folks involved.

mcardosos:
Translated!

mcardosos:
Ah, Travis is not passing. The gas scanning is throwing issues because I'm using `exec.Command()` with variables as arguments :(

.travis.yml
```diff
@@ -27,3 +29,5 @@ script:
   - go test -v ./management/...
   - test -z "$(golint ./management/... |  grep -v 'should have comment' | grep -v 'stutters' | tee /dev/stderr)"
   - go vet ./management/...
+  - test -z "$(golint ./Gododir/... | tee /dev/stderr)"
+  - go vet ./Gododir/...
```
jhendrixMSFT: nit: fix line ending

rakefile.rb
```diff
@@ -0,0 +1,292 @@
+# This file is use to generate Azure SDK from swagger.
```
jhendrixMSFT: Was the adding of the rake file intentional?

mcardosos: No, I forgot to remove it :p

PR #416, opened 2016-10-16
Bloblease

kpfaulkner:
Noticed in Issue 406 that blob leases hadn't been implemented in the Go client library, so thought I'd give it a go (although a long time dev am fairly new to Go).

Have added functionality for acquire, renew, change, release and break lease for blobs. I've also added in relevant tests to the blob_test.go file. 

Hope it's useful.

Thanks

Ken

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -701,12 +701,46 @@ func (b BlobStorageClient) ChangeLease(container string, name string, currentLea
 }
 
 // ReleaseLease releases the lease for a blob as per https://msdn.microsoft.com/en-us/library/azure/ee691972.aspx
-func (b BlobStorageClient) ReleaseLease(container string, name string) error {
+func (b BlobStorageClient) ReleaseLease(container string, name string, currentLeaseID string) error {
+	params := url.Values{"comp": {"lease"}}
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), params)
+	headers := b.client.getStandardHeaders()
+
+	headers[leaseAction] = releaseLease
+	headers[leaseID] = currentLeaseID
+
+	resp, err := b.client.exec("PUT", uri, headers, nil)
+	if err != nil {
+		return err
+	}
+	defer resp.body.Close()
+
+	if err := checkRespCode(resp.statusCode, []int{http.StatusOK}); err != nil {
+		return err
+	}
+
 	return nil
 }
 
 // RenewLease renews the lease for a blob as per https://msdn.microsoft.com/en-us/library/azure/ee691972.aspx
-func (b BlobStorageClient) RenewLease(container string, name string) error {
+func (b BlobStorageClient) RenewLease(container string, name string, currentLeaseID string) error {
+	params := url.Values{"comp": {"lease"}}
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), params)
+	headers := b.client.getStandardHeaders()
+
+	headers[leaseAction] = renewLease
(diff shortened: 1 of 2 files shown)


storage/blob.go
```diff
@@ -560,6 +577,151 @@ func (b BlobStorageClient) getBlobRange(container, name, bytesRange string, extr
...
+	headers[leaseProposedID] = proposedLeaseID
+	headers[leaseDuration] = strconv.Itoa(leaseTimeInSeconds)
+
+	respHeaders, err := b.leaseCommonPut(container, name, headers, http.StatusCreated)
+	if err != nil {
+		return "", err
+	}
+
+	for k, v := range respHeaders {
+		k = strings.ToLower(k)
```
garimakhulbe: I might understood it wrong but why are you going loop here? `respHeaders` is `map[string][]string` type. Can't you just do respHeaders[leaseID] to get lease id?

kpfaulkner: You're absolutely right. That's a hang over from when I was tinkering with checking for header prefixes. I will remove/replace that.

garimakhulbe: Lets keep it as it is. (like `GetBlobMetadata`). :)

garimakhulbe:
@mcardosos I am going OOF from tomorrow. Will try to complete it tonight. Could you also take a look once? Thanks!

garimakhulbe:
@kpfaulkner For CLA signing..
    https://cla.microsoft.com/

storage/blob.go
ahmetalpbalkan: @kpfaulkner Go does not guarantee case sensitivity of header values. Please use `http.CanonicalizedHeaderKey()` function. If you're trying to get headers from a request, use `resp.Headers.Get()` method. It handles that for you.

kpfaulkner: So the comments in GetBlobMetadata about not being able to trust CanonicalHeaderKey() are no longer valid? Admittedly I didn't put it to the test. If it can be trusted, then I'll change it.

Thoughts?

ahmetalpbalkan: @kpfaulkner I'm not sure about those. Basically Go will change the headers you specify and you should access to headers via resp.headers.Get(). I'd suggest writing some tests to validate behavior.

kpfaulkner: Will do. My concern was that if it *sometimes* modifies case and other times doesn't, will I be able to catch all cases. 

I'll have a dig around.

storage/blob.go
```diff
@@ -560,6 +577,151 @@ func (b BlobStorageClient) getBlobRange(container, name, bytesRange string, extr
...
+			return v[0], nil
+		}
+	}
+
+	// what should we return in case of HTTP 201 but no lease ID?
+	// or it just cant happen? (brave words)
+	return "", nil
+}
+
+// BreakLease breaks the lease for a blob as per https://msdn.microsoft.com/en-us/library/azure/ee691972.aspx
```
mcardosos: Include what is the int you are returning in the doc comments or the method's signature. Something like
```go
// BreakLease breaks the lease for a blob as per https://msdn.microsoft.com/en-us/library/azure/ee691972.aspx.
// It also returns the timeout remaining in the lease in seconds.
func (b BlobStorageClient) BreakLease(container string, name string) (timeout int, error)
```
The same goes for `AcquireLease`, `BreakLeaseWithBreakPeriod` and `ChangeLease`

kpfaulkner:
Will do.

mcardosos:
LGTM :D

PR #425, opened 2016-10-25
Version v6.0.0-beta

mcardosos:
Please, take a look :3
@vishrutshah @jhendrixMSFT


mcardosos:
I tweaked a little the travis file. It looks like godo uses a bunch of somewhat old packages that also changed names, and have been moved around, so, instead of `go build` everything, I am not building the vendors folder that at some point generates in travis. I hope this is not a problem.

arm/authorization/version.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	major = "5"
+	major = "6"
```
vishrutshah: Just curious, do you guys track the breaking changes? if so it'd be great to put it somewhere with the release (in Changelog or breakingchanges file)

glide.lock
```diff
@@ -11,13 +11,43 @@ imports:
   - autorest/validation
 - name: github.com/dgrijalva/jwt-go
   version: 24c63f56522a87ec5339cc3567883f1039378fdb
+- name: github.com/howeyc/gopass
```
vishrutshah: just curious, what are the reasons of adding more dependencies on packages? Are these for development or production usage?

arm/redis/models.go
```diff
@@ -160,50 +158,14 @@ type PatchSchedulesResponse struct {
...
-
-// ReadablePropertiesWithAccessKey is properties generated only in response to
-// CreateOrUpdate redis operation.
-type ReadablePropertiesWithAccessKey struct {
-	RedisVersion       *string             `json:"redisVersion,omitempty"`
-	Sku                *Sku                `json:"sku,omitempty"`
 	RedisConfiguration *map[string]*string `json:"redisConfiguration,omitempty"`
 	EnableNonSslPort   *bool               `json:"enableNonSslPort,omitempty"`
 	TenantSettings     *map[string]*string `json:"tenantSettings,omitempty"`
 	ShardCount         *int32              `json:"shardCount,omitempty"`
```
tombuildsstuff: same as above: this should also be a string not an int32?

arm/redis/models.go
```diff
@@ -266,3 +227,20 @@ type Sku struct {
...
+type UpdateParameters struct {
+	Properties *UpdateProperties `json:"properties,omitempty"`
+}
+
+// UpdateProperties is properties supplied to Update redis operation.
+type UpdateProperties struct {
+	RedisConfiguration *map[string]*string `json:"redisConfiguration,omitempty"`
+	EnableNonSslPort   *bool               `json:"enableNonSslPort,omitempty"`
+	TenantSettings     *map[string]*string `json:"tenantSettings,omitempty"`
+	ShardCount         *int32              `json:"shardCount,omitempty"`
```
tombuildsstuff: same as above: this should also be a string not an int32?

glide.lock
mcardosos: All these new dependencies come from gen.go. I think we can get rid of them

arm/authorization/version.go
mcardosos: Ah, yes, there are some notes in the changelog :3

stack72:
@mcardosos why has this SDK again been bumped another major version? We just had 5.0.0-beta recently. Are there really a lot of breaking changes again here?

mcardosos:
@stack72 There are breaking changes yes, some methods changed names, others are organized in a different way or have new parameters.

PR #427, opened 2016-10-25
management: check x509 decoding errors

mischief:
errors should not be ignored in this codebase, and https://github.com/kisielk/errcheck should probably be run over the whole repo.

this caused a hard to debug auth problem where i mishandled certificate decoding from `azureProfile.json` from the CLI.


mcardosos:
LGTM, @jhendrixMSFT ?

jhendrixMSFT:
Yep looks good, thanks for the fix.  :)

PR #430, opened 2016-10-26
Snapshot

kpfaulkner:
Added blob snapshot functionality + tests


storage/blob.go
```diff
@@ -595,13 +595,51 @@ func (b BlobStorageClient) leaseCommonPut(container string, name string, headers
...
+	if timeout > 0 {
+		params.Add("timeout", strconv.Itoa(timeout))
+	}
+
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), params)
+	resp, err := b.client.exec("PUT", uri, headers, nil)
+	if err != nil {
+		return nil, err
+	}
+
```
mcardosos: Check for correct status code too

storage/blob.go
```diff
@@ -595,13 +595,51 @@ func (b BlobStorageClient) leaseCommonPut(container string, name string, headers
 	return resp.headers, nil
 }
 
+// SnapshotBlob creates a snapshot for a blob as per https://msdn.microsoft.com/en-us/library/azure/ee691971.aspx
+func (b BlobStorageClient) SnapshotBlob(container string, name string, timeout int, currentLeaseID string) (snapshotTimestamp *time.Time, err error) {
```
mcardosos: Add the `extraHeaders` parameter present in other functions (actually, leaseID could go there too). Check how this works with other funcs, for example `PutAppendBlob`

kpfaulkner:
Added changes

storage/blob.go
```diff
@@ -621,6 +621,10 @@ func (b BlobStorageClient) SnapshotBlob(container string, name string, timeout i
 			return nil, err
 		}
 
+		if err := checkRespCode(resp.statusCode, []int{http.StatusCreated}); err != nil {
```
mcardosos: I think this should be between
```go 
resp, err := b.client.exec("PUT", uri, headers, nil)
  	if err != nil {
 		return nil, err
 	}
 ```
and
```go
snapshotResponse := resp.headers.Get(http.CanonicalHeaderKey("x-ms-snapshot"))
```

storage/blob.go
```diff
@@ -750,13 +750,55 @@ func (b BlobStorageClient) leaseCommonPut(container string, name string, headers
...
+	}
+
+	snapshotResponse := resp.headers.Get(http.CanonicalHeaderKey("x-ms-snapshot"))
+	if snapshotResponse != "" {
+		snapshotTimestamp, err := time.Parse(time.RFC3339, snapshotResponse)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := checkRespCode(resp.statusCode, []int{http.StatusCreated}); err != nil {
```
mcardosos: I think the `checkRespCode` makes more sense between
```go
resp, err := b.client.exec("PUT", uri, headers, nil)
if err != nil {
	return nil, err
}
```
and
```go
snapshotResponse := resp.headers.Get(http.CanonicalHeaderKey("x-ms-snapshot"))
```

kpfaulkner:
Oops did I leave in my change?
Afaik that emulator vs real has been there for a long while. I just occasionally tinker with that line during local testing.

s
```diff
@@ -0,0 +1,107 @@
+[33mcommit 7b0aa258822ec949db14c8e858f3ee0aa7255e4f[m
```
vishrutshah: Did you check-in this file by mistake?

kpfaulkner: Yes, one of my checkins was a mistake (was supposed to be just a couple of debug lines for myself). But changes were reversed. 

Do you still see changes that shouldn't be there?

vishrutshah: Yup i still see that changes. Deleting the file named `s` and then committing & pushing that change should remove that change.

kpfaulkner:
done

kpfaulkner:
Can anyone explain that error?
The command "test -z "$(grep 'Severity:\s*HIGH' gas-scan.txt)"" exited with 1.

mcardosos:
It is because of the gas scan. Probably gas got updates recently so, now it classifies some code as not very secure (not you code, your code is ok). I'll take a look.

mcardosos:
Yes, gas was [updated](https://github.com/GoASTScanner/gas/commit/a3fcd96f57246314004c9c0d0e747bfd29171e57) for the credentials stuff. We could update the Travis file to `gas -exclude=G101 -skip=management/examples/*.go -skip=*vendor* -skip=Gododir/* ./... | tee gas-scan.txt` so the credentials rules are ignored. What do you think? @garimakhulbe @jhendrixMSFT @vishrutshah

mcardosos:
Fixing gas on [PR 461](https://github.com/Azure/azure-sdk-for-go/pull/461).

mcardosos:
Hello @kpfaulkner !
I fixed Travis but it looks like it is still executing with the old Travis file. Your PR should pass with Travis newest version. Could you please `git pull origin master` and push again to your fork?

kpfaulkner:
done :)

mcardosos:
@vishrutshah @jhendrixMSFT ready to merge?

jhendrixMSFT:
Yep be sure to squash.  :)

mcardosos:
@kpfaulkner Thanks for all your awesome contributions!

Issue #428, opened 2016-10-25
Ensure that errors don't go unchecked


jhendrixMSFT:
We should scrub the SDK to ensure that we don't have unchecked errors, see https://github.com/Azure/azure-sdk-for-go/pull/427 for more info.

PR #431, opened 2016-10-29
Check errors

mcardosos:
Fixes [issue 428](https://github.com/Azure/azure-sdk-for-go/issues/428)
Please take a look @vishrutshah @jhendrixMSFT


management/vmutils/extensions.go
```diff
@@ -58,7 +58,9 @@ func AddAzureDockerVMExtensionConfiguration(role *vm.Role, dockerPort int, versi
 		return fmt.Errorf(errParamNotSpecified, "role")
 	}
 
-	ConfigureWithExternalPort(role, "docker", dockerPort, dockerPort, vm.InputEndpointProtocolTCP)
+	if err := ConfigureWithExternalPort(role, "docker", dockerPort, dockerPort, vm.InputEndpointProtocolTCP); err != nil {
+		panic(err)
```
jhendrixMSFT: Why not return err here like the other branches in this function?

mcardosos: Ohh, true that. Will change :D

storage/blob.go
```diff
@@ -368,41 +368,46 @@ var (
...
+	defer func() {
+		err = resp.body.Close()
+	}()
 	return checkRespCode(resp.statusCode, []int{http.StatusCreated})
 }
 
 // CreateContainerIfNotExists creates a blob container if it does not exist. Returns
 // true if container is newly created or false if container already exists.
-func (b BlobStorageClient) CreateContainerIfNotExists(name string, access ContainerAccessType) (bool, error) {
+func (b BlobStorageClient) CreateContainerIfNotExists(name string, access ContainerAccessType) (notExists bool, err error) {
```
jhendrixMSFT: Why was the bool return valued named?  I don't see the name referenced in the method body.  I see this same pattern in other methods too.

mcardosos: I named the bool because if I name the error, all returning stuff should be named.
The error is named so errors from `defer` can be returned and checked.

jhendrixMSFT: I see, makes sense.

jhendrixMSFT:
Changes look good.

.travis.yml
```diff
@@ -31,3 +32,10 @@ script:
   - go vet ./management/...
   - test -z "$(golint ./Gododir/... | tee /dev/stderr)"
   - go vet ./Gododir/...
+  - errcheck -ignoretests ./arm/...
```
ahmetalpbalkan: can be shortened: `set -e; for v in ./arm... ./datalake-store/... ./Gododir/...; do errcheck -ignoretests "$v"; done`

management/util.go
```diff
@@ -5,7 +5,9 @@ import (
 	"net/http"
 )
 
-func getResponseBody(response *http.Response) ([]byte, error) {
-	defer response.Body.Close()
+func getResponseBody(response *http.Response) (out []byte, err error) {
+	defer func() {
```
ahmetalpbalkan: I'm not super ok with this one 1) `response.Body` can be nil 2) returning `Close()` error is a behavior change. also why do you add named returns?

ahmetalpbalkan: > Question, will Close() return an error on an empty / nil response.Body?

it will panic with nil dereference error.

.travis.yml
```diff
@@ -31,3 +32,6 @@ script:
   - go vet ./management/...
   - test -z "$(golint ./Gododir/... | tee /dev/stderr)"
   - go vet ./Gododir/...
+  - test -z "$(set -e; for v in ./arm... ./datalake-store/... ./Gododir/...; do (errcheck -ignoretests "$v" | grep -v 'resp.body.Close()' | tee /dev/stderr); done)""
```
ahmetalpbalkan: 1) if you will have test -z, set -e is not necessary.

2) instead of this grep hack, perhaps you can do:

```go
defer func(){
     _ = resp.Body.Close()
}()
```

3) can you please check all callers of `getRequestBody()` make sure that error from the request is nil and then go on reading the body? (if error from http.Do/Get/... is non-nil, then resp.Body will be nil, causing the panic in `defer resp.Body.Close()`

4) You have `grep -v 'resp.body.Close()'`  but I think `body` should be `Body`?

mcardosos:
Travis is passing. How does this look @ahmetalpbalkan ?

.travis.yml
```diff
@@ -30,3 +31,6 @@ script:
   - go vet ./management/...
   - test -z "$(golint ./Gododir/... | tee /dev/stderr)"
   - go vet ./Gododir/...
+  - set -e; for v in ./arm/... ./datalake-store/... ./Gododir/... ./management/... ./storage/...; do errcheck -ignoretests "$v"; done
```
ahmetalpbalkan: are you sure `errcheck` returns with non-zero exitcode when it finds issues? if so you may still need the `test -z`.

storage/blob.go
```diff
@@ -439,7 +439,9 @@ func (b BlobStorageClient) ListContainers(params ListContainersParameters) (Cont
 	if err != nil {
 		return out, err
 	}
-	defer resp.body.Close()
+	defer func() {
```
ahmetalpbalkan: @mcardosos I strongly suggest turning off `errcheck` for this package. It makes the code really ugly with this change. It's idiomatic go to call `defer resp.body.Close()` and discard the output implicitly.

I am sorry that I did not see this coming. I thought you were just going to change 1-2 places in `management` package. But this results in quite ugly code for this package.

We must use static analysis tools like this occasionally to catch bugs but for cases like these, they are deliberate decisions and we should not be using hacks like this all over the place to suppress the static analysis tool. I think `errcheck` is a bit extreme. Perhaps just run `errcheck | grep -v Close` or something again. :(

.travis.yml
mcardosos: Travis fails with `errcheck`, take a look at [this build](https://travis-ci.org/Azure/azure-sdk-for-go/builds/182463249#L1570). And from errcheck [documentation](https://github.com/kisielk/errcheck/blob/master/README.md#exit-codes).

ahmetalpbalkan: That's fair. I think I tried integrating errcheck long time ago manually and it just came with tons of changes which were not necessary (see my other comment). I just started to recall that...

storage/util.go
```diff
@@ -15,10 +15,12 @@ import (
 	"time"
 )
 
-func (c Client) computeHmac256(message string) string {
+func (c Client) computeHmac256(message string) (string, error) {
 	h := hmac.New(sha256.New, c.accountKey)
-	h.Write([]byte(message))
-	return base64.StdEncoding.EncodeToString(h.Sum(nil))
+	if _, err := h.Write([]byte(message)); err != nil {
```
ahmetalpbalkan: Kinda mixed feelings about changing this. the hmac.inner == sha256 type's Write() will never return an error as everything is in-memory. 😕

storage/queue.go
```diff
@@ -319,7 +319,7 @@ func (c QueueServiceClient) DeleteMessage(queue, messageID, popReceipt string) e
 	if err != nil {
 		return err
 	}
-	defer resp.body.Close()
+	deferresp.body.Close()
```
ahmetalpbalkan: error?

storage/util.go
ahmetalpbalkan: yeah this change causes many changes in the upstream callers whereas the error can be just discarded. wdyt?

storage/queue.go
mcardosos: Clumsy me

storage/util.go
mcardosos: `errcheck` doesn't like it. In the other hand, `gas` [doesn't really care](https://github.com/GoASTScanner/gas/blob/master/rules/errors.go#L77). Looking at all this again, I actually think `gas` is more reasonable. I think it would be better if instead of `gas -skip=*/arm/*/models.go -skip=*/management/examples/*.go -skip=*vendor* -skip=*/Gododir/* ./...`, I rewrote that line so for all this skipped files I just exclude whatever [rule](https://github.com/GoASTScanner/gas#available-rules) they break. For example, I'm skipping all the `models.go` files because they "break" rule G101.

Issue #424, opened 2016-10-25
Abort Copy Blob operation.


kpfaulkner:
Currently the library supports the starting of a copy blob operation (CopyBlob function) which starts a single copy and then waits. 

I'd like to propose a new function for Aborting such a copy (see https://msdn.microsoft.com/en-us/library/azure/jj159098.aspx) as well as modifying the existing code a tiny bit.

Currently the exported function CopyBlob calls a non-exported (still trying to get my Go jargon right) startBlobCopy, then once the copying has started it loops checking the status.

I'd like to propose the following changes:
- make startBlobCopy exportable (ie name it StartBlobCopy) that way people can easily fire off a bunch of them without having to perform the wait.
- Add AbortBlobCopy which takes the copyID generated by startBlobCopy and can abort the operation if needed.

No functionality (CopyBlob function) is impacted but gives the users 2 methods (StartBlobCopy and AbortBlobCopy) which are closer (and for me personally more useful) to the REST API.

I've already coded up a PR with the above, but wanted to get opinions before I actually submit it.

Thoughts?

ahmetalpbalkan:
@kpfaulkner sounds good. Folks can help you review if you submit a patch (with tests).

kpfaulkner:
Closing. PR merged

PR #434, opened 2016-10-31
Abortcopy

kpfaulkner:
Adds ability to abort copyblob functionality. 

See issue https://github.com/Azure/azure-sdk-for-go/issues/424


kpfaulkner:
Have removed the test for abort copy since being able to reliably test it is difficult to say the least. More often than not the copy is complete before the abort code can run.

Are we either happy to not specifically test the abort code (feels dirty) or does anyone know of a way to force the slowdown of a copyblob operation?

storage/blob_test.go
```diff
@@ -269,6 +269,30 @@ func (s *StorageBlobSuite) TestBlobCopy(c *chk.C) {
...
+	body := []byte(randString(1024))
+
+	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
+	defer cli.deleteContainer(cnt)
+
+	c.Assert(cli.putSingleBlockBlob(cnt, src, body), chk.IsNil)
+	defer cli.DeleteBlob(cnt, src, nil)
+
+	// given we dont know when it will start, can we even test destination creation?
+	// will just test that an error wasn't thrown for now.
```
mcardosos: Try to test with `BlobExists`.
To test aborting maybe you'll need to copy a bigger blob, so it actually isn't copied so fast that it cannot be aborted.

storage/blob.go
```diff
@@ -1137,6 +1140,36 @@ func (b BlobStorageClient) startBlobCopy(container, name, sourceBlob string) (st
...
+// copyID is generated from StartBlobCopy function.
+// currentLeaseID is required IF the destination blob has an active lease on it.
+// As defined in https://msdn.microsoft.com/en-us/library/azure/jj159098.aspx
+func (b BlobStorageClient) AbortBlobCopy(container, name, copyID, currentLeaseID string, timeout int) error {
+	uri := b.client.getEndpoint(blobServiceName, pathForBlob(container, name), url.Values{"comp": {"copy"}, "copyid": {copyID}})
+	headers := b.client.getStandardHeaders()
+	headers["x-ms-copy-action"] = "abort"
+
+	if timeout > 0 {
+		headers["timeout"] = strconv.Itoa(timeout)
```
kpfaulkner: you're absolutely right...  I blame the scotch :)  Fixing now.
Adding extra Scotch resistant tests :)

storage/blob_test.go
kpfaulkner: I've already tried with blobs up to 2G in size. Am wondering about having to have multiple accounts. Wouldn't want to go cross DC though and up the charges..  Thoughts?

kpfaulkner: Do you think we need a multi gig/different DC test to proceed with this?

mcardosos: Ok, after reading some tests from [other SDKs](https://github.com/Azure/azure-storage-net/blob/68c3ee55a3a6f62a0159cea58005d3fe027312a8/Test/ClassLibraryCommon/Blob/CopyBlobTest.cs#L177), what you can do is try to abort the copy operation when you are sure the operation has ended... It should throw status code [409](https://msdn.microsoft.com/en-us/library/azure/jj159098.aspx#Anchor_3). Sorry for taking so long to answer.

kpfaulkner: Ahhh good idea. I can produce *those* at a drop of a hat (and nothing else). Will update sometime today. Cheers!

kpfaulkner:
Added appropriate negative test for abort copy. Also made WaitForBlobCopy to be exported. Made sense given StartBlobCopy is available.

mcardosos:
LGTM, @jhendrixMSFT @vishrutshah anything else to add?

jhendrixMSFT:
Looks good!

mcardosos:
@kpfaulkner thanks for the contribution!

PR #437, opened 2016-11-02
Escape path when creating Canonicalized Resource for Table

ubercow:
This fixes a bug where `buildCanonicalizedResourceTable` would create an
invalid string when using a PartitionKey or RowKey with a special
character in it, such as "=".

As per the documentation for creating CanonicalizedResource strings for
SharedKey and SharedKeyLite, found at:
https://msdn.microsoft.com/en-gb/library/azure/dd179428.aspx

"Any portion of the CanonicalizedResource string that is derived from
the resource's URI should be encoded exactly as it is in the URI."

`buildCanonicalizedResource` used in other places, was already using
this same method to conform to the documentation.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -305,7 +305,7 @@ func (c Client) buildCanonicalizedResourceTable(uri string) (string, error) {
 	cr := "/" + c.getCanonicalizedAccountName()
 
 	if len(u.Path) > 0 {
-		cr += u.Path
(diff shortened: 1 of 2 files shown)


mcardosos:
LGTM, anything else to add @vishrutshah @jhendrixMSFT ?

mcardosos:
@ubercow Thanks for your contribution!

PR #443, opened 2016-11-03
Add support for queue UpdateMessage operation in Azure Go SDK

amanohar:
Add support for queue UpdateMessage operation in Azure Go SDK

### DIFF
--- a/storage/queue.go
+++ b/storage/queue.go
@@ -82,6 +82,24 @@ func (p PeekMessagesParameters) getParameters() url.Values {
 	return out
 }
 
+// UpdateMessageParameters is the set of options can be specified for Update Messsage
+// operation. A zero struct does not use any preferences for the request.
+type UpdateMessageParameters struct {
+	PopReceipt        string
+	VisibilityTimeout int
+}
+
+func (p UpdateMessageParameters) getParameters() url.Values {
+	out := url.Values{}
+	if len(p.PopReceipt) != 0 {
+		out.Set("popreceipt", p.PopReceipt)
+	}
+	if p.VisibilityTimeout != 0 {
+		out.Set("visibilitytimeout", strconv.Itoa(p.VisibilityTimeout))
+	}
+	return out
+}
+
 // GetMessagesResponse represents a response returned from Get Messages
 // operation.
 type GetMessagesResponse struct {
@@ -298,6 +316,26 @@ func (c QueueServiceClient) DeleteMessage(queue, messageID, popReceipt string) e
 	uri := c.client.getEndpoint(queueServiceName, pathForMessage(queue, messageID), url.Values{
 		"popreceipt": {popReceipt}})
 	resp, err := c.client.exec("DELETE", uri, c.client.getStandardHeaders(), nil)
+	if err != nil {
+		return err
+	}
+	defer resp.body.Close()
+	return checkRespCode(resp.statusCode, []int{http.StatusNoContent})
+}
+
+// UpdateMessage operation deletes the specified message.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/hh452234.aspx
+func (c QueueServiceClient) UpdateMessage(queue string, messageID string, message string, params UpdateMessageParameters) error {
+	uri := c.client.getEndpoint(queueServiceName, pathForMessage(queue, messageID), params.getParameters())
+	req := putMessageRequest{MessageText: message}
+	body, nn, err := xmlMarshal(req)
+	if err != nil {
+		return err
+	}
+	headers := c.client.getStandardHeaders()
+	headers["Content-Length"] = strconv.Itoa(nn)
+	resp, err := c.client.exec("PUT", uri, headers, body)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


storage/queue.go
```diff
@@ -82,6 +82,24 @@ func (p PeekMessagesParameters) getParameters() url.Values {
...
+// UpdateMessageParameters is the set of options can be specified for Update Messsage
+// operation. A zero struct does not use any preferences for the request.
+type UpdateMessageParameters struct {
+	PopReceipt        string
+	VisibilityTimeout int
+}
+
+func (p UpdateMessageParameters) getParameters() url.Values {
+	out := url.Values{}
+	if len(p.PopReceipt) != 0 {
```
ahmetalpbalkan: use `== ""`

storage/queue.go
```diff
@@ -82,6 +82,24 @@ func (p PeekMessagesParameters) getParameters() url.Values {
...
+		out.Set("popreceipt", p.PopReceipt)
+	}
+	if p.VisibilityTimeout != 0 {
```
ahmetalpbalkan: is it a valid use case somebody actually wants to set visibilitytimeout to 0? If so we'll need to use *int which has default value of nil here I think.

storage/queue.go
```diff
@@ -304,3 +322,23 @@ func (c QueueServiceClient) DeleteMessage(queue, messageID, popReceipt string) e
...
+// See https://msdn.microsoft.com/en-us/library/azure/hh452234.aspx
+func (c QueueServiceClient) UpdateMessage(queue string, messageID string, message string, params UpdateMessageParameters) error {
+	uri := c.client.getEndpoint(queueServiceName, pathForMessage(queue, messageID), params.getParameters())
+	req := putMessageRequest{MessageText: message}
+	body, nn, err := xmlMarshal(req)
+	if err != nil {
+		return err
+	}
+	headers := c.client.getStandardHeaders()
+	headers["Content-Length"] = strconv.Itoa(nn)
```
ahmetalpbalkan: fmt.Sprintf("%d",nn) would  be preferred. I realized some other `Itoa`s slipped through already but let's fix it while we still can.

storage/queue_test.go
```diff
@@ -87,7 +87,7 @@ func (s *StorageQueueSuite) TestQueueExists(c *chk.C) {
 	c.Assert(ok, chk.Equals, true)
 }
 
-func (s *StorageQueueSuite) TestPostMessage_PeekMessage_DeleteMessage(c *chk.C) {
+func (s *StorageQueueSuite) TestPostMessage_PeekMessage_PutMessage_DeleteMessage(c *chk.C) {
```
ahmetalpbalkan: maybe add _UpdateMessage as well?

storage/queue.go
amanohar: My understanding is the using len is more efficient vs doing a "". Any specific reason for preference to ""?

storage/queue.go
amanohar: From the documentation: "The new value must be larger than or equal to 0". 

I followed the pattern of getParameters() in GetMessagesParameters. Let me know how you want me to change it.

storage/queue.go
amanohar: Updated.

amanohar:
@ahmetalpbalkan addressed most of your comments. Please take a look. Thanks.

storage/queue.go
ahmetalpbalkan: @amanohar since `0` is a valid value, I think you should make this value `*int` and check here to test if it's nil, then don't set it. but if it is not nil, then dereference its value to set it in `out`.

storage/queue_test.go
amanohar: Sure I agree. Happy to rename this. I was merely following the naming pattern of this test.
Name before I updated the test:  TestPostMessage_PeekMessage_DeleteMessage
After my change: TestPostMessage_PeekMessage_PutMessage_DeleteMessage

I can rename this to:
TestPutMessage_PeekMessage_UpdateMessage_DeleteMessage

Does this work?

storage/queue.go
amanohar: @ahmetalpbalkan, it would be strange that in PutMessageParameters VisibilityTimeout is an int and in UpdateMessageParameters VisibilityTimeout is an *int. However, if you still want it to be that way I can change it. Let me know.

storage/queue_test.go
amanohar: Updated the test name.

storage/queue.go
ahmetalpbalkan: @amanohar there's no efficiency argument here really. It's more about readability. I'm sure the perf difference is negligible.

storage/queue.go
ahmetalpbalkan: @amanohar I don't have a strong preference. You're right about `PutMessageParameters`. I guess while writing that struct I didn't think of `0` being a valid value people can specify... We can go with this code for now I guess. If somebody tries to specify it as `0`, the value will not be sent in the request though. So we'd have to take a breaking change by making these fields *int, but for now I guess it's fine.

amanohar:
Thanks @ahmetalpbalkan for the quick review.

@mcardosos can I get help merging this PR?

PR #444, opened 2016-11-03
storage#DeleteBlobIfExists: fix panic if resp==nil

ahmetalpbalkan:
A resp.body.Close() statement was causing nil pointer dereference
panics when deleteBlob() returned an error and hence resp was nil.

Signed-off-by: Ahmet Alp Balkan <ahmetalpbalkan@gmail.com>


ahmetalpbalkan:
Travis is pretty bad today. I canceled/restarted build 3 times now in the past 45 minutes and it doesn't even kick off the build. https://travis-ci.org/Azure/azure-sdk-for-go/builds/173056994 I will just go ahead and skip CR as this is a trivial change and we need to vendor this in `docker/distribution` repo for a release ASAP.

PR #445, opened 2016-11-04
Generated DocumentDB via AutoRest

tombuildsstuff:
DocumentDB was available in the Azure Specs Repo - so I've generated the code for it via AutoRest (and then set the version number to 6.0 to match the other components)

I've no idea how the other bits of the SDK have been generated - and this appears very different to some of the other ones I've checked.. but this is based on the docs over at autorest?


mcardosos:
Hello @tombuildsstuff 
Exactly which ones are the differences you are referring to? Was it the linting, I guess?

tombuildsstuff:
hey @mcardosos

I was referring to the way the [client sdk constructors work](https://github.com/Azure/azure-sdk-for-go/pull/445/files#diff-c8c21f30b228db90c10036f34a5ed687R22) where they require you to provide the resource group name and account name up front; rather than [like the EventHubs one](https://github.com/Azure/azure-sdk-for-go/blob/master/arm/eventhub/eventhubs.go#L34) for instance where it just takes the SubscriptionID - providing the resource group name / account name when calling the relevant methods. Do I need to manually extend this auto-generated code to be in the correct format? I'm not sure if there's any docs for how to add another auto-generated sdk that I should be looking at? :)

An aside / feedback for auto-rest - I assumed that the generated code would comply with the linting rules by default - but I can raise an issue over at autorest for that :)

mcardosos:
I see. That is because of this [AutoRest issue](https://github.com/Azure/autorest/issues/1532). Actually, that is the only problem this PR could have (another detail, easy to fix, would be to include the Apache license). 
For generating following the linting rules, I run [this file](https://github.com/Azure/azure-sdk-for-go/blob/master/Gododir/gen.go). It generates the SDK, and also runs `go fmt`, `golint`, etc. It would be cool if linting was included in AutoRest, so feel free to add an issue.

tombuildsstuff:
@mcardosos oh cool thanks - I'll take a look into that, do you mind if I send a PR to add that to the Contribution section of the README?

mcardosos:
PRs are always welcome

jhendrixMSFT:
@tombuildsstuff Sorry to take so long.  I've fixed the issue in the generator (https://github.com/Azure/autorest/issues/1532) and was able to generate this package as part of the v7.0.0-beta release which I just published.  Please let us know if it works for you.

Issue #455, opened 2016-11-14
Add support for GraphRbac api


TamalSaha:
We would like to call GraphRbac api via GO. I noticed that it was commented out in the gen.go file. 
https://github.com/Azure/azure-sdk-for-go/blob/master/Gododir/gen.go#L118

Can we get this api added to GO SDK?

Thanks.

jhendrixMSFT:
This is dependent on fixing support for composite swaggers in the Go generator, see https://github.com/Azure/autorest/issues/1746.

mcardosos:
Fixed in https://github.com/Azure/azure-sdk-for-go/pull/561
I should add that it is not using composite swaggers, so some operations might not be supported. Feel free to reopen the issue if you think it is not correctly solved.

PR #456, opened 2016-11-14
Generate GO client for GraphRBAC

TamalSaha:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/455


mcardosos:
Hello @TamalSaha ! Thanks for submitting a PR
Quick question, have you tested this code?

TamalSaha:
@mcardosos , Yes. The generated code works.

TamalSaha:
@mcardosos, FYI, I used the graph RBAC client here: https://github.com/appscode/appctl/blob/master/pkg/cmd/credential/cloud/azure.go#L125

marstr:
Howdy, it looks like continuous integration is failing. Mind rebasing this on-top of master and seeing if that is enough to make CI pass? If so, we'll take another look and get this merged in promptly.

TamalSaha:
@marstr , I rebased the pr and that seems to fix CI tests. PTAL.

salameer:
Hi @TamalSaha, thanks so much for the contribution here, we will not be able to accept this PR right now for the following reasons:
- This is based on a composite swagger.
- This version Go Autorest Generator doesn't support these  swaggers.

We will ping you once the generator is updated to regenerate this code

TamalSaha:
@salameer is there any issue that I can follow regarding "the generator is updated to regenerate composite swagger".?

colemickens:
Can we just generate without composite swagger? I've still seen nothing to convince me that composite swagger is at all a good idea. As far as I can tell it's just going to make the already-precarious versioning problem even worse (and then adding versioning across Azure Stack on top...)

colemickens:
It seems a shame that we would limit functionality in our SDKs because the C# folks don't want to ship too many packages (despite other teams in the companies happily publishing small .NET nuget packages so that customers can selectively opt-in...)

salameer:
Hi @colemickens  @TamalSaha ,

- Sooo, regarding composite swaggers there's definitely alot of different views and opinions around it. however the good thing here is that we're driving toward a model where every language Autorest generator can decide however it want to process composite swaggers (So Go doesn't have to follow the same direction as .Net)

- As for this PR what you suggested earlier of releasing this on the side (seperately) is something that we're planning on doing this Februry release of the SDK to unblock customers. however the risk here @TamalSaha  is that when we finlaize the composite swagger processing strategy for Go SDK we might be introducing some breaing changes there.

- For  now I will go ahead and close this PR and you can track the issue in the original ticket #543 which I scheduled for this release

TamalSaha:
> As for this PR what you suggested earlier of releasing this on the side (seperately) is something that we're planning on doing this Februry release of the SDK to unblock customers. however the risk here @TamalSaha is that when we finlaize the composite swagger processing strategy for Go SDK we might be introducing some breaing changes there.

I think we are used to breaking changes at this point . :) Thanks for explaining the way forward.

PR #463, opened 2016-11-17
Add support for more file storage APIs.

jhendrixMSFT:
This change adds support for file storage operations on directories and
files; this includes creation, deletion and getting/setting of properties
and metadata.

### DIFF
--- a/storage/file.go
+++ b/storage/file.go
@@ -150,22 +150,26 @@ type ShareHeaders struct {
 }
 
 type compType string
-type resourceType string
 
 const (
-	compNone          compType     = ""
-	compList          compType     = "list"
-	compMetadata      compType     = "metadata"
-	compProperties    compType     = "properties"
-	compRangeList     compType     = "rangelist"
-	resourceDirectory resourceType = "directory"
-	resourceFile      resourceType = ""
-	resourceShare     resourceType = "share"
+	compNone       compType = ""
+	compList       compType = "list"
+	compMetadata   compType = "metadata"
+	compProperties compType = "properties"
+	compRangeList  compType = "rangelist"
 )
 
 func (ct compType) String() string {
 	return string(ct)
 }
+
+type resourceType string
+
+const (
+	resourceDirectory resourceType = "directory"
+	resourceFile      resourceType = ""
+	resourceShare     resourceType = "share"
+)
 
 func (rt resourceType) String() string {
 	return string(rt)
--- a/storage/file.go
+++ b/storage/file.go
@@ -2,9 +2,12 @@ package storage
 
 import (
 	"encoding/xml"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 )
 
@@ -19,12 +22,57 @@ type Share struct {
 	Properties ShareProperties `xml:"Properties"`
 }
 
+// A Directory is an entry in DirsAndFilesListResponse.
+type Directory struct {
+	Name string `xml:"Name"`
+}
+
+// A File is an entry in DirsAndFilesListResponse.
+type File struct {
+	Name       string         `xml:"Name"`
+	Properties FileProperties `xml:"Properties"`
+}
+
 // ShareProperties contains various properties of a share returned from
 // various endpoints like ListShares.
 type ShareProperties struct {
 	LastModified string `xml:"Last-Modified"`
 	Etag         string `xml:"Etag"`
 	Quota        string `xml:"Quota"`
+}
+
+// DirectoryProperties contains various properties of a directory returned
+// from various endpoints like GetDirectoryProperties.
+type DirectoryProperties struct {
+	LastModified string `xml:"Last-Modified"`
+	Etag         string `xml:"Etag"`
+}
+
+// FileProperties contains various properties of a file returned from
+// various endpoints like ListDirsAndFiles.
+type FileProperties struct {
+	CacheControl       string `header:"x-ms-cache-control"`
+	ContentLength      uint64 `xml:"Content-Length"`
+	ContentType        string `header:"x-ms-content-type"`
+	CopyCompletionTime string
+	CopyID             string
+	CopySource         string
+	CopyProgress       string
+	CopyStatusDesc     string
+	CopyStatus         string
+	Disposition        string `header:"x-ms-content-disposition"`
+	Encoding           string `header:"x-ms-content-encoding"`
+	Etag               string
+	Language           string `header:"x-ms-content-language"`
+	LastModified       string
+	MD5                string `header:"x-ms-content-md5"`
+}
+
+// FileStream contains file data returned from a call to GetFile.
+type FileStream struct {
+	Body       io.ReadCloser
+	Properties *FileProperties
+	Metadata   map[string]string
 }
 
 // ShareListResponse contains the response fields from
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
@mcardosos Please take a look.  :)

storage/file.go
```diff
@@ -53,12 +101,83 @@ type ListSharesParameters struct {
...
+type httpVerb string
+type resourceType string
+
+const (
+	compNone          compType     = ""
+	compList          compType     = "list"
+	compMetadata      compType     = "metadata"
+	compProperties    compType     = "properties"
+	compRangeList     compType     = "rangelist"
+	httpGet           httpVerb     = "GET"
```
mcardosos: Refactor! yesss! I bring [this](https://golang.org/pkg/net/http/#pkg-constants) to the table.

storage/file.go
```diff
@@ -81,9 +192,94 @@ func (p ListSharesParameters) getParameters() url.Values {
...
+
+	var out DirsAndFilesListResponse
+	resp, err := f.listContent(path, q, nil)
+	if err != nil {
+		return out, err
+	}
+
+	defer resp.body.Close()
+	err = xmlUnmarshal(resp.body, &out)
+	return out, nil
```
mcardosos: Aren't you returning the unmarshal error?

storage/file.go
```diff
@@ -81,9 +192,94 @@ func (p ListSharesParameters) getParameters() url.Values {
...
+// See https://msdn.microsoft.com/en-us/library/azure/dn166980.aspx
+func (f FileServiceClient) ListDirsAndFiles(path string, params ListDirsAndFilesParameters) (DirsAndFilesListResponse, error) {
+	q := mergeParams(params.getParameters(), getURLInitValues(compList, resourceDirectory))
+
+	var out DirsAndFilesListResponse
+	resp, err := f.listContent(path, q, nil)
+	if err != nil {
+		return out, err
+	}
+
```
mcardosos: Check for response code

storage/file.go
```diff
@@ -81,9 +192,94 @@ func (p ListSharesParameters) getParameters() url.Values {
...
+		headers = make(map[string]string)
+		headers["Range"] = listRange.String()
+	}
+
+	var out FileRanges
+	resp, err := f.listContent(path, params, headers)
+	if err != nil {
+		return out, err
+	}
+
```
mcardosos: I would move `defer resp.body.Close()` here.
What about the `Last-Modified` header?
The same as above, check for response code, and return the unmarshal error.

storage/file.go
```diff
@@ -92,40 +288,171 @@ func pathForFileShare(name string) string {
...
+	resp, err := f.listContent("", q, nil)
 	if err != nil {
 		return out, err
 	}
-	defer resp.body.Close()
 
+	defer resp.body.Close()
 	err = xmlUnmarshal(resp.body, &out)
-	return out, err
+	return out, nil
```
mcardosos: Return the error too

storage/file.go
jhendrixMSFT: Fixed.

storage/file.go
jhendrixMSFT: Fixed.

storage/file.go
jhendrixMSFT: I added the response code check to the listContent method so we only have to do it in one place (nice catch).

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+// See https://msdn.microsoft.com/en-us/library/azure/dn166993.aspx
+func (f FileServiceClient) CreateDirectory(path string) error {
+	return f.createResource(path, resourceDirectory, nil)
+}
+
+// CreateFile operation creates a new file or replaces an existing one.  Note that this
+// only initializes the file, call PutRange to add content.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dn194271.aspx
+func (f FileServiceClient) CreateFile(path string, maxSize uint64) error {
```
mcardosos: I would add an extraHeaders parameter so other headers can be specified

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+}
+
+// CreateDirectory operation creates a new directory in the specified share. If a
+// directory with the same name already exists, the operation fails.
 //
-// See https://msdn.microsoft.com/en-us/library/azure/dn167008.aspx
-func (f FileServiceClient) CreateShare(name string) error {
-	resp, err := f.createShare(name)
+// See https://msdn.microsoft.com/en-us/library/azure/dn166993.aspx
+func (f FileServiceClient) CreateDirectory(path string) error {
```
mcardosos: Add support for `x-ms-meta-name:value`

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+
+// modifies a range of bytes in the specified file
+func (f FileServiceClient) modifyRange(path string, bytes io.Reader, fileRange FileRange) error {
+	if err := f.checkForStorageEmulator(); err != nil {
+		return err
+	}
+	if fileRange.End < fileRange.Start {
+		return errors.New("the value for rangeEnd must be greater than or equal to rangeStart")
+	}
+	if fileRange.End-fileRange.Start > 4194304 {
```
mcardosos: This constraint is just for `x-ms-write: update`

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+// See https://msdn.microsoft.com/en-us/library/azure/dn194276.aspx
+func (f FileServiceClient) ClearRange(path string, fileRange FileRange) error {
+	return f.modifyRange(path, nil, fileRange)
+}
+
+// PutRange writes a range of bytes to a file.  Note that the length of bytes must
+// match (rangeEnd - rangeStart) + 1 with a maximum size of 4MB.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dn194276.aspx
+func (f FileServiceClient) PutRange(path string, bytes io.Reader, fileRange FileRange) error {
```
mcardosos: Will you add support for `Content-MD5` ?

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
 // See https://msdn.microsoft.com/en-us/library/azure/dd179352.aspx
```
mcardosos: This is the correct link `https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/list-shares`

jhendrixMSFT: It appears that all of the links are redirecting to the new docs location.  Let's do a scrub of these in a separate PR.

storage/file.go
jhendrixMSFT: Fixed.

storage/file.go
jhendrixMSFT: CreateFile and CreateDirectory were modeled from CreateShare which doesn't allow one to specify metadata at the time of creation.  If we want to fix this for files and dirs I think we should also fix it for shares even though it's a breaking change.  What do you think?

mcardosos: Oh D: maybe in a separate PR? all breaking changes together?

jhendrixMSFT: Sure we could include this change as part of the v7 SDK update I'm working on.

storage/file.go
mcardosos: Agree

storage/file.go
mcardosos: It could be

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+			"Range": fileRange.String(),
+		}
+	}
+
+	resp, err := f.getResourceNoClose(path, compNone, resourceFile, http.MethodGet, extraHeaders)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = checkRespCode(resp.statusCode, []int{http.StatusOK, http.StatusPartialContent}); err != nil {
```
mcardosos: PartialContent?

storage/file.go
```diff
@@ -92,40 +292,176 @@ func pathForFileShare(name string) string {
...
+
 	defer resp.body.Close()
 	return checkRespCode(resp.statusCode, []int{http.StatusCreated})
 }
 
+// GetFile operation reads or downloads a file from the system, including its
+// metadata and properties.
+//
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/get-file
+func (f FileServiceClient) GetFile(path string, fileRange *FileRange) (*FileStream, error) {
```
mcardosos: Just commenting this so we don't forget, this operation also supports md5.

jhendrixMSFT:
@mcardosos Any other feedback before I merge?

storage/file.go
```diff
@@ -53,12 +101,76 @@ type ListSharesParameters struct {
...
 // ShareHeaders contains various properties of a file and is an entry
 // in SetShareProperties
 type ShareHeaders struct {
 	Quota string `header:"x-ms-share-quota"`
 }
 
+type compType string
+type resourceType string
+
+const (
```
marstr: What would you think of splitting this up into two const blocks to denote the two separate types represented here? Also, you may be able to bring this tighter to two enums by following the following pattern:
[printable-enums-in-go](http://stackoverflow.com/questions/23917916/printable-enums-in-go)

Thoughts?

jhendrixMSFT: Good idea on splitting these into two blocks, I'll make that change.  I don't think switching to printable enums here is worth the added complexity (albeit not much) as for this case the string values are what's actually important (we place them into the request header).

mcardosos:
LGTM

PR #478, opened 2016-11-30
Version v7.0.1-beta

jhendrixMSFT:


jhendrixMSFT:
@mcardosos @marstr Here's the updated SDK with the flattening fix.

arm/cognitiveservices/accounts.go
```diff
@@ -61,7 +61,7 @@ func (client AccountsClient) Create(resourceGroupName string, accountName string
 			Constraints: []validation.Constraint{{Target: "parameters.Sku", Name: validation.Null, Rule: true,
 				Chain: []validation.Constraint{{Target: "parameters.Sku.Tier", Name: validation.ReadOnly, Rule: true, Chain: nil}}},
 				{Target: "parameters.Location", Name: validation.Null, Rule: true, Chain: nil},
-				{Target: "parameters.Properties", Name: validation.Null, Rule: true, Chain: nil}}}}); err != nil {
+				{Target: "parameters.map[string]interface{}", Name: validation.Null, Rule: true, Chain: nil}}}}); err != nil {
```
mcardosos: I wonder if this actually works? Maybe the problem comes from a bad defined swagger?

tombuildsstuff:
@jhendrixMSFT out of interest, given the size of these PR's, what kind of tests are you running to validate this?

arm/cognitiveservices/accounts.go
mcardosos: Found the [guilty line](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-cognitiveservices/2016-02-01-preview/swagger/cognitiveservices.json#L582)

jhendrixMSFT: @mcardosos and I debugged this, it was a bug in the support for flattening.

jhendrixMSFT:
@tombuildsstuff Unfortunately our coverage is a bit ad-hoc right now which is how this regression slipped in.  :(  We have a work item to improve our coverage and get it into our CI for the next release.

PR #480, opened 2016-12-01
storage: return AzureStorageServiceError for empty responses

ahmetalpbalkan:
Currently, if somebody was trying to implement an `is404(error )`
for errors returned from "storage" package, they would miss
out the errors coming from HEAD requests, such as GetBlobProperties
as these responses do not have a response body that contains an error
message.

Therefore, reusing AzureStorageServiceError to return a sensible
error message that still contains the status code (int) to implement
a `err.(AzureStorageServiceError).Code==404` and the status text (str)
that contains a helpful error message instead of just "404 Not Found" with
an indicator saying "no response body was available".

That simplifies the code that would implement an is404(error) in the
downstream clients.

cc: @yuwamsft FYI

### DIFF
--- a/storage/client_test.go
+++ b/storage/client_test.go
@@ -218,6 +218,22 @@ func (s *StorageClientSuite) TestReturnsStorageServiceError(c *chk.C) {
 	c.Assert(v.StatusCode, chk.Equals, 404)
 	c.Assert(v.Code, chk.Equals, "ContainerNotFound")
 	c.Assert(v.Code, chk.Not(chk.Equals), "")
+	c.Assert(v.RequestID, chk.Not(chk.Equals), "")
+}
+
+func (s *StorageClientSuite) TestReturnsStorageServiceError_withoutResponseBody(c *chk.C) {
+	// HEAD on non-existing blob
+	_, err := getBlobClient(c).GetBlobProperties("non-existing-blob", "non-existing-container")
+
+	c.Assert(err, chk.NotNil)
+	c.Assert(err, chk.FitsTypeOf, AzureStorageServiceError{})
+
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
Needs another pair of eyes and merging. :)

storage/client_test.go
```diff
@@ -218,6 +218,22 @@ func (s *StorageClientSuite) TestReturnsStorageServiceError(c *chk.C) {
...
+func (s *StorageClientSuite) TestReturnsStorageServiceError_withoutResponseBody(c *chk.C) {
+	// HEAD on non-existing blob
+	_, err := getBlobClient(c).GetBlobProperties("non-existing-blob", "non-existing-container")
+
+	c.Assert(err, chk.NotNil)
+	c.Assert(err, chk.FitsTypeOf, AzureStorageServiceError{})
+
+	v, ok := err.(AzureStorageServiceError)
+	c.Check(ok, chk.Equals, true)
+	c.Assert(v.StatusCode, chk.Equals, 404)
```
mcardosos: Only tiny detail... `http.StatusNotFound`.
Still LGTM, @jhendrixMSFT @marstr would you add something?
From my side, feel free to merge.

ahmetalpbalkan: Fixed.

marstr:
I checked the Travis failures and confirmed that they are strictly a result of issue #485. Once these are rebased or merged, the checks will pass.

ahmetalpbalkan:
@marstr It was an oversight. The service error is constructed from response body, so the case when there is no response body was not shoehorned into this error type and was returned as a separate error as if it would never happen. But apparently HEAD requests is exactly when it happens.

PR #487, opened 2016-12-07
Updating gas skip paths to accomodate new globs.

marstr:
gas changed the way their globs are resolved to no longer use
Go/filepath, but instead a package published by an individual. This
caused a change in behavior, where there must be a prefix '/' when
providing a pattern that starts with a path.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
   - glide install
 
 script:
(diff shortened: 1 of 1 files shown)


mcardosos:
LGTM

marstr:
Danke ☺

PR #488, opened 2016-12-08
Adding missing import that wasn't caught in last pull-request.

marstr:
Issue 485 masked the issue because it appeared that gas was the only
cause of failure in the integration tests.


marstr:
Because of the trivial nature of this change, combined with the fact that it is only to fix integration tests, I'm going to do the generally shady act of approving and merging my own pull-request.

PR #497, opened 2017-01-03
Adding Go Report Card

marstr:
I added support for Go Report Card, which is a popular lint-like service that helps build confidence in folks hoping to adopt our package.

We currently have an A+ rating, but are not at 100%. I made some of the easy changes that will help bring us closer to 100. It is worth noting that I did not fix issues that are results of mistakes in swagger files (mis-spelled words) and I did not yet go add documentation to a handful of public methods that have no comment.

### DIFF
--- a/storage/file.go
+++ b/storage/file.go
@@ -425,6 +425,10 @@ func (f FileServiceClient) GetFile(path string, fileRange *FileRange) (*FileStre
 	}
 
 	props, err := getFileProps(resp.headers)
+	if err != nil {
+		return nil, err
+	}
+
 	md := getFileMDFromHeaders(resp.headers)
 	return &FileStream{Body: resp.body, Properties: props, Metadata: md}, nil
 }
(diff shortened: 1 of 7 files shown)


mcardosos:
Cool! LGTM
On the other hand, now looking at this code I think some errors will look a bit mysterious, but also think those details should be fixed in another PR.

storage/file.go
```diff
@@ -425,6 +425,10 @@ func (f FileServiceClient) GetFile(path string, fileRange *FileRange) (*FileStre
 	}
 
 	props, err := getFileProps(resp.headers)
+	if err != nil {
+		return nil, err
```
mcardosos: This is one of those mysterious looking errors. If `getFileProps` returns an error and then it is returned by `GetFile`, we'll just get a `ParseUint` error. We won't really know what the problem is unless we dive into the implementation.

jhendrixMSFT: Agreed the error would be vague.  It would probably help to have getFileProps() return a detailed message in its error object.  I don't think we need to fix it in this PR though.

PR #498, opened 2017-01-03
Adding gas to glide definitions.

marstr:
Updating CI so that changes so that we rely on a consistent version of gas.


marstr:
I've decided not to attempt vendoring gas in this manner.

PR #507, opened 2017-01-17
Adopting latest go-autorest patch.

marstr:
Updated glide.yaml to demand at least go-autorest v7.2.2, but accept any minor version or patch updates.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,23 @@
-hash: 7407050cee9bb9ce89e23ef26bce4051cce63d558338a4937f027a18b789e3a1
-updated: 2016-12-08T17:31:55.7578123-08:00
+hash: 3a11da81c75582dfd5068c841666b0efe621a6191b1b976d389b0223aa118472
+updated: 2017-01-17T12:47:26.9853413-08:00
 imports:
+- name: github.com/Azure/azure-sdk-for-go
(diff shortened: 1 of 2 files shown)


glide.lock
```diff
@@ -1,50 +1,65 @@
-hash: 7407050cee9bb9ce89e23ef26bce4051cce63d558338a4937f027a18b789e3a1
-updated: 2016-12-08T17:31:55.7578123-08:00
+hash: 3a11da81c75582dfd5068c841666b0efe621a6191b1b976d389b0223aa118472
+updated: 2017-01-17T12:47:26.9853413-08:00
 imports:
+- name: github.com/Azure/azure-sdk-for-go
```
mcardosos: Do you know why this changed?

uthark: Probably this is a change in end of line characters (\r\n vs \n)

colemickens: The whole import is new though... Is it expected now that glide shows its main project as an "import" ?

marstr: Ah, I suspect this is related to the fact that I did this from a fork. When I ran glide up, it probably recognized that I was taking dependencies on github.com/Azure/azure-sdk-for-go somehow, perhaps there are cross-package dependencies.

Issue #370, opened 2016-08-02
[docs] Instructions on how to create service principal account


ahmetalpbalkan:
Apparently Node SDK maintains a doc here: https://github.com/Azure/azure-sdk-for-node/blob/master/Documentation/Authentication.md

We should link to this instead of blog posts etc in https://github.com/Azure/azure-sdk-for-go/tree/master/arm#first-a-sidenote-authentication-and-the-azure-resource-manager

marstr:
Fixed. If you do not feel my PR truly fixed your issue, feel free to reopen.

PR #511, opened 2017-01-20
Updating ARM README to ref. the Node SDK's Auth tutorial

marstr:
This in response to issue Azure/azure-sdk-for-go#370


mcardosos:
LGTM
About Travis, looks like gas is not behaving and it isn't skipping the vendor folder :I

PR #515, opened 2017-01-25
Got rid of ReadOnly validation

mcardosos:
@colemickens @jhendrixMSFT @marstr 
PTAL!


jhendrixMSFT:
I'm good with the change once @colemickens verifies it resolves their issue.

marstr:
Before we tag this as a release, let's make sure to have @colemickens play with this commit to ensure there aren't other issues.

marstr:
Actually, instead of immediately putting this on `master`, let's commit this to a test branch on top of master for @colemickens to play with. That way folks using `go get -u github.com/Azure/azure-sdk-for-go` don't pull down these bits until they're more validated.

colemickens:
It shouldn't be a problem for me to pull from @mcardosos's remote. No need to put it explicitly on this repo unless you really want to.

marstr:
Ah, good point. Let us know when you've finished validating.

colemickens:
I apologize for the delay on testing this. Someone hotfixed Kubernetes after this PR was submitted so I deprioritized it a bit.

I can confirm that this (along with removing the workaround) results in successful operation. The PR is good.

Thanks again for this.

mcardosos:
No worries!
Thanks for taking the time to verify the PR worked as expected :D

colemickens:
@mcardosos Two follow up quesitons:

1. Is this "fix" permanent? There's no risk of these popping back up on next re-gen?

2. Is there any reason you know of that I should not go ahead and move Kubernetes over to the latest commit? Maybe I should wait for a tagged release? Is that coming?

Thanks!

mcardosos:
@colemickens 
2 - I am tagging this today. If you use the storage dataplane, take a look into the latest commits. For the ARM side you shouldn't have problems.

marstr:
(I'm not @mcardosos, but I can answer too!)

1. Yep, we're making this particular change permanent. If we bring back validation, it would look radically different and we'll definitely keep this and related threads in mind.
2. A tagged release is coming right quick. Expect to see this Monday. (We're currently working on it, and you can see our progress in PR #528 )

colemickens:
Per today's releases, I plan to have Kubernetes move to: 

* azure-sdk-for-go: `v8.0.0-beta`
* go-autorest: `v7.2.5`

Please let me know if this is not advisable. Thanks.

Issue #514, opened 2017-01-24
Update gas to run only on master branch


marstr:
Gas has failed intermittently, preventing PR CI from completing. These failures can discourage contributions. To prevent this, we'll only run gas on the master branch.

PR #521, opened 2017-01-27
Updating gas to only run during CI on master

marstr:
As per issue #514, I've written a script to mask running gas unless we are on the release branch "master"


marstr:
Sorry for the noise in this PR folks, I'm trying to get Travis to trigger a new build.

PR #526, opened 2017-01-31
Refactored Authorization

mcardosos:
Authorization was a bit messy and incomplete, moved all of it to its own files too.
PTAL!
@marstr @jhendrixMSFT

### DIFF
--- a/storage/authorization.go
+++ b/storage/authorization.go
@@ -51,7 +51,7 @@ func (c *Client) getSharedKey(verb, url string, headers map[string]string, auth 
 		return "", err
 	}
 
-	canString := c.buildCanonicalizedString(verb, headers, canRes, auth)
+	canString := buildCanonicalizedString(verb, headers, canRes, auth)
 	return c.createAuthorizationHeader(canString, auth), nil
 }
 
@@ -114,7 +114,7 @@ func (c *Client) getCanonicalizedAccountName() string {
 	return strings.TrimSuffix(c.accountName, "-secondary")
 }
 
-func (c *Client) buildCanonicalizedString(verb string, headers map[string]string, canonicalizedResource string, auth authentication) string {
+func buildCanonicalizedString(verb string, headers map[string]string, canonicalizedResource string, auth authentication) string {
 	contentLength := headers[headerContentLength]
 	if contentLength == "0" {
 		contentLength = ""
@@ -143,7 +143,7 @@ func (c *Client) buildCanonicalizedString(verb string, headers map[string]string
 			headers[headerIfNoneMatch],
 			headers[headerIfUnmodifiedSince],
 			headers[headerRange],
-			c.buildCanonicalizedHeader(headers),
+			buildCanonicalizedHeader(headers),
 			canonicalizedResource,
 		}, "\n")
 	case sharedKeyForTable:
@@ -160,7 +160,7 @@ func (c *Client) buildCanonicalizedString(verb string, headers map[string]string
 			headers[headerContentMD5],
 			headers[headerContentType],
 			date,
-			c.buildCanonicalizedHeader(headers),
+			buildCanonicalizedHeader(headers),
 			canonicalizedResource,
 		}, "\n")
 	case sharedKeyLiteForTable:
@@ -172,7 +172,7 @@ func (c *Client) buildCanonicalizedString(verb string, headers map[string]string
 	return canString
 }
 
-func (c *Client) buildCanonicalizedHeader(headers map[string]string) string {
+func buildCanonicalizedHeader(headers map[string]string) string {
 	cm := make(map[string]string)
 
 	for k, v := range headers {
@@ -195,13 +195,14 @@ func (c *Client) buildCanonicalizedHeader(headers map[string]string) string {
 
 	ch := bytes.NewBufferString("")
 
-	completeHeaders := []string{}
 	for _, key := range keys {
-		completeHeaders = append(completeHeaders, fmt.Sprintf("%s:%s", key, cm[key]))
+		ch.WriteString(key)
+		ch.WriteRune(':')
+		ch.WriteString(cm[key])
+		ch.WriteRune('\n')
 	}
-	ch.WriteString(strings.Join(completeHeaders, "\n"))
 
-	return string(ch.Bytes())
+	return strings.TrimSuffix(string(ch.Bytes()), "\n")
 }
 
 func (c *Client) createAuthorizationHeader(canonicalizedString string, auth authentication) string {
--- a/storage/authorization.go
+++ b/storage/authorization.go
@@ -2,6 +2,7 @@
 package storage
 
 import (
+	"bytes"
 	"fmt"
 	"net/url"
 	"sort"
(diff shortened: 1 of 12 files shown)


storage/authorization.go
```diff
@@ -0,0 +1,220 @@
...
+	ch := ""
+
+	for i, key := range keys {
+		if i == len(keys)-1 {
+			ch += fmt.Sprintf("%s:%s", key, cm[key])
+		} else {
+			ch += fmt.Sprintf("%s:%s\n", key, cm[key])
+		}
+	}
+	return ch
```
marstr: To get more performant results, I'd rewrite the above few lines to:
``` go
ch := bytes.NewBufferString("")
for _, key := range keys {
    ch.WriteString(key)
    ch.WriteRune(':')
    ch.WriteString(cm[key])
    ch.WriteRune('\n')
}
return string(bytes.TrimSuffix(ch.Bytes(), []byte("\n")))
```

storage/authorization.go
```diff
@@ -0,0 +1,220 @@
...
+				if i == len(keys)-1 {
+					cr += fmt.Sprintf("%s:%s", key, strings.Join(params[key], ","))
+				} else {
+					cr += fmt.Sprintf("%s:%s\n", key, strings.Join(params[key], ","))
+				}
+			}
+		}
+	} else {
+		// search for "comp" parameter, if exists then add it to canonicalizedresource
+		for key := range params {
```
marstr: Params is type Values which is just an alias for a map. There's no need to do this in a loop, it can be simplified to:

``` go
else if val, isPresent := params["comp"]; isPresent {
    cr += "?comp=" + val[0]
}
```

storage/authorization.go
```diff
@@ -0,0 +1,220 @@
...
+	canRes, err := c.buildCanonicalizedResource(url, auth)
+	if err != nil {
+		return "", err
+	}
+
+	canString := c.buildCanonicalizedString(verb, headers, canRes, auth)
+	return c.createAuthorizationHeader(canString, auth), nil
+}
+
+func (c Client) buildCanonicalizedResource(uri string, auth authentication) (string, error) {
```
marstr: Like below, this method would achieve better performance and not make as many allocations if it was rewritten to use buffers like .NET StringBuilder instead of just using string concatenation.

storage/authorization.go
```diff
@@ -0,0 +1,220 @@
...
+	headerContentType       = "Content-Type"
+	headerContentMD5        = "Content-MD5"
+	headerIfModifiedSince   = "If-Modified-Since"
+	headerIfMatch           = "If-Match"
+	headerIfNoneMatch       = "If-None-Match"
+	headerIfUnmodifiedSince = "If-Unmodified-Since"
+	headerRange             = "Range"
+)
+
+func (c Client) addAuthorizationHeader(verb, url string, headers *map[string]string, auth authentication) error {
```
mcardosos: Fixed

storage/authorization.go
mcardosos: Fixed. Also used `strings.Join()`

storage/authorization.go
mcardosos: Fixed

storage/authorization.go
mcardosos: Fixed

storage/authorization.go
```diff
@@ -0,0 +1,217 @@
...
+	canRes, err := c.buildCanonicalizedResource(url, auth)
+	if err != nil {
+		return "", err
+	}
+
+	canString := c.buildCanonicalizedString(verb, headers, canRes, auth)
+	return c.createAuthorizationHeader(canString, auth), nil
+}
+
+func (c *Client) buildCanonicalizedResource(uri string, auth authentication) (string, error) {
```
marstr: I don't see a reference to the receiver in this function. Perhaps this should be "static"?

storage/authorization.go
```diff
@@ -0,0 +1,217 @@
...
+		keys = append(keys, key)
+	}
+
+	sort.Strings(keys)
+
+	ch := bytes.NewBufferString("")
+
+	completeHeaders := []string{}
+	for _, key := range keys {
+		completeHeaders = append(completeHeaders, fmt.Sprintf("%s:%s", key, cm[key]))
```
marstr: Given that there is already a builder present, it seems odd to not take advantage of it. How about we express this this as:
``` go
for _, key := range keys {
    ch.WriteString(key)
    ch.WriteRune(':')
    ch.WriteString(cm[key])
    ch.WriteRune('\n')
}
```
Then you could remove line 202.

storage/authorization.go
mcardosos: Needed for `getCanonicalizedAccountName`

mcardosos: Still, got rid of it in other funcs

storage/authorization.go
mcardosos: Fixed

storage/authorization.go
```diff
@@ -0,0 +1,218 @@
...
+
+type authentication string
+
+const (
+	sharedKey             authentication = "sharedKey"
+	sharedKeyForTable     authentication = "sharedKeyTable"
+	sharedKeyLite         authentication = "sharedKeyLite"
+	sharedKeyLiteForTable authentication = "sharedKeyLiteTable"
+
+	// headers
```
jhendrixMSFT: We should consolidate these headers into a common source file as they are repeated elsewhere.  We don't have to do it now but I would open an issue to track it so we can clean it up later.

storage/authorization.go
```diff
@@ -0,0 +1,218 @@
...
+	headerContentType       = "Content-Type"
+	headerContentMD5        = "Content-MD5"
+	headerIfModifiedSince   = "If-Modified-Since"
+	headerIfMatch           = "If-Match"
+	headerIfNoneMatch       = "If-None-Match"
+	headerIfUnmodifiedSince = "If-Unmodified-Since"
+	headerRange             = "Range"
+)
+
+func (c *Client) addAuthorizationHeader(verb, url string, headers *map[string]string, auth authentication) error {
```
jhendrixMSFT: Does headers need to be a pointer?

storage/authorization.go
```diff
@@ -0,0 +1,218 @@
...
+	date := headers[headerDate]
+	if v, ok := headers[headerXmsDate]; ok {
+		if auth == sharedKey || auth == sharedKeyLite {
+			date = ""
+		} else {
+			date = v
+		}
+	}
+	var canString string
+	switch auth {
```
jhendrixMSFT: Is a default case needed to handle the case where a new auth type is added but we haven't yet implemented support?

jhendrixMSFT:
A few minor comments, looks good!

storage/authorization.go
mcardosos: Issue created

storage/authorization.go
mcardosos: Fixed

storage/authorization.go
mcardosos: Fixed

mcardosos:
Merging as soon last commit passes Travis...

PR #531, opened 2017-02-01
ACLs now work correctly with lists

mcardosos:
Found this on the way.
ACLs were working with a single permission, but they should also work with many permissions.
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -501,7 +501,7 @@ func (b BlobStorageClient) SetContainerPermissions(container string, containerPe
 // If timeout is 0 then it will not be passed to Azure
 // leaseID will only be passed to Azure if populated
 // Returns permissionResponse which is combined permissions and AccessPolicy
-func (b BlobStorageClient) GetContainerPermissions(container string, timeout int, leaseID string) (permissionResponse *ContainerPermissions, err error) {
+func (b BlobStorageClient) GetContainerPermissions(container string, timeout int, leaseID string) (*ContainerPermissions, error) {
 	params := url.Values{"restype": {"container"},
 		"comp": {"acl"}}
 
@@ -528,12 +528,11 @@ func (b BlobStorageClient) GetContainerPermissions(container string, timeout int
 		return nil, err
 	}
 
-	permissionResponse = updateContainerAccessPolicy(out, &resp.headers)
-
-	return permissionResponse, nil
+	permissionResponse := updateContainerAccessPolicy(out, &resp.headers)
+	return &permissionResponse, nil
 }
 
-func updateContainerAccessPolicy(ap AccessPolicy, headers *http.Header) *ContainerPermissions {
+func updateContainerAccessPolicy(ap AccessPolicy, headers *http.Header) ContainerPermissions {
 	// containerAccess. Blob, Container, empty
 	containerAccess := headers.Get(http.CanonicalHeaderKey(ContainerAccessHeader))
 
@@ -545,14 +544,14 @@ func updateContainerAccessPolicy(ap AccessPolicy, headers *http.Header) *Contain
 			StartTime:  policy.AccessPolicy.StartTime,
 			ExpiryTime: policy.AccessPolicy.ExpiryTime,
 		}
-		updatePermissions(&capd.CanRead, policy.AccessPolicy.Permission, "r")
-		updatePermissions(&capd.CanWrite, policy.AccessPolicy.Permission, "w")
-		updatePermissions(&capd.CanDelete, policy.AccessPolicy.Permission, "d")
+		capd.CanRead = updatePermissions(policy.AccessPolicy.Permission, "r")
+		capd.CanWrite = updatePermissions(policy.AccessPolicy.Permission, "w")
+		capd.CanDelete = updatePermissions(policy.AccessPolicy.Permission, "d")
 
 		cp.AccessPolicies = append(cp.AccessPolicies, capd)
 	}
 
-	return &cp
+	return cp
 }
 
 // DeleteContainer deletes the container with given name on the storage
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -260,7 +260,7 @@ const (
 // lease constants.
 const (
 	leaseHeaderPrefix = "x-ms-lease-"
-	leaseID           = "x-ms-lease-id"
+	headerLeaseID     = "x-ms-lease-id"
 	leaseAction       = "x-ms-lease-action"
 	leaseBreakPeriod  = "x-ms-lease-break-period"
 	leaseDuration     = "x-ms-lease-duration"
@@ -302,13 +302,6 @@ const (
 	ContainerAccessTypeContainer ContainerAccessType = "container"
 )
 
-// ContainerAccessOptions are used when setting ACLs of containers (after creation)
-type ContainerAccessOptions struct {
-	ContainerAccess ContainerAccessType
-	Timeout         int
-	LeaseID         string
-}
-
 // ContainerAccessPolicyDetails are used for SETTING container policies
 type ContainerAccessPolicyDetails struct {
 	ID         string
@@ -321,15 +314,8 @@ type ContainerAccessPolicyDetails struct {
 
 // ContainerPermissions is used when setting permissions and Access Policies for containers.
 type ContainerPermissions struct {
-	AccessOptions ContainerAccessOptions
-	AccessPolicy  ContainerAccessPolicyDetails
-}
-
-// ContainerAccessResponse is returned for the GetContainerPermissions function.
-// This contains both the permission and access policy for the container.
-type ContainerAccessResponse struct {
-	ContainerAccess ContainerAccessType
-	AccessPolicy    SignedIdentifiers
+	AccessType     ContainerAccessType
+	AccessPolicies []ContainerAccessPolicyDetails
 }
 
 // ContainerAccessHeader references header used when setting/getting container ACL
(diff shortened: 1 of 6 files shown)


storage/blob_test.go
```diff
@@ -798,24 +798,23 @@ func (s *StorageBlobSuite) TestSetBlobProperties(c *chk.C) {
...
 }
 
-func (s *StorageBlobSuite) createContainerPermissions(accessType ContainerAccessType,
-	timeout int, leaseID string, ID string, canRead bool,
-	canWrite bool, canDelete bool) ContainerPermissions {
-	perms := ContainerPermissions{}
-	perms.AccessOptions.ContainerAccess = accessType
-	perms.AccessOptions.Timeout = timeout
-	perms.AccessOptions.LeaseID = leaseID
+func (perms *ContainerPermissions) createContainerPermissions(accessType ContainerAccessType,
```
marstr: I'm struggling with the usage on this one. It seems like this is basically a constructor, but it executes on an existing struct in memory. I suspect this should be updated to:
``` go
func newContainerPermissions(accessType ContainerAccessType, ID string, start, expiry time.Time, canRead, canWrite, canDelete bool) *ContainerPermissions
```

storage/blob.go
```diff
@@ -561,13 +528,33 @@ func (b BlobStorageClient) GetContainerPermissions(container string, timeout int
...
 
-	permissionResponse = &ContainerAccessResponse{}
-	permissionResponse.AccessPolicy = out.SignedIdentifiersList
-	permissionResponse.ContainerAccess = ContainerAccessType(containerAccess)
+	permissionResponse = b.updateAccessPolicy(out, &resp.headers)
 
 	return permissionResponse, nil
 }
 
+func (b BlobStorageClient) updateAccessPolicy(ap AccessPolicy, headers *http.Header) *ContainerPermissions {
```
marstr: I'm not sure why this function is tied on to `BlobStorageClient`. It seems like this just returns a ContainerPermissions with the appropriate `AccessPolicy` enforced. Maybe this should be another ContainerPermissions constructor as mentioned below.

storage/blob.go
```diff
@@ -473,44 +459,29 @@ func (b BlobStorageClient) ContainerExists(name string) (bool, error) {
...
 
-	var resp *storageResponse
-	if accessPolicyXML != "" {
-		headers["Content-Length"] = strconv.Itoa(len(accessPolicyXML))
-		resp, err = b.client.exec(http.MethodPut, uri, headers, strings.NewReader(accessPolicyXML))
-	} else {
-		resp, err = b.client.exec(http.MethodPut, uri, headers, nil)
-	}
+	body, length, err := generateContainerACLpayload(containerPermissions.AccessPolicies)
+	headers["Content-Length"] = fmt.Sprintf("%v", length)
```
marstr: `fmt.Sprintf(...)` is not needed here. I'd update this to be
``` go
headers["Content-Length"] = string(length)
```

storage/blob.go
```diff
@@ -473,44 +459,29 @@ func (b BlobStorageClient) ContainerExists(name string) (bool, error) {
 }
 
 // SetContainerPermissions sets up container permissions as per https://msdn.microsoft.com/en-us/library/azure/dd179391.aspx
-func (b BlobStorageClient) SetContainerPermissions(container string, containerPermissions ContainerPermissions) (err error) {
+func (b BlobStorageClient) SetContainerPermissions(container string, containerPermissions ContainerPermissions, timeout int, leaseID string) (err error) {
```
marstr: [nit] I'd make `timeout` a `time.Duration` and make the call using something like this:
``` go
int(math.Ceil(timeout.Seconds()))
```

storage/blob_test.go
```diff
@@ -838,9 +838,10 @@ func (s *StorageBlobSuite) TestSetContainerPermissionsSuccessfully(c *chk.C) {
 	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
 	defer cli.deleteContainer(cnt)
 
-	perms := s.createContainerPermissions(ContainerAccessTypeBlob, 0, "", "MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTa=", true, true, true)
+	perms := ContainerPermissions{}
+	perms.createContainerPermissions(ContainerAccessTypeBlob, "GolangRocksOnAzure", time.Now(), time.Now().Add(10*time.Hour), true, true, true)
 
-	err := cli.SetContainerPermissions(cnt, perms)
+	err := cli.SetContainerPermissions(cnt, perms, 0, "")
```
marstr: Maybe we should make this into a method family, so that if one doesn't want to provided a lease or timeout, one doesn't have to specify `0` and `""`.

storage/blob_test.go
```diff
@@ -973,12 +987,12 @@ func (s *StorageBlobSuite) TestSnapshotBlobWithInvalidLease(c *chk.C) {
...
 	// generate lease.
-	_, err := cli.AcquireLease(cnt, blob, 30, "")
+	leaseID, err := cli.AcquireLease(cnt, blob, 30, "")
 	c.Assert(err, chk.IsNil)
-	c.Assert(leaseID, chk.NotNil)
+	c.Assert(leaseID, chk.Not(chk.Equals), "")
 
 	extraHeaders := map[string]string{
-		leaseID: "718e3c89-da3d-4201-b616-dd794b0bd7c1",
+		headerLeaseID: "718e3c89-da3d-4201-b616-dd794b0bd7c1",
```
marstr: We should probably make a note of what this lease is referring to. At the moment its very Magic Number like.

storage/blob.go
mcardosos: Timeout is already an `int` (or a `uint`) in a bunch of places. I also suspect some operations that support the timeout are not including it in their parameters. We can fix this, but I think it is better in another PR.

storage/blob_test.go
mcardosos: Fixed

storage/blob_test.go
```diff
@@ -825,9 +824,10 @@ func (s *StorageBlobSuite) TestSetContainerPermissionsWithTimeoutSuccessfully(c
 	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
 	defer cli.deleteContainer(cnt)
 
-	perms := s.createContainerPermissions(ContainerAccessTypeBlob, 30, "", "MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTa=", true, true, true)
+	perms := ContainerPermissions{}
+	perms.createContainerPermissions(ContainerAccessTypeBlob, "GolangRocksOnAzure", time.Now(), time.Now().Add(10*time.Hour), true, true, true)
```
mcardosos: Fixed

storage/blob_test.go
mcardosos: Replaced with a string that its obviously an invalid leaseID

storage/table.go
```diff
@@ -158,43 +156,38 @@ func (c *TableServiceClient) SetTablePermissions(table AzureTable, accessPolicy
...
-		if err := checkRespCode(resp.statusCode, []int{http.StatusNoContent}); err != nil {
-			return err
-		}
+	if err := checkRespCode(resp.statusCode, []int{http.StatusNoContent}); err != nil {
+		return err
 	}
 	return nil
 }
 
+func generateTableACLpayload(policies []TableAccessPolicy) (io.Reader, int, error) {
```
mcardosos: Fixed

storage/table_test.go
```diff
@@ -287,10 +287,11 @@ func randTable() string {
 	return string(bytes)
 }
 
-func (s *StorageBlobSuite) createTablePermissions(ID string, canRead bool, canAppend bool, canUpdate bool,
-	canDelete bool, startTime time.Time, expiryTime time.Time) TableAccessPolicy {
+func (s *StorageBlobSuite) createTablePermissions(policies *[]TableAccessPolicy, ID string,
```
mcardosos: Fixed

storage/blob.go
mcardosos: Fixed

storage/storagepolicy.go
```diff
@@ -31,38 +31,17 @@ type AccessPolicy struct {
...
-		xmlByteArray, err := ioutil.ReadAll(body)
-		if err != nil {
-			return "", err
-		}
-		accessPolicyXML = string(xmlByteArray)
-		return accessPolicyXML, nil
-	}
-
-	return "", nil
+func updatePermissions(policy *bool, permissions, permission string) {
```
jhendrixMSFT: Does policy need to be a pointer?

storage/table.go
```diff
@@ -203,42 +196,62 @@ func (c *TableServiceClient) GetTablePermissions(table AzureTable, timeout int)
...
+	err = xmlUnmarshal(resp.body, &ap.SignedIdentifiersList)
 	if err != nil {
 		return nil, err
 	}
 
-	return &out, nil
+	return updateTableAccessPolicy(ap), nil
+}
+
+func updateTableAccessPolicy(ap AccessPolicy) *[]TableAccessPolicy {
```
jhendrixMSFT: Since slices are passed by reference I don't think this needs to be a pointer.

storage/storagepolicy.go
mcardosos: Fixed

storage/table.go
mcardosos: Fixed

PR #532, opened 2017-02-02
Nit... changed client and suite names

mcardosos:
Most nitpicky PR ever.
Jokes aside, the suite name change is useful for testing.
@jhendrixMSFT @marstr


jhendrixMSFT:
While I like the change to blob storage client it is a breaking change, and since we want to refactor the blob storage APIs anyways I'm thinking we should make all those changes at once so there's only one breaking change instead of two.  Thoughts?
The test changes look good.

mcardosos:
Made the change in client name in the [refactor container](https://github.com/Azure/azure-sdk-for-go/pull/537) PR.

PR #534, opened 2017-02-04
Table refactor

mcardosos:
Finally!
PTAL
@jhendrixMSFT @marstr


mcardosos:
@alukyan I know you asked about ETags and OData types. I am fixing that in this PR. Feel free to take a look here.

storage/entity.go
```diff
@@ -0,0 +1,350 @@
...
+/* #nosec */
+const (
+	partitionKeyNode  = "PartitionKey"
+	rowKeyNode        = "RowKey"
+	etagErrorTemplate = "Etag didn't match: %v"
+)
+
+// Entity represents an entity inside an Azure table.
+type Entity struct {
+	tsc           *TableServiceClient
```
alukyan: Why there is a need to link each entity to table service client and the table? Tables can have millions of rows.
Entity is just data. What if I want to insert the same data into two tables in two different storage accounts?

storage/entity.go
```diff
@@ -0,0 +1,350 @@
...
+	return fmt.Sprintf("%s(PartitionKey='%s', RowKey='%s')", e.table.buildPath(), e.PartitionKey, e.RowKey)
+}
+
+// MarshalJSON is a custom marshaller for entity
+func (e *Entity) MarshalJSON() ([]byte, error) {
+	completeMap := map[string]interface{}{}
+	completeMap[partitionKeyNode] = e.PartitionKey
+	completeMap[rowKeyNode] = e.RowKey
+	for k, v := range e.Properties {
+		typeKey := fmt.Sprintf("%s%s", k, OdataTypeSuffix)
```
alukyan: In my opinion Sprintf is too heavy to use just to concatenate two strings (not only in Go). Simple + will work just fine for cases like this. There is also strings.Join which is also more efficient than +

storage/entity.go
```diff
@@ -0,0 +1,350 @@
...
+		switch v.(type) {
+		case []byte:
+			completeMap[typeKey] = OdataBinary
+			completeMap[k] = string(v.([]byte))
```
alukyan: Type switch in Go already performs a cast to a specific type, so your casts in 'case' braches are unnecessary
You just need to introduce typed variable like `switch t := t.(type) {`

>  If the switch declares a variable in the expression, **the variable will have the corresponding type in each clause**. It's also idiomatic to reuse the name in such cases, in effect declaring a new variable with the same name but a different type in each case.

https://golang.org/doc/effective_go.html#type_switch

mcardosos: Fixed

storage/entity.go
mcardosos: Fixed

storage/entity.go
mcardosos: What do you think about exporting `Table` field? That way it is easy to change the Table in which the entity operations are performed.
For using different storage accounts, until now the thing to do is create a new client. Adding the functionality to easily change storage accounts I think is better suited for a different PR.
About the client and table as fields, we want to follow a similar approach to what the [C# SDK](https://github.com/Azure/azure-storage-net/blob/master/Samples/GettingStarted/VisualStudioQuickStarts/DataFileStorage/Program.cs) is doing.

storage/entity.go
```diff
@@ -0,0 +1,350 @@
...
+			}
+			delete(props, k)
+		}
+	}
+
+	e.Properties = props
+	return nil
+}
+
+func getAndDelete(props *map[string]interface{}, key string) string {
```
jhendrixMSFT: Does props need to be a pointer (maps are nilable)?  Same question applies to the below functions.

storage/odata.go
```diff
@@ -0,0 +1,41 @@
+package storage
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// This consts are meant to help with opdata supported operations
```
jhendrixMSFT: Nit, I think "opdata" is supposed to be "odata" yes?

storage/entity.go
jhendrixMSFT: If the Table field is never allowed to be nil you could use the TableServiceClient from that field (this is just off the cuff, I haven't completely digested the APIs so perhaps this doesn't make sense).

storage/table.go
```diff
@@ -40,64 +33,55 @@ type TableAccessPolicy struct {
...
 
-// CreateTable creates the table given the specific
-// name. This function fails if the name is not compliant
+// Create creates the referenced table.
+// This function fails if the name is not compliant
 // with the specification or the tables already exists.
-func (c *TableServiceClient) CreateTable(table AzureTable) error {
-	uri := c.client.getEndpoint(tableServiceName, tablesURIPath, url.Values{})
+// See https://docs.microsoft.com/rest/api/storageservices/fileservices/create-table
+func (t *Table) Create(getResponse bool) error {
```
jhendrixMSFT: Does this need to be its own parameter?

storage/table.go
```diff
@@ -146,24 +137,46 @@ func (c *TableServiceClient) DeleteTable(table AzureTable) error {
 	return nil
 }
 
-// SetTablePermissions sets up table ACL permissions as per REST details https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/Set-Table-ACL
-func (c *TableServiceClient) SetTablePermissions(table AzureTable, policies []TableAccessPolicy, timeout uint) (err error) {
+// QueryEntities returns the entities in the table.
+// You can use query options defined by the OData Protocol specification.
+//
+// See: https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/query-entities
+func (t *Table) QueryEntities(odataQuery url.Values) (*EntityQueryResult, error) {
```
jhendrixMSFT: Is this analogous to ExecuteQuery() in the C# implementation?

storage/tableserviceclient.go
```diff
@@ -0,0 +1,106 @@
...
+// TableQueryResult contains the response from
+// QueryTables and QueryTablesNextResults functions.
+type TableQueryResult struct {
+	OdataMetadata string  `json:"odata.metadata"`
+	Tables        []Table `json:"value"`
+	NextLink      *string
+}
+
+// GetTableReference returns a Table object for the specified table name.
+func (t TableServiceClient) GetTableReference(name string) Table {
```
jhendrixMSFT: This should be passed by reference to align with the other receivers.

storage/table.go
mcardosos: Yes

storage/table.go
mcardosos: Actually I think a `type AcceptMetadata string` would be better, so it works better with how the API works.

mcardosos: Ok, it looks like the default for C# is minimal metadata, but users can still modify it. It is an optional parameter.

storage/entity.go
mcardosos: Fixed

storage/odata.go
mcardosos: Fixed

storage/table.go
mcardosos: Added a different solution

storage/table.go
mcardosos: Changed name

storage/tableserviceclient.go
mcardosos: Fixed

mcardosos:
Applying @jhendrixMSFT [comment](https://github.com/Azure/azure-sdk-for-go/pull/537#pullrequestreview-21309325) here as well.

mcardosos:
Fixed to use readAndCloseBody()

storage/entity.go
```diff
@@ -0,0 +1,353 @@
...
+	if !ok {
+		return fmt.Errorf(errorTemplate, "Timestamp casting error")
+	}
+	t, err := time.Parse(time.RFC3339Nano, str)
+	if err != nil {
+		return fmt.Errorf(errorTemplate, err)
+	}
+	e.TimeStamp = t
+	delete(props, "Timestamp")
+	delete(props, "Timestamp@odata.type")
```
marstr: It looks like there is a const declared below for `"@odata.type"` Perhaps we should clean this up to something like:

``` Go
const timestamp = "Timestamp" // There's probably a better name and place for this const
delete(props, timstamp)
delete(props, timestamp + OdataTypeSuffix)
```

storage/entity.go
```diff
@@ -0,0 +1,353 @@
...
+	props = assignFromMap(props, &e.OdataMetadata, "odata.metadata")
+	props = assignFromMap(props, &e.OdataType, "odata.type")
+	props = assignFromMap(props, &e.OdataID, "odata.id")
+	props = assignFromMap(props, &e.OdataEtag, "odata.etag")
+	props = assignFromMap(props, &e.OdataEditLink, "odata.editLink")
+	props = assignFromMap(props, &e.PartitionKey, partitionKeyNode)
+	props = assignFromMap(props, &e.RowKey, rowKeyNode)
+
+	// deserialize timestamp
+	str, ok := props["Timestamp"].(string)
```
marstr: Maybe this should be named `entityTimestamp` or something more representative.

storage/entity.go
```diff
@@ -0,0 +1,353 @@
...
+
+func getAndDelete(props map[string]interface{}, key string) (map[string]interface{}, string) {
+	str, ok := props[key].(string)
+	if ok {
+		delete(props, key)
+		return props, str
+
+	}
+	return props, ""
+}
```
marstr: The method above needs to be repaired.

At the moment, it looks to me like getAndDelete is supposed to test to see if a key is in the map then delete it if present. Otherwise, return the same map and an empty string.
However, what it will actually do is validate that the value stored at `props[key]` is a string, then delete it, otherwise return the unaltered map and an empty string.
I wrote up a sample to demonstrate the issue here: https://play.golang.org/p/R1PsVWH_AV

The updated function could look something like this:
``` Go
func getAndDelete(props map[string]interface{}, key string) (map[string]interface{}, interface{}) {
    val, ok := props[key]
    if ok {
        delete(props, key)
        return props, val
    }
    return props, ""
}
```

mcardosos:
@DavidObando Hello~

mcardosos:
Moved to https://github.com/Azure/azure-storage-go/pull/12

PR #536, opened 2017-02-09
Added some love for sovereign clouds

mcardosos:
- Added consts for their base URLs
- Added funcs to create clients as easy as it is for the public cloud

PTAL @marstr @jhendrixMSFT

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -15,24 +15,14 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+
+	"github.com/Azure/go-autorest/autorest/azure"
 )
 
 const (
 	// DefaultBaseURL is the domain name used for storage requests in the
 	// public cloud when a default client is created.
 	DefaultBaseURL = "core.windows.net"
-
-	// ChinaBaseURL is the domain name used for storage requests in the
-	// China cloud when a client is created.
-	ChinaBaseURL = "core.chinacloudapi.cn"
-
-	// GermanyBaseURL is the domain name used for storage requests in the
-	// Germany cloud when a client is created.
-	GermanyBaseURL = "core.cloudapi.de"
-
-	// USGovBaseURL is the domain name used for storage requests in the
-	// US government cloud when a client is created.
-	USGovBaseURL = "core.usgovcloudapi.net"
 
 	// DefaultAPIVersion is the Azure Storage API version string used when a
 	// basic client is created.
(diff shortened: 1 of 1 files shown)


storage/client.go
```diff
@@ -18,11 +18,23 @@ import (
...
-	// DefaultBaseURL is the domain name used for storage requests when a
-	// default client is created.
+	// DefaultBaseURL is the domain name used for storage requests in the
+	// public cloud when a default client is created.
 	DefaultBaseURL = "core.windows.net"
 
-	// DefaultAPIVersion is the  Azure Storage API version string used when a
+	// ChinaBaseURL is the domain name used for storage requests in the
+	// China cloud when a client is created.
+	ChinaBaseURL = "core.chinacloudapi.cn"
```
colemickens: You could just read these from go-autorest potentially: https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L21

colemickens: And then I realize that this is storage client -- maybe you don't want to add a new dependency on `go-autorest`.

mcardosos: I got the strings from there yes.

colemickens: (Sorry, yeah, I meant you could just reference them directly rather than copying, but that introduces a new dependency that's probably not worth it.)

mcardosos: I know :) No worries!

storage/client.go
```diff
@@ -131,7 +143,33 @@ func NewBasicClient(accountName, accountKey string) (Client, error) {
...
+func NewBasicClientChina(accountName, accountKey string) (Client, error) {
+	if accountName == StorageEmulatorAccountName {
+		return NewEmulatorClient()
+	}
+	return NewClient(accountName, accountKey, ChinaBaseURL, DefaultAPIVersion, defaultUseHTTPS)
+}
 
+// NewBasicClientGermany constructs a Client with given storage service name and
+// key in the Germany cloud.
+func NewBasicClientGermany(accountName, accountKey string) (Client, error) {
```
marstr: I'd prefer to not duplicate these methods in order to support different clouds. What about using something like:

``` Go
func NewBasicClient(accountName, accountKey string) {
    return NewBasicClientc(accountName, accountKey, DefaultBaseURL)
}

func NewBasicClientc(accountName, accountKey, cloud string) {
    if accountName == StorageEmulatorAccountName {
        return NewEmulatorClient()
    }
    return NewClient(accountName, accountKey, cloud, DefaultAPIVersion, defaultUseHTTPS)
}
```

Also, I'm not married to the idea of using the name `NewClientc` I'd definitely love to come up with something better!

mcardosos: Fixed (but introducing dependency on go-autorest)

PR #537, opened 2017-02-10
Refactor blob: containers

mcardosos:
Since blob is a bit of a monster, dividing the refactor in smaller PRs.
PTAL @marstr @jhendrixMSFT


storage/container.go
```diff
@@ -0,0 +1,381 @@
...
+// ContainerAccessHeader references header used when setting/getting container ACL
+const (
+	ContainerAccessHeader string = "x-ms-blob-public-access"
+)
+
+// Create creates a blob container within the storage account
+// with given name and access level. Returns error if container already exists.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179468.aspx
+func (c *Container) Create(access ContainerAccessType) error {
```
jhendrixMSFT: The access type is optional and defaults to private.  I would move this to a SetPermissions() function similar to the C# implementation so that people wanting the default behavior (presumably this is the majority) don't have to bother with this parameter.

storage/container.go
```diff
@@ -0,0 +1,381 @@
...
+		defer resp.body.Close()
+		if resp.statusCode == http.StatusOK || resp.statusCode == http.StatusNotFound {
+			return resp.statusCode == http.StatusOK, nil
+		}
+	}
+	return false, err
+}
+
+// SetPermissions sets up container permissions as per https://msdn.microsoft.com/en-us/library/azure/dd179391.aspx
+func (c *Container) SetPermissions(timeout int, leaseID string) error {
```
jhendrixMSFT: I would change this to take a BlobContainerPermissions object.

storage/blob.go
```diff
@@ -387,236 +218,9 @@ var (
...
-	defer resp.body.Close()
-
-	err = xmlUnmarshal(resp.body, &out)
-	return out, err
-}
-
 // BlobExists returns true if a blob with given name exists on the specified
 // container of the storage account.
-func (b BlobStorageClient) BlobExists(container, name string) (bool, error) {
+func (b BlobServiceClient) BlobExists(container, name string) (bool, error) {
```
jhendrixMSFT: This doesn't follow the pattern.  Is the plan to fix this up in a separate PR?

mcardosos: Yes yes

storage/container.go
mcardosos: OK, will do the same with table then./=

storage/container.go
mcardosos: Fixed

storage/container.go
mcardosos: Fixed (table still pending)

mcardosos: Fixed in table too

storage/blob.go
```diff
@@ -633,7 +237,7 @@ func (b BlobStorageClient) BlobExists(container, name string) (bool, error) {
 // specified container. This method does not create a publicly accessible URL if
 // the blob or container is private and this method does not check if the blob
 // exists.
-func (b BlobStorageClient) GetBlobURL(container, name string) string {
+func (b BlobServiceClient) GetBlobURL(container, name string) string {
```
marstr: This change got revoked before #532 got merged. Those changes should be refactored back here as well.

storage/container.go
```diff
@@ -0,0 +1,378 @@
...
+
+// ContainerProperties contains various properties of a container returned from
+// various endpoints like ListContainers.
+type ContainerProperties struct {
+	LastModified  string `xml:"Last-Modified"`
+	Etag          string `xml:"Etag"`
+	LeaseStatus   string `xml:"LeaseStatus"`
+	LeaseState    string `xml:"LeaseState"`
+	LeaseDuration string `xml:"LeaseDuration"`
+	// TODO (ahmetalpbalkan) remaining fields
```
marstr: Derp, we should probably clean up the TODO reference to a no-longer active team member.

mcardosos:
Fixed to use readAndCloseBody()

Issue #535, opened 2017-02-05
File Storage CopyFile implementation


kpfaulkner:
Azure File storage doesn't currently implement CopyFile ( https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/copy-file )

I've implemented this and will be submitting a PR soon. I have a question about the preferred approach for timeout parameters.

The timeout parameter for most of the functionality in file.go is missing. I'm assuming this is by choice and will be implemented later once people are interested in it. I just wanted to check if that was a conscious decision or was just missed?

If I add it in for CopyFile (which uses the createResource function like many others)  then it will impact said others. Would it be preferable that I just skip timeout similar to the other functions, or go through and add this?  I merely bring this up since I'm aware of some discussion about stabilising the API and didn't want to go through making non-critical API signature changes.

Thoughts?

/cc @jhendrixMSFT

jhendrixMSFT:
I wouldn't add it as a stand-alone parameter, instead I'd make it part of some type that is passed to the copy operation (take a look at the C# implementation).

kpfaulkner:
Thanks, will check that out.

mcardosos:
Hello @kpfaulkner :D
If you can do this in February we can include it in the next release.
Please make sure to submit the PR against the dev branch.
Thanks for your enthusiasm!

kpfaulkner:
The copying side of things is already working. Just need to add in the FileRequestOptions idea from @jhendrixMSFT . Hopefully sort it out on the weekend.

salameer:
Closing the PR is merged that Captain @kpfaulkner for the contribution!!!

PR #541, opened 2017-02-12
CopyFile implemented and tests

kpfaulkner:
Implement PR for https://github.com/Azure/azure-sdk-for-go/issues/535  (CopyFile).

Timeout is passed via a struct FileRequestOptions. Currently it only has a single option (Timeout) but can start expanding as more functionality added.

### DIFF
--- a/storage/file.go
+++ b/storage/file.go
@@ -35,6 +35,7 @@ type FileProperties struct {
 	MD5          string `header:"x-ms-content-md5"`
 	Type         string `header:"x-ms-content-type"`
 	CopyID       string `header:"x-ms-copy-id"`
+	CopyStatus   string `header:"x-ms-copy-status"`
 }
 
 // FileCopyState contains various properties of a file copy operation.
@@ -124,7 +125,7 @@ func (f *File) Create(maxSize uint64) error {
 		"x-ms-type":           "file",
 	}
 
-	headers, err := f.fsc.createResource(f.buildPath(), resourceFile, nil, mergeMDIntoExtraHeaders(f.Metadata, extraHeaders))
+	headers, err := f.fsc.createResource(f.buildPath(), resourceFile, nil, mergeMDIntoExtraHeaders(f.Metadata, extraHeaders), []int{http.StatusCreated})
 	if err != nil {
 		return err
 	}
@@ -148,12 +149,12 @@ func (f File) CopyFile(sourceURL string, options *FileRequestOptions) error {
 		parameters = options.getParameters()
 	}
 
-	headers, err := f.fsc.createResource(f.buildPath(), resourceFile, parameters, mergeMDIntoExtraHeaders(f.Metadata, extraHeaders))
+	headers, err := f.fsc.createResource(f.buildPath(), resourceFile, parameters, mergeMDIntoExtraHeaders(f.Metadata, extraHeaders), []int{http.StatusAccepted})
 	if err != nil {
 		return err
 	}
 
-	f.updateEtagLastModifiedAndCopyID(headers)
+	f.updateEtagLastModifiedAndCopyHeaders(headers)
 	return nil
 }
 
@@ -362,10 +363,11 @@ func (f *File) updateEtagAndLastModified(headers http.Header) {
 }
 
 // updates Etag, last modified date and x-ms-copy-id
-func (f *File) updateEtagLastModifiedAndCopyID(headers http.Header) {
+func (f *File) updateEtagLastModifiedAndCopyHeaders(headers http.Header) {
 	f.Properties.Etag = headers.Get("Etag")
 	f.Properties.LastModified = headers.Get("Last-Modified")
 	f.Properties.CopyID = headers.Get("x-ms-copy-id")
+	f.Properties.CopyStatus = headers.Get("x-ms-copy-status")
 }
 
 // updates file properties from the specified HTTP header
--- a/storage/file.go
+++ b/storage/file.go
@@ -14,12 +14,13 @@ const oneTB = uint64(1099511627776)
 
 // File represents a file on a share.
 type File struct {
-	fsc        *FileServiceClient
-	Metadata   map[string]string
-	Name       string `xml:"Name"`
-	parent     *Directory
-	Properties FileProperties `xml:"Properties"`
-	share      *Share
+	fsc                *FileServiceClient
+	Metadata           map[string]string
+	Name               string `xml:"Name"`
+	parent             *Directory
+	Properties         FileProperties `xml:"Properties"`
+	share              *Share
+	FileCopyProperties FileCopyState
 }
 
 // FileProperties contains various properties of a file.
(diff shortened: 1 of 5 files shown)


kpfaulkner:
Can anyone clarify the issue I'm hitting with the PR?

The command "go test -v ./storage/... -check.v" exited with 1.
test -z "$(golint ./storage/... | tee /dev/stderr)"


Thanks

Ken

storage/file_test.go
```diff
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"azure-sdk-for-go/storage"
```
mcardosos: Get rid of this line for Travis to pass :)

kpfaulkner: Ahhh thanks!

Out of interest, how did you figure that out? Looking at the Travis output I didn't see it mentioned/hinted anywhere. Did you just read the code and see the issue or should I be looking somewhere else for clues?

Thanks

Ken

mcardosos: Everything in Travis. Look at the lines before the one in red (in this case 344 - 350). That is the failed command output.

kpfaulkner:
As I hit "send" on my comment I'd realised that I'd been looking at the
Travis summary logs and not the raw logs. Doh :)

storage/file.go
```diff
@@ -51,6 +51,24 @@ type FileStream struct {
 	ContentMD5 string
 }
 
+// FileRequestOptions will be passed to misc file operations.
+// Currently just Timeout but will expand.
+type FileRequestOptions struct {
+	Timeout uint
```
marstr: I'd update this to either be a time.Duration or add documentation that states which unit this is expressed in. (i.e. Timeout is a number of seconds.)

storage/file.go
```diff
@@ -114,6 +132,24 @@ func (f *File) Create(maxSize uint64) error {
 	return nil
 }
 
+// CopyFile operation copied a file/blob from the sourceURL to the path provided.
+//
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/copy-file
+func (f File) CopyFile(sourceURL string, options *FileRequestOptions, metadata map[string]string) error {
```
marstr: A `nil` guard should be added for `options` here.

marstr: Also, the `metadata` argument goes unused in this method. Are we sure it should be included in the header?

storage/file.go
mcardosos: +1 on adding documentation. Not on using `time.Duration` because all other funcs use `int` or `uint`.

storage/file.go
```diff
@@ -114,6 +132,24 @@ func (f *File) Create(maxSize uint64) error {
...
+	extraHeaders := map[string]string{
+		"x-ms-type":        "file",
+		"x-ms-copy-source": sourceURL,
+	}
+
+	headers, err := f.fsc.createResource(f.buildPath(), resourceFile, options.getParameters(), mergeMDIntoExtraHeaders(f.Metadata, extraHeaders))
```
mcardosos: Success code is 202, not 201 (201 is the one implemented in `createResource`)

storage/file.go
```diff
@@ -114,6 +132,24 @@ func (f *File) Create(maxSize uint64) error {
...
+	if err != nil {
+		return err
+	}
+
+	f.updateEtagAndLastModified(headers)
+	return nil
```
mcardosos: Returning the data in `x-ms-copy-id` and `x-ms-copy-id` can be useful for other operations too.

storage/file_test.go
```diff
@@ -272,3 +272,120 @@ func newByteStream(count uint64) (io.Reader, string) {
...
+	exists, err = file.Exists()
+	c.Assert(err, chk.IsNil)
+	c.Assert(exists, chk.Equals, true)
+
+	otherFile := dir2.GetFileReference("someother.file")
+
+	options := FileRequestOptions{}
+	options.Timeout = 60
+
+	// copy the file, no timeout parameter
```
mcardosos: With timeout parameter.

kpfaulkner:
Fought with some merges. Will sort this out tonight :)

storage/file.go
kpfaulkner: commented, leaving as uint to be consistent with other funcs. Not pushed yet.

storage/file.go
kpfaulkner: Doh. added

kpfaulkner: Agreed. This was a bad mistake by me. The splitting of file.go out to directory/share/file and services jumbled up a bunch of stuff. Fixing those :)

storage/file.go
kpfaulkner: Returning x-ms-copy-id.... but you listed that twice. Was there a second you were thinking of?   Status?

storage/file_test.go
kpfaulkner: removed comment :)

storage/file.go
kpfaulkner: Have a question around this one.  I modified createResource and createResourceNoClose to accept the url.Values as parameters since all the various create functions can technically pass in a timeout. They don't do it YET but the spec says they can so was thinking they might be added later. 

I was picturing that copying a file is creating a new file hence reusing the createResource function. Given the HTTP return code is different (201 vs 202) is there a preference to how this is handled?

I could create another method similar to createFunction that calls createFunctionNoClose but simply checks for 202 instead of 201 OR I could make the expected status code to just be a parameter.

Either works... one is duplicated code (yuck) and the other is parameter with expected status (still maybe yuck).

I know it's a small/simple problem, but just wondering if there is a preferred approach here?

storage/file.go
mcardosos: Oops, it was `x-ms-copy-status`, sorry

storage/file.go
mcardosos: I think making it a parameter is OK. Thoughts @jhendrixMSFT ?

jhendrixMSFT: Could it be the case that there's more than one expected status code?  If so then perhaps adding a function argument that takes a lamba that performs the status code validation would allow for flexibility.

kpfaulkner: I'll go by that recommendation if it's preferred, but was thinking that if we want to handle multiple status codes, why not just pass an array/slice? createResource already simply creates an array of 1 (StatusCreated), couldn't we just make the parameter an array to begin with?

Merely curious.

jhendrixMSFT: Yes that would probably be better considering we just check that the response code is within a slice of valid values (I didn't look at the implementation when I commented previously, shame on me).  I was thinking the logic might be complex than this.

kpfaulkner:
Cool, I'll do that change then push later this morning.

Thanks

salameer:
@jhendrixMSFT will take a last  look

THanks FOR THE CONTRIBUTION @kpfaulkner

storage/file.go
```diff
@@ -34,6 +35,8 @@ type FileProperties struct {
 	Length       uint64 `xml:"Content-Length"`
 	MD5          string `header:"x-ms-content-md5"`
 	Type         string `header:"x-ms-content-type"`
+	CopyID       string `header:"x-ms-copy-id"`
```
jhendrixMSFT: Instead of adding these fields to the FileProperties struct we should use FileCopyState instead, adding it as a field to the File struct.

jhendrixMSFT: I see that FileCopyState was added, just need to use its fields instead.

jhendrixMSFT:
Thanks @kpfaulkner this is great.

Issue #404, opened 2016-09-27
Connection Reuse issue for PUT operations


bcc:
The calls that make use of an HTTP PUT are not properly closing the body, which is preventing reuse of HTTPS connections. This badly hurts performance, and can cause ephemeral connection exhaustion on the client. This is most obvious if you call something like PutBlob, SetBlobProperties or CopyBlob a lot of times. 

I've bodged around this for now by adding `io.Copy(ioutil.Discard, resp.body)` after the operations that have been a problem for me as in https://github.com/bcc/azure-sdk-for-go/commit/df611657ca0c940d8f1c7d79959e3afa7f6aff38

It strikes me that this isn't a great solution though - none of the PUT operations on blobs actually do anything with the response body, and a skim through the other non-blob functions suggests that might be universally true. If that's the case, it might be better to fix this in client.exec by clearing out the body if verb == PUT, but I'm not sure if that might be a mistake later on.

marstr:
@salameer This needs to be re-triaged.

marstr:
I've spent some time this morning investigating, and I can confirm that this is a real issue. The root of the issue can be seen here: [godoc http.Response](https://golang.org/src/net/http/response.go) which says:
> The default HTTP client's Transport does not attempt to reuse HTTP/1.0 or HTTP/1.1 TCP connections ("keep-alive") unless the Body is read to completion and is closed.

For the vast majority of the methods in Azure/azure-sdk-for-go/storage, we call `resp.body.Close()` but never read the body; thus we complete only half of the contract that would result in connection reuse. 

As @bcc noted above, it would be _easiest_ to fix this in the client, by universally reading the response body then closing. We could keep the same contract by creating a ReadCloser over a buffer that we had copied the body into. However, the more thorough fix is to dig through all Storage operations and ensure the Body is read before it is closed where appropriate. I much prefer the latter solution, and believe it is worth the extra cost. I'll start working on a PR to fix this issue today.

marstr:
This issue has now been fixed in storage (as mentioned in the original issue) and a PR is out to fix it in Azure/go-autorest.

Azure/go-autorest#118

I'll close this issue out for now. If the issue you were experiencing wasn't resolved, feel free to re-open.

meirf:
@bcc did this actually fix the issue for you?

jhendrixMSFT:
@meirf which version of the SDK are you using?  The latest is v15.0.0.  Can you also paste the APIs you're using please?

[...]

PR #544, opened 2017-02-14
Fixing #404, which prevented connection reuse.

marstr:


storage/client.go
```diff
@@ -424,9 +424,9 @@ func (c Client) execInternalJSON(verb, url string, headers map[string]string, bo
 	return respToRet, nil
 }
 
-func readResponseBody(resp *http.Response) ([]byte, error) {
-	defer resp.Body.Close()
-	out, err := ioutil.ReadAll(resp.Body)
+func readBody(body io.ReadCloser) ([]byte, error) {
```
mcardosos: I would rename to `readAndCloseBody` to make it more clear. Maybe add a link to the godoc in which it is explained why this must be like that.
Other that that LGTM.

jhendrixMSFT:
Does any of our generated code have this bug?

mcardosos:
Looks like yes :I

marstr:
I'll investigate what it'll take to fix the generated stuff before I close #404. Though depending on the fix, it may be that it makes more sense to file a bug on the generator or in Azure/go-autorest instead of tying this issue.

jhendrixMSFT:
Agreed fixing the generator should be tracked with a separate issue.

mcardosos:
I think we'll need to add a `ByReading()` RespondDecorator. I think it is not needed in funcs that use `ByUnmarshallingJSON()`.

marstr:
Agreed, I took particular care to never apply the `readBody()` (now `readAndCloseBody()`) to any function that has its own use for the response body.

marstr:
I just tinkered around a little (and copied my experiment [here.](https://play.golang.org/p/jnFf7RhqCu)) I verified that there is no harm in reading something that has already been read to completion, but hasn't yet been closed. For that reason, I believe the correct fix for other parts of the repository is to update the `ByClosing()` decorator to first read and discard the contents of `response.Body`.

marstr:
Once I actually started playing around, I decided it may fit the pattern better to create a new decorator called `ByDiscardingBody()`. Thoughts, @mcardosos ?

mcardosos:
I like it. I would prefer not to modify `ByClosing()`.

Issue #518, opened 2017-01-26
Operations on Account missing


mcardosos:
The ones in [here](https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/operations-on-the-account--table-service-)

mcardosos:
Also the ones in [file service](https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/operations-on-the-account--file-service-)

mcardosos:
Ok, all services are missing operations.

marstr:
For further conversation around this topic, please see the issue linked above.

jhendrixMSFT:
These will be added when we create all the swaggers for dataplane, we will not add it to the existing storage package.

PR #545, opened 2017-02-14
Service properties

mcardosos:
Related to https://github.com/Azure/azure-sdk-for-go/issues/518
There are still some funcs missing for all services. Started with get and set service properties.
PTAL @jhendrixMSFT @marstr

### DIFF
--- a/storage/storageservice.go
+++ b/storage/storageservice.go
@@ -112,7 +112,7 @@ func (c Client) setServiceProperties(props ServiceProperties, service string, au
 	if err != nil {
 		return err
 	}
-	defer resp.body.Close()
+	defer readAndCloseBody(resp.body)
 
 	return checkRespCode(resp.statusCode, []int{http.StatusAccepted})
 }
--- a/storage/storageservice.go
+++ b/storage/storageservice.go
@@ -1,0 +1,118 @@
+package storage
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// ServiceProperties represents the storage account service properties
+type ServiceProperties struct {
+	Logging       *Logging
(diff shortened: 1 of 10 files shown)


storage/storageservice.go
```diff
@@ -0,0 +1,118 @@
...
+func (c Client) setServiceProperties(props ServiceProperties, service string, auth authentication) error {
+	query := url.Values{
+		"restype": {"service"},
+		"comp":    {"properties"},
+	}
+	uri := c.getEndpoint(service, "", query)
+
+	// Ideally, StorageServiceProperties would be the output struct
+	// This is to avoid golint stuttering, while generating the correct XML
+	type StorageServiceProperties struct {
```
jhendrixMSFT: Not clear to me why this is needed with respect to stuttering?

mcardosos: Because if this struct was outside the func, if someone used it, it would look like `storage.StorageServiceProperties`. Also, golint does not pass.

jhendrixMSFT: Ok that part makes sense.  But why can't we pass props to xmlMarshal()?

mcardosos: Because of instead of generating
```xml
<StorageServiceProperties>
...
</StorageServiceProperties>
```
it generates
```xml
<ServiceProperties>
...
</ServiceProperties>
```

jhendrixMSFT: Ah ok makes sense, thanks for the explanation.

jhendrixMSFT:
Looks good!

mcardosos:
Solved conflicts, feel free to merge @marstr @jhendrixMSFT

PR #546, opened 2017-02-15
Add support to generate SAS URI for entire containers

lirao:
We are using this as a dependency in Yammer's fork of hashicorp/vault which includes support for Azure Blob Storage SAS URI generation (I'll send them a PR once I get the OSS approval)

Slightly modifies GetBlobSASURI behavior to generate a URI for the entire container instead of just a single blob, if blob name isn't specified.

-- Rao Li (raol@microsoft.com)

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -234,14 +234,16 @@ func (b BlobStorageClient) BlobExists(container, name string) (bool, error) {
 }
 
 // GetBlobURL gets the canonical URL to the blob with the specified name in the
-// specified container. This method does not create a publicly accessible URL if
-// the blob or container is private and this method does not check if the blob
-// exists.
+// specified container. If name is not specified, the canonical URL for the entire
+// container is obtained.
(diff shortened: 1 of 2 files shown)


salameer:
Hi @lirao, please join the Microsoft Github org to avoid getting the CLA warning, or else you'll have to sign for us to move forward here

lirao:
Joined, how do I get the bot to verify me?

marstr:
@msftclas Could you take another look?

marstr:
I'm going to close and re-open to attempt to get @msftclas to take another look.

marstr:
Looks like there is a missing function `pathForContainer()` From CI:
> storage/blob.go:1023: undefined: pathForContainer
   FAIL	github.com/Azure/azure-sdk-for-go/storage [build failed]

lirao:
Wow looks like there was a major refactor today, I'll rebase/merge it

PR #547, opened 2017-02-15
Gen.go update

mcardosos:
This works with the cowboy's latest changes.
PTAL @marstr @jhendrixMSFT

### DIFF
--- a/Gododir/gen.go
+++ b/Gododir/gen.go
@@ -364,14 +364,14 @@ func initAndAddService(service *service, inputPrefix, plane string) {
 	}
 	packages := append(service.Packages, service.Name)
 	service.TaskName = fmt.Sprintf("%s>%s", plane, strings.Join(packages, ">"))
-	service.Fullname = fmt.Sprintf("%s/%s", plane, strings.Join(packages, "/"))
+	service.Fullname = filepath.Join(plane, strings.Join(packages, "/"))
 	if service.Input == "" {
-		service.Input = fmt.Sprintf("%s%s/%s/swagger/%s", inputPrefix, strings.Join(packages, "/"), service.Version, service.Swagger)
+		service.Input = filepath.Join(inputPrefix+strings.Join(packages, "/"), service.Version, "swagger", service.Swagger)
 	} else {
-		service.Input = fmt.Sprintf("%s%s/%s/swagger/%s", inputPrefix, service.Input, service.Version, service.Swagger)
+		service.Input = filepath.Join(inputPrefix+service.Input, service.Version, "swagger", service.Swagger)
 	}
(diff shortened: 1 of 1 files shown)


Gododir/gen.go
```diff
@@ -404,14 +414,16 @@ func generate(service *service) {
 	fmt.Printf("Generating %s...\n\n", service.Fullname)
 	delete(service)
 
-	autorest := exec.Command(fmt.Sprintf("%s/autorest/src/core/AutoRest/bin/Debug/net451/win7-x64/autorest", autorestDir),
+	autorest := exec.Command("gulp",
```
marstr: Given that this is pretty new stuff, and some of our computers (okay, well mine) haven't had gulp setup yet. We may want to add in some guard rails like this:
[exec.LookPath(string)](https://godoc.org/os/exec#LookPath)

Gododir/gen.go
```diff
@@ -404,14 +414,16 @@ func generate(service *service) {
...
+		"autorest",
 		"-Input", fmt.Sprintf("%s/azure-rest-api-specs/%s.json", swaggersDir, service.Input),
 		"-CodeGenerator", "Go",
 		"-Header", "MICROSOFT_APACHE",
 		"-Namespace", service.Name,
 		"-OutputDirectory", service.Output,
 		"-Modeler", "Swagger",
 		"-pv", sdkVersion)
+	autorest.Dir = fmt.Sprintf("%s/autorest", autorestDir)
```
marstr: We're all cross-platform like now! How about we update it to:

``` Go
autorest.Dir = filepath.Join(autores.Dir, "autorest")
```

[filepath.Join(string...)](https://golang.org/pkg/path/filepath/#Join)

Gododir/gen.go
```diff
@@ -404,15 +415,22 @@ func generate(service *service) {
...
 
-	autorest := exec.Command(fmt.Sprintf("%s/autorest/src/core/AutoRest/bin/Debug/net451/win7-x64/autorest", autorestDir),
+	_, err := exec.LookPath("gulp")
+	if err != nil {
+		panic("You need gulp (and some other tools) to run AutoRest. See https://github.com/Azure/autorest/pull/1827")
+	}
+
+	autorest := exec.Command("gulp",
+		"autorest",
 		"-Input", fmt.Sprintf("%s/azure-rest-api-specs/%s.json", swaggersDir, service.Input),
```
jhendrixMSFT: Another candidate for filepath.Join.

mcardosos:
Wait, it is not working as I expected yet :I

Gododir/gen.go
mcardosos: Fixed (and used filepath.Join in some other places too)

Gododir/gen.go
mcardosos: Fixed

mcardosos:
It works!

Issue #548, opened 2017-02-16
Unable to acquire an infinite blob lease


pmcatominey:
There as a check added in #430 which only sets the lease duration header if the provided value is > 0, this prevents infinite leases from being acquired as the request is rejected for missing the header.

marstr:
Nice catch.

Relevant documentation can be found here: [Microsoft Docs, storageservices/fileservices/lease-blob](https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/lease-blob)

marstr:
PR #550 solves this issue.

PR #550, opened 2017-02-16
Addressing issue #548

marstr:
Updating `blob.AcquireLease(...)` to ensure that leaseTimeInSeconds is
submitted within the bounds of what the Azure REST API expects. I
decided clamping the value to the min/max acceptable value was more user
friendly than throwing an error without adding too much complexity.

The real alternative to what I did was to simply remove the check
all-together and just pass through the value to Azure. The reason I
didn't do that, is that I presume Azure would adopt the same behavior I
enumerated. Having this operation done client side will allow folks who
are debugging to see the operation.


marstr:
It should now be safe to close #548

PR #556, opened 2017-02-22
Moving Storage to an Independent Repository

marstr:
Go 1.8 was released on Thursday, February 16th. Many of our partners
have adopted and we need to ensure that we support the most recent
version.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "storage"]
```
mcardosos: Should we get rid of funcs related to storage in gen.go?

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
```
mcardosos: You will need to get rid of some lines here

.gitmodules
marstr: yes :) I'll get on that.

.travis.yml
marstr: Done.

.gitmodules
marstr: Done

marstr:
Hey @kpfaulkner, I wanted to give you a heads up about this PR and the new repository. If you have a little time, I'd love for you to do a once-over on [Azure/azure-storage-go ](https://github.com/Azure/azure-storage-go) to ensure that there isn't something critical that we're missing.

kpfaulkner:
Hi

Will do! Glad to see a separate repo! :)

Ken

kpfaulkner:
Definitely thumbs up from me.
Have ported one of my Azure/Go projects to azure-storage-go with minimal hassles. Admittedly most of the porting was due to that I hadn't upgraded that particular project since the whole storage split to separate structs/files for containers/blobs etc. 

I like the idea of having all the storage functionality separate to the whole management side of things.

kpfaulkner:
Any idea when azure-storage-go would become the official storage repo and the storage dir in azure-sdk-for-go would be deprecated?

Or it's still early stages?

marstr:
Hey Ken, we're targeting Version 9.0 in this repository (so end of the month.) I've been holding out on pulling in this request until closer to when we plan on releasing this change because there is a little chaffing with sub-modules when you switch between branches that have them and don't. However, it may be time to just pull the trigger. It should make clear to folks sending PRs and filing issues that this repo is moved.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "storage"]
 	path = storage
 	url = https://github.com/Azure/azure-storage-go
+	branch = v0.1.0
```
mcardosos: Is branch correct? Cannot glide take tags?

mcardosos: Oh, lol, this is not glide. But still, the question is the same! Cannot it be tag instead of branch?

marstr: The name `branch` is misleading here. See this thread:
http://stackoverflow.com/questions/19986075/git-submodule-to-track-remote-branch

Alternatively, try cloning my fork to ensure it works :D

marstr: @mcardosos and I just tinkered with this offline and can confirm go get has the correct behavior.

Issue #555, opened 2017-02-22
Add telemetry for ASM


mcardosos:

marstr:
This was solved by the User-Agent PR above.

PR #557, opened 2017-02-23
More complete User-Agent for ASM

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/555
PTAL @jhendrixMSFT @marstr

### DIFF
--- a/Gododir/gen.go
+++ b/Gododir/gen.go
@@ -386,7 +386,7 @@ func initAndAddService(service *service, inputPrefix, plane string) {
 }
 
 func tasks(p *do.Project) {
-	p.Task("default", do.S{"setvars", "generate:all", "storage"}, nil)
+	p.Task("default", do.S{"setvars", "generate:all", "storage", "management"}, nil)
 	p.Task("setvars", nil, setVars)
 	p.Use("generate", generateTasks)
 	p.Use("gofmt", formatTasks)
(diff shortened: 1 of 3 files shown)


management/version.go
```diff
@@ -0,0 +1,5 @@
+package management
+
+var (
+	sdkVersion = "8.0.0-beta"
```
jhendrixMSFT: Any changes in the update script we need to make so this is updated automatically?  I know that ASM doesn't change much so it's not really high on the priority list.

jhendrixMSFT: Sign maybe I should read the review closer.

jhendrixMSFT:
LGTM.  On a side-note I know @marstr was looking at updating version reporting to avoid allocations so we should apply that same pattern here at some point.

mcardosos:
OK, waiting for @marstr then.

management/version.go
marstr: Ugh, ideally I would have wanted to skip reporting these broken version numbers. However, this is a chicken & egg problem. When I refactor out management into its own repository, I'll have to remember to change the UserAgent string to "Azure-SDK-For-Go-ASM/<updatedVersionNumber>" so that we can differentiate between this version and the refactored version.

Gododir/gen.go
```diff
@@ -506,15 +507,23 @@ func vet(service *service) {
 }
 
 func storageVersion(c *do.Context) {
-	versionFile := "storage/version.go"
+	version("storage")
```
marstr: I guess it makes sense to have this change for now. However, once everything has been submoduled out, managementVersion and storageVersion should no longer be present.

mcardosos: Should not forget

Issue #494, opened 2016-12-19
Running the Management example "example.go" throws an internal server error


salameer:
When I run the "example.go" in the management\examples\ folder I get the following output:

Create client
Create hosted service
Create storage account
Create virtual machine
Deploy
panic: Error response from Azure. Code: InternalError, Message: The server encountered an internal error. Please retry the request.

goroutine 1 [running]:
panic(0x25e440, 0xc42037b560)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
main.main()
	/Users/Samer/Documents/workspace/TestingGo1/src/MyProgram/main.go:69 +0x10f2

as you can see the resources get created but when it seems to fail on "Create Deployment" with a message that's really descriptive

marstr:
It seems that the service itself is broken. I do not think the suggested remedy from the error message is appropriate, just retrying into perpetuity.

Having a non-functional example isn't really appropriate. I believe we should close this issue by deleting the example script and throwing an item on the backlog to fix this at some-point.

marstr:
Just spoke offline with @mcardosos, she agrees with this assessment.

PR #558, opened 2017-02-23
Forgot to get rid of these lines in readme

mcardosos:
@marstr @jhendrixMSFT


marstr:
This is follow up from #494

PR #563, opened 2017-03-02
Fix table entities error checking

johnglover:
There is a nasty bug in the current TableService code that results in some errors being silently dropped.

The following functions are only checking the response code if a call to `c.execTable` returns an error:

* `InsertEntity`
* `UpdateEntity`
* `InsertOrReplaceEntity`
* `InsertOrMergeEntity`

In some cases calls to `c.execTable` can fail without returning errors. In particular, it calls `client.execInternalJSON`,  which does return an error for bad status codes with no response body (https://github.com/Azure/azure-sdk-for-go/blob/master/storage/client.go#L414). However if there is some response body, it then unmarshals it, and only returns an error if this unmarshalling fails (https://github.com/Azure/azure-sdk-for-go/blob/master/storage/client.go#L420).

It is unclear if this behaviour (for `execInternalJSON`) is intentional or not, so this PR just makes sure that the response code is checked at all times. It also adds a new test verifying that trying to insert entities into nonexistent tables should fail as expected.

### DIFF
--- a/storage/table_entities.go
+++ b/storage/table_entities.go
@@ -124,11 +124,12 @@ func (c *TableServiceClient) QueryTableEntities(tableName AzureTable, previousCo
 // The function fails if there is an entity with the same
 // PartitionKey and RowKey in the table.
 func (c *TableServiceClient) InsertEntity(table AzureTable, entity TableEntity) error {
-	if sc, err := c.execTable(table, entity, false, http.MethodPost); err != nil {
-		return checkRespCode(sc, []int{http.StatusCreated})
+	sc, err := c.execTable(table, entity, false, http.MethodPost)
+	if err != nil {
+		return err
 	}
 
-	return nil
+	return checkRespCode(sc, []int{http.StatusCreated})
 }
 
 func (c *TableServiceClient) execTable(table AzureTable, entity TableEntity, specifyKeysInURL bool, method string) (int, error) {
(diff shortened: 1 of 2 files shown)


storage/table_test.go
```diff
@@ -77,6 +77,16 @@ func (s *StorageBlobSuite) Test_InsertEntities(c *chk.C) {
 	}
 }
 
+func (s *StorageBlobSuite) Test_InsertEntitiesRandomTableFails(c *chk.C) {
+	cli := getTableClient(c)
+
+	tn := AzureTable(randTable())
+
+	ce := &CustomEntity{Name: "Luke", Surname: "Skywalker", Number: 1543, PKey: "pkey", RKey: "5"}
```
marstr: ^ Very nice name choice.

storage/table_entities.go
```diff
@@ -162,21 +163,25 @@ func (c *TableServiceClient) execTable(table AzureTable, entity TableEntity, spe
...
 }
 
 // MergeEntity merges the contents of an entity with the
 // one passed as parameter.
 // The function fails if there is no entity
 // with the same PartitionKey and RowKey in the table.
 func (c *TableServiceClient) MergeEntity(table AzureTable, entity TableEntity) error {
-	if sc, err := c.execTable(table, entity, true, "MERGE"); err != nil {
-		return checkRespCode(sc, []int{http.StatusNoContent})
+	sc, err := c.execTable(table, entity, true, "MERGE")
```
marstr: I'd like to pull `"MERGE"` out as a const. However, I suspect that this isn't the right package to host its definition. I'll file an issue to track this follow up item.

mcardosos: I would like the `http` package to include it as a const 😢

mcardosos:
Sure, I agree with merging here and then cherry-pick.

mcardosos:
BTW, the repo for storage is now https://github.com/Azure/azure-storage-go :)

johnglover:
@mcardosos: thanks, I wasn't aware of the separate storage repo.

storage/table_test.go
johnglover: Taken from an existing test case so I can't take credit :)

mcardosos:
Fixed in the other repo too!

Issue #559, opened 2017-02-23
Code generated by AutoRest has race conditions


katcipis:
Hi,

The following code generated by AutoRest:

```
// UserAgent returns the UserAgent string to use when sending http.Requests.
func UserAgent() string {
	if userAgent == "" {
		userAgent = fmt.Sprintf(userAgentFormat, Version(), "resources", "2016-09-01")
	}
	return userAgent
}
```

Is producing a race condition detected by my tests as can be seen here:

```
WARNING: DATA RACE
Read at 0x000000ac6770 by goroutine 20:
  github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources.UserAgent()
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources/version.go:43 +0x6d
  github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources.NewWithBaseURI()
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources/client.go:53 +0x69
  github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources.NewGroupsClientWithBaseURI()
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources/groups.go:41 +0x9c
  github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources.NewGroupsClient()
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/vendor/github.com/Azure/azure-sdk-for-go/arm/resources/resources/groups.go:36 +0x97
  github.com/NeowayLabs/klb/tests/lib/azure/fixture.NewResourceGroup()
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/lib/azure/fixture/resourcegroup.go:27 +0x75
      /home/katcipis/workspace/go/src/github.com/NeowayLabs/klb/tests/lib/azure/fixture/resourcegroup.go:27 +0x75
[...]

Issue #517, opened 2017-01-26
Versioning Proposal: Customer Feedback Requested


marstr:
We've been thinking a lot about improvements that we could make to improve upon our current package management story. There are three problems in particular that have jumped out to us, which think need to be tackled head on in order to improve our story:

1. The version numbers need to actually conform to [semver 2.0](http://semver.org/spec/v2.0.0.html) in order to integrate best with glide and other tools. The biggest change here is that pre-release status should be indicated by major-version 0, not a tag at the end. In the future, there could be “-rc” or “-beta” tags that mean specifically this is a test version of a patch or new revision.
2. At the moment, too many packages are bunched up into a single repository. This makes acquisition convenient, but blows up version numbers and doesn't communicate changes' scope or severity well. 
3. As pointed out in other threads, a lot of swaggers/packages are stable enough to consider release quality. Our support of those shouldn't be questioned because they happen to be adjacent to new swaggers that are unstable.
4. Without intervention, releases will continue to be at a relatively slow monthly cadence with a single version number across all packages.

We want to fix all of those problems ASAP. However, your feedback is important to us. We want to ensure that we don't introduce new worse problems while we're fixing some remittable issues.

Here are the changes that we'd like you all to consider: (Please note, the numbering here doesn't line up with the problems above.)

1.  Does it make sense to split up this repository into several smaller ones?
     If so, there are a couple of options, and none of them really require breaking changes. With all of the proposed changes below git submodules could be used to keep this repository alive and targeting each of the smaller repositories. In this way, there need not be breaking changes.
     - Storage, management, and each ARM package each get their own repositories. Their paths would be:
       GitHub.com/Azure/azure-sdk-for-go-arm-[package name]
       GitHub.com/Azure/azure-sdk-for-go-storage
       GitHub.com/Azure/azure-sdk-for-go-management
    - Storage, management, and ARM each get their own repositories. Their paths would be:
       GitHub.com/Azure/azure-sdk-for-go-arm
       GitHub.com/Azure/azure-sdk-for-go-storage
       GitHub.com/Azure/azure-sdk-for-go-management
    - We attempt to match the Azure portal’s groupings, and have families packaged together. Some repository paths may be:
       GitHub.com/Azure/azure-sdk-for-go-storage
       GitHub.com/Azure/azure-sdk-for-go-arm-networking
       GitHub.com/Azure/azure-sdk-for-go-arm-compute
       Etc…
[...]

PR #569, opened 2017-03-09
V9

marstr:
For Version 9.0.0-beta, there are a bunch of breaking changes that we've been holding off on until we could give folks more of a proper heads-up.

So, the goal of this PR is to capture a running list of categories of changes (breaking and otherwise) that will be present in v9 and the packages impacted by them. CHANGELOG will contain a summary of the changes as well, but could never hope to be as precise as the diffs in a PR.

I'll be keeping the v9 branch in my fork until this is merged in. So if you have questions about whether or not a package you're interested in will be impacted, I recommend cloning my fork then doing a diff of a specific package like so:

``` sh
go get -u github.com/marstr/azure-sdk-for-go
cd $GOPATH/src/github.com/marstr/azure-sdk-for-go
git checkout v9
git diff v8.1.1-beta v9 -- arm/<package name>
```

Note, there will be more revisions to this PR through-out March until we release. The current status is intended to demonstrate name changes, etc in packages where the Swaggers were not updated.


marstr:
First Major Change to note:
Storage has been moved into its own repository. We've added a submodule so that this isn't a breaking change, but consuming it through Azure/azure-sdk-for-go is deprecated. We encourage you to directly consume it via [Azure/azure-storage-go](https://github.com/Azure/azure-storage-go)

marstr:
There are breaking name changes in the following packages:
- arm/authorization
- arm/batch

(I'll edit this list to add the full list of impacted packages as I discover them. Please check back here before you adopt v9.0.0-beta)

marstr:
#559 is fixed in this PR. Version number and UserAgent are now handled by the generator instead of being created at runtime. This may have to change if we want to start capturing Go Version or OS as part of the UserAgent field.

randomcamel:
- Is there a way to manage compatibility issues among the newly-separated libraries? How do we know what versions are safe to use together?
  -  This is a pretty heavy change: is there a doc somewhere with the backstory on what problems it's solving and how this solution emerged?

marstr:
Howdy @randomcamel,

Very valid concern! For some context around why we're pulling storage into a new repository, you can read #517. At the end of the day, what it really comes down to is protecting our storage/ASM users from the frequent massive PRs and Major Version bumps that come along with the auto-generated ARM packages.

In terms of compatibility, our decision was made easier by the fact that they are completely disparate packages. Other than spinning up new storage accounts, there isn't really any overlap between the storage and ARM packages. All versions are compatible with one another. We'll eventually be making a similar change to the "management" (ASM) packages.

As to the other changes in this PR, they are the result of an aggregate monthly publish from swagger files in the Azure/azure-rest-api-specs repository. Getting as many customers away from these bloated PRs as possible is important to us. Long term, we're working on a continuous deployment story for the ARM packages.

I hope that answers your questions! Feel free to follow up or ask more.

-Martin.

pmcatominey:
While the storage and arm packages are wholly separate, they do both depend on the `Azure/go-autorest` package. The only concern there would be if a breaking change was made to autorest and only one of arm or storage had been updated.

arm/servicebus/models.go
```diff
@@ -453,29 +375,24 @@ func (client TopicListResult) TopicListResultPreparer() (*http.Request, error) {
...
-	EntityAvailabilityStatus            EntityAvailabilityStatus `json:"entityAvailabilityStatus ,omitempty"`
-	CreatedAt                           *date.Time               `json:"createdAt,omitempty"`
-	CountDetails                        *MessageCountDetails     `json:"countDetails,omitempty"`
-	DefaultMessageTimeToLive            *string                  `json:"defaultMessageTimeToLive,omitempty"`
-	DuplicateDetectionHistoryTimeWindow *string                  `json:"duplicateDetectionHistoryTimeWindow ,omitempty"`
-	EnableBatchedOperations             *bool                    `json:"enableBatchedOperations,omitempty"`
-	EnableExpress                       *bool                    `json:"enableExpress,omitempty"`
-	EnablePartitioning                  *bool                    `json:"enablePartitioning,omitempty"`
-	EnableSubscriptionPartitioning      *bool                    `json:"enableSubscriptionPartitioning,omitempty"`
-	FilteringMessagesBeforePublishing   *bool                    `json:"filteringMessagesBeforePublishing,omitempty"`
```
tombuildsstuff: We're currently exposing this field, but were unaware it was Deprecated / scheduled to be removed.

I can see [it's been removed from the Swagger in this commit](https://github.com/v-Ajnava/azure-rest-api-specs/commit/5d6a17103c57e57b2b5e49e77ec118b684f8366e#diff-654d402701463525866911e60a6308a1L2443) - however presumably it's still being returned in the API for the moment, is that likely to change?

arm/compute/models.go
```diff
@@ -397,7 +374,6 @@ type AvailabilitySet struct {
 	Location                   *string             `json:"location,omitempty"`
 	Tags                       *map[string]*string `json:"tags,omitempty"`
 	*AvailabilitySetProperties `json:"properties,omitempty"`
-	Sku                        *Sku `json:"sku,omitempty"`
```
tombuildsstuff: we're currently reliant on this field to expose whether an Availability Set is Managed or not - has it been Deprecated?

colemickens: (OT: what does it mean for an AS to be managed or not, any docs on that?)

marstr:
Howdy @tombuildsstuff ,

@kirthik and I are digging into these particular changes for you. We don't have all of the information yet, but from our initial conversations, it seems like these fields are not going to comeback for v9.0.0-beta.

For the broader conversation around deprecation, I think you have the right read on things. The services for a given API Version are supposed to only change for bugs, the contracts should never change. Between API Versions, all bets are off, the entire world can change. Because former API Versions will still be supported by the Services, tying yourself to an API Version for an operation 

At the moment, that is pretty awkward for the Azure-SDK-for-Go. We only generate the SDK from the most recent API Version, then support for former versions is left to just over-writing the API Version field (which isn't really an answer to the problem at all.) #517 brought up an interesting point, which is that we really need to support all API Versions that the Services themselves support. Our team is currently floating around a few ideas for how we're going to do that, and are running some experiments right now. Particularly, the semantics of composite swaggers don't play super well with how we were thinking of packaging multiple API Versions. I'll open up an issue to track that conversation here very soon.

tl;dr In theory, deprecation should apply to whole API Versions, not individual fields. API Versions should each have long-term support. We're working on updating the Azure-SDK-for-Go to expose that awesome theoretical stability to our customers.

arm/servicebus/models.go
marstr: Good news on this front, @tombuildsstuff. We've reached out to the service team, and they've agreed to publish these changes as part of a new API Version instead of modify the existing one. (Which is more consistent with the way API Versions are supposed to work.) that means that this change will not be part of V9

tombuildsstuff: @marstr that's great - thanks! :)

marstr:
Hey @tombuildsstuff, mind taking another look to ensure that the most recent version has fixed the Service Bus issues that you were seeing?

tombuildsstuff:
@marstr taking a quick look at the GitHub diff it LGTM - I'll update my fork and do another merge tomorrow :)

arm/compute/models.go
tombuildsstuff: Sorry - missed this! Digging our use - we're setting the SKU to `aligned` which [according to the initial PR](https://github.com/hashicorp/terraform/pull/12532) means the Availability Set can contain VM's with Managed Disks?

tombuildsstuff:
👋 hey @marstr 

I've taken a look into this today (excluding the Storage API changes, which are being handled separately) - and there's one major blocker from migrating to v9 at our end - which is [the removal of the SKU property from the Availability Set](https://github.com/Azure/azure-sdk-for-go/pull/569#discussion_r106481890). From what I can see, we'd need this functionality in order to be able to ship [the Managed Disks PR](https://github.com/hashicorp/terraform/pull/12455), which is pretty highly requested. 

Is there any chance this will be reverted by the time that the v9 SDK is released? (I was hoping to link to it, but I'm struggling to find the Swagger PR associated with that change)

FWIW one subtle thing I've noticed whilst working through this is that fields in the Swagger can have spaces in the names, i.e. the API field `duplicateDetectionHistoryTimeWindow` was in the Swagger as `duplicateDetectionHistoryTimeWindow ` (note the space). If there isn't already, is there anything AutoRest could do to handle/catch these kind of issues in future?

FYI I've pushed [the branch with the changes here](https://github.com/hashicorp/terraform/compare/master...azurerm-sdkv9-upgrade) - and we're going to run the full test suite overnight :)

Thanks! :)

tombuildsstuff:
Hey @marstr 

Just to follow up: I've run the AzureRM test suite a couple of times and aside from the issues I've pointed out above - it looks good :)

Thanks!

marstr:
Thanks @tombuildsstuff! I'll follow up by EOD with status of what I've learned about that field.

arm/compute/models.go
marstr: @tombuildsstuff, looks like this was a mistake on my part! I generated compute from the latest non-preview API Version. It looks like we've been generating from the version 2016-04-30-preview in the past. I've updated our manifest and am regenerating now. For v10, we'll generate from the composite swagger and this won't be an issue.

tombuildsstuff:
@marstr thanks a ton for all the extra effort here btw :)

marstr:
Not a problem, @tombuildsstuff. :)

PR #574, opened 2017-03-28
Modified gen.go to support comp swaggers gen

mcardosos:
There it is!
@marstr @jhendrixMSFT


Gododir/gen.go
```diff
@@ -26,8 +26,16 @@ type service struct {
 	Output      string
 	Swagger     string
 	SubServices []service
+	Modeler     modeler
 }
 
+type modeler string
+
+const (
```
marstr: 👍

PR #575, opened 2017-03-29
Updating gen.go to most recent swaggers.

marstr:


mcardosos:
The stuff that is still commented, why is it like that?
Now we can gen graph, network, etc properly

jhendrixMSFT:
Please update containerregistry to the latest swagger (2017-03-01).

marstr:
Take another look? @mcardosos @jhendrixMSFT

PR #578, opened 2017-03-31
Fix BlobStorageClient.AcquireLease to work with infinite lease times.

### LINKED ISSUE #577: Bug: BlobStorageClient.AcquireLease fails with infinite lease time
When using lease time set to `-1`, the call to `BlobStorageClient.AcquireLease` fails with:

```
storage.AzureStorageServiceError{Code:"MissingRequiredHeader", Message:"An HTTP header that's mandatory for this request is not specified.\nRequestId:2aebd74d-0001-0061-5c0c-aa75a6000000\nTime:2017-03-31T10:52:54.2325938Z", AuthenticationErrorDetail:"", QueryParameterName:"", QueryParameterValue:"", Reason:"", StatusCode:400, RequestID:"2aebd74d-0001-0061-5c0c-aa75a6000000"} ("storage: service returned error: StatusCode=400, ErrorCode=MissingRequiredHeader, ErrorMessage=An HTTP header that's mandatory for this request is not specified.\nRequestId:2aebd74d-0001-0061-5c0c-aa75a6000000\nTime:2017-03-31T10:52:54.2325938Z, RequestId=2aebd74d-0001-0061-5c0c-aa75a6000000, QueryParameterName=, QueryParameterValue=")
```

The reason is `AcquireLease` sets the mandatory `x-ms-lease-duration` header only when the supplied lease time is positive. However, this header must always be set, and it must be set to `"-1"` when lease is infinite. See docs: https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/Lease-Blob?redirectedfrom=MSDN

ppanyukov:
This resolves #577

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -755,6 +755,8 @@ func (b BlobStorageClient) AcquireLease(container string, name string, leaseTime
 
 	if leaseTimeInSeconds > 0 {
 		headers[leaseDuration] = strconv.Itoa(leaseTimeInSeconds)
+	} else {
+		headers[leaseDuration] = "-1"
 	}
 
 	if proposedLeaseID != "" {
--- a/storage/blob_test.go
+++ b/storage/blob_test.go
@@ -1046,6 +1046,20 @@ func (s *StorageBlobSuite) TestAcquireLeaseWithBadProposedLeaseID(c *chk.C) {
 	c.Assert(err, chk.NotNil)
 }
 
+func (s *StorageBlobSuite) TestAcquireLeaseWithInfiniteLeaseTime(c *chk.C) {
+	cli := getBlobClient(c)
+	cnt := randContainer()
+
+	c.Assert(cli.CreateContainer(cnt, ContainerAccessTypePrivate), chk.IsNil)
+	defer cli.deleteContainer(cnt)
+
+	blob := randName(5)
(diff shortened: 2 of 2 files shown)


ppanyukov:
Not sure why Travis build failed. Have I done something wrong?

marstr:
Howdy @ppanyukov, 

Travis CI was failing due to Azure/azure-storage-go#41. However, as you'll see now, we've moved the storage package into it's own repository. I hope this hasn't caused too much confusion!

We still value your contribution! If you'd like to have it considered, please re-target it against the new storage repository.

-Martin

ppanyukov:
@marstr you bet it did cause a lot of confusion! :)  

More importantly, how would you like to proceed?  I have this fixed in our vendored tree using "hacks" so not super urgent for us, but this would be nice to get fixed properly.  Would you prefer me to submit this PR against that other repo? Or you think you'd prefer to fix this yourselves: this fix and tests are trivial as you will see.

Happy to go either way, please advise.

marstr:
@mcardosos, can you comment on whether this is something you're already fixing as part of your refactoring, or if it would be more efficient to cherry-pick this PR in the other repo?

mcardosos:
@ppanyukov Thanks for submitting a PR!
This has been already fixed in https://github.com/Azure/azure-storage-go
Still, I think your test is a good contribution to the SDK :) Please submit a PR with the test against the other repo.

Issue #590, opened 2017-04-18
There doesn't seem to be a way to get-blob and get content length.


jrwren:
There doesn't seem to be a way to get-blob and get content length without resorting to multiple calls.

storageResponse isn't exposed. GetBlobProperties is a separate call.

It would be nice if there were a call like GetBlob(...) exposed which included access to all of the Response Headers as listed at `https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/Get-Blob?redirectedfrom=MSDN`

The API as is requires 2 calls to obtain a blob & the blobs length, without resorting to reading the entire blob to caclulate the length. Regardless of length, it would be nice to have other headers such as ETag available.

jhendrixMSFT:
This has been fixed and will be included with the v10 release of the SDK.

jrwren:
I can't find the fix/feature in this repo or at https://github.com/Azure/azure-storage-go. Will you please link me to the commit with the feature?

jhendrixMSFT:
It was fixed in this PR https://github.com/Azure/azure-storage-go/pull/46

PR #591, opened 2017-04-18
GetBlobAndProperties to prevent extra req

jrwren:


jrwren:
Addresses issue 590 https://github.com/Azure/azure-sdk-for-go/issues/590

storage/blob.go
```diff
@@ -513,6 +513,57 @@ func (b BlobStorageClient) GetBlob(container, name string) (io.ReadCloser, error
 	return resp.body, nil
 }
 
+// GetBlob returns a stream to read the blob. Caller must call Close() the
+// reader to close on the underlying connection.
+//
+// See https://msdn.microsoft.com/en-us/library/azure/dd179440.aspx
+func (b BlobStorageClient) GetBlobAndProperties(container, name string) (io.ReadCloser, *BlobProperties, error) {
```
ahmetb: you should probably write an unexported helper method that does this, and use that in GetBlobAndProperties (directly) as well as in GetBlob (by discarding BlobProperties). Otherwise it's duplicating code for no good reason.

mcardosos: Alternatively it can be an unexported `getProperties` that is used both by this method and `GetBlobProperties`.

jrwren:
I closed this because the related issue says that it will be fixed in the V10 tools release.

Issue #584, opened 2017-04-12
Verify client account name instead of panicing


mjibson:
The following program panics at `service.CreateBlockBlob` because of the newline in the client account name. NewBasicClient should perhaps verify the account name is a valid format and return an error if it is not, instead of ignoring it and causing a panic later on.

```
package main

import (
	"github.com/Azure/azure-sdk-for-go/storage"
)

func main() {
	client, err := storage.NewBasicClient("\n", "password")
	if err != nil {
		panic(err)
	}
	service := client.GetBlobService()
	if err := service.CreateBlockBlob("c", "n"); err != nil {
		panic(err)
	}
}
```

mcardosos:
Fixed!

PR #594, opened 2017-04-21
Check validity of storage accounts

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/584
PTAL @jhendrixMSFT @marstr

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -160,16 +160,12 @@ func NewEmulatorClient() (Client, error) {
 // storage endpoint than Azure Public Cloud.
 func NewClient(accountName, accountKey, blobServiceBaseURL, apiVersion string, useHTTPS bool) (Client, error) {
 	var c Client
-	if accountName == "" {
-		return c, fmt.Errorf("azure: account name required")
+	if !IsValidStorageAccount(accountName) {
+		return c, fmt.Errorf("azure: account name is not valid: it must be between 3 and 24 characters, and only may contain numbers and lowercase letters: %v", accountName)
 	} else if accountKey == "" {
 		return c, fmt.Errorf("azure: account key required")
 	} else if blobServiceBaseURL == "" {
 		return c, fmt.Errorf("azure: base storage service url required")
-	}
-
-	if !validStorageAccount.MatchString(accountName) {
-		return c, fmt.Errorf("azure: account name is not valid: it must be between 3 and 24 characters, and only may contain numbers and lowercase letters: %v", accountName)
 	}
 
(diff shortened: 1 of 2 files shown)


storage/client.go
```diff
@@ -48,6 +49,10 @@ const (
 	userAgentHeader = "User-Agent"
 )
 
+var (
+	validStorageAccount = regexp.MustCompile("^[0-9a-z]{3,24}$")
```
jhendrixMSFT: I'm concerned about having this logic in our SDK.  If/when this restriction changes we will need to update our logic too.  Is there a way we can have Azure validate the name?

jhendrixMSFT: Well I suppose if we had a swagger for this it would be codified there and would be handled per normal validation logic, and if it changes in the swagger we'd get updated validation.

marstr: I don't want folks to have to duplicate this logic in their code, and I can imagine folks wanting to validate before making the NewClient call. Would something like this be too fancy? [Go Playground](https://play.golang.org/p/malkhcYxzr)

If it is too fancy, maybe just expose a method named `ValidAccountName(name string) bool`

marstr: Without adding anything, Azure would validate the account name by returning a failure. :) I think the idea is just to save people a request. Though, I do agree with the principal of not encoding unnecessary logic.

marstr: Upon further consideration, it get's in the way too much of just calling a function. I'll type up another way to do this.

mcardosos: Too fancy. Not a big fan of making AccountName something that is not a string

mcardosos: [For the curious](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-storage/2016-12-01/swagger/storage.json#L69). Swaggers actually include regex sometimes.
ARM storage swagger already includes a method that does a [very similar thing](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-storage/2016-12-01/swagger/storage.json#L28).... As a plus it also checks if the storage account name is available.

storage/client.go
```diff
@@ -163,6 +168,10 @@ func NewClient(accountName, accountKey, blobServiceBaseURL, apiVersion string, u
 		return c, fmt.Errorf("azure: base storage service url required")
 	}
 
+	if !validStorageAccount.MatchString(accountName) {
```
mcardosos: fixed

storage/client.go
mcardosos: added exposed method

PR #595, opened 2017-04-21
storage: avoid transiting url.URL through string

tamird:
Inspired by #584, #594.

@mcardosos @jhendrixMSFT @marstr


tamird:
This may also allow reverting #594; I think the validation will take place on the server, in that case.

mcardosos:
@tamird Thanks for the contribution!

tamird:
Thanks for the quick reviews!

Issue #581, opened 2017-04-03
submodule file conflicts break glide/git


vikstrous:
We use glide to vendor this repo and a recent change broke glide's ability to depend on old versions of this repo.

The issue is that glide clones the latest version of Azure/azure-sdk-for-go and then tries to check out an old version. When it does that, it gets the following errors:

```
[ERROR]	Failed to set initial config references: Unable to update checked out version
[DEBUG]	Output was: error: The following untracked working tree files would be overwritten by checkout:
	storage/blob.go
	storage/blob_test.go
	storage/client.go
	storage/client_test.go
	storage/file.go
	storage/file_test.go
	storage/queue.go
	storage/queue_test.go
	storage/util.go
	storage/util_test.go
```

This happens because the storage directory was replaced with a submodule.

See https://github.com/Masterminds/glide/issues/749

We can blame git, glide or azure-sdk-for-go, but I think the simplest solution that doesn't require new code to be written is here. That's why I'm hoping that you guys can simply rename the storage directory on master to something else. That way these conflicts won't exist and we can move on without having to wait for code changes elsewhere.

Meanwhile, we'll probably just fork this repo and remove the latest commits as a workaround :(

Edit: I believe the breaking change was introduced in https://github.com/Azure/azure-sdk-for-go/pull/569

marstr:
Howdy, @vikstrous sorry you're running into trouble.

**Assuming you don't have any alterations to the SDK that you need to keep**, the _easiest_ way forward is to checkout a fresh copy of the SDK. For the benefit of others who have found this thread, I'll include instructions for folks using glide and otherwise.

## Glide
1. `cd <project directory>/vendor/github.com/Azure`
2. `rm -rf azure-sdk-for-go`
3. `cd ../../..`
4. `glide install`

## Go Get
1. `cd $GOPATH/src/github.com/Azure/`
2. `rm -rf azure-sdk-for-go`
3. `go get -u github.com/Azure/azure-sdk-for-go`

vikstrous:
This is not sufficient. `glide update` works by cloning the repo first, running get submodule init/update, then checking out the version specified in the glide.yaml file. By the time it has cloned master, and initialized submodules it's too late and checking out an old version will fail unless the submodule directory is deleted. There is no workaround without code changes to glide from what I can tell.

On April 4, 2017 10:49:54 AM PDT, Martin Strobel <notifications@github.com> wrote:
>Howdy, @vikstrous sorry you're running into trouble.
[...]

PR #597, opened 2017-04-26
Return storage package to this repository.

marstr:
One of the conditions for moving storage into its own repository was continued support of folks consuming it out of this repository. Our mechanism for doing that was adding a git sub-module as seen here: #556. We tested moving forward to adopt the latest bits, but missed an important scenario: using glide/git to tie yourself to a commit that predated the sub-module. This caused people to stop being able to use our SDK as seen in #581.
The immediate solution was to back-out the sub-module, replacing it with copies of the files as they existed at the commit the sub-module referred to. However, we were then left with a choice. Do we permanently move to the new repository, deprecating the copy here, or deprecate the new repository and return to developing out of a single repository?
While there were supporters of both options on the team, we made the decision to return the code here based on the download numbers we saw for the new repository as seen in GitHub's Graphs. Almost all of the clones of `Azure/azure-storage-go` came indirectly through the sub-module, meaning almost everybody using that package was doing so via this repository.

For that reason, I've rebased all commits from the old repository onto our history here.


marstr:
What are your thoughts, @kpfaulkner?

kpfaulkner:
Although personally I prefer it as a separate repo, if reality is that most people using it are getting it through azure-sdk-for-go then I'd say merge it back.

kpfaulkner:
Just to clarify, gong forward will we still be working off the dev branch I assume?

.travis.yml
```diff
@@ -18,12 +18,14 @@ script:
...
   - test -z "$(gofmt -s -l -w management | tee /dev/stderr)"
+  - test -z "$(gofmt -s -l -w storage | tee /dev/stderr)"
   - test -z "$(gofmt -s -l -w Gododir | tee /dev/stderr)"
   - test -z "$(go build $(find ./* -type d -print | grep -v '^./vendor$' | grep -v '^./storage$') | tee /dev/stderr)"
   - test -z "$(go vet $(find ./arm/* -type d -print) | tee /dev/stderr)"
   - test -z "$(golint ./arm/... | tee /dev/stderr)"
   - go test -v ./management/...
   - test -z "$(golint ./management/... |  grep -v 'should have comment' | grep -v 'stutters' | tee /dev/stderr)"
   - go vet ./management/...
+  - go vet ./storage/...
```
mcardosos: >go vet ./storage/... [](start = 4, length = 20)

At some point we have to add `go test ./storage/...`

mcardosos:
> [GoDoc](https://godoc.org/github.com/Azure/azure-sdk-for-go) 

Why is this file inside a folder?

---
Refers to: README.md:2 in 0574446. [](commit_id = 05744467ef80e9bb8faf7159908a39f181141cdf, deletion_comment = False)

marstr:
Yes, @kpfaulkner. The plan is to continue working off of the dev branch.

PR #600, opened 2017-05-02
Fixed examples to use new go-autorest adal package

xtophs:
Examples didn't build after go-autorest https://github.com/Azure/go-autorest/pull/122 was merged

### DIFF
--- a/arm/examples/check/check.go
+++ b/arm/examples/check/check.go
@@ -9,7 +9,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/arm/examples/helpers"
 	"github.com/Azure/azure-sdk-for-go/arm/storage"
 	"github.com/Azure/go-autorest/autorest"
-	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/azure"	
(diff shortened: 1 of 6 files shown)


xtophs:
adding @colemickens, @marstr and @cosmincojocar

What's the trick to setting the environment variables to get the CI build to pass?

marstr:
Howdy @xtophs, I'm not sure what you mean here:
> What's the trick to setting the environment variables to get the CI build to pass?

However, taking a glance at CI, it seems that `go fmt` is finding violations in the following files:
- arm/examples/check/check.go
- arm/examples/create/create.go
- arm/examples/check/check.go
- arm/examples/create/create.go

You should be able to run `go fmt` locally to see the specific violations. That said, it's sorta shady that one doesn't see the specific violations. I'll have a look to see what I can do about that moving forward.

xtophs:
@marstr Thanks for the pointer ... I didn't click far enough down the CI build report. 

and apologies for the slow response time. My notification settings were a bit lax. That's fixed now. 

With the last commit, I also updated handling the return value of the ac.Create call, which is now <-chan  
error instead of error.

Thanks

PR #601, opened 2017-05-03
Regenerated the SDK to include the latest Service Bus changes

### LINKED ISSUE #593: Regenerate the Service Bus SDK
👋 hey Azure SDK for Go team!

I've been going through trying to upgrade us to v9 of the SDK - and noticed a blocker 😞

When previously dry-running the SDK upgrade - I'd commented out a breaking change in the Service Bus SDK, where the Location field has been removed from the Response models. Unfortunately this meant that I'd only reported 1 of the 2 issues with the Service Bus SDK, which means we're currently unable to upgrade to v9 of the Azure SDK for Go 😞 

I was hoping that we may be able to work around this by not posting the Location field at Creation/Update time - but unfortunately it's a required field. I've opened a PR to [the Rest API Specs Repository which fixes this](https://github.com/Azure/azure-rest-api-specs/pull/1138) - once this is merged, would it be possible to re-generate and re-release the SDK?

Could we also add this to [the list of breaking changes](https://github.com/Azure/azure-sdk-for-go/pull/569#issuecomment-285460597) in v9 too until this is resolved?

Thanks!

tombuildsstuff:
👋 hey y'all

Feel free to close this if it's going to cause more pain at your end, but I had some spare time and ended up taking a stab at regenerating the Service Bus SDK - and subsequently updating the rest of the SDK with the version number `9.1.0-beta`

I wasn't 100% sure what the original commit used to generate this was, but an educated guess says that `1257d04ddfe03521d83d73627a1897d570e09037` might be a good candidate (which is what I've used)? I've then updated the Service Bus SDK to include the changes included in https://github.com/Azure/azure-rest-api-specs/pull/1138 (commit `0ca88c0c276ea4fd96db1a3b69951638099db503`)

Related: is there a means of telling which version of the Repository was used to generate the Swaggers?

Fixes #593

cc @marstr


marstr:
Thanks for taking the initiative here @tombuildsstuff! I'll take a look at this PR and make the next moves sometime tomorrow.

mcardosos:
Hello @tombuildsstuff ! I have verified, and we generated the same Service Bus SDK, so the changes you need will be included in v10 :D
Right now there is no way to know which commit was used to gen, unless you want to play detective. This time we are including in the changelog which commits were used to gen the SDK, for both AutoRest and the swaggers.

tombuildsstuff:
@mcardosos / @marstr cool, thanks for that - I'll close this PR out then :)

So that I can manage expectations on our end, any idea of a timeframe for when v10 will be merged/tagged into the master branch? I'll also update https://github.com/hashicorp/terraform/pull/14004 to use v10 instead of v9 as soon as I get a chance :)

Thanks!

mcardosos:
@tombuildsstuff 
Tagged and released :D

Issue #587, opened 2017-04-12
KeyVault - Bad data types used for ints


khenidak:
the values for NotBefore, Created and Updates are incorrect and cause unmarshmaling errors [here]( https://github.com/Azure/azure-sdk-for-go/blob/master/dataplane/keyvault/models.go#L518). They should be json.Number

Edit: This apply on SecretBundle:SecretAttributes & KeyBundle:KeyAttributes
Edit #2: also applies on CertificateBundle

marstr:
At the moment, it looks like our generator doesn't support "unixtime" format when it is seen in Swagger. (The defintions in question can be found here: [Azure/azure-rest-api-specs/keyvault/2015-06-01/swagger/keyvault.json](https://github.com/Azure/azure-rest-api-specs/blob/c0c9d6353a7d87ada601a609460a11b67f6f15ae/keyvault/2015-06-01/swagger/keyvault.json#L1805)) When that happens, we bomb out into strings to support the broadest set of data types. As you point out, that can cause marshaling issues.

To get around these issues, I'm adding support into [Azure/go-autorest ](https://github.com/Azure/go-autorest) for the UnixTime. Once that PR is accepted, I'll replace [this line](https://github.com/Azure/autorest/blob/v1.0.1-20170417-2300-nightly/src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs#L115) in the generator with the newly created type.

marstr:
This is fixed in #602

marstr:
Thanks for bringing this to our attention @khenidak! I believe I have it fixed.

PR #602, opened 2017-05-04
Regenerating to support UnixTime

marstr:
This in Response to Azure/autorest#2245

- [x] Update CHANGELOG.md to use the result of Azure/autorest#2245 as the AutoRest commit.


marstr:
@mcardosos any insight as to why the storage tests would be failing?

marstr:
THis PR fixes #587

PR #619, opened 2017-05-10
Bump patch version to v10.0.2

jhendrixMSFT:


mcardosos:
Is it because of the storage version change?

PR #622, opened 2017-05-13
How the SAS client could look

mcardosos:
Fancy fancy client that can do operations with SAS URIs!
Right now I have only added this functionality to a couple of funcs (and its tests). Right now `TestContainerExists` is not passing.
PTAL
@jhendrixMSFT @marstr @DavidObando


storage/blobserviceclient.go
```diff
@@ -45,6 +47,21 @@ func (b *BlobStorageClient) GetContainerReference(name string) *Container {
 	}
 }
 
+// GetContainerReferenceFromSASURI returns a Container object for the specified
+// container SASURI
+func GetContainerReferenceFromSASURI(sasuri url.URL) (*Container, error) {
```
jhendrixMSFT: Does this need to be public?

Issue #626, opened 2017-05-18
CreateBlockBlobFromReader fails due to missing content length parameter


radu-matei:
Usage:
`blob.CreateBlockBlobFromReader(reader, nil)`

Output:
`storage: service returned error: StatusCode=411, ErrorCode=MissingContentLengthHeader, ErrorMessage=Content-Length HTTP header is missing`

`fmt.Printf("Content length: %d", blob.Properties.ContentLength)` - this outputs 0

radu-matei:
Quick fix - https://github.com/radu-matei/azure-sdk-for-go/commit/7bfa39141a170036a54ae46c1c89ad92a539c570

radu-matei:
PR #627

marstr:
Howdy @radu-matei,

I'm investigating this issue now, thanks for bringing this to our attention. I'll get back to you ASAP on how we should move forward.

My gut reaction is that this should be solved by better calculating blob.Properties.ContentLength, not making that calculation the responsibility of the caller.

-Martin

marstr:
For the life of me, I'm not sure why our client.go is even setting Content-Length. It seems to me that if we just leave it unset, the HTTP stack in go will correctly set it.

marstr:
@radu-matei, can you confirm that this issue is fixed by v10.0.3-beta?

Blackbaud-ChrisJenkins:
@marstr I just ran into this issue, running
```
- name: github.com/azure/azure-sdk-for-go
  version: 26132835cbefa2669a306b777f34b929b56aa0a2
```

so I think the answer to your question is no

radu-matei:
@Blackbaud-ChrisJenkins - until there is a fix in the SDK you can pass the size yourself, as in https://github.com/Azure/azure-sdk-for-go/pull/627

radu-matei:
Any updates on this issue? I am unable to use the SDK because of it (I have to use my fork instead)

marstr:
Howdy, I didn't get to this. I've been on vacation without internet access for the last 5 days or so. @mcardosos and I will look into this today and get another patch out to really get this issue fixed.

marstr:
To mitigate this, @mcardosos wrote up another Pull Request that counts the size of the blob before submitting a request, you can see it in #638. In the long run, I really think this is something that should get pushed off on the Go standard library as much as possible. However, we'll do this in the spirit of unblocking folks.

marstr:
[...]

PR #627, opened 2017-05-18
Add content size to header of request for block blob

radu-matei:
Fix for #626


radu-matei:
The build doesn't seem to be right - `no environment variables set`

marstr:
Looking through the log, it looks like travis failed due to tests due to the breaking nature of the changes in this PR. The message "no environment variables set" seems to be a red herring.

marstr:
I'm going to close this in favor of #628 which does not introduce breaking changes. Thanks for this contribution though!

PR #628, opened 2017-05-18
Removing storage/client.go Content-Length calc.

marstr:
As far as I can tell, Go will calculate the correct value for
Content-Length independently if we just leave it unset when we create a
new request.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -17,7 +17,6 @@ import (
 	"net/url"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


marstr:
This PR is made to address #626

marstr:
I've tested this locally as well so that I can test against the live services instead of the recordings. I believe I was able to test against the live services by setting the environment variables ACCOUNT_NAME and ACCOUNT_KEY but I'm having trouble verifying that my efforts were successful. @mcardosos, can you confirm that I have tested this appropriately?

mcardosos:
If you want to test against Azure, the best thing to do is, delete de recordings folder, run the tests and let's see how it goes. After everything goes right, you can undo all the changes to the recordings with git.

marstr:
Thanks, @mcardosos, you're the best.

I have verified that my change doesn't break any of the tests after removing the recordings and testing against the live services. Mind double checking my conclusion before I release a patch, @jhendrixMSFT?

jhendrixMSFT:
Does this work with Go 1.7?  There were some changes between 1.7 and 1.8 with respect to Content-Length.

marstr:
Yep, I've tested the change locally (on two different machines) using both 1.7 and 1.8. All tests pass in both cases.

marstr:
Also, for interested parties, these are the Go 1.8 Release Notes that pertain to Joel's comment:
> The Transport no longer reads a byte of a non-nil Request.Body when the Request.ContentLength is zero to determine whether the ContentLength is actually zero or just undefined. To explicitly signal that a body has zero length, either set it to nil, or set it to the new value NoBody. The new NoBody value is intended for use by Request constructor functions; it is used by NewRequest.

PR #636, opened 2017-05-23
Rewriting generator script

marstr:
I've included the resulting differences to the packages that we generate from the generator that I've written.


marstr:
The most important files to review are in the `generator` folder

brjohnstmsft:
@marstr Please exclude Azure Search from the generated data plane libraries. While our spec under arm-search is complete, the other specs are not. See this issue for details and current status: https://github.com/Azure/azure-rest-api-specs/issues/1085

mcardosos:
Will we get rid of the Gododir?

marstr:
> Will we get ride of the Gododir?

Yep :)

jhendrixMSFT:
@marstr you should take a look at how the specs repo is being refactored.  Depending on the timing it might make sense to hold off on this change then retool it for the refactor.
https://github.com/Azure/azure-rest-api-specs/tree/reorg

generator/program.go
```diff
@@ -0,0 +1,389 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/Masterminds/semver"
+	"github.com/marstr/swagger"
+)
+
+const (
+	defaultRemoteAzureRestAPISpecsPath = "https://github.com/Azure/azure-rest-api-specs.git"
```
mcardosos: Nit in consistency, `Rest` or `REST`? The same applies to some vars

generator/program.go
```diff
@@ -0,0 +1,389 @@
...
+	optionalTools := []string{"gofmt", "golint"}
+	requiredTools := []string{"autorest", "git"}
+
+	for _, tool := range optionalTools {
+		if _, err := exec.LookPath(tool); err != nil {
+			log.Printf("WARNING: Could not find \"%s\" usage of this tool will be skipped.", tool)
+		}
+	}
+
+	anyMissing = false
```
mcardosos: bools are set to false when created. You don't need this line.

marstr: Right you are! Personally, I like to set these things explicitly for readability and just in case it gets set somewhere else.

generator/program.go
marstr: Derp, you're totally correct. This should be REST.

marstr: Fixed locally

marstr:
> ...you should take a look at how the specs repo is being refactored. Depending on the timing it might make sense to hold off on this change then retool it for the refactor.

True @jhendrixMSFT, I've tried to make this as agnostic to directory organization as possible so that when it does happen there are only a few lines that need to be changed. I'm also intentionally keeping this off the `master` branch for now.

PR #638, opened 2017-05-26
Fixes issue 626

mcardosos:
PTAL
@jhendrixMSFT  @marstr


marstr:
I'm okay with this as a mitigatory fix for #626, but in general I'd like to stay away from us manually setting content-length as much as possible.

PR #644, opened 2017-06-08
Reduce allocations for request retry logic.

jhendrixMSFT:
Using ioutil.ReadAll() to read the request body can introduce a large
number of allocations increasing heap pressure.  For cases where the
content length of the request is known we can read the body into an
allocated buffer of exact size.
Update go-autorest dependency to 8.1.0 to use RetriableRequest feature.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -75,19 +75,13 @@ type DefaultSender struct {
 
 // Send is the default retry strategy in the client
 func (ds *DefaultSender) Send(c *Client, req *http.Request) (resp *http.Response, err error) {
-	b := []byte{}
-	if req.Body != nil {
-		b, err = ioutil.ReadAll(req.Body)
+	rr := autorest.NewRetriableRequest(req)
+	for attempts := 0; attempts < ds.RetryAttempts; attempts++ {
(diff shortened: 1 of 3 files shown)


storage/client.go
```diff
@@ -75,17 +75,27 @@ type DefaultSender struct {
...
 	for attempts := 0; attempts < ds.RetryAttempts; attempts++ {
-		if len(b) > 0 {
-			req.Body = ioutil.NopCloser(bytes.NewBuffer(b))
+		if reader != nil {
+			// move the reader back to the beginning of the buffer
+			_, err = reader.Seek(0, io.SeekStart)
+			if err != nil {
+				return resp, err
+			}
+			req.Body = ioutil.NopCloser(reader)
```
marstr: I _believe_ you can pull this line outside of the loop, which will result in N fewer allocations of a blank NopCloser.

jhendrixMSFT: Yes I think you're right, I'll make this change.

marstr:
Beautiful :)

jhendrixMSFT:
@marstr @mcardosos can you please take a look at the latest iteration?

Issue #645, opened 2017-06-08
SetBlobMetadata may change the case of the existing meta data name


yuwaMSFT2:
In C# client, we put a metadata "name=key".
In Go, we could query the metadata "name=key". Now if we can SetBlobMetadata with the same metadata, the metadata on the blob will be changed to "Name=key".

This is because in client.exec, we call req.Header.Add, which will convert the key to MIMEHeader which will make the first chacater of a word uppercase.

For Storage REST API, the metadata is set through header:
x-ms-meta-name

With the canonization, the header becomes:
X-Ms-Meta-Name

Server will change the metadata name to "Name" now.

Now if C# client tries to get the metadata with "name", it gets nothing.

Go sdk should not change the case of the name (although I think C# sdk can also do some improvement here).

yuwaMSFT2:
The fix would be to avoid using MIMEHeader convention for these metadata headers.
Instead of calling req.Header.Add, can call req.Header["name"] = "key" directly.

yuwaMSFT2:
@marstr @DavidObando @SajayAntony @mcardosos  @sivagms

marstr:
Howdy @yuwaMSFT2, this is a really interesting bug. We've decided your fix is correct, and I'll try to get out a PR out here quickly to fix it.

The relevant Azure documentation is here: https://docs.microsoft.com/en-us/rest/api/storageservices/setting-and-retrieving-properties-and-metadata-for-blob-resources which says:
> Names are case-insensitive. Note that metadata names preserve the case with which they were created, but are case-insensitive when set or read.

It's less easy to see here, but as you point out the Go Standard Library is munging the case as the user sets it in the [Header type](https://godoc.org/net/http#Header).

As you point out, the service handles this fine, but apparently both clients need to be better citizens about interacting with this service. (As an aside, given that HTTP Headers are supposed to be case-insensitive, I think the behavior the service has chosen is kinda silly.)

SajayAntony:
@marstr Does this mean storage's behavior should change?

marstr:
Yep! working on the PR now

marstr:
The PR to address this is #646

PR #646, opened 2017-06-09
Fixing #645

marstr:
This bypasses the Go Standard Library munging of Header casing.


marstr:
Ah ha, so this is is interesting. CI is failing, even though test will pass against the live services. @mcardosos and I chased this down to being down to the fact that the recordings get tested in a case-sensitive way.

marstr:
(For the sake of GitHub tagging, I'll repeat this information here.) This PR addresses #645

storage/client_test.go
```diff
@@ -114,39 +138,55 @@ func compareHeaders(r *http.Request, i cassette.Request) bool {
...
 		}
-		requestHeaders.Del("Content-Type")
-		cassetteHeaders.Del("Content-Type")
+
+		deleteHeaders(requestHeaders, isContentType)
+		deleteHeaders(cassetteHeaders, isContentType)
 	}
 
-	return reflect.DeepEqual(requestHeaders, cassetteHeaders)
+	b := reflect.DeepEqual(requestHeaders, cassetteHeaders)
```
mcardosos: >b := reflect.DeepEqual(requestHeaders, cassetteHeaders) [](start = 1, length = 55)

Better `return reflect.DeepEqual(requestHeaders, cassetteHeaders)`

marstr: Derp, looks like debugging nonsense got included.

storage/client.go
```diff
@@ -399,7 +399,7 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 	}
 
 	for k, v := range headers {
-		req.Header.Add(k, v)
+		req.Header[k] = append(req.Header[k], v)
```
jhendrixMSFT: Can you add a comment here explaining why we don't use Add() here.

marstr: Good call, we don't want history to repeat itself.

Issue #656, opened 2017-06-16
Queue SAS generation doesn't exist


kpfaulkner:
Need to be able to create SAS URL for Storage queues.

About to submit PR.

mcardosos:
Hello @kpfaulkner !
I added a little comment in you PR :D

PR #422, opened 2016-10-21
Containeracl

kpfaulkner:
This covers setting/getting ACLs for containers and also AccessPolicy creation.
This is referenced in Issue 14.

Hope it is of some use.

PR #657, opened 2017-06-16
Queuesas

kpfaulkner:
Generating SAS URLs for Storage queues. Similar approach to the blob SAS generation.
Related issue https://github.com/Azure/azure-sdk-for-go/issues/656

One problem I haven't been able to resolve is that if I'm reading the specs ( https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas ) correctly I *should* be able to specify the protocol to use (http and/or https). Currently I cannot get that working, but leaving it as blank IS accepted by the spec (and works).


mcardosos:
Hello @kpfaulkner !
Thanks for submitting a PR.
While looking at you PR, I noticed some other issues there are in the already checked-in code, so I addressed them on [this other PR](https://github.com/Azure/azure-sdk-for-go/pull/662). Your PR and mine fix stuff in a somewhat different way, but it would be cool if both PRs were consistent (for example, for the SAS permissions, I liked very much the pattern you used in https://github.com/Azure/azure-sdk-for-go/pull/422 with the `CanRead`, `CanWrite` and `CanDelete` bools, so I preferred to follow that pattern for SAS implementation too). I have also discussed this with @jhendrixMSFT . Do the changes in https://github.com/Azure/azure-sdk-for-go/pull/662 make sense to you? Feel free to comment it if something does not feel right :) Meanwhile I will take a look on the issues you have mentioned in your PR's comments and how to fix them (the identifier, start time and protocols).

kpfaulkner:
Cool, yes am able to restructure this PR to be a similar style to #662 and #422.

Is #652 getting merged soon?

storage/queuesasuri.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// assumption that start time is now.
+	//signedStart := time.Now().UTC().Format(time.RFC3339)
+	signedExpiry := expiry.UTC().Format(time.RFC3339)
+
+	// Cannot get this working yet. Any values entered generates bad URL.
+	protocols := ""
```
mcardosos: When does this generate a bad URL? I played a little with the protocols option but I did not find problems. Is there something I missed?

kpfaulkner: whenever I had a value for "protocols" it wouldn't work. Empty string would be the only value I could get working.  You're not seeing that?

kpfaulkner: I'll try and repro later today. But whenever I had a value I'd never get it to generate a URL successfully.

kpfaulkner: For example if protocol is populated with "https" I generate the SAS URL but when trying to use it I get the error:

<Error>
<Code>AuthenticationFailed</Code>
<Message>
Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature. RequestId:0e3aae9d-0003-002f-3e81-ec31f4000000 Time:2017-06-24T00:31:37.0443316Z
</Message>
<AuthenticationErrorDetail>
Signature did not match. String to sign used was p 2017-06-24T01:31:26Z /queue/kenfau/temp 2016-05-31
</AuthenticationErrorDetail>
</Error>

But protocol of an empty string is fine.

kpfaulkner: Updated PR to follow similar lines to @mcardosos BlobSAS changes....  although I still cannot get protocol to work properly.

mcardosos: Hmmm, so the protocol should be included both in the string to sign and on the query. Curious question, when trying to use the SAS URI, are you sending the request with an authorization header?

mcardosos: My string to sign looks like this...
```
raup

2050-12-20T21:55:06Z
/queue/golangrocksonazure/name


https,http
2016-05-31
```
And if I do (manually) a Get queue metadata operation, the URL looks like this
```
https://golangrocksonazure.queue.core.windows.net/name?comp=metadata&se=2050-12-20T21%3A55%3A06Z&sig=nGP%2BOUcNIc5i%2Fcb1sFD2Ksnpz1CHazjdVy2ewurTUsE%3D&sp=raup&spr=https%2Chttp&sv=2016-05-31
```
When I get the same error as you is when I don't include the protocols in the URL as query parameters.

mcardosos:
Merged https://github.com/Azure/azure-sdk-for-go/pull/662 :D

storage/queuesasuri.go
kpfaulkner: Ahh ok, my bad, I wasn't including them as a parameter as well.

kpfaulkner:
Everything looking good now.

storage/queuesasuri.go
```diff
@@ -0,0 +1,132 @@
...
+	// -- https://msdn.microsoft.com/en-us/library/azure/dn140255.aspx
+	// We need to replace + with %2b first to avoid being treated as a space (which is correct for query strings, but not the path component).
+	canonicalizedResource = strings.Replace(canonicalizedResource, "+", "%2b", -1)
+	canonicalizedResource, err = url.QueryUnescape(canonicalizedResource)
+	if err != nil {
+		return "", err
+	}
+
+	// assumption that start time is now.
+	signedStart := ""
```
mcardosos: This assumption is done by Azure, not the SDK. I would prefer this comment to go away.

storage/queuesasuri.go
```diff
@@ -0,0 +1,132 @@
...
+	uri := q.qsc.client.getEndpoint(queueServiceName, q.buildPath(), nil)
+	sasURL, err := url.Parse(uri)
+	if err != nil {
+		return "", err
+	}
+	sasURL.RawQuery = sasParams.Encode()
+	return sasURL.String(), nil
+}
+
+func queueSASStringToSign(signedVersion, canonicalizedResource, signedStart, signedExpiry, signedIP string, signedPermissions string, protocols string, signedIdentifier string) (string, error) {
```
mcardosos: >string [](start = 100, length = 6)

Nit, "string" is not needed by all parameters here.

mcardosos:
@kpfaulkner Cool! just some last tiny comments, but other than that, LGTM :D

mcardosos:
Awesome! @kpfaulkner 
Thanks for contributing!

PR #662, opened 2017-06-21
GetSASURI is more complete and has grouped params

mcardosos:
GetSASURI func now receive grouped parameters
GetSASURI also includes signed identifier and overridable headers
Got rid of GetSASURIWithSignedIPAndProtocol

PTAL
@marstr @jhendrixMSFT

### DIFF
--- a/storage/blobsasuri.go
+++ b/storage/blobsasuri.go
@@ -8,6 +8,9 @@ import (
 	"time"
 )
 
+// OverrideHeaders defines overridable response heaedrs in
+// a request using a SAS URI.
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
 type OverrideHeaders struct {
 	CacheControl       string
 	ContentDisposition string
@@ -16,12 +19,17 @@ type OverrideHeaders struct {
 	ContentType        string
 }
 
+// BlobSASOptions are options to construct a blob SAS
+// URI.
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
 type BlobSASOptions struct {
 	BlobSASPermissions
 	OverrideHeaders
 	SASOptions
 }
 
+// SASOptions includes options used by SAS URIs for diffrenet
+// services and resources.
 type SASOptions struct {
 	APIVersion string
 	Start      time.Time
@@ -31,6 +39,8 @@ type SASOptions struct {
 	Identifier string
 }
 
+// BlobSASPermissions includes the available permissions for
+// a blob SAS URI.
 type BlobSASPermissions struct {
 	Read   bool
 	Add    bool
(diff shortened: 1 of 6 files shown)


storage/container.go
```diff
@@ -37,18 +41,60 @@ func (c *Container) GetURL() string {
...
+type ContainerSASOptions struct {
+	ContainerSASPermissions
+	OverrideHeaders
+	SASOptions
 }
 
-func (c *Container) GetSASURIWithSignedIPAndProtocol(expiry time.Time, permissions string, signedIPRange string, HTTPSOnly bool) (string, error) {
+// ContainerSASPermissions includes the available permissions for
+// a container SAS URI.
+type ContainerSASPermissions struct {
```
jhendrixMSFT: Seems like this could be a superset of BlobSASPermissions (or perhaps some other "base" type) to avoid duplication.

storage/container.go
```diff
@@ -37,18 +41,60 @@ func (c *Container) GetURL() string {
...
 	uri := c.GetURL()
 	signedResource := "c"
 	canonicalizedResource, err := c.bsc.client.buildCanonicalizedResource(uri, c.bsc.auth)
 	if err != nil {
 		return "", err
 	}
-	return c.bsc.client.commonSASURI(expiry, uri, permissions, signedIPRange, canonicalizedResource, signedResource, HTTPSOnly)
+
+	// build permissions string
+	permissions := ""
```
jhendrixMSFT: Move this logic into a base type and make it a method so you don't have to repeat it.

storage/blobsasuri.go
```diff
@@ -8,25 +8,87 @@ import (
...
+// BlobSASOptions are options to construct a blob SAS
+// URI.
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
+type BlobSASOptions struct {
+	BlobSASPermissions
+	OverrideHeaders
+	SASOptions
+}
+
+// SASOptions includes options used by SAS URIs for diffrenet
```
kpfaulkner: different :)

storage/blobsasuri.go
```diff
@@ -8,25 +8,87 @@ import (
...
+	}
+	if options.Delete {
+		permissions += "d"
+	}
+
+	return b.Container.bsc.client.commonSASURI(options.SASOptions, uri, permissions, canonicalizedResource, signedResource, options.OverrideHeaders)
 }
 
-func (c *Client) commonSASURI(expiry time.Time, uri, permissions, signedIPRange, canonicalizedResource, signedResource string, HTTPSOnly bool) (string, error) {
+func (c *Client) commonSASURI(options SASOptions, uri, permissions, canonicalizedResource, signedResource string, headers OverrideHeaders) (string, error) {
```
kpfaulkner: commonSASURI isn't really completely common. It's valid to use for blob and fileservice but not tables or queues (unless we add a few conditionals in there). Maybe rename it to reflect blob/fileservice usage only?

storage/container.go
mcardosos: Fixed

storage/container.go
mcardosos: Fixed

storage/blobsasuri.go
mcardosos: Fixed

storage/blobsasuri.go
mcardosos: Fixed

storage/blobsasuri.go
```diff
@@ -8,25 +8,90 @@ import (
...
+	return permissions
+}
+
+// GetSASURI creates an URL to the blob which contains the Shared
+// Access Signature with the specified options.
 //
-// See https://msdn.microsoft.com/en-us/library/azure/ee395415.aspx
-func (b *Blob) GetSASURIWithSignedIPAndProtocol(expiry time.Time, permissions string, signedIPRange string, HTTPSOnly bool) (string, error) {
+// See https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
+func (b *Blob) GetSASURI(options BlobSASOptions) (string, error) {
```
jhendrixMSFT: Be sure to document this breaking change in the change log.

PR #664, opened 2017-06-23
Fix byte range string

diptadas:
Use **"bytes=start-"** instead of **"bytes=start-end"** as byte range string if **BlobRange.End** is not specified or, set to zero.


mcardosos:
Hello @diptadas !
Thanks for submitting a PR!
Right now Travis is failing because the PR did not pass the storage tests. Don't worry, it is not a problem with your code. It is not complicated to make Travis pass in this PR :)
For a little bit of context, the SDK uses HTTP recordings to test the storage package in CI.
1. Delete file `storage\recordings\StorageBlobSuite\TestGetBlobRange.yaml`.
1. Set env vars `ACCOUNT_NAME` and `ACCOUNT_KEY` to your test storage account name and key.
1. Run the tests again: `go test ./storage/...`. You will get a slightly different `TestGetBlobRange.yaml` file.
With this change you are ready to update the PR. Travis should go green after this.

diptadas:
@mcardosos  I have generated ```TestGetBlobRange.yaml``` as you said, but still Travis is failing, would you please take a look?

mcardosos:
Investigating...

mcardosos:
So the new test cases have failed. The error from Azure is `StatusCode=400, ErrorCode=InvalidHeaderValue, ErrorMessage=The value for one of the HTTP headers is not in the correct format.`. I still don't know what is wrong in the headers.

tamalsaha:
I tried using curl on a public blob on Azure. That seems to work with range headers.

mcardosos:
Hello again @diptadas !
I just merged a small change that should fix the tests :) If you pull master and rerecord the test again, Travis should go green.

mcardosos:
@diptadas Thanks for the awesome contribution!
@tamalsaha Yes, the problem was another header was not really needed :D Good comments!

PR #684, opened 2017-07-11
Fix Snapshot parameter format

cswarth:
Attempting to use `Blob.Get()` to retrieve blob snapshots fails.
`addSnapshot()` is using the wrong format for the timestamp parameter.

This PR fixes `addSnapshot()` to use the correct timestamp format.


mcardosos:
Hello @cswarth ! Thanks for your contribution!

PR #702, opened 2017-07-28
Add ContentType to PutBlobOptions

ncw:
I haven't written tests for this yet but I'm happy to if you think this approach is OK.

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -32,6 +32,7 @@ type PutBlobOptions struct {
 	IfMatch           string     `header:"If-Match"`
 	IfNoneMatch       string     `header:"If-None-Match"`
 	RequestID         string     `header:"x-ms-client-request-id"`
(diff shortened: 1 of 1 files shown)


ncw:
Note that I've updated the commit with the other missing headers.

mcardosos:
Hello again @ncw !
The way you can modify this header (and many other headers) is like this...
``` go
b := container.GetBlobReference("myBlob")
b.Properties := BlobPoroperties{
    ContentType:     "application/json",
}
```

ncw:
Ah, OK!  Yes that works :-)

And that should have been obvious since I'm updating the metadata the same way...

Want me to send a PR adding a bit more docs to `CreateBlockBlobFromReader`?

mcardosos:
@ncw PRs are always welcome :D

ncw:
OK I'll send some changes to the docs for the bits I found confusing!

Thanks for your help.

PR #703, opened 2017-07-28
Add Range to GetBlobOptions and fix return response

ncw:
I haven't written tests for this yet but I'm happy to if you think this approach is OK.


mcardosos:
Hello @ncw !
The operation you are looking for is [`GetRange`](https://github.com/Azure/azure-sdk-for-go/blob/220faf9a2337c0cfbc7d20291c8870cb23e184b5/storage/blob.go#L209), and it is already supported.

ncw:
@mcardosos thank you - I've tested that and it works fine :-)

mcardosos:
@ncw Awesome!
In any case, thanks for submitting a PR :D

Issue #683, opened 2017-07-09
AuthenticationFailed error in PutBlockWithLength


cswarth:
storage.PutBlockWithLength() returns AuthenticationFailed error when passed an [io.LimitedReader](https://golang.org/pkg/io/#LimitedReader), but works when passed an [io.Reader](https://golang.org/pkg/io/#Reader).
```
--- FAIL: TestBlobStorage (0.09s)
	azure_test.go:30: storage: service returned error: StatusCode=403, ErrorCode=AuthenticationFailed, ErrorMessage=Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
		RequestId:d54be8b8-0001-00b9-15fc-f8fc17000000
		Time:2017-07-09T21:45:36.8668954Z, RequestInitiated=Sun, 09 Jul 2017 21:45:36 GMT, RequestId=d54be8b8-0001-00b9-15fc-f8fc17000000, API Version=, QueryParameterName=, QueryParameterValue=
```
The substitution of io.LimitedReader for io.Reader should be transparent to PutBlockWithLength().

Substitute your own account name, access key, and pre-made container name in the follow code and note the failure. Run with `go test`.

```
package azure

import (
	"io"
	"strings"
	"testing"

	"github.com/Azure/azure-sdk-for-go/storage"
)

func TestBlobStorage(t *testing.T) {

	accountName := "account_name"
	accountKey := "account_key"
	container := "preconstructed_container"

	azsClient, err := storage.NewBasicClient(accountName, accountKey)
	if err != nil {
		t.Fatal(err)
	}

	blobsvc := azsClient.GetBlobService()
	contref := blobsvc.GetContainerReference(container)
	b := contref.GetBlobReference("some/key")

	// this succeeds.
	//lr := strings.NewReader("abcde")

	// this fails with AuthenticationFailed
	lr := io.LimitReader(strings.NewReader("abcde"), 5)

	err = b.PutBlockWithLength("0000", 5, lr, nil)
	if err != nil {
		t.Fatal(err)
	}
}
```

majedev:
[...]

PR #704, opened 2017-07-28
Ensure content length is set when body is a io.LimitedReader.

jhendrixMSFT:
The ctor for http.Request will perform type assertions for a handful of
types in order to set the content length.  Unfortunately it doesn't test
for io.LimitedReader so a check has been added.
Fix for issue https://github.com/Azure/azure-sdk-for-go/issues/683

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -398,6 +398,20 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 		return nil, errors.New("azure/storage: error creating request: " + err.Error())
 	}
 
+	// if a body was provided ensure that the content length was set.
+	// http.NewRequest() will automatically do this for a handful of types
+	// and for those that it doesn't we will handle here.
+	if body != nil && req.ContentLength < 1 {
+		if lr, ok := body.(*io.LimitedReader); ok {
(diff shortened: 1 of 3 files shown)


storage/client.go
```diff
@@ -398,6 +398,15 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 		return nil, errors.New("azure/storage: error creating request: " + err.Error())
 	}
 
+	// if a body was provided ensure that the content length was set.
```
mcardosos: I could be wrong, but since this is to fix authentication, doesn't this need to be place right before adding that header?

jhendrixMSFT: I would have thought so however the fix does work.  I will dig into this more to understand why.

jhendrixMSFT: The Content-Length header was set in PutBlockWithLength so it's included when the authentication token is created.

majedev: This only works for io.LimitReader (plus the other reader types handled by `http.NewRequest()`). I suggest parsing the value from `headers["Content-Length"]` and use it (since it's passed down from `PutBlockWithLength`). This will make it more flexible and cover any type of io.Reader.

PR #707, opened 2017-07-31
gen.go uses new AutoRest CLI

mcardosos:
Fixing stuff by removing code :O
PTAL @marstr @jhendrixMSFT 
Would this break you @sergey-shandar ?
Thanks for all your patience here @olydis


sergey-shandar:
Go code generation for live tests are broken antway https://github.com/Azure/azure-rest-api-specs-tests/issues/58 and I don't plan to fix it for now. We will fix it when we have better .Net live test coverage. So no blockers from me.

Gododir/gen.go
```diff
@@ -550,26 +272,30 @@ func generateTasks(p *do.Project) {
 }
 
 func generate(service *service) {
-	codegen := "Go"
+	codegen := "--go"
 	if testGen {
-		codegen = "Go.TestGen"
+		codegen = "--go.namespace=TestGen"
```
olydis: pretty sure that's not an equivalent translation

mcardosos: @sergey-shandar Hello~

olydis: think it should be `--go.testgen`

Gododir/gen.go
```diff
@@ -563,26 +273,29 @@ func generateTasks(p *do.Project) {
...
 	execCommand := "autorest"
 	commandArgs := []string{
-		"-Input", filepath.Join(swaggersDir, "azure-rest-api-specs", service.Input+"."+string(service.Extension)),
-		"-CodeGenerator", codegen,
-		"-Header", "MICROSOFT_APACHE",
-		"-Namespace", service.Name,
-		"-OutputDirectory", service.Output,
-		"-Modeler", string(service.Modeler),
-		"-PackageVersion", sdkVersion,
+		fmt.Sprintf("https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/%s", service.Input),
```
jhendrixMSFT: So I take it we'll be generating from github instead of having to clone the specs repo.  Can we make this configurable with a default to github?  Also when we generate from github can we retrieve the hash so we can place it in the change log?

Gododir/gen.go
```diff
@@ -563,26 +273,29 @@ func generateTasks(p *do.Project) {
...
+		codegen = "--go.testgen"
 		service.Fullname = strings.Join([]string{service.Fullname, testsSubDir}, string(os.PathSeparator))
 		service.Output = filepath.Join(service.Output, testsSubDir)
 	}
 
-	fmt.Printf("Generating %s...\n\n", service.Fullname)
-
-	delete(service)
+	fmt.Printf("Working on %s...\n\n", service.Fullname)
+	fmt.Printf("Generating on %s...\n\n", service.Fullname)
```
jhendrixMSFT: Seems like this is a dupe of the previous line?

Gododir/gen.go
mcardosos: Okok, I will add the option to regen from local repo

sergey-shandar: Agree. Another way is to keep hash/commitId in configuration and then update it using specific command. I've implement the mechanism in Azure SDK for .NET (not sure if it's used) and I need something similar for PHP.

olydis: on the SDK for .Net repo it now records the commit ID and everything necessary to repro the generation: https://github.com/Azure/azure-sdk-for-net/tree/psSdkJson6/src/SDKs/_metadata (only SQL seems to have updated its SDK since then). See the `readme.md` in that folder for further detail 😉

Gododir/gen.go
mcardosos: Fixed

PR #709, opened 2017-07-31
V10.2.1

mcardosos:
PTAL
@marstr @jhendrixMSFT


arm/mysql/models.go
```diff
@@ -232,31 +316,108 @@ type ServerProperties struct {
 	FullyQualifiedDomainName *string            `json:"fullyQualifiedDomainName,omitempty"`
 }
 
+// IServerPropertiesForCreate is interface used for polymorphic Properties in ServerForCreate
+type IServerPropertiesForCreate interface {
```
jhendrixMSFT: The 'I' prefix is a C#-ism, I think we should drop it and simply call the type ServerPropertiesForCreate.

mcardosos: Fixed

arm/postgresql/models.go
```diff
@@ -210,10 +223,81 @@ type Server struct {
...
+		if err != nil {
+			return err
+		}
+		sfc.Tags = &tags
+	}
+
+	return nil
+}
+
+func unmarshalIServerPropertiesForCreate(b []byte) (ServerPropertiesForCreate, error) {
```
marstr: Shouldn't this be renamed to not include the "I" in the SerePropertiesForCreate?

mcardosos: Fixed

PR #710, opened 2017-07-31
Adding "latest" profile.

marstr:
This is an example of what a profile would look like in the world where our SDK supported multiple API Versions.


profile/latest/advisor/management/advisor/models.go
```diff
@@ -0,0 +1,40 @@
+package advisor
+
+import (
+	 original "github.com/Azure/azure-sdk-for-go/service/advisor/management/2017-04-19/advisor"
+)
+
+type (
+	 Category = original.Category
```
mcardosos: >Category = original.Category [](start = 2, length = 28)

qq, would golint pass? In the alias case, do exported structs need to be include a comment?

profile/latest/advisor/management/advisor/models.go
```diff
@@ -0,0 +1,40 @@
+package advisor
+
+import (
+	 original "github.com/Azure/azure-sdk-for-go/service/advisor/management/2017-04-19/advisor"
```
mcardosos: I am not a big fan of duplicating `advisor` in the import path, however, I think I understand why you did it that way.

jhendrixMSFT:
What about methods?

marstr:
Aha, good point @jhendrixMSFT. Functions that aren't hanging off of a type aren't represented in this PR. I'll add those right-quick.

profile/latest/advisor/management/advisor/models.go
marstr: Yep, agreed in general I do think there is some unfortunate redundancy in 90% of our packages. However there are two main reasons that I structure the packages this way:
1. Readability: If the last element of the path here were "2017-04-19" that would mean folks would have to reference the package as "2017" all throughout their code; that approach is clearly ridiculous.
2. Not all packages have this redundancy. Compute for example has two sets of packages: `compute` & `containerService`. To disambiguate in these cases and have consistent structure, we must have some redundancy.

profile/latest/advisor/management/advisor/models.go
marstr: So, it's hard to say because golint hasn't been updated to handle type aliases yet. At the moment it fails. However, I'm pretty certain that the stuttering shouldn't be triggered here. Even if it was, the value of having the type names be the same is high enough we would ignore those violations.

In terms of comments, it would be pretty easy to modify my tool to pull in the original comments. I'll make sure that happens for before this hits master.

profile/latest/advisor/management/advisor/models.go
mcardosos: qq, for `compute`, are you generating `compute` and `containerservice` separately? I tink this would be a good moment to merge `containerservice` and `disks` back to `compute`

marstr: Can't do it! With multi-APIVersion support, they must be separate by design. :) Composite swaggers were a huge mistake that should never have come around.

mcardosos: Oh well, 💩 
But disk can me merged back right?

PR #712, opened 2017-08-01
Improve storage docs

ncw:
I've added some bits to the docs for things I found confusing when doing my integration.

### DIFF
--- a/storage/blockblob.go
+++ b/storage/blockblob.go
@@ -68,6 +68,8 @@ type BlockResponse struct {
 
 // CreateBlockBlob initializes an empty block blob with no blocks.
 //
+// See CreateBlockBlobFromReader for more info on creating blobs.
+//
 // See https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/Put-Blob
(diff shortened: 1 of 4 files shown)


storage/blockblob.go
```diff
@@ -77,10 +79,17 @@ func (b *Blob) CreateBlockBlob(options *PutBlobOptions) error {
...
 // create an empty blob, use size==0 and reader==nil.
 //
+// Any headers set in blob.Properties or metadata in blob.Metadata
+// will be set on the blob.
+//
 // The API rejects requests with size > 256 MiB (but this limit is not
 // checked by the SDK). To write a larger blob, use CreateBlockBlob,
 // PutBlock, and PutBlockList.
 //
+// To create a blob from scratch, call container.GetBlobReference() to
```
marstr: @mcardosos, can you confirm that this is the preferable way to handle this?

mcardosos: This is correct

---
In reply to: [130678050](https://github.com/Azure/azure-sdk-for-go/pull/712#discussion_r130678050) [](ancestors = 130678050)

mcardosos:
Awesome!
Thanks for contributing @ncw !

PR #724, opened 2017-08-14
Fix build errors on earlier versions of Go.

jhendrixMSFT:
Add Go 1.6 and 1.7 to travis.
Refactored some 1.8-specific code into version-specific files.
Updated dependencies.
Disable table batch tests on Go 1.6; this is because the MIME multipart
writer fixed the sort order on headers starting in 1.7 so the recordings
are not comparible between 1.6 and later versions.
Add Go 1.7+ build tags to management package.
Updated README for management package indicating that it's deprecated.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,8 @@
-hash: bdeb606eca05304f80e5577159227ae4f00a002213f353e304066c593419a31b
-updated: 2017-07-31T15:32:36.5529621-07:00
+hash: f52476dff67b66fe9fffe35b63ac0c06962fcaacd326b66820a33edf53b01a16
+updated: 2017-08-14T15:40:55.888398-07:00
 imports:
 - name: github.com/Azure/go-autorest
-  version: f6e08fe5e4d45c9a66e40196d3fed5f37331d224
+  version: 48a8154950f853ae1f2a9c409e0eb1fb71119bc8
   subpackages:
   - autorest
   - autorest/adal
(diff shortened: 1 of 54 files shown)


.travis.yml
```diff
@@ -23,12 +25,14 @@ script:
   - test -z "$(go vet $(find ./arm/* -type d -print) | tee /dev/stderr)"
   - test -z "$(golint ./arm/... | tee /dev/stderr)"
   - go build ./arm/examples/...
-  - go test -v ./management/...
+  - if [[ $TRAVIS_GO_VERSION != 1.6* ]]; then
```
mcardosos: I would explain why ASM is not supported in 1.6

marstr: I worry that this fix won't help people using go 1.6 who are installing our SDK as such:

`go get -u github.com/Azure/azure-sdk-for-go/...`

If we're willing to pay that price, I'm open to it.

jhendrixMSFT: @mcardosos I will update the README for ASM to include a note about requiring Go 1.7+ (this dependency was introduced in commit 2cb6120 and since it's legacy I don't think it's worth investing the effort to get it working on 1.6 again).
@marstr What is the error that will happen in this case?

mcardosos: :D

marstr: Ah, I think we're alluding to the same thing. For folks who install the SDK with the command I mention above, it'll will try to build each package, including the management ones. They'll get the same error that Travis was running into.

We can solve that by documenting and have folks expect that err, or we could add build tags to all of the packages in management. :) I imagine a little scripting could make that happen in no time flat.

jhendrixMSFT: I see your point, will fix.

PR #735, opened 2017-08-23
Speling

md2perpe:
"opptions" --> "options"


marstr:
Nice catch, thank you @md2perpe .

PR #737, opened 2017-08-23
Adding required Apache NOTICE file.

marstr:
I created this NOTICE file using the guidance found here: https://www.apache.org/dev/licensing-howto.html#simple

I specifically believe that we do not qualify as "bundling" our dependencies, therefore our NOTICE file MUST NOT include references to our dependencies.


salameer:
Related to https://github.com/Azure/azure-sdk-for-go/issues/738

marstr:
Giving up on all y'all! I think it looks good, so I'm going to merge this.

PR #740, opened 2017-08-23
Delete Terraform job from Travis

mcardosos:
Deleting the Terraform build from travis cron jobs.
However, the same job will stay on our internal Jenkins.
@marstr @jhendrixMSFT


marstr:
@tombuildsstuff, just a heads up if you didn't see this thread :)

As @mcardosos says above, we're just changing up the procedure of where we run these tests, not actually disregarding Terraform.

Issue #731, opened 2017-08-21
Generated wrong SAS url for blob service using emulator


zyzheng:
I am trying to generate a SAS url for blob service to upload a file. I tried both this sdk and [Azure Storage Explorer](https://azure.microsoft.com/en-us/features/storage-explorer/). The url (sig part) generated are different and the one from this sdk cannot upload file with curl.

Here is my code to generate the sas url (skip error handling):
```
client, _ := storage.NewEmulatorClient()
blobService := client.GetBlobService()
container := blobService.GetContainerReference("testfolder")
options := storage.CreateContainerOptions{Access: storage.ContainerAccessTypePrivate}
container.CreateIfNotExists(&options)
blob := container.GetBlobReference(name)
blob.CreateBlockBlob(nil)
url, _ := blob.GetSASURI(time.Date(2017, 10, 1, 0, 0, 0, 0, time.Local), "w")
```

Generated url is `http://127.0.0.1:10000/devstoreaccount1/testfolder/testfile?se=2017-09-30T16%3A00%3A00Z&sig=sVcV11mVeHkfUnjwMDInOyVn4VYpb%2BycdjISXBRICEs%3D&sp=w&spr=https%2Chttp&sr=b&sv=2016-05-31`.

Here is how I generated sas url from Azure Storage Explorer:
![default](https://user-images.githubusercontent.com/6612754/29502032-89b14f9c-865f-11e7-81b7-705fbbeb8232.png)
Generated url is `http://127.0.0.1:10000/devstoreaccount1/testfolder/testfile?se=2017-09-30T16%3A00%3A00Z&sp=w&sv=2016-05-31&sr=b&sig=Tyrg2ccc0RXyRz5xfkcSVDvjjoRivygrGb%2ByTLf0jJY%3D`.

The `sig` part is different. Now I use curl to test the two urls:
```
# the url from go sdk
> curl -XPUT -H "Content-Length:11" -d "hello world" -H "x-ms-blob-type:BlockBlob" "http://127.0.0.1:10000/devstoreaccount1/testfolder/testfile?se=2017-09-30T16%3A00%3A00Z&sig=sVcV11mVeHkfUnjwMDInOyVn4VYpb%2BycdjISXBRICEs%3D&sp=w&spr=https%2Chttp&sr=b&sv=2016-05-31"
[...]

PR #745, opened 2017-08-29
Fixes stringToSign for SASURI

mcardosos:
PTAL @jhendrixMSFT @marstr 
Fixes https://github.com/Azure/azure-sdk-for-go/issues/731

### DIFF
--- a/storage/blobsasuri.go
+++ b/storage/blobsasuri.go
@@ -44,7 +44,7 @@ func (b *Blob) GetSASURIWithSignedIPAndProtocol(expiry time.Time, permissions st
 		signedResource = "b"
 	}
 
-	protocols := "https,http"
+	protocols := ""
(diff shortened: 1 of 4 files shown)


storage/blobsasuri.go
```diff
@@ -44,7 +44,7 @@ func (b *Blob) GetSASURIWithSignedIPAndProtocol(expiry time.Time, permissions st
 		signedResource = "b"
 	}
 
-	protocols := "https,http"
+	protocols := ""
```
jhendrixMSFT: Why we don't need "http" anymore?

mcardosos: The protocols are set to `https,http` by default, so setting them explicitly is not really needed.
On the other hand, this is also useful to generate exactly the same SAS as the Storage Explorer does.

PR #750, opened 2017-08-31
Upgrading to Go 1.9

tombuildsstuff:
👋 

I'm upgrading Terraform to Go 1.9 and noticed the SDK wasn't building against this yet, as such I've updated the Travis file to do this. I'm not sure if this needs to be updated anywhere else, but hopefully this is useful :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 language: go
 
 go:
+  - 1.8
   - 1.9
 
 install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 language: go
 
 go:
-  - 1.8
+  - 1.9
```
marstr: No reason to move to 1.9, might as well have both 1.8 and 1.9.

Other than that, thanks for doing this, @tombuildsstuff!

tombuildsstuff: @marstr good point - updated to include 1.8 too :)

Issue #752, opened 2017-09-01
Incorrect package name on datalake-store/filesystem


mcardosos:
As explained by @kirankumarbv in https://github.com/Azure/azure-sdk-for-go/pull/729

> Hey,
You have changed the file /root/godep/src/github.com/Azure/azure-sdk-for-go/datalake-store/filesystem/version.go to have package datalakestore instead of package filesystem, but client.go still has package filesystem which caused this error in my system.
can't load package: package github.com/Azure/azure-sdk-for-go/datalake-store/filesystem: found packages filesystem (client.go) and datalakestore (version.go) in github.com/Azure/azure-sdk-for-go/datalake-store/filesystem

PR #753, opened 2017-09-01
Update and fix version.go files

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/752
PTAL
@jhendrixMSFT @marstr


Gododir/gen.go
```diff
@@ -405,7 +406,9 @@ func managementVersion(c *do.Context) {
 func version(packageName string) {
 	versionFile := filepath.Join(packageName, "version.go")
 	os.Remove(versionFile)
-	template := `package %s
+	template := `// +build go1.7
```
jhendrixMSFT: Why the build constraint (not pushing back just wondering)?

mcardosos: This build constraint was added to `version.go` some time ago by hand. But as `gen.go` gens that file and overwrites, well, I included this so there would be no diff.

jhendrixMSFT: Makes sense.

Issue #748, opened 2017-08-29
Handling of API Errors in the SDK


tombuildsstuff:
:wave:

I've been meaning to raise this one for a while.. in this case I'm using the SQL Databases SDK - however this also applies to CosmosDB Accounts and I believe Container Service (amongst other services) too.

When making the following request using the SQL Databases SDK with an invalid configuration:

```
PUT /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctest_rg_5909366699526987910/providers/Microsoft.Sql/servers/acctestsqlserver5909366699526987910/databases/acctestdb5909366699526987910?api-version=2014-04-01 HTTP/1.1
Host: management.azure.com
User-Agent: HashiCorp-Terraform-v0.10.0-dev
Content-Length: 179
Content-Type: application/json
Accept-Encoding: gzip

{
	"tags": {},
	"location": "eastus",
	"properties": {
		"collation": "SQL_Latin1_General_CP1_CI_AS",
		"createMode": "Default",
		"edition": "DataWarehouse",
		"requestedServiceObjectiveName": "System"
	}
}
```

which returns the response from the API:

```
HTTP/1.1 400 Bad Request
Content-Length: 323
Cache-Control: no-store, no-cache
Content-Type: application/json
Dataserviceversion: 3.0;
Date: Tue, 29 Aug 2017 23:18:28 GMT
Preference-Applied: return-content
Server: Microsoft-HTTPAPI/2.0
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Content-Type-Options: nosniff
X-Ms-Correlation-Request-Id: b6059a2e-6c97-4f31-aa5f-37689644e58a
X-Ms-Ratelimit-Remaining-Subscription-Writes: 1197
X-Ms-Request-Id: 936005d0-13ee-4142-8039-43b972c2df7c
X-Ms-Routing-Request-Id: UKWEST:20170829T231829Z:b6059a2e-6c97-4f31-aa5f-37689644e58a

{
	"code": "40808",
	"message": "The edition \u0027DataWarehouse\u0027 does not support the service objective \u0027System\u0027.",
	"target": null,
	"details": [{
		"code": "40808",
		"message": "The edition \u0027DataWarehouse\u0027 does not support the service objective \u0027System\u0027.",
		"target": null,
		"severity": "16"
	}],
	"innererror": []
}
```

When checking the error response from the SDK for this call, the error returned is:

```
[...]

PR #754, opened 2017-09-04
Check `errIn`, not `err`

md2perpe:
It really looks like `errIn` should be checked, not `err`.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -437,12 +437,12 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 			// response contains storage service error object, unmarshal
 			if resp.Header.Get("Content-Type") == "application/xml" {
 				errIn := serviceErrFromXML(respBody, &storageErr)
-				if err != nil { // error unmarshaling the error response
+				if errIn != nil { // error unmarshaling the error response
 					err = errIn
 				}
 			} else {
 				errIn := serviceErrFromJSON(respBody, &storageErr)
-				if err != nil { // error unmarshaling the error response
+				if errIn != nil { // error unmarshaling the error response
 					err = errIn
 				}
 			}


storage/client.go
```diff
@@ -437,12 +437,12 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 			// response contains storage service error object, unmarshal
 			if resp.Header.Get("Content-Type") == "application/xml" {
 				errIn := serviceErrFromXML(respBody, &storageErr)
-				if err != nil { // error unmarshaling the error response
+				if errIn != nil { // error unmarshaling the error response
 					err = errIn
```
marstr: Good catch seeing the masked error. However, I'm not sure this fix goes far enough to fix it. As it is in this PR, `errIn` gets blown away anyway.

@mcardosos, what are your thoughts? Maybe we should be extending the AzureStorageServiceError to include an inner error in the case of marshaling trouble?

tombuildsstuff:
👋🏻 out of interest does this fix #748?

mcardosos:
@tombuildsstuff These issues are not related :(

storage/client.go
mcardosos: Take a look into the `serviceErrFromXML` and `serviceErrFromJSON` funcs. They both include the unmarshal errors in `AzureStorageServiceError.Message`. I actually think we could get rid of `if err != nil`. Code would look like this...
```go
if resp.Header.Get("Content-Type") == "application/xml" {
	serviceErrFromXML(respBody, &storageErr)
} else {
	serviceErrFromJSON(respBody, &storageErr)
}
```

marstr: Talking IRL with @mcardosos, she showed me that we do actually percolate up the un-marshaling error already. I like the blurb that she has above, let's do that :)

md2perpe: @marstr I noticed afterwards that the error is blown away, but thought that you need to take a look at the whole code block. As I did this at work I didn't want to put too much time into it.

mcardosos:
@md2perpe ping

md2perpe:
@mcardosos pong

marstr:
It seems like this PR has gone stale. I'm going to close it out.

PR #755, opened 2017-09-04
Support os.File as body

md2perpe:
Trying to understand why I got authentication error when doing `WriteRange` although other calls worked, I finally found that it was using `*os.File` for `body` that didn't work. The code couldn't get the content length from it. The content length is used in the authentication signature.

This patch uses the `Stat` method to find the size of a file so that files are supported as bodies.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -15,6 +15,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"runtime"
 	"strings"
@@ -403,7 +404,12 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 	// and for those that it doesn't we will handle here.
 	if body != nil && req.ContentLength < 1 {
 		if lr, ok := body.(*io.LimitedReader); ok {
-			setContentLengthFromLimitedReader(req, lr)
+			err = setContentLengthFromLimitedReader(req, lr)
+		} else if f, ok := body.(*os.File); ok {
+			err = setContentLengthFromFile(req, f)
+		}
+		if err != nil {
+			return nil, err
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


storage/client.go
```diff
@@ -403,7 +404,12 @@ func (c Client) exec(verb, url string, headers map[string]string, body io.Reader
 	// and for those that it doesn't we will handle here.
 	if body != nil && req.ContentLength < 1 {
 		if lr, ok := body.(*io.LimitedReader); ok {
-			setContentLengthFromLimitedReader(req, lr)
+			err = setContentLengthFromLimitedReader(req, lr)
+		} else if f, ok := body.(*os.File); ok {
```
mcardosos: At this point, a type switch is cleaner.

storage/util_1.7.go
```diff
@@ -5,8 +5,19 @@ package storage
 import (
 	"io"
 	"net/http"
+	"os"
 )
 
-func setContentLengthFromLimitedReader(req *http.Request, lr *io.LimitedReader) {
+func setContentLengthFromLimitedReader(req *http.Request, lr *io.LimitedReader) error {
 	req.ContentLength = lr.N
+	return nil
```
mcardosos: Changes to this func are not really needed.

storage/util_1.7.go
```diff
@@ -5,8 +5,19 @@ package storage
...
+}
+
+func setContentLengthFromFile(req *http.Request, f *os.File) error {
```
mcardosos: This func would better live in `client.go` file. This file is for code that would only work on certain go versions.

mcardosos:
@md2perpe ping

md2perpe:
@mcardosos pong

marstr:
Howdy Folks, 

This PR is stale. I'm going to close it out, but if there are still important contributions, feel free to either open another or ask us to re-open this. Either way, if these changes are going to be brought in, please address the review comments.

Cheers!

PR #766, opened 2017-09-11
generates SDK is parallel, also reports failures at the end

mcardosos:
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/Gododir/gen.go
+++ b/Gododir/gen.go
@@ -239,8 +239,8 @@ var (
 			},
 		},
 	}
-	mutex = &sync.Mutex{}
-	fails []string
+	failLock = &sync.Mutex{}
+	fails    []string
 )
 
 func init() {
@@ -478,9 +478,9 @@ func runner(cmd *exec.Cmd) (string, string, error) {
 }
 
 func addFail(fail string) {
-	mutex.Lock()
+	failLock.Lock()
 	fails = append(fails, fail)
-	mutex.Unlock()
+	failLock.Unlock()
 }
 
 func report(c *do.Context) {
--- a/Gododir/gen.go
+++ b/Gododir/gen.go
@@ -37,6 +37,13 @@ type mapping struct {
 	Services    []service
 }
 
+type failList []string
+
+type failLocker struct {
+	sync.Mutex
+	failList
+}
+
 var (
 	start           time.Time
 	gopath          = os.Getenv("GOPATH")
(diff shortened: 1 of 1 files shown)


Gododir/gen.go
```diff
@@ -36,12 +37,13 @@ type mapping struct {
...
 
 var (
+	start           time.Time
 	gopath          = os.Getenv("GOPATH")
 	sdkVersion      string
 	autorestDir     string
 	swaggersDir     string
 	testGen         bool
-	deps            do.S
+	deps            do.P
```
jhendrixMSFT: This is the magic that makes it parallel?

mcardosos: Yes

Gododir/gen.go
```diff
@@ -318,8 +328,8 @@ func generate(service *service) {
 
 	fmt.Println(commandArgs)
 
-	if _, err := runner(autorest); err != nil {
-		panic(fmt.Errorf("Autorest error: %s", err))
+	if _, stderr, err := runner(autorest); err != nil {
+		failed = append(failed, fmt.Sprintf("%s: autorest error: %s: %s", service.Fullname, err, stderr))
```
marstr: As mentioned offline, this is a race condition 😄

mcardosos: Fixed

Gododir/gen.go
```diff
@@ -216,24 +239,36 @@ var (
 			},
 		},
 	}
+	mutex = &sync.Mutex{}
```
marstr: Let's rename this to be a little more specific. Maybe `fails_lock`? I'm relatively certain I've seen somewhere that there is a standard idiom for this exact sorta scenario, but I can't remember exactly what it was. :/

mcardosos: Fixed

Gododir/gen.go
```diff
@@ -216,24 +239,36 @@ var (
 			},
 		},
 	}
+	failLock = &sync.Mutex{}
+	fails    []string
```
jhendrixMSFT: This is mostly academic in this case, but seeing that these two objects are tightly coupled I would put them in a struct and provide methods for getting/setting so that any invariants can be transparently maintained.

marstr: Even better, then you wouldn't need to name the Mutex. You'd be able to Lock/Unlock directly off of the object.

mcardosos: Fixed

PR #767, opened 2017-09-11
Introducing profileBuilder and rewriting generator

marstr:
This is laying the foundation to move to the Multiple API Version and Profile future of our API. Once this is accepted, I'll send another PR with the generated new packages including each API Version and profile.

You can see examples of what the generated code looks like here:
https://github.com/Azure/azure-sdk-for-go/tree/experimental/allAPIVersions


glide.lock
```diff
@@ -1,5 +1,10 @@
+<<<<<<< HEAD
 hash: 7a8f8652a17526f090124ace3974ce9bff3d6d09a83ab94841d377c1125111f7
 updated: 2017-06-12T16:27:58.5460985-07:00
+=======
+hash: fc11c273a4bfae9a077f2238a4693fa3af3d815dfc1b1e9ec75d9eb8e6428993
+updated: 2017-05-23T10:47:59.5894489-07:00
+>>>>>>> c9acadb... Fixing glide defs to include Masterminds/semver
```
mcardosos: Merge branches failed

tools/generator/program.go
```diff
@@ -0,0 +1,424 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/Masterminds/semver"
+)
+
+const (
+	defaultRemoteAzureRestAPISpecsPath = "https://github.com/Azure/azure-rest-api-specs.git"
+	defaultAzureRESTAPIBranch          = "master"
```
mcardosos: >master [](start = 39, length = 6)

The branch we use is "current"

tools/generator/program.go
```diff
@@ -0,0 +1,424 @@
...
+
+	fmt.Println("Summary:")
+	fmt.Println("\tFound:    \t", found)
+	fmt.Println("\tGenerated:\t", generated)
+	fmt.Println("\tFormatted:\t", formatted)
+	fmt.Println("\tVetted:   \t", vetted)
+}
+
+func cloneAPISpecs(origin, local string) (string, error) {
+	_, cloneLoc := filepath.Split(local)
```
mcardosos: Autorest does not need the repo to be cloned on local. You can do `autorest [https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/graphrbac/data-plane/readme.md`](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/current/specification/graphrbac/data-plane/readme.md%60).

tools/generator/program.go
```diff
@@ -0,0 +1,424 @@
...
+		formatted++
+		return vet(x.(string)) == nil
+	})
+
+	for range processor {
+		vetted++
+	}
+
+	fmt.Println("Summary:")
+	fmt.Println("\tFound:    \t", found)
```
mcardosos: Would be useful to also `go build` packages, I think.

tools/generator/program.go
```diff
@@ -0,0 +1,424 @@
...
+	autorest := exec.Command(
+		"autorest",
+		"-Input", swag.Path,
+		"-CodeGenerator", "Go",
+		"-Header", "MICROSOFT_APACHE",
+		"-Namespace", namespace[strings.LastIndex(namespace, "/")+1:],
+		"-OutputDirectory", outputDir,
+		"-Modeler", "Swagger",
+		"-pv", version.String(),
+		"-SkipValidation")
```
mcardosos: This is from the legacy autorest CLI

tools/generator/program.go
```diff
@@ -0,0 +1,424 @@
...
+
+func getAzureSpecsPattern(plane string) *regexp.Regexp {
+	elements := []string{
+		`specification`,
+		`(?P<package>[\w\-\d\.]+)`,
+		plane,
+		`[Mm]icrosoft\.(?P<category>[\w\-]+)`,
+		`(?P<apiVersion>v?\d{4}-\d{2}-\d{2}(?:[\-\.][\w\d\-\.]+)?)`,
+		`(?P<group>[\w\-\d\.]+)\.json`,
+	}
```
mcardosos: Will you support md files as input?

tools/profileBuilder/latest.go
```diff
@@ -0,0 +1,178 @@
...
+// LatestStrategy evaluates the Azure-SDK-for-Go repository for the latest available API versions of each service.
+type LatestStrategy struct {
+	Root      string
+	Predicate func(packageName string) bool
+}
+
+// AcceptAll is a predefined value for `LatestStrategy.Predicate` which always returns true.
+func AcceptAll(name string) (result bool) {
+	result = true
+	return
```
mcardosos: I would prefer to keep it simple and `return true`

tools/profileBuilder/latest.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// IgnorePreview searches a packages "API Version" to see if it contains the word "preview". It only returns true when a package is not a preview.
+func IgnorePreview(name string) (result bool) {
+	matches := packageName.FindStringSubmatch(name)
+	if len(matches) >= 4 {
```
mcardosos: >>= 4 [](start = 17, length = 4)

These number look a bit magical, add a comment explaining a little bit.

tools/profileBuilder/latest.go
```diff
@@ -0,0 +1,178 @@
...
+
+// ErrNotVersionString is instantiated when a string not conforming to known Azure API Version patterns is parsed is if it did.
+type ErrNotVersionString string
+
+func (err ErrNotVersionString) Error() string {
+	return fmt.Sprintf("`%s` is not a recognized Azure version string", string(err))
+}
+
+var versionle = func() func(string, string) (bool, error) {
+	versionPattern := regexp.MustCompile(`^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})(?:[\.\-](?P<tag>.+))?$`)
```
mcardosos: >versionPattern [](start = 1, length = 14)

Graphrbac uses a different pattern

tools/profileBuilder/program.go
```diff
@@ -0,0 +1,341 @@
...
+		outputLog.Print("profileBuilder Version: ", version)
+	}
+
+	// Sort out the Profile Name to be used.
+	if profileName == defaultName {
+		profileName = randname.AdjNoun{}.Generate()
+		outputLog.Print("Profile Name Set to: ", profileName)
+	}
+
+	//
```
mcardosos: >// [](start = 1, length = 2)

O.O

tools/profileBuilder/README.md
```diff
@@ -0,0 +1,48 @@
...
+go install
+```
+Taking things a step further, if you'd like the profileBuilder to stamp the version of itself in the code it generates, you can install as below:
+
+``` bash
+# bash
+go get -d github.com/Azure/azure-sdk-for-go/tools/profileBuilder
+cd $GOPATH/src/github.com/Azure/azure-sdk-for-go/tools/profileBuilder
+glide install
+export currentCommit=$(git rev-parse HEAD)
```
mcardosos: >git rev-parse HEAD [](start = 23, length = 18)

Could this be executed inside the profile builder, and just have a bool flag for including or not the commit?

marstr: Actually, it cannot be done this way :) That would result in the profileBuilder always telling you it was the most up-to-date version even if it wasn't. We really need to set this variable at build time the way I've done it.

glide.lock
marstr: Whoops whoops whoops, I'll get this cleaned up.

tools/generator/program.go
marstr: Ah, this is a new developement, that's great :)

tools/profileBuilder/latest.go
marstr: You're telling me you don't appreciate the magic in life? Sure thing, I'll add some documentation here.

I was actually quite frustrated with Go's Regexp library, because in C# you can reference groups by the name you gave them instead of some index. The way goes about it, it's quite prone to these sorts of magic numbers.

tools/profileBuilder/program.go
marstr: Fixed locally

tools/generator/program.go
marstr: Fixed locally

tools/profileBuilder/latest.go
marstr: I'll make this change because I like you and Joel. 😝

marstr: Fixed locally

mcardosos: =^w^=

tools/profileBuilder/latest.go
marstr: I think the amount of work for this versus the reward warrants fixing this in a separate PR.

tools/profileBuilder/latest.go
marstr: Fixed locally

Issue #738, opened 2017-08-23
Apache file Headers missing from many files


marstr:
All of the [Autorest](https://github.com/Azure/autorest) generated code contains the suggested Apache header, but the manually maintained files don't.

To close out this item, we should add the header to any file missing it and update Travis to flag any files that don't have it in both this repository and in [Azure/go-autorest](https://github.com/Azure/go-autorest).

PR #771, opened 2017-09-14
Headers

marstr:
Adding a check for Missing copyright headers.


marstr:
Solves #738

marstr:
Look at this build for an example of what happens when files are added without the copyright notification: https://travis-ci.org/Azure/azure-sdk-for-go/builds/275698423

PR #772, opened 2017-09-15
Add deployment helper

mcardosos:
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/arm/deployment/deployment.go
+++ b/arm/deployment/deployment.go
@@ -7,11 +7,11 @@ import (
 	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
 )
 
-// BuildDeployment is a helper that creates a resources.Deployment, which can
+// Build is a helper that creates a resources.Deployment, which can
 // be used as a parameter for a CreateOrUpdate deployment operation.
 // templateFile is a local Azure template.
 // See https://github.com/Azure-Samples/resource-manager-go-template-deployment
-func BuildDeployment(mode resources.DeploymentMode, templateFile string, parameters map[string]interface{}) (deployment resources.Deployment, err error) {
+func Build(mode resources.DeploymentMode, templateFile string, parameters map[string]interface{}) (deployment resources.Deployment, err error) {
 	template, err := parseJSONFromFile(templateFile)
 	if err != nil {
 		return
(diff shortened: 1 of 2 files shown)


arm/helpers/helpers.go
```diff
@@ -0,0 +1,47 @@
+package helpers
```
jhendrixMSFT: I think we should give this package a more meaningful name, something that describes its use.  I'm bad at naming but will throw a few out here: template, deployment.

arm/helpers/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+)
+
+// BuildDeployment is a helper method that create a deployment object
```
jhendrixMSFT: I think this needs to be beefed up (why would somebody want to call this method) and include links about ARM templates.  Also describe the return params, e.g. what do I actually do with the returned deployment object?

jhendrixMSFT:
A more general comment, this creates some data structures but doesn't actually do anything in Azure.  Do we need another method that simply takes a template file and deploys it in Azure?

arm/helpers/helpers.go
mcardosos: Fixed

arm/helpers/helpers.go
mcardosos: Fixed

arm/deployment/deployment.go
```diff
@@ -0,0 +1,49 @@
...
+	"io/ioutil"
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+)
+
+// BuildDeployment is a helper that creates a resources.Deployment, which can
+// be used as a parameter for a CreateOrUpdate deployment operation.
+// templateFile is a local Azure template.
+// See https://github.com/Azure-Samples/resource-manager-go-template-deployment
+func BuildDeployment(mode resources.DeploymentMode, templateFile string, parameters map[string]interface{}) (deployment resources.Deployment, err error) {
```
jhendrixMSFT: Since the package name is deployment I think we can change this function name to Build.

PR #777, opened 2017-09-20
feat(storage): allow creating clients from connection strings

connor4312:
Based on the [Node.js version](https://github.com/Azure/azure-storage-node/blob/a2a26c7c70acb1eee0a77fb68fc1547145afec28/lib/common/services/servicesettings.js#L710), allows instantiating storage clients from connection strings.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -179,6 +179,44 @@ func (e UnexpectedStatusCodeError) Got() int {
 	return e.got
 }
 
+// NewClientFromConnectionString creates a Client from the connection string.
+func NewClientFromConnectionString(input string) (Client, error) {
+	var (
+		accountName, accountKey, endpointSuffix string
+		useHTTPS                                = defaultUseHTTPS
+	)
+
+	for _, pair := range strings.Split(input, ";") {
+		if pair == "" {
+			continue
+		}
+
+		equalDex := strings.IndexByte(pair, '=')
+		if equalDex <= 0 {
+			return Client{}, fmt.Errorf("Invalid connection segment %q", pair)
+		}
+
+		value := pair[equalDex+1:]
+		switch strings.ToLower(pair[:equalDex]) {
+		case "accountname":
+			accountName = value
+		case "accountkey":
+			accountKey = value
+		case "endpointsuffix":
+			endpointSuffix = value
+		case "defaultendpointsprotocol":
+			useHTTPS = value == "https"
+		default:
+			// ignore
+		}
+	}
+
+	if accountName == StorageEmulatorAccountName {
+		return NewEmulatorClient()
+	}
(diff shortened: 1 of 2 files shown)


storage/client.go
```diff
@@ -179,6 +179,44 @@ func (e UnexpectedStatusCodeError) Got() int {
...
+		case "accountname":
+			accountName = value
+		case "accountkey":
+			accountKey = value
+		case "endpointsuffix":
+			endpointSuffix = value
+		case "defaultendpointsprotocol":
+			useHTTPS = value == "https"
+		default:
+			// ignore
```
marstr: Is No-oping the correct decision here? I wound have thought it would be more appropriate to error out if an unrecognized key is encountered.

storage/client.go
```diff
@@ -179,6 +179,44 @@ func (e UnexpectedStatusCodeError) Got() int {
...
+		}
+
+		equalDex := strings.IndexByte(pair, '=')
+		if equalDex <= 0 {
+			return Client{}, fmt.Errorf("Invalid connection segment %q", pair)
+		}
+
+		value := pair[equalDex+1:]
+		switch strings.ToLower(pair[:equalDex]) {
+		case "accountname":
```
marstr: I'd pull these out into consts.

storage/client.go
```diff
@@ -179,6 +179,44 @@ func (e UnexpectedStatusCodeError) Got() int {
...
+	)
+
+	for _, pair := range strings.Split(input, ";") {
+		if pair == "" {
+			continue
+		}
+
+		equalDex := strings.IndexByte(pair, '=')
+		if equalDex <= 0 {
+			return Client{}, fmt.Errorf("Invalid connection segment %q", pair)
```
marstr: Nit: we know this is a string, so I'd rather have the output format surround it in quotes and have the format specifier `%s`.

connor4312: `%q` does format the parameter correct:

> %q	a double-quoted string safely escaped with Go syntax

For this reason I generally prefer to use it over manually quoting a %s; there's no confusion if the string happens to also contain quotes.

marstr: Ah, beautiful. Thanks.

marstr: For people curious: https://play.golang.org/p/lVBlTRB2OH

storage/client.go
marstr: @jhendrixMSFT / @mcardosos, mind taking a look to see if there are other possible values here?

mcardosos: Looks like there are a bunch of other possible values
https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string

connor4312: Thanks for looking into it! I'll undo my change which errors in this case.

Issue #765, opened 2017-09-11
Incorrect formatting of x-ms-blob-content-length header in blob.SetProperties


kvstark:
When trying to set the ContentLength of a blob via the SetProperties() call, I get a 400 'header formatted incorrectly' error.  The queryParameterName and queryParameterValue are both blank.  When I modify the code to print the actual return error I see that the problematic header is the x-ms-blob-content-length.  I checked the blob type is BlobTypePage and the other headers seem formatted correctly.  The SetProperties() call works as expected if I remove the ContentLength field from the blob.

I believe that when attempting to update ContentLength in SetProperties the word "byte" should NOT be included before the new length.  The documention specifies [here](https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-properties) that the request header format is x-ms-blob-content-length: byte value.  Though it's a bit confusing, I suggest that what the documentation intends is byte_value (value in bytes).  I eventually came to this conclusion via testing and by looking at the azure-sdk for other languages.

Ruby in particular, in [blob.rb](https://github.com/Azure/azure-storage-ruby/blob/master/lib/azure/storage/blob/blob.rb) sets up its content lenght header for SetProperties using the following line:
` StorageService.with_header headers, 'x-ms-blob-content-length', options[:content_length].to_s if options[:content_length]`
No mention of 'byte' anywhere.

When I removed the word 'byte' from the line in storage/blob.go containing 
`headers = addToHeaders(headers, "x-ms-blob-content-length", fmt.Sprintf("byte %v", b.Properties.ContentLength))`
then the function ceased returning errors and correctly updated the size of the blob.

kvstark:
I would like to additionally note that directly after the line mentioned above in SetProperties in blob.go is a line that crashes.
`if options != nil || options.SequenceNumberAction != nil` will crash if options is nil.  I believe this should have been an && instead of ||.

kvstark:
Pull request submitted with the changes specified above.

kvstark:
Thanks so much for the quick turn around on the merge commit!

PR #781, opened 2017-09-21
blob.SetProperties fixes (Issue 765)

kvstark:
SetProperties uses correct format for ContentLength header.
options.SequenceNumberAction is only accessed when options != nil.

See comments in Issue 765 for detailed description.  Hope this helps!


jhendrixMSFT:
Fix for https://github.com/Azure/azure-sdk-for-go/issues/765.

mcardosos:
@kvstark Cool! Thanks for contributing!

PR #782, opened 2017-09-21
Blobtest

mcardosos:
PTAL
@jhendrixMSFT @marstr 
Test related to https://github.com/Azure/azure-sdk-for-go/pull/781
Also setting blob type after creating, so we don't need to get blob properties.

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -627,3 +627,13 @@ func pathForResource(container, name string) string {
 	}
 	return fmt.Sprintf("/%s", container)
 }
+
+func (b *Blob) respondCreation(resp *storageResponse, bt BlobType) error {
+	readAndCloseBody(resp.body)
+	err := checkRespCode(resp.statusCode, []int{http.StatusCreated})
+	if err != nil {
+		return err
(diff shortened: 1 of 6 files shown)


storage/blob.go
```diff
@@ -627,3 +627,13 @@ func pathForResource(container, name string) string {
...
 	return fmt.Sprintf("/%s", container)
 }
+
+func (b *Blob) respondCreation(resp *storageResponse, bt BlobType) error {
+	readAndCloseBody(resp.body)
+	err := checkRespCode(resp.statusCode, []int{http.StatusCreated})
+	if err != nil {
+		return err
+	}
+	b.Properties.BlobType = bt
```
jhendrixMSFT: Should this be set before returning if err != nil?

mcardosos: I prefer not to modify the receiver if the operation was not successful

jhendrixMSFT: Thinking about this more I agree with you, if the creation fails then no blob exists so it's ok to not set the type.

PR #793, opened 2017-10-04
Updated `tools/generator` to read from literate files

marstr:
This is the generator for the next generation package in our SDK that provide multi-API Version and profile support.


marstr:
Alright @jhendrixMSFT / @mcardosos: This PR is ready for feedback and merge. It is no longer a WIP

tools/profileBuilder/latest.go
```diff
@@ -78,11 +78,10 @@ func (latest LatestStrategy) Enumerate(cancel <-chan struct{}) collection.Enumer
 				return
 			}
 
-			version := pathMatches[2]
+			version := pathMatches[3]
 			currentGroup := operationGroup{
 				provider:     pathMatches[1],
-				resourceType: pathMatches[2],
-				group:        pathMatches[3],
+				resourceType: pathMatches[4],
```
mcardosos: >resourceType: pathMatches[4], [](start = 4, length = 29)

does this also work with services that have a `parentResourcePath` ? What about graphrbac? Their paths are not like the rest of ARM

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
-	useDebug := flag.Bool("debug", false, "Include this flag to print debug messages as the program executes.")
-	rawVersion := flag.String("version", "0.0.0", "The version that should be stamped on the generated code. Common usage will be to report user agent.")
-
-	flag.Parse()
-
-	if help {
-		return
-	}
+// version should be set by the linker when compiling this program by providing the following arguments:
+// -X main.version=<version>
```
mcardosos: ><version> [](start = 19, length = 9)

Here do you mean the semantic version? or the git commit?

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
-		`(?P<package>[\w\-\d\.]+)`,
-		plane,
-		`[Mm]icrosoft\.(?P<category>[\w\-]+)`,
-		`(?P<apiVersion>v?\d{4}-\d{2}-\d{2}(?:[\-\.][\w\d\-\.]+)?)`,
-		`(?P<group>[\w\-\d\.]+)\.json`,
-	}
-	return regexp.MustCompile(strings.Join(elements, `[/\\]`))
+// getDefaultOutputBase returns the default location of the Azure-SDK-for-Go on your filesystem.
+func getDefaultOutputBase() string {
+	return strings.Replace(path.Join(goPath(), "src", "github.com", "Azure", "azure-sdk-for-go"), `\`, "/", -1)
```
mcardosos: `normalizePath` :D ?

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
 		}
 	}
-	return retval
-}
 
-func generate(swag SwaggerFile, outputRootPath, specsRootPath string, output io.Writer) (outputDir string, err error) {
-	if output == nil {
-		output = ioutil.Discard
-	}
+	targetFiles = collection.AsEnumerable(flag.Args())
```
mcardosos: >Args [](start = 44, length = 4)

Example?

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
+		// The following function compiles the regexp which finds Go related package tags in a literate file, and creates a collection.Unfolder.
+		// This function has been declared this way so that the relatively expensive act of compiling a regexp is only done once.
+		func() collection.Unfolder {
+			const patternText = "```" + `\s+yaml\s+\$\(\s*tag\s*\)\s*==\s*'([\d\w\-\.]+)'\s*&&\s*\$\(go\)[\w\d\-\s:]*\s+output-folder:\s+\$\(go-sdk-folder\)([\w\d\-_\\/\.]+)\s+[\w\d\-\s:]*` + "```"
+
+			goConfigPattern := regexp.MustCompile(patternText)
+
+			// This function is a collection.Unfolder which takes a literate file as a path, and retrieves all configuration which applies to a package tag and Go.
+			return func(subject interface{}) collection.Enumerator {
+				results := make(chan interface{})
```
mcardosos: can it be a `generationTuple` from the beginning?

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
+			// Intenionally Left Blank
+		}
 
-func cloneAPISpecs(origin, local string) (string, error) {
-	_, cloneLoc := filepath.Split(local)
-	if verbose {
-		fmt.Println("Cloning Azure REST API Specs to: ", cloneLoc)
+		fmt.Println("Execution Time: ", time.Now().Sub(start))
+		fmt.Println("Generated: ", generatedCount)
+		fmt.Println("Formatted: ", formattedCount)
```
mcardosos: `golint` and `go vet` are useful too :D

tools/generator/program.go
marstr: These are the arguments to the program. https://godoc.org/flag#Args

Is that what you mean?

tools/generator/program.go
marstr: I've added more clarifying lines later locally.

Though to answer the immediate question, the git commit :)

tools/generator/program.go
marstr: Derp, yes good catch! Fixed locally

tools/generator/program.go
marstr: I'm afraid not :/

marstr: That's not to say there isn't anything we could do to make it nicer! But the generationTuple type would need to be significantly modified.

tools/generator/program.go
marstr: Fair enough!

tools/generator/program.go
mcardosos: I mean an example of a program argument, what can I input as argument?

tools/generator/program.go
mcardosos: Remind me plz, where do you use the git commit?

tools/generator/program.go
```diff
@@ -26,399 +37,280 @@ import (
...
-	help                        bool
-	anyMissing                  bool
-	noClone                     bool
-	verbose                     bool
-	wait                        bool
-	targetFile                  string
-	debugLog                    *log.Logger
-	version                     *semver.Version
+	targetFiles    collection.Enumerable
+	packageVersion string
```
mcardosos: >packageVersion [](start = 1, length = 14)

Where is this set?

marstr: Fixed locally.

jhendrixMSFT:
To take a step back, how come we need our own generation program?  With all the language-specific generation options in the config files it seems that we should only need one tool to perform generation for any language.

mcardosos:
BTW, how are you whitelisting dataplane stuff?

marstr:
I already discussed all of the following with @mcardosos  and @jhendrixMSFT offline, but I wanted to copy this over for community involvement.

---

> @jhendrixMSFT says:
> To take a step back, how come we need our own generation program? With all the language-specific generation options in the config files it seems that we should only need one tool to perform generation for any language.

Agreed, ideally we wouldn't have to have such a tool. However, at the moment it's the best thing we've got. I'm in talks with @johanste, @lmazuel, and @sergey-shandar about how we could extend our current tool chain to have some reusable means of generating SDKs from README tags.

> @mcardosos says:
> BTW, how are you whitelisting dataplane stuff?

Specifically, having a `yaml $(tag)=='<tag name>' && $(go)` with an `output-folder` attribute white lists a package to be generated. Generally though, I've basically taken the position that being in `master` of the public repository that documentation is being generated off of constitutes as being ready to be generated in Go. We get a lot of Issues files on us asking why we don't generate such Swaggers, I'd like to be out of that business. I believe @salameer is working on improving the Specs Repo processes to accomodate this thinking.

tools/profileBuilder/latest.go
marstr: Yep, I've been adding in support for broader and broader sets. The regexp itself is structured to take care of those differences at the moment.

tools/generator/program.go
marstr: I just added a new commit which replaces the `flag.Usage` global variable function, the new tells the caller that the expected arguments are literate files.

marstr:
@mcardosos / @jhendrixMSFT any final thoughts before I merge this in?

mcardosos:
Not many, not many

Issue #794, opened 2017-10-05
Traffic Manager API Version in SDK isn't available


tombuildsstuff:
👋

I've spent the last couple of days trying to migrate us over to using v11 of the SDK - and have run into what I believe is a blocker with the Traffic Manager SDK. v11 switches over to using API version `2017-09-01-preview` (from `2015-11-01`) - which it appears isn't actually available in Azure yet:

```
trafficmanager.ProfilesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="NoRegisteredProviderFound" Message="No registered resource provider found for location 'global' and API version '2017-09-01-preview' for type 'trafficmanagerprofiles'. The supported api-versions are '2015-04-28-preview, 2015-11-01, 2017-03-01, 2017-05-01'. The supported locations are ', global'."
```

This means our acceptance tests are failing:

```
$ acctests azurerm TestAccAzureRMTrafficManagerProfile_priority
=== RUN   TestAccAzureRMTrafficManagerProfile_priority
--- FAIL: TestAccAzureRMTrafficManagerProfile_priority (53.12s)
	testing.go:459: Step 0 error: Error applying: 1 error(s) occurred:

		* azurerm_traffic_manager_profile.test: 1 error(s) occurred:

		* azurerm_traffic_manager_profile.test: trafficmanager.ProfilesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="NoRegisteredProviderFound" Message="No registered resource provider found for location 'global' and API version '2017-09-01-preview' for type 'trafficmanagerprofiles'. The supported api-versions are '2015-04-28-preview, 2015-11-01, 2017-03-01, 2017-05-01'. The supported locations are ', global'."
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-azurerm/azurerm	53.147s
```

However when switching out the API version used in V11 of the SDK to `2015-11-01` - the tests appear to pass fine:

```
 $ acctests azurerm TestAccAzureRMTrafficManagerProfile_priority
=== RUN   TestAccAzureRMTrafficManagerProfile_priority
--- PASS: TestAccAzureRMTrafficManagerProfile_priority (60.44s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azurerm/azurerm	60.460s
```

[...]

PR #800, opened 2017-10-06
Trafficmanager

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/794
This does not introduce breaking changes.
PTAL
@jhendrixMSFT @marstr


arm/trafficmanager/usermetricskeys.go
```diff
@@ -0,0 +1,222 @@
...
+	return
+}
+
+// CreateOrUpdatePreparer prepares the CreateOrUpdate request.
+func (client UserMetricsKeysClient) CreateOrUpdatePreparer() (*http.Request, error) {
+	pathParameters := map[string]interface{}{
+		"subscriptionId": autorest.Encode("path", client.SubscriptionID),
+	}
+
+	const APIVersion = "2017-09-01-preview"
```
marstr: This operation group looks like it is still has the old API version baked into it. Is that correct?

mcardosos: It is intentional
https://github.com/Azure/azure-rest-api-specs/pull/1668

PR #806, opened 2017-10-10
Auth

mcardosos:


marstr:
Chatted offline, we decided to close this in favor of just adding utility functions in go-autorest for working with auth files.

PR #814, opened 2017-10-15
Updating travis build check to be global.

marstr:


marstr:
I'm going to abandon this PR and instead submit another where we use a `makefile`, as can be seen here in @devigned's project: [https://github.com/devigned.veil/Makefile](https://github.com/devigned/veil/blob/4d88b5a40b6b6ea9a6ba46f6ed5a7ca89d9adfef/Makefile)

The advantage of doing it this way would be two fold:
 1. Allow us to write the rules in such a way that only relevant information is spit out in the CI log, also skipping irrelevant tests depending on the environment.
2. Mac/Linux users will be able to run the tests as if they were executing in Travis simply by running `make`. Windows users will not be impacted by this change either for the better or worse.

PR #815, opened 2017-10-16
Generating out 'latest' and 'preview' profiles

marstr:
Still to come is the snapshot profile which is supported in Stack


mcardosos:
`visualstudio` package is missing in the `preview` profile (I don't know if there are more missing packages). Is the profile like that?

marstr:
> @mcardosos says:
visualstudio package is missing in the preview profile (I don't know if there are more missing packages). Is the profile like that?

Hmm, when I look in my Dev branch I see Visual Studio in the preview profile: ://github.com/marstr/azure-sdk-for-go/tree/dev/profiles/preview/visualstudio/mgmt/visualstudio

Is that the package you mean? Was there another one that was missing?

mcardosos:
Sorry, I meant `latest`

marstr:
Ahhhh! yes, that is by design. [If you go look](https://github.com/Azure/azure-rest-api-specs/tree/4b65060d24c05de5c8776722618050fa85fc0363/specification/visualstudio/resource-manager/Microsoft.VisualStudio), `visualstudio` only has `preview` packages, which are filtered out of the the `latest` profile.

mcardosos:
Ok ok. Only comment is I would explain what this profiles are

marstr:
I'll address documentation concerns in future PRs. For now, I'm going to merge this in :)

PR #817, opened 2017-10-20
Adding previously missing services package.

marstr:
Adding previously untagged package for the Microsoft.Resource 2016-02-01.

This PR should not be merged before https://github.com/Azure/azure-rest-api-specs/pull/1901


marstr:
Updating the profile definition of 2017-03-09 as per Azure Stack Team's guidance.

PR #819, opened 2017-10-23
Adding a Template to help guide new PRs.

marstr:
There are a lot of submissions to the `master` branch, which will get
immediately released to folks using `go get` for that reason, we want to
always make sure that `master` stays in release quality.


jhendrixMSFT:
@marstr How about including that the PR links to the issue it fixes and that the change description contains an adequate summary of the fix?  So many times we get PRs with very little context.

marstr:
Change made, @jhendrixMSFT :)

Issue #822, opened 2017-10-24
PutBlockWithLength fails for many io.Reader implementations


majedev:
`blob.PutBlockWithLength()` returns AuthenticationFailed error when passed any implementations of `io.Reader` other than `bytes.Reader`, `bytes.Buffer`, `strings.Reader` and `io.LimitedReader`. This is caused by the `Content-Length` header not being set correctly for the outgoing request, causing a 403 response from the server.

This is a generalization for issue #683 (which was resolved in #704 but only for io.LimitedReader type). The following example demonstrates this behavior for `os.File` reader type (assuming test.txt file exists):

```
f, err := os.Open("test.txt")
if err != nil {
    panic(err)
}
defer f.Close()

stat, err := f.Stat()
if err != nil {
    panic(err)
}

size := stat.Size()

cli, err := storage.NewBasicClient("testAccount", "topSecretAwesomeSauce")
if err != nil {
    panic(err)
}

blobSvc := cli.GetBlobService()
container := blobSvc.GetContainerReference("azure-test")

blob := container.GetBlobReference("test.txt")
err = blob.PutBlockWithLength("0000", uint64(size), f, nil)
if err != nil {
    panic(err)
}
```

jhendrixMSFT:
Fix has been merged and will be included in the next release.

PR #823, opened 2017-10-24
Set req.ContentLength from headers regardless of body type

majedev:
This is a fix for #822. This makes requests for any generic io.Reader type possible without having to check the underlying Reader implementation. The `headers` map will already have the correct Content-Length value set, so I just parsed it and passed it to req.ContenLength value to match the access signature.

I've also removed util implementation for io.LimitedReader since it's no longer needed - although I still have other concerns for rewinding arbitrary reader types during retries.


storage/util_1.8.go
```diff
@@ -1,32 +0,0 @@
...
-import (
-	"io"
-	"io/ioutil"
-	"net/http"
-)
-
-func setContentLengthFromLimitedReader(req *http.Request, lr *io.LimitedReader) {
-	req.ContentLength = lr.N
-	snapshot := *lr
-	req.GetBody = func() (io.ReadCloser, error) {
```
jhendrixMSFT: Removing this means we can no longer rely on it for retrying requests and have to fall back to reading the entire contents of lr into a buffer which seems like a regression (the rest of the change looks good).

majedev: I don't think this implementation works with retries. There is no rewinding when GetBody is called, so the second retry will fail if the reader was consumed (partially or completely) during the 1st attempt. Here's a demonstration:

https://play.golang.org/p/6gllphBhVK

jhendrixMSFT: Yep you are right (thanks for the sample BTW it was very helpful).  The copies of LimitReader point to the same underlying reader which is really the thing that needs to be the snapshot.  So the current implementation is broken.  :(

majedev: My pleasure! I guess some more work need to be done on go-autorest project side for retriable requests, I've mentioned a couple of suggestions here: https://github.com/Azure/go-autorest/issues/157#issuecomment-339113595

 How do you propose to proceed?

jhendrixMSFT: We are getting ready to release an updated storage SDK which fixes this issue (we use io.ReadSeekers to solve the retry problem).

Issue #688, opened 2017-07-17
Support Auto-registration of RPs


salameer:

mcardosos:
Fixed in https://github.com/Azure/azure-sdk-for-go/pull/835

PR #835, opened 2017-10-26
Auto rp

mcardosos:
PTAL @jhendrixMSFT @marstr @vladbarosan 

 - [x] I've tested my changes, adding unit tests where applicable.
 - [x] If I'm targeting the `master` branch, I've also added a note to [CHANGELOG.md](https://github.com/Azure/azure-sdk-for-go/blob/master/README.md).
 - [x] I've mentioned any relevant open issues in this PR, making clear the context for the contribution.


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/Azure/azure-sdk-for-go
 import:
 - package: github.com/Azure/go-autorest
-  version: ~8.4.0
+  version: ^9.1.0
```
marstr: Let's keep this as `~9.1.0`

marstr: Oh also, in the changelog you mention that you need 9.1.1 not 9.1.0, which is right?

mcardosos: Fixed

mcardosos: well, it is at least 9.1.0, will fix

mcardosos: 9.1.1 is for generator tests

mcardosos: Fixed

jhendrixMSFT: I would be interested to hear from both sides on the motivations behind the change.

mcardosos: Okok, looked for the [difference](https://github.com/Masterminds/glide/blob/master/docs/versions.md)
Now that I understand it, `^` would be better I think

marstr: Agreed, caret seems to do what I thought tilde did. :)

marstr:
Once `glide.yaml` is how you like it, feel free to merge and tag as a release.

marstr:
This PR fixes #688

PR #844, opened 2017-11-02
Add MongoDb Client for CosmosDb

vladbarosan:
A convenience layer for getting a MongoDb client to be used with CosmosDb. Supports auth
through
 - username and password
 - connection string
 - MSI 
 - SP Token
---


vladbarosan:
Questions:
 - Is there `versions file necessary`  and if so, anything related to the one in the management plane ?
 - Do we need to unit-test this ? if so is there any example on how to use the `recorder` feature ?

dataplane/cosmos-db/client.go
```diff
@@ -0,0 +1,119 @@
...
+	result, err := cosmosDbClient.ListConnectionStrings(resourceGroup, account)
+
+	if err != nil {
+		return nil, err
+	}
+
+	connectionStrings := *result.ConnectionStrings
+
+	fmt.Println(fmt.Sprintf("this isthe conn string: %s", *(connectionStrings[0].ConnectionString)))
+
```
mcardosos: Used for debugging I guess?

dataplane/cosmos-db/client.go
```diff
@@ -0,0 +1,119 @@
...
+	return NewMongoDBClient(dialInfo)
+}
+
+// NewMongoDBClientWithSPToken returns a  session to communicate with CosmosDB using an auth token.
+func NewMongoDBClientWithSPToken(spToken *adal.ServicePrincipalToken, subscriptionID, resourceGroup, account string) (*mgo.Session, error) {
+
+	authorizer := autorest.NewBearerAuthorizer(spToken)
+
+	cosmosDbClient := cosmosdb.NewDatabaseAccountsClient(subscriptionID)
+	cosmosDbClient.Authorizer = authorizer
```
mcardosos: Wont work on sovereign clouds

dataplane/cosmos-db/client.go
```diff
@@ -0,0 +1,119 @@
...
+
+	cosmosdb "github.com/Azure/azure-sdk-for-go/arm/cosmos-db"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/adal"
+	"gopkg.in/mgo.v2"
+)
+
+const (
+	cosmosDbConnectionPort = 10255
+	managementResource     = "https://management.azure.com/"
```
mcardosos: >managementResource [](start = 1, length = 18)

This only works on the public cloud

dataplane/cosmos-db/client.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// NewMongoDBClientWithConnectionString returns a MongoDb session to communicate with CosmosDB using a connection string.
+func NewMongoDBClientWithConnectionString(connectionString string) (*mgo.Session, error) {
```
mcardosos: >NewMongoDBClientWithConnectionString [](start = 5, length = 36)

I would add a test for this func

mcardosos:
Versions files are used for useragents.
The recorder I used for storage tests was this one https://github.com/dnaeon/go-vcr
But you could also use [net/http/httptest](https://golang.org/pkg/net/http/httptest) from the standard library

dataplane/cosmos-db/client.go
```diff
@@ -0,0 +1,119 @@
+// Package cosmosdb provides clients for Microsoft Azure CosmosDb Services.
+package cosmosdb
```
jhendrixMSFT: CosmosDB encompasses a variety of storage technologies.  Since this addition is strictly for MongoDB on Cosmos I think we should give the package a more meaningful name.

devigned: I don't know. Each function let's you know you are getting a Mongo API client and it's within the cosmosdb package. Seems pretty solid. 

The other functions for constructing clients could be named similarly, but with their own protocol name.

dataplane/cosmos-db/client.go
devigned: Why would this not work on sovereign clouds?

dataplane/cosmos-db/client.go
jhendrixMSFT: To be clear, I think they should all be packages under a cosmosbd top-level directory.  Something like this (names can use improvement but you get the idea).
```
cosmosdb/mongo
cosmosdb/docdb
cosmosdb/table
cosmosdb/graph
```
For storage we didn't segregate by technologies (blobs/files/queues etc) and it's a big mess, in terms of discoverability and maintenance.

dataplane/cosmos-db/client.go
mcardosos: Because creating a client with that func defaults to the public cloud.

dataplane/cosmos-db/client.go
mcardosos: BTW, now that we have all APIs support, I think this should go in the services folder, not the dataplane folder.

vladbarosan:
Addresses #859

services/cosmos-db/mongodb/client.go
```diff
@@ -0,0 +1,116 @@
...
+	return NewMongoDBClient(dialInfo)
+}
+
+// NewMongoDBClientWithSPToken returns a  session to communicate with CosmosDB using an auth token.
+func NewMongoDBClientWithSPToken(spToken *adal.ServicePrincipalToken, subscriptionID, resourceGroup, account string) (*mgo.Session, error) {
+
+	authorizer := autorest.NewBearerAuthorizer(spToken)
+
+	cosmosDbClient := documentdb.NewDatabaseAccountsClient(subscriptionID)
+	cosmosDbClient.Authorizer = authorizer
```
mcardosos: This part still wont work with sovereign clouds

vladbarosan: Ah yes, thanks for catching it  :)

Issue #866, opened 2017-11-20
Move from Glide to `dep`


marstr:
The tide seems to be changing, and it seems like we'd be better off moving to use [`dep`](https://github.com/golang/dep) instead of  [`glide`](https://github.com/Masterminds/glide). We'd like to make that move for v12.

marstr:
Fixed in the `dev` branch

PR #876, opened 2017-11-22
Moving to use `dep` instead of `glide`

marstr:
This contribution satisfies #866.

---

 - [x] I've tested my changes, adding unit tests where applicable.
 - [x] If I'm targeting the `master` branch, I've also added a note to [CHANGELOG.md](https://github.com/Azure/azure-sdk-for-go/blob/master/README.md).
 - [x] I've mentioned any relevant open issues in this PR, making clear the context for the contribution.


mcardosos:
There is a reference to glide in the documentation I added recently, please take a look there too.

.travis.yml
```diff
@@ -2,17 +2,16 @@ sudo: false
 
 language: go
 go:
-  - 1.7.x
```
mcardosos: Why are we removing 1.7?

marstr: I mention that in the commit message :)

>Using go dep requires support for the `sort` packages which was first
available in in go1.8

mcardosos: I see. The idea is to release this with v12 right?

marstr: Correct :)

marstr:
> @mcardosos says:
There is a reference to glide in the documentation I added recently, please take a look there too.

Fixed!

Gopkg.toml
```diff
@@ -0,0 +1,70 @@
+
+# Gopkg.toml example
```
vladbarosan: Is this intentionally here ? ( the example part)

Gopkg.lock
```diff
@@ -0,0 +1,153 @@
...
+  version = "v1.0.0"
+
+[[projects]]
+  name = "github.com/nozzle/throttler"
+  packages = ["."]
+  revision = "d9b45f19996c645d38c9266d1f5cf1990e930119"
+  version = "v1.0"
+
+[[projects]]
+  name = "github.com/satori/go.uuid"
```
vladbarosan: from here https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md only one of the `version`, `revision` or `branch` can be specified

vladbarosan: Same for the other constraints

Gopkg.toml
marstr: Yep, it is auto-generated as part of the `dep` bootstrapping process. I think for now at least, because it is a new format for most folks that it makes sense to keep it here.

Gopkg.lock
vladbarosan: or that is for the toml file and this will have everything ?

Gopkg.toml
vladbarosan: Synced with Martin offline we can leave this part for a while as it is a new format

Gopkg.lock
marstr: You are correct, that is only for the `.toml`. For the `.lock` file, it is auto-generated by the tool and specifies everything it needs for context, including version info and commit ids.

PR #892, opened 2017-12-12
rewrite README

joshgav:
Please review but don't merge till the new samples repo is published. Thanks!

cc @sptramer


SERVICES.md
```diff
@@ -0,0 +1,457 @@
...
+  "batch/mgmt/2015-12-01/batch"<br />
+  "batch/mgmt/2017-01-01/batch"<br />
+  "batch/mgmt/2017-05-01/batch"<br />
+  "batch/mgmt/2017-09-01/batch"<br />
+  </td></tr>
+
+  <tr><td>
+  Billing
+    </td><td>
+  "billing/mgmt/2017-02-27-preview/billing"<br />
```
marstr: Would it be too much effort to make these links? :)

README.md
```diff
@@ -1,72 +1,204 @@
...
+  },
+  nil,
+)
+```
+
+For a complete VM example see
+[github.com/Azure-Samples/azure-sdk-for-go-samples/compute][2].
+
+For examples of using more services see
+[Azure-Samples/azure-sdk-for-go-samples][3].
```
marstr: IMHO, having both of these links feels redundant. I'd rephrase to something like:
``` Markdown
For complete examples, including details on VM management, see:
[Azure-Samples/azure-sdk-for-go-samples](https://github.com/Azure-Samples/azure-sdk-for-go-samples)
```

README.md
```diff
@@ -1,72 +1,204 @@
...
-- Azure SDK for Go Documentation is available at [GoDoc.org](http://godoc.org/github.com/Azure/azure-sdk-for-go/).
-- Azure REST APIs used by packages in this repository are documented at [Microsoft Docs, Azure REST](https://docs.microsoft.com/en-us/rest/api/).
-- Azure Services are discussed in detail at [Microsoft Docs, Azure Services](https://docs.microsoft.com/en-us/azure/#pivot=services).
-
-# Code samples
-
-- [Getting Started with Azure Blob Service in Go](https://github.com/Azure-Samples/storage-blob-go-getting-started)
+
+Both SDK and API versions should be considered.
+
```
marstr: Maybe a link to the wiki could be useful here? https://github.com/Azure/azure-sdk-for-go/wiki/Versioning

SERVICES.md
```diff
@@ -0,0 +1,457 @@
+# Azure services and package import paths
```
sptramer: Is this file necessary? The `godoc` documentation covers all services in a table and is substantially less fragile since it is automatically updated.

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+# Contributing
+
+While in preview we maintain only a `master` branch for releases and `dev` branch for development.
+
+After release we will maintain a branch for each major version to backport important fixes.
+
+Please submit pull requests to `dev` to update the latest version or to a version branch for backports.
```
sptramer: Are there any other specific developer requirements, such as using `godep` for handling dependencies to be able to run tests, requirements for writing tests and integrating them, automated checks that need to pass, etc?

README.md
```diff
@@ -1,72 +1,204 @@
-# Microsoft Azure SDK for Go
-[![GoDoc](https://godoc.org/github.com/Azure/azure-sdk-for-go?status.svg)](https://godoc.org/github.com/Azure/azure-sdk-for-go) 
+// azure-sdk-for-go provides Go packages for using Azure services
```
sptramer: Delete? I don't think markdown supports non-HTML comments.

README.md
```diff
@@ -1,72 +1,204 @@
...
+:exclamation: **NOTE:** This project is in preview and breaking changes are 
+introduced frequently. Therefore, vendoring dependencies is even more
+important than usual. We use [dep](https://github.com/golang/dep).
 
-> **NOTE:** This repository is under heavy ongoing development and should be considered a preview. Vendoring your
-dependencies is always a good idea, but it is doubly important if you're consuming this library.
+To be notified about updates and changes, subscribe to the [Azure update feed][].
 
-# Installation
-- If you don't already have it, install [the Go Programming Language](https://golang.org/dl/).
```
sptramer: I would recommend including this as a prerequisite, still. Mention the minimum supported version of go.

CONTRIBUTING.md
sptramer: Also make sure to mention the minimum required version of Go for development. This might be different than the minimum version to use the libraries, if I remember right.

README.md
```diff
@@ -1,72 +1,204 @@
...
+
+1. Import a package from the [services][1] directory. See [SERVICES.md][] for a list.
+1. Create and authenticate a client with a `New*Client` func, e.g.
+   `c := compute.NewVirtualMachinesClient(...)`.
+1. Invoke API methods using the client, e.g. `c.CreateOrUpdate(...)`.
+1. Handle responses.
+
+[1]: https://github.com/Azure/azure-sdk-for-go/tree/master/services
+
+For example, to create a new VM (substitute your own values for angle brackets):
```
sptramer: I decided to check and no other SDK has an inline example like this. They instead link out to the samples directory, which is what we should do here as well to:

1. Direct users to where they can actually find samples
2. Avoid drift between the SDK development and the sample code

README.md
```diff
@@ -1,72 +1,204 @@
...
+[3]: https://github.com/Azure-Samples/azure-sdk-for-go-samples
+
+## Authentication
+
+Most operations require an OAuth token for authentication and authorization.
+Following is how to get one using the
+[adal](https://github.com/Azure/go-autorest/tree/master/autorest/adal)
+package.
+
+If you need a service principal (aka Client, App), run `az ad sp create-for-rbac -n "<app_name>"`.
```
sptramer: Does not say where or how you run this. Should link to the portal, mention Cloud Shell, and also link to the CLI overview page. Can also link to the CLI's "create a service principal" article.

README.md
joshgav: i'm being cute and including a go-style comment at the top :)

sptramer: Fine with me, maybe move it after the H1?

SERVICES.md
joshgav: a reasonable request but I'd like to defer since it's error-prone. for now the main purpose of this page is awareness of what's required to select a version so we can get feedback. if this sticks do you think we could automate creation of columns for this table via [profileBuilder](https://github.com/Azure/azure-sdk-for-go/tree/master/tools/profileBuilder)?

CONTRIBUTING.md
joshgav: thanks for raising these @sptramer! @marstr we should discuss. i'll work on this, but would it be okay to land the basic page as proposed now and add these later (e.g. by 14.0.0)? in particular i'm not sure we have a well-documented test process yet.

README.md
joshgav: i think it should be shorter but i think a small example helps readers better understand the "Getting Started" steps. I replaced VM with Vnet, what do you think?
we'll have to be sure this small snippet keeps pace with code changes, good point.

SERVICES.md
joshgav: for now I'd like to include it as we need to hold users' hands through the transition from `arm/` to `services/` and there are a lot of services+versions permutations. i think it's helpful to be upfront about that. let's consider moving or relying on something else over time though, thanks!

README.md
joshgav: done!

README.md
joshgav: done! i bumped the link to the top of the "use" section too.

README.md
```diff
@@ -1,72 +1,183 @@
...
+For complete examples see [Azure-Samples/azure-sdk-for-go-samples][samples_repo].
+
+1. Import a package from the [services][services_dir] directory. See [SERVICES.md][] for
+   a list.
+1. Create and authenticate a client with a `New*Client` func, e.g.
+   `c := compute.NewVirtualMachinesClient(...)`.
+1. Invoke API methods using the client, e.g. `c.CreateOrUpdate(...)`.
+1. Handle responses.
+
+For example, to create a new virtual network (substitute your own values for
```
sptramer: I think that this can be dropped in favor of just keeping in the auth example, but there is the very real concern that using the thin wrappings around the REST data structures can get confusing. Best way to fix that is documentation improvements, though.

README.md
```diff
@@ -1,72 +1,183 @@
...
+## Authentication
+
+Most operations require an OAuth token for authentication and authorization.
+Following is how to get one using the
+[adal](https://github.com/Azure/go-autorest/tree/master/autorest/adal)
+package.
+
+If you need a service principal (aka Client, App), run `az ad sp create-for-rbac -n "<app_name>"` 
+in [azure-cli](https://github.com/Azure/azure-cli), see
+[docs](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli?view=azure-cli-latest
```
sptramer: Incorrectly formatted MD url, also the text is difficult to parse. Would suggest:

````
Most operations require authentication. The following snippet shows how to get one using the [adal](...) package,
as long as you have an available service principal. If you need to create a service principal, you can do so with the
[Azure CLI 2.0](https://docs.microsoft.com/en-us/cli/azure/overview) with the command:

```
... command here ...
```

After creating the service principal, write down the ID, secret, and tenant ID to use in your Go code.
````

Even this may not provide enough context to be useful. We should include a link to the full service principal docs at some point because they cover things like the differentiation between roles and when to use what.

File: README.md (hunk 10 above)
Comment (joshgav) [reply README.md]: I think the README is an opportunity to familiarize people with our paradigms early so I'd like to keep a straightforward, short example of client usage here. We could consider combining this with the auth example as long as they can still be read clearly.

File: README.md [hunk 12]
```diff
@@ -1,72 +1,183 @@
...
+# Azure SDK for Go
+
+[![godoc](https://godoc.org/github.com/Azure/azure-sdk-for-go?status.svg)](https://godoc.org/github.com/Azure/azure-sdk-for-go) 
 [![Build Status](https://travis-ci.org/Azure/azure-sdk-for-go.svg?branch=master)](https://travis-ci.org/Azure/azure-sdk-for-go) 
 [![Go Report Card](https://goreportcard.com/badge/github.com/Azure/azure-sdk-for-go)](https://goreportcard.com/report/github.com/Azure/azure-sdk-for-go)
 
-This is Microsoft Azure's core repository for hosting Go packages which offer a more convenient way of targeting Azure
-REST endpoints. Here, you'll find a mix of code generated by [Autorest](https://github.com/Azure/autorest) and hand
-maintained packages.
+// azure-sdk-for-go provides Go packages for using Azure services
```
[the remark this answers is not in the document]
joshgav: Yeah, I agree. Removing the `//`.

README.md
joshgav: Cleaned up language, PTAL. Thanks!

joshgav:
@Azure/go-admins please take a look, thanks!

README.md
```diff
@@ -1,72 +1,185 @@
...
+specify an API version in your import path, for example:
+
+```go
+import "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2016-03-01/compute"
+import "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+```
+
+See [SERVICES.md][] for a full list of services and paths.
+
+Azure **API profiles** specify a set of API versions tested for
```
vladbarosan: Maybe specify that you can use `latest` for the latest stable api versions and `preview` for the latest preview api versions

SERVICES.md
```diff
@@ -0,0 +1,457 @@
...
+
+All package paths begin with **github.com/Azure/azure-sdk-for-go/services/**.
```
vladbarosan: Does this list need to be updated ?

README.md
```diff
@@ -1,72 +1,185 @@
...
+      },
+    },
+  },
+  nil)
+```
+
+## Authentication
+
+Most operations require an OAuth token for authentication and authorization.
+You can get one from Azure AD using the
```
vladbarosan: Coming this sprint hopefully a less verbose way of auth

vladbarosan: @sptramer it will still rely on the present authentication methods present in `go-autorest`, they will be a convenience layer to simplify the experience.
A PR with some of the work is here https://github.com/Azure/go-autorest/pull/221 will update it this week after we get an agreement on the best way to expose things, but mostly using what is defined there we would either pass a config to the constructor functions or have a method on them to pass this.

CONTRIBUTING.md
joshgav: added this. for now we support 1.8 and 1.9, we're working on 1.10 (#964) and considering 1.7 (#965).

README.md
joshgav: will leave as is for now but please consider submitting a PR with alternative suggestions!

joshgav:
landed in https://github.com/Azure/azure-sdk-for-go/commit/4494f695920ad731d9b976c8fe5037695798342f

working on landing in master now

PR #907, opened 2017-12-14
remove Gododir/gen.go

joshgav:
looks like this isn't used anymore, okay to remove?


joshgav:
I also removed the entry from Gopkg.toml and ran `dep ensure` (without -update) to update Gopkg.lock as minimally as possible and included both in another commit. can you double-check that?

PR #946, opened 2018-01-05
Handles error from uuid.NewV1

jerpa:
Can't do go get without this fix since uuid now returns error for NewV1.


jhendrixMSFT:
Thanks for submitting a PR.  This has been fixed in https://github.com/Azure/azure-sdk-for-go/releases/tag/v12.1.1-beta.

PR #948, opened 2018-01-05
Moving UUID generation to github.com/marstr/guid

### LINKED ISSUE #947: address breaking change in satori/go.uuid
Several issues have been opened: https://github.com/Azure/azure-sdk-for-go/issues/945, https://github.com/Azure/azure-sdk-for-go/issues/940, https://github.com/Azure/azure-sdk-for-go/pull/946.

Also: https://github.com/satori/go.uuid/issues/66

Let's get a fix in perhaps starting with #946 and release a new minor. Thanks!

@carolynvs do you have suggestions on how we can best shield users of our library from changes in [transitive dependencies](https://github.com/golang/dep/blob/master/docs/FAQ.md#what-is-a-direct-or-transitive-dependency)? Is dep's recommended practice for users to explicitly declare the dependency and version in their own Gopkg.toml file? Thank you!

marstr:
Fixes #947


storage/table_batch.go
```diff
@@ -131,14 +131,32 @@ func (t *TableBatch) MergeEntity(entity *Entity) {
 // the changesets.
 // As per document https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/performing-entity-group-transactions
 func (t *TableBatch) ExecuteBatch() error {
-	changesetBoundary := fmt.Sprintf("changeset_%s", uuid.NewV1())
+
+	// Using `github.com/marstr/guid` is in response to issue #947 (https://github.com/Azure/azure-sdk-for-go/issues/947).
```
jhendrixMSFT: I think this comment block should go in the change description and not here.

vladbarosan: @marstr  I would suggest to create an issue to track that fact the we need to change this once satori releases v2

marstr: Issue added.

marstr: I moved the bulk of the comment to address your note, @jhendrixMSFT  :)

Gopkg.toml
```diff
@@ -37,6 +37,10 @@
   branch = "master"
   name = "github.com/marstr/goalias"
 
+[[constraint]]
+  revision = "8bdf7d1a087ccc975cf37dd6507da50698fd19ca"
```
jhendrixMSFT: Is there a reason the revision is to a commit and not a semver?

marstr: Yep, so that it's even more locked down than a semver. This way, even if I make the same mistake Satori did, folks using `dep` won't get broken.

marstr: If you'd prefer, I can move it to a semver

jhendrixMSFT: Well I do think that semvers are easier to reason about however if we think this is a short-term fix then I think it's fine to leave it as-is.

jhendrixMSFT:
If this is going into master don't we need to rev the version number for all the packages?

Gopkg.toml
marstr: I'll add a note to evaluate this in #949

marstr:
Shoot, I don't have time for that, though I do think it's a good idea. Mind taking over that and pushing to my branch? Folks will have to merge this PR when I'm OOF.

vladbarosan:
I can take over later. Do we have permissions ? otherwise i would create a new PR . But should be just a ctrl-shift-f and replace in VS code to make the change :)

Issue #889, opened 2017-12-11
[storage] Client does not authenticate properly with an SAS generated from a container's access policy


StrikerX3:
I've been trying to use this library to write blobs and messages to a queue on an Azure account using a Shared Access Signature token generated with one of my container's access policies. The issue I'm facing is that it does not seem to support Shared Access Signatures based on a container's access policy at all.

If I try to use a client created with `storage.NewClientFromConnectionString`, following [these instructions](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#authenticating-from-a-client-application-with-a-sas) to build the connection string, any attempt to communicate with Azure services fail because they expect an account name which is actually embedded in the service endpoints. It seems that this method always uses the regular, non-SAS authentication scheme.

Using `storage.NewAccountSASClient` with the SAS token converted into `url.Values` and the appropriate storage account name also doesn't work. This time, every operation on the container returns 404 and ResourceNotFound. Doing `httputil.DumpRequest` on the request generated by the library I can see that the SAS token is never appended to the resource URL and there is no `Authorization` header.

I am currently using https://github.com/Azure/azure-storage-blob-go and it works fine for blobs, but doesn't support queues.

joshgav:
cc @seguler

jhendrixMSFT:
Fixed in https://github.com/Azure/azure-sdk-for-go/releases/tag/v12.2.0-beta.  Thanks for reporting this issue.

PR #954, opened 2018-01-09
Fix SAS support in storage package.

jhendrixMSFT:
Add support for creating a SAS client from an endpoint and SAS token.
Fixed bug that wasn't appending SAS token to URI query parameters in all
cases.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -75,6 +75,12 @@ const (
 	connectionStringAccountKey       = "accountkey"
 	connectionStringEndpointSuffix   = "endpointsuffix"
 	connectionStringEndpointProtocol = "defaultendpointsprotocol"
+
+	connectionStringBlobEndpoint  = "blobendpoint"
+	connectionStringFileEndpoint  = "fileendpoint"
+	connectionStringQueueEndpoint = "queueendpoint"
+	connectionStringTableEndpoint = "tableendpoint"
+	connectionStringSAS           = "sharedaccesssignature"
 )
 
 var (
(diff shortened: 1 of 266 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/889.

vladbarosan:
LGTM , waiting on @mcardosos to approve as she has better context.  To test it I guess I just need to call `NewClientFromConnectionString` with the SAS url ?

jhendrixMSFT:
@vladbarosan You are right, and it also points out I forgot to add a test (will fix).

storage/client.go
```diff
@@ -330,6 +346,47 @@ func NewAccountSASClient(account string, token url.Values, env azure.Environment
 	return c
 }
 
+// NewAccountSASClientFromEndpointToken constructs a client that uses accountSAS authorization
+// for its operations using the specified endpoint and SAS token.
+func NewAccountSASClientFromEndpointToken(endpoint string, sasToken string) (Client, error) {
```
mcardosos: >NewAccountSASClientFromEndpointToken [](start = 5, length = 36)

is this doing the same as `NewAccountSASClient`, but from other parameters, right?

jhendrixMSFT: That's correct.  Take a look [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#authenticating-from-a-client-application-with-a-sas) for more details, but the TL;DR version is you can specify and endpoint and SAS token in a connection string; this handles that case when called from ```NewClientFromConnectionString``` and exporting it seemed like a nice thing to do.

PR #908, opened 2017-12-15
Fix bug: Binary data is not properly encoded.

shiweiz:
Binary data should be encoded using base64.

PR #957, opened 2018-01-11
Fix bug: Binary data is not properly encoded.

shiweiz:
Binary data should be encoded using base64.

Previous conversions see PR #908.

### DIFF
--- a/storage/entity.go
+++ b/storage/entity.go
@@ -16,6 +16,7 @@ package storage
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


shiweiz:
@jhendrixMSFT

storage/entity.go
```diff
@@ -246,7 +247,7 @@ func (e *Entity) MarshalJSON() ([]byte, error) {
 		switch t := v.(type) {
 		case []byte:
 			completeMap[typeKey] = OdataBinary
-			completeMap[k] = string(t)
+			completeMap[k] = t
```
jhendrixMSFT: This doesn't need to be base-64 encoded?

jhendrixMSFT: I suppose it already is.

shiweiz: t is of type []byte, and golang automatically converts []byte to base64 string when calling json.Marshal().

PR #961, opened 2018-01-14
Fix profiler

### LINKED ISSUE #843: ProfileBuilder Items Order is Non-Deterministic
When Profiles are generated, the order that that the types/consts are written to the "models.go" file is inconsistent from run to run. This is causing noise in PRs, but no behavior changes.

The root cause is the invalid Token Position in the alias AST. The [`go/printer`](https://godoc.org/go/printer) sees that the positions are either not set, or are all conflicting, and assigns them as it sees fit. Apparently, it doesn't always make the same decision, hence the noise.

edit: formatting

### LINKED ISSUE #937: Adapt `latest` and `preview` profiles according to the new specs repo organization
Given the new way of determining  if a swagger is stable or not, we should adapt the profile to reflect this

vladbarosan:
Fixes #843 
- The reason for the non-deterministic behavior was the fact that the AST library is calling range on maps
which is implemented to return the elements in random order which causes the behavior to be non deterministic. For solutions we could have used a different library ( https://godoc.org/golang.org/x/tools/go/ast/astutil ). Didn't want to take a dependency on the implementation since they make no guarantee of being deterministic, so second solution is to sort the declaration nodes by position before actually writing it.

Fixes #937 
-  For the new repo we need to determine that an API is stable based on the new specs repo organization. To do this:
   - in the generator for each package tag we see the referenced input files.
  - If there is no input file referenced as coming from a preview folder we mark the package tag as stable
 - When processing the packages marked for go also, we check if the package tag is "stable". If so we write it in the latest\stableApis.txt file. 
 - TO generate the stable API's we use the list strategy for the profile builder with the generated list.


tools/profileBuilder/program_test.go
```diff
@@ -92,3 +98,31 @@ func Test_getAliasPath(t *testing.T) {
 		})
 	}
 }
+
+func Test_generateAliasPackages(t *testing.T) {
+	files := token.NewFileSet()
+	gopath := strings.Replace(os.Getenv("GOPATH"), "\\", "/", -1)
+	folder := path.Join(gopath, "src", "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2015-06-15/network")
+	iterations := 10
```
marstr: Total nit, this could be declared as a constant.

``` Go
const iterations int = 10
```

tools/profileBuilder/program_test.go
```diff
@@ -92,3 +98,31 @@ func Test_getAliasPath(t *testing.T) {
...
+	if err != nil {
+		errLog.Printf("Couldn't open %q because: %v", folder, err)
+		return
+	}
+
+	for _, entry := range parsed {
+		packagesEnumerator.Add(entry)
+	}
+
+	want := generateAliasPackages(packagesEnumerator.Enumerate(nil)).ElementAt(0).(*alias).AliasPackage
```
marstr: I'd rather have this be more of an end-to-end test, and instead of comparing at the model level, compare the text output that is spit into a file or text buffer.

vladbarosan: I would say we want to ensure that the model we build is always the same which we do, but unit testing that printing is always the same seems unnecessary.

marstr: I'm okay agreeing to disagree.

PR #971, opened 2018-01-18
Auto import used packages in profiles

### LINKED ISSUE #920: imports missing in generated profiles
Some code that gets generated in the profiles requires certain imports. Currently there is no logic in the profile builder for that.

Ex:  `profiles\preview\resources\mgmt\management\models.go`

vladbarosan:
Fixes #920 
 - Using the same package goimports uses we can resolve missing packages.


vladbarosan:
I think I need to update toml here also for the dependency.

PR #977, opened 2018-01-20
Run build against latest go version

vladbarosan:
- Adding `master` to the build matrix for go.
 - Removing `GO15VENDOREXPERIMENT` as it is on by defaullt since go 1.6.


vladbarosan:
Should we make `master` an allowed failure ?

vladbarosan:
Fixes #977

Issue #958, opened 2018-01-12
Use new mongo for go driver


vladbarosan:
As mentioned by @devigned  , the mongo driver we are using is currently unmantained : https://github.com/go-mgo/mgo/blob/v2-unstable/README.md

vladbarosan:
Marked is as for GA for visibility

mcardosos:
Fixed on dev branch

Issue #949, opened 2018-01-05
Clean up GUIDs


marstr:
As covered in #947, the Satori UUID library broke underneath us.

Depending on how that is resolved (e.g. he adds a v2.0 label,) we can remove the code commited in #948.

Be careful though, as removing this code too early, or in response to an inadequate resolution by Satori could cause trouble for people installing directly, or targeting a version of `github.com/satori/go.uuid` from before the breaking changes.

jhendrixMSFT:
Long-term we need to remove our dependency on this package.

PR #979, opened 2018-01-23
Replacing mongoDB driver

mcardosos:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/958

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -54,12 +54,6 @@
   revision = "3026ca7e18e829805b990b41e4ce28c92cb80caa"
 
 [[projects]]
-  name = "github.com/marstr/guid"
-  packages = ["."]
-  revision = "8bd9a64bf37eb297b492a4101fb28e80ac0b290f"
-  version = "v1.1.0"
-
-[[projects]]
   branch = "master"
   name = "github.com/marstr/randname"
   packages = ["."]
@@ -84,7 +78,7 @@
     "pkcs12",
     "pkcs12/internal/rc2"
   ]
-  revision = "d94f6bc902c2970500950e88ab1a54a0c26bec48"
+  revision = "3d37316aaa6bd9929127ac9a527abf408178ea7b"
 
 [[projects]]
   branch = "master"
@@ -93,7 +87,7 @@
     "go/ast/astutil",
     "imports"
   ]
-  revision = "99037e3760ed7d9c772c980caee42b17779b80ce"
+  revision = "f86b507a7e91633e9ed99e4071f91a62e78f05f6"
 
 [[projects]]
   branch = "v1"
@@ -110,6 +104,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "56c5ece551cc0fe53dfe99350acb7a705b3538e056ccb711bd8cf2950a1a4d8f"
+  inputs-digest = "a17d78d70ec53003960990755d97545a369f737a59d57f8948dd2f682753f745"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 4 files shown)


Gopkg.lock
```diff
@@ -44,7 +56,8 @@
 [[projects]]
   name = "github.com/marstr/guid"
   packages = ["."]
-  revision = "8bdf7d1a087ccc975cf37dd6507da50698fd19ca"
+  revision = "8bd9a64bf37eb297b492a4101fb28e80ac0b290f"
```
vladbarosan: I think there is an item #949. is anyone doing this ? Otherwise can you pick it up ? If you  don't have time just assign it to me.

mcardosos: Interesting. So when we removed the references to Satori's package, did we replace them with Martin's package?

vladbarosan: @mcardosos there was only 1 place where we were actually generating a UUID so to make that the behavior is consistent for people who use go get or not, we switched to that temporarily yes

mcardosos: OKok, pushed changes about the guid issue here.
Release 1.2.0 from satori's package removed breaking changes.
This is being tracked here https://github.com/satori/go.uuid/issues/66

storage/table_batch.go
```diff
@@ -131,26 +131,14 @@ func (t *TableBatch) MergeEntity(entity *Entity) {
...
 func (t *TableBatch) ExecuteBatch() error {
-
-	// Using `github.com/marstr/guid` is in response to issue #947 (https://github.com/Azure/azure-sdk-for-go/issues/947).
-	id, err := guid.NewGUIDs(guid.CreationStrategyVersion1)
-	if err != nil {
-		return err
-	}
-
-	changesetBoundary := fmt.Sprintf("changeset_%s", id.String())
+	changesetBoundary := fmt.Sprintf("changeset_%s", uuid.NewV1())
```
marstr: We cannot take this change, because the fundamental error in @satori's code introduced here: https://github.com/satori/go.uuid/commit/0ef6afb2f6cdd6cdaeee3885a95099c63f18fc8c is still in his `master` branch and will still break folks using our repository in conjunction with `go get`.

marstr: https://github.com/Azure/azure-sdk-for-go/issues/947 is related to this problem.

vladbarosan: I think he released a new version. we can take a dependency on that and use the new way  of doing things.

mcardosos: But he has tagged a 1.2.0 release without breaking changes.
If we do `go get sdk` and `go get guid`, of course it would be broken, but truth is commits with breaking changes can happen anytime in any of our dependencies. `go get` will never protect us from breaking changes.
That is why the dep constraint is not against `master`, but `1.2.0`.

vladbarosan: @mcardosos so you are saying that release doesn't have the breaking changes or just that he upgraded only to minor version with the breaking change ?

mcardosos: The upgrade to minor version does not have breaking changes. Master has breaking changes.

mcardosos: I have talked offline with @marstr , the work done on this PR for uuid should be postponed.

vladbarosan: ok so you are saying that there is no release yet with the breaking change ? yes in that case we should postpone, but if he has let me know what would be the problem.

mcardosos: There is no release yet with the breaking change.

Gopkg.lock
```diff
@@ -29,6 +29,18 @@
   ]
   revision = "87d4990451a858cc210399285be976e63bc3c364"
 
+[[projects]]
+  branch = "master"
+  name = "github.com/globalsign/mgo"
```
marstr: Just to be clear, this is still a breaking change and will need to wait to be merged into `master` until that time.

mcardosos: This branch targets dev :D

marstr: I know :) I just want to make sure that anybody looking through the history sees that we were aware of the breaking changes.

PR #993, opened 2018-01-25
Fix profile builder

### LINKED ISSUE #963: generation cannot handle multiple folders after the version [impacts LUIS in Cognitive Services]

vladbarosan:
Fixes #963 
- Fixed regex to include the whole folder path for the grouping
- Generated the LUIS packages after the fixes in the generator 
- Generated profiles


profiles/latest/cognitiveservices/face/models.go
```diff
@@ -177,13 +177,13 @@ type IdentifyRequest = original.IdentifyRequest
 type IdentifyResult = original.IdentifyResult
 type ImageURL = original.ImageURL
 type Landmarks = original.Landmarks
+type List = original.List
 type ListDetectedFace = original.ListDetectedFace
 type ListIdentifyResult = original.ListIdentifyResult
-type ListListType = original.ListListType
+type ListList = original.ListList
```
mcardosos: It is more a question on the code generation, why do we get a ListList model? Is it like that in the swagger?

vladbarosan: pretty much. This is what is in swagger: https://github.com/Azure/azure-rest-api-specs/blob/0bef2b99ad6d52781e3c308bc6b420f6eef59ef7/specification/cognitiveservices/data-plane/Face/stable/v1.0/Face.json#L2378

PR #1019, opened 2018-01-30
[DO NOT MERGE] Singular version prototype.

jhendrixMSFT:
This is a prototype for what a singular version location might look like.


jhendrixMSFT:
I'm thinking that if we went this route we'd update the generator to make this behavior opt-in, so that by default you get the old behavior, allowing generated packages to stand on their own.

mcardosos:
:O cool :D It is a good thing the SDKd are generated from the md files, which have the `sdk-path` parameter

jhendrixMSFT:
@tombuildsstuff I was wondering how this change would impact Terraform.  IIRC Terraform has a checked in vendored subset of the Azure SDK.  It would mean you'd need to include this new version package and update it when you update the other packages.  The only place this breaks down is if you mix/match packages from different versions of the Azure SDK.

marstr:
In addition to what @jhendrixMSFT mentioned, we're hoping that having a single version becomes less of a problem for Terraform now that we have multi-api version support.

tombuildsstuff:
@jhendrixMSFT @marstr sorry for the delayed response here (I was moving house)

>  I was wondering how this change would impact Terraform. IIRC Terraform has a checked in vendored subset of the Azure SDK. It would mean you'd need to include this new version package and update it when you update the other packages. The only place this breaks down is if you mix/match packages from different versions of the Azure SDK.

Given we upgrade the entire Azure SDK at the same time, I don't believe this should cause us any issues (we can ensure we bump the `/version` folder at the same time) - so this gets a 👍 from our end :)

jhendrixMSFT:
@brendandburns Any feedback on this from the kubernetes side?

jhendrixMSFT:
We've decided to go with this approach.  The change has been merged in dev in preparation for the v14 GA release.

PR #1030, opened 2018-02-01
Updating version number in Swagger -> SDK config.

marstr:
UserAgent was also updated.

---

### DIFF
--- a/swagger_to_sdk_config.json
+++ b/swagger_to_sdk_config.json
@@ -9,8 +9,8 @@
       "verbose": "",
       "sdkrel:go-sdk-folder": ".",
       "multiapi": "",
-      "package-version": "v12.2.1-beta",
-      "user-agent": "Azure-SDK-For-Go/v12.2.1-beta services"
+      "package-version": "v12.3.1-beta",
+      "user-agent": "Azure-SDK-For-Go/v12.3.1-beta services"
     },
     "version": "0.2.0"
   }


swagger_to_sdk_config.json
```diff
@@ -9,8 +9,8 @@
       "verbose": "",
       "sdkrel:go-sdk-folder": ".",
       "multiapi": "",
-      "package-version": "v12.2.1-beta",
-      "user-agent": "Azure-SDK-For-Go/v12.2.1-beta services"
+      "package-version": "v12.3.1-beta",
```
jhendrixMSFT: Why is this v12.3.1-beta?  The latest release is v12.3.0-beta.

marstr: Righto, I deliberately have it incremented by one so that it's potentially ready to go if we need to release.

jhendrixMSFT: I see.  What is this file used for?

marstr: This is a configuration file for the Swagger -> SDK bot.

jhendrixMSFT: This is the per commit thing or something else?  When does a person have to update this version info?

vladbarosan: @marstr per @jhendrixMSFT  comments I think this raises the issues to properly document in a readme or in OneNote how to use the configs and the generator.

marstr: The idea was that temporarily, this file would stay one ahead of the released version in anticipation of the changes being released as patches. Ideally the version number would be automatically set, but we had trouble doing it that way.

I was going to continue updating it, but it looks like I may have a little spare time this morning to skip this step and instead submit a PR with all of the configuration changes we were looking for.

jhendrixMSFT: I thought we were going to use a different version format, something like 20180201.00-nightly or something.  Did that change?

marstr: I'm going to cross-post some of this conversation to the new PR that I opened up.

PR #1071, opened 2018-02-09
Normalizing `management` package path name.

marstr:
All of the code that lived in the `management` package before this
commit was used to interact with Azure via an out-date model called
Azure Service Management (ASM) for short. Since we've moved to the Azure
Resource Management (ARM) model, the ASM packages living in the root of the
repository are becoming more and more misleading. While ASM is no longer
a recommended path to interact with Azure, we don't want to literally
delete these packages. For those reasons, we're moving them into the
services directory, where they will remain available for the forseeable
future.


jhendrixMSFT:
Now the directory name doesn't match the package name, can we keep the director name as "management"?

services/management/README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Service Management packages for Go
 
-The `github.com/Azure/azure-sdk-for-go/services/servicemanagement` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](../arm) folder.
+The `github.com/Azure/azure-sdk-for-go/services/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](../arm) folder.
```
tombuildsstuff: 🤔 with a generic name like `management` I could see some folks (read: me 😛) getting confused with something else - as such, would it make sense to name this package `classic` (or `classic/management`)?

marstr: Makes sense to me!

PR #1073, opened 2018-02-09
+Viper instances for each profileBuilder command.

### LINKED ISSUE #1060: profileBuilder can't bind viper to string cobra arguments
Quite embarrassingly, and for reasons that are completely unclear to me, I cannot get `github.com/spf13/viper` to bind to `github.com/spf13/cobra` `string` flags. `bool` flags seem to be working fine... for now I've worked around it, so this is a low-pri issue.

marstr:
Fixes #1060

### DIFF
--- a/tools/profileBuilder/cmd/latest.go
+++ b/tools/profileBuilder/cmd/latest.go
@@ -42,6 +42,8 @@ const (
 
 var rootDefault = model.DefaultInputRoot()
 
+var latestFlags = viper.New()
+
 // latestCmd represents the latest command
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
I assume everything works the same, this just cleans up some cruft yes?

marstr:
> @jhendrixMSFT says:
I assume everything works the same, this just cleans up some cruft yes?

Yep! There were actually some bugs I ran into the with the cruft that this fixes too.

PR #1074, opened 2018-02-12
update README for v14.0.0

joshgav:
notable updates:
* removed preview notices but still hinted at possible breaking changes and recommended dep
* replace references to "ADAL" with "auth"
* cleaned up versioning section to reflect latest consensus
* added blobs and app insights repo to Resources

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,33 +4,53 @@
 [![Build Status](https://travis-ci.org/Azure/azure-sdk-for-go.svg?branch=master)](https://travis-ci.org/Azure/azure-sdk-for-go) 
 [![Go Report Card](https://goreportcard.com/badge/github.com/Azure/azure-sdk-for-go)](https://goreportcard.com/report/github.com/Azure/azure-sdk-for-go)
 
-azure-sdk-for-go provides Go packages for using Azure services. It has been
-tested with Go 1.8 and 1.9. To be notified about updates and changes, subscribe
-to the [Azure update feed][].
+azure-sdk-for-go provides Go packages for managing and using Azure services. It has been
+tested with Go 1.8, 1.9 and 1.10.
 
-:exclamation: **NOTE:** This project is in preview and breaking changes are 
-introduced frequently. Therefore, vendoring dependencies is even more
-important than usual. We use [dep](https://github.com/golang/dep).
+To be notified about updates and changes, subscribe to the [Azure update
+feed](https://azure.microsoft.com/updates/).
 
-### Install:
+Users of the SDK may prefer to jump right in to our samples repo at
+[github.com/Azure-Samples/azure-sdk-for-go-samples][samples_repo].
+
+### Build Details
+
+Most packages in the SDK are generated from [Azure API specs][azure_rest_specs]
+using [Azure/autorest.go][] and [Azure/autorest][]. These generated packages
+depend on the HTTP client implemented at [Azure/go-autorest][].
+
+[azure_rest_specs]: https://github.com/Azure/azure-rest-api-specs
+[Azure/autorest]: https://github.com/Azure/autorest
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,43 +4,68 @@
 [![Build Status](https://travis-ci.org/Azure/azure-sdk-for-go.svg?branch=master)](https://travis-ci.org/Azure/azure-sdk-for-go) 
 [![Go Report Card](https://goreportcard.com/badge/github.com/Azure/azure-sdk-for-go)](https://goreportcard.com/report/github.com/Azure/azure-sdk-for-go)
 
-azure-sdk-for-go provides Go packages for using Azure services. It has been
-tested with Go 1.8 and 1.9. To be notified about updates and changes, subscribe
-to the [Azure update feed][].
+azure-sdk-for-go provides Go packages for managing and using Azure services. It has been
+tested with Go 1.8, 1.9 and 1.10.
```
jhendrixMSFT: I think it would be good to reword this so that we don't have to update it every time a new version of Go is released.  Maybe something along the lines of the latest version plus the previous two?

README.md
```diff
@@ -117,84 +146,97 @@ func getServicePrincipalToken() (adal.OAuthTokenProvider, error) {
...
-import "github.com/Azure/azure-sdk-for-go/profiles/2017-03-09/resources/mgmt/..."
 import "github.com/Azure/azure-sdk-for-go/profiles/2017-03-09/storage/mgmt/storage"
 ```
 
-We also provide two special profiles: `latest` and `preview`. These will always
-refer to the most recent stable or preview API versions for each service. For
-example, to automatically use the most recent Compute APIs, use one of the following
-imports:
+The 2017-03-09 profile is the only one currently available and is for use with Azure
+Stack. More profiles are under development.
```
marstr: I wouldn't use the phrasing, I would call out that the `2017-03-09` profile is supposed to help with "Hybrid Cloud" scenarios. "Azure Stack" is only half of that equation. :)

README.md
```diff
@@ -117,84 +146,97 @@ func getServicePrincipalToken() (adal.OAuthTokenProvider, error) {
...
+
+In addition to versioned profiles, we also provide two special profiles
+`latest` and `preview`. These *always* include the most recent respective stable or
+preview API versions for each service, even when updating them to do so causes
+breaking changes.
+
+The `latest` and `preview` profiles can help you stay up to date with API
+updates as you build applications. Since they are by definition not stable,
+however, they **should not** be used in production apps. Instead, choose the
```
marstr: To be super explicit, I may add particular mention here that the SemVer on the repository doesn't cover these profiles.

README.md
```diff
@@ -117,84 +146,97 @@ func getServicePrincipalToken() (adal.OAuthTokenProvider, error) {
...
+
 ## License
 
-Apache 2.0, see [LICENSE][].
+Apache 2.0, see [LICENSE](./LICENSE).
 
 ## Contribute
 
-See [CONTRIBUTING.md][]. 
+See [CONTRIBUTING.md](./CONTRIBUTING.md).
```
marstr: Aren't we supposed to include a "Code of Conduct" sectio?

README.md
joshgav: agree, but can we fix this later?

README.md
joshgav: yes good idea, will fix now

README.md
joshgav: okay will add

README.md
joshgav: maybe, it's in the Contributing guide I think. let's defer fixing this now though

joshgav:
fixed. for later:
* use a [current - 2] formula for supported runtime versions
* check into adding Code of Conduct to top-level README

thanks!

PR #1075, opened 2018-02-12
update storage README to reflect preview state

joshgav:
cc @seguler @marstr 

updated links, removed the test notes since not needed for most users, and put notice about new blobs repo at the top.


joshgav:
added Tables to APIs under development per feedback from @mcardosos

PR #1084, opened 2018-02-13
update storage README to reflect recommended practices

joshgav:
Looks like this didn't land correctly in master so trying again for inclusion in next point release. Thanks!


mcardosos:
Restarted travis job
LGTM

PR #1093, opened 2018-02-15
Marshal TimeRFC1123 as RFC1123 instead of empty string.

tomclegg:
With this, `xml.Marshal(BlobListResponse{...})` (for example) produces XML that can be decoded by the client. This makes it much easier to write a stub server.


mcardosos:
Hello @tomclegg !
Thanks for submitting a PR :)
Change looks good

PR #1097, opened 2018-02-16
Release profileBuilder and storage changes

marstr:
This is a roll-up of changes made in the `dev` branch.

It is necessary to get the automatic PRs working against `latest` and generating the most correct profiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## `v14.1.1`
+
+- Fixing timestamp marshalling bug in the `storage` package.
+- Updating `profileBuilder` to clear-output folders when it is run by `go generate`.
+- Tweaking Swagger -> SDK config to use "latest" instead of "nightly" and be tied to a particular version of autorest.go.
+
 ## `v14.1.0`
 
 ### Changes
--- a/swagger_to_sdk_config.json
+++ b/swagger_to_sdk_config.json
@@ -6,18 +6,18 @@
       "dep ensure",
       "go install",
       "cd ../..",
-      "go generate",
+      "go generate ./profiles/...",
       "gofmt -w ./services/",
       "gofmt -w ./profiles/"
     ],
     "autorest_options": {
-      "use": "@microsoft.azure/autorest.go@preview",
+      "use": "@microsoft.azure/autorest.go@~2",
       "go": "",
       "verbose": "",
       "sdkrel:go-sdk-folder": ".",
       "multiapi": "",
-      "package-version": "nightly",
-      "user-agent": "Azure-SDK-For-Go/nightly services"
+      "package-version": "latest",
+      "user-agent": "Azure-SDK-For-Go/latest services"
     },
     "version": "0.2.0"
   }
--- a/tools/profileBuilder/model/delete.go
+++ b/tools/profileBuilder/model/delete.go
@@ -1,0 +1,48 @@
+// +build go1.9
+
+// Copyright 2018 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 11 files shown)


tools/profileBuilder/model/delete.go
```diff
@@ -0,0 +1,48 @@
...
+	var children []os.FileInfo
+
+	children, err = ioutil.ReadDir(dir)
+	if err != nil {
+		return
+	}
+
+	for _, child := range children {
+		if child.IsDir() {
+			childPath := strings.Replace(path.Join(dir, child.Name()), `\`, `/`, -1)
```
jhendrixMSFT: Why is this necessary?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## `v14.1.1`
+
+- Fixing timestamp marshalling bug in the `storage` package.
+- Updating `profileBuilder` to clear-output folders when it is run by `go generate`.
+- Tweaking Swagger -> SDK config to use "latest" instead of "nightly" and be tied to a particular version of autorest.go.
```
vladbarosan: Nit: from what I see it is tied to particular minor version of autorest.go. Any reason we don't want to tie it to the major version ? That would still mean no breaking changes

tools/profileBuilder/model/delete.go
mcardosos: My guess it depends on OS. I would use the filepath package instead.

---
In reply to: [168828461](https://github.com/Azure/azure-sdk-for-go/pull/1097#discussion_r168828461) [](ancestors = 168828461)

swagger_to_sdk_config.json
```diff
@@ -6,18 +6,18 @@
...
       "go install",
       "cd ../..",
-      "go generate",
+      "go generate ./profiles/...",
       "gofmt -w ./services/",
       "gofmt -w ./profiles/"
     ],
     "autorest_options": {
-      "use": "@microsoft.azure/autorest.go@preview",
+      "use": "@microsoft.azure/autorest.go@~2",
```
vladbarosan: Not sure if autorest just calls npm install on this, but last minor version of autorest.go is 2.1.x .  Usually these versions are specified as ~2.1.0 , not sure what is the expected behavior on ~2

tools/profileBuilder/model/delete.go
vladbarosan: When you say thins you mean the whole delete thing ? because if some profiles go away, they will stick around if we don't delete them

jhendrixMSFT: I was referring to fixing up the path (I get that deleting is required to clear out old packages).  I guess when running on Windows we end up with a mix of Windows and Linux dir separators which is unfortunate.  If the filepath package abstracts all of this away then I agree let's use that.

marstr: Neither `filepath` nor `path` are enough, the separators still don't get normalized without a `strings.Replace` call. https://play.golang.org/p/_dupP3C0ek3

swagger_to_sdk_config.json
marstr: ~2 ties it to the major version, as you propose above :)

CHANGELOG.md
marstr: It is tied to the major version, see below. :)

vladbarosan: I was referring that this yes, to say "particular major version" :D

swagger_to_sdk_config.json
vladbarosan: ok apparently just doing ~2 ties it to the major version just that standard node practice is doing ^2.1.0 but if this works I guess its fine.

Issue #1061, opened 2018-02-08
Create changes report generation tool for commits


jhendrixMSFT:
For an SDK commit we need a tool that will report on the changes introduced.

- APIs added
- Any breaking changes introduced

The tool should generate a categorized (and sorted!) report in either markdown or HTML.  It should also report an exit code that can be used in automation to determine if breaking changes were introduced.

jhendrixMSFT:
Tool added in v14.3.0.

PR #1179, opened 2018-03-01
Add apidiff tool.

jhendrixMSFT:
This tool will generate a report in JSON format describing changes to
public surface area between two specified commits.  It can work on
individual packages or a set of packages under a specified directory.
Please see the help for details on usage of the tool.

### DIFF
--- a/tools/apidiff/cmd/packages.go
+++ b/tools/apidiff/cmd/packages.go
@@ -1,0 +1,225 @@
+// Copyright 2018 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/tools/apidiff/exports"
+	"github.com/Azure/azure-sdk-for-go/tools/apidiff/repo"
+	"github.com/spf13/cobra"
+)
+
+var packagesCmd = &cobra.Command{
+	Use:   "packages [package search dir] [base commit] [target commit]",
+	Short: "Generates report for all packages under the specified directory.",
+	Long: `The packages command generates a report for all of the packages under the directory specified in [package dir].
+The package content in [target commit] is compared against the package content in [base commit]
+to determine what changes were introduced in [target commit].`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		rootDir, cloneRepo, cleanupFn, err := processArgsAndClone(args)
+		if err != nil {
+			return err
+		}
+		defer cleanupFn()
+
+		baseCommit := args[1]
+		targetCommit := args[2]
+
+		// get for lhs
+		vprintf("checking out base commit %s and gathering exports\n", baseCommit)
+		lhs, err := getRepoContentForCommit(cloneRepo, rootDir, baseCommit)
+		if err != nil {
+			return err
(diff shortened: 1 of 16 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/1061.

tools/apidiff/repo/repo.go
```diff
@@ -0,0 +1,90 @@
...
+				return
+			}
+			dir = dir[:i]
+		}
+	}
+	wt.dir = dir
+	return
+}
+
+// Clone calls "git clone", cloning the working tree into the specified directory.
```
vladbarosan: >it clone",  [](start = 17, length = 11)

is this something we can use instead of repo [https://github.com/src-d/go-git](https://github.com/src-d/go-git) ?

tools/apidiff/cmd/packages.go
```diff
@@ -0,0 +1,225 @@
...
+
+	r, err = getExportsForPackages(pkgDirs)
+	if err != nil {
+		err = fmt.Errorf("failed to get exports for commit '%s': %s", commit, err)
+	}
+	return
+}
+
+// contains repo content, it's structured as "package":"apiversion":content
+type repoContent map[string]map[string]exports.Content
```
vladbarosan: >map[string]map[string]exports.Content [](start = 17, length = 37)

cant we have just the package path here and the content ? (i.e. map[string]exports.Content ) or is it helpful later ?

tools/apidiff/repo/repo.go
jhendrixMSFT: I tried this package, unfortunately the perf is terrible.  In my measurements it was literally four times slower (~20 seconds vs ~88 seconds).

tools/apidiff/cmd/packages.go
jhendrixMSFT: It's used to group packages by API version.

tools/apidiff/repo/repo.go
vladbarosan: bummer 👎

tools/apidiff/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (c cmdExitCode) Error() string {
+	return fmt.Sprintf("%v", c.exitCode)
+}
+
+func init() {
+	rootCmd.PersistentFlags().BoolVarP(&onlyAdditionsFlag, "additions", "a", false, "only include additive changes in the report")
+	rootCmd.PersistentFlags().BoolVarP(&onlyBreakingChangesFlag, "breakingchanges", "b", false, "only include breaking changes in the report")
+	rootCmd.PersistentFlags().BoolVarP(&quietFlag, "quiet", "q", false, "suppress console output")
+	rootCmd.PersistentFlags().BoolVarP(&verboseFlag, "verbose", "v", false, "verbose output")
```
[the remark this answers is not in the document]
jhendrixMSFT: I thought about that but as I didn't see an initial need for it I omitted it.  If it turns out we need it later then we can add it.

tools/apidiff/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+var verboseFlag bool
+
+var rootCmd = &cobra.Command{
+	Use:   "apidiff",
+	Short: "Generates a diff of exported package content between two commits.",
+	Long: `This tool will generate a report in JSON format describing changes to
+public surface area between two specified commits.  It can work on
+individual packages or a set of packages under a specified directory.`,
+	SilenceErrors: true,
+	SilenceUsage:  true,
```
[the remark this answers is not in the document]
jhendrixMSFT: By default if a command returns a non-nil error then the cobra infrastructure will dump the help messages.  Normally this is fine however I've used non-nil error from a command to specify an exit code from a command (see the cmdExitCode type below), and in this case I don't want to spew the help message.  I'll admit this is a bit of a hack but I couldn't find a better way to do it (if you know of one I will update the code).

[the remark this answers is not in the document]
marstr: Ah, got it. My usual strategy here is to write an error message to `os.Stderr`, then `os.Exit(/* your desired exit status code */)`. I reserve returning a non-nil error to cobra for the event that my command couldn't parse the parameters.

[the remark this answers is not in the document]
jhendrixMSFT: This has been fixed.

Issue #899, opened 2017-12-14
services/ and profiles/ not showing up in godoc


joshgav:
see e.g. https://godoc.org/github.com/Azure/azure-sdk-for-go/services

expected: services/ and profiles/ show up there.

jhendrixMSFT:
This is apparently a known issue with godoc not recursively scanning for all packages, see https://github.com/golang/gddo/issues/494 for more details.

jhendrixMSFT:
For now the only option we have is to manually visit each package so it will get indexed (I did this for the advisor package, it's now indexed).

jhendrixMSFT:
I should be able to write a tool to automate this.

jhendrixMSFT:
Indexer tool added in v14.3.0.

PR #1181, opened 2018-03-01
Add indexer tool

jhendrixMSFT:
This is used to add missing SDK packages to godoc.org.  Simply run it
from its source directory to index packages under the services folder.

### DIFF
--- a/tools/indexer/main.go
+++ b/tools/indexer/main.go
@@ -1,0 +1,186 @@
+// Copyright 2018 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type packageSet map[string]bool
+
+// adds any missing SDK packages to godoc.org
+func main() {
+	dir := ""
+	if len(os.Args) > 1 {
+		// assume second arg is source dir
+		dir = os.Args[1]
+	} else {
+		// if no args specified assume we're running from the source dir
+		// and calculate the relative path to the services directory.
+		var err error
+		dir, err = filepath.Abs("../../services")
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	pkgs, err := getPackagesForIndexing(dir)
+	if err != nil {
+		panic(err)
+	}
+
+	indexedPkgs, err := getIndexedPackages()
+	if err != nil {
+		panic(err)
+	}
+
+	// for each package in pkgs, check if it's already been
+	// indexed.  if it hasn't been indexed then do so
+
+	for pkg := range pkgs {
+		if _, already := indexedPkgs[pkg]; already {
+			pkgs[pkg] = true
+			continue
+		}
+
+		fmt.Printf("indexing %s...", pkg)
+		resp, err := http.DefaultClient.Get(fmt.Sprintf("https://godoc.org/%s", pkg))
+		if err != nil {
+			panic(err)
+		}
+
+		resp.Body.Close()
+		if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
+			// a 404 means that the package exists locally but not yet in github
+			fmt.Printf("completed with status '%s'\n", resp.Status)
+			pkgs[pkg] = true
+		} else {
+			fmt.Printf("FAILED with status '%s'\n", resp.Status)
+		}
+
+		// sleep a bit between indexing
+		time.Sleep(60 * time.Second)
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/899.

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if failed {
+		fmt.Println("not all packages were indexed")
+		os.Exit(1)
+	}
+	fmt.Println("successfully indexed all packages")
+}
+
+// returns the set of packages that have already been indexed
+func getIndexedPackages() (packageSet, error) {
```
vladbarosan: I presume this gets the href from the html page ? seems ok given that the format is "/github.com/Azure/azure-sdk-for-go/services/advisor/mgmt/2017-03-31/advisor" ?

If it works i think its ok but was wondering if it would be easier to just have a regex match on the pattern with a capture group for the actual pacakge name ( something like  this [https://regex101.com/r/r45IYk/1](https://regex101.com/r/r45IYk/1) ( needs the non greedy option set )

vladbarosan: >getIndexedPackages [](start = 5, length = 18)

also maybe we can have a local html file and have a test on it for this function

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+		pkgs[pkg] = true
+
+		// advance to the next
+		bodyStr = bodyStr[terminal:]
+	}
+	return pkgs, nil
+}
+
+// returns the set of packages, calculated from the repo, to be indexed
+func getPackagesForIndexing(dir string) (packageSet, error) {
```
vladbarosan: >getPackagesForIndexing [](start = 5, length = 22)

I was about to mention to use filepath.Glob but it does not support "**" :'(   [https://github.com/golang/go/issues/11862](https://github.com/golang/go/issues/11862)

vladbarosan: I guess a test for this also  ?

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+		if !v {
+			failed = true
+			break
+		}
+	}
+
+	if failed {
+		fmt.Println("not all packages were indexed")
+		os.Exit(1)
+	}
```
vladbarosan: Ok so I we don't want to loop here until all of them are fine  ? We will loop on the executable itself ? Maybe we can extract the refresh logic in a function and give the option to loop on it until is ready

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+		if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
+			// a 404 means that the package exists locally but not yet in github
+			fmt.Printf("completed with status '%s'\n", resp.Status)
+			pkgs[pkg] = true
+		} else {
+			fmt.Printf("FAILED with status '%s'\n", resp.Status)
+		}
+
+		// sleep a bit between indexing
+		time.Sleep(60 * time.Second)
```
vladbarosan: >time.Sleep(60 * time.Second) [](start = 2, length = 28)

is this to avoid getting blocked by some kind of bot protection ? Maybe we can do a random number between 30-70 or something if you saw that there is still an issue :D

tools/indexer/main.go
jhendrixMSFT: In the original C# implementation that's how it worked however for release process I don't want that behavior by default.  We can add a switch to have it run forever if we think that's necessary.

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	indexedPkgs, err := getIndexedPackages()
+	if err != nil {
+		panic(err)
+	}
+
+	// for each package in pkgs, check if it's already been
+	// indexed.  if it hasn't been indexed then do so
+
```
vladbarosan: maybe add a comment that indexing will happen if there is a get on the package or/and why is that necessary ( maybe link to a docorg github issue if there is anything like that )

tools/indexer/main.go
mcardosos: But some of them might be in local, right?

tools/indexer/main.go
```diff
@@ -0,0 +1,186 @@
...
+	// dirs will look something like "D:\work\src\github.com\Azure\azure-sdk-for-go\services\..."
+	// strip off the stuff before the github.com and change the whacks so it looks like a package import
+
+	pkgs := packageSet{}
+	for _, dir := range leafDirs {
+		i := strings.Index(dir, "github.com")
+		if i < 0 {
+			return nil, fmt.Errorf("didn't find github.com in directory '%s'", dir)
+		}
+		pkg := strings.Replace(dir[i:], "\\", "/", -1)
```
mcardosos: Dont the whacks depend on where this is run?

jhendrixMSFT: True, on non-Windows OS this would end up being a no-op.

tools/indexer/main.go
jhendrixMSFT: You know this might not be an issue any more, I will bump the sleep to 10 seconds and if we have problems in the future we can adjust.

tools/indexer/main.go
jhendrixMSFT: Fixed.

jhendrixMSFT: Fixed.

tools/indexer/main.go
jhendrixMSFT: Fixed.

tools/indexer/main.go
jhendrixMSFT: Fixed.

PR #1376, opened 2018-03-21
Tweaks to profile builder.

jhendrixMSFT:
Include the bad path in the error message.
Close the file handle to avoid race conditions.
Moved formatting to be per file so that it only touches the contents of
the profile being generated.

### DIFF
--- a/tools/profileBuilder/model/transforms.go
+++ b/tools/profileBuilder/model/transforms.go
@@ -23,7 +23,6 @@ package model
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"go/ast"
 	"go/parser"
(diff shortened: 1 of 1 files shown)


tools/profileBuilder/model/transforms.go
```diff
@@ -119,7 +112,7 @@ func GetAliasPath(subject, profile string) (transformed string, err error) {
 
 	matches := packageName.FindAllStringSubmatch(subject, -1)
 	if matches == nil {
-		err = errors.New("path does not resemble a known package path")
+		err = fmt.Errorf("path '%s' does not resemble a known package path", subject)
```
mcardosos: >%s [](start = 26, length = 2)

Cool, shiny new info :D

PR #1390, opened 2018-03-21
Fixes to apidiff tool.

jhendrixMSFT:
Fixed a bug in breakingChanges.isEmpty() that was causing it to return
true when it should have returned false.
Convert relative package(s) path to absolute path.
Added switch "copyrep" to copy the source repo instead of cloning it;
useful for CI so things like FETCH_HEAD are preserved.
Include new struct fields and interface methods in GetExports().
Added package ioext containing various IO extension funcs.

### DIFF
--- a/tools/apidiff/cmd/root.go
+++ b/tools/apidiff/cmd/root.go
@@ -20,6 +20,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var copyRepoFlag bool
 var onlyAdditionsFlag bool
 var onlyBreakingChangesFlag bool
 var quietFlag bool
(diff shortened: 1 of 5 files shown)


tools/apidiff/cmd/root.go
```diff
@@ -34,6 +35,7 @@ individual packages or a set of packages under a specified directory.`,
 }
 
 func init() {
+	rootCmd.PersistentFlags().BoolVarP(&copyRepoFlag, "copyrepo", "c", false, "copy the repo instead of cloning it")
```
vladbarosan: >copyRepoFlag [](start = 37, length = 12)

when would this be useful ? also is the breaking change tool doing any distructive behavior on the code itself ?

tools/apidiff/delta/delta.go
```diff
@@ -46,6 +47,20 @@ func GetExports(lhs, rhs exports.Content) exports.Content {
 		}
 	}
 
+	structFields := GetStructFields(lhs, rhs)
+	if len(structFields) > 0 {
```
vladbarosan: >structFields [](start = 8, length = 12)

hmmm so if there is a new field we would say its a brekaing change ? would be better to do it just if the field is not a pointer ? If so is there  a way to check if its a pointer or not ?

tools/apidiff/cmd/root.go
jhendrixMSFT: It's useful for CI (see the change description where I go into more detail).

tools/apidiff/delta/delta.go
jhendrixMSFT: No, this just includes new struct fields when diff'ing the sides (same goes for the interface method code I added below).

Issue #1487, opened 2018-04-02
Install Go SDK fails


seyadava:
Running bellow command to install Go SDK fails with the following error.
> go get -u github.com/Azure/azure-sdk-for-go/...

ERROR:

# github.com/Azure/azure-sdk-for-go/profiles/preview/preview/iotspaces/mgmt/iotspaces
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:34:21: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.New
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:37:32: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.NewWithBaseURI
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:43:27: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.NewClient
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:46:38: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.NewClientWithBaseURI
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:110:37: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.NewOperationsClient
DRIVE:\Users\USER\go\src\github.com\Azure\azure-sdk-for-go\profiles\preview\preview\iotspaces\mgmt\iotspaces\models.go:113:48: cannot use subscriptionID (type "github.com/satori/uuid".UUID) as type "github.com/satori/go.uuid".UUID in argument to iotspaces.NewOperationsClientWithBaseURI
[...]

Issue #1136, opened 2018-02-22
Handle preview specs properly


vladbarosan:
Right now in the https://github.com/Azure/azure-rest-api-specs specs go in 2 folders:
 - **stable** - spec is ready to be consumed and it is not expecting breaking changes.
 - **preview** - spec is ready to be consumed but breaking changes are acceptable and expected.

To handle this, for API version which are in the preview folder ( regardless if the api version is preview or not) we will put them in a `services/preview` folder.  NOTE:
  - everything here is expected to have breaking changes. We will mark this explicitly that semantic versioning does not apply here.
  - api versions that currently exist and need to be in preview folder will exist in 2 places (in the preview folder and outside). The path outside of the preview folder will be marked as deprecated and wont receive any updates.  At one point it will be removed.

vladbarosan:
@jhendrixMSFT , @joshgav here is the issue for handling previews.

jhendrixMSFT:
Scrub is done and merged into latest.

PR #1498, opened 2018-04-03
v15.1.0

jhendrixMSFT:


vladbarosan:
#1487 I reporoed the issue. 

2 causes:
1. We didn't regenerate the preview folder so some types are mismatchign ( the repo name switch from uuid to go.uuid from Satori) . Let's fot 15.1 maybe manually fix the profile so at least go get works. then after https://github.com/Azure/azure-sdk-for-go/issues/1136 is done it should work.

2. For some reason some Automation operations and Iot although they have a rsourcegroup parameter in all operations , one of them thinks is a client property not a parameter. I had a look at the specs for Automation in `Dscconfiguration` and couldn't see a reason for it.  Fixing it would mean making one client call a breaking change but at this point is not working either way.

vladbarosan:
@mcardosos can you have a look at why the CI didn't catch these ? I know there was a similar issue in the samples repo

vladbarosan:
@mcardosos apparently it takes the exit code of the last command we need to check for the status code of each command. and fail if its non 0

PR #1533, opened 2018-04-09
Adding CI line which runs the breaking change tool.

marstr:
For now, it will print to the log whether or not it found any, but its
exit code is ignored. Later, the mask will be flipped. This could make
it harder temporarily to collect vetting information around the success
or failure of the breaking change tool. However, publishing status to an
EventGrid topic could allow us to stow results in a DB then report on it
and create alerting rules.


marstr:
I dug into what's going on with the `default` job, and I've attempted to reconfigured Jenkins to not add this commit status in the future.

mcardosos:
🐑 it!

Issue #1500, opened 2018-04-04
Unexpected quantity of TCP connections


meirf:
I ran this code at https://github.com/Azure/azure-sdk-for-go/commit/56332fec5b308fbb6615fa1af6117394cdba186d

I'm seeing an unexpected number of TCP connections when copying files.

```
    package main
    
    import (
    	"net/http"
    	"sync"
    	"time"
    
    	"github.com/Azure/azure-sdk-for-go/storage"
    )
    
    func main() {
    	// create and initialize a client
    	client, _ := storage.NewBasicClient(ACCOUNT_NAME, ACCOUNT_KEY)
    
    	// use DefaultTransport with configs that limit conns
    	client.HTTPClient.Transport = http.DefaultTransport
    	transp, _ := client.HTTPClient.Transport.(*http.Transport)
    	client.HTTPClient.Transport = http.DefaultTransport
    	transp.MaxIdleConnsPerHost = 200
    	transp.MaxIdleConns = 200
    
    	rootClient := client.GetBlobService().GetContainerReference("foo")
    	numRoutines := 200
    	numRequestsPerRoutine := 100
    
    	var wg sync.WaitGroup
    	for i := 0; i < numRoutines; i++ {
    		wg.Add(1)
    		go func() {
    			for j := 0; j < numRequestsPerRoutine; j++ {
    				sourceBlob := rootClient.GetBlobReference("fileA")
    				destBlob := rootClient.GetBlobReference("dirX/fileA")
    				destBlob.Copy(sourceBlob.GetURL(), nil)
    				time.Sleep(250 * time.Millisecond)
    			}
    			wg.Done()
    		}()
    	}
    	wg.Wait()
    }
    time.Sleep(5 * time.Minute)
```

Because `MaxIdleConnsPerHost` is set to a high number we can reuse a lot more descriptors, but I still see more than what I expected. 

**Expected**
I expected there to be a maximum of 200 connections since I assume each goroutine is executing http requests in serial. 

**Observed**
Until the end of the execution, the number of TCP connections stays under 200. But towards the end that number consistently jumps to over 400 and sometimes 800 and I start getting `dial tcp: lookup `FOO.blob.core.cloudapi.de on 10.1.2.3:53: dial udp 10.1.2.3:53: socket: too many open files)`
[...]

PR #1540, opened 2018-04-10
Drain response body without allocating a buffer.

jhendrixMSFT:
Added func drainRespBody() that copies the HTTP response body into
nothingness then closes it, avoiding allocations.
When reading/closing the HTTP response body, for cases that don't
require the data call drainRespBody() instead of readAndCloseBody().

### DIFF
--- a/storage/blob.go
+++ b/storage/blob.go
@@ -140,7 +140,7 @@ func (b *Blob) Exists() (bool, error) {
 	headers := b.Container.bsc.client.getStandardHeaders()
 	resp, err := b.Container.bsc.client.exec(http.MethodHead, uri, headers, nil, b.Container.bsc.auth)
 	if resp != nil {
-		defer readAndCloseBody(resp.Body)
+		defer drainRespBody(resp)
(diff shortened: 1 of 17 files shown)


marstr:
I was reviewing this code and thought about making a similar PR. This is definitely a step up from what exists right now, because it should save a bunch of allocations. I didn't submit the PR though, because I couldn't convince myself that `resp.Body.Close()` is insufficient. I seem to remember there being a reason we needed to actually ready the Body (as you do here with the call `io.Copy(ioutil.Discard, resp.Body)`, but do you know why that is?

jhendrixMSFT:
It's to reuse keep-alive connections, see https://golang.org/pkg/net/http/#Response.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/1500

marstr:
Doc Comment above the `Body` of the the `Response` type:
> // ... The default HTTP client's Transport may not
> // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
> // not read to completion and closed.

That's what I was trying to find! Thanks for finding it for me @jhendrixMSFT

PR #1911, opened 2018-05-29
Update README

### LINKED ISSUE #1445: add table of Azure packages to README
e.g.:

* https://github.com/Azure/azure-event-hubs-go
* https://github.com/Azure/azure-storage-blobs-go
* https://github.com/Microsoft/ApplicationInsights-Go

joshgav:
* adds table of other Azure Go packages (fixes #1445)
* edits for clarity in authentication section and a couple other small additions

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,16 @@
 [![Build Status](https://travis-ci.org/Azure/azure-sdk-for-go.svg?branch=master)](https://travis-ci.org/Azure/azure-sdk-for-go)
 [![Go Report Card](https://goreportcard.com/badge/github.com/Azure/azure-sdk-for-go)](https://goreportcard.com/report/github.com/Azure/azure-sdk-for-go)
 
-azure-sdk-for-go provides Go packages for managing and using Azure services. It has been
-tested with Go 1.8, 1.9 and 1.10.
+azure-sdk-for-go provides Go packages for managing and using Azure services.
+It is continuously tested with Go 1.8, 1.9, 1.10 and master.
 
 To be notified about updates and changes, subscribe to the [Azure update
 feed](https://azure.microsoft.com/updates/).
 
-Users of the SDK may prefer to jump right in to our samples repo at
+Users may prefer to jump right in to our samples repo at
 [github.com/Azure-Samples/azure-sdk-for-go-samples][samples_repo].
 
-### Build Details
+## Package Updates
 
 Most packages in the SDK are generated from [Azure API specs][azure_rest_specs]
 using [Azure/autorest.go][] and [Azure/autorest][]. These generated packages
@@ -25,16 +25,30 @@ depend on the HTTP client implemented at [Azure/go-autorest][].
 [Azure/go-autorest]: https://github.com/Azure/go-autorest
 
 The SDK codebase adheres to [semantic versioning](https://semver.org) and thus
-avoids breaking changes other than at major (x.0.0) releases. However,
-occasionally Azure API fixes require breaking updates within an individual
-package; these exceptions are noted in release changelogs.
+avoids breaking changes other than at major (x.0.0) releases. Because Azure's
+APIs are updated frequently, we release a **new major version at the end of
+each month** with a full changelog. For more details and background see [SDK Update
+Practices](https://github.com/Azure/azure-sdk-for-go/wiki/SDK-Update-Practices).
 
 To more reliably manage dependencies like the Azure SDK in your applications we
 recommend [golang/dep](https://github.com/golang/dep).
 
+## Other Azure Go Packages
+
+Azure provides several other packages for using services from Go, listed below.
+If a package you need isn't available please open an issue and let us know.
+
+| Service | Import Path/Repo |
+|---------|------------------|
+| Storage - Blobs | [github.com/Azure/azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) |
+| Storage - Files | [github.com/Azure/azure-storage-file-go](https://github.com/Azure/azure-storage-file-go) |
+| Storage - Queues | [github.com/Azure/azure-storage-queue-go](https://github.com/Azure/azure-storage-queue-go) |
+| Event Hubs | [github.com/Azure/azure-event-hubs-go](https://github.com/Azure/azure-event-hubs-go) |
+| Application Insights | [github.com/Microsoft/ApplicationInsights-go](https://github.com/Microsoft/ApplicationInsights-go) |
+
 # Install and Use:
 
-### Install
+## Install
 
 ```sh
 $ go get -u github.com/Azure/azure-sdk-for-go/...
@@ -48,23 +62,26 @@ $ dep ensure -add github.com/Azure/azure-sdk-for-go
 
 If you need to install Go, follow [the official instructions](https://golang.org/dl/).
 
-### Use
+## Use
 
-For complete examples of many scenarios see [Azure-Samples/azure-sdk-for-go-samples][samples_repo].
+For many more scenarios and examples see
+[Azure-Samples/azure-sdk-for-go-samples][samples_repo].
+
+Apply the following general steps to use packages in this repo. For more on
+authentication and the `Authorizer` interface see [the next
+section](#authentication).
 
 1. Import a package from the [services][services_dir] directory.
 2. Create and authenticate a client with a `New*Client` func, e.g.
    `c := compute.NewVirtualMachinesClient(...)`.
-3. Invoke API methods using the client, e.g. `c.CreateOrUpdate(...)`.
-4. Handle responses.
+3. Invoke API methods using the client, e.g.
+   `res, err := c.CreateOrUpdate(...)`.
+4. Handle responses and errors.
 
 [services_dir]: https://github.com/Azure/azure-sdk-for-go/tree/master/services
 
 For example, to create a new virtual network (substitute your own values for
 strings in angle brackets):
-
-Note: For more on authentication and the `Authorizer` interface see [the next
-  section](#authentication).
 
 ```go
 package main
@@ -72,18 +89,22 @@ package main
 import (
 	"context"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2017-09-01/network"
+
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
 func main() {
+	// create a VirtualNetworks client
 	vnetClient := network.NewVirtualNetworksClient("<subscriptionID>")
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
 
+	// create an authorizer from env vars or Azure Managed Service Idenity
+	authorizer, err := auth.NewAuthorizerFromEnvironment()
 	if err == nil {
 		vnetClient.Authorizer = authorizer
 	}
 
+	// call the VirtualNetworks CreateOrUpdate API
 	vnetClient.CreateOrUpdate(context.Background(),
 		"<resourceGroupName>",
 		"<vnetName>",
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,15 +29,32 @@ avoids breaking changes other than at major (x.0.0) releases. However,
...
+| Application Insights | [github.com/Microsoft/ApplicationInsights-go](https://github.com/Microsoft/ApplicationInsights-go) |
+
 # Install and Use:
 
-### Install
+## Install
 
 ```sh
-$ go get -u github.com/Azure/azure-sdk-for-go/...
+$ go get -u github.com/Azure/azure-sdk-for-go
```
marstr: This will fail to fetch any transitive dependencies. The most sure fire way to get everything you need is to include the `/...`.

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+        - `AZURE_TENANT_ID`: Specifies the Tenant to which to authenticate.
+        - `AZURE_CLIENT_ID`: Specifies the app client ID to use.
+        - `AZURE_CERTIFICATE_PATH`: Specifies the certificate Path to use.
+        - `AZURE_CERTIFICATE_PASSWORD`: Specifies the certificate password to use.
+
+    3. **Resource Owner**: Azure AD User and Password. This grant type is *not
+       recommended*, use client credentials (service principals) instead.
+
+        - `AZURE_TENANT_ID`: Specifies the Tenant to which to authenticate.
+        - `AZURE_CLIENT_ID`: Specifies the app client ID to use.
```
marstr: I really really wish that we had two separate environment variables for this. It doesn't make any sense to me that we've overloaded it.

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+        - `AZURE_PASSWORD`: Specifies the password to use.
+
+    4. **Azure Managed Service Identity**: Delegate credential management to the
+       platform. Requires that code is running in Azure, e.g. on a VM. All
+       configuration is handled by Azure. See [Azure Managed Service
+       Identity](https://docs.microsoft.com/en-us/azure/active-directory/msi-overview)
+       for more details.
+
+* The `auth.NewAuthorizerFromFile()` method creates an authorizer using tokens
+  serialized and cached by the [Azure CLI][]. Follow these steps to utilize:
```
marstr: I believe adding this method was a mistake. It means that we've defacto taken a dependency on the CLI without actually declaring it or enforcing it in any way. For the sake of the README, I'd try to stay vague about the source of these files. Honestly, I may just skip it for now, and look into whether or not we have a means of generalizing the source of files with the schema.

For getting at the tokens as generated by the CLI, I'd rather take a true dependency on the CLI, and shell out with the command `az account get-access-token`.

storage/README.md
```diff
@@ -1,9 +1,13 @@
 # Azure Storage SDK for Go (Preview)
 
 :exclamation: IMPORTANT: This package is in maintenance only and will be deprecated in the
-future. Consider using the new package for blobs currently in preview at
-[github.com/Azure/azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go).
-New Table, Queue and File packages are also in development.
+future. Please use one of the following packages instead.
```
vladbarosan: since we have all 3 of them now, dont we want to say they are deprecated and will be removed in the future ?

README.md
```diff
@@ -29,15 +29,32 @@ avoids breaking changes other than at major (x.0.0) releases. However,
 occasionally Azure API fixes require breaking updates within an individual
 package; these exceptions are noted in release changelogs.
```
vladbarosan: maybe we can merge this sentence with the next one you added ?

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+vnetClient.CreateOrUpdate(context.Background(),
+// ...
+```
+
+The following environment variables help determine authentication configuration:
+
+- `AZURE_ENVIRONMENT`: Specifies the Azure Environment to use. If not set, it
+  defaults to `AzurePublicCloud`. Not applicable to authentication with Managed
+  Service Identity (MSI).
+- `AZURE_AD_RESOURCE`: Specifies the AAD resource ID to use. If not set, it
```
vladbarosan: >AZURE_AD_RESOURCE [](start = 3, length = 17)

maybe specify that in cases that multiple resources need to be added we have a helper also `NewAuthorizerFromEnvironmentWithResource()` where they can specify the resource directly ) [https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L61](https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L61)

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+  defaults to `ResourceManagerEndpoint` for operations with Azure Resource
+  Manager.
+
+## Authentication Details
+
```
vladbarosan: you dont want this to be a sub section of auth ?

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+        - `AZURE_CLIENT_SECRET`: Specifies the app secret to use.
+
+    2. **Client Certificate**: Azure AD Application ID and X.509 Certificate.
+
+        - `AZURE_TENANT_ID`: Specifies the Tenant to which to authenticate.
+        - `AZURE_CLIENT_ID`: Specifies the app client ID to use.
+        - `AZURE_CERTIFICATE_PATH`: Specifies the certificate Path to use.
+        - `AZURE_CERTIFICATE_PASSWORD`: Specifies the certificate password to use.
+
+    3. **Resource Owner**: Azure AD User and Password. This grant type is *not
```
vladbarosan: >Resource Owner [](start = 9, length = 14)

is this the recommended name for it ?

vladbarosan: if not maybe User Credentials ?

---
In reply to: [195191337](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195191337) [](ancestors = 195191337)

README.md
vladbarosan: with what are we overloading it ? doesn't always refer to the app id to use ?

---
In reply to: [195175985](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195175985) [](ancestors = 195175985)

README.md
vladbarosan: First of all the description seems wrong. not sure what you mean that it uses tokens serialized and cached by the Azure CLI. All this method does is taking doing client credentials auth but with the info in a file instead of env var

---
In reply to: [195178171](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195178171) [](ancestors = 195178171)

vladbarosan: @marstr@microsoft.com  if that was your concern its not an issue. That being said, im not a big fan of the auth file thing since its format originated from CLI not from AAD and then it seems all the SDK's kind of went along with it.

---
In reply to: [195193861](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195193861) [](ancestors = 195193861,195178171)

vladbarosan: so maybe we can remove it but maybe for some customers its useful. so maybe remove it form the readme or add an experimental or some kind of info showing that is not something 100% standard from what I am aware at least.

---
In reply to: [195194277](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195194277) [](ancestors = 195194277,195193861,195178171)

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+       recommended*, use client credentials (service principals) instead.
```
vladbarosan: maybe add that if this is desired for interactive user access they can use Device Flow for 2FA access

README.md
```diff
@@ -112,52 +136,111 @@ func main() {
...
+       create-for-rbac --sdk-auth > client_credentials.json`.
+    2. Set environment variable `AZURE_AUTH_LOCATION` to the path of the saved
+       output file.
+    3. Use the authorizer returned by `auth.NewAuthorizerFromFile()` in your
+       client as described above.
+
+[Azure CLI]: https://github.com/Azure/azure-cli
+
+* Finally, you can use OAuth's [Device Flow][] by calling
+  `auth.NewDeviceFlowConfig()` and extracting the Authorizer as follows:
```
vladbarosan: >NewDeviceFlowConfig [](start = 8, length = 19)

maybe specify that this is for 2FA access for people who are not aware.

storage/README.md
vladbarosan: and also actually deprecate it. and I saw you are doing the PR directly to master, dont we want to put it in latest ?

---
In reply to: [195187726](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195187726) [](ancestors = 195187726)

README.md
joshgav: I think they're about different things. The previous sentence is talking about breaking changes *within* a major, the next sentence is talking about us releasing new majors frequently.

Do you perhaps mean combine the paragraphs? I prefer to break up paragraphs pretty frequently cause I believe people tend to skim the end of long paragraphs, so I'd rather keep them separate.

README.md
joshgav: Since we're evaluating authentication algorithms across SDKs let's leave this as is for now (it's based on the current version) and consider adding more later.

README.md
joshgav: https://tools.ietf.org/html/rfc6749#section-1.3.3

README.md
joshgav: It means the same thing here as in the other flows - the ID of the app requesting access "on behalf of" the resource owner.

README.md
vladbarosan: Resource Owner Password 👎   :D

README.md
vladbarosan: yes but the thing is we are actually not doing that right, we have monthly releases on which we bump the major version

README.md
joshgav: @marstr this isn't using `~/.azure/accessTokens.json`, the CLI's internal serialization file. It's using the format generated by `az ad sp create-for-rbac --name 'delete-me' --sdk-auth`. This is supported in other SDKs and we should continue supporting it here at least for now.

@vladbarosan good point that it would be better if this came from AAD, but we've decided to define and pursue this format in the SDKs/CLIs team now, in fact we're working on standards to apply across all languages.

README.md
joshgav: @vladbarosan 2FA is dependent on the user account that's logging in, not the login method utilized. For example, if I sign in with authz code flow I also have to approve via 2FA if it's configured for my user account. If I don't have 2FA enabled I don't need it for device flow either.

storage/README.md
joshgav: maybe, but I'm not yet sure how we'll handle tables so didn't change this yet. will open an issue

README.md
joshgav: true, but the first sentence is to notify people in advance of the reality. perhaps we no longer do this cause we can defer to end of the month and latest? then I can take the sentence out.

README.md
vladbarosan: 👍

README.md
vladbarosan: yea either way this thing will always be confusing for users and not only. in the AAD Graph API there   there is even no client id field for neither the SP object or App Object. Not sure why they went with calling it client id instead of app id

README.md
joshgav: actually I'll add this directly to this paragraph, makes sense

storage/README.md
joshgav: https://github.com/Azure/azure-sdk-for-go/issues/2037

joshgav:
@marstr can you take a look again? Also, should I be targeting latest or master (see [comment](https://github.com/Azure/azure-sdk-for-go/pull/1911#discussion_r195196475))?

vladbarosan:
@jhendrixMSFT do we want to target this to latest ? Otherwise I guess we need to make a release for this.

jhendrixMSFT:
Targeting latest would simplify the release, i.e. we can just cherry-pick it along with the rest of the non-breaking changes and release them in one revision.  I don't feel super-strong about it though so if you want to merge it into master be sure to update the patch number and tag it accordingly.

joshgav:
landed in latest as b9f76dbde4ccc31ebee29b1a2974885a613f8f87

PR #1942, opened 2018-06-01
Release v17.0

vladbarosan:


vladbarosan:
I'll have another look but I ran a script on the report and I think all updated have a breaking change

PR #2009, opened 2018-06-11
storage/Client.useHTTPS should default to true when using SAS URLs

florianholzapfel:
Hi,
I have a blob storage that only allows access using HTTPS. When I generate a SAS URL to a container in that blob storage (using the Azure Storage Explorer), the generated SAS URL does not include the ```spr``` field. I can use the SAS URL to access the container in the Azure Storage Explorer, but neither ```NewAccountSASClient``` nor ```NewAccountSASClientFromEndpointToken```.

I could not find any documentation on the default value of the ```spr``` field, but I assume it is ```https```. Therefore I made this PR.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -343,7 +343,7 @@ func NewAccountSASClient(account string, token url.Values, env azure.Environment
 
 	// Get API version and protocol from token
 	c.apiVersion = token.Get("sv")
-	c.useHTTPS = token.Get("spr") == "https"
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Hello!  Thanks for your contribution.  Please note that the blob storage functionality in this package has been deprecated, thus we aren't accepting any new contributions to it.  The functionality has been replaced by the [azblob](https://github.com/azure/azure-storage-blob-go) package which has much better support for SAS authentication.

Issue #2119, opened 2018-06-25
storage.GetContainerReferenceFromSASURI doesn't seem to work properly


ncw:
If I try this program with a SAS URL generated from the Azure portal for a single container I get this error

    GetProperties failed: Head http://.blob./myorganization-test1/hello.txt: dial tcp: lookup .blob.: no such host

The identical code works if I use an account/key or connection string, which leads me to think there must be something from with `storage.GetContainerReferenceFromSASURI`.  Some things do work using the SAS URL like listing the container.

```go
package main

import (
	"fmt"
	"log"
	"net/url"

	"github.com/Azure/azure-sdk-for-go/storage"
)

const SAS_URL = `https://myorganization.blob.core.windows.net/myorganization-test1?sp=rwdl&st=2018-06-25T13:21:14Z&se=2018-06-26T13:21:14Z&sv=2017-11-09&sig=XXX%3D&sr=c`

func main() {
	URL, err := url.Parse(SAS_URL)
	if err != nil {
		log.Fatalf("failed to parse SAS URL: %v", err)
	}
	cc, err := storage.GetContainerReferenceFromSASURI(*URL)
	if err != nil {
		log.Fatalf("failed get container reference: %v", err)
	}
	blob := cc.GetBlobReference("hello.txt")
	getPropertiesOptions := storage.GetBlobPropertiesOptions{}
	err = blob.GetProperties(&getPropertiesOptions)
	if err != nil {
		log.Fatalf("GetProperties failed: %v", err)
	}
	fmt.Printf("blob = %#v", blob)
}
```

Debugging the HTTP transaction you can see it appears to use an incorrect value for the `Host:` field, hence the error.

```
2018/06/25 15:08:38 DEBUG : HTTP REQUEST (req 0xc420444200)
2018/06/25 15:08:38 DEBUG : HEAD /rclone-test1/hello.txt HTTP/1.1
Host: .blob.
User-Agent: rclone/v1.42-DEV
x-ms-date: Mon, 25 Jun 2018 14:08:38 GMT
x-ms-version: 2016-05-31

2018/06/25 15:08:38 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
[...]

PR #2123, opened 2018-06-25
Fixes for blob storage SAS URI parsing

jhendrixMSFT:
storage.GetContainerReferenceFromSASURI() didn't set the storage account
name and other fields when creating a SAS client from a URI.

### DIFF
--- a/storage/blobserviceclient.go
+++ b/storage/blobserviceclient.go
@@ -69,7 +69,11 @@ func GetContainerReferenceFromSASURI(sasuri url.URL) (*Container, error) {
 	if len(path) <= 1 {
 		return nil, fmt.Errorf("could not find a container in URI: %s", sasuri.String())
 	}
-	cli := newSASClient().GetBlobService()
+	c, err := newSASClientFromURL(&sasuri)
+	if err != nil {
+		return nil, err
+	}
+	cli := c.GetBlobService()
 	return &Container{
 		bsc:    &cli,
 		Name:   path[1],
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2119

vladbarosan:
version update ?

jhendrixMSFT:
Will include that in v17.5 (this is into latest)

storage/blobserviceclient.go
```diff
@@ -69,7 +69,11 @@ func GetContainerReferenceFromSASURI(sasuri url.URL) (*Container, error) {
 	if len(path) <= 1 {
 		return nil, fmt.Errorf("could not find a container in URI: %s", sasuri.String())
 	}
-	cli := newSASClient().GetBlobService()
+	c, err := newSASClientFromURL(&sasuri)
+	if err != nil {
+		return nil, err
+	}
+	cli := c.GetBlobService()
```
vladbarosan: >GetBlobService [](start = 10, length = 14)

can't SAS be only for specific types of azure storage api's ( i.e. just for queue service ). In that case when is the lack of permission discovered ?

vladbarosan:
@jhendrixMSFT ok. I didnt notice it was against latest.

storage/blobserviceclient.go
jhendrixMSFT: should be discovered when the API call is made.

vladbarosan: hmm it should be discoverable based on the URI I think, so we should be able to fail faster, either way this is the deprecated package so maybe not worth investing here.

ncw:
I have tested this code and it looks like it fixes the issue - thanks :-)

PR #2154, opened 2018-06-29
v18

vladbarosan:


vladbarosan:
Postponing for Monday so we get a breaking change coming from app insights specs.

PR #2204, opened 2018-07-09
add GitHub templates

joshgav:
Updates and adds PR and issue templates. Multiple issue templates provide experience described [here](https://help.github.com/articles/creating-issue-templates-for-your-repository/); I determined GitHub's expected format by trying it with the described UI :).

Targeting `master` cause these templates only work when on the master branch.

PR #2234, opened 2018-07-16
[backport:master] add GitHub templates

joshgav:
Backport of #2204, lands cleanly :). I think the process once someone approves would be to land in master and tag a new patch (0.0.**++x**) release. @jhendrixMSFT can you confirm? Thanks!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,29 @@
+---
+name: Bug Report
+about: Report bugs in Azure Go SDK packages.
+---
+
+### Bug Report
+
+<!--
+Thank you for opening a bug report. For faster processing, please include:
+-->
+
+* import path of package in question, e.g. `.../services/compute/mgmt/2018-06-01/compute`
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Do we need to tag a release?  None of this touches shipping assets.

joshgav:
I was wondering the same :). To be sure, I think you're saying we only need to tag patches and minors which users would want to receive on `go get -u` or `dep ensure -update`, i.e. code changes.

To say it in other words, when users specify "a version compatible with 18.0.0" in their dependencies list, e.g. `~18.0.0` in dep, they'll receive all *tagged* patches after 18, but not *untagged* patches (until that untagged patch is included in a later tag). Patches like the one in this PR which only effect docs can remain untagged since users don't need to update and get them; and they'll begin to be automatically included when a later patch with code changes is tagged.

If you agree, once this PR is approved I'll just land in master and go. When you merge in changes from latest at the end of the month it should be a fast-forward commit since the changes in this PR match the changes in #2204.

joshgav:
ping @jhendrixMSFT can you comment and/or approve? Thanks!

PR #2361, opened 2018-08-01
update dep to v0.5.0

jhendrixMSFT:
 
### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ go:
 
 env:
   global:
-    - DEP_VERSION="0.4.1"
-    - DEP_SHA=31144e465e52ffbc0035248a10ddea61a09bf28b00784fd3fdd9882c8cbb2315
+    - DEP_VERSION="0.5.0"
+    - DEP_SHA=287b08291e14f1fae8ba44374b26a2b12eb941af3497ed0ca649253e21ba2f83
     - IGNORE_BREAKING_CHANGES=true
 
 matrix:


.travis.yml
```diff
@@ -9,8 +9,8 @@ go:
 
 env:
   global:
-    - DEP_VERSION="0.4.1"
-    - DEP_SHA=31144e465e52ffbc0035248a10ddea61a09bf28b00784fd3fdd9882c8cbb2315
+    - DEP_VERSION="0.5.0"
+    - DEP_SHA=287b08291e14f1fae8ba44374b26a2b12eb941af3497ed0ca649253e21ba2f83
```
vladbarosan: Now that there is an install script, we could do something like this:

```
DEP_RELEASE_TAG=v0.5.0 # so the script knows which version to use 
curl -sSL [https://raw.githubusercontent.com/golang/dep/master/install.sh](https://raw.githubusercontent.com/golang/dep/master/install.sh) | sh \

vladbarosan:
>   - grep -L -r --include *.go --exclude-dir vendor -P "Copyright (\d{4}|\(c\)) Microsoft" ./ | tee /dev/stderr | test -z "$(< /dev/stdin)"

also  i presume there is no update on the lock files because those already came in from the Auto PR ?

---
Refers to: .travis.yml:29 in c468073. [](commit_id = c46807304bedcabbb2ded1b6a68dccc954404d6b, deletion_comment = False)

.travis.yml
jhendrixMSFT: Good idea will fix.

PR #2786, opened 2018-09-20
update dependencies

jhendrixMSFT:
 
### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -22,7 +22,7 @@
 
 [[constraint]]
   name = "github.com/Azure/go-autorest"
-  version = "10.15.3"
+  version = "10.15.4"
 
 [[constraint]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -22,7 +22,7 @@
 
 [[constraint]]
   name = "github.com/Azure/go-autorest"
-  version = "10.15.3"
```
marstr: I'd rather update just the lock file, and set the constraint to "anything above this min version we need for the sdk"

jhendrixMSFT: Derp indeed.  I'll update the constraint to `^10.15.4`.

jhendrixMSFT: Hmm actually `dep` will assume `^` operator by default.  I think we can merge this as-is but in the future we don't need to update the `.toml` file.

Issue #2907, opened 2018-10-04
Add package validation to CI


jhendrixMSFT:
Add a CI task to catch some common mistakes we see.

- Package and directory name mismatch
- Upper-case letters in package directories
- Using build tag instead of API version in package directory
- Incomplete API version in package directory (e.g. `2018-01` instead of `2018-01-01`)
- Incorrect package directory (e.g. /services/< not supposed to be here >/foo/mgmt/...)
- Preview packages not being placed under preview subdirectory (this task will have to go in the swagger->sdk bot)

jhendrixMSFT:
Also update CI to verify that there are no modified files.

jhendrixMSFT:
Fixed.

PR #2925, opened 2018-10-08
Add pkgchk tool and integrate into CI

jhendrixMSFT:
pkgchk will verify the following things:
- package and directory names match
- all lower-case characters in package path
- directory for ARM packages are well-formed

Updated swagger->sdk to specify the preview-chk switch to verify that a
preview swagger has its codegen under a preview subdirectory.

### DIFF
--- a/tools/pkgchk/cmd/root.go
+++ b/tools/pkgchk/cmd/root.go
@@ -1,0 +1,254 @@
+// Copyright 2018 Microsoft Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"bufio"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"unicode"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var excepFileFlag string
+
+var rootCmd = &cobra.Command{
+	Use:   "pkgchk <dir>",
+	Short: "Performs package validation tasks against all packages found under the specified directory.",
+	Long: `This tool will perform various package validation checks against all of the packages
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Note this will fail until I finish cleaning up the repo (can't happen until https://github.com/Azure/azure-rest-api-specs/pull/4089 is merged).

jhendrixMSFT:
Fixes most of https://github.com/Azure/azure-sdk-for-go/issues/2907

jhendrixMSFT:
autorest.go 2.1.115 has dropped so we can move ahead with this PR.

tools/pkgchk/cmd/root.go
```diff
@@ -0,0 +1,254 @@
...
+	})
+	return pkgs, err
+}
+
+type verifier func(p pkg) error
+
+// returns a list of verifiers to execute
+func getVerifiers() []verifier {
+	return []verifier{
+		verifyPkgName(),
```
vladbarosan: can't we have functions directly here without returning them from another function ? it seems that the wrapper function doesn't much except returning the actual verifier

tools/pkgchk/cmd/root.go
```diff
@@ -0,0 +1,254 @@
...
+		verifyLowerCase(),
+		verifyDirectorySturcture(),
+	}
+}
+
+// ensures that the leaf directory name matches the package name
+func verifyPkgName() verifier {
```
vladbarosan: nit: maybe more explicit name ( e.g. verifyPkgMatchesDir)

tools/pkgchk/cmd/root.go
jhendrixMSFT: Good catch, this is a remnant from an earlier iteration I forgot to refactor.

tools/pkgchk/cmd/root.go
jhendrixMSFT: Fixed.

tools/pkgchk/cmd/root.go
```diff
@@ -0,0 +1,254 @@
...
+	return func(p pkg) error {
+		// for ARM the package directory structure is highly deterministic:
+		// /redis/mgmt/2015-08-01/redis
+		// /resources/mgmt/2017-06-01-preview/policy
+		// /preview/signalr/mgmt/2018-03-01-preview/signalr
+		if !p.isARMPkg() {
+			return nil
+		}
+		regexStr := strings.Join([]string{
+			`^(?:/preview)?`,
```
vladbarosan: is this regex tested ?

vladbarosan: by that I mean the examples above are exhaustive  ?

jhendrixMSFT: for ARM yes and the majority of packages match this pattern.  there are just a handful of exceptions which is why I added the exceptions.txt mechanism.

PR #2992, opened 2018-10-12
fix golint path

karthikraobr:
Golint has been moved.

More info here -> https://github.com/golang/lint#installation and golang/lint#417


jhendrixMSFT:
@vladbarosan in fact I don't think we even use `golint` in CI anymore so if true feel free to remove it when creating v21.2.0.

jhendrixMSFT:
This was fixed in the v21.2.0 release.

PR #3012, opened 2018-10-15
v21.2.0

jhendrixMSFT:


README.md
```diff
@@ -315,51 +341,6 @@ import "github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
 import "github.com/Azure/azure-sdk-for-go/profiles/preview/compute/mgmt/compute"
 ```
 
-## Inspecting and Debugging
```
vladbarosan: why remove this part ?

jhendrixMSFT: This might have been due to a bad merge, will check...

jhendrixMSFT: This is due to some changes that were made directly in master thus they aren't in latest.  Will fix then update this PR.

Issue #3111, opened 2018-10-23
breaking api changes post-release


flyinprogrammer:
### Bug Report
https://github.com/Azure/azure-sdk-for-go/commit/b253a0ed38da9645a818bbb6e491c1f7ec1f0bd7

Is a breaking change for our 2018-06-01 API release.

So are these dates completely meaningless? And should we only rely on the package version `v21.3.0` ?

In which case, why even bother with dating these packages since their model contracts can change at any time.

flyinprogrammer:
> seems to be only additive
is simply not true:
https://cl.ly/657ceac352f6

`VirtualMachineSizeTypes` was added as a prefix to all the constants virtual machine size type constants, and the old names were removed.

i.e.

```
compute.StandardD2sV3
```

now becomes:

```
compute.VirtualMachineSizeTypesStandardD2sV3
```

Which broke my build - which validates that `you should rely on a specific API version` is just untrue with our current code generators, PR process, and lack of testing.

flyinprogrammer:
this isn't the first time either that these generators have completely broken things, and our lack of testing has caused breaking releases: https://github.com/Azure/azure-sdk-for-go/issues/2556

vladbarosan:
@flyinprogrammer please read my reply fully. What you linked is from the profile which,  like I mentioned, is expected to change  [as the readme specifies](https://github.com/Azure/azure-sdk-for-go#profiles). 

Also as mentioned there: 

> The latest and preview profiles can help you stay up to date with API updates as you build applications. Since they are by definition not stable, however, they should not be used in production apps. Instead, choose the latest specific API version (or an older one if necessary) from the services/ path.

Also please read my previous reply about how to go about the SDK versions and the Service API versions to avoid breaking changes.

If here is anything unclear left please reopen the issue.

flyinprogrammer:
```
import "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
```

So this import is not considered to be versioned under a `Service API version` of `2018-06-01` ?

flyinprogrammer:
https://github.com/Azure/azure-sdk-for-go/releases/tag/v20.0.0

you guys noted it as a breaking change in the change log, my apologies. i'll work on a kind PR to the readme.

vladbarosan:
[...]

PR #3122, opened 2018-10-24
docs: adding more clarity to backwards compatibility

flyinprogrammer:
Hoping that this PR ensures something like this: https://github.com/Azure/azure-sdk-for-go/issues/3111 doesn't happen again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,8 @@ import (
 ```
 
 Occasionally service-side changes require major changes to existing versions.
-These cases are noted in the changelog.
+These cases are noted in the changelog, and for this reason `Service API versions`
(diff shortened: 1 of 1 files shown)


vladbarosan:
@joshgav are you ok with these changes?

README.md
```diff
@@ -294,7 +294,8 @@ import (
 ```
 
 Occasionally service-side changes require major changes to existing versions.
-These cases are noted in the changelog.
+These cases are noted in the changelog, and for this reason cannot be used
```
joshgav: for this reason <what> cannot be used alone - "version numbers" perhaps?

flyinprogrammer:
thanks for question - hopefully being explicit helps too?

PR #3130, opened 2018-10-25
regenerate whole sdk with latest autorest and update dep

vladbarosan:
- update dependencies
- regenerate whole sdk ( specs commit: 884d9027a69334049b2d8a2beacbb8cbc8f7612c)
- regenerate profiles
- update swagger to sdk autorest version


vladbarosan:
@jhendrixMSFT please check:
- only deprecated packages have been removed ( note that deprecated packages from last release are still here)
- that the packages that have been deprecated indeed don't exist or should have been(are) in the preview folder

PR #3423, opened 2018-11-27
Fix packages command for apidiff tool

jhendrixMSFT:
The addition of the *api subdirectories broke some assumptions the tool
made on how to find packages.  Update the logic to not recurse into the
api subdirectories.

### DIFF
--- a/tools/apidiff/cmd/packages.go
+++ b/tools/apidiff/cmd/packages.go
@@ -26,6 +26,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const apiDirSuffix = "api"
+
 var packagesCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


tools/apidiff/cmd/packages.go
```diff
@@ -119,7 +119,8 @@ func getRepoContentForCommit(wt repo.WorkingTree, dir, commit string) (r repoCon
 			}
 			hasSubDirs := false
 			for _, f := range fi {
-				if f.IsDir() {
+				// check if this is the interfaces subdir, if it is don't recurse into it
+				if f.IsDir() && f.Name() != filepath.Base(path)+"api" {
```
vladbarosan: nit: maybe have the "api" as a constant for interfaces directory

PR #3424, opened 2018-11-27
v23.0.0

jhendrixMSFT:


README.md
```diff
@@ -15,6 +15,12 @@ feed](https://azure.microsoft.com/updates/).
 Users may prefer to jump right in to our samples repo at
 [github.com/Azure-Samples/azure-sdk-for-go-samples][samples_repo].
 
+Questions and feedback? Chat with us in the **[#Azure SDK
+channel](https://gophers.slack.com/messages/CA7HK8EEP)** on the [Gophers
+Slack](https://gophers.slack.com/). Sign up
+[here](https://invite.slack.golangbridge.org) first if necessary.
```
vladbarosan: only thing is i would have added some more info in the readme about using the interfaces for mocking, but i guess its too late ?

services/web/mgmt/2016-09-01/web/webapi/interfaces.go
```diff
@@ -0,0 +1,603 @@
+package webapi
```
vladbarosan: just to check is this the intended name for the package ? I thought it would just be `api`

jhendrixMSFT: This is intentional so that all the `api` package names don't collide.

README.md
jhendrixMSFT: I will add a sample for this.

PR #3440, opened 2018-11-29
Refactor profileBuilder tool

jhendrixMSFT:
Simplified the code and in the process uncovered several bugs where an
operation would fail but the error was ignored.
Removed dependency on GOPATH.
For the `list` command changed the input format to JSON; this is
required for upcoming changes to support hybrid profiles.
Removed dependency on viper as it wasn't used and cluttered up the code.
The content in aliasPackage.go is a copy from an external dependency so
the core logic is the same but simplified.
Removed all default values and made their flags required.
Properly support packages that are versioned like `v1`.
The sort order for aliased content has been fixed once and for all.

Update profiles with latest profileBuilder tool
    
Converted test list files to JSON.
Added root generate.go file and swagger->sdk to use it.
Some missing profile content has been added due to bug fixes.
Sort order fixes has rearranged some content.


tools/profileBuilder/Gopkg.toml
```diff
@@ -1,50 +1,7 @@
...
-  branch = "master"
-  name = "github.com/marstr/randname"
-
-[[constraint]]
-  branch = "master"
-  name = "github.com/mitchellh/go-homedir"
-
 [[constraint]]
   name = "github.com/spf13/cobra"
   version = "0.0.1"
```
vladbarosan: >"0.0.1 [](start = 12, length = 6)

there is 0.0.3 if we care to update. kind of sucks the guy doesnt provide any release notes.

jhendrixMSFT: Default behavior is this version is treated as `^`, so it's the minimum required version.  So when you do `dep ensure` for the first time or `dep ensure -update` the tool will find the latest version that respects `^`.  If you look in the lock file you'll see that we're using 0.0.3.

tools/profileBuilder/cmd/root.go
```diff
@@ -72,42 +56,10 @@ func Execute() {
...
-	}
-
-	viper.AutomaticEnv() // read in environment variables that match
-
-	// If a config file is found, read it in.
-	if err := viper.ReadInConfig(); err == nil {
-		fmt.Println("Using config file:", viper.ConfigFileUsed())
-	}
+	rootCmd.PersistentFlags().BoolVarP(&clearOutputFlag, "clear-output", "c", false, "Removes any directories in the output-folder before writing a profile.")
+	rootCmd.PersistentFlags().BoolVarP(&verboseFlag, "verbose", "v", false, "Use stderr to log verbose output.")
```
vladbarosan: >Flag [](start = 44, length = 4)

nit: adding `Flag`  to the var name seems unnecessary

tools/profileBuilder/cmd/list.go
```diff
@@ -54,65 +49,50 @@ $> ../model/testdata/smallProfile.txt > profileBuilder list --name small_profile
...
+		var listDef model.ListDefinition
+		err = json.Unmarshal(data, &listDef)
+		if err != nil {
+			errLog.Fatalf("failed to unmarshal JSON: %v", err)
+		}
+
+		if clearOutputFlag {
+			if err := model.DeleteChildDirs(outputRootDir); err != nil {
+				errLog.Fatalf("Unable to clear output-folder: %v", err)
 			}
```
vladbarosan: should we fail in this case ?

jhendrixMSFT: Call to `Fatalf` will log the message then call `os.Exit(-1)`, or are you saying this shouldn't be considered a fatal error?

vladbarosan: well so yea, was asking if that is what we want, to just fail or warn and continue

---
In reply to: [238000977](https://github.com/Azure/azure-sdk-for-go/pull/3440#discussion_r238000977) [](ancestors = 238000977)

vladbarosan: hmm I guess if the user put that flag, I guess its fair enough to fail since we are not doing what he requested 

---
In reply to: [238003925](https://github.com/Azure/azure-sdk-for-go/pull/3440#discussion_r238003925) [](ancestors = 238003925,238000977)

jhendrixMSFT: The way it works right now it would be safe to warn and continue since we only write the one file with the same name (well, that assumes that we open the file for truncate).  However this could be problematic in the future if we ever emit more than one file and/or the file name can change.

tools/profileBuilder/model/transforms.go
```diff
@@ -35,155 +35,127 @@ import (
...
+			if err != nil {
+				errLog.Fatalf("failed to convert to absolute path: %v", err)
 			}
-			files := token.NewFileSet()
-			parsed, err := parser.ParseDir(files, cast, nil, 0)
+			pkgDir = abs
+		}
+		go func(pd string) {
+			fs := token.NewFileSet()
+			packages, err := parser.ParseDir(fs, pd, func(f os.FileInfo) bool {
```
vladbarosan: hmm can we extract the goroutine as a seperate function ?

tools/profileBuilder/model/transforms.go
```diff
@@ -35,155 +35,127 @@ import (
...
-		}
+			if len(packages) > 1 {
+				errLog.Fatalf("found more than one package in '%s'", pd)
+			}
+			for pn := range packages {
+				p := packages[pn]
+				// trim any non-exported nodes
+				if exp := ast.PackageExports(p); !exp {
+					errLog.Fatalf("package '%s' doesn't contain any exports", pn)
+				}
```
vladbarosan: i guess we fail here because it should never happen ?

tools/profileBuilder/model/transforms.go
```diff
@@ -35,155 +35,127 @@ import (
...
+				// construct the import path from the outputLocation
+				// e.g. D:\work\src\github.com\Azure\azure-sdk-for-go\profiles\2017-03-09\compute\mgmt\compute
+				// becomes github.com/Azure/azure-sdk-for-go/profiles/2017-03-09/compute/mgmt/compute
+				i := strings.Index(pd, "github.com")
+				if i == -1 {
+					errLog.Fatalf("didn't find 'github.com' in '%s'", pd)
+				}
```
vladbarosan: nit:  I wonder if we want to parametrize this in any way ( for testing for example ) and have a default github.com/Azure/azure-sdk-for-go

tools/profileBuilder/model/transforms.go
jhendrixMSFT: Correct.

tools/profileBuilder/model/transforms.go
jhendrixMSFT: To be honest I'm not super happy with how this works as it makes the tool less general-purpose.  Originally I was thinking about adding the import path explicitly in the JSON to avoid this magic however for the SDK the information is redundant which also bothered me.  :/

tools/profileBuilder/model/transforms.go
jhendrixMSFT: We can however I did it this way since nobody else needs to call this function (it's an implementation detail of `BuildProfile()`).  If you feel strongly about it I will move it.

tools/profileBuilder/model/aliasPackage.go
```diff
@@ -0,0 +1,322 @@
...
+func (alias AliasPackage) ModelFile() *ast.File {
+	if alias.Files != nil {
+		return alias.Files[modelFile]
+	}
+	return nil
+}
+
+// NewAliasPackage creates an alias package from the specified input package.
+// Parameter importPath is the import path specified to consume the package.
+func NewAliasPackage(original *ast.Package, importPath string) (*AliasPackage, error) {
```
vladbarosan: >NewAliasPackage [](start = 5, length = 15)

are the functions here testable, only thing would be to maybe add a test

tools/profileBuilder/model/transforms.go
vladbarosan: oh was mostly just for better readability

tools/profileBuilder/model/transforms.go
vladbarosan: well not super important but, would provide a way to override for testing or for working outside of the gopath. Actually to think about it, we might need to change it a little more to make sure that works I guess. at least lets open an issue for this.

PR #3730, opened 2018-12-14
Fix latest profile

jhendrixMSFT:
Update profile builder `latest` command to respect the preview
subdirectory when omitting preview content.
Switched `latest` profile to use the `latest` command instead of a fixed
list that was outdated.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/1066

tools/profileBuilder/model/latest.go
```diff
@@ -31,11 +31,16 @@ func acceptAll(name string) bool {
...
+func ignorePreview(name string) bool {
+	// check if the path contains a /preview/ subdirectory
+	if strings.Contains(name, previewSubdir) {
+		return false
+	}
 	matches := packageName.FindStringSubmatch(name)
 	version := matches[3]
-	result = !strings.Contains(version, "-preview") && !strings.Contains(version, "-beta") // matches[2] is the `version` group
-	return
+	return !strings.Contains(version, "-preview") && !strings.Contains(version, "-beta") // matches[2] is the `version` group
```
vladbarosan: ok just to be sure, the policy is. if folder includes preview OR if its a preview api version we wont included. can you check please that the readme specifies this also for the latest profile

tools/profileBuilder/model/latest.go
```diff
@@ -31,11 +31,16 @@ func acceptAll(name string) bool {
 	return true
 }
 
-func ignorePreview(name string) (result bool) {
+var previewSubdir = fmt.Sprintf("%spreview%s", string(os.PathSeparator), string(os.PathSeparator))
+
+func ignorePreview(name string) bool {
```
vladbarosan: hmm not sure if there is a better name, but ignorePreview doesnt really tell me if it should return true or false if it is preview ( maybe `isPreview` )?

tools/profileBuilder/model/latest.go
jhendrixMSFT: The readme doesn't really explain the difference between the `latest` and `preview` profiles, will update.

jhendrixMSFT:
I can confirm the content is deleted as there was a few preview APIs versions under latest that are now removed in this PR (they were in the list by mistake).

tools/profileBuilder/model/latest.go
jhendrixMSFT: Agreed the naming is confusing, will add a comment a clean up the names.

PR #3785, opened 2018-12-31
Support for cloud services

nshampur:
Several large enterprises still extensively use Azure PaaS V1 (Cloud Services a.k.a. Web/Worker Roles) through the classic management API (ASM). These are early adopters of the Azure cloud platform and will take singificant migration effort to move to newer Compute options available via Azure Resource Manager (ARM). 
Newer cross platform automation options (azure cli, Az Powershell Module) do not provide support for ASM, which makes it non feasible to run on non windows environment (e.g:- Linux Containers as part of CI/CD). 
This pull request wraps the ASM API and adds support for the common workflows such as deploying cloud service packages to production or staging slots of a cloud service and performing blue/green style deployment with VIP Swap. It also adds support for managing extensions for cloud services which are needed for enabling extension such as PaaS Diagnostics (a.k.a. WAD a.k.a. Windows Azure Diagnostics) and Microsoft AntiMalware extensions.


jhendrixMSFT:
@nshampur thanks for this PR, can you please target this against the `latest` branch instead?

jhendrixMSFT:
Thanks will follow up on that PR.

PR #3818, opened 2019-01-07
Support asm cloud services (Targeting latest branch)

nshampur:
Several large enterprises still extensively use Azure PaaS V1 (Cloud Services a.k.a. Web/Worker Roles) through the classic management API (ASM). These are early adopters of the Azure cloud platform and will take singificant migration effort to move to newer Compute options available via Azure Resource Manager (ARM).
Newer cross platform automation options (azure cli, Az Powershell Module) do not provide support for ASM, which makes it non feasible to run on non windows environment (e.g:- Linux Containers as part of CI/CD).
This pull request wraps the ASM API and adds support for the common workflows such as deploying cloud service packages to production or staging slots of a cloud service and performing blue/green style deployment with VIP Swap. It also adds support for managing extensions for cloud services which are needed for enabling extension such as PaaS Diagnostics (a.k.a. WAD a.k.a. Windows Azure Diagnostics) and Microsoft AntiMalware extensions.

Targeting latest branch

### DIFF
--- a/services/classic/management/virtualmachine/client.go
+++ b/services/classic/management/virtualmachine/client.go
@@ -25,14 +25,18 @@ import (
 )
 
 const (
-	azureDeploymentListURL        = "services/hostedservices/%s/deployments"
-	azureDeploymentURL            = "services/hostedservices/%s/deployments/%s"
-	azureListDeploymentsInSlotURL = "services/hostedservices/%s/deploymentslots/Production"
-	deleteAzureDeploymentURL      = "services/hostedservices/%s/deployments/%s?comp=media"
-	azureAddRoleURL               = "services/hostedservices/%s/deployments/%s/roles"
-	azureRoleURL                  = "services/hostedservices/%s/deployments/%s/roles/%s"
-	azureOperationsURL            = "services/hostedservices/%s/deployments/%s/roleinstances/%s/Operations"
-	azureRoleSizeListURL          = "rolesizes"
+	azureDeploymentListURL                    = "services/hostedservices/%s/deployments"
+	azureDeploymentURL                        = "services/hostedservices/%s/deployments/%s"
+	azureUpdateDeploymentURL                  = "services/hostedservices/%s/deployments/%s?comp=%s"
+	azureDeploymentSlotSwapURL                = "services/hostedservices/%s"
+	azureDeploymentSlotURL                    = "services/hostedservices/%s/deploymentslots/%s"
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Sadly we don't have tests for these packages.  Were you able to locally test your additions?

services/classic/management/virtualmachine/client.go
```diff
@@ -89,7 +311,33 @@ func (vm VirtualMachineClient) GetDeploymentName(cloudServiceName string) (strin
 	if cloudServiceName == "" {
 		return "", fmt.Errorf(errParamNotSpecified, "cloudServiceName")
 	}
-	requestURL := fmt.Sprintf(azureListDeploymentsInSlotURL, cloudServiceName)
+	requestURL := fmt.Sprintf(azureDeploymentSlotURL, cloudServiceName, "Production")
```
jhendrixMSFT: Shouldn't the third param be `DeploymentSlotProduction`?

jhendrixMSFT: Or perhaps `GetDeploymentName()` calls `GetDeploymentNameForSlot()`?

services/classic/management/virtualmachine/entities.go
```diff
@@ -71,7 +71,7 @@ type DeploymentResponse struct {
 	ExtendedProperties     []ExtendedProperty `xml:">ExtendedProperty"`
 	PersistentVMDowntime   PersistentVMDowntime
 	VirtualIPs             []VirtualIP `xml:">VirtualIP"`
-	ExtensionConfiguration string      // cloud service extensions not fully implemented
+	ExtensionConfiguration ExtensionConfiguration
```
jhendrixMSFT: This is a breaking change so we'll have to release this as part of v25 at the end of this month.

nshampur:
@jhendrixMSFT .. Yes.  Tested it fairly well and even integrated into our CI/CD pipeline

services/classic/management/virtualmachine/entities.go
nshampur: @jhendrixMSFT . Sounds good if we need to wait for v25, even though the property would always be null previously.

services/classic/management/virtualmachine/client.go
nshampur: good suggestions. I'll make the changes

jhendrixMSFT:
@nshampur thanks for your contribution.

PR #3858, opened 2019-01-11
Reuse http client used with ASM API 

nshampur:
This PR addresses re-using the underlying httpClient when making API calls to Azure Service Management (ASM). 

It also adds some throttling when ASM returns a ServiceUnavailable response (Code: TooManyRequests, Message: Too many requests received. Retry after some time.)

### DIFF
--- a/services/classic/management/http.go
+++ b/services/classic/management/http.go
@@ -107,7 +107,7 @@ func (client client) createHTTPClient() (*http.Client, error) {
 		Transport: &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
-				Renegotiation: tls.RenegotiateFreelyAsClient,
+				Renegotiation: tls.RenegotiateOnceAsClient,
 				Certificates:  []tls.Certificate{cert},
 			},
 		},
--- a/services/classic/management/http.go
+++ b/services/classic/management/http.go
@@ -87,12 +87,7 @@ func (client client) sendAzureRequest(method, url, contentType string, data []by
 		return nil, fmt.Errorf(errParamNotSpecified, "url")
 	}
 
-	httpClient, err := client.createHTTPClient()
-	if err != nil {
-		return nil, err
-	}
-
-	response, err := client.sendRequest(httpClient, url, method, contentType, data, 5)
+	response, err := client.sendRequest(client.httpClient, url, method, contentType, data, 5)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


nshampur:
per Go docs, this appears to be the recommended pattern

https://golang.org/pkg/net/http/

The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines

services/classic/management/http.go
```diff
@@ -100,22 +101,26 @@ func (client client) sendAzureRequest(method, url, contentType string, data []by
...
 // createHTTPClient creates an HTTP Client configured with the key pair for
-// the subscription for this client.
+// the subscription for this client. If a client already exists, then it returns
+// the instance
 func (client client) createHTTPClient() (*http.Client, error) {
-	cert, err := tls.X509KeyPair(client.publishSettings.SubscriptionCert, client.publishSettings.SubscriptionKey)
-	if err != nil {
-		return nil, err
-	}
+	if client.httpClient == nil {
```
jhendrixMSFT: Looks like this is only called from `sendAzureRequest()`.  Since you're now calling it in `makeClient()` I think you could remove the call from `sendAzureRequest()` and eliminate this nil check (also the receiver is passed by value so modifying it here doesn't have any effect).

services/classic/management/http.go
```diff
@@ -161,6 +157,10 @@ func (client client) sendRequest(httpClient *http.Client, url, requestType, cont
 				if numberOfRetries == 0 {
 					return nil, azureErr
 				}
+				if response.StatusCode == http.StatusServiceUnavailable {
+					// Wait before retrying the operation
+					time.Sleep(30 * time.Second)
```
jhendrixMSFT: How about using `client.config.OperationPollInterval` instead of a fixed value?  It defaults to 30 seconds and it user-configurable.

services/classic/management/http.go
```diff
@@ -161,6 +157,10 @@ func (client client) sendRequest(httpClient *http.Client, url, requestType, cont
 				if numberOfRetries == 0 {
 					return nil, azureErr
 				}
+				if response.StatusCode == http.StatusServiceUnavailable {
```
jhendrixMSFT: I'm not super-familiar with ASM but for ARM we have a list of standard HTTP status codes to automatically retry on, you can find it [here](https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L48).  Up to you if you want to expand this or not.

services/classic/management/http.go
nshampur: Sounds good.  I missed that nuance.  Will add.

services/classic/management/http.go
nshampur: ASM has some funny behavior. 
I initially went with 429 (StatusTooManyRequests) as one would expect from a throttling message. But ASM does not seem to return that in this scenario.

services/classic/management/http.go
```diff
@@ -111,7 +107,7 @@ func (client client) createHTTPClient() (*http.Client, error) {
 		Transport: &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
-				Renegotiation: tls.RenegotiateOnceAsClient,
+				Renegotiation: tls.RenegotiateFreelyAsClient,
```
jhendrixMSFT: Curious about this change and if it's really necessary?  Perhaps it would be better to make it user-configurable, i.e. add it to `ClientConfig` and leave the default as `RenegotiateOnceAsClient`.

nshampur: This was an oversight. I'd reverted in my commit 092dc5b a short while back.

PR #3863, opened 2019-01-13
Read deploymentID property in GetDeployment API responses

nshampur:
Certain workflows rely on the deploymentID (a.k.a. PrivateID) from the Deployment response. 

https://docs.microsoft.com/en-us/rest/api/compute/cloudservices/rest-get-deployment
<PrivateID>identifier-of-deployment</PrivateID>  

This was missing in the entity definition


nshampur:
PR to the Wrong branch. Will resubmit

PR #3951, opened 2019-01-28
update dependencies

jhendrixMSFT:


README.md
```diff
@@ -7,7 +7,9 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/Azure/azure-sdk-for-go)](https://goreportcard.com/report/github.com/Azure/azure-sdk-for-go)
 
 azure-sdk-for-go provides Go packages for managing and using Azure services.
-It is continuously tested with Go 1.8, 1.9, 1.10, 1.11 and master.
+It officially supports the last two major releases of Go.  Older versions of
```
vladbarosan: i wonder if we should specify when we will move to supporting a new released go version , otherwise LGTM

PR #3965, opened 2019-01-29
v25.0.0

jhendrixMSFT:


profiles/2017-03-09/definition.json
```diff
@@ -16,7 +16,7 @@
         "../../services/resources/mgmt/2016-09-01/links/linksapi",
         "../../services/resources/mgmt/2015-01-01/locks",
         "../../services/resources/mgmt/2015-01-01/locks/locksapi",
-        "../../services/resources/mgmt/2015-10-01-preview/policy",
+        "../../services/preview/resources/mgmt/2015-10-01-preview/policy",
         "../../services/resources/mgmt/2016-02-01/resources",
```
vladbarosan: hmm so the profile contains a preview api ?

jhendrixMSFT: Sadly yes.

jhendrixMSFT: Will fix per discussion to remove `preview` directory.

PR #4254, opened 2019-03-11
Check in vendored dependencies

jhendrixMSFT:
Running `dep ensure` in CI has been a point of failure causing CI builds
to be unreliable.  Enable pruning of dependencies and check them in.
Removed all dep-related commands from travis.
Removed the vendor directory from .gitignore so changes are visible.


aviaoh:
This PR breaks latest `azure-sdk-for-go`. we are now getting the following error:
> cannot use env (type "github.com/Azure/go-autorest/autorest/azure".Environment) as type "github.com/Azure/azure-sdk-for-go/vendor/github.com/Azure/go-autorest/autorest/azure".Environment in argument to storage.NewAccountSASClient

PR #4440, opened 2019-03-29
v27.0.0

jhendrixMSFT:


tools/profileBuilder/Gopkg.toml
```diff
@@ -1,7 +0,0 @@
-[[constraint]]
```
vladbarosan: no need anymore ?

jhendrixMSFT: No, it was a subset of the root Gopkg.toml.

PR #4514, opened 2019-04-10
chore: Add gitattributes for line endings

nschonni:
Ran `git add --renormalize .`


jhendrixMSFT:
@nschonni can you retarget this to the `latest` branch?  It will then flow into `master`.

nschonni:
@jhendrixMSFT done. I does seem like "latest" is behind in some ways from master, so the CODEOWNERS fixup is gone now

PR #4588, opened 2019-04-19
fix file permissions when generating profiles

jhendrixMSFT:


jhendrixMSFT:
File permissions were incorrect when building on macOS.  @vladbarosan could you try this on Windows to ensure it works?  I don't have access to a Windows machine at present.

PR #4689, opened 2019-04-28
Use correct uuid.NewV1() API

orlangure:
This PR fixes invalid usage of `uuid.NewV1()`.
This method returns a value and an error, and it was used inline, like it returns a single value.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/4690

### DIFF
--- a/storage/table_batch.go
+++ b/storage/table_batch.go
@@ -131,15 +131,22 @@ func (t *TableBatch) MergeEntity(entity *Entity) {
 // the changesets.
 // As per document https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/performing-entity-group-transactions
 func (t *TableBatch) ExecuteBatch() error {
-
-	changesetBoundary := fmt.Sprintf("changeset_%s", uuid.NewV1().String())
+	uuidV1, err := uuid.NewV1()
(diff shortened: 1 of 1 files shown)


orlangure:
The builds using `azure-sdk-for-go` fail with the following message:
```
# github.com/Azure/azure-sdk-for-go/storage
../../Azure/azure-sdk-for-go/storage/table_batch.go:135:61: multiple-value uuid.NewV1() in single-value context
../../Azure/azure-sdk-for-go/storage/table_batch.go:142:48: multiple-value uuid.NewV1() in single-value context
```

aviaoh:
this change was made by auto generated code. see https://github.com/Azure/azure-sdk-for-go/issues/4690.
after fixing it, can you please fix codegen to include this new code?
thanks :)

jhendrixMSFT:
The problem here is that HEAD and tag v1.2.0 (the version we reference) have different signatures for `uuid.NewV1()`.  If you're not using `dep` to manage dependencies can you please pin your module dependency to v1.2.0 to `uuid`?

jhendrixMSFT:
Closed in favor of https://github.com/Azure/azure-sdk-for-go/pull/4699

PR #4854, opened 2019-05-22
fix a problem that token generated from storage explorer, without spr in query, is using http

jeffguorg:
the go sdk checks for [spr parameter](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#parameters-common-to-account-sas-and-service-sas-tokens), but when it comes to tokens generated from Storage Explorer, there is no such spr parameter, and the sdk get empty string for spr. so after compares empty string and "https", it [sets useHTTPS to false](https://github.com/Azure/azure-sdk-for-go/blob/master/storage/client.go#L374), while it actually needs to set it to true

this patch just altered that and add a test case. the test account has been deleted to prevent malformed requests but the test case should not be influenced

it's my first PR here. tell me if there is something need to be done and i'll do it.


jhendrixMSFT:
Hello!  This PR is merging master into latest which is dragging in a bunch of extra changes.  Can you please create a new PR from your fork of latest (or a branch off of latest) into upstream/latest?

jeffguorg:
Hi, i just opened another [PR](https://github.com/Azure/azure-sdk-for-go/pull/4870) and I'm closing this one. sorry for my mistake

PR #4886, opened 2019-05-28
Update default storage API version to 2018-03-28

jhendrixMSFT:
Updated string-to-sign algorithm for SAS tokens to support 2018-11-09.


jhendrixMSFT:
@tombuildsstuff @kt how hard would it be to ingest this into Terraform to see if anything breaks?  My hope is to add some of the missing functionality in this API version to unblock Terraform scenarios.

tombuildsstuff:
@jhendrixMSFT oh cool, thanks! To get the tests running is pretty easy - I’ll try and kick off a test run today. That said it may be worth holding off adding support for the new functionality for the moment.

I started looking into generating our own Data Plane SDK for Storage, but couldn’t find the Swagger files to generate from - so instead I ended up prototyping our own using the standard go-discover libraries (eg including Context). Whilst we’re just prototyping it (and adding tests) to see if that’s viable at this stage, I’m wondering if that approach (and possibly upstreaming it) would be better longer-term than trying to determine what features are missing from the old SDK - WDYT?

Thanks!

jhendrixMSFT:
So there is a swagger for this floating about internally but has not been made public for reasons I don't understand.  I plan on following up on that but I'm not super-optimistic about it.  In addition, we never back-ported XML support to the v2 code generator so even with a swagger we would still be blocked until this happened.
I forgot about lack of context in the current storage package, that's a bummer and splicing this in might be clunky.  Really my hope was that I could spend half a day to implement some low-hanging fruit to get you unblocked, e.g. static web sites or whatever APIs you needed.

tombuildsstuff:
@vladbarosan @jhendrixMSFT tests look good to me:

Screenshot 2019-05-29 at 17 17 05

PR #5413, opened 2019-08-01
Fix table query continuation token

djyou:
Azure Storage table query may return results with non-empty `x-ms-continuation-NextPartitionKey` but empty `x-ms-continuation-NextRowKey`, see https://docs.microsoft.com/en-us/rest/api/storageservices/query-timeout-and-pagination

The current Go SDK implementation couldn't follow the continuation token for these cases.

Reference to .NET implementation https://github.com/Azure/azure-storage-net/blob/fdc07726910ead3281b14b2769a92b2fea68bd97/Lib/Common/Table/TableContinuationToken.cs

### DIFF
--- a/storage/table.go
+++ b/storage/table.go
@@ -355,8 +355,12 @@ func (t *Table) queryEntities(uri string, headers map[string]string, ml Metadata
 			return nil, err
 		}
 		v := originalURI.Query()
-		v.Set(nextPartitionKeyQueryParameter, contToken.NextPartitionKey)
-		v.Set(nextRowKeyQueryParameter, contToken.NextRowKey)
+		if contToken.NextPartitionKey != "" {
+			v.Set(nextPartitionKeyQueryParameter, contToken.NextPartitionKey)
+		}
+		if contToken.NextRowKey != "" {
+			v.Set(nextRowKeyQueryParameter, contToken.NextRowKey)
+		}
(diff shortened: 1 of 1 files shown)


storage/table.go
```diff
@@ -371,7 +375,7 @@ func extractContinuationTokenFromHeaders(h http.Header) *continuationToken {
 		NextRowKey:       h.Get(headerNextRowKey),
 	}
 
-	if ct.NextPartitionKey != "" && ct.NextRowKey != "" {
+	if ct.NextPartitionKey != "" || ct.NextRowKey != "" {
```
jhendrixMSFT: Can it ever be the case that `NextRowKey` is not empty but `NextParitionKey` is?

jhendrixMSFT:
Can you please retarget this PR against the `latest` branch?

storage/table.go
djyou: Storage docs didn't specifically mention if `NextParitionKey` could be empty, but it's a valid query to storage to have just `NextRowKey`.

Also from how other SDKs handle it, these three pagination headers `NextTableName`, `NextPartitionKey`, and `NextRowKey` don't have dependencies on each other. The non-empty ones will be added to the URL of the next request.

djyou:
Retarget to `latest` branch with this PR https://github.com/Azure/azure-sdk-for-go/pull/5423
Closing this one.

PR #5415, opened 2019-08-02
v32.1.0

ArcturusZhang:
This minor release releases a new package qnamakerruntime v4.0


ArcturusZhang:
> The deleted profile files looks wrong, can you please take a look?

Now they come back.

PR #5464, opened 2019-08-08
Prevent apitool from producing empty tables

ArcturusZhang:
Prevent apitool from creating empty tables when there are only breaking changes but no additional changes

### DIFF
--- a/tools/apidiff/cmd/changelog.go
+++ b/tools/apidiff/cmd/changelog.go
@@ -82,28 +82,32 @@ func reportUpdatedPkgs(pr pkgsReport) {
 	if !pr.modPkgHasAdditions {
 		return
 	}
-	fmt.Printf("### Updated Packages\n\n")
 	updated := []string{}
 	for pkgName, pkgRpt := range pr.ModifiedPackages {
 		if pkgRpt.HasAdditiveChanges() && !pkgRpt.HasBreakingChanges() {
 			updated = append(updated, pkgName)
 		}
 	}
-	createTable(createTableRows(updated))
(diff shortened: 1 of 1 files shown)


tools/apidiff/cmd/changelog.go
```diff
@@ -82,28 +82,32 @@ func reportUpdatedPkgs(pr pkgsReport) {
...
 	}
-	fmt.Printf("### Updated Packages\n\n")
 	updated := []string{}
 	for pkgName, pkgRpt := range pr.ModifiedPackages {
 		if pkgRpt.HasAdditiveChanges() && !pkgRpt.HasBreakingChanges() {
 			updated = append(updated, pkgName)
 		}
 	}
-	createTable(createTableRows(updated))
+	if len(updated) > 0 {
```
jhendrixMSFT: This will work but it shouldn't be necessary.  On line 82 `pr.modPkgHasAdditions` should be `false` so the function exits early.  I think the bug is that it's getting set to `true` when there are no additionsl.

ArcturusZhang: In the case I tested (base commit: v32.2.0, target commit: v32.3.0), `modPkgHasAddtions` is true, so is `pkgRpt.HasAdditiveChanges()` and `pkgRpt.HasBreakingChanges()`, therefore this part of code prints an empty table of updated packages.
Actually, breaking changes here is some functions renaming, which could also be recognized as deleting and adding some function. This is also the reason initially I manually add an entry to the updated package table. It does updated and happens to be a breaking change, and I suppose these two characters can co-exist.

PR #5663, opened 2019-08-30
feature: support PutBlockFromURL API:

swordqiu:
Support PutBlockFromURL API: https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-from-url

- [x] The PR targets the `latest` branch.
- [-] Updates to [CHANGELOG.md][] are included.
- [-] Apache v2 license headers are included in each file.
 
### DIFF
--- a/storage/blockblob.go
+++ b/storage/blockblob.go
@@ -197,6 +197,47 @@ func (b *Blob) PutBlockWithLength(blockID string, size uint64, blob io.Reader, o
 	return b.respondCreation(resp, BlobTypeBlock)
 }
 
+// PutBlockFromURLOptions includes the options for a put block from URL operation
+type PutBlockFromURLOptions struct {
+	PutBlockOptions
+
+	SourceContentMD5   string `header:"x-ms-source-content-md5"`
+	SourceContentCRC64 string `header:"x-ms-source-content-crc64"`
+}
+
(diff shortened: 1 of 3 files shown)


storage/blockblob.go
```diff
@@ -197,6 +197,47 @@ func (b *Blob) PutBlockWithLength(blockID string, size uint64, blob io.Reader, o
 	return b.respondCreation(resp, BlobTypeBlock)
 }
 
+// PutBlockFromURLOptions includes the options for a put block from URL operation
+type PutBlockFromURLOptions struct {
+	PutBlockOptions
+
+	SourceContentMD5   string `header:"x-ms-source-content-md5"`
+	SourceContentCRC64 string `header:"x-ms-source-content-crc64"`
```
tombuildsstuff: it might be worth a note that [this is only supported in API version `2019-02-02`](https://docs.microsoft.com/en-gb/rest/api/storageservices/versioning-for-the-azure-storage-services#version-2019-02-02), rather than in the default API version the SDK ships with?

jhendrixMSFT: Agreed.  In fact we might want to omit this value for earlier API versions unless we're confident it won't cause any issues, i.e. will earlier API versions reject the request due to this value?  It will be used in the signing algorithm.  @ArcturusZhang can you please follow up on this?

PR #5790, opened 2019-09-17
Update versioner for module implementation

ArcturusZhang:
 

tools/versioner/cmd/root.go
```diff
@@ -113,8 +129,40 @@ func theCommandImpl(args []string) (string, error) {
 	return tag, err
 }
 
+func ensureGoModFileExistence(path string) error {
```
jhendrixMSFT: It's not clear to me why this is necessary as the generator creates this content.  Plus I think we should keep the two separate.

tools/versioner/cmd/root.go
```diff
@@ -77,16 +88,13 @@ func Execute() {
...
-		fmt.Printf("tag: %s\n", tag)
-	}
+	_, err := theCommandImpl(args)
 	return err
 }
 
 // does the actual work
 func theCommandImpl(args []string) (string, error) {
-	stage := filepath.Clean(args[0])
+	stage, _ := filepath.Abs(args[0])
```
jhendrixMSFT: Don't discard potential error.

tools/versioner/cmd/root.go
```diff
@@ -180,6 +235,51 @@ func forInplaceUpdate(lmv, stage string, mod modinfo.Provider) (string, error) {
 	return tag, nil
 }
 
+func updateVersion(path, tag string) error {
```
jhendrixMSFT: Please add a test for this.

tools/versioner/cmd/root.go
ArcturusZhang: Actually, in my local test, the autorest does not generate go.mod file. Therefore I added this function to generate this file

tools/versioner/cmd/root.go
ArcturusZhang: Done

tools/versioner/cmd/root.go
ArcturusZhang: Test added

tools/versioner/cmd/root.go
jhendrixMSFT: You have to specify the `--gomod-root` option with the root location, so `--gomod-root=github.com/Azure/azure-sdk-for-go`

ArcturusZhang: Removed this method

ArcturusZhang:
I made some changes, please have a look again

ArcturusZhang:
Rebased to master

tools/versioner/cmd/root.go
```diff
@@ -128,19 +137,38 @@ func forSideBySideRelease(stage string, mod modinfo.Provider) (string, error) {
...
+	file, err := os.OpenFile(goMod, os.O_RDWR, 0666)
+	if err != nil {
+		return "", fmt.Errorf("failed to open for read '%s': %v", goMod, err)
+	}
+	ver := modinfo.FindVersionSuffix(mod.DestDir())
+	if err = updateGoModVer(file, ver); err != nil {
+		file.Close()
+		return "", fmt.Errorf("failed to update go.mod file: %v", err)
+	}
+	file.Close()
```
ArcturusZhang: Inplaced update also need to update go.mod file, since it may be a update to v2 version.
I suppose the go.mod file generated by autorest will always be as `v1`, am I right? @jhendrixMSFT

jhendrixMSFT: You are correct autorest always generates go.mod as v1 since it has no version information available to it.
Func `forInplaceUpdate()` does not need to modify the go.mod file, see the comment for that func.  `forSideBySideRelease()` is what updates go.mod for v2+

tools/versioner/cmd/root.go
```diff
@@ -196,9 +201,22 @@ func forInplaceUpdate(lmv, stage string, mod modinfo.Provider) (string, error) {
...
 		return "", fmt.Errorf("failed to rename '%s' to '%s': %v", temp, dest, err)
 	}
+	// format code
+	if err = formatCode(mod.DestDir()); err != nil {
+		return "", fmt.Errorf("failed to format code: %v", err)
+	}
 	return tag, nil
 }
 
+func formatCode(path string) error {
```
jhendrixMSFT: Why is this necessary?

ArcturusZhang: Running format in after_script for the whole SDK folder is time consuming, therefore I move the format here, and only run in the modified folder

tools/versioner/cmd/root.go
ArcturusZhang: You mean if there is already a major version of v2, say v2.0.0, and the content in stage folder do not have breaking changes comparing to v2, the tool will use `forInplaceUpdate()`. 
The test on scenarioe will fail if `forInplaceUpdate` do not modify `go.mod` file.

jhendrixMSFT: ScenarioE is an in-place update so the version should end up being v2.1.0.  If the test is now failing then a regression has been introduced someplace.  Or maybe I misunderstand?

tools/versioner/cmd/root.go
jhendrixMSFT: The problem now is that the versioner tool is also responsible for formatting the generated code which IMO is not a good separation of concerns.  Given that this perf problem is not present in swagger-to-sdk-v2, please open a bug to track removing this piece of code once we switch to v2.

tools/versioner/cmd/root.go
ArcturusZhang: Will the inplace update suppose to move every file other than `go.mod` to the old folder? Since autorest always generate `go.mod` as `v1`, I suppose in inplace update, we should not move the `go.mod` in stage folder to the corresponding major version folder. 
I will investigate this more carefully.

jhendrixMSFT: Good point, I don't recall if this case is handled properly; it might not be.  I agree though we shouldn't move go.mod in this scenario so we can just delete it from the stage directory.

tools/versioner/cmd/root.go
ArcturusZhang: I removed this from this tool. And I talked with Phoenix (owner of automation bot), he will change  the behaviour in v2 to make the after script only run once for every commit. Previously the after script will run multiple times if there are multiple tags changed.

ArcturusZhang:
Hi @jhendrixMSFT I made some changes, please have a look.
Since in v2, the after_script will only run once, even if there are multiple tags changed, I made some changes in versioner tool correspondingly. It will now search in a directory for `stage` folders, and run the `unstage` command (same as the previous versioner tool) for each of them.

ArcturusZhang:
There is a problem in CHANGELOG file. In the `New Funcs` section, if the receiver of that func is a pointer, a redundant space will appear. For instance, the item should be 
``` go
1. *receiver.Func() error
```
but actually it is
``` go
1. * receiver.Func() error
```

tools/versioner/cmd/root.go
```diff
@@ -76,243 +61,68 @@ func Execute() {
...
-	} else {
-		// no new exports, this is a patch update
-		n := sv.IncPatch()
-		sv = &n
-	}
-	return strings.Replace(tag, v, "v"+sv.String(), 1), nil
-}
+//
+//// does the actual work
+//func theCommandImpl(args []string) (string, error) {
```
jhendrixMSFT: Why is all this commented out?

tools/versioner/cmd/unstage.go
```diff
@@ -0,0 +1,355 @@
+package cmd
```
jhendrixMSFT: Missing license comment.

tools/versioner/cmd/unstage.go
```diff
@@ -0,0 +1,355 @@
...
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var unstageCmd = &cobra.Command{
```
jhendrixMSFT: While I'm not opposed to moving this content to `unstage.go` the problem is any code changes made here don't show up in the diff.  If you really want to move this content can you please do it in a separate PR that introduces zero code changes?

tools/versioner/cmd/root.go
ArcturusZhang: I moved this piece of code to unstage.go. Comments will be removed.

tools/versioner/cmd/unstage.go
ArcturusZhang: How about I put this command in the same file with rootCmd, therefore the diff will be calculate normally. And after reviewed, they can go in separate files

ArcturusZhang: I move these changes to [another pr](https://github.com/Azure/azure-sdk-for-go/pull/5981)

ArcturusZhang:
Hi @jhendrixMSFT I realized another thing of concern that when a package becomes module with breaking changes, a `v2` module will create for it, whose remain part should play as `v1`. But the problem is there will not be a `go.mod` file for it. So for this case maybe we should have to ensure `go.mod` file existence, or put a `go.mod` file in every package before any pr comes in.

jhendrixMSFT:
I don't understand why there wouldn't be a go.mod file for it, the generator creates one (as v1) and this tool should update it to v2 (there are test scenarios that already cover this).  Can you please clarify?

ArcturusZhang:
> I don't understand why there wouldn't be a go.mod file for it, the generator creates one (as v1) and this tool should update it to v2 (there are test scenarios that already cover this). Can you please clarify?

Let me take this [pr](https://github.com/test-repo-arcturus/azure-sdk-for-go/pull/27) in test scenario as example. This pr introduces breaking changes to package `machinelearningservices/mgmt/2019-06-01/machinelearningservices`, therefore a `v2` of this package should be created and left the original folder as `v1`. Since the original state is not implemented for modules, the original folder does not contains a `go.mod` file.

ArcturusZhang:
> There is a problem in CHANGELOG file. In the `New Funcs` section, if the receiver of that func is a pointer, a redundant space will appear. For instance, the item should be
> 
> ```go
> 1. *receiver.Func() error
> ```
> 
> but actually it is
> 
> ```go
> 1. * receiver.Func() error
> ```

This is resolved in #5981 .
The root cause of this is the stage folder is not formatted.

LorbusChris:
Hi, there hasn't been any movement on this in a while. Is there anything that can be done to speed this up?

PR #5878, opened 2019-09-25
The moduler tool

ArcturusZhang:
This is a new tool used for searching version.go files in the SDK for tags, and push them to github.
I wrote a conceptual pipeline using the moduler tool [here](https://github.com/ArcturusZhang/test-azure-pipeline/blob/master/azure-pipelines.yaml)


jhendrixMSFT:
Can you add some tests?

ArcturusZhang:
Rebased this branch to `refactor-versioner` because the result of this tool based on the result of versioner tool.

ArcturusZhang:
> Can you add some tests?

Tests added, please have a look

tools/moduler/cmd/root.go
```diff
@@ -0,0 +1,238 @@
...
+	vprintln("Pushing new tags")
+	for _, tag := range newTags {
+		if err := addNewTag(tag); err != nil {
+			return err
+		}
+	}
+	return executeCommand("git", "push", "origin", "--tags")
+}
+
+func generateProfiles(profilePath string) error {
```
jhendrixMSFT: This should happen as part of the build pipeline.  Each tool should be responsible for one job, then they can be composed via scripts etc.  See also https://en.wikipedia.org/wiki/Unix_philosophy

ArcturusZhang: Sure, I will move this back to after_script.

PR #6031, opened 2019-10-15
Modify config for modulization

ArcturusZhang:
 
### DIFF
--- a/swagger_to_sdk_config.json
+++ b/swagger_to_sdk_config.json
@@ -1,10 +1,9 @@
 {
   "$schema": "https://openapistorageprod.blob.core.windows.net/sdkautomation/prod/schemas/swagger_to_sdk_config.schema.json",
   "meta": {
-    "after_scripts": [
+    "after_scripts_in_repo": [
       "dep ensure",
-      "go generate ./profiles/generate.go",
-      "gofmt -w ./profiles/",
+      "go run ./tools/versioner/main.go ./services -v",
       "gofmt -w ./services/"
     ],
     "autorest_options": {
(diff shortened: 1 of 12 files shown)


moduler-pipelines.yml
```diff
@@ -0,0 +1,94 @@
+pool:
```
jhendrixMSFT: How does this pipeline get invoked?

swagger_to_sdk_config.json
```diff
@@ -1,10 +1,9 @@
 {
   "$schema": "https://openapistorageprod.blob.core.windows.net/sdkautomation/prod/schemas/swagger_to_sdk_config.schema.json",
   "meta": {
-    "after_scripts": [
+    "after_scripts_in_repo": [
```
jhendrixMSFT: How does `after_scripts_in_repo` change things?

swagger_to_sdk_config.json
```diff
@@ -13,8 +12,10 @@
       "verbose": "",
       "sdkrel:go-sdk-folder": ".",
       "multiapi": "",
-      "use-onever": "",
-      "preview-chk": ""
+      "preview-chk": "",
+      "go.clear-output-folder": false,
+      "stage": "",
```
jhendrixMSFT: So we're ready to turn this on then?

moduler-pipelines.yml
ArcturusZhang: This is a new CI, which only triggers when master branch gets new commits.
We have a testing repo [here](https://github.com/test-repo-arcturus/azure-sdk-for-go) which already gets this pipeline enabled

swagger_to_sdk_config.json
ArcturusZhang: This is per what Phoenix recommends, to emphasize this script runs in the root dir and only run once, even there are multiple changed services

PR #6364, opened 2019-11-16
refactor clients

jhendrixMSFT:
 
### DIFF
--- a/sdk/storage/azblob/service.go
+++ b/sdk/storage/azblob/service.go
@@ -15,32 +15,78 @@ const (
 	scope = "https://storage.azure.com/.default"
 )
 
+// ServiceVersion enumerates the possible service API versions.
+type ServiceVersion string
+
+const (
+	// V2019_02_02 specifies the 2019-02-02 service API version.
+	V2019_02_02 = "2019-02-02"
+)
+
+//ServiceClient provides access to Azure Storage service resources and blob containers.
 type ServiceClient struct {
 	u *url.URL
 	p azcore.Pipeline
+	o ServiceClientOptions
 }
 
-func NewServiceClient(endpoint string, cred azcore.Credential, options azcore.PipelineOptions) (*ServiceClient, error) {
-	p := azcore.NewPipeline(options.HTTPClient,
+// ServiceClientOptions contains client configuration options for connecting to blob storage.
+type ServiceClientOptions struct {
+	azcore.PipelineOptions
+	// Version is the service API version to use when making requests.
+	// Leaving this empty will default to the latest API version.
+	Version ServiceVersion
+
+	// SecondaryEndpoint is the secondary storage endpoint that can be read from if the storage
+	// account is enabled for RA-GRS.
+	// If this value is set, the secondary endpoint will be used for GET or HEAD requests during retries.
+	// If the status of the response from the secondary endpoint is a 404, then subsequent retries for
+	// the request will not use the secondary endpoint again, as this indicates that the resource
+	// may not have propagated there yet. Otherwise, subsequent retries will alternate back and forth
+	// between the primary and secondary endpoint.
+	SecondaryEndpoint string
+}
+
+func setDefaultClientOptions(s *ServiceClientOptions) *ServiceClientOptions {
+	if s == nil {
+		s = &ServiceClientOptions{}
+	}
+	if len(s.Version) == 0 {
+		s.Version = V2019_02_02
+	}
+	return s
+}
+
+// NewServiceClient creates a new instance of ServiceClient with the specified values.
+// Passing nil for cred will default to anonymous access.
+// Pass nil for options to use the default configuration.
+func NewServiceClient(endpoint string, cred azcore.Credential, options *ServiceClientOptions) (*ServiceClient, error) {
+	if cred == nil {
+		cred = azcore.AnonymousCredential()
+	}
+	options = setDefaultClientOptions(options)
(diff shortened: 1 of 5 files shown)


sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
+const (
+	// V2019_02_02 specifies the 2019-02-02 service API version.
+	V2019_02_02 = "2019-02-02"
+)
+
+//ServiceClient provides access to Azure Storage service resources and blob containers.
 type ServiceClient struct {
 	u *url.URL
 	p azcore.Pipeline
+	o ServiceClientOptions
```
JeffreyRichter: I assume this was mostly added due to ServiceVersion? Maybe we should just put that here and let the rest of the options be GC'd.

sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
+	Version ServiceVersion
+
+	// SecondaryEndpoint is the secondary storage endpoint that can be read from if the storage
+	// account is enabled for RA-GRS.
+	// If this value is set, the secondary endpoint will be used for GET or HEAD requests during retries.
+	// If the status of the response from the secondary endpoint is a 404, then subsequent retries for
+	// the request will not use the secondary endpoint again, as this indicates that the resource
+	// may not have propagated there yet. Otherwise, subsequent retries will alternate back and forth
+	// between the primary and secondary endpoint.
+	SecondaryEndpoint string
```
JeffreyRichter: In Track 1.8, this is a retry policy option and I think that is where this belongs; not here. Why was this moved here?

sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
+}
+
+func setDefaultClientOptions(s *ServiceClientOptions) *ServiceClientOptions {
+	if s == nil {
+		s = &ServiceClientOptions{}
+	}
+	if len(s.Version) == 0 {
```
JeffreyRichter: Can't you compare against "" instead - feels more natural to me.

sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
+// Pass nil for options to use the default configuration.
+func NewServiceClient(endpoint string, cred azcore.Credential, options *ServiceClientOptions) (*ServiceClient, error) {
+	if cred == nil {
+		cred = azcore.AnonymousCredential()
+	}
+	options = setDefaultClientOptions(options)
+	p := azcore.NewPipeline(options.HTTPTransport,
 		azcore.NewTelemetryPolicy(options.Telemetry),
 		azcore.NewUniqueRequestIDPolicy(),
 		azcore.NewRetryPolicy(options.Retry),
```
JeffreyRichter: This should be a RetryPolicy that is unique to azblob; not the azcore RetryPolicy.
The azblob policy takes a SecondaryEndpoint in its options

sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
 		cred.AuthenticationPolicy(azcore.AuthenticationPolicyOptions{Scopes: []string{scope}}),
 		azcore.NewRequestLogPolicy(options.LogOptions))
-	return NewServiceClientWithPipeline(endpoint, p)
+	return NewServiceClientWithPipeline(endpoint, p, options)
```
JeffreyRichter: OK, I think I see the issue. lets create a ServiceApiVersion pipeline policy and add it to the pipeline. It is configured with the service api version. Now, we don't need to add options fields to any XxxClient and we don't pass options to any XxxClientWithPipeline functions. 

The options are for configuring the pipeline. It doesn't feel right to have a configured pipeline AND then to also have to pass options to a XxxClientWithPipeline function.

sdk/storage/azblob/service.go
```diff
@@ -15,32 +15,78 @@ const (
...
-	return &ServiceClient{u: u, p: p}, nil
+	options = setDefaultClientOptions(options)
+	return &ServiceClient{u: u, p: pipeline, o: *options}, nil
 }
 
-func (c *ServiceClient) ServiceVersion() string {
-	// this is a method instead of a package-level const to handle the case of composite services
-	return "2018-11-09"
+// ServiceVersion returns the service API version used when making requests.
+func (c *ServiceClient) ServiceVersion() ServiceVersion {
```
JeffreyRichter: I don't think we need this function at all. Do you disagree?
If you think we do, then we'd need a way to return ALL of options and I really don't think this is necessary.

sdk/storage/azblob/service.go
```diff
@@ -95,7 +141,7 @@ func (c *ServiceClient) listContainersPreparer(options *ListContainersOptions) *
 		}
 	}
 	req.SetQueryParam("comp", "list")
-	req.Header.Set("x-ms-version", c.ServiceVersion())
+	req.Header.Set("x-ms-version", string(c.ServiceVersion()))
```
JeffreyRichter: This would be in a pipeline policy now.

sdk/storage/azblob/service.go
jhendrixMSFT: Agreed it can be removed.

sdk/storage/azblob/service.go
jhendrixMSFT: Will fix.

jhendrixMSFT:
Closing as no longer needed.

PR #6377, opened 2019-11-19
use a lock-free algorithm to refresh bearer tokens

jhendrixMSFT:


sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,62 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
+	bt := bearerTokenPolicy{creds: creds, scopes: scopes}
+	// set initial values to their zero-value
+	bt.header.Store("")
+	bt.expiresOn.Store(time.Time{})
+	return &bt
 }
 
 func (b *bearerTokenPolicy) Do(ctx context.Context, req *azcore.Request) (*azcore.Response, error) {
+	// TODO: is window really constant, what about short-lived tokens?
+	const window = 5 * time.Minute
```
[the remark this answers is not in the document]
catalinaperalta: So there's no constraint stopping it from being 5 min, but other sdks do have about 2 min.

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,62 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
 }
 
 type bearerTokenPolicy struct {
-	// take lock when manipulating header/expiresOn fields
-	lock      sync.RWMutex
-	header    string
-	expiresOn time.Time
+	updating  int64                  // atomically set to 1 to indicate a refresh is in progress
+	header    atomic.Value           // string
```
JeffreyRichter: Because atomic stuff is so error prone and it's easy for someone to assert incorrectly or just access the field directly without using atomic.GetValue, could we add some some structs with members like what Uber has done here: https://github.com/uber-go/atomic/blob/master/atomic.go.

We just need to add the types/methods we require and it can grow over time.

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,62 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
 	now := time.Now().UTC()
-	if now.Equal(b.expiresOn) || now.After(b.expiresOn) {
-		// token has expired, take the write lock then check again
-		b.lock.RUnlock()
-		// don't defer Unlock(), we want to release it ASAP
-		b.lock.Lock()
-		if now.Equal(b.expiresOn) || now.After(b.expiresOn) {
-			// token has expired, get a new one and update shared state
+	exp := b.expiresOn.Load().(time.Time)
+	if now.After(exp) {
```
JeffreyRichter: I think you want to use the window here; if the expiration is within 5 seconds of NOW

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,62 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
+		// token has expired, set update flag and begin the refresh
+		if atomic.CompareAndSwapInt64(&b.updating, 0, 1) {
```
JeffreyRichter: Add comment: If no other goroutine is refreshing the token, the calling goroutine will do it

sdk/azidentity/client_secret_credential.go
jhendrixMSFT: The value stored in `expiresOn` has the window subtracted from it (see line 95).

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,62 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
 			tk, err := b.creds.GetToken(ctx, b.scopes)
 			if err != nil {
-				b.lock.Unlock()
+				atomic.StoreInt64(&b.updating, 0)
```
JeffreyRichter: Comment: We're done updating the token; if the new token expires, another goroutine can refresh it.
ALSO: This should be done AFTER setting the new token & expiration time to prevent another goroutine from also fetching it due to stale data.

sdk/azidentity/client_secret_credential.go
JeffreyRichter: I would do it this way. I'd stop the actual expiration time that the service gives back to us in the field. And after getting the value, subtract the windows and then compare to NOW. 

This way, the field always stores the actual value returned from the service which is really useful for debugging - we know what the service returned (not some mutated value). We mutate the value ephemerally just for comparison.

Also, imagine someday, we dynamically adjust the window as the program runs. The expires time is always the actual expires times and we can dynamically adjust the window over time WHEN we do the comparison.

sdk/azidentity/client_secret_credential.go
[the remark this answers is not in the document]
jhendrixMSFT: What if we made this configurable via `AuthenticationPolicyOptions` with a default of two minutes.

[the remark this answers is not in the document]
JeffreyRichter: Yes, we could but I prefer to just hard code it for now and NOT offer it as an option to customers. We can always add it later but I'm afraid customers can too easily screw it up.

sdk/azidentity/client_secret_credential.go
jhendrixMSFT: Fixed.

sdk/azidentity/client_secret_credential.go
jhendrixMSFT: Fixed.

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,47 +55,65 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
+		for {
+			bt = b.header.Load()
+			// for the very first call to Do() the header will be
+			// the empty string.  in this case we need to spin-wait
+			// until header contains a value.  subsequent calls to
+			// Do() will never spin-wait.
+			if bt != "" {
+				break
+			}
+			runtime.Gosched()
```
JeffreyRichter: This Gosched bothers me. I'd prefer a time.Sleep() with 100ms or so.
There is another potential problem. If a thread tried to get the token and this errors out, then we might have goroutines spinning forever and none of them ever try to get the token again and then never error out.

So, if bt == "", then these go routine should also spin larger trying to do the CAS to try to get the token.

jhendrixMSFT: A few other spin-lock implementations I found online use Gosched(), it yields the processor without suspending the calling go routine compared to sleep which will suspend the calling go routine.
You're right about the error case, I was thinking about that this morning.  Ideally we should exit the spin-wait and retry from the beginning (will work on this).

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,48 +54,65 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
+	bt.empty <- true
+	return &bt
 }
 
 func (b *bearerTokenPolicy) Do(ctx context.Context, req *azcore.Request) (*azcore.Response, error) {
-	var bt string
-	// take read lock and check if the token has expired
-	b.lock.RLock()
+	// check if the token is empty
+	if <-b.empty {
```
JeffreyRichter: I'd add some more comments here for other developers. Something like:
// This return true if this is the first goroutine to try to use this token; it returns false to all other goroutines after the 1st token has been obtained.

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,48 +54,65 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
+		tk, err := b.creds.GetToken(ctx, b.scopes)
+		if err != nil {
+			// failed to get a token, let other go routines try
```
JeffreyRichter: ... let one goroutine try

sdk/azidentity/client_secret_credential.go
```diff
@@ -54,48 +54,65 @@ func (c *ClientSecretCredential) AuthenticationPolicy(options azcore.Authenticat
...
 	now := time.Now().UTC()
-	if now.Equal(b.expiresOn) || now.After(b.expiresOn) {
-		// token has expired, take the write lock then check again
-		b.lock.RUnlock()
-		// don't defer Unlock(), we want to release it ASAP
-		b.lock.Lock()
-		if now.Equal(b.expiresOn) || now.After(b.expiresOn) {
-			// token has expired, get a new one and update shared state
+	exp := b.expiresOn.Load()
+	if now.After(exp.Add(-window)) {
```
JeffreyRichter: you could do this:
time.Now().UTC().After(exp.Add(-window))

And get rid of line 93

PR #6384, opened 2019-11-19
refactor azblob

jhendrixMSFT:
move generated code to internal package
simplified codegen to contain only request creation/response handling
internal types are exposed via type aliasing in the public package
updated page iterator per latest discussion

### DIFF
--- a/sdk/storage/azblob/models.go
+++ b/sdk/storage/azblob/models.go
@@ -27,7 +27,7 @@ func (iter *ListContainersIterator) NextPage(ctx context.Context) (*ListContaine
 	if err != nil {
 		return nil, err
 	}
-	page, err := iter.client.s.ListContainersCreateResponse(resp)
+	page, err := iter.client.s.ListContainersHandleResponse(resp)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


sdk/storage/azblob/models.go
```diff
@@ -27,7 +27,7 @@ func (iter *ListContainersIterator) NextPage(ctx context.Context) (*ListContaine
 	if err != nil {
 		return nil, err
 	}
-	page, err := iter.client.s.ListContainersCreateResponse(resp)
+	page, err := iter.client.s.ListContainersHandleResponse(resp)
```
JeffreyRichter: Yes, all looks good but remember we need a CurrentPageMarker "property" on the iterator too.

jhendrixMSFT: It's already in the returned page, you want to add it to the iterator itself?

JeffreyRichter:
No, you’re right – I forgot. Do not add the property.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://azure.microsoft.com/training/distributed-cloud-apps/>)

Issue #6387, opened 2019-11-19
Add tests for pipeline policies


jhendrixMSFT:
Need to add tests for each pipeline policy.

jhendrixMSFT:
Fixed in https://github.com/Azure/azure-sdk-for-go/pull/6410

PR #6410, opened 2019-11-22
add internal mock package for testing

jhendrixMSFT:
This gets us to 39.9% code coverage of `azcore`

### DIFF
--- a/sdk/azcore/internal/mock/mock.go
+++ b/sdk/azcore/internal/mock/mock.go
@@ -1,0 +1,186 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package mock
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+)
+
+// Server is a wrapper around an httptest.Server.
+type Server struct {
+	srv    *httptest.Server
+	static *mockResponse
+	resp   []mockResponse
+	req    int
+}
+
+// NewServer creates a new Server object.
+// The returned close func must be called when the server is no longer needed.
+func NewServer() (*Server, func()) {
+	s := Server{}
+	s.srv = httptest.NewServer(http.HandlerFunc(s.serveHTTP))
+	return &s, func() { s.srv.Close() }
+}
+
+// returns true if the next response is an error response
+func (s *Server) isErrorResp() bool {
+	if s.static == nil && len(s.resp) == 0 {
+		panic("no more responses")
+	}
+	// always favor static response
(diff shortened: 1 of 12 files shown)


jhendrixMSFT:
45.4% with latest commit

jhendrixMSFT:
74.1% with latest commits

jhendrixMSFT:
75.6%

jhendrixMSFT:
We should consider moving the `mock` package into a shared internal module so that it can be used by other modules (probably the `uuid` functionality as well).  We should discuss next week.

sdk/azcore/internal/mock/mock.go
```diff
@@ -0,0 +1,186 @@
...
+	"net/http/httptest"
+	"net/url"
+)
+
+// Server is a wrapper around an httptest.Server.
+type Server struct {
+	srv    *httptest.Server
+	static *mockResponse
+	resp   []mockResponse
+	req    int
```
JeffreyRichter: Some comments would be good; I don't know what all these fields are for.
Also, do some of them (like req) need to be thread safe?

sdk/azcore/policy_retry.go
```diff
@@ -21,6 +21,10 @@ const (
 	RetryPolicyFixed RetryPolicy = 1
 )
 
+const (
+	defaultMaxRetries = 4
```
JeffreyRichter: Is it MaxRetries or max tries? I used MaxTries in the Storage libraries because 0 tries makes no sense and 0 is the default value so I could set it to default max tries. If it's max retries then 0 DOES make sense; 1 try and no retries. So then 0 can't be a sentinel value to change it to default max tries.

sdk/azcore/policy_retry.go
```diff
@@ -87,7 +91,7 @@ func (o RetryOptions) defaults() RetryOptions {
 
 	// Set defaults if unspecified
 	if o.MaxTries == 0 {
```
JeffreyRichter: See previous comment

sdk/azcore/internal/mock/mock.go
jhendrixMSFT: Good point I'll add some comments here on the next round of test implementation.

sdk/azcore/policy_retry.go
jhendrixMSFT: You are right it should be `defaultMaxTries`, will fix in next round of clean-up.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/6387

Issue #6363, opened 2019-11-15
Fix problem when sending nil to AuthenticationFailedError constructor


catalinaperalta:
Panic when sending nil response payload to newAuthenticationFailedError()

catalinaperalta:
Added in linked PR #6432

Issue #6360, opened 2019-11-15
Fix problem when passing non-nil Identity Client Options 


catalinaperalta:
When passing non-nil Identity Client Options the authority host needs to be set or else the application will panic. Add appropriate checks when creating the new AAD identity client and look into the corresponding changes in MSI client.

catalinaperalta:
Added in linked PR #6432

Issue #6347, opened 2019-11-15
Fix ChainedTokenCredential logic and error handling


catalinaperalta:
The loop should only continue if it received a Credential Unavailable error, otherwise it should return the Token or the other error.
Fix error handling to return a credential unavailable error that includes the slice of errors. Return a nice message stating which credential failed and why.

catalinaperalta:
This is related to #6506

Issue #6346, opened 2019-11-15
Add init functions in azidentity


catalinaperalta:
Create init functions for URL initialization for AAD Identity Client authority host and for Managed Identity Client IMDS endpoint.

catalinaperalta:
Added in linked PR #6432

PR #6432, opened 2019-11-26
Azidentity review fixes

catalinaperalta:
This PR includes the following changes:
1. Adding DefaultTokenCredentialOptions to the DefaultTokenCredential (#6342)
2. Adding TokenRequestOptions to GetToken methods (#6344)
3. Adding ManagedIdentityCredentialOptions to ManagedIdentityCredential (#6345)
4. Add init functions (#6346):
- Init in azidentity: initialize defaultAuthorityHostURL and defaultIdentityClientOpts 
- Init in managed_identity_client: imdsURL and defaultMSIOpts
5.  Corrected ChainedTokenCredential GetToken logic (#6347 )
6. Fixed problem when passing empty Authority Host in Identity Client Options (#6360)

Other fixes:
1. Fixed internal func newAuthenticationFailedError (#6363)
2. Fixed instances of errors.Is when they should be errors.As


sdk/azcore/credential.go
```diff
@@ -6,14 +6,16 @@ package azcore
...
 	"time"
 )
 
 // AuthenticationPolicyOptions contains various options used to create a credential policy.
 type AuthenticationPolicyOptions struct {
 	// Scopes is the list of OAuth2 authentication scopes used when requesting a token.
 	// This field is ignored for other forms of authentication (e.g. shared key).
-	Scopes []string
+	Scopes        []string
+	AuthorityHost url.URL
```
catalinaperalta: Check if this is actually needed

sdk/azcore/credential.go
```diff
@@ -35,7 +37,7 @@ func (cf credentialFunc) AuthenticationPolicy(options AuthenticationPolicyOption
 type TokenCredential interface {
 	Credential
 	// GetToken requests an access token for the specified set of scopes.
-	GetToken(ctx context.Context, scopes []string) (*AccessToken, error)
+	GetToken(ctx context.Context, opts *TokenRequestOptions) (*AccessToken, error)
```
catalinaperalta: Change to pass by val because you always have to pass at least one scope

sdk/azidentity/aad_identity_client.go
```diff
@@ -77,8 +77,7 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 		return nil, err
 	}
 
-	// TODO: look into HasStatusCode
-	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusCreated {
+	if hasStatusCode(resp, successStatusCodes[:]...) {
```
catalinaperalta: Export HasStatusCode from azcore on Response type

sdk/azidentity/azidentity.go
```diff
@@ -14,6 +15,39 @@ const (
 	defaultAuthorityHost = "https://login.microsoftonline.com/"
 )
 
+var (
+	// StatusCodesForRetry is the default set of HTTP status code for which the policy will retry.
+	StatusCodesForRetry = [6]int{
```
catalinaperalta: Remove this

sdk/azidentity/azidentity.go
```diff
@@ -14,6 +15,39 @@ const (
...
+)
+
+var (
+	defaultAuthorityHostURL   *url.URL
+	defaultIdentityClientOpts *IdentityClientOptions
+)
+
+func init() {
+	var err error
+	defaultAuthorityHostURL, err = url.Parse(defaultAuthorityHost)
```
catalinaperalta: Keep the error check in the test and remove it here

sdk/azidentity/azidentity.go
```diff
@@ -30,7 +64,12 @@ func (e *AuthenticationFailedError) Error() string {
 }
 
 func newAuthenticationFailedError(resp *azcore.Response) error {
-	var authFailed *AuthenticationFailedError
+	authFailed := &AuthenticationFailedError{}
+	if resp == nil {
```
catalinaperalta: Remove this

sdk/azidentity/azidentity.go
```diff
@@ -42,27 +81,30 @@ func newAuthenticationFailedError(resp *azcore.Response) error {
...
-// AggregateError a specific error type for the chained token credential
-type AggregateError struct {
-	Msg     string
-	Err     error
-	ErrList []error
+// ChainedCredentialError an error specific to ChainedTokenCredential and DefaultTokenCredential
+// this error type will return a list of Credential Unavailable errors
+type ChainedCredentialError struct {
+	ErrorList []*CredentialUnavailableError
+	Message   string
```
catalinaperalta: Remove Message field

sdk/azidentity/aad_identity_client.go
```diff
@@ -50,15 +50,15 @@ type aadIdentityClient struct {
 // options: IdentityClientOptions that adds policies for the pipeline and the authority host that
 // will be used to retrieve tokens and authenticate
 func newAADIdentityClient(options *IdentityClientOptions) *aadIdentityClient {
-	options = options.setDefaultValues()
+	if options == nil {
```
catalinaperalta: Remove these checks and leave them centralized in setDefaultValues()

sdk/azidentity/azidentity.go
```diff
@@ -94,6 +138,34 @@ func newDefaultPipeline(o azcore.PipelineOptions) azcore.Pipeline {
 		azcore.NewRequestLogPolicy(o.LogOptions))
 }
 
+// hasStatusCode returns true if the Response's status code is one of the specified values.
+func hasStatusCode(r *azcore.Response, statusCodes ...int) bool {
```
catalinaperalta: Use the one from azcore

sdk/azidentity/azidentity_test.go
```diff
@@ -14,3 +15,12 @@ func TestAuthorityHostParse(t *testing.T) {
 		t.Fatalf("Failed to parse default authority host: %v", err)
 	}
 }
+
+func TestNilAuthFailedParam(t *testing.T) {
```
catalinaperalta: remove this

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,37 +19,39 @@ type ChainedTokenCredential struct {
...
-		// CP: This currently works the same way as the other languages
-		// TODO: instead of appending to an error slice, check if the error returned is an auth failure then stop the function completely and return the auth failure, credunavailable is still ok to continue looping
-		if err != nil {
-			errList = append(errList, err)
+	var errList []*CredentialUnavailableError
+
+	for i := 0; i < len(c.sources); i++ {
+		token, err = c.sources[i].GetToken(ctx, opts)
+		if errors.As(err, &credErr) {
+			errList = append(errList, err.(*CredentialUnavailableError))
```
catalinaperalta: use credErr in slice append, declare cred error in loop

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,37 +19,39 @@ type ChainedTokenCredential struct {
...
+		} else {
+			if token != nil && err == nil {
```
catalinaperalta: update logic, use else if and return token outside of if block

catalinaperalta: else if err != nil { nil, err}

return token, nil

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,37 +19,39 @@ type ChainedTokenCredential struct {
...
+				return token, nil
+			}
+			return token, &ChainedCredentialError{Message: err.Error(), ErrorList: errList}
 		}
 	}
 
 	if token == nil && len(errList) > 0 {
```
catalinaperalta: Change to if token == nil && len(errList) == 0 
Add comment above saying that we should never reach this point

catalinaperalta: Leave ChainedCredentialError as default return statement after the if block

sdk/azidentity/client_secret_credential.go
```diff
@@ -81,7 +71,7 @@ func (b *bearerTokenPolicy) Do(ctx context.Context, req *azcore.Request) (*azcor
 	// the first read.  all other reads will block until the
 	// token has been obtained at which point it returns false
 	if <-b.empty {
-		tk, err := b.creds.GetToken(ctx, b.scopes)
+		tk, err := b.creds.GetToken(ctx, &azcore.TokenRequestOptions{Scopes: b.scopes})
```
catalinaperalta: Change scopes field to TokenRequestOptions

sdk/azidentity/client_secret_credential.go
```diff
@@ -125,3 +115,6 @@ func (b *bearerTokenPolicy) Do(ctx context.Context, req *azcore.Request) (*azcor
 }
 
 var _ azcore.TokenCredential = (*ClientSecretCredential)(nil)
+
+// TODO: rename credentialpolicyoptions
```
catalinaperalta: Remove these comments

sdk/azidentity/default_azure_credential.go
```diff
@@ -3,64 +3,85 @@
...
 // types will be tried, in order:
 // - EnvironmentCredential
 // - ManagedIdentityCredential
 // Consult the documentation of these credential types for more information on how they attempt authentication.
-func NewDefaultTokenCredential(o *IdentityClientOptions) (*ChainedTokenCredential, error) {
-	// CP: This is fine because we are not calling GetToken we are simple creating the new EnvironmentClient
-	envClient, err := NewEnvironmentCredential(o)
+func NewDefaultTokenCredential(options *DefaultTokenCredentialOptions) (*ChainedTokenCredential, error) {
+	cred := &DefaultTokenCredential{}
+	err := cred.getDefaultTokenCredentialChain(options)
```
catalinaperalta: Add nil options check before calling getDefaultTokenCredentialChain()

sdk/azidentity/default_azure_credential.go
```diff
@@ -3,64 +3,85 @@
...
+	return NewChainedTokenCredential(cred.sources...)
+}
+
+func (c *DefaultTokenCredential) getDefaultTokenCredentialChain(options *DefaultTokenCredentialOptions) error {
+	errCount := 0
+	var envCred *ClientSecretCredential
+	var msiCred *ManagedIdentityCredential
+	var err error
+
+	if options == nil {
```
catalinaperalta: This goes away when passing in initialized opts above

sdk/azidentity/default_azure_credential.go
```diff
@@ -3,64 +3,85 @@
...
 	}
 
-	return NewChainedTokenCredential(
-		envClient,
-		msiClient,
-		&credentialNotFoundGuard{})
-}
+	if !options.ExcludeEnvironmentCredential {
+		envCred, err = NewEnvironmentCredential(nil)
+		if err != nil {
```
catalinaperalta: Check if err == nil { return cred, nil}

sdk/azidentity/fingerprint.go
```diff
@@ -49,11 +49,11 @@ func spkiFingerprint(cert string) (fingerprint, error) {
 	_, err = buffer.Read(pemBytes)
 	// Get first block of PEM file
 	data, rest := pem.Decode([]byte(pemBytes))
-
-	if data.Type != "CERTIFICATE" {
+	const key = "CERTIFICATE"
```
catalinaperalta: Rename

sdk/azidentity/managed_identity_client.go
```diff
@@ -43,53 +49,94 @@ const (
...
+}
+
+var (
+	imdsURL        *url.URL
+	defaultMSIOpts *ManagedIdentityCredentialOptions
+)
+
+func init() {
+	var err error
+	imdsURL, err = url.Parse(imdsEndpoint)
```
catalinaperalta: Fail only in test for parse not here

sdk/azidentity/managed_identity_client.go
```diff
@@ -43,53 +49,94 @@ const (
 type managedIdentityClient struct {
 	options                IdentityClientOptions
 	pipeline               azcore.Pipeline
-	sIMDSEndpoint          *url.URL
+	imdsEndpoint           *url.URL
```
catalinaperalta: Remove and leave parsed url as an unexported var parsed in init

sdk/azidentity/managed_identity_client.go
```diff
@@ -43,53 +49,94 @@ const (
...
+		// TODO change this
+		panic(err)
+	}
+
+	defaultMSIOpts = newDefaultManagedIdentityOptions()
+}
+
+func newDefaultManagedIdentityOptions() *ManagedIdentityCredentialOptions {
+	// CP: fix this implementation
+	opts := &ManagedIdentityCredentialOptions{}
```
catalinaperalta: Consider removing IdentityClientOptions and possibly unexported retry options

sdk/azidentity/managed_identity_client.go
```diff
@@ -43,53 +49,94 @@ const (
...
-	sType, err := c.getMSIType(ctx)
+	currentMSI, err := c.getMSIType(ctx)
 	if err != nil {
 		return nil, err
 	}
 	// if msi is unavailable or we were unable to determine the type return a nil access token
-	if sType == unavailable || sType == unknown {
-		// TODO: add message
-		return nil, &CredentialUnavailableError{}
+	if currentMSI == unavailable || currentMSI == unknown {
```
catalinaperalta: Add comment that this should never happen

sdk/azidentity/managed_identity_client.go
```diff
@@ -43,53 +49,94 @@ const (
...
+		// CP: maybe this should be an auth failed error?
+		return nil, &CredentialUnavailableError{CredentialType: "Managed Identity Credential", Message: "Please make sure you are running in a managed identity environment, such as a VM, Azure Functions, Cloud Shell, etc..."}
```
catalinaperalta: Should this be a SDK error of some kind?

sdk/azidentity/managed_identity_client.go
```diff
@@ -147,14 +194,14 @@ func (c *managedIdentityClient) createAuthRequest(msiType msiType, clientID stri
 	case cloudShell:
 		req, err = c.createCloudShellAuthRequest(clientID, scopes)
 	default:
-		// TODO: return an error CredentialUnavailable
+		return nil, &CredentialUnavailableError{CredentialType: "Managed Identity Credential", Message: "Make sure you are running in a valid Managed Identity Environment"}
```
catalinaperalta: Include msiType in message

sdk/azidentity/managed_identity_client.go
```diff
@@ -197,81 +244,40 @@ func (c *managedIdentityClient) createCloudShellAuthRequest(clientID string, sco
...
 		}
 	}
-	return c.sMSIType, nil
+	return c.msiType, nil
 }
 
 func (c *managedIdentityClient) imdsAvailable(ctx context.Context) bool {
-	request := c.pipeline.NewRequest(http.MethodGet, *c.sIMDSEndpoint)
-	request.Header.Set(azcore.HeaderMetadata, "true")
+	tempCtx, cancel := context.WithTimeout(ctx, 500*time.Second)
```
catalinaperalta: time.Millisecond

sdk/azidentity/managed_identity_client.go
```diff
@@ -197,81 +244,40 @@ func (c *managedIdentityClient) createCloudShellAuthRequest(clientID string, sco
...
+	defer cancel()
+	request := c.pipeline.NewRequest(http.MethodGet, *c.imdsEndpoint)
 	q := request.URL.Query()
 	q.Add("api-version", c.imdsAPIVersion)
 	request.URL.RawQuery = q.Encode()
-	// TODO: missing setting timeout and handling it
-	_, err := request.Do(ctx)
-	if err != nil {
+	_, err := request.Do(tempCtx)
+	if err != nil || errors.Is(err, context.DeadlineExceeded) {
```
catalinaperalta: remove or context.DeadlineExceeded

catalinaperalta: Just return err == nil

sdk/azcore/credential.go
catalinaperalta: Fixed

sdk/azidentity/azidentity.go
catalinaperalta: Fixed

sdk/azidentity/aad_identity_client.go
catalinaperalta: Fixed

sdk/azidentity/azidentity.go
catalinaperalta: Fixed, added resp == nil code in the authenticate() methods in aad_identity_client

sdk/azidentity/azidentity.go
catalinaperalta: Fixed

sdk/azidentity/azidentity.go
catalinaperalta: Fixed

sdk/azidentity/azidentity_test.go
catalinaperalta: Done

sdk/azidentity/client_secret_credential.go
catalinaperalta: Done

sdk/azidentity/client_secret_credential.go
catalinaperalta: Done

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go
catalinaperalta: Fixed

sdk/azidentity/managed_identity_client.go
catalinaperalta: Added a switch to check if the MSI env in unavailable or as a default unknown

sdk/azidentity/managed_identity_client.go

sdk/azidentity/fingerprint.go
catalinaperalta: Done

sdk/azcore/credential.go
catalinaperalta: I don't think so, since we pass in a custom authority host to the credential through Identity Client Options. If someone has a different perspective on this I'm glad to hear, but I'll remove the field for now and resolve this.

sdk/azidentity/azidentity.go
catalinaperalta: Done

sdk/azidentity/aad_identity_client.go

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done, except I put the return token, nil in the else statement. Feel free to comment again if you disagree.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done

sdk/azidentity/default_azure_credential.go
catalinaperalta: Done

sdk/azidentity/default_azure_credential.go
catalinaperalta: Done

sdk/azcore/credential.go
```diff
@@ -44,3 +45,8 @@ type AccessToken struct {
 	ExpiresIn json.Number `json:"expires_in"`
```
schaabs: Shouldn't have both ExpiresIn and ExpiresOn, Only ExpiresOn

sdk/azidentity/azidentity.go
```diff
@@ -102,6 +129,27 @@ func newDefaultPipeline(o azcore.PipelineOptions) azcore.Pipeline {
...
+// for a Managed Identity, such as a MSI specific retry policy
+func newDefaultMSIPipeline(o azcore.PipelineOptions) azcore.Pipeline {
+	if o.HTTPClient == nil {
+		o.HTTPClient = azcore.DefaultHTTPClientTransport()
+	}
+
+	// Set defaults needed for MSI retry policy
+	o.Retry.MaxTries = 5
+	o.Retry.Policy = azcore.RetryPolicyExponential
+	o.Retry.RetryDelay = 0
```
schaabs: Why retry delay of 0? Will the delay always be 0 in this case even with exponential backoff? i.e. 0^2 == 0

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,44 +19,49 @@ type ChainedTokenCredential struct {
...
 	}
 
 	return &ChainedTokenCredential{sources: sources}, nil
 }
 
 // GetToken sequentially calls TokenCredential.GetToken on all the specified sources, returning the first non default AccessToken.
-func (c *ChainedTokenCredential) GetToken(ctx context.Context, scopes []string) (*azcore.AccessToken, error) {
-	// CP: Notes from session with Jeff, have two funcs for the DefaultAzureCredential the main default func will issue the call that hits the wire, however internally first we create the chain and return a credUnavailable error if no credentials are found, if one works then we immediately return the one that works. End goal to always try to use the same credential type so we dont switch between credentials during execution, or at least reduce the possibility of that happening?
+func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	if len(opts.Scopes) == 0 {
```
schaabs: In other languages we don't do this validation in the client. if no scopes are specified we still send the request with no scopes (i.e. ...&scopes=&...

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,44 +19,49 @@ type ChainedTokenCredential struct {
...
-			errList = append(errList, err)
+	var errList []*CredentialUnavailableError
+
+	for i := 0; i < len(c.sources); i++ {
+		var credErr *CredentialUnavailableError
+		token, err = c.sources[i].GetToken(ctx, opts)
+		if errors.As(err, &credErr) {
+			errList = append(errList, credErr)
+		} else if err != nil {
+			return nil, err
```
schaabs: If errList is not empty you might want to return that information in some way here as well, so if they get an error from a credential they weren't expecting would be used they can see why preceding credentials were skipped.

sdk/azidentity/azidentity.go
```diff
@@ -82,9 +105,13 @@ type IdentityClientOptions struct {
 // NewIdentityClientOptions initializes an instance of IdentityClientOptions with default settings
 func (c *IdentityClientOptions) setDefaultValues() *IdentityClientOptions {
```
schaabs: you should rename IdentityClientOptions to TokenCredentialOptions

sdk/azidentity/managed_identity_client.go
schaabs: It should be a `CredentialUnavailbleError` as you have it here. It's the most accurate since the managed identity service isn't available. Also this allows for the correct flow when including the `ManagedIdentityCredential` in a `ChainedTokenCredential` or `DefaultAzureCredential`

sdk/azidentity/azidentity.go
catalinaperalta: I'm actually not using that variable. I'll delete it from the code.

sdk/azidentity/azidentity.go
catalinaperalta: Ok, that sounds good to me.

sdk/azidentity/managed_identity_client.go
catalinaperalta: Ok, I agree since ChainedTokenCredential and DefaultAzureCredential will use CredentialUnavailableError in their error handling.

sdk/azidentity/chained_token_credential.go
catalinaperalta: You're right. I removed it.

sdk/azcore/credential.go
catalinaperalta: I might change this if I see I do need to handle ExpiresIn differently for the MSI credential, but for Go this is a good solutions since we need to use ExpiresIn to unmarshal the JSON response (and the unmarshal needs to be done on an exported field). I'll look into the token responses for MSI and get back to this

schaabs: I do seem to recall that one type of endpoint only returned `ExpiresIn`, but if that's the case you should deserialize to an internal struct which has both fields and then calculate `ExpiresOn` from `ExpiresIn`. The AccessToken struct which is return from `TokenCredential.GetToken` should only have `ExpiresOn`

catalinaperalta: Adding internal struct for marshaling purposes in next PR

sdk/azidentity/chained_token_credential.go
catalinaperalta: Will fix error handling in next PR

PR #6449, opened 2019-11-28
v37.0.0

ArcturusZhang:
 

ArcturusZhang:
Hi @jhendrixMSFT I notice that in this update, the parameter order in `appplatform` has changed. 
Is there recently a change on autorest that could change this? I asked the service team that owns appplatform, there is no change on their swaggers.

jhendrixMSFT:
The change in ordering is due to the following commit.
https://github.com/Azure/azure-rest-api-specs/commit/98c74125ced2106024ec5a837d882c56e46765f5#diff-b79a9f207fe92ed29a3d10963a8d9c38
The parameter changed from optional to required.  Optional parameters are always sequenced after required parameters, and since this is no longer required it's placed in sort-order.

ArcturusZhang:
> The change in ordering is due to the following commit.
> [Azure/azure-rest-api-specs@98c7412#diff-b79a9f207fe92ed29a3d10963a8d9c38](https://github.com/Azure/azure-rest-api-specs/commit/98c74125ced2106024ec5a837d882c56e46765f5#diff-b79a9f207fe92ed29a3d10963a8d9c38)
> The parameter changed from optional to required. Optional parameters are always sequenced after required parameters, and since this is no longer required it's placed in sort-order.

I see. Actually currently, every parameters for this service is required and then the order of these parameters is same as the order they are defined, right? I just did a little experiment about this, which is saying so.
So I may need to ask the service team to alternate their order of parameters to make things right before i can release this major version.

ArcturusZhang:
> The change in ordering is due to the following commit.
> [Azure/azure-rest-api-specs@98c7412#diff-b79a9f207fe92ed29a3d10963a8d9c38](https://github.com/Azure/azure-rest-api-specs/commit/98c74125ced2106024ec5a837d882c56e46765f5#diff-b79a9f207fe92ed29a3d10963a8d9c38)
> The parameter changed from optional to required. Optional parameters are always sequenced after required parameters, and since this is no longer required it's placed in sort-order.

I submitted a [pr](https://github.com/Azure/azure-rest-api-specs/pull/7874) in spec to fix this.

ArcturusZhang:
The parameter order of appplatform has been fixed

ArcturusZhang:
It seems all the problems in this major version release are fixed! I manually excluded those malformed swagger when generating this major release.
@jhendrixMSFT please have a look, thanks 🚀

ArcturusZhang:
Hi @tombuildsstuff we have this SDK new version v37.0.0, and the issue with 32-bit compilation failure described [here](https://github.com/Azure/azure-rest-api-specs/pull/7847) should be resolved 🚀
Please have a look~

Issue #6351, opened 2019-11-15
Add testing for credential types in azidentity


catalinaperalta:
Implement tests for AAD identity client and Client Secret, Client Certificate, Environment, Default Azure and Chained Token Credentials. 
Implement tests for ManagedIdentityCredential

catalinaperalta:
Added tests for:
- ClientSecretCredential
- ClientCertificateCredential
- EnvironmentCredential
- DefaultTokenCredential
- ChainedTokenCredential
- UsernamePasswordCredential
- ManagedIdentityCredential

Tests were added in the following PRs:
#6493 
#6496 
#6497 
#6502 
#6503

PR #6493, opened 2019-12-04
Track2 identitytests

catalinaperalta:
Added tests for the following credential types:

- ClientSecretCredential
- ClientCertificateCredential
- ChainedTokenCredential

Also made some naming corrections and switched to use consts for repeated string values in tests. 
Related to issue: #6351 
Total azidentity test coverage so far: 47.5%


sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
+)
+
+const (
+	certificatePath      = "../../../testPEM.pem"
```
catalinaperalta: NOTE: I am working on adding a local PEM file for testing and will add it shortly

sdk/azidentity/managed_identity_credential_test.go
```diff
@@ -2,40 +2,22 @@ package azidentity
...
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-func Test_GetToken_CloudShell(t *testing.T) {
+func TestManagedIdentityCredential_GetTokenInCloudShell(t *testing.T) {
 	managedClient, err := NewManagedIdentityCredential("", newDefaultManagedIdentityOptions())
 	if err != nil {
 		fmt.Println("Managed ID error: ", err)
```
jhendrixMSFT: Shouldn't this be a `t.Fatalf()`?

sdk/azidentity/managed_identity_credential_test.go
```diff
@@ -2,40 +2,22 @@ package azidentity
...
 	} else {
-		managedAT, err := managedClient.GetToken(context.Background(), azcore.TokenRequestOptions{Scopes: []string{"https://storage.azure.com"}})
+		managedAT, err := managedClient.GetToken(context.Background(), azcore.TokenRequestOptions{Scopes: []string{scope}})
 		if err != nil {
 			fmt.Println("Error: ", err)
```
jhendrixMSFT: `t.Fatalf()`?

sdk/azidentity/managed_identity_credential_test.go
```diff
@@ -2,40 +2,22 @@ package azidentity
...
 		} else {
 			fmt.Println(managedAT)
```
jhendrixMSFT: I think this can be removed.

sdk/azidentity/managed_identity_credential_test.go
catalinaperalta: Yes, but this is one of those tests that must be run manually in the environment at the moment. So I left if with a Println. 

What I could do is check if the MSI_endpoint is set and handle testing depending on the environment we're in, do you think that would work for testing? 

I'd basically add a check for the endpoint and the result of the test would change depending on where it's available or not.

sdk/azidentity/managed_identity_credential_test.go
catalinaperalta: Same problem as before, this test would need to be run manually and would fail all the time when outside of the right environment

sdk/azidentity/managed_identity_credential_test.go
catalinaperalta: Agreed, I'll remove it

sdk/azidentity/managed_identity_credential_test.go
jhendrixMSFT: Ah I see.  If we don't want this test to run when that env var is set then you can call `t.Skip()` to skip running the test.

PR #6495, opened 2019-12-04
MSI pipeline uses retry policy in azcore

jhendrixMSFT:
added retry policy logging
removed extra sleep at end of retry


sdk/azcore/log.go
```diff
@@ -11,6 +11,7 @@ const (
 	LogError        LogClassification = "Error"
 	LogRequest      LogClassification = "Request"
 	LogResponse     LogClassification = "Response"
+	LogRetry        LogClassification = "Retry"
```
JeffreyRichter: This si really just for debugging the retry logic, right? If so, a comment to that effect would be nice to have here.

sdk/azcore/policy_retry.go
```diff
@@ -160,9 +161,12 @@ func (p *retryPolicy) Do(ctx context.Context, req *Request) (resp *Response, err
 		}
 		defer rwbody.realClose()
 	}
-	for try := int32(1); try <= p.options.MaxTries; try++ {
+	try := int32(1)
+	for {
 		resp = nil // reset
-		logf("\n=====> Try=%d\n", try)
+		if Log().Should(LogRetry) {
```
JeffreyRichter: You could capture the result of this once at the top of the method and then not call Should() multiple times in this method.

sdk/azcore/policy_retry.go
```diff
@@ -192,21 +198,27 @@ func (p *retryPolicy) Do(ctx context.Context, req *Request) (resp *Response, err
...
 
+		if try == p.options.MaxTries {
+			// max number of tries has been reached, don't sleep again
+			return
+		}
+
 		// use the delay from retry-after if available
 		delay, ok := resp.RetryAfter()
 		if !ok {
 			delay = p.options.calcDelay(try)
```
JeffreyRichter: By the way, we should only support exponential retries and nothing else so no need to have a retry option for the retry type. Also, we probably want a minimum delay; that is, a delay of 0 is never appropriate. Let's make a const for minimum delay and set it initially to 100 milliseconds.

sdk/azcore/policy_retry.go
jhendrixMSFT: Will refactor however I'd be surprised if the compiler didn't hoist this call out of the for body (loop invariant code motion).

sdk/azcore/policy_retry.go
jhendrixMSFT: For ARM we have to honor the `Retry-After` header if available (it takes precedence over our own delay calculation).

sdk/azcore/log.go
jhendrixMSFT: Will rename to `LogRetryPolicy`.  All of these enums need comments :(

JeffreyRichter:
The compiler can’t hoist this out because the Should list could be modified by other goroutines which means that some calls could return true and others could return false – this would also be strange to us because what gets logged could dynamically change within the function. Capturing the bool up front means consistent behavior with a single function invocation.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://azure.microsoft.com/training/distributed-cloud-apps/>)

JeffreyRichter:
Yes, for ARM but I was talking about non-ARM.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://azure.microsoft.com/training/distributed-cloud-apps/>)

PR #6496, opened 2019-12-04
EnvironmentCredential and DefaultTokenCredentialTests

catalinaperalta:
Added tests for EnvironmentCredential and DefaultTokenCredential.
Fixed some formatting.


sdk/azidentity/default_token_credential_test.go
```diff
@@ -0,0 +1,63 @@
+package azidentity
```
jhendrixMSFT: missing license header

sdk/azidentity/environment_credential_test.go
```diff
@@ -0,0 +1,108 @@
+package azidentity
```
jhendrixMSFT: missing license header

sdk/azidentity/default_token_credential_test.go
catalinaperalta: Added

sdk/azidentity/environment_credential_test.go
catalinaperalta: Added

PR #6503, opened 2019-12-06
Bearer Token Policy tests

catalinaperalta:
This PR includes: 

- Tests for BearerTokenPolicy
- Tests for Credentials that implement AuthenticationPolicy method

Test coverage for azidentity (not including IMDS test which needs to be live): 87.8%


sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
+)
+
+const (
+	accessTokenRespSuccess    = `{"access_token": "new_token", "expires_in": 3600}`
+	accessTokenRespError      = `{"access_token": 0, "expires_in": 3600}`
```
jhendrixMSFT: I think this is really `accessTokenRespMalformed` or something similar.  What does an actual error response look like?

catalinaperalta: I like your suggestion. I'll update the const. There isn't really an access token error, it's just an error from the service, like this:
{
    "error": "invalid_client",
    "error_description": "AADSTS7000215: Invalid client secret is provided.\r\nTrace ID: 2d091b11-76de-4bdb-8865-a5a9c0a20100\r\nCorrelation ID: a9983003-c09f-4122-b1d1-699de2feac49\r\nTimestamp: 2019-12-06 19:42:57Z",
    "error_codes": [
        7000215
    ],
    "timestamp": "2019-12-06 19:42:57Z",
    "trace_id": "2d091b11-76de-4bdb-8865-a5a9c0a20100",
    "correlation_id": "a9983003-c09f-4122-b1d1-699de2feac49",
    "error_uri": "https://login.microsoftonline.com/error?code=7000215"
}

jhendrixMSFT: Do we have tests that cover this scenario?

catalinaperalta: Yes, I mocked it by sending an unauthorized status code as a response to check this functionality, but I can also add another test that sends a json similar to this one as a response. I'll add it to this PR.

sdk/azidentity/client_secret_credential_test.go
```diff
@@ -75,7 +75,7 @@ func TestClientSecretCredential_GetTokenSuccess(t *testing.T) {
 func TestClientSecretCredential_GetTokenInvalidCredentials(t *testing.T) {
 	srv, close := mock.NewServer()
 	defer close()
-	srv.SetResponse(mock.WithStatusCode(http.StatusUnauthorized))
+	srv.SetResponse(mock.WithBody([]byte(`{"error": "invalid_client","error_description": "Invalid client secret is provided.","error_codes": [7000215],"timestamp": "2019-12-01 19:00:00Z","trace_id": "2d091b0","correlation_id": "a999","error_uri": "https://login.contoso.com/error?code=7000215"}`)), mock.WithStatusCode(http.StatusUnauthorized))
```
jhendrixMSFT: If any of the error values are exposed in the error object it would be good to update the test to verify the values.

Issue #6506, opened 2019-12-06
Pending Identity Review Fixes


catalinaperalta:
- Create private struct for AccessToken unmarshaling
- Change AuthenticationFailedError to include a pointer to an error and add AuthenticationResponseError
- MSI retry policy consider removing 503 errors from acceptable retry codes...look into this case

catalinaperalta:
These fixes should be quick to implement, however they depend on the design we land on.

catalinaperalta:
Added in #6536

PR #6536, opened 2019-12-17
Changed error handling and updated AccessToken unmarshaling

catalinaperalta:
- Changed AuthenticationFailedError to only point to an inner error
- Created AuthenticationResponseError for errors returned from the service
- Added an internal AccessToken struct for unmarshaling purposes and for help in only passing the right values to the exported AccessToken type that exists in azcore. 

This PR is related to the review fixes stated in #6506


sdk/azidentity/azidentity.go
```diff
@@ -45,21 +55,30 @@ type AuthenticationFailedError struct {
...
 	msg := e.Message
 	if len(e.Description) > 0 {
 		msg += " " + e.Description
 	}
 	return msg
 }
 
+// AuthenticationFailedError is a struct used to marshal responses when authentication has failed
+type AuthenticationFailedError struct {
+	Err error
```
jhendrixMSFT: This should be `inner error` and provide an `Unwrap() error` method to properly support `As()` and `Is()`.

sdk/azidentity/azidentity.go
```diff
@@ -156,12 +175,17 @@ func newDefaultMSIPipeline(o ManagedIdentityCredentialOptions) azcore.Pipeline {
...
 		o.HTTPClient = azcore.DefaultHTTPClientTransport()
 	}
-
+	var statusCodes []int
 	// retry policy for MSI is not end-user configurable
 	retryOpts := azcore.RetryOptions{
 		MaxTries:   5,
 		RetryDelay: 2 * time.Second,
-		StatusCodes: append(azcore.StatusCodesForRetry[:],
+		StatusCodes: append(statusCodes,
```
jhendrixMSFT: Why did these codes get duplicated?

sdk/azidentity/azidentity.go
jhendrixMSFT: Looks like there should also be a `msg string` field for the case where there is no error.

sdk/azidentity/chained_token_credential.go
```diff
@@ -51,7 +50,7 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
 	if token == nil && len(errList) == 0 {
 		return nil, nil
 	}
-	return nil, &ChainedCredentialError{ErrorList: errList}
```
jhendrixMSFT: Is `ChainedCredentialError` still used someplace or can its type definition be removed now?

sdk/azidentity/managed_identity_client.go
```diff
@@ -120,43 +120,52 @@ func (c *managedIdentityClient) sendAuthRequest(ctx context.Context, msiType msi
...
-		return nil, fmt.Errorf("azcore.AccessToken: %w", err)
+		return nil, fmt.Errorf("internalAccessToken: %w", err)
+	}
+
+	if value.ExpiresIn == "" {
+		if value.ExpiresOn == "" {
+			return nil, &AuthenticationFailedError{Err: errors.New("did not receive a valid token expiration time")}
+		}
+		accessToken.Token = value.Token
+		// TODO: missing here is parsing expires on to a time.Time value
```
jhendrixMSFT: Does this TODO need to be addressed now?

sdk/azidentity/azidentity.go
catalinaperalta: Ok I added an Unwrap method on AuthenticationFailedError and a msg field which is returned when calling the Error method.

sdk/azidentity/managed_identity_client.go
catalinaperalta: It has been addressed, but it'll be merged in one of my next PRs.

sdk/azidentity/chained_token_credential.go
catalinaperalta: It's still used when we fail to find any available credentials in GetToken for ChainedTokenCredentials.

sdk/azidentity/azidentity.go
catalinaperalta: So, during the last review, we discussed possibly removing the 503 status code for retry since it would probably consume more time than we would (ideally?) want. I decided to copy the values over instead of doing an operation over the slice, because I think it'll be safer this way and this code doesnt need to check the other slice in case something is added before or after the 503 status code.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Actually we wont be using ChainedCredentialError anymore. I will add a task to remove it from the code in my next PR.

catalinaperalta: Here's the issue: #6549

sdk/azidentity/azidentity.go
jhendrixMSFT: Ah I didn't realize this is a subset, makes sense that it's copied.  Might be worth adding a comment indicating it's a subset of `StatusCodesForRetry`.

sdk/azidentity/aad_identity_client.go
```diff
@@ -74,14 +74,14 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 
 	// This should never happen under normal conditions
 	if resp == nil {
-		return nil, &AuthenticationFailedError{Message: "Something unexpected happened with the request and received a nil response"}
+		return nil, &AuthenticationFailedError{inner: errors.New("Something unexpected happened with the request and received a nil response")}
```
jhendrixMSFT: I think the `msg` field should be set to this.  `inner` should only be used if it's wrapping an error returned from some other operation.

sdk/azidentity/aad_identity_client.go
```diff
@@ -104,28 +104,30 @@ func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantI
...
```
jhendrixMSFT: Same as above, use `msg`.

sdk/azidentity/aad_identity_client.go
```diff
@@ -202,14 +204,14 @@ func (c *aadIdentityClient) authenticateUsernamePassword(ctx context.Context, te
...
```
jhendrixMSFT: Use `msg`

sdk/azidentity/aad_identity_client.go
catalinaperalta: Totally agree.

sdk/azidentity/aad_identity_client.go

sdk/azidentity/aad_identity_client.go

sdk/azidentity/azidentity.go
catalinaperalta: Added

sdk/azidentity/bearer_token_policy.go
```diff
@@ -31,9 +32,7 @@ func newBearerTokenPolicy(creds azcore.TokenCredential, opts azcore.Authenticati
 func (b *bearerTokenPolicy) Do(ctx context.Context, req *azcore.Request) (*azcore.Response, error) {
 	if req.URL.Scheme != "https" {
 		// HTTPS must be used, otherwise the tokens are at the risk of being exposed
-		return nil, &AuthenticationFailedError{
-			Message: "token credentials require a URL using the HTTPS protocol scheme",
-		}
+		return nil, &AuthenticationFailedError{inner: errors.New("token credentials require a URL using the HTTPS protocol scheme")}
```
jhendrixMSFT: Use `msg`.

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -84,7 +85,8 @@ func TestChainedTokenCredential_GetTokenSuccess(t *testing.T) {
 	if tk.Token != tokenValue {
 		t.Fatalf("Received an incorrect access token")
 	}
-	if tk.ExpiresIn != tokenExpiresIn {
+	emptyTime := time.Time{}
+	if tk.ExpiresOn == emptyTime {
```
jhendrixMSFT: `tk.ExpiresOn.IsZero()`

sdk/azidentity/managed_identity_client.go
```diff
@@ -120,43 +120,52 @@ func (c *managedIdentityClient) sendAuthRequest(ctx context.Context, msiType msi
 		return nil, err
 	}
 
+	// This should never happen under normal conditions
+	if resp == nil {
+		return nil, &AuthenticationFailedError{inner: errors.New("Something unexpected happened with the request and received a nil response")}
```
jhendrixMSFT: Use `msg`

sdk/azidentity/managed_identity_client.go
```diff
@@ -120,43 +120,52 @@ func (c *managedIdentityClient) sendAuthRequest(ctx context.Context, msiType msi
...
+	value := internalAccessToken{}
+	accessToken := &azcore.AccessToken{}
 	if err := json.Unmarshal(res.Payload, &value); err != nil {
-		return nil, fmt.Errorf("azcore.AccessToken: %w", err)
+		return nil, fmt.Errorf("internalAccessToken: %w", err)
+	}
+
+	if value.ExpiresIn == "" {
+		if value.ExpiresOn == "" {
+			return nil, &AuthenticationFailedError{inner: errors.New("did not receive a valid token expiration time")}
```
jhendrixMSFT: Use `msg`

sdk/azidentity/chained_token_credential_test.go
catalinaperalta: Done

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go

sdk/azidentity/chained_token_credential_test.go
catalinaperalta: Done

sdk/azidentity/bearer_token_policy.go
catalinaperalta: Done

PR #6551, opened 2019-12-18
Adding API Version parameter

paulomarquesc:
- [X] The PR targets the `latest` branch.
- [X] Updates to [CHANGELOG.md][] are included.
- [X] Apache v2 license headers are included in each file.
 
This PR fixes a bug related to hard coded APIVersion which is a constant that as of now reflects the GO SDK version instead of each specific resource provider version, which in most cases will fail the functions to retrieve resources when using the functions at resources.go. The changes are basically exposing a new parameter called APIVersion and passing it down to other functions (and of course adding it on each function) as needed. In order to get closer to the old behavior a test is made to check if APIVersion is "", if true, it will default to the value that was previosly defined in the constant of the same name.

Changes were applied to the most obvious functions, for the untouched functions (with exception of the ones that this surely doesn't apply), deeper investigation is needed and can be done in the future.

List of functions updated:

- CheckExistence
- CheckExistenceByID
- CreateOrUpdate
- CreateOrUpdateByID
- Delete
- DeleteByID
- Get
- GetByID
- Update
- UpdateByID

Note: I could not find the test cases for resources module.


jhendrixMSFT:
The API version constant is the service API version, it is not the Go SDK version.

The majority of the Go SDK is generated from swagger files.  The package being modified in this PR is generated from the following swagger.

https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Resources/stable/2019-05-01/resources.json

As you can see, the service API version for this package is 2019-05-01.

Any changes made to the source code will be lost the next time the package is generated, so if there's a bug it should be fixed in the swagger which is the source of truth.

Are you looking for a different API version?

paulomarquesc:
Hi @jhendrixMSFT, I didn't know this was auto-generated code. The problem is that whatever is adding the api version is adding a version that does not exist for that particular resource provider, I think we may have to work on the auto generation mechanism. This module I need fixed is about generic resources, and it is relaying on the API version for networking which is not 2019-05-01 that is being hard-coded, since I'm querying for a vnet using the general resource module Get method and since it is adding the same version of the SDK itself, I get the error that the API version is not supported for vnets on westus2 for example.

paulomarquesc:
BTW, I can workaround this by using the networks module (https://github.com/Azure/azure-sdk-for-go/blob/master/services/network/mgmt/2019-09-01/network) but I am worried about customers that needs to use generic resources, so if we could work on something that would be great.

jhendrixMSFT:
Just to confirm, you're calling the following API which is failing?

https://github.com/Azure/azure-sdk-for-go/blob/master/services/resources/mgmt/2019-05-01/resources/resources.go#L567

What are you using for parameters?  I'm not familiar with the working of this API (my team owns SDK generation and release) but I can find out.

paulomarquesc:
Yes, that's correct, but other similar methods will have the same issue. If you look at the code snippet you pointed out, it calls the [GetPreparer](https://github.com/Azure/azure-sdk-for-go/blob/781d9fb592a1818cb7fb8e0855ea9c00b095a9e9/services/resources/mgmt/2019-05-01/resources/resources.go#L608) and on that method, at line [618](https://github.com/Azure/azure-sdk-for-go/blob/781d9fb592a1818cb7fb8e0855ea9c00b095a9e9/services/resources/mgmt/2019-05-01/resources/resources.go#L618) there is the resources SDK module version and then the request is built using that api version [here](https://github.com/Azure/azure-sdk-for-go/blob/781d9fb592a1818cb7fb8e0855ea9c00b095a9e9/services/resources/mgmt/2019-05-01/resources/resources.go#L623-L627). But somehow it looks like it is redirecting to the RP related to Network and that one is checking the API version coming and that does not match, thus the error message I pasted previously.

paulomarquesc:
This is the code snippet I'm working on, of course there is the APIVersion I added there as the last argument.

```go
func GetResourceByID(ctx context.Context, resourceID string, APIVersion string) (resources.GenericResource, error) {

	resourcesClient, err := getResourcesClient()
	if err != nil {
		return resources.GenericResource{}, err
	}

	parentResource := ""
	resourceGroup, _ := uri.GetResourceGroup(resourceID)
	resourceProvider, _ := uri.GetResourceValue(resourceID, "providers")
	resourceName, _ := uri.GetResourceName(resourceID)
	resourceType, _ := uri.GetResourceValue(resourceID, resourceProvider)

	if strings.Contains(resourceID, "/subnets/") {
		parentResource, _ = uri.GetResourceValue(resourceID, resourceType)
		resourceType = "subnets"
	}

	return resourcesClient.Get(
		ctx,
		resourceGroup,
		resourceProvider,
		parentResource,
		resourceType,
		resourceName,
		APIVersion,
	)
}
```

jhendrixMSFT:
Thanks for the info, I was able to repro this using a service bus instance in my subscription.  I have no idea what to pass for the `parentResourcePath` parameter so I left it blank which I'm sure is incorrect.  Wondering if this is simply the case of bad parameters and a misleading error message.

paulomarquesc:
You did it correctly, if you pass the parentResourcePath as blank it will work fine if the API of the service bus matches the API version being hard-coded as the resources module. In my code snippet this is used for child resources like a subnet, custom script extension, etc.

jhendrixMSFT:
I spoke to the service owner, indeed the Go codegen is incorrect (other languages don't have this problem).  I'll work on getting the codegen fixed up.
CC @ArcturusZhang

jhendrixMSFT:
This will be fixed via https://github.com/Azure/autorest.go/pull/184.  Once this is merged we can generate updated packages, however since it introduces a breaking change we'll have to wait for the next major release (v38 I believe due the end of this month).

PR #6552, opened 2019-12-18
decouple request creation from pipeline

jhendrixMSFT:
added Pipeline.Do() for sending requests through a pipeline
renamed Request.Do() to Request.Next() for advancing to the next policy
added ErrNoMorePolicies if Request.Next() is called on an empty pipeline

### DIFF
--- a/sdk/azcore/policy_anonymous_credential_test.go
+++ b/sdk/azcore/policy_anonymous_credential_test.go
@@ -17,8 +17,8 @@ func TestAnonymousCredential(t *testing.T) {
 	defer close()
 	srv.SetResponse(mock.WithStatusCode(http.StatusOK))
 	pl := NewPipeline(srv, AnonymousCredential().AuthenticationPolicy(AuthenticationPolicyOptions{}))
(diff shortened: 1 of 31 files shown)


sdk/azcore/policy_anonymous_credential_test.go
```diff
@@ -17,8 +17,8 @@ func TestAnonymousCredential(t *testing.T) {
 	defer close()
 	srv.SetResponse(mock.WithStatusCode(http.StatusOK))
 	pl := NewPipeline(srv, AnonymousCredential().AuthenticationPolicy(AuthenticationPolicyOptions{}))
-	req := pl.NewRequest(http.MethodGet, srv.URL())
-	resp, err := req.Do(context.Background())
+	req := NewRequest(http.MethodGet, srv.URL())
```
catalinaperalta: Are we leaving this as a publicly exported function?

jhendrixMSFT: Yep!  We no longer have to create a request via a pipeline, just create it directly.

Issue #6549, opened 2019-12-18
Remove ChainedCredentialError from Azure Identity 


catalinaperalta:

catalinaperalta:
Removed in the referenced PRs.

Issue #6433, opened 2019-11-26
Add Device Code Credential


catalinaperalta:
Implement and test Device Code Credential in azidentity

catalinaperalta:
Completed in referenced PR #6560

Issue #6350, opened 2019-11-15
Add separate handling for different Access Tokens 


catalinaperalta:
Add functionality to parse the information returned in the expires_in field of the access token returned from MSI.

catalinaperalta:
Included in this PR: #6560

catalinaperalta:
This was added in the referenced PR #6560

PR #6560, opened 2019-12-19
Adding DeviceCodeCredential, MSI Access Token creation, fixed error handling in ChainedTokenCredential

catalinaperalta:
This PR includes: 
- Adding DeviceCodeCredential and corresponding unit/mock tests  #6433
- Adding MSI Access Token parsing to include ExpiresOn cases #6350
- Updating error handling for ChainedTokenCredential
- Fixed MSI tests to use MSI specific scopes
- Added IMDS request creation test
- Updated DefaultTokenCredential chain creation
- Updated MSI type const names
- Removed ChainedCredentialError #6549

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -283,8 +283,8 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
 	return msg, nil
 }
 
-func createDeviceCodeResult(res *azcore.Response) (*DeviceCodeResult, error) {
-	value := &DeviceCodeResult{}
+func createDeviceCodeResult(res *azcore.Response) (*deviceCodeResult, error) {
+	value := &deviceCodeResult{}
 	if err := json.Unmarshal(res.Payload, &value); err != nil {
 		return nil, fmt.Errorf("DeviceCodeResult: %w", err)
 	}
@@ -342,7 +342,7 @@ func (c *aadIdentityClient) createDeviceCodeAuthRequest(tenantID string, clientI
 	return msg, nil
 }
 
-func (c *aadIdentityClient) requestNewDeviceCode(ctx context.Context, tenantID, clientID string, scopes []string) (*DeviceCodeResult, error) {
+func (c *aadIdentityClient) requestNewDeviceCode(ctx context.Context, tenantID, clientID string, scopes []string) (*deviceCodeResult, error) {
 	msg, err := c.createDeviceCodeNumberRequest(tenantID, clientID, scopes)
 	if err != nil {
 		return nil, err
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -31,8 +31,10 @@ const (
 	qpClientAssertion     = "client_assertion"
 	qpClientID            = "client_id"
 	qpClientSecret        = "client_secret"
+	qpDeviceCode          = "device_code"
 	qpGrantType           = "grant_type"
 	qpPassword            = "password"
+	qpRefreshToken        = "refresh_token"
 	qpResponseType        = "response_type"
 	qpScope               = "scope"
 	qpUsername            = "username"
@@ -61,7 +63,37 @@ func newAADIdentityClient(options *TokenCredentialOptions) *aadIdentityClient {
 // clientID: The client (application) ID of the service principal
 // clientSecret: A client secret that was generated for the App Registration used to authenticate the client
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, clientID string, clientSecret string, scopes []string) (*azcore.AccessToken, error) {
+func (c *aadIdentityClient) refreshAccessToken(ctx context.Context, tenantID string, clientID string, clientSecret string, refreshToken string, scopes []string) (*tokenResponse, error) {
+	msg, err := c.createRefreshTokenRequest(tenantID, clientID, clientSecret, refreshToken, scopes)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.pipeline.Do(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	// This should never happen under normal conditions
+	if resp == nil {
+		return nil, &AuthenticationFailedError{msg: "Something unexpected happened with the request and received a nil response"}
+	}
+
+	if resp.HasStatusCode(successStatusCodes[:]...) {
+		return c.createAccessToken(resp)
+	}
+
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
+}
+
+// Authenticate creates a client secret authentication request and returns the resulting Access Token or
+// an error in case of authentication failure.
+// ctx: The current request context
+// tenantID: The Azure Active Directory tenant (directory) Id of the service principal
+// clientID: The client (application) ID of the service principal
+// clientSecret: A client secret that was generated for the App Registration used to authenticate the client
+// scopes: The scopes required for the token
+func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, clientID string, clientSecret string, scopes []string) (*tokenResponse, error) {
 	msg, err := c.createClientSecretAuthRequest(tenantID, clientID, clientSecret, scopes)
 	if err != nil {
 		return nil, err
@@ -81,7 +113,7 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 		return c.createAccessToken(resp)
 	}
 
-	return nil, &AuthenticationFailedError{inner: newAuthenticationResponseError(resp)}
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
 }
 
 // AuthenticateCertificate creates a client certificate authentication request and returns an Access Token or
@@ -91,7 +123,7 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 // clientID: The client (application) ID of the service principal
 // clientCertificatePath: The path to the client certificate PEM file
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantID string, clientID string, clientCertificatePath string, scopes []string) (*azcore.AccessToken, error) {
+func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantID string, clientID string, clientCertificatePath string, scopes []string) (*tokenResponse, error) {
 	msg, err := c.createClientCertificateAuthRequest(tenantID, clientID, clientCertificatePath, scopes)
 	if err != nil {
 		return nil, err
@@ -111,14 +143,19 @@ func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantI
 		return c.createAccessToken(resp)
 	}
 
-	return nil, &AuthenticationFailedError{inner: newAuthenticationResponseError(resp)}
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
 }
 
-func (c *aadIdentityClient) createAccessToken(res *azcore.Response) (*azcore.AccessToken, error) {
-	value := &internalAccessToken{}
+func (c *aadIdentityClient) createAccessToken(res *azcore.Response) (*tokenResponse, error) {
+	value := struct {
+		Token        string      `json:"access_token"`
+		RefreshToken string      `json:"refresh_token"`
+		ExpiresIn    json.Number `json:"expires_in"`
+		ExpiresOn    string      `json:"expires_on"`
+	}{}
 	accessToken := &azcore.AccessToken{}
 	if err := json.Unmarshal(res.Payload, &value); err != nil {
-		return nil, fmt.Errorf("internalAccessToken: %w", err)
+		return nil, fmt.Errorf("internal AccessToken: %w", err)
 	}
 	t, err := value.ExpiresIn.Int64()
 	if err != nil {
@@ -127,7 +164,32 @@ func (c *aadIdentityClient) createAccessToken(res *azcore.Response) (*azcore.Acc
 	// NOTE: look at go-autorest
 	accessToken.Token = value.Token
 	accessToken.ExpiresOn = time.Now().Add(time.Second * time.Duration(t)).UTC()
-	return accessToken, nil
+	return &tokenResponse{token: accessToken, refreshToken: value.RefreshToken}, nil
+}
+
+func (c *aadIdentityClient) createRefreshTokenRequest(tenantID, clientID, clientSecret, refreshToken string, scopes []string) (*azcore.Request, error) {
+	urlStr := c.options.AuthorityHost.String() + tenantID + tokenEndpoint
+	urlFormat, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	data := url.Values{}
+	data.Set(qpGrantType, "refresh_token")
+	data.Set(qpClientID, clientID)
+	data.Set(qpClientSecret, clientSecret)
+	data.Set(qpRefreshToken, refreshToken)
+	data.Set(qpScope, strings.Join(scopes, " "))
+	dataEncoded := data.Encode()
+	body := azcore.NopCloser(strings.NewReader(dataEncoded))
+	msg := azcore.NewRequest(http.MethodPost, *urlFormat)
+	msg.Header.Set(azcore.HeaderContentType, azcore.HeaderURLEncoded)
+	err = msg.SetBody(body)
+	if err != nil {
+		return nil, err
+	}
+
+	return msg, nil
 }
 
 func (c *aadIdentityClient) createClientSecretAuthRequest(tenantID string, clientID string, clientSecret string, scopes []string) (*azcore.Request, error) {
@@ -191,7 +253,7 @@ func (c *aadIdentityClient) createClientCertificateAuthRequest(tenantID string, 
 // username: User's account username
 // password: User's account password
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticateUsernamePassword(ctx context.Context, tenantID string, clientID string, username string, password string, scopes []string) (*azcore.AccessToken, error) {
+func (c *aadIdentityClient) authenticateUsernamePassword(ctx context.Context, tenantID string, clientID string, username string, password string, scopes []string) (*tokenResponse, error) {
 	msg, err := c.createUsernamePasswordAuthRequest(tenantID, clientID, username, password, scopes)
 	if err != nil {
 		return nil, err
@@ -211,7 +273,7 @@ func (c *aadIdentityClient) authenticateUsernamePassword(ctx context.Context, te
 		return c.createAccessToken(resp)
 	}
 
-	return nil, &AuthenticationFailedError{inner: newAuthenticationResponseError(resp)}
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
 }
 
 func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, clientID string, username string, password string, scopes []string) (*azcore.Request, error) {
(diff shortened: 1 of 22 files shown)


sdk/azidentity/device_code_credential.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
+	deviceCodeGrantType = "urn:ietf:params:oauth:grant-type:device_code"
+)
+
+// DeviceCodeResult is used to store device code related information to help the user login and allow the device code flow to continue
+// to request a token to authenticate a user
+type DeviceCodeResult struct {
```
catalinaperalta: I'm thinking we don't need this struct, at least not now. At the moment we only need the login message for the user and  the interval is used for the delay in the loop. Or leave it with only the fields we use.

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,42 +19,45 @@ type ChainedTokenCredential struct {
 
 // NewChainedTokenCredential creates an instance of ChainedTokenCredential with the specified TokenCredential sources.
 func NewChainedTokenCredential(sources ...azcore.TokenCredential) (*ChainedTokenCredential, error) {
```
jhendrixMSFT: This is no-fail now so the `error` return value can be removed.

sdk/azidentity/chained_token_credential.go
```diff
@@ -18,42 +19,45 @@ type ChainedTokenCredential struct {
...
+				return nil, &AuthenticationFailedError{msg: "Received an AuthenticationFailedError, there is an invalid credential in the chain. " + createChainedErrorMessage(errList), inner: err}
+			}
+			return nil, fmt.Errorf("Received an unexpected error: %w", err) // if we receive some other error type this is unexpected and we simple return the unexpected error
 		} else {
-			return token, nil
+			return token, nil // if we did not receive an error then we return the token
 		}
 	}
 	// This condition should never be true
 	if token == nil && len(errList) == 0 {
```
jhendrixMSFT: Is this actually true now, that it can't happen?  Since we no longer ensure `len(sources) > 0` it seems you can now hit this case if `NewChainedTokenCredential()` is called without any parameters.

sdk/azidentity/aad_identity_client.go
```diff
@@ -41,8 +43,10 @@ const (
 // AADIdentityClient provides the base for authenticating with Client Secret Credentials, Client Certificate Credentials
 // and Environment Credentials. This type initializes a default azcore.Pipeline and IdentityClientOptions.
 type aadIdentityClient struct {
-	options  TokenCredentialOptions
-	pipeline azcore.Pipeline
+	lock         sync.RWMutex
+	options      TokenCredentialOptions
+	pipeline     azcore.Pipeline
+	refreshToken string
```
jhendrixMSFT: This field, and all the device code-related methods, should go on the `DeviceCodeCredential` type.  Then you won't need the lock.

sdk/azidentity/default_token_credential.go
```diff
@@ -41,15 +31,17 @@ func NewDefaultTokenCredential(options *DefaultTokenCredentialOptions) (*Chained
 	if !options.ExcludeEnvironmentCredential {
 		envCred, err := NewEnvironmentCredential(nil)
 		if err == nil {
-			cred.sources = append(cred.sources, envCred)
-			return NewChainedTokenCredential(cred.sources...)
+			creds = append(creds, envCred)
 		}
+		errMsg += "Make sure you have set the environment variables necessary for authentication: AZURE_TENANT_ID, AZURE_CLIENT_ID, AZURE_CLIENT_SECRET."
```
jhendrixMSFT: Is this error message redundant?  Seems like it should be returned from `NewEnvironmentCredential()`.

sdk/azidentity/aad_identity_client.go
jhendrixMSFT: Or perhaps some other internal struct, maybe even a `deviceCodeBearerAuthorizer`?

catalinaperalta: I agree if we only decide to use the refresh token for DeviceCodeCredential. I was looking into what information is returned when requesting a token for the other credentials and several other credentials also return a refresh token. I can make a list of the information returned by the other credential types, then we can all discuss what to do with the information that's returned.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Very right, thanks for catching that! I'm removing it.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Also true, I'm updating the comment to explain why this could happen now

sdk/azidentity/default_token_credential.go
catalinaperalta: Yes, I agree I'll simply append err.Error() to errMsg

sdk/azidentity/aad_identity_client.go
schaabs: The refresh token will be specific to each request possibly as it's specific to the user which logs in using the device code, hence it should definitely be stored on the `DeviceCodeCredential` and not on the client which can presumably handle request for many different credential instances.

sdk/azidentity/chained_token_credential.go
```diff
@@ -17,43 +18,46 @@ type ChainedTokenCredential struct {
...
+			return nil, fmt.Errorf("Received an unexpected error: %w", err) // if we receive some other error type this is unexpected and we simple return the unexpected error
 		} else {
-			return token, nil
+			return token, nil // if we did not receive an error then we return the token
 		}
 	}
-	// This condition should never be true
+	// this situation can occur when the length of sources is 0
 	if token == nil && len(errList) == 0 {
 		return nil, nil
```
schaabs: Perhaps this should return an error that no token was returned rather than simply also returning nil? If we think that's the case you could modify `createChainedErrorMessage` to return a special message when errList is empty, then you could just remove this condition entirely.

jhendrixMSFT: Agreed an error should be returned here.

sdk/azidentity/aad_identity_client.go
schaabs: The refresh token will be specific to each request possibly as it's specific to the user which logs in using the device code, hence it should definitely be stored on the DeviceCodeCredential and not on the client which can presumably handle request for many different credential instances.

sdk/azidentity/chained_token_credential.go
catalinaperalta: I reverted the changes so I also reverted the comment. This condition should never be true now

catalinaperalta:
> Can you rebase this onto track2? All the extraneous commits make it hard to review.

Yes I'm on it.

sdk/azidentity/aad_identity_client.go
```diff
@@ -61,7 +63,37 @@ func newAADIdentityClient(options *TokenCredentialOptions) *aadIdentityClient {
...
+		return nil, err
+	}
+
+	resp, err := c.pipeline.Do(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	// This should never happen under normal conditions
+	if resp == nil {
```
jhendrixMSFT: I think we should remove this block.  If `Do()` ever returns a nil error and a nil response it's a bug and we should fix it.

sdk/azidentity/aad_identity_client.go
```diff
@@ -61,7 +63,37 @@ func newAADIdentityClient(options *TokenCredentialOptions) *aadIdentityClient {
...
+		return nil, &AuthenticationFailedError{msg: "Something unexpected happened with the request and received a nil response"}
+	}
+
+	if resp.HasStatusCode(successStatusCodes[:]...) {
+		return c.createAccessToken(resp)
+	}
+
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
```
jhendrixMSFT: I don't think we should wrap the error here.  Instead, put the authentication failure info directly in the `AuthenticationFailedError`.

sdk/azidentity/aad_identity_client.go
```diff
@@ -81,7 +113,7 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 		return c.createAccessToken(resp)
 	}
 
-	return nil, &AuthenticationFailedError{inner: newAuthenticationResponseError(resp)}
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
```
jhendrixMSFT: Avoid wrapping.

sdk/azidentity/aad_identity_client.go
```diff
@@ -111,14 +143,19 @@ func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantI
...
```
jhendrixMSFT: Avoid wrapping.

sdk/azidentity/aad_identity_client.go
```diff
@@ -111,14 +143,19 @@ func (c *aadIdentityClient) authenticateCertificate(ctx context.Context, tenantI
...
+		ExpiresIn    json.Number `json:"expires_in"`
+		ExpiresOn    string      `json:"expires_on"`
+	}{}
 	accessToken := &azcore.AccessToken{}
 	if err := json.Unmarshal(res.Payload, &value); err != nil {
-		return nil, fmt.Errorf("internalAccessToken: %w", err)
+		return nil, fmt.Errorf("internal AccessToken: %w", err)
 	}
 	t, err := value.ExpiresIn.Int64()
 	if err != nil {
```
jhendrixMSFT: I would wrap this error with some context so we know what failed (similar to the above wrapping).

sdk/azidentity/aad_identity_client.go
```diff
@@ -127,7 +164,32 @@ func (c *aadIdentityClient) createAccessToken(res *azcore.Response) (*azcore.Acc
 	// NOTE: look at go-autorest
 	accessToken.Token = value.Token
 	accessToken.ExpiresOn = time.Now().Add(time.Second * time.Duration(t)).UTC()
-	return accessToken, nil
+	return &tokenResponse{token: accessToken, refreshToken: value.RefreshToken}, nil
+}
+
+func (c *aadIdentityClient) createRefreshTokenRequest(tenantID, clientID, clientSecret, refreshToken string, scopes []string) (*azcore.Request, error) {
+	urlStr := c.options.AuthorityHost.String() + tenantID + tokenEndpoint
```
jhendrixMSFT: Does this properly handle the case with a custom `AuthorityHost` that doesn't end with a trailing /?  You might want to use `ResolveReference()`, see https://stackoverflow.com/questions/34668012/combine-url-paths-with-path-join

sdk/azidentity/aad_identity_client.go
```diff
@@ -127,7 +164,32 @@ func (c *aadIdentityClient) createAccessToken(res *azcore.Response) (*azcore.Acc
...
+	urlFormat, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	data := url.Values{}
+	data.Set(qpGrantType, "refresh_token")
+	data.Set(qpClientID, clientID)
+	data.Set(qpClientSecret, clientSecret)
```
jhendrixMSFT: @schaabs Is client secret required in all cases?  I see [this](https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-protocols-oauth-code#refreshing-the-access-tokens) and it says only for web apps.

sdk/azidentity/aad_identity_client.go
```diff
@@ -211,7 +273,7 @@ func (c *aadIdentityClient) authenticateUsernamePassword(ctx context.Context, te
 		return c.createAccessToken(resp)
 	}
 
-	return nil, &AuthenticationFailedError{inner: newAuthenticationResponseError(resp)}
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
```
jhendrixMSFT: Avoid wrapping.

sdk/azidentity/device_code_credential.go
```diff
@@ -0,0 +1,100 @@
...
```
jhendrixMSFT: Does this actually need to be exported?  Isn't all of this implementation details for how device code polling works?

sdk/azidentity/aad_identity_client.go
```diff
@@ -238,6 +300,120 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
...
+		return nil, err
+	}
+
+	resp, err := c.pipeline.Do(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	// This should never happen under normal conditions
+	if resp == nil {
```
jhendrixMSFT: Remove block.

sdk/azidentity/aad_identity_client.go
```diff
@@ -238,6 +300,120 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
...
+		return nil, &AuthenticationFailedError{msg: "Something unexpected happened with the request and received a nil response"}
+	}
+
+	if resp.HasStatusCode(successStatusCodes[:]...) {
+		return c.createAccessToken(resp)
+	}
+
+	return nil, &AuthenticationFailedError{inner: newAADAuthenticationFailedError(resp)}
```
jhendrixMSFT: Avoid wrapping.

sdk/azidentity/aad_identity_client.go
```diff
@@ -238,6 +300,120 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
...
+}
+
+func (c *aadIdentityClient) createDeviceCodeAuthRequest(tenantID string, clientID string, deviceCode string, scopes []string) (*azcore.Request, error) {
+	if len(tenantID) == 0 { // if the user did not pass in a tenantID then the default value is set
+		tenantID = "organizations"
```
jhendrixMSFT: Does this need to be documented in the exported API that calls this method?  I'm not saying it has to, but it seems magical.  If this is the normal expectation then no need to doc IMO but I'll defer to @schaabs

sdk/azidentity/aad_identity_client.go
```diff
@@ -238,6 +300,120 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
...
+	if err != nil {
+		return nil, err
+	}
+	data := url.Values{}
+	data.Set(qpGrantType, deviceCodeGrantType)
+	data.Set(qpClientID, clientID)
+	data.Set(qpDeviceCode, deviceCode)
+	scope := strings.Join(scopes, " ")
+	if !strings.Contains(scope, "offline_access") { // offline_access scope is set for device code credential so that a refresh token will be returned so that future token requests can be carried out sliently
+		scope = scope + " offline_access"
```
jhendrixMSFT: Another magical thing to doc?

sdk/azidentity/aad_identity_client.go
```diff
@@ -238,6 +300,120 @@ func (c *aadIdentityClient) createUsernamePasswordAuthRequest(tenantID string, c
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.pipeline.Do(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+	// This should never happen under normal conditions
+	if resp == nil {
```
jhendrixMSFT: Remove block.

sdk/azidentity/aad_identity_client.go
jhendrixMSFT: I would add this to the scopes slice in `DeviceCodeCredential.GetToken()` so this only has to be done in one place (I see it was duplicated elsewhere).

sdk/azidentity/chained_token_credential.go
jhendrixMSFT: Can this block just be removed then?

sdk/azidentity/device_code_credential.go
```diff
@@ -0,0 +1,100 @@
...
+// GetToken obtains a token from the Azure Active Directory service, following the device code authentication
+// flow. This function first requests a device code and requests that the user login to continue authenticating.
+// This function will keep polling the service for a token meanwhile the user logs.
+// - scopes: The list of scopes for which the token will have access.
+// - ctx: controlling the request lifetime.
+// Returns an AccessToken which can be used to authenticate service client calls.
+func (c *DeviceCodeCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	if len(c.refreshToken) != 0 {
+		tk, err := c.client.refreshAccessToken(ctx, c.tenantID, c.clientID, "", c.refreshToken, opts.Scopes)
+		if tk == nil { // this would only happen if there was an unmarshal error or a failure to parse ExpiresOn
```
jhendrixMSFT: This should be `err != nil`, we should only predicate failure on a non-nil error.

sdk/azidentity/device_code_credential.go
```diff
@@ -0,0 +1,100 @@
...
+		if !errors.As(err, &authRespErr) {
+			return nil, err
+		}
+		switch authRespErr.Message {
+		// wait for the interval specified from the initial device code endpoint and then poll for the token again
+		case "authorization_pending":
+			time.Sleep(time.Duration(dc.Interval) * time.Second)
+		default:
+			// any other error should be returned
+			return nil, err
```
jhendrixMSFT: Shouldn't this return an `AuthenticationFailedError` to be consistent with the other credential types?

sdk/azidentity/managed_identity_client.go
```diff
@@ -226,31 +226,31 @@ func (c *managedIdentityClient) createCloudShellAuthRequest(clientID string, sco
...
 			c.endpoint = endpoint
 			if secretEnvVar := os.Getenv(msiSecretEnvironemntVariable); secretEnvVar != "" { // if BOTH the env vars MSI_ENDPOINT and MSI_SECRET are set the MsiType is AppService
-				c.msiType = appService
+				c.msiType = msiTypeAppService
 			} else { // if ONLY the env var MSI_ENDPOINT is set the MsiType is CloudShell
-				c.msiType = cloudShell
+				c.msiType = msiTypeCloudShell
 			}
-		} else if c.imdsAvailable(ctx) { // if MSI_ENDPOINT is NOT set AND the IMDS endpoint is available the MsiType is Imds
+		} else if c.imdsAvailable(ctx) { // if MSI_ENDPOINT is NOT set AND the IMDS endpoint is available the MsiType is Imds. This will timeout after 500 milliseconds
```
jhendrixMSFT: Weren't we going to change this to do it during MSI credential construction?  It can happen in a separate PR.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Are you thinking we should remove this everywhere? (i.e. I also check for this condition for client secret credentials, certificate credentials, etc) I'm not opposed to removing this from all of them.

catalinaperalta: I removed it in all instances, since this should actually never come to pass.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I read that as well. This code  is going through the general refresh token flow. I believe it should be alright to send an empty value for client ID. I tested this with device code which does not send a client id and had no problems. However, I could simply add an if statement to check if clientID is empty. I'll wait to see what Scott says about this.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I think it should be documented regardless of whether it is normally expected, simply to add clarity for users who haven't delved into the specifics of the flow they are using. Thanks for pointing this out. I'll add this to the info for the exported API.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Yes

sdk/azidentity/device_code_credential.go
catalinaperalta: Yes, that's what my original comment said. I don't think it needs to be exported.

sdk/azidentity/managed_identity_client.go
catalinaperalta: Yup, I'm waiting for this PR to be completed. I have this change ready to go in a local branch.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I'm not sure I understood what you mean. You want me to wrap this unmarshal error in an error type??

sdk/azidentity/device_code_credential.go
catalinaperalta: Ok, got it 👍

sdk/azidentity/device_code_credential.go
catalinaperalta: Yes

sdk/azidentity/aad_identity_client.go
jhendrixMSFT: On line 144 we wrap the error from `json.Unmarshal()` with an "internal AccessToken" message.  I think we should do the same thing here.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Jeff says: "AuthenticationFailedError is authentication mechanism agnostic and should therefore not have fields that are specific to AAD. The AADAuthenticationFailedError has those fields and then is wrapped with AuthenticationFailedError to indicate the the authentication failed. If recovery code wants to find out why the authentication failed it would have to look specifically for an AAD failure and if found, then the fields correspond to that."

sdk/azidentity/aad_identity_client.go
catalinaperalta: Will add an if statement

sdk/azidentity/aad_identity_client.go
jhendrixMSFT: Agreed that AAD-specific fields shouldn't be in AuthenticationFailedError.  However with the code the way it is at present, if this API call fails and you `panic(err)` or `fmt.Printf("%v", err)` you get zero information (it literally prints an empty string) because the message is empty, so the failure is not actionable.  At the very least we should have `Error()` include the error message from `inner.Error()` to avoid this.
If the inner error will contain more useful info beyond a message/description then agreed that keeping them distinct is valuable.  I just want to make sure we aren't adding a needless level of indirection.

sdk/azidentity/device_code_credential.go
catalinaperalta: Update: this is actually how it should be because the wrapping is carried out within authenticateDeviceCode. The errors that could be returned are json unmarshal errors, pipeline errors OR AuthenticationFailedErrors. So, when this is in fact an AuthenticationFailedError the wrapping will be there. See line: 
https://github.com/Azure/azure-sdk-for-go/pull/6560/files#diff-a7c4c602cbc8f32f6e9e22430ee205e1R316

sdk/azidentity/aad_identity_client.go
catalinaperalta: Leaving wrapping for now. If you want I can open an issue to discuss this.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Leaving wrapping for now.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Leaving wrapping for now.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Leaving wrapping for now.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I agree I'll add the inner error description to the message

catalinaperalta: Added the inner error description to AuthenticationFailedError: 
https://github.com/Azure/azure-sdk-for-go/pull/6560/files#diff-d1256cf2ef15d119cac7ccfaf6931cc9R79-R81

sdk/azidentity/aad_identity_client.go
```diff
@@ -155,11 +227,12 @@ func (c *aadIdentityClient) createClientSecretAuthRequest(tenantID string, clien
 }
 
 func (c *aadIdentityClient) createClientCertificateAuthRequest(tenantID string, clientID string, clientCertificate string, scopes []string) (*azcore.Request, error) {
-	urlStr := c.options.AuthorityHost.String() + tenantID + tokenEndpoint
+	urlStr := c.options.AuthorityHost.String()
```
jhendrixMSFT: Can this be removed now?

sdk/azidentity/azidentity.go
```diff
@@ -34,18 +35,13 @@ func init() {
...
-	TokenType    string      `json:"token_type"`
+type tokenResponse struct {
+	token        *azcore.AccessToken
+	refreshToken string
 }
 
-// AuthenticationResponseError is a struct used to marshal responses when authentication has failed
-type AuthenticationResponseError struct {
+// AADAuthenticationFailedError is a struct used to marshal responses when authentication has failed
+type AADAuthenticationFailedError struct {
```
jhendrixMSFT: Does this actually need to be exported?  If we don't expect consumers to `errors.As()` on this then it can be left internal.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Yes, I'll update the code wherever this happens

sdk/azidentity/managed_identity_credential.go
```diff
@@ -54,5 +54,6 @@ func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts azcore.To
 
 // AuthenticationPolicy implements the azcore.Credential interface on ManagedIdentityCredential.
 func (c *ManagedIdentityCredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	options.Options.Scopes = []string{scopesToResource(options.Options.Scopes[0])}
```
jhendrixMSFT: Why does the `/.default` need to be removed from the scope?  If this is the only place it happens I suggest doing that here inline and remove the function.  Also add a comment why this is necessary.

sdk/azidentity/device_code_credential.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// GetToken obtains a token from the Azure Active Directory service, following the device code authentication
+// flow. This function first requests a device code and requests that the user login to continue authenticating.
+// This function will keep polling the service for a token meanwhile the user logs.
+// scopes: The list of scopes for which the token will have access. The "offline_access" scope is checked for and automatically added in case it isn't present to allow for silent token refresh.
+// ctx: controlling the request lifetime.
+// Returns an AccessToken which can be used to authenticate service client calls.
+func (c *DeviceCodeCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	for i, scope := range opts.Scopes {
```
jhendrixMSFT: This block can be removed now yes?

sdk/azidentity/aad_identity_client.go
catalinaperalta: Ok, so there is one problem with changing this to simply append information to the path of the request. Any subsequent calls to create a new request will also try to append information to the path of the AuthorityHost (that means we are changing the default value of the authority host on the client, I dont think that is the best idea). However, if we dont mind changing the path on the authority host then I could add a check to compare the current path in the client to an expected path.

sdk/azidentity/device_code_credential.go
jhendrixMSFT: Oh I thought this had been moved.  It probably makes more sense to do this in the constructor so that it only happens once and not on every call to `GetToken()`.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I'm leaning towards leaving it as it was, that is with a copy of the path of the authority host

sdk/azidentity/device_code_credential.go
catalinaperalta: We dont know the scopes in the constructor though.

jhendrixMSFT: Sigh yes you're right (I hate this scopes problem).

sdk/azidentity/aad_identity_client.go
catalinaperalta: I think it would make the code a bit clunky to add some sort of check like: 
if !HasSuffix(c.options.AuthorityHost.Path, "oauth/etc../token") { 
    then append
}

I think creating a copy of the default authority host url is better, because we are also not changing the default value that we use for reference for all requests related to each credential.

jhendrixMSFT: Will the value ever change?  If not can we set `AuthorityHost` to the correct value in the constructor?

catalinaperalta: I think that can work for most credentials (although, in the future, if we decide to export tenantID or some other relevant field that affects the building of the URL then it wont work, but for now that is not a problem), but Device Code uses two different endpoints, so that one definitely cannot be created in the constructor.

sdk/azidentity/azidentity.go
catalinaperalta: Leaving exported for now because AADAuthenticationFailedError exposes relevant fields for identity related errors

sdk/azidentity/managed_identity_credential.go
catalinaperalta: It does, this credential expects a resource string and not a scope. So I'll just put the code directly in the authpolicy func

jhendrixMSFT: Is this specific to IMDS, i.e. does it use a slice of resources and not a slice of scopes?  Wondering if we should have a separate field for slice of resources.  At the very least, if this is unique behavior for IMDS and we want to use the `Scopes` slice as a resources slice we should comment that.

catalinaperalta: It will expect a slice of resource strings that must match the URI exactly as it is specified in Azure AD. 
An example request from the docs: $response = Invoke-WebRequest -Uri 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/' -Method GET -Headers @{Metadata="true"}

And the note that went with it: 
The value of the "resource" parameter must be an exact match for what is expected by Azure AD. When using the Azure Resource Manager resource ID, you must include the trailing slash on the URI.

I could document in the credential's info that this credential expects a slice of resources instead of a slice of scopes

Also, if we were to add a specific field for resources it would be on TokenRequestOptions which is in AuthenticationPolicyOptions

catalinaperalta: Here are examples of how to send MSI requests in each environment: 
IMDS: curl 'http://169.254.169.254/metadata/identity/oauth2/token?resource=https://vault.azure.net&api-version=2018-02-01' -H "Metadata: true"

AppService: curl "$MSI_ENDPOINT?resource=https://vault.azure.net&api-version=2017-09-01" -H "secret: $MSI_SECRET"

CloudShell: curl -H 'Metadata: true' -d 'resource=https://vault.azure.net' $MSI_ENDPOINT

PR #6576, opened 2019-12-24
v38.0.0

ArcturusZhang:
 

ArcturusZhang:
Hi @jhendrixMSFT I have two codegen related questions about this major version release code.
1. It seems that the APIVersion change is taking effect. But the position of APIVersion seems unstable, is there any solution that we could stabilize the location of APIVersion parameter in the parameter list?
You can take [this](https://github.com/Azure/azure-sdk-for-go/blob/5bf1033a8a3896cc91433181d03e6c1f3b60713d/services/preview/managedservices/mgmt/2018-06-01/managedservices/managedservicesapi/interfaces.go#L27) as an example.
2. What kind of service will have an APIVersion string as their parameters? Is there a rule of this?

ArcturusZhang:
The third question is that do we have default values for those APIVersion parameters? Or the customer must explicitly assign one? This may cause some inconvenience when using the SDK.

jhendrixMSFT:
I looked at the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/managedservices/resource-manager/Microsoft.ManagedServices/preview/2018-06-01-preview/managedservices.json), the ordering has this inconsistency there too.  While we could enforce a certain ordering I'm hesitant to do so (maybe it's ok, I haven't thought too deeply about it).
I don't think there's a rule, it up to how the authors wrote the swagger.
I don't know anything about this API, so I don't know if having some kind of default value makes sense.

ArcturusZhang:
> I looked at the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/managedservices/resource-manager/Microsoft.ManagedServices/preview/2018-06-01-preview/managedservices.json), the ordering has this inconsistency there too. While we could enforce a certain ordering I'm hesitant to do so (maybe it's ok, I haven't thought too deeply about it).
> I don't think there's a rule, it up to how the authors wrote the swagger.
> I don't know anything about this API, so I don't know if having some kind of default value makes sense.

So is API version must add to this method, is because of the [API version parameter in swagger]() has a `"x-ms-parameter-location": "method",` attribute?
I think we can release this major version in this way. This parameter ordering problem is not the first time (last time I know is v37.0.0)
Or if we setup a rule for how those parameter orders it will be good, i think, to keep those different methods consistent.

Issue #6586, opened 2019-12-27
There is a bug with typecasting in Azure Cognitive part


thesunwave:
### Bug Report

- import path `.../services/cognitiveservices/v1.0/face"`
- SDK version: `34.0.0`
- Go Version: `go1.13.1 darwin/amd64`

#### What happened?
When I try to call AddPhoto method of Cognitive through a picture that includes more than 1 face remote service returns me an error with description like "there is more 1 face", after that I call the Detect method to get coordinates of required face and then send it again, but AddFace answers me that coordinates are incorrect.

Instead of sending correct coordinates it sends value like this `<int32 Value> <int32 Value> <int32 Value> <int32 Value>`

#### What did you expect or want to happen?
I send coordinates and picture then receive a correct answer with persisted_face_id
#### How can we reproduce it?
Use `face.LargeFaceListClient` and call `AddFaceFromURL` set the valid targetFace with `[]int32{}` and send a request to a mock server to have an opportunity to see the result of that request.

#### Btw, I've found the reason. 

Here we are calling `autorest.Encode("query", targetFace, ",")` and this is the reason of that behavior.
https://github.com/Azure/azure-sdk-for-go/blob/781d9fb592a1818cb7fb8e0855ea9c00b095a9e9/services/cognitiveservices/v1.0/face/largefacelist.go#L288-L290

As we know, `targetFace` is a slice of `int32`, but `autorest.Encode` under the hood wants to get a slice of strings or some types derived from string.
Look: 

https://github.com/Azure/go-autorest/blob/b965d219bd9d10676384ef8f05da2e2441df9829/autorest/utility.go#L143-L157

So, now I don't really understand where we have to fix it. Either in `go-autorest` or here?

ArcturusZhang:
Hi @thesunwave thanks for opening this issue!
I do not think this is a code gen problem, more of the probability is the service side bug. 
When I am checking the [swagger definition](https://github.com/Azure/azure-rest-api-specs/blob/4dc4ff6b562d666782e11b66aaa5fefe3ebef399/specification/cognitiveservices/data-plane/Face/stable/v1.0/Face.json#L4112) for this service, the `targetFace` is defined as:
```
    "targetFace": {
      "name": "targetFace",
      "in": "query",
[...]

PR #6730, opened 2020-01-08
Change type from int to string to set a valid data for autorest #6586

thesunwave:
- [] The PR targets the `latest` branch.
 
Fix for this issue #6586. It can be resolved like that or will be fixed in the autorest package


jhendrixMSFT:
This package, and the majority of the SDK, is generated from swagger files.  We cannot make hand-edits as they will be lost when the SDK is regenerated.
Looking at the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cognitiveservices/data-plane/Face/stable/v1.0/Face.json#L4160) it's properly defined as a string so this appears to be a codegen issue.  Can you please open an issue to track getting this fixed?

thesunwave:
Hi @jhendrixMSFT, I'm not sure I understood right what you mean. I mean that issue is already created https://github.com/Azure/azure-sdk-for-go/issues/6586

jhendrixMSFT:
Oh I see, I missed the issue in the original comment, sorry for the confusion.

PR #6884, opened 2020-01-14
Update ManagedIdentityCredential constructor

catalinaperalta:
Updated ManagedIdentityCredential constructor to check for MSI environment upon creation. Also updated tests and other references to Managed Identity Credential

### DIFF
--- a/sdk/azidentity/managed_identity_credential.go
+++ b/sdk/azidentity/managed_identity_credential.go
@@ -6,6 +6,7 @@ package azidentity
 import (
 	"context"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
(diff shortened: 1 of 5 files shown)


sdk/azidentity/managed_identity_credential.go
```diff
@@ -44,8 +45,20 @@ type ManagedIdentityCredential struct {
 // clientID: The client id to authenticate for a user assigned managed identity.  More information on user assigned managed identities cam be found here:
 // https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#how-a-user-assigned-managed-identity-works-with-an-azure-vm
 // options: Options that allow to configure the management of the requests sent to the Azure Active Directory service.
-func NewManagedIdentityCredential(clientID string, options *ManagedIdentityCredentialOptions) *ManagedIdentityCredential {
-	return &ManagedIdentityCredential{clientID: clientID, client: newManagedIdentityClient(options)}
+func NewManagedIdentityCredential(clientID string, options *ManagedIdentityCredentialOptions) (*ManagedIdentityCredential, error) {
+	// Create a new Managed Identity Client with default options
+	client := newManagedIdentityClient(options)
+	// Create a context that will timeout after 500 milliseconds (that is the amount of time designated to find out if the IMDS endpoint is available)
+	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Duration(client.imdsAvailableTimeoutMS)*time.Millisecond)
```
jhendrixMSFT: Any reason why `imdsAvailableTimeoutMS` is a field on the client and not a constant value?  It doesn't appear to be configurable.

catalinaperalta: Honestly, I had it as a constant at some point but cant remember why I moved it over to the client. At the moment I'm willing to make it a constant again. In the meantime I'm going to check out other implementations.

catalinaperalta: I'm not sure if timeout should be configurable at some point: https://github.com/Azure/azure-sdk-for-js/blob/db333ca8a763a27c425353e8b60237f5571fd532/sdk/identity/identity/src/credentials/managedIdentityCredential.ts#L182

jhendrixMSFT: We can leave it as-is for now.

PR #6969, opened 2020-01-15
Add support for marshalling/unmarshalling JSON

jhendrixMSFT:
Removed Response.Payload field, replacing it with an internal
implementation nopClosingBytesReader.

### DIFF
--- a/sdk/azcore/response.go
+++ b/sdk/azcore/response.go
@@ -7,6 +7,7 @@ package azcore
 
 import (
 	"bytes"
+	"encoding/json"
 	"encoding/xml"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


sdk/azcore/response.go
```diff
@@ -64,8 +64,9 @@ func (r *Response) HasStatusCode(statusCodes ...int) bool {
 // UnmarshalAsJSON calls json.Unmarshal() to unmarshal the received payload into the value pointed to by v.
 // If no payload was received a RequestError is returned.  If json.Unmarshal fails a UnmarshalError is returned.
 func (r *Response) UnmarshalAsJSON(v interface{}) error {
+	// TODO: verify early exit is correct
 	if len(r.payload()) == 0 {
-		return newRequestError("missing payload", r)
+		return nil
```
catalinaperalta: Just a question, to understand the reason behind returning nil, wouldn't it be better to return an error is the payload empty? Since clearly calling the function means that there was the expectation that something would be unmarshaled?

jhendrixMSFT: That was my original assumption, however the `byte_getNull` test expects this to be valid.  I actually think the swagger is malformed, i.e. the MIME type should be `application/text` and not `application/json` as none of the models are actually JSON.  We also have this same behavior in track 1, but again it might be predicated on some bad assumptions (i.e. is there ever a valid success case where the swagger specifies a JSON model but the service is allowed to return null).

PR #7023, opened 2020-01-17
Add CliCredential and it's unit test

wantedfast:
Follow PR: https://github.com/Azure/azure-sdk-for-python/pull/8962
Add Azure Cli credential into azidentity and add it as part of default_token_credential.
Don't include any generated files.
This PR would target track2 branch.
unit tests are included.
No any packages changed, so it don't include the CHANGELOG.md
 Included


tzhanl:
@jongio

sdk/azidentity/cli_credential_client.go
```diff
@@ -0,0 +1,142 @@
...
+// CliCredentialClient provides the base for authenticating with Cli Credential.
+type CliCredentialClient struct {
+}
+
+// NewCliCredentialClient creates a new instance of the CliCredentialClient.
+func NewCliCredentialClient() *CliCredentialClient {
+	return &CliCredentialClient{}
+}
+
+// Authenticate runs a Azure Cli command for a Azure CLi Credential and returns the resulting Access Token or
```
jongio: Please use consistent casing for Cli

sdk/azidentity/cli_credential_client.go
```diff
@@ -0,0 +1,142 @@
...
+// an error in case of authentication failure.
+// ctx: The current request context
+// scopes: The scopes required for the token
+func (c *CliCredentialClient) authenticate(ctx context.Context, scopes []string, credentialClient ShellClient) (*azcore.AccessToken, error) {
+
+	// Validate the resource to make sure it don't include shell-meta characters since it gets sent as a command line argument to Azure CLI
```
jongio: Change don't to doesn't

sdk/azidentity/cli_credential_client.go
```diff
@@ -0,0 +1,142 @@
...
+func (c *CliCredentialClient) createAccessToken(output []byte) (*azcore.AccessToken, error) {
+	token := cliAccessToken{}
+	accessToken := &azcore.AccessToken{}
+
+	err := json.Unmarshal(output, &token)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse experisOn string to date as required time format
```
jongio: expiresOn spelling

sdk/azidentity/cli_credential_client.go
```diff
@@ -0,0 +1,142 @@
...
+	dateString := token.ExpiresOn
+	timeformat := "2006-01-02 15:04:05.999999"
+	experisOnValue, err := time.ParseInLocation(timeformat, dateString, time.Local)
```
jongio: expiresOnValue spelling.

sdk/azidentity/cli_credential_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+type winAzureCliNotInstalledMock struct {
+}
+
+func (c *winAzureCliNotInstalledMock) getCliAccessToken(command string) ([]byte, string, error) {
+	return nil, "'az' is not recognized", errors.New("mockError")
+}
+
+func TestCliCredential_WinAzureCLINotInstalled(t *testing.T) {
```
jongio: Cli casing

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,47 @@
...
+package azidentity
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// Enables authentication to Azure Active Directory using Azure cli to generated an access token.
+type AzureCliCredential struct {
```
catalinaperalta: This should be AzureCLICredential. CLI should be in all caps wherever it appear to follow convention.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,47 @@
...
+	client      *AzureCliCredentialClient
+	shellClient ShellClient
+}
+
+// AzureCliCredentialOption contains parameter that Shell execute Azure Cli command to return a json output what include accessToken and expiresOn.
+type AzureCliCredentialOption struct {
```
catalinaperalta: AzureCLICredentialOptions

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,161 @@
...
+	Tenant       string `json:"tenant"`
+	TokenType    string `json:"tokenType"`
+}
+
+type ShellClient interface {
+	getAzureCliAccessToken(command string) ([]byte, string, error)
+}
+
+// AzureCliCredentialClient provides the base for authenticating with Azure Cli Credential.
+type AzureCliCredentialClient struct {
```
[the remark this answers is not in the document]
catalinaperalta: I'm wondering whether this should have a pipeline to handle the token request and whether this should simple exist in aadIdentityClient. @jhendrixMSFT what do you think?

UPDATE: looking at the code I don't think this particular credential needs a pipeline since it simply executes a shell command

sdk/azidentity/azure_cli_credential.go
jongio: Let’s hold on changing this until we get consensus across Langs as it is Cli across all right now.  Thanks.

catalinaperalta: This is a Golang specific requirement. It's idiomatic to the language to use common casing when using acronyms. In this, we differ from the rest of the languages, since the Gopher community and the linter look for this. The current go linter already complains for common acronyms such as ID, which is why we have tenantID and not tenantId like other languages.

jongio: Gotcha.  Thanks for the explanation.  @tzhanl, @wantedfast this if for Go only, change to CLI.  Thank you.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// Enables authentication to Azure Active Directory using Azure CLI to generated an access token.
+type AzureCLICredential struct {
+	Client      *azureCLICredentialClient
```
catalinaperalta: unexport

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+package azidentity
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// Enables authentication to Azure Active Directory using Azure CLI to generated an access token.
+type AzureCLICredential struct {
```
catalinaperalta: Rename this to DeveloperCredential (we may change this later)

schaabs: Don't rename this right now. Let's get everything checked in as `AzureCliCredential` and then we can lay `DeveloperCredential` on top of this afterwards.

schaabs: Also in all other languages as well as in the cli itself the casing Cli is used vs. CLI. Is there a go specific reason to not follow this convention?

catalinaperalta: Copying the same answer I gave Jon: This is a Golang specific requirement. It's idiomatic to the language to use common casing when using acronyms. In this, we differ from the rest of the languages, since the Gopher community and the linter look for this. The current go linter already complains for common acronyms such as ID, which is why we have tenantID and not tenantId like other languages.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+	Client      *azureCLICredentialClient
+	shellClient shellClient
+}
+
+// AzureCLICredentialOptions contains the parameters for the ShellClient to execute the Azure CLI command that will return an Access Token.
+type AzureCLICredentialOption struct {
```
JeffreyRichter: *Options (add 's' suffix)

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+	shellClientOption shellClient
```
JeffreyRichter: First, *Option should have an 's' suffix.
Second, if the type is an *Options, then why does the field name refer to a client and not to an options? I think the field name should be shellClientOptions.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// NewAzureCLICredential creates an instance of AzureCLICredential to authenticate against Azure Active Directory with Azure CLI Credential's token.
+func NewAzureCLICredential(options *AzureCLICredentialOption) *AzureCLICredential {
+	var Client = newAzureCLICredentialClient()
```
JeffreyRichter: local vars should start with lowercase. Plus, this code is very confusing to me. It creates a *Client which is than passed to create an *Options and then you grab just 1 field (shellClientOption) out of the options struct? AND, on top of this, the client that is pass to set the default options is then ignored in the returned options but used to initialize the AzireCLICredential's Client field?

In addition, the way that default options are being initialized/obtained is not in line with our guidelines.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,48 @@
...
+func (c *AzureCLICredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	return c.Client.authenticate(ctx, opts.Scopes, c.shellClient)
+}
+
+// AuthenticationPolicy implements the azcore.Credential interface on AzureCLICredential.
+func (c *AzureCLICredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	return newBearerTokenPolicy(c, options)
+}
+
+func (c *AzureCLICredentialOption) setDefaultCLIOptions(CLIClient *azureCLICredentialClient) *AzureCLICredentialOption {
```
JeffreyRichter: This is confusing. What is an option setting and what is a client?
Also, this doesn't follow our guidelines.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	"regexp"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
+	DefaultSuffix        = "/.default"
```
JeffreyRichter: Why are these public? Is a customer expected to refer to these strings? If so, why? It feels to me that these would be strings in an error value. Maybe we want public sentinel error values instead? Since there are no comments, I can't discern intent. Please also add comments so that other devs know what you're trying to accomplish here.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	AzureCLINotInstalled = "Azure CLI not installed"
+	AzNotLogIn           = "ERROR: Please run 'az login'"
+	WinAzureCLIError     = "'az' is not recognized"
+	invalidResourceError = "Resource is not in expected format. Only alphanumeric characters, '.', ':', '-', and '/' are allowed"
+)
+
+type azureCLIAccessToken struct {
+	Token        string `json:"accessToken"`
```
JeffreyRichter: Let's make the fields private too so that readers of the code know that the fields are intended not to be public.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	ExpiresOn    string `json:"expiresOn"`
+	Subscription string `json:"subscription"`
+	Tenant       string `json:"tenant"`
+	TokenType    string `json:"tokenType"`
+}
+
+type shellClient interface {
```
JeffreyRichter: What is the intent of this interface? What other types might implement it?

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	getAzureCLIAccessToken(command string) ([]byte, string, error)
+}
+
+// AzureCLICredentialClient provides the base for authenticating with Azure CLI Credential.
+type azureCLICredentialClient struct {
```
JeffreyRichter: The type is not public but the comment refers to a public type. Also, what does it mean to be a "base" since Go does not support inheritance. Is this type required for some reason?

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// NewAzureCLICredentialClient creates a new instance of the AzureCLICredentialClient.
```
JeffreyRichter: Again, cases don't match. I'm even surprised that this compiles cleanly because the compiler checks case with comments and types.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		command = command + " --resource " + resource
+	}
+	// Execute Azure CLI command(az account get-access-token --output json --resource) to get a token to authenticate
+	out, errout, err := credentialClient.getAzureCLIAccessToken(command)
+
+	// Determining Azure CLI errors
+	if err != nil {
+		isLoginError := strings.HasPrefix(errout, AzNotLogIn)
```
JeffreyRichter: Normally in Go, if an error is not nil, then the other returned value should not be examined. Here, if err is not nil (an error occurred), then you ARE looking at errout. This doesn't feel right to me. Although it may be but there MUST be some comments explaining how this should be used if it goes against the normal Go way of doing things.

Also, it is very uncommon in Go to have a method return more than 2 values. Maybe there is a better way of accomplishing what you're trying to do here?

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+		executePara = "-c"
+	}
+
+	// Run Azure CLI command to return resulting Access Token or error
+	cmd = exec.Command(shell, executePara, command)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+
+	return stdout.Bytes(), stderr.String(), err
```
JeffreyRichter: Maybe define a struct that has stdout & stderr fields in it and return that?

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+// AzureCLICredentialClient provides the base for authenticating with Azure CLI Credential.
+type azureCLICredentialClient struct {
+}
+
+// NewAzureCLICredentialClient creates a new instance of the AzureCLICredentialClient.
+func newAzureCLICredentialClient() *azureCLICredentialClient {
+	return &azureCLICredentialClient{}
+}
+
+// Authenticate runs a Azure CLI command for a Azure CLI Credential and returns the resulting Access Token or
```
JeffreyRichter: Method name doesn't match

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+
+	err := json.Unmarshal(output, &token)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse expiresOn string to date as required time format
+	dateString := token.ExpiresOn
+	timeformat := "2006-01-02 15:04:05.999999"
+	expiresOnValue, err := time.ParseInLocation(timeformat, dateString, time.Local)
```
JeffreyRichter: I'm sure you do NOT want time.Local here; you want UTC instead. So, you should just call Parse to get UTC.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	accessToken.ExpiresOn = expiresOnValue.In(time.UTC)
```
JeffreyRichter: I don't understand what's happening here. token.ExpiresOn is a string (like 134). Then you parse it (line 136), then you convert it back to a string (this line). I don't think you need to do any of this. My guess is you're trying to convert it from UTC to UTC but if it's already in UTC then there is nothing to do.

sdk/azidentity/azure_cli_credential_client.go
```diff
@@ -0,0 +1,155 @@
...
+	accessToken.Token = token.Token
+
+	return accessToken, err
+}
+
+func (c *azureCLICredentialClient) scopeToResource(scopes []string) string {
+	if !strings.HasSuffix(scopes[0], DefaultSuffix) {
```
JeffreyRichter: Some comments explaining what this method is trying to do would be useful. It looks like if the 0th scope doesn't end with a suffix, you return it. Why? What does this mean?

If it does end with that suffix, then you scan the string looking for ".default" (which is not in a const variable unlike DefaultSuffix) and returns a string up to that. But, what if ".default" is not in the string at all?

sdk/azidentity/azure_cli_credential.go
wantedfast: Thanks for your review, We are looking into those comments carefully . As we were notified that @schabbs is finalize the design. We will resolve all comments later.

sdk/azidentity/azure_cli_credential.go
wantedfast: We have a new design,  azAccessTokenProvider implements interface azureCLIAccessTokenProvider would be as the parameter of NewAzureCLICredentialClient. No extra para to NewAzureCLICredential.  
And there are some comments been added to explain the logic.

sdk/azidentity/azure_cli_credential.go
wantedfast: This func was remmoved since it's unused.

sdk/azidentity/azure_cli_credential_client.go
wantedfast: This interface was re-named to azureCLIAccessTokenProvider what helps mock tests. Also comments added in here to help explaining it's purpose.

sdk/azidentity/azure_cli_credential_client.go
wantedfast: Thanks. 
Return this error as CredentialUnavaiableError driectly.

sdk/azidentity/azure_cli_credential_client.go
wantedfast: The expiresOnValue return from the Azure CLI is local time. So, get local time first, then parse to UTC.

sdk/azidentity/azure_cli_credential_client.go
wantedfast: The following code will remove the suffix from any scopes passed into the method since the Azure CLI command expect a resource string instead of a scope string. 
And some comments added to explaining what is this func purpose.

wantedfast:
Checkes not successful becuase TestManagedIdentityCredential_GetTokenInCloudShellLive test run failed.
This is wired since this PR don't change this test and ManagedIentityCredential code.

jhendrixMSFT:
Closing in favor of https://github.com/Azure/azure-sdk-for-go/pull/10645

PR #7044, opened 2020-01-21
add support for authfilecredential

Luyunmt:
https://github.com/Azure/azure-sdk-for-net/pull/9312

 The purpose of this PR is explained in this or a referenced issue.
Add Azure AuthFile credential into azidentity.
 
 The PR does not update generated files. 
These files are managed by the codegen framework at [Azure/autorest.go][].
Don't include any generated files.
 
 The PR targets the latest branch.
This PR would target track2 branch.
 
 Tests are included and/or updated for code changes.
unit tests are included.
 
 Updates to [CHANGELOG.md][] are included.
No any packages changed, so it don't include the CHANGELOG.md
 
 Apache v2 license headers are included in each file.
Included

### DIFF
--- a/sdk/azidentity/auth_file_credential.go
+++ b/sdk/azidentity/auth_file_credential.go
@@ -26,30 +26,26 @@ type authFileAccesstoken struct {
 	ManagementEndpointUrl          string `json:"managementEndpointUrl"`
 }
 
-// Enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
+// AuthFileCredential enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
 type AuthFileCredential struct {
 	filePath        string
 	credential      azcore.TokenCredential
 	authfileoptions *TokenCredentialOptions
 }
 
-// Creates an instance of the SdkAuthFileCredential class based on information in given SDK Auth file.
+// NewAuthFileCredential creates an instance of the AuthFileCredential class based on information in given SDK Auth file.
 func NewAuthFileCredential(filePath string, options *TokenCredentialOptions) (*AuthFileCredential, error) {
-	if filePath == "" {
-		return nil, fmt.Errorf("The parameter 'filePath' must set a value")
-	}
-
 	return &AuthFileCredential{filePath: filePath, authfileoptions: options}, nil
 }
 
-// Obtains a token from the Azure Active Directory service, using the specified client detailed specified in the SDK Auth file.
+// GetToken Obtains a token from the Azure Active Directory service, using the specified client detailed specified in the SDK Auth file.
 func (c *AuthFileCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
-	_, err := c.ensureCredential()
+	cred, err := c.ensureCredential()
 	if err != nil {
 		return nil, &AuthenticationFailedError{msg: "Error parsing SDK Auth File", inner: err}
 	}
 
-	return c.credential.GetToken(ctx, opts)
+	return cred.GetToken(ctx, opts)
 }
 
 // AuthenticationPolicy implements the azcore.Credential interface on AuthFileCredential.
@@ -57,16 +53,17 @@ func (c *AuthFileCredential) AuthenticationPolicy(options azcore.AuthenticationP
 	return newBearerTokenPolicy(c, options)
 }
 
-// Ensures that credential information is loaded from the SDK Auth file. This method should be called to initialize.
+// ensureCredential ensures that credential information is loaded from the SDK Auth file.
 func (c *AuthFileCredential) ensureCredential() (azcore.TokenCredential, error) {
 	if c.credential != nil {
 		return c.credential, nil
 	}
 
-	authData, err := c.parseCredentialsFile(c.filePath)
+	authData, err := ioutil.ReadFile(c.filePath)
 	if err != nil {
 		return nil, err
 	}
+
 
 	c.credential, err = c.buildCredentialForCredentialsFile(authData)
 	if err != nil {
@@ -76,17 +73,7 @@ func (c *AuthFileCredential) ensureCredential() (azcore.TokenCredential, error) 
 	return c.credential, nil
 }
 
-// Parse credential file from local file path to byte.
-func (c *AuthFileCredential) parseCredentialsFile(filePath string) ([]byte, error) {
-	authData, err := ioutil.ReadFile(filePath)
-	if err != nil {
-		return nil, err
-	}
-
-	return authData, nil
-}
-
-// Use the SDK auth file info to build a ClientSecretCredential
+// buildCredentialForCredentialsFile use the SDK auth file info to build a ClientSecretCredential
 func (c *AuthFileCredential) buildCredentialForCredentialsFile(authData []byte) (*ClientSecretCredential, error) {
 	token := &authFileAccesstoken{}
 
@@ -99,10 +86,6 @@ func (c *AuthFileCredential) buildCredentialForCredentialsFile(authData []byte) 
 	clientSecret := token.ClientSecret
 	tenantId := token.TenantId
 	activeDirectoryEndpointUrl := token.ActiveDirectoryEndpointUrl
-
-	if clientId == "" || clientSecret == "" || tenantId == "" || activeDirectoryEndpointUrl == "" {
-		return nil, fmt.Errorf("Malformed Azure SDK Auth file. The file should contain 'clientId', 'clientSecret', 'tenantId' and 'activeDirectoryEndpointUrl' values.")
-	}
 
 	// Parse string activeDirectoryEndpointUrl to a Url.
 	AuthorityHost, err := url.Parse(activeDirectoryEndpointUrl)
--- a/sdk/azidentity/auth_file_credential.go
+++ b/sdk/azidentity/auth_file_credential.go
@@ -13,19 +13,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-type authFileAccesstoken struct {
-	ClientId                       string `json:"clientId"`
-	ClientSecret                   string `json:"clientSecret"`
-	SubscriptionId                 string `json:"subscriptionId"`
-	TenantId                       string `json:"tenantId"`
-	ActiveDirectoryEndpointUrl     string `json:"activeDirectoryEndpointUrl"`
-	ResourceManagerEndpointUrl     string `json:"resourceManagerEndpointUrl"`
-	ActiveDirectoryGraphResourceId string `json:"activeDirectoryGraphResourceId"`
-	SqlManagementEndpointUrl       string `json:"sqlManagementEndpointUrl"`
-	GalleryEndpointUrl             string `json:"galleryEndpointUrl"`
-	ManagementEndpointUrl          string `json:"managementEndpointUrl"`
-}
-
 // AuthFileCredential enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
 type AuthFileCredential struct {
 	filePath        string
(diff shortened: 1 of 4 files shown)


sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return &AuthFileCredential{filePath: filePath}, nil
+}
+
+// Obtains a token from the Azure Active Directory service, using the specified client detailed specified in the SDK Auth file.
+func (c *AuthFileCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	_, err := c.ensureCredential()
+	if err != nil {
+		return nil, &AuthenticationFailedError{msg: "Error parsing SDK Auth File: " + err.Error()}
```
ellismg: NIT:  I think this should be `&AuthenticationFailedError{msg: "Error parsing SDK Auth File", inner: err}` so we don't drop the inner error (which may have additional context on top of what `.Error()` returns.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return c.credential, nil
+}
+
+// Parse credential file from local file path to byte.
+func (c *AuthFileCredential) parseCredentialsFile(filePath string) ([]byte, error) {
+	authData, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		log.Fatal(err)
```
ellismg: I don't believe calling `log.Fatal(err)` is what we want here.  It will cause the process to exit  I think this should return `nil, err` if `error` is non `nil`.  If you agree, we could get rid of this method completely and just call `ioutil.ReadFile` in `ensureCredential`

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
+type AuthFileCredential struct {
+	filePath   string
+	credential azcore.TokenCredential
+}
+
+// Creates an instance of the SdkAuthFileCredential class based on information in given SDK Auth file.
+func NewAuthFileCredential(filePath string) (*AuthFileCredential, error) {
```
ellismg: I think this also taken a `options *TokenCredentialOptions`, which are passed to the `NewClientSecretCredential` (but have the `AuthorityHost` that is used be the one specified in the file.  Similar to our other credential types.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,114 @@
...
+	activeDirectoryEndpointUrl := token.ActiveDirectoryEndpointUrl
+
+	if clientId == "" || clientSecret == "" || tenantId == "" || activeDirectoryEndpointUrl == "" {
+		return nil, fmt.Errorf("Malformed Azure SDK Auth file. The file should contain 'clientId', 'clientSecret', 'tenantId' and 'activeDirectoryEndpointUrl' values.")
+	}
+
+	// Parse string activeDirectoryEndpointUrl to a Url.
+	AuthorityHost, err := url.Parse(activeDirectoryEndpointUrl)
+	if err != nil {
+		return nil, err
```
ellismg: The addition of some context here might be helpful when trying to use the returned error to understand what went wrong.  Consider doing something like `return nil, fmt.Errorf("failed to parse activeDirectoryEndpointUrl in auth file: %w", err)`.

You could consider doing something similar on L95 to allow us to provide more context to the JSON decoding error.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	return c.credential.GetToken(ctx, opts)
+}
+
+// AuthenticationPolicy implements the azcore.Credential interface on AuthFileCredential.
+func (c *AuthFileCredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	return newBearerTokenPolicy(c, options)
+}
+
+// Ensures that credential information is loaded from the SDK Auth file. This method should be called to initialize.
+func (c *AuthFileCredential) ensureCredential() (azcore.TokenCredential, error) {
```
ellismg: I wonder if it's reasonable to pull all of the logic in `ensureCredential` into `NewAuthFileCredential`.  The fact it's a separate method was tied to how we did this in .NET (where doing blocking IO in a constructor felt a little weird) but it feels a little more idiomatic, but interested in what @jhendrixMSFT and other go folks think.

That would also let us front-load more error returning to when the credential is created instead of deferring until the fist time `GetToken` is called.

Luyunmt: Since exception was thrown in getToken in other languages. We can do in same way here.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	TenantId                       string `json:"tenantId"`
+	ActiveDirectoryEndpointUrl     string `json:"activeDirectoryEndpointUrl"`
+	ResourceManagerEndpointUrl     string `json:"resourceManagerEndpointUrl"`
+	ActiveDirectoryGraphResourceId string `json:"activeDirectoryGraphResourceId"`
+	SqlManagementEndpointUrl       string `json:"sqlManagementEndpointUrl"`
+	GalleryEndpointUrl             string `json:"galleryEndpointUrl"`
+	ManagementEndpointUrl          string `json:"managementEndpointUrl"`
+}
+
+// Enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
```
JeffreyRichter: The comment MUST start with AuthFileCredential to compile without errors.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+type AuthFileCredential struct {
+	filePath        string
+	credential      azcore.TokenCredential
+	authfileoptions *TokenCredentialOptions
```
JeffreyRichter: Usually these options are stored by value so that they cannot be changed after the credential is created to avoid race conditions causing data corruption.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Creates an instance of the SdkAuthFileCredential class based on information in given SDK Auth file.
```
JeffreyRichter: Comment must start with "NewAuthFileCredential". Is there such a thing as SdkAuthFileCredential? I don't see it.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+func NewAuthFileCredential(filePath string, options *TokenCredentialOptions) (*AuthFileCredential, error) {
+	if filePath == "" {
+		return nil, fmt.Errorf("The parameter 'filePath' must set a value")
```
JeffreyRichter: You mean it must not be the empty string; it has a value, it is "". I also think this check is not useful because someone can always write the code you have on line 42 and just pass "". So, if you have this check it should be where filePath is first used.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return &AuthFileCredential{filePath: filePath, authfileoptions: options}, nil
+}
+
+// Obtains a token from the Azure Active Directory service, using the specified client detailed specified in the SDK Auth file.
```
JeffreyRichter: Must start with GetToken

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+func (c *AuthFileCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	_, err := c.ensureCredential()
```
JeffreyRichter: Why does ensureCredential return a value that this ignores?
I see no other calls to this method that actually uses the 1st return value.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return c.credential.GetToken(ctx, opts)
+}
+
+// AuthenticationPolicy implements the azcore.Credential interface on AuthFileCredential.
+func (c *AuthFileCredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	return newBearerTokenPolicy(c, options)
+}
+
+// Ensures that credential information is loaded from the SDK Auth file. This method should be called to initialize.
```
JeffreyRichter: should be called by who to initialize?

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return c.credential, nil
+}
+
+// Parse credential file from local file path to byte.
+func (c *AuthFileCredential) parseCredentialsFile(filePath string) ([]byte, error) {
+	authData, err := ioutil.ReadFile(filePath)
```
JeffreyRichter: Why have a separate function with one line of code in it? You need the error check here AND also in the code that calls this duplicating the error checking.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	clientId := token.ClientId
+	clientSecret := token.ClientSecret
+	tenantId := token.TenantId
+	activeDirectoryEndpointUrl := token.ActiveDirectoryEndpointUrl
+
+	if clientId == "" || clientSecret == "" || tenantId == "" || activeDirectoryEndpointUrl == "" {
```
JeffreyRichter: I'd kill these checks. There are all kinds of other checks we're not doing and these checks look for a specific kind of bug in the service but not all kinds of other potential service bugs. I think we can trust the service. If we can't, then we have much worse problems.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("Malformed Azure SDK Auth file. The file should contain 'clientId', 'clientSecret', 'tenantId' and 'activeDirectoryEndpointUrl' values.")
+	}
+
+	// Parse string activeDirectoryEndpointUrl to a Url.
+	AuthorityHost, err := url.Parse(activeDirectoryEndpointUrl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse activeDirectoryEndpointUrl in auth file: %w", err)
+	}
+	c.authfileoptions.AuthorityHost = AuthorityHost
```
JeffreyRichter: It is scary to me to update option fields dynamically at runtime. If this is done, it MUST be done in a goroutine-safe way as other goroutines may be trying to use this client & its options at the same time.

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return NewClientSecretCredential(tenantId, clientId, clientSecret, c.authfileoptions)
```
JeffreyRichter: So, this is built by ensureCredential which it called by GetToken. This implies that each call to GetToken could create a new ClientSecretCredential which doesn't make sense. The logic flow is too hard to follow and opens the door for race conditions and creating garbage. I think the credential should be determined much earlier once, not in GetToken.

Luyunmt:
Thanks for your review, We are looking into those comments carefully . As we were notified that @schabbs is finalizing the design. We will resolve all comments later.

sdk/azidentity/auth_file_credential.go
Luyunmt: Thanks for your review, We are looking into those comments carefully . As we were notified that @schabbs is finalizing the design. We will resolve all comments later.

jongio:
@JeffreyRichter / @catalinaperalta - Can you please review these and help us get to sign-off for April release?  Thanks

JeffreyRichter:
Done


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fazure.microsoft.com%2Ftraining%2Fdistributed-cloud-apps%2F&data=02%7C01%7Cjeffreyr%40microsoft.com%7C82df4ff9e5ae46a5095908d7345e957b%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637035453166233741&sdata=4GnkpFUKxlZXGXFr6oQo5GB%2FFNIbDXIZDXUij0q4AEE%3D&reserved=0>)

sdk/azidentity/auth_file_credential.go
```diff
@@ -0,0 +1,101 @@
...
+// AuthFileCredential enables authentication to Azure Active Directory using configuration information stored Azure SDK Auth File.
+type AuthFileCredential struct {
+	filePath        string
+	credential      azcore.TokenCredential
+	authfileoptions TokenCredentialOptions
+}
+
+// NewAuthFileCredential creates an instance of the AuthFileCredential class based on information in given SDK Auth file.
+func NewAuthFileCredential(filePath string, options *TokenCredentialOptions) (*AuthFileCredential, error) {
+	options, err := options.setDefaultValues()
```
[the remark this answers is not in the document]
Luyunmt: The method setDefaultValues() I used is already exsited in azidentity.go . And other credential in 
method newAADIdentityClient in aad_identity_client.go use it.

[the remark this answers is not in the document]
JeffreyRichter: @catalinaperalta & @jhendrixMSFT : Please look into setDefaultValue - it seems that this is not following our accepted options pattern.

jongio:
This is on hold pending further discussion

PR #7350, opened 2020-02-14
Track2 chainedcredentialtest

catalinaperalta:
Adding a test to check for CredentialUnavailable errors in the ChainedTokenCredential test suite. 
Also adding a function to append errors as responses for the test server.

### DIFF
--- a/sdk/azidentity/chained_token_credential_test.go
+++ b/sdk/azidentity/chained_token_credential_test.go
@@ -107,7 +107,7 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
 	if err != nil {
 		t.Fatalf("Unable to create credential. Received: %v", err)
 	}
-	// It doesn't matter what credential we add to the chain as long as it is not a nil credential.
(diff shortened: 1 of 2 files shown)


sdk/azidentity/chained_token_credential_test.go
```diff
@@ -97,6 +97,37 @@ func TestChainedTokenCredential_GetTokenFail(t *testing.T) {
 	}
 }
 
+func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	srv.AppendResponse(mock.WithError(&CredentialUnavailableError{CredentialType: "MockCredential", Message: "Mocking a credential unavailable error"}))
```
jhendrixMSFT: There is `srv.AppendError()`, did that not work?

catalinaperalta: Oops! I must have skipped over that when I was using my intellisense. Removing it!

PR #7541, opened 2020-02-25
Ensure per-try timeout is properly cancelled

jhendrixMSFT:
Explicitly call cancel the per-try timeout when the response body has
been read/closed by the body download policy.
When the response body is returned to the caller for reading/closing,
wrap it in a responseBodyReader that will cancel the timeout when the
body is closed.
Logger.Should() will return false if no listener is set.

### DIFF
--- a/sdk/azcore/request.go
+++ b/sdk/azcore/request.go
@@ -144,6 +144,13 @@ func (req *Request) SkipBodyDownload() {
 	req.SetOperationValue(bodyDownloadPolicyOpValues{skip: true})
 }
 
+// returns true if auto-body download policy is enabled
+func (req *Request) bodyDownloadEnabled() bool {
+	var opValues bodyDownloadPolicyOpValues
+	req.OperationValue(&opValues)
+	return !opValues.skip
+}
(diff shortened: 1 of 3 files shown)


sdk/azcore/request.go
```diff
@@ -144,6 +144,13 @@ func (req *Request) SkipBodyDownload() {
 	req.SetOperationValue(bodyDownloadPolicyOpValues{skip: true})
 }
 
+// returns true if auto-body download policy is enabled
+func (req *Request) bodyDownloadEnabled() bool {
+	var opValues bodyDownloadPolicyOpValues
+	req.OperationValue(&opValues)
```
catalinaperalta: What is this line doing?

jhendrixMSFT: The request object contains a property bag `values` where you can stash various things.  This line is retrieving it from the property bag if available and populating the value.

catalinaperalta: Ah I see. Ok cool. Thanks for the explanation! 👍

PR #7546, opened 2020-02-26
Fix nil dereference error

ashvindeodhar:
e.Table.tsc.client.exec can return nil resp and non-nil err. This causes
resp.StatusCode to crash with nil dereference. This change adds the
nil check on resp before accessing StatusCode.


jhendrixMSFT:
@ArcturusZhang can you please release an updated SDK with this fix?

ashvindeodhar:
Thanks @jhendrixMSFT! @ArcturusZhang our Container Networking release for GA depends on this. Appreciate you releasing the updated SDK tonight.

PR #7886, opened 2020-03-14
GraphRbacApi-List-Add-Expand

UdiRaitblat:
GraphRbacApi - Add the $expand flag to List/ListComplete requests.


ArcturusZhang:
Thanks for contribution for go SDK @UdiRaitblat !
But unfortunately this PR cannot be merged because (almost) everything in the `services` directory is auto-generated from the [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs) using [autorest](https://github.com/Azure/autorest), you changes will be overridden on the next iteration of code generation.

Or if this PR is generated by you using autorest from the latest specs, please submit a release request on [Open API Hub](aka.ms/openapihub).

Therefore I have to close this PR, thanks for your contribution again.

UdiRaitblat:
@ArcturusZhang I've opened this change for a missing flag in the api call, will this logic ever be added? I prefer not to work with a fork. Can I somehow facilitate this code addition?

ArcturusZhang:
The code you are changing are generated from this [swagger](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/graphrbac/data-plane/Microsoft.GraphRbac/stable/1.6/graphrbac.json)

I am not sure if you have tested the REST API would work with your modifications. If it works, then this should be a bug on the rest api specs. I suggest you to open a new issue in this [repo](https://github.com/Azure/azure-rest-api-specs) with some context. And before they actually fix this in the swagger, I can see the only way to fix this is to working with a fork. Sorry about this.

PR #8108, opened 2020-03-30
Logging for Azure Identity credential types

catalinaperalta:
Adding initial logging for Azure Identity credential types: 
- ClientSecretCredential
- ClientCertificateCredential
- ChainedTokenCredential
- DeviceCodeCredential
- DefaultTokenCredential
- UsernamePasswordCredential
- ManagedIdentityCredential
- EnvironmentCredential


sdk/azidentity/chained_token_credential.go
```diff
@@ -41,10 +42,21 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
 		} else if err != nil { // if we receive some other type of error then we must stop looping and process the error accordingly
 			var authenticationFailed *AuthenticationFailedError
 			if errors.As(err, &authenticationFailed) { // if the error is an AuthenticationFailedError we return the error related to the invalid credential and append all of the other error messages received prior to this point
-				return nil, &AuthenticationFailedError{msg: "Received an AuthenticationFailedError, there is an invalid credential in the chain. " + createChainedErrorMessage(errList), inner: err}
+				authErr := &AuthenticationFailedError{msg: "Received an AuthenticationFailedError, there is an invalid credential in the chain. " + createChainedErrorMessage(errList), inner: err}
+				log := azcore.Log()
+				msg := fmt.Sprintf("Azure Identity => ERROR in GetToken() call for %T: %s", cred, authErr.Error())
+				log.Write(CredentialClassification, msg)
```
jhendrixMSFT: This should probably be a `ErrorClassification`.

sdk/azidentity/client_secret_credential.go
```diff
@@ -39,7 +41,21 @@ func NewClientSecretCredential(tenantID string, clientID string, clientSecret st
 // scopes: The list of scopes for which the token will have access.
 // Returns an AccessToken which can be used to authenticate service client calls.
 func (c *ClientSecretCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
-	return c.client.authenticate(ctx, c.tenantID, c.clientID, c.clientSecret, opts.Scopes)
+	tk, err := c.client.authenticate(ctx, c.tenantID, c.clientID, c.clientSecret, opts.Scopes)
+	log := azcore.Log()
+	if err != nil {
+		msg := fmt.Sprintf("Azure Identity => ERROR in GetToken() call for %T: %s", c, err.Error())
+		log.Write(CredentialClassification, msg)
```
jhendrixMSFT: This should probably be a `azcore.ErrorClassification`.

sdk/azidentity/client_secret_credential_test.go
```diff
@@ -79,6 +85,11 @@ func TestClientSecretCredential_GetTokenSuccess(t *testing.T) {
 }
 
 func TestClientSecretCredential_GetTokenInvalidCredentials(t *testing.T) {
+	l := azcore.Log()
+	l.SetClassifications(CredentialClassification, CredentialClassificationVerbose)
```
jhendrixMSFT: Just FYI this will impact other logging tests.  If it becomes a problem just add a `defer l.SetClassification()` to reset the classifications at the end of the test.

sdk/azidentity/credential_log_classification.go
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License.
+
+package azidentity
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
+	CredentialClassification        azcore.LogClassification = "credential"
```
jhendrixMSFT: Log classification names should have the prefix `Log`, so this would be `LogCredential`.

sdk/azidentity/chained_token_credential.go
```diff
@@ -41,10 +42,21 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
...
+				return nil, authErr
 			}
 			return nil, fmt.Errorf("Received an unexpected error: %w", err) // if we receive some other error type this is unexpected and we simple return the unexpected error
 		} else {
+			log := azcore.Log()
+			msg := fmt.Sprintf("Azure Identity => GetToken() result for %T: SUCCESS", cred)
+			log.Write(CredentialClassification, msg)
+			if log.Should(CredentialClassificationVerbose) {
+				vmsg := fmt.Sprintf("Azure Identity => Scopes: [%s]", strings.Join(opts.Scopes, ", "))
```
jhendrixMSFT: I would include this with the standard log message and omit `LogCredentialVerbose`.

sdk/azidentity/credential_log_classification.go
catalinaperalta: Updated.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Something I was thinking about related to a previous comment. Shouldn't we rename azcore.ErrorClassification to azcore.LogError? So that all log classifications do in fact begin with Log

catalinaperalta: Just checked it's already in that format! I've updated this in identity as well

sdk/azidentity/chained_token_credential.go
```diff
@@ -41,10 +42,19 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
...
+				log.Write(azcore.LogError, msg)
+				return nil, authErr
 			}
 			return nil, fmt.Errorf("Received an unexpected error: %w", err) // if we receive some other error type this is unexpected and we simple return the unexpected error
 		} else {
+			log := azcore.Log()
+			msg := fmt.Sprintf("Azure Identity => GetToken() result for %T: SUCCESS", cred)
+			log.Write(LogCredential, msg)
+			vmsg := fmt.Sprintf("Azure Identity => Scopes: [%s]", strings.Join(opts.Scopes, ", "))
+			log.Write(LogCredentialVerbose, vmsg)
```
jhendrixMSFT, line 57: For client certificate the scopes isn't part of verbose while it is here.

sdk/azidentity/default_token_credential.go
```diff
@@ -21,6 +25,7 @@ type DefaultTokenCredentialOptions struct {
 // - ManagedIdentityCredential
 // Consult the documentation of these credential types for more information on how they attempt authentication.
 func NewDefaultTokenCredential(options *DefaultTokenCredentialOptions) (*ChainedTokenCredential, error) {
+	log := azcore.Log()
```
jhendrixMSFT, line 28: The underlying credential types will log success/errors.  Do we think adding the same logging here is valuable?

sdk/azidentity/environment_credential.go
```diff
@@ -4,27 +4,40 @@
...
 
 import (
+	"fmt"
 	"os"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
 // NewEnvironmentCredential creates an instance of the EnvironmentCredential type and reads client secret details from environment variables.
 // If the expected environment variables are not found at this time, the GetToken method will return the default AccessToken when invoked.
```
jhendrixMSFT, line 14: Unrelated to logging, but is this part of the comment actually true?  Looks like if the environment vars aren't found there will be no future call to GetToken().

sdk/azidentity/default_token_credential.go
catalinaperalta: Well the logging I'm adding here is for the error that occurs when we cannot instantiate any credentials. 
I also thought it would be helpful to have the call for when DefaultTokenCredential invokes the NewChainedTokenCredential in the logs, because it might not be clear to some users why they are getting ChainedTokenCredential errors when they called DefaultTokenCredential. However, I will say that this information is in the comment of the function call, so I can see why this could be considered redundant. But I'm thinking through the scenario where the user is completely new to the identity library and does not understand exactly how DefaultTokenCredential works.

sdk/azidentity/environment_credential.go
catalinaperalta: You're right, this was the original comment. I will remove it.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Yes, forgot to remove this call to LogCredentialVerbose

catalinaperalta: Changed.

sdk/azidentity/default_token_credential.go
jhendrixMSFT: That's a good point.  It might be good to run through a failure scenario here to see what the logging spew looks like.

catalinaperalta: You're right, part of the log message was redundant in the error case. Here's the spew without the redundant error message:
```
Azure Identity => ERROR in NewEnvironmentCredential() call: Environment Credential: Missing environment variable AZURE_TENANT_ID
Azure Identity => ERROR in GetToken() call for NewDefaultTokenCredential(): an error was returned from an underlying credential type
```

sdk/azidentity/azidentity.go
```diff
@@ -16,6 +18,31 @@ const (
...
+	envVars := []string{}
+	log := azcore.Log()
+	if envCheck := os.Getenv("AZURE_TENANT_ID"); len(envCheck) > 0 {
+		envVars = append(envVars, "AZURE_TENANT_ID")
+	}
+	if envCheck := os.Getenv("AZURE_CLIENT_ID"); len(envCheck) > 0 {
+		envVars = append(envVars, "AZURE_CLIENT_ID")
+	}
+	if envCheck := os.Getenv("AZURE_CLIENT_SECRET"); len(envCheck) > 0 {
+		envVars = append(envVars, "AZURE_CLIENT_SECRET")
```
jianghaolu, line 33: Is AZURE_CLIENT_SECRET considered PII?

jhendrixMSFT: Good point, we should replace the actual value with `<REDACTED>`.

catalinaperalta: I'm not printing the value of the environment variable. I'm literally just printing the name of the variable for the user's knowledge so that they know this variable is set in the environment. If you look at line 42 you can see that I'm just printing the names of the variables here for the user to know what the environment is like that they're operation in. Also this array of env vars is not used anywhere else so it won't run the risk of adding client secret anywhere else.

jianghaolu: Got it! Thanks for the clarification.

catalinaperalta:
Closing will create a new PR

PR #8151, opened 2020-04-02
More updates to storage examples

jhendrixMSFT:
Add support for shared key authorization.
Added example for listing containers.
Fixed append blob creation.
Updated blob listing for both blobs.

### DIFF
--- a/sdk/storage/blob/2019-07-07/azblob/examples_test.go
+++ b/sdk/storage/blob/2019-07-07/azblob/examples_test.go
@@ -26,6 +26,16 @@ const (
 
 // returns a credential that can be used to authenticate with blob storage
 func getCredential() azcore.Credential {
+	// check environment for shared key credential info
+	accountName := os.Getenv("AZURE_STORAGE_ACCOUNT_NAME")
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/334

sdk/storage/blob/2019-07-07/azblob/examples_test.go
```diff
@@ -104,38 +135,14 @@ func ExampleBlockBlobOperations_Upload() {
...
 func ExampleAppendBlobOperations_Create() {
 	endpoint := pathJoin(getEndpoint(), containerName, appendBlobName)
 	client, err := NewClient(endpoint, getCredential(), nil)
 	if err != nil {
 		panic(err)
 	}
 	blobClient := client.AppendBlobOperations()
-	s := "text/plain"
-	a, err := blobClient.Create(context.Background(), int64(10), &AppendBlobCreateOptions{BlobContentType: &s})
+	a, err := blobClient.Create(context.Background(), 0, nil)
```
catalinaperalta, line 145: I think we can close the issue related to content type being required in the autorest.go repo due to this. Strange behavior that it worked previously, but clearly it's not required.

PR #8274, opened 2020-04-10
Adding block download test

catalinaperalta:


sdk/storage/blob/2019-07-07/azblob/examples_test.go
```diff
@@ -127,6 +127,21 @@ func ExampleBlockBlobOperations_Upload() {
...
 }
 
+func ExampleBlobOperations_Download() {
+	endpoint := pathJoin(getEndpoint(), containerName, blockBlobName)
+	client, err := NewClient(endpoint, getCredential(), nil)
+	if err != nil {
+		panic(err)
+	}
+	blobClient := client.BlobOperations(nil)
+	b, err := blobClient.Download(context.Background(), nil)
```
jhendrixMSFT, line 137: Maybe show reading from the response body to get the blob content.

catalinaperalta: So I added a print of the block type. But I have a question about reading the blob content. 
Question: I'm not sure what field the body is in? I looked and don't see the body field. 

For now I'm printing the blob type and adding that check in the output of the example. I'll merge this for now. If you can answer I'll create a separate PR to add reading the blob's content.

jhendrixMSFT: there should be a `Body` field on the `http.Response` object returned.

catalinaperalta: The operation returns a BlobDownloadResponse and I dont see the Body field on it :(

jhendrixMSFT: It would be on the `RawResponse`, so `BlobDownloadResponse.RawResponse.Body`.  We might want to consider adding a `Body` field on `BlobDownloadResponse`; we should discuss this in scrum today.

PR #8562, opened 2020-04-24
Updating NOTICE.txt to use Microsoft recommended template

kurtzeborn:
Guidance can be found here: https://docs.opensource.microsoft.com/content/using/required-notice-template.html


RickWinter:
CodeOwners and Notice was recently updated

PR #8666, opened 2020-04-29
Updating RetryAfter on azcore.Response to only return duration

catalinaperalta:


sdk/azcore/response.go
```diff
@@ -97,22 +97,22 @@ func (r *Response) removeBOM() {
 }
 
 // RetryAfter returns (non-zero, true) if the response contains a Retry-After header value.
```
jhendrixMSFT, line 99: Update doc comment to match new behavior.

PR #9811, opened 2020-06-02
Renaming DefaultTokenCredential to DefaultAzureCredential

catalinaperalta:
This PR renames DefaultTokenCredential to DefaultAzureCredential.

### DIFF
--- a/sdk/azidentity/default_azure_credential.go
+++ b/sdk/azidentity/default_azure_credential.go
@@ -9,10 +9,14 @@ const (
 	developerSignOnClientID = "04b07795-8ddb-461a-bbee-02f9e1bf7b46"
 )
 
-// DefaultAzureCredentialOptions contain information that can configure Default Token Credentials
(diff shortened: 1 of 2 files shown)


sdk/azidentity/default_azure_credential.go
```diff
@@ -9,23 +9,23 @@ const (
 	developerSignOnClientID = "04b07795-8ddb-461a-bbee-02f9e1bf7b46"
 )
 
-// DefaultTokenCredentialOptions contain information that can configure Default Token Credentials
-type DefaultTokenCredentialOptions struct {
+// DefaultAzureCredentialOptions contain information that can configure Default Token Credentials
```
jhendrixMSFT, line 12: `DefaultAzureCredentialOptions contains options for configuring how credentials are acquired.` (something like that)  Probably also add comments to each field in this struct.

catalinaperalta: Done

PR #9957, opened 2020-06-05
Azure Identity: Updating doc comments

catalinaperalta:


sdk/azidentity/chained_token_credential.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-// ChainedTokenCredential provides a TokenCredential implementation which chains multiple TokenCredential implementations to be tried in order
-// until one of the GetToken methods returns a non-default AccessToken.
+// ChainedTokenCredential provides a TokenCredential implementation that chains multiple TokenCredential implementations to be tried in order
```
jhendrixMSFT, line 14: Replace `TokenCredential implementation` with `TokenCredential sources`.

sdk/azidentity/chained_token_credential.go
```diff
@@ -11,8 +11,8 @@ import (
...
+// until one of the GetToken methods returns a non-nil AccessToken.
```
jhendrixMSFT, line 15: "non-nil" and "non-default" are both really implementation details of success.  I'd word it more like `It returns the token from the first successful call to GetToken().`

sdk/azidentity/chained_token_credential.go
```diff
@@ -30,7 +30,7 @@ func NewChainedTokenCredential(sources ...azcore.TokenCredential) (*ChainedToken
 	return &ChainedTokenCredential{sources: sources}, nil
 }
 
-// GetToken sequentially calls TokenCredential.GetToken on all the specified sources, returning the first non default AccessToken.
+// GetToken sequentially calls TokenCredential.GetToken on all the specified sources, returning the first non-nil AccessToken.
```
jhendrixMSFT, line 33: `returning the token from the first successful call to GetToken()`

sdk/azidentity/azidentity.go
```diff
@@ -54,18 +54,19 @@ func (e *AADAuthenticationFailedError) Error() string {
...
 
-// AuthenticationFailedError is a struct used to marshal responses when authentication has failed
+// AuthenticationFailedError is returned when the authentication request has failed.
 type AuthenticationFailedError struct {
 	inner error
 	msg   string
 }
 
-// Unwrap method on AuthenticationFailedError provides access to the inner error
+// Unwrap method on AuthenticationFailedError provides access to the inner error authentication
```
jhendrixMSFT, line 63: Will this always be an inner authentication error?  Maybe it's enough to say `provides access to the inner error if available.`

sdk/azidentity/azidentity.go
```diff
@@ -54,18 +54,19 @@ func (e *AADAuthenticationFailedError) Error() string {
...
+// error returned from Azure Active Directory
 func (e *AuthenticationFailedError) Unwrap() error {
 	return e.inner
 }
 
-// IsNotRetriable allows retry policy to stop execution in case it receives a AuthenticationFailedError
+// IsNotRetriable allows retry policy to stop execution in case it receives an AuthenticationFailedError
```
jhendrixMSFT, line 69: IsNotRetriable returns true indicating that this is a terminal error.

sdk/azidentity/azidentity.go
```diff
@@ -54,18 +54,19 @@ func (e *AADAuthenticationFailedError) Error() string {
...
 func (e *AuthenticationFailedError) IsNotRetriable() bool {
```
jhendrixMSFT, line 70: We should change this to `Temporary()`, see https://golang.org/pkg/net/#Error for the pattern.

jhendrixMSFT: Which also means it would return `false`.

sdk/azidentity/azidentity.go
```diff
@@ -88,10 +89,13 @@ func newAADAuthenticationFailedError(resp *azcore.Response) error {
 	return authFailed
 }
 
-// CredentialUnavailableError an error type returned when the conditions required to create a credential do not exist
+// CredentialUnavailableError is an error type returned when the conditions required to
```
jhendrixMSFT, line 92: `is the error type returned...`

sdk/azidentity/azidentity.go
jhendrixMSFT: We can leave this discussion for a separate PR.

catalinaperalta: Ok sounds good. I'll leave it as is for now, but I'm open to changing it to Temporary()

sdk/azidentity/azidentity.go
catalinaperalta: Done

sdk/azidentity/azidentity.go
catalinaperalta: Done

sdk/azidentity/azidentity.go
catalinaperalta: Done

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done

sdk/azidentity/chained_token_credential.go
catalinaperalta: Done

PR #10143, opened 2020-06-09
Adding sample for azcompute

catalinaperalta:
Adding compute sample that creates/updates a VM, gets a VM and lists VMs in a resource group. Also demonstrates the use of pollers and pagers.

### DIFF
--- a/sdk/compute/mgmt/2019-12-01/azcompute/example_test.go
+++ b/sdk/compute/mgmt/2019-12-01/azcompute/example_test.go
@@ -1,0 +1,164 @@
+package azcompute
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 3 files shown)


sdk/compute/mgmt/2019-12-01/azcompute/example_test.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+			Properties: &VirtualMachineProperties{
+				HardwareProfile: &HardwareProfile{
+					VMSize: VirtualMachineSizeTypesStandardA0.ToPtr(),
+				},
+				StorageProfile: &StorageProfile{
+					ImageReference: &ImageReference{
+						Publisher: to.StringPtr("MicrosoftWindowsServer"),
+						Offer:     to.StringPtr("WindowsServer"),
+						Sku:       to.StringPtr("2012-R2-Datacenter"),
```
catalinaperalta, line 76: This needs to be updated.

PR #10448, opened 2020-06-15
Adding firewall policies sample with more complex examples

catalinaperalta:
This PR includes:
- Example for creating a FirewallPolicy
- Example for create a FirewallPolicyRuleGroup that first creates a resource group that will then have the firewall resources added to it. This example also uses FirewallPolicyRuleClassifications which are discriminated types and also uses the previously created FirewallPolicy for the FirewallPolicyRuleGroup.
- Example for deleting FirewallPolicyRuleGroup
- Example for deleting a ResourceGroup with a manual example for polling

### DIFF
--- a/sdk/arm/network/2020-03-01/armnetwork/example_firewallpolicies_test.go
+++ b/sdk/arm/network/2020-03-01/armnetwork/example_firewallpolicies_test.go
@@ -8,8 +8,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/arm/resources/2019-05-01/armresources"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/to"
 )
 
@@ -17,14 +15,6 @@ const (
 	policyName    = "samplepolicy"
 	ruleGroupName = "sampleRuleGroup"
 )
-
-func getResourceGroupsOperations() armresources.ResourceGroupsOperations {
-	client, err := armresources.NewDefaultClient(getCredential(), nil)
-	if err != nil {
-		panic(err)
-	}
-	return client.ResourceGroupsOperations(subscriptionID)
-}
 
 func getFirewallPolicyRuleGroupsOperations() FirewallPolicyRuleGroupsOperations {
 	client, err := NewDefaultClient(getCredential(), nil)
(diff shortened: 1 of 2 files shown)


sdk/arm/network/2020-03-01/armnetwork/example_firewallpolicies_test.go
```diff
@@ -0,0 +1,157 @@
...
+	client, err := NewDefaultClient(getCredential(), nil)
+	if err != nil {
+		panic(err)
+	}
+	return client.FirewallPoliciesOperations(subscriptionID)
+}
+
+func ExampleFirewallPolicyRuleGroupsOperations_BeginCreateOrUpdate() {
+	// create a new resource group to create resource in
+	rg := createResourceGroup(resourceGroupName, location)
```
jhendrixMSFT, line 47: I'm kinda leaning towards omitting this to try and keep the example focused on just network.  I believe we omitted it for `armcompute`.  We can have a larger end-to-end sample that creates all the things elsewhere.  What do you think?

sdk/arm/network/2020-03-01/armnetwork/go.mod
```diff
@@ -5,7 +5,10 @@ go 1.13
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.8.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.0.0-00010101000000-000000000000
+	github.com/Azure/azure-sdk-for-go/sdk/arm/resources/2019-05-01/armresources v0.0.0-00010101000000-000000000000
```
jhendrixMSFT, line 8: go mod tidy

sdk/arm/network/2020-03-01/armnetwork/example_firewallpolicies_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/to"
+)
+
+const (
+	policyName    = "samplepolicy"
+	ruleGroupName = "sampleRuleGroup"
```
jhendrixMSFT, line 16: I think this should come from an env var like in the other sample.

jhendrixMSFT: Disregard, this isn't what I thought it was.

sdk/arm/network/2020-03-01/armnetwork/example_firewallpolicies_test.go
catalinaperalta: Agreed. I'm thinking we can add one in the general ARM directory. We can talk about it later.

sdk/arm/network/2020-03-01/armnetwork/go.mod
catalinaperalta: Done

PR #10454, opened 2020-06-16
Adding resources samples

catalinaperalta:
This PR includes: 
- Sample to create a resource group
- Sample for deleting a resource group that shows manual polling usage

### DIFF
--- a/sdk/arm/resources/2019-05-01/armresources/example_test.go
+++ b/sdk/arm/resources/2019-05-01/armresources/example_test.go
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package armresources
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/to"
+)
+
+const (
+	resourceGroupName = "sampleRG"
+	location          = "westus"
+)
+
+var (
+	// The subscription ID where the resource group exists
+	subscriptionID = os.Getenv("AZURE_SUBSCRIPTION_ID")
+)
+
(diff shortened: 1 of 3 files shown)


sdk/arm/resources/2019-05-01/armresources/example_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"os"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/to"
+)
+
+const (
+	resourceGroupName = "sampleRG"
```
jhendrixMSFT, line 18: We should decide as a team how we want to handle names like this, i.e. should they always be constants or should they be configurable via environment vars?  Or perhaps a mix of both.

catalinaperalta: So I'm using constants as non-secret, general information that can be used for the purpose of the sample. For environment vars, I'm putting information that will be specific to the sample user. So in this case, since the resource group doesn't exist, the resource group name and location are not specific to the sample user and can be used to show the values that they would need to use when actually carrying out the activity. For the other samples when (I assume) they already have their resource group and it's location set, then these are environment vars since they are user specific. What do you think?

jhendrixMSFT: Agreed that secrets must be env only.

So things that are preconditions (e.g. resource group name) are also env vars, and things that are specific to the sample (e.g. VM name, or RG in this specific sample) are constants?

catalinaperalta: Yes, that's how I've been thinking about it.

jhendrixMSFT: Makes sense, we should probably codify this in the design guidelines.

catalinaperalta: Agreed, will do 👍

PR #10645, opened 2020-06-19
Add initial work for CLI Credential

catalinaperalta:
This PR ports the track1 CLI Credential implementation to track2 format, so that users can use the token provided when authenticating with the CLI.

### DIFF
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -60,8 +60,7 @@ func (c *AzureCLICredential) AuthenticationPolicy(options azcore.AuthenticationP
 	return newBearerTokenPolicy(c, options)
 }
 
-// TODO: should the CLI request have a timeout? This is currently not being used
-const timeoutCLIRequest = 10000
+const timeoutCLIRequest = 10000 * time.Millisecond
 
 // authenticate creates a client secret authentication request and returns the resulting Access Token or
 // an error in case of authentication failure.
@@ -97,7 +96,7 @@ func defaultTokenProvider() func(ctx context.Context, resource string) ([]byte, 
 			return nil, fmt.Errorf(invalidResourceErrorTemplate, resource)
 		}
 
-		ctx, cancel := context.WithTimeout(ctx, timeoutCLIRequest*time.Second)
+		ctx, cancel := context.WithTimeout(ctx, timeoutCLIRequest)
 		defer cancel()
 
 		// Execute Azure CLI to get token
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -18,11 +18,8 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-type AzureCLITokenProvider interface {
-	GetCLIToken(ctx context.Context, resource string) ([]byte, error)
-}
-
-type azureCLITokenProvider struct{}
+// AzureCLITokenProvider can be used to supply the AzureCLICredential with an alternate token provider
+type AzureCLITokenProvider func(ctx context.Context, resource string) ([]byte, error)
 
 // AzureCLICredentialOptions contains options used to configure the AzureCLICredential
 type AzureCLICredentialOptions struct {
@@ -34,17 +31,16 @@ type AzureCLICredential struct {
 	tokenProvider AzureCLITokenProvider
 }
 
-// TODO: do we want an options bag for this credential that includes a developer specified Azure CLI path or only specify with env var?
 // NewAzureCLICredential constructs a new AzureCLICredential with the details needed to authenticate against Azure Active Directory
 // options: configure the management of the requests sent to Azure Active Directory.
 func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
-	if options != nil {
-		return &AzureCLICredential{
-			tokenProvider: options.TokenProvider,
-		}, nil
+	if options == nil {
+		var o AzureCLICredentialOptions
+		o.TokenProvider = defaultTokenProvider()
+		options = &o
 	}
 	return &AzureCLICredential{
-		tokenProvider: &azureCLITokenProvider{},
+		tokenProvider: options.TokenProvider,
 	}, nil
 }
 
@@ -74,7 +70,7 @@ const timeoutCLIRequest = 10000
 // ctx: The current request context
 // scopes: The scopes for which the token has access
 func (c *AzureCLICredential) authenticate(ctx context.Context, resource string) (*azcore.AccessToken, error) {
-	output, err := c.tokenProvider.GetCLIToken(ctx, resource)
+	output, err := c.tokenProvider(ctx, resource)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+	tokenProvider AzureCLITokenProvider
+}
+
+// TODO: do we want an options bag for this credential that includes a developer specified Azure CLI path or only specify with env var?
+// NewAzureCLICredential constructs a new AzureCLICredential with the details needed to authenticate against Azure Active Directory
+// options: configure the management of the requests sent to Azure Active Directory.
+func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
+	if options != nil {
+		return &AzureCLICredential{
+			tokenProvider: options.TokenProvider,
```
catalinaperalta, line 43: maybe this would be better with a setTokenProvider() to check that the token provider is not empty?

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+// AzureCLICredential enables authentication to Azure Active Directory using the Azure CLI command "az account get-access-token".
+type AzureCLICredential struct {
+	tokenProvider AzureCLITokenProvider
+}
+
+// TODO: do we want an options bag for this credential that includes a developer specified Azure CLI path or only specify with env var?
+// NewAzureCLICredential constructs a new AzureCLICredential with the details needed to authenticate against Azure Active Directory
+// options: configure the management of the requests sent to Azure Active Directory.
+func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
+	if options != nil {
```
jhendrixMSFT, line 41: I would change this to use our new pattern, see https://github.com/Azure/azure-sdk-for-go/blob/track2/sdk/arm/compute/2019-12-01/armcompute/client.go#L52 for an example

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+type AzureCLITokenProvider interface {
+	GetCLIToken(ctx context.Context, resource string) ([]byte, error)
+}
+
+type azureCLITokenProvider struct{}
```
jhendrixMSFT, line 25: If this empty type is just to provide a single method, you can implement the method as a first-class function and remove this type.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+	const invalidResourceErrorTemplate = "Resource %s is not in expected format. Only alphanumeric characters, [dot], [colon], [hyphen], and [forward slash] are allowed."
+	match, err := regexp.MatchString("^[0-9a-zA-Z-.:/]+$", resource)
+	if err != nil {
+		return nil, err
+	}
+	if !match {
+		return nil, fmt.Errorf(invalidResourceErrorTemplate, resource)
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeoutCLIRequest*time.Second)
```
jhendrixMSFT, line 105: I don't see the `ctx` actually being used to control the lifetime of the request.

catalinaperalta: Yeah, it isn't being used. I want to know if we do have to implement a timeout for this. Wondering what @schaabs thinks about whether we need the timeout? If we do then I can add the context to the command that will be executed.

jhendrixMSFT: Well, since this is doing I/O then really IMO it should, especially since the call to `Output()` will wait for the invoked process to exit.

I don't see a variant of `Output()` that takes a context, looks like you'll have to roll your own.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+	}, nil
+}
+
+// GetToken obtains a token from Azure Active Directory, using the Azure CLI command to authenticate.
+// ctx: Context used to control the request lifetime.
+// opts: TokenRequestOptions contains the list of scopes for which the token will have access.
+// Returns an AccessToken which can be used to authenticate service client calls.
+func (c *AzureCLICredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	// The following code will remove the /.default suffix from the scope passed into the method since AzureCLI expect a resource string instead of a scope string
+	if strings.HasSuffix(opts.Scopes[0], defaultSuffix) {
```
jhendrixMSFT, line 57: This seems a bit wonky, we shouldn't be repurposing `Scopes` to really mean an Azure resource.  If scopes aren't applicable to CLI credentials we should simply ignore them and pass the resource via other means.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,177 @@
...
+type AzureCLICredentialOptions struct {
+	TokenProvider AzureCLITokenProvider
+}
+
+// AzureCLICredential enables authentication to Azure Active Directory using the Azure CLI command "az account get-access-token".
+type AzureCLICredential struct {
+	tokenProvider AzureCLITokenProvider
+}
+
+// TODO: do we want an options bag for this credential that includes a developer specified Azure CLI path or only specify with env var?
```
jhendrixMSFT, line 37: Is this comment still relevant?  Looks like options are now in the constructor's signature.

sdk/azidentity/azure_cli_credential.go
catalinaperalta: We also do this in MSI credential, agree it's not great, but if we want a specific resources var then we would need to add it in TokenRequestOptions

sdk/azidentity/azure_cli_credential.go
catalinaperalta: I wanted to get a general opinion on this. I think it's important so that we can add the alternate TokenProvider.

sdk/azidentity/azure_cli_credential.go
catalinaperalta: It works with CommandContext. Done.

sdk/azidentity/azure_cli_credential.go
catalinaperalta: Done

sdk/azidentity/azure_cli_credential.go
catalinaperalta: Done

sdk/azidentity/azure_cli_credential.go
catalinaperalta: Removed the comment, since we'll have the alternate token provider in the options struct

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,175 @@
...
+	tokenProvider AzureCLITokenProvider
+}
+
+// NewAzureCLICredential constructs a new AzureCLICredential with the details needed to authenticate against Azure Active Directory
+// options: configure the management of the requests sent to Azure Active Directory.
+func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
+	if options == nil {
+		var o AzureCLICredentialOptions
+		o.TokenProvider = defaultTokenProvider()
+		options = &o
```
jhendrixMSFT, line 40: Nit, you could condense this to the following.
```go
options = &AzureCLICredentialOptions{TokenProvider: defaultTokenProvider()}
```

sdk/azidentity/azure_cli_credential.go
jhendrixMSFT: I would prefer we do that then as the intent is (somewhat) clear.  We'll have to ensure to properly document the cases where `Scopes` and `Resource` are required and gracefully fail when they're missing.

sdk/azidentity/azure_cli_credential.go
catalinaperalta: Done

sdk/azidentity/azure_cli_credential.go
catalinaperalta: So in order to do this I think we would probably have to have a separate `resources` field on TokenRequestOptions in order to differentiate between scopes and resources. Then when we call the AuthenticationPolicy pass in any scopes or resource strings we might have. I'll make this proposed change in a separate PR.

sdk/azidentity/azure_cli_credential.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// AuthenticationPolicy implements the azcore.Credential interface on AzureCLICredential and calls the Bearer Token policy
+// to get the bearer token.
+func (c *AzureCLICredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	return newBearerTokenPolicy(c, options)
+}
+
+// TODO: should the CLI request have a timeout? This is currently not being used
+const timeoutCLIRequest = 10000
```
jhendrixMSFT, line 64: I would express this in the resultant duration so it's clear what kind of value we're dealing with.
```go
const timeoutCLIRequest = 1*time.Minute
```
Or whatever the duration ends up being (right now it's ten thousand seconds, seems excessive).

Also, the comment is outdated as this is now used.

catalinaperalta: Done, I removed the comment and realized there was a mistake. It's 10000 milliseconds. I updated it to be that. I could also set it as 10 seconds. I used the dotnet implementation as a reference for this value.

PR #10974, opened 2020-06-29
Add logging for identity

catalinaperalta:
Add logging for Azure Identity credential types:

- ClientSecretCredential
- ClientCertificateCredential
- ChainedTokenCredential
- DeviceCodeCredential
- DefaultAzureCredential
- UsernamePasswordCredential
- ManagedIdentityCredential
- EnvironmentCredential


sdk/azidentity/azidentity.go
```diff
@@ -25,6 +27,31 @@ const (
 	defaultSuffix = "/.default"
 )
 
+// log environment variables that can be used for credential types
+func logEnvVars() {
+	// Log available environment variables
```
jhendrixMSFT, line 32: You might want to add the following to the beginning of this func.
```go
if !Log().Should(LogCredential) {
    return
}
```
This will skip the following code if logging is disabled.

sdk/azidentity/credential_log_classification.go
```diff
@@ -0,0 +1,13 @@
...
+
+package azidentity
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
+	LogCredential        azcore.LogClassification = "credential"
+	LogCredentialVerbose azcore.LogClassification = "credentialVerbose"
```
jhendrixMSFT, line 12: Other than one test this isn't used anywhere, can it be removed?  It can always be added later if needed.

sdk/azidentity/azidentity.go
catalinaperalta: Done

sdk/azidentity/credential_log_classification.go
catalinaperalta: So I was keeping it because I would really like to log scopes with the more verbose classification instead of the main logCredential classification. But, for now, I think it's ok to leave it all in LogCredential since there isn't more information that could also be added to the verbose level.

catalinaperalta: Removed.

sdk/azidentity/credential_log_classification.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azidentity
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
```
jhendrixMSFT, line 10: Nit: Having this file for just the log classification const seems excessive.  Maybe move this to azidentity.go, or if there are to be (internal) logging helper funcs that go here maybe rename the file to something like logging.go?

catalinaperalta: I'll move this and the log env vars func to a logging.go file. I think azidentity.go is a bit overloaded now.

catalinaperalta: Done

sdk/azidentity/chained_token_credential.go
```diff
@@ -41,10 +42,19 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
 		} else if err != nil { // if we receive some other type of error then we must stop looping and process the error accordingly
 			var authenticationFailed *AuthenticationFailedError
 			if errors.As(err, &authenticationFailed) { // if the error is an AuthenticationFailedError we return the error related to the invalid credential and append all of the other error messages received prior to this point
-				return nil, &AuthenticationFailedError{msg: "Received an AuthenticationFailedError, there is an invalid credential in the chain. " + createChainedErrorMessage(errList), inner: err}
+				authErr := &AuthenticationFailedError{msg: "Received an AuthenticationFailedError, there is an invalid credential in the chain. " + createChainedErrorMessage(errList), inner: err}
+				log := azcore.Log()
+				msg := fmt.Sprintf("Azure Identity => ERROR in GetToken() call for %T: %s", cred, authErr.Error())
```
jhendrixMSFT, line 47: Looks like this is repeated a lot, helper func?

sdk/azidentity/chained_token_credential.go
```diff
@@ -41,10 +42,19 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts azcore.Token
...
+				log.Write(azcore.LogError, msg)
+				return nil, authErr
 			}
 			return nil, fmt.Errorf("Received an unexpected error: %w", err) // if we receive some other error type this is unexpected and we simple return the unexpected error
 		} else {
+			log := azcore.Log()
+			msg := fmt.Sprintf("Azure Identity => GetToken() result for %T: SUCCESS", c)
```
jhendrixMSFT, line 54: Looks like this part could also be moved into a helper.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Added.

sdk/azidentity/chained_token_credential.go
catalinaperalta: Added.

sdk/azidentity/logging.go
```diff
@@ -0,0 +1,77 @@
...
+	case 3:
+		return "Azure Identity => Managed Identity environment: MSI_ENDPOINT"
+	case 4:
+		return "Azure Identity => Managed Identity environment: Unavailable"
+	default:
+		return "Azure Identity => Managed Identity environment: Unknown"
+	}
+}
+
+func addGetTokenFailureLogs(log *azcore.Logger, credName string, err error) {
```
jhendrixMSFT, line 74: It looks like this will result in duplicate calls to `err.Error()` being logged.

catalinaperalta: Removed it.

sdk/azidentity/default_azure_credential.go
```diff
@@ -51,7 +53,10 @@ func NewDefaultAzureCredential(options *DefaultAzureCredentialOptions) (*Chained
 	}
 	// if no credentials are added to the slice of TokenCredentials then return a CredentialUnavailableError
 	if len(creds) == 0 {
-		return nil, &CredentialUnavailableError{CredentialType: "Default Token Credential", Message: errMsg}
+		err := &CredentialUnavailableError{CredentialType: "Default Azure Credential", Message: errMsg}
+		azcore.Log().Write(azcore.LogError, logCredentialError(err.CredentialType, err))
+		return nil, err
 	}
+	azcore.Log().Write(LogCredential, "Azure Identity => NewDefaultAzureCredential() invoking NewChainedTokenCredential()")
```
jhendrixMSFT, line 60: Does this add any value?  Looks like there is logging that happens in `NewChainedTokenCredential()`, is that enough?

catalinaperalta: Yes, I think we might have talked about this in V1 of the logging PR 😅. It's important so users can understand why they will get ChainedTokenCredential log messages after this point. Since DAC will call that.

PR #11318, opened 2020-07-09
Add new error types to sdk/internal/runtime package

jhendrixMSFT:
Added NewFrameError() for wrapping errors with stack frame information.
Added NewResponseError() for wrapping errors and/or failed HTTP requests
with the underlying HTTP response.

### DIFF
--- a/sdk/internal/runtime/frame_error.go
+++ b/sdk/internal/runtime/frame_error.go
@@ -12,12 +12,13 @@ import (
 
 // NewFrameError wraps the specified error with an error that provides stack frame information.
 // Call this at the inner error's origin to provide file name and line number info with the error.
+// You MUST supply an inner error.
 // DO NOT ARBITRARILY CALL THIS TO WRAP ERRORS!  There MUST be only ONE error of this type in the chain.
 func NewFrameError(inner error, stackTrace bool, skipFrames, totalFrames int) error {
 	fe := frameError{inner: inner, info: "stack trace unavailable"}
 	if stackTrace {
-		// the skipFrames+2 is to skip StackTrace and ourselves
-		fe.info = StackTrace(skipFrames+2, totalFrames)
+		// the skipFrames+3 is to skip runtime.Callers(), StackTrace and ourselves
+		fe.info = StackTrace(skipFrames+3, totalFrames)
 	} else if pc, file, line, ok := runtime.Caller(skipFrames + 1); ok {
 		// the skipFrames + 1 is to skip ourselves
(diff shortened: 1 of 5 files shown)


sdk/internal/runtime/frame_error.go
```diff
@@ -0,0 +1,43 @@
...
+package runtime
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// NewFrameError wraps the specified error with an error that provides stack frame information.
+// Call this at the inner error's origin to provide file name and line number info with the error.
+// DO NOT ARBITRARILY CALL THIS TO WRAP ERRORS!  There MUST be only ONE error of this type in the chain.
```
catalinaperalta, line 15: So only use this with the top-most error we return? How do we guarantee that?

sdk/internal/runtime/frame_error.go
```diff
@@ -0,0 +1,43 @@
...
+
+// contains stack frame info
+type frameError struct {
+	inner error
+	info  string
+}
+
+// Error implements the error interface for type frameError.
+func (f *frameError) Error() string {
+	return fmt.Sprintf("%s: \n%s\n", f.inner.Error(), f.info)
```
catalinaperalta, line 37: Nit: why is there a space after the error message if there is also a new line?

sdk/internal/runtime/frame_error.go
jhendrixMSFT: Unfortunately I don't think we can really guarantee it, outside of walking the error chain and verifying that there's only one frameError in the chain, which would also mean we have to export this type.  Maybe we want to do this in tests which means the type would need to be exported (it's still in an internal package so only we can access it).

sdk/internal/runtime/frame_error.go
jhendrixMSFT: Fixed

catalinaperalta:
One last thing, we should have some tests showing the use of the error types? Possibly make sure that we are getting a stack trace? Also getting an appropriate error message.

sdk/internal/runtime/frame_error.go
catalinaperalta: Hmm I see, well we can think about exporting it, but I think it's better in the internal package for now.

jhendrixMSFT: It would remain in the exported from the internal package.  Exporting it would allow tests to walk the error chain looking for multiple `*FrameError` types.

PR #11428, opened 2020-07-09
Add HTTPResponse interface

jhendrixMSFT:
To be used by callers to access the raw HTTP response from an error in
the event of an API call failure.
Updated sdk/internal dependency to latest version.

### DIFF
--- a/sdk/azcore/error.go
+++ b/sdk/azcore/error.go
@@ -9,7 +9,7 @@ import (
 	"errors"
 	"net/http"
 
-	azruntime "github.com/Azure/azure-sdk-for-go/sdk/internal/runtime"
+	sdkruntime "github.com/Azure/azure-sdk-for-go/sdk/internal/runtime"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


sdk/azcore/error.go
```diff
@@ -7,47 +7,29 @@ package azcore
 
 import (
 	"errors"
-	"fmt"
+	"net/http"
+
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/internal/runtime"
```
catalinaperalta, line 12: Just wondering if the alias should be azruntime? Doesn't really matter but it implies azure runtime, not necessarily bad, but possibly misleading. Anyways, not going to block on that, just some thoughts about it.

jhendrixMSFT: I did this to disambiguate between our internal `runtime` package and the one in the standard library.  I picked `az` as we seem to prefix all of our hand-authored track 2 content with it.

catalinaperalta: Yes I figured, just thought if we ever have an azruntime package in the future it could be weird. Not that it would ever be imported here. I thought maybe 'sdkruntime' or 'internalruntime' to differentiate internal packages, but I dont think either of those are good aliases either. So we can leave it like this.

jhendrixMSFT: I like `sdkruntime`, I'll switch to that.

catalinaperalta: Ok sounds good to me 😊

PR #11851, opened 2020-07-23
Correct archetype-sdk-client from service level ci.yml

scbedd:
Title describes.

Build definitions are present, so we should get checks firing on this one.

### DIFF
--- a/sdk/azcore/ci.yml
+++ b/sdk/azcore/ci.yml
@@ -9,6 +9,6 @@ pr:
     - sdk/azcore/
     
 stages:
-- template: ../eng/pipelines/templates/jobs/archetype-sdk-client.yml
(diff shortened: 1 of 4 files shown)


sdk/azcore/ci.yml
```diff
@@ -9,6 +9,6 @@ pr:
     - sdk/azcore/
     
 stages:
-- template: ../eng/pipelines/templates/jobs/archetype-sdk-client.yml
+- template: ../eng/../pipelines/templates/jobs/archetype-sdk-client.yml
```
jhendrixMSFT, line 12: Doesn't look right (applies to others).

scbedd: Not certain why this was fine for one of them, but not the rest. Pushed an update.

PR #12027, opened 2020-07-30
Adding Liftr SDK

Harsh-Wardhan-Agarwal:


ArcturusZhang:
Hi @Harsh-Wardhan-Agarwal thanks for your contribution!

Would you please provide the location of the swagger that generates your new packages? If this swagger is not in [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs), I am afraid we could not release that package until it goes into azure-rest-api-specs.

And if it is in azure-rest-api-specs, the typical procedure of releasing new SDK packages should be:
Go to `aka.ms/openapihub` and submit a release request (it could only work if you are a MSFT employee)
After that I will handle the release process

ArcturusZhang:
Closed since no response from the author

Issue #7542, opened 2020-02-25
Don't retry non-idempotent operations when response body download fails


jhendrixMSFT:
Body download policy will trigger a retry if reading the response body fails.  We should not retry in the case of non-idempotent operations.  Still need to indicate to the caller that the operation succeeded.

jhendrixMSFT:
Fixed

PR #12036, opened 2020-07-31
Fix cancellation of per try timeout

jhendrixMSFT:
Per try timeout is used to ensure that an HTTP operation doesn't take
too long, e.g. that a GET on some URL doesn't take an inordinant amount
of time.
Once the HTTP request returns, the per try timeout should be cancelled,
not when the response has been read to completion.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/7542

PR #12200, opened 2020-08-10
Fixed the broken links

sima-zhu:
 

services/classic/management/README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Service Management packages for Go
 
-The `github.com/Azure/azure-sdk-for-go/services/classic/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](../arm) folder.
+The `github.com/Azure/azure-sdk-for-go/services/classic/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](https://github.com/Azure/azure-sdk-for-go/tree/master/sdk/armcore) folder.
```
jhendrixMSFT, line 3: Link should be https://github.com/Azure/azure-sdk-for-go/tree/master/services

eng/common/pipelines/templates/steps/verify-links.yml
```diff
@@ -1,5 +1,6 @@
 parameters:
   Directory: 'not-specified'
+  IgnoreLinksFile: "$(Build.SourcesDirectory)/eng/ignore-links.txt"
```
weshaggard, line 3: We shouldn't need to make these changes manually. Lets try to get the sync PR merged https://github.com/Azure/azure-sdk-for-go/pull/11024/files.

azure-pipelines.yml
```diff
@@ -60,6 +59,14 @@ jobs:
     - script: go test $(dirname $(find . -path ./vendor -prune -o -path ./sdk -prune -o -path ./sdk -prune -o -name '*_test.go' -print) | sort -u)
       workingDirectory: '$(sdkPath)'
       displayName: 'Run Tests'
+    - task: PowerShell@2
```
weshaggard, line 62: Why aren't we using the template here?

weshaggard: @jhendrixMSFT is the the best place to add in a link verification check?

sima-zhu: 1. The yml here reads the $sdkPath to get the script. We are using ${Build.SourceDirectory} in template. The two paths are not the same 
2. We have to exclude path of 'vendor' when read .md files. So the one in template cannot apply this case.

eng/common/pipelines/templates/steps/verify-links.yml
sima-zhu: Yes, will rebase my changes once the sync PR gets in. @KarishmaGhiya is working on merging the sync PR.

azure-pipelines.yml
weshaggard: I'm not very familar with the go repo it doesn't even look like this pipeline is for the track 2 libraries which I believe are under the root sdk folder. So it then becomes a question of should we run the link validation on these libraries.

If we should run it on these libraries we should try and run it on all files and not exclude directories as that defeats the purpose of the checks.  If for some reason we do end up needing to exclude some files under the path then we should add the glob pattern as a template parameter.

jhendrixMSFT: We are planning on unifying the pipelines in this repo to avoid this bifurcation.  @scbedd is aware of the work that needs to happen here.

sima-zhu: @weshaggard They are having two yml files which mutual exclude with each other.
Codes under sdk folder runs architype-sdk-client.yml
Codes outside of sdk folder (including the README.md in root) runs azure-pipeline.yml
The vendor folder is generated one during build time, so we cannot check the .md files under vendor. 

Do we want to check sdk folder under azure-pipeline.yml as well? I saw most of the steps exclude sdk path.

weshaggard: No I'd suggest lets limit our link verification to archetype-sdk-client.yml for now and once things converge they will be get checked.

As for md files getting generated as part of the build those should still be checked and valid, especially if we plan to publish/release them anywhere. If the generate md files have bad links we should follow-up with the generation tools and get them fixed. We can of course do that as a follow-up item.

sima-zhu: Vendor includes 3rd party dependencies. We do not publish anything under vendor, so we cannot fix the broken links introduce by 3rd party. 

I'd suggest to leave the check in azure-pipeline as well until they find a way to unify the path. Otherwise, we have plenty of codes without coverage. @weshaggard thoughts?

weshaggard: I'd be OK with leaving the check here but I'd still like to use our template, even if we need to make some tweaks to the template. Can we do the validation before the build so that we don't have pulled down all the 3rd party dependencies yet?

sima-zhu: I have no clues what's happened on the vm. But command failed to read the script under System root ({Build.SourceDirectory}/{system.default}), throw me an NoSuchFile exception.

sima-zhu: Tried before install dependencies and after. Not working anywhere.

services/classic/management/README.md
ArcturusZhang: And also, should this be `services` folder?

services/classic/management/README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Service Management packages for Go
 
-The `github.com/Azure/azure-sdk-for-go/services/classic/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](../arm) folder.
+The `github.com/Azure/azure-sdk-for-go/services/classic/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [arm](https://github.com/Azure/azure-sdk-for-go/tree/master/services) folder.
```
ArcturusZhang, line 3: This folder name is kind of misleading - should be `services`
```suggestion
The `github.com/Azure/azure-sdk-for-go/services/classic/management` packages are used to perform operations using the Azure Service Management (ASM), aka classic deployment model. Read more about [Azure Resource Manager vs. classic deployment](https://azure.microsoft.com/documentation/articles/resource-manager-deployment-model/). Packages for Azure Resource Manager are in the [services](https://github.com/Azure/azure-sdk-for-go/tree/master/services) folder.
```

azure-pipelines.yml
weshaggard: Ok I'm fine with keeping this calling the script directly for now, especially given that it will likely go away in the future.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -64,6 +64,10 @@ stages:
         parameters: 
           ServiceDirectory: ${{ parameters.ServiceDirectory }}
           GoWorkspace: $(GO_WORKSPACE_PATH)
+          
+      - template: ../../../common/pipelines/templates/steps/verify-links.yml
```
weshaggard, line 68: Generally a better idea to use a full path here as it is easier to see the exact path so I suggest you use `/eng/common/pipelines/templates/steps/verify-links.yml`

azure-pipelines.yml
weshaggard: Out of curiosity does this work if you use a full path to the template like `/eng/common/pipelines/templates/steps/verify-links.yml`.

services/classic/management/README.md
sima-zhu: @ArcturusZhang, Checked in your suggestions.

azure-pipelines.yml
sima-zhu: `/eng/common/pipelines/templates/steps/verify-links.yml` works.
I have trouble to read `verify-links.ps1`. 
I made some experiment in test pipeline. It has trouble to read the `sdkPath` I passed to verify-links.yml.
https://dev.azure.com/azure-sdk/playground/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=1886&branch=azure-pipelines&nonce=R8Pr%2FoPlyhiu88lmCpfuzg%3D%3D

sima-zhu:
Tested over the architype-sdk-client.yml. Working as expected.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=493745&view=logs&jobId=b70e5e73-bbb6-5567-0939-8415943fadb9

azure-pipelines.yml
sima-zhu: Meanwhile, we have to change the template (Verify-Links.yml) to accommodate the changes.

sima-zhu: Code changes can refer here: https://github.com/sima-zhu/azure-sdk-for-go/tree/azure-pipelines

sima-zhu: Failed pipeline: https://dev.azure.com/azure-sdk/playground/_build/results?buildId=493797&view=logs&j=c30e86aa-a579-54ac-9cc0-7664eb300cda&t=e9bff205-e0a2-5082-2954-a81ea98353a2

weshaggard: Passing:
```
      parameters:
        Workspace: '$(sdkPath)'
```
will not set `$(WorkSpace)` variable. You would need to reference it like `${{ parameters.WorkSpace }}`.

weshaggard: By the way I'm fine with doing this as a follow-up I do think we might want to enable some more extension points to the template but we don't need to block on that fir this work.

Issue #5412, opened 2019-08-01
Authenticating with Azure using Certificate


udbhav-sharma:
I am trying to authenticate with Azure Active Directory using Certificate. I get following error when I try authentication using SDK 
`Authorizer failed: failed to get oauth token from certificate auth: failed to decode pkcs12 certificate while creating spt: pkcs12: expected exactly two safe bags in the PFX PDU` 

Same certificate works if I use powershell commandlet to login to azure.

[Code Link](https://gist.github.com/udbhav-sharma/3aaff6c673af918e17cd18a82614958e)

There is one more concern I like to raise - Golang SDK forces to provide the certificate in pfx format where as `Login-AzureRmAccount` powershell commandlet not only works with certificate without private key but also only requires certificate thumbprint. Why is there a discrepancy in the API for different languages?

loarabia:
Hi @udbhav-sharma  I've tagged this bug so that the right team member can take a look.

udbhav-sharma:
Is there a update?

ArcturusZhang:
@jhendrixMSFT Hi would you please check if this is related with auth in go-autorest?

jhendrixMSFT:
Yes this is from the `adal` package in `go-autorest`.  Do you have a code sample how you're using the API?

jhendrixMSFT:
Thanks.  We use `pkcs12.Decode()` to extract the cert/private key which expects there to be only one cert/private key pair in the data.  How are you creating the cert?

Regarding authentication with certificate thumbprint I've opened https://github.com/Azure/go-autorest/issues/468 to track it.

gapariha:
Hi @jhendrixMSFT - I'm also seeing similar issue as above while trying to authN with certificate using Go auth SDK: "github.com/Azure/go-autorest/autorest/azure/auth"   NewAuthorizerFromEnvironment()

`"Error": "failed to get oauth token from certificate auth: failed to decode pkcs12 certificate while creating spt: pkcs12: expected exactly two safe bags in the PFX PDU"`

Wondering if there has been an update for this issue or any workaround to that matter? 
Regards.

jhendrixMSFT:
No update here as we don't have a repro to investigate further.  How was your cert created?

gapariha:
I see. I have generated the cert in the KeyVault with OneCert CA. 
In the PEM file, I can see one private key and but 3 certificate sections.

tamilmani1989:
[...]

PR #12208, opened 2020-08-12
Add support for PFX files

jhendrixMSFT:
This is to keep parity with track 1 that only supported PFX.
Moved reading and parsing of certificate data into the
NewClientCertificateCredential constructor so it will fail earlier.
Fixed an issue with certificates containing intermediate authorities.


jhendrixMSFT:
Forward-port to fix https://github.com/Azure/azure-sdk-for-go/issues/5412

PR #12246, opened 2020-08-17
Improvements to telemetry policy

jhendrixMSFT:
Add support for application ID string.
Preserve any existing user-agent string data.
Add flag to disable telemetry info.

### DIFF
--- a/sdk/azcore/policy_telemetry.go
+++ b/sdk/azcore/policy_telemetry.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 )
 
 // TelemetryOptions configures the telemetry policy's behavior.
@@ -18,6 +19,13 @@ type TelemetryOptions struct {
 	// Value is a string prepended to each request's User-Agent and sent to the service.
 	// The service records the user-agent in logs for diagnostics and tracking of client requests.
 	Value string
+
+	// ApplicationID is an application-specific identification string used in telemetry.
+	// It has a maximum length of 24 characters and must not contain any spaces.
+	ApplicationID string
+
+	// Disabled will prevent the addition of any telemetry data to the User-Agent.
+	Disabled bool
 }
 
 type telemetryPolicy struct {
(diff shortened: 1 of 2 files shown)


sdk/azcore/policy_telemetry.go
```diff
@@ -11,31 +11,62 @@ import (
...
 	// Value is a string prepended to each request's User-Agent and sent to the service.
 	// The service records the user-agent in logs for diagnostics and tracking of client requests.
 	Value string
+
+	// ApplicationID is an application-specific identification string used in telemetry.
+	// It has a maximum length of 24 characters and must not contain any spaces.
+	ApplicationID string
+
+	// Disabled will prevent the addition of any telemetry data to the User-Agent.
+	Disabled bool
```
JeffreyRichter, line 28: I think it would be better to have the code that creates the pipeline just exclude this policy entirely if disabled. I don't thin it's the "job" of the policy to ignore itself.

jhendrixMSFT: Having the policy's configuration be self-contained means we only have to change it in one place (azcore) without making any changes to the code generator.  I don't see this being any different than setting the retry policy's MaxRetry count to zero to turn it off.

JeffreyRichter: Yes, that's true (about maxretries = 0) - OK, keep Disable as is.

sdk/azcore/policy_telemetry_test.go
```diff
@@ -42,3 +42,69 @@ func TestPolicyTelemetryWithCustomInfo(t *testing.T) {
...
+	if v := resp.Request.Header.Get(HeaderUserAgent); v != fmt.Sprintf("%s %s", appID, platformInfo) {
+		t.Fatalf("unexpected user agent value: %s", v)
+	}
+}
+
+func TestPolicyTelemetryWithAppIDSanitized(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	srv.SetResponse()
+	const appID = "This will get the spaces removed an truncated."
```
catalinaperalta, line 82: Nit: "and"

jhendrixMSFT: Fixed

PR #12255, opened 2020-08-18
Improvements to body download policy

jhendrixMSFT:
Always download the response body for error responses, i.e. HTTP status
codes >= 400.

### DIFF
--- a/sdk/azcore/policy_body_download.go
+++ b/sdk/azcore/policy_body_download.go
@@ -22,17 +22,19 @@ func newBodyDownloadPolicy() Policy {
 			return resp, err
 		}
 		var opValues bodyDownloadPolicyOpValues
-		if req.OperationValue(&opValues); !opValues.skip && resp.Body != nil {
-			// Either bodyDownloadPolicyOpValues was not specified (so skip is false)
-			// or it was specified and skip is false: don't skip downloading the body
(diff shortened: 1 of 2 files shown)


sdk/azcore/policy_body_download.go
```diff
@@ -22,17 +22,18 @@ func newBodyDownloadPolicy() Policy {
...
+		// don't skip downloading error response bodies
+		if req.OperationValue(&opValues); opValues.skip && resp.StatusCode < 400 {
+			return resp, err
+		}
+		// Either bodyDownloadPolicyOpValues was not specified (so skip is false)
+		// or it was specified and skip is false: don't skip downloading the body
+		b, err := ioutil.ReadAll(resp.Body)
+		resp.Body.Close()
+		if err != nil {
+			return resp, newBodyDownloadError(err, req)
```
catalinaperalta, line 34: Why do we return the response here and not nil?

jhendrixMSFT: In case other stages in the pipeline need to inspect the response (e.g. logging).  The only time we should be returning a nil response in the pipeline is if the response is in fact nil.

catalinaperalta: Got it, thanks! :)

PR #12256, opened 2020-08-18
RPRegistrationPolicy improvements

jhendrixMSFT:
Set Attempts to zero to disable the policy and added a test for it.

### DIFF
--- a/sdk/armcore/policy_register_rp.go
+++ b/sdk/armcore/policy_register_rp.go
@@ -32,6 +32,7 @@ type RegistrationOptions struct {
 	// Attempts is the total number of times to attempt automatic registration
 	// in the event that an attempt fails.
 	// The default value is 3.
+	// Set to zero to disable the policy.
 	Attempts int
 
 	// PollingDelay is the amount of time to sleep between polling intervals.
@@ -66,7 +67,9 @@ func DefaultRegistrationOptions() RegistrationOptions {
 }
 
 // NewRPRegistrationPolicy creates a policy object configured using the specified pipeline
-// and options. Pass nil to accept the default options; this is the same as passing the result
+// and options.  The policy controls if an unregistered resource provider should automatically
+// be registered. See https://aka.ms/rps-not-found for more information.
+// Pass nil to accept the default options; this is the same as passing the result
 // from a call to DefaultRegistrationOptions().
 func NewRPRegistrationPolicy(cred azcore.Credential, o *RegistrationOptions) azcore.Policy {
 	if o == nil {
@@ -87,6 +90,10 @@ type rpRegistrationPolicy struct {
 }
 
 func (r *rpRegistrationPolicy) Do(ctx context.Context, req *azcore.Request) (*azcore.Response, error) {
+	if r.options.Attempts == 0 {
+		// policy is disabled
+		return req.Next(ctx)
+	}
 	const unregisteredRPCode = "MissingSubscriptionRegistration"
 	const registeredState = "Registered"
 	var rp string
--- a/sdk/armcore/policy_register_rp.go
+++ b/sdk/armcore/policy_register_rp.go
@@ -29,11 +29,11 @@ const (
 
 // RegistrationOptions configures the registration policy's behavior.
 type RegistrationOptions struct {
-	// Attempts is the total number of times to attempt automatic registration
+	// MaxAttempts is the total number of times to attempt automatic registration
 	// in the event that an attempt fails.
 	// The default value is 3.
 	// Set to zero to disable the policy.
-	Attempts int
+	MaxAttempts int
 
 	// PollingDelay is the amount of time to sleep between polling intervals.
 	// The default value is 15 seconds.
@@ -58,7 +58,7 @@ type RegistrationOptions struct {
 // DefaultRegistrationOptions returns an instance of RegistrationOptions initialized with default values.
 func DefaultRegistrationOptions() RegistrationOptions {
 	return RegistrationOptions{
-		Attempts:        3,
+		MaxAttempts:     3,
 		PollingDelay:    15 * time.Second,
 		PollingDuration: 5 * time.Minute,
 		HTTPClient:      azcore.DefaultHTTPClientTransport(),
@@ -90,7 +90,7 @@ type rpRegistrationPolicy struct {
 }
 
 func (r *rpRegistrationPolicy) Do(ctx context.Context, req *azcore.Request) (*azcore.Response, error) {
-	if r.options.Attempts == 0 {
+	if r.options.MaxAttempts == 0 {
 		// policy is disabled
 		return req.Next(ctx)
 	}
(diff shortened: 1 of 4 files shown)


sdk/armcore/policy_register_rp.go
```diff
@@ -32,6 +32,7 @@ type RegistrationOptions struct {
 	// Attempts is the total number of times to attempt automatic registration
 	// in the event that an attempt fails.
 	// The default value is 3.
+	// Set to zero to disable the policy.
 	Attempts int
```
tombuildsstuff, line 36: (more a general thought than for this resource) `0` is typically used in other libraries/languages to mean "unlimited" - would this make more sense as a pointer having `nil` meaning disabled?

(unrelated: I'm trying to open the GH issues we discussed last week - 🤞 later this week)

jhendrixMSFT: Interesting idea.  I picked zero to mean "disable" to align with our retry policy; i.e. setting `MaxRetries` to zero there means "disable retries" (maybe I should also rename this to `MaxAttempts` for more similarity).  Generally speaking, I'm a bit leery about infinite retries as I feel it's too easy for that to mask other bugs.  Do you think there's a compelling case for infinite retries?

tombuildsstuff: The VM Provisioning API has a hard-coded internal timeout of ~3-5 hours for a machine to finish booting (which is reasonable), at which point it's marked as "Failed" in the provisioningState.

However it's possible for a machine to finish booting /way/ later than this (for example if you're using one of the older encryption techniques, where the OS Disk is encrypted on launch on Windows - depending on the size of the OS Disk Bitlocker can take /hours/ [up to 36h iirc] to encrypt this disk upon boot [at which point it's still "booting"])

In this instance it could make sense for clients to continually poll, if we know that the 9in this case older) encryption mechanism is enabled on that VM - since the provisioningState eventually switches to "succeeded"

To be honest the provisioningState for a VM in the API should probably expose "encrypting" as a state to work around this specific bug - but these API bugs tend to be the awkward ones where this kind of thing would be useful (and today we have to write custom pollers to workaround)

jhendrixMSFT: Can you please link to one of the custom pollers you have to work around this?  I'd like to understand it better to see if there's anything we can improve in our track 2 poller implementation.

How do you feel about using `-1` to indicate infinite retries?

tombuildsstuff: this is a good one: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/maintenance/maintenance_assignment_virtual_machine_resource.go#L94-L104 - there's another I can't find right now (it's possible it's since been fixed) where the Delete API is only functionality 50% of the time, so we just keep calling Delete indefinitely until it's gone (or rather, until the context has expired - which'd ideally be the same here)

imo `-1` feels like a magic number here - given that other languages use `0` to mean unlimited I could see this causing confusion - presumably `0` attempts (e.g. in the current setup of 0 != unlimited) shouldn't be a valid configuration?

jhendrixMSFT: Agreed on the magic number bit.  Really what I was trying to do was avoid adding a `Disabled` field (trying to not have too many knobs, maybe that's misguided though).

Your comment about the context makes me wonder if we should rethink some of this.  All of these options (MaxRetries, etc) predate the existence of context.  Maybe we should simply let the context control the lifetime and be done with it?  I can discuss with the team, what do you think?

tombuildsstuff: that makes sense to me (and is essentially how we use this today, fwiw) 👍

sdk/armcore/policy_register_rp_test.go
```diff
@@ -316,3 +316,47 @@ func TestRPRegistrationPolicyCanCancel(t *testing.T) {
...
+
+func TestRPRegistrationPolicyDisabled(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	// initial response that RP is unregistered
+	srv.AppendResponse(mock.WithStatusCode(http.StatusConflict), mock.WithBody([]byte(rpUnregisteredResp)))
+	// polling responses to Register() and Get(), in progress
+	srv.RepeatResponse(5, mock.WithStatusCode(http.StatusOK), mock.WithBody([]byte(rpRegisteringResp)))
+	// polling response, successful registration
+	srv.AppendResponse(mock.WithStatusCode(http.StatusOK), mock.WithBody([]byte(rpRegisteredResp)))
```
catalinaperalta, line 328: I think this is extra you already appended 5 responses to the mock server, although I think you could just append 3 for the max attempts.

jhendrixMSFT: Good point, removed extra responses.

sdk/armcore/policy_register_rp.go
jhendrixMSFT: We discussed the idea of using context in lieu of things like max attempts.  The consensus was that it would be too easy for customers to inadvertently introduce hangs through the use of `context.Background()`, i.e. try indefinitely.  In addition, it's really masking a problem upstream that should be addressed.  That said, we get there's a need for this while the upstream problems are addressed.  With the pipeline design in track 2 you should be able to replace the default retry policy with one that uses the context as the mechanism to control the retry duration.  Would that work for you?

tombuildsstuff: maybe I'm missing something but doesn't that mean users would need to:

a) pass in a timeout for polling: https://github.com/Azure/azure-sdk-for-go/blob/track2/sdk/arm/network/2020-03-01/armnetwork/example_publicipaddress_test.go#L74
b) ensure the max attempts matches that value
c) ensure the retry duration matches that value

that appears to be (at least) 3 fields required to configure this behaviour - which given there's no base client anymore (and thus no way to make this logic reusable) feels like it's going to become a source of subtle bugs for us?

jhendrixMSFT: I see what you're saying now, I think there's a slight disconnect.  Let my clarify.

The polling interval passed to PollUntilDone() is a fallback if the service doesn't return a retry-after header.  It's the interval to wait between calls to "are you done yet" for a long-running operation.

The MaxAttempts, PollingDelay, and PollingDuration in this policy's options are used to configure how many times to attempt auto-registration of the RP, the interval to wait between calls to "are you done yet" for the registration, and the amount of time to wait before giving up on auto-registration respectively.  It's all orthogonal to polling for long-running operations (although the auto-registration is an LRO itself, which is why it needs a fair amount of knobs).

I'm still intrigued by the idea of using the context to control the lifetime of the operation, although some of the above knobs wouldn't go away (PollingDelay would have to remain, but I think the other two could be removed).  I just don't know how to reconcile the potential for hangs.  We could add a timeout to contexts that don't specify one, but we'd need to make that configurable, and now we're pretty much back to where we started.

jhendrixMSFT: Regarding the lack of a common base client.  The pipeline would be the analogous reusable component across clients.  If the default pipeline doesn't work for you, you'd create a custom one configured the way you want then call <rp>.NewClientWithPipeline(endpoint, pipeline).

jhendrixMSFT: By default, WaitForCompletionRef() will time out based on the value of client.PollingDuration (15 min is the default).  You are correct though, that if you set PollingDuration to zero the context will control the duration meaning you can have an infinite wait.  I'm ok with this as it's opt-in behavior.  This has made me realize that PollUntilDone() doesn't have such an opt-in mechanism, so passing context.Background() will infinitely poll.

Thanks for the links to code.  While I think track 2 could be made to work, I need some time to wrap my head around this to be sure.

PR #12424, opened 2020-09-07
Fix some format issues in apidiff changelog command

ArcturusZhang:
 
### DIFF
--- a/tools/apidiff/report/markdown.go
+++ b/tools/apidiff/report/markdown.go
@@ -141,7 +141,7 @@ func (t *MarkdownTable) writeHeader() {
 	}
 	t.sb.WriteString("| ")
 	t.sb.WriteString(strings.Join(t.headers, " | "))
-	t.sb.WriteString(" |\n")
+	t.sb.WriteString(" |")
 }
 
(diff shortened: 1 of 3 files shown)


tools/apidiff/report/markdown.go
```diff
@@ -141,7 +141,7 @@ func (t *MarkdownTable) writeHeader() {
 	}
 	t.sb.WriteString("| ")
 	t.sb.WriteString(strings.Join(t.headers, " | "))
-	t.sb.WriteString(" |\n")
+	t.sb.WriteString(" |")
```
jhendrixMSFT, line 144: What was the problem here?

ArcturusZhang: These changes would make the table.String() to produce a table without a linebreak at the end of string. You could regard this line break was moved to the next line's beginning.

In the MarkdownWriter, we only have a `WriteLine` method, which will append a line break at the end of string, therefore to keep consistency, I would like the `table.String()` does not have a line break at the end.

jhendrixMSFT (approved):

PR #12539, opened 2020-09-17
Add new App Service version and improve Managed Identity Credential

catalinaperalta:
This PR adds support for App Service api-version 2019-08-01 in ManagedIdentityCredential, as well as cleans up redundant code in the same credential type. 
Add missing MaxRetryDelay to MSI credential pipelines.

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -22,20 +22,22 @@ const (
 )
 
 const (
-	msiEndpointEnvironemntVariable = "MSI_ENDPOINT"
-	msiSecretEnvironemntVariable   = "MSI_SECRET"
-	appServiceMsiAPIVersion        = "2017-09-01"
-	imdsAPIVersion                 = "2018-02-01"
+	identityEndpoint = "IDENTITY_ENDPOINT"
+	identityHeader   = "IDENTITY_HEADER"
+	msiEndpoint      = "MSI_ENDPOINT"
+	msiSecret        = "MSI_SECRET"
+	imdsAPIVersion   = "2018-02-01"
 )
 
 type msiType int
 
 const (
-	msiTypeUnknown     msiType = 0
-	msiTypeIMDS        msiType = 1
-	msiTypeAppService  msiType = 2
-	msiTypeCloudShell  msiType = 3
-	msiTypeUnavailable msiType = 4
+	msiTypeUnknown             msiType = 0
+	msiTypeIMDS                msiType = 1
+	msiTypeAppServiceV20170901 msiType = 2
+	msiTypeCloudShell          msiType = 3
+	msiTypeUnavailable         msiType = 4
+	msiTypeAppServiceV20190801 msiType = 5
 )
 
 // managedIdentityClient provides the base for authenticating in managed identity environments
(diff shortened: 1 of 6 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -88,25 +90,21 @@ func newManagedIdentityClient(options *ManagedIdentityCredentialOptions) *manage
...
 // scopes: The scopes required for the token.
 func (c *managedIdentityClient) authenticate(ctx context.Context, clientID string, scopes []string) (*azcore.AccessToken, error) {
-	currentMSI, err := c.getMSIType(ctx)
-	if err != nil {
-		return nil, err
-	}
 	// This condition should never be true since getMSIType returns an error in these cases
 	// if MSI is unavailable or we were unable to determine the type return a nil access token
-	if currentMSI == msiTypeUnavailable || currentMSI == msiTypeUnknown {
+	if c.msiType == msiTypeUnavailable || c.msiType == msiTypeUnknown {
```
catalinaperalta, line 95: I think this is redundant....considering removing it

jhendrixMSFT: If the two states are conceptually the same and there's no need to differentiate the two then agreed.

sdk/azidentity/managed_identity_credential.go
```diff
@@ -48,10 +47,9 @@ type ManagedIdentityCredential struct {
 func NewManagedIdentityCredential(clientID string, options *ManagedIdentityCredentialOptions) (*ManagedIdentityCredential, error) {
 	// Create a new Managed Identity Client with default options
 	client := newManagedIdentityClient(options)
-	// Create a context that will timeout after 500 milliseconds (that is the amount of time designated to find out if the IMDS endpoint is available)
-	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Duration(client.imdsAvailableTimeoutMS)*time.Millisecond)
-	defer cancelFunc()
-	msiType, err := client.getMSIType(ctx)
+	// The context passed in here will only be used to check for the IMDS endpoint and will be set to timeout after
+	// 500 milliseconds in imdsAvailable()
+	msiType, err := client.getMSIType(context.Background())
```
jhendrixMSFT, line 52: A minor issue, I feel like this context can be removed since it's only used in `imdsAvailable()` and no other method cares about it.

sdk/azidentity/managed_identity_client.go
catalinaperalta: No the two states are distinct, I was referring to the whole check and it returning an error. I think it's sufficient with the error returned from getMSIType when it's called in the ctor.

catalinaperalta: As the comment says, the condition should never be true.

sdk/azidentity/managed_identity_client.go
```diff
@@ -154,23 +152,24 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
...
 
-func (c *managedIdentityClient) createAuthRequest(ctx context.Context, msiType msiType, clientID string, scopes []string) (*azcore.Request, error) {
-	switch msiType {
+func (c *managedIdentityClient) createAuthRequest(ctx context.Context, clientID string, scopes []string) (*azcore.Request, error) {
+	switch c.msiType {
 	case msiTypeIMDS:
 		return c.createIMDSAuthRequest(ctx, scopes)
-	case msiTypeAppService:
+	case msiTypeAppServiceV20170901:
+		return c.createAppServiceAuthRequest(ctx, clientID, scopes)
```
jhendrixMSFT, line 160: This looks to do the same thing as the other case.  If this is intentional then you can write it in one case to make it clear.

```go
case msiTypeAppServiceV20170901, msiTypeAppServiceV20190801:
```

catalinaperalta: I have been ping ponging between having two different create request methods but decided to just leave it in the same one for now. This might change in the future though if the requests end up looking more different. I'll do what you say for now.

sdk/azidentity/managed_identity_client.go
catalinaperalta: Removed

sdk/azidentity/managed_identity_credential.go
catalinaperalta: Done

sdk/azidentity/managed_identity_client.go
catalinaperalta: Done

jhendrixMSFT (approved):

PR #12590, opened 2020-09-23
Update environment credential

catalinaperalta:
This PR updates environment credential to include clientCertificateCredential and usernamePasswordCredential.

### DIFF
--- a/sdk/azidentity/environment_credential.go
+++ b/sdk/azidentity/environment_credential.go
@@ -9,7 +9,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-// NewEnvironmentCredential creates an instance of azcore.TokenCredential type and reads credential details from environment variables.
+// NewEnvironmentCredential creates an instance that implements the azcore.TokenCredential interface and reads credential details from environment variables.
 // If the expected environment variables are not found at this time, then a CredentialUnavailableError will be returned.
 // options: The options used to configure the management of the requests sent to Azure Active Directory.
 func NewEnvironmentCredential(options *TokenCredentialOptions) (azcore.TokenCredential, error) {
@@ -40,7 +40,7 @@ func NewEnvironmentCredential(options *TokenCredentialOptions) (azcore.TokenCred
 		}
 	}
 	err := &CredentialUnavailableError{CredentialType: "Environment Credential", Message: "Missing environment variable AZURE_CLIENT_SECRET or AZURE_CLIENT_CERTIFICATE_PATH or AZURE_USERNAME and AZURE_PASSWORD"}
-	azcore.Log().Write(azcore.LogError, logCredentialError(err.CredentialType, err))
+	logCredentialError(err.CredentialType, err)
 	return nil, err
 
 }
--- a/sdk/azidentity/environment_credential.go
+++ b/sdk/azidentity/environment_credential.go
@@ -9,30 +9,38 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-// NewEnvironmentCredential creates an instance of the ClientSecretCredential type and reads credential details from environment variables.
+// NewEnvironmentCredential creates an instance of azcore.TokenCredential type and reads credential details from environment variables.
 // If the expected environment variables are not found at this time, then a CredentialUnavailableError will be returned.
 // options: The options used to configure the management of the requests sent to Azure Active Directory.
-func NewEnvironmentCredential(options *TokenCredentialOptions) (*ClientSecretCredential, error) {
+func NewEnvironmentCredential(options *TokenCredentialOptions) (azcore.TokenCredential, error) {
 	tenantID := os.Getenv("AZURE_TENANT_ID")
 	if tenantID == "" {
 		err := &CredentialUnavailableError{CredentialType: "Environment Credential", Message: "Missing environment variable AZURE_TENANT_ID"}
 		logCredentialError(err.CredentialType, err)
 		return nil, err
 	}
-
 	clientID := os.Getenv("AZURE_CLIENT_ID")
 	if clientID == "" {
 		err := &CredentialUnavailableError{CredentialType: "Environment Credential", Message: "Missing environment variable AZURE_CLIENT_ID"}
 		logCredentialError(err.CredentialType, err)
 		return nil, err
 	}
-
-	clientSecret := os.Getenv("AZURE_CLIENT_SECRET")
-	if clientSecret == "" {
-		err := &CredentialUnavailableError{CredentialType: "Environment Credential", Message: "Missing environment variable AZURE_CLIENT_SECRET"}
-		logCredentialError(err.CredentialType, err)
-		return nil, err
+	if clientSecret := os.Getenv("AZURE_CLIENT_SECRET"); clientSecret != "" {
+		azcore.Log().Write(LogCredential, "Azure Identity => NewEnvironmentCredential() invoking ClientSecretCredential")
+		return NewClientSecretCredential(tenantID, clientID, clientSecret, options)
 	}
-	azcore.Log().Write(LogCredential, "Azure Identity => NewEnvironmentCredential() invoking ClientSecretCredential")
-	return NewClientSecretCredential(tenantID, clientID, clientSecret, options)
+	if clientCertificate := os.Getenv("AZURE_CLIENT_CERTIFICATE_PATH"); clientCertificate != "" {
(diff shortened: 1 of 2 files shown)


sdk/azidentity/environment_credential.go
```diff
@@ -9,30 +9,38 @@ import (
...
+		azcore.Log().Write(LogCredential, "Azure Identity => NewEnvironmentCredential() invoking ClientCertificateCredential")
+		return NewClientCertificateCredential(tenantID, clientID, clientCertificate, nil, options)
+	}
+	if username := os.Getenv("AZURE_USERNAME"); username != "" {
+		if password := os.Getenv("AZURE_PASSWORD"); password != "" {
+			azcore.Log().Write(LogCredential, "Azure Identity => NewEnvironmentCredential() invoking UsernamePasswordCredential")
+			return NewUsernamePasswordCredential(tenantID, clientID, username, password, options)
+		}
+	}
+	err := &CredentialUnavailableError{CredentialType: "Environment Credential", Message: "Missing environment variable AZURE_CLIENT_SECRET or AZURE_CLIENT_CERTIFICATE_PATH or AZURE_USERNAME and AZURE_PASSWORD"}
```
jhendrixMSFT, line 42: We already log the env vars found before this right, so between that and this error message is should be clear what's missing?

sdk/azidentity/environment_credential.go
```diff
@@ -9,30 +9,38 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-// NewEnvironmentCredential creates an instance of the ClientSecretCredential type and reads credential details from environment variables.
+// NewEnvironmentCredential creates an instance that implements the azcore.TokenCredential interface and reads credential details from environment variables.
 // If the expected environment variables are not found at this time, then a CredentialUnavailableError will be returned.
 // options: The options used to configure the management of the requests sent to Azure Active Directory.
-func NewEnvironmentCredential(options *TokenCredentialOptions) (*ClientSecretCredential, error) {
+func NewEnvironmentCredential(options *TokenCredentialOptions) (azcore.TokenCredential, error) {
```
jhendrixMSFT, line 15: Does this change need to be made elsewhere?  Also, do we need to continue to export ClientSecretCredential?

catalinaperalta: All of our credential types are exported. Only NewEnvironmentCredential and NewDefaultAzureCredential arent types because they return other credentials. I can see a case for return azcore.TokenCredential from all of the constructors if we dont want to export any credential types.

sdk/azidentity/environment_credential.go
catalinaperalta: Yes but we also log when a credential returns an error and this is the error returned from NewEnvironmentCredential. I do think it's hard to miss it, particularly if the user is checking the error that is returned, but adding the error to the log completes the story for this credential dont you think?

jhendrixMSFT: So we log what was found, and here (below really) we log what was missing so it sounds like the caller should be able to figure out from the logs which env vars they need to add.

sdk/azidentity/environment_credential.go
jhendrixMSFT: Having the constructors return interfaces gives us the ability to change the underlying implementation without affecting callers.  It also allows mocking, especially given that all these credential types don't export any content.

catalinaperalta: Yeah I think that could provide us some flexibility, but I do think we need to get @JeffreyRichter to approve on this idea as well. @JeffreyRichter, what do you think of having all of the constructors return azcore.TokenCredential and unexporting the credential types?

sdk/azidentity/environment_credential.go
jhendrixMSFT: No keep logging the error.  I just want to ensure that the log messages are useful in tracking down what env vars are missing.  E.g. I'm using client secret env vars, if the previous log message states that it found the tenant, subscription, and client ID env vars, and this error logs that client secret (and other) env vars are missing it should be clear that I omitted (or misspelled) the env var for the client secret.

sdk/azidentity/environment_credential.go
jhendrixMSFT: Thinking about this more, there are some advantages to returning `ClientSecretCredential`.  E.g. today it supports `azcore.TokenCredential`.  Say we end up with some new credential interface type in `azcore`.  If `ClientSecretCredential` can support both then this constructor continues to be useful for both cases.

The mocking of credentials can always be done for the API consuming the credential as it takes an interface, so I think that's less useful here.

sdk/azidentity/environment_credential.go
catalinaperalta: Ah I understood you know 😅. Yes before we log the env vars that are found and here we log those that are missing. Do you think I should expand on the message? Saying please make sure you include one of these and that the env var is spelled correctly? Something along those lines

jhendrixMSFT: Ok so it sounds like we both agree the logs produce enough info to help diagnose this error.

I'm fine with the message as-is but feel free to modify it if you think it would help.

sdk/azidentity/environment_credential.go
catalinaperalta: True, the current implementation is better if we ever want to check credential types with interfaces other than TokenCredential. But then, is it a good idea to return azcore.TokenCredential from NewEnvironmentCredential?

jhendrixMSFT: You might want to leave it as `ClientSecretCredential` for now until we discuss further (can always switch to interfaces later).

sdk/azidentity/environment_credential.go
catalinaperalta: My only doubt with the error message as is, is if it's clear enough that AZURE_USERNAME and AZURE_PASSWORD both need to be present at the same time?

sdk/azidentity/environment_credential.go
catalinaperalta: Hmmm, well the whole point of this change was to add support for ClientCertificateCredential and UsernamePasswordCredential from the env credential...I'll leave this PR open while we discuss

sdk/azidentity/environment_credential.go
jhendrixMSFT: Good point.  Maybe customize the message for that case.

sdk/azidentity/environment_credential.go
jhendrixMSFT: Derp my bad, I overlooked that part.

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -26,13 +26,13 @@ func TestChainedTokenCredential_InstantiateSuccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not find appropriate environment credentials")
 	}
-	cred, err := NewChainedTokenCredential(secCred, envCred)
+	cred, err := NewChainedTokenCredential(secCred.(azcore.TokenCredential), envCred.(azcore.TokenCredential))
```
jhendrixMSFT, line 29: This is unfortunate.  It means that when callers create a chained token credential they have to perform this type-assertion.  I feel like this is a bit of a step backwards.

jhendrixMSFT: This makes me wonder if this abstraction is wrong.  Should it be `NewChainedCredential()` instead, that takes multiple `azcore.Credential` instead?

catalinaperalta: I agree, this makes ChainedTokenCredential harder to use and it's definitely not ideal. I would not like to leave it like this. But the abstraction for ChainedTokenCredential is correct in being `azcore.TokenCredential` since it is expecting a chain of only TokenCredentials with the GetToken method.

jhendrixMSFT: Agreed that `ChainedTokenCredential()` takes a slice of `azcore.TokenCredential`.  With this change I'm suggesting it should be `ChainedCredential()` that takes a slice of `azcore.Credential`.  How does a customer know they need to perform this type assertion?

Another option would be to leave it as `ChainedTokenCredential()` that takes a slice of `azcore.Credential` and perform the type-assertion in the constructor, returning an error if it fails.  I don't like this though as you lose the type safety.

I still think the best option is to return concrete types from the credential constructors as it allows them to be covariant.

jhendrixMSFT (approved):

PR #12635, opened 2020-09-30
Adding interactive browser credential to azidentity

catalinaperalta:
This PR adds interactive browser credential along with its tests to azidentity

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -384,10 +384,11 @@ func (c *aadIdentityClient) createDeviceCodeNumberRequest(ctx context.Context, t
 
 // authenticateInteractiveBrowser opens an interactive browser window, gets the authorization code and requests an Access Token with the
 // authorization code and returns the token or an error in case of authentication failure.
-// ctx: The current request context
-// tenantID: The Azure Active Directory tenant (directory) ID of the service principal
-// clientID: The client (application) ID of the service principal
+// ctx: The current request context.
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal.
+// clientID: The client (application) ID of the service principal.
 // clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// redirectURI: The redirect URI that was used to request the authorization code. Must be the same URI that is configured for the App Registration.
 // scopes: The scopes required for the token
 func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, tenantID string, clientID string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
 	cfg, err := authCodeReceiver(c.options.AuthorityHost, tenantID, clientID, redirectURI, scopes)
@@ -398,10 +399,12 @@ func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, 
 }
 
 // authenticateAuthCode requests an Access Token with the authorization code and returns the token or an error in case of authentication failure.
-// ctx: The current request context
-// tenantID: The Azure Active Directory tenant (directory) ID of the service principal
-// clientID: The client (application) ID of the service principal
+// ctx: The current request context.
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal.
+// clientID: The client (application) ID of the service principal.
+// authCode: The authorization code received from the authorization code flow. The authorization code must not have been used to obtain another token.
 // clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// redirectURI: The redirect URI that was used to request the authorization code. Must be the same URI that is configured for the App Registration.
 // scopes: The scopes required for the token
 func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
 	req, err := c.createAuthorizationCodeAuthRequest(ctx, tenantID, clientID, authCode, clientSecret, redirectURI, scopes)
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -390,7 +390,7 @@ func (c *aadIdentityClient) createDeviceCodeNumberRequest(ctx context.Context, t
 // clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
 // redirectURI: The redirect URI that was used to request the authorization code. Must be the same URI that is configured for the App Registration.
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, tenantID string, clientID string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
+func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, tenantID string, clientID string, clientSecret *string, redirectURI *string, scopes []string) (*azcore.AccessToken, error) {
 	cfg, err := authCodeReceiver(c.options.AuthorityHost, tenantID, clientID, redirectURI, scopes)
 	if err != nil {
 		return nil, err
@@ -406,7 +406,7 @@ func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, 
 // clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
 // redirectURI: The redirect URI that was used to request the authorization code. Must be the same URI that is configured for the App Registration.
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
+func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret *string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
 	req, err := c.createAuthorizationCodeAuthRequest(ctx, tenantID, clientID, authCode, clientSecret, redirectURI, scopes)
 	if err != nil {
 		return nil, err
@@ -425,12 +425,12 @@ func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID s
 }
 
 // createAuthorizationCodeAuthRequest creates a request for an Access Token for authorization_code grant types.
-func (c *aadIdentityClient) createAuthorizationCodeAuthRequest(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.Request, error) {
+func (c *aadIdentityClient) createAuthorizationCodeAuthRequest(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret *string, redirectURI string, scopes []string) (*azcore.Request, error) {
 	data := url.Values{}
 	data.Set(qpGrantType, "authorization_code")
 	data.Set(qpClientID, clientID)
-	if clientSecret != "" {
-		data.Set(qpClientSecret, clientSecret) // only for web apps
+	if clientSecret != nil {
+		data.Set(qpClientSecret, *clientSecret) // only for web apps
 	}
 	data.Set(qpRedirectURI, redirectURI)
 	data.Set(qpScope, strings.Join(scopes, " "))
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -389,12 +389,22 @@ func (c *aadIdentityClient) createDeviceCodeNumberRequest(ctx context.Context, t
 // clientID: The client (application) ID of the service principal
 // clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
 // scopes: The scopes required for the token
-func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, tenantID string, clientID string, clientSecret string, scopes []string) (*azcore.AccessToken, error) {
-	cfg, err := authCodeReceiver(tenantID, clientID, scopes)
+func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, tenantID string, clientID string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
+	cfg, err := authCodeReceiver(c.options.AuthorityHost, tenantID, clientID, redirectURI, scopes)
 	if err != nil {
 		return nil, err
 	}
-	req, err := c.createAuthorizationCodeAuthRequest(ctx, *cfg, tenantID, clientID, clientSecret, scopes)
+	return c.authenticateAuthCode(ctx, tenantID, clientID, cfg.authCode, clientSecret, cfg.redirectURI, scopes)
+}
+
+// authenticateAuthCode requests an Access Token with the authorization code and returns the token or an error in case of authentication failure.
+// ctx: The current request context
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal
+// clientID: The client (application) ID of the service principal
+// clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// scopes: The scopes required for the token
+func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
+	req, err := c.createAuthorizationCodeAuthRequest(ctx, tenantID, clientID, authCode, clientSecret, redirectURI, scopes)
 	if err != nil {
 		return nil, err
 	}
@@ -412,16 +422,16 @@ func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, 
 }
 
 // createAuthorizationCodeAuthRequest creates a request for an Access Token for authorization_code grant types.
-func (c *aadIdentityClient) createAuthorizationCodeAuthRequest(ctx context.Context, config interactiveConfig, tenantID string, clientID string, clientSecret string, scopes []string) (*azcore.Request, error) {
+func (c *aadIdentityClient) createAuthorizationCodeAuthRequest(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.Request, error) {
 	data := url.Values{}
 	data.Set(qpGrantType, "authorization_code")
 	data.Set(qpClientID, clientID)
 	if clientSecret != "" {
 		data.Set(qpClientSecret, clientSecret) // only for web apps
 	}
-	data.Set(qpRedirectURI, config.redirectURI)
+	data.Set(qpRedirectURI, redirectURI)
 	data.Set(qpScope, strings.Join(scopes, " "))
-	data.Set(qpCode, config.authCode)
+	data.Set(qpCode, authCode)
 	dataEncoded := data.Encode()
 	body := azcore.NopCloser(strings.NewReader(dataEncoded))
 	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(c.options.AuthorityHost, tenantID, tokenEndpoint))
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -7,10 +7,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
+
+	"github.com/pkg/browser"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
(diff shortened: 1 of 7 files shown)


sdk/azidentity/aad_identity_client.go
```diff
@@ -373,3 +384,99 @@ func (c *aadIdentityClient) createDeviceCodeNumberRequest(ctx context.Context, t
...
+	}
+	if clientID == "" {
+		clientID = "04b07795-8ddb-461a-bbee-02f9e1bf7b46"
+	}
+	state := func() string {
+		// generate a 20-char random alpha-numeric string
+		const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+		buff := make([]byte, 20)
+		for i := range buff {
+			buff[i] = charset[rand.Intn(len(charset))]
```
jhendrixMSFT, line 432: Be sure to seed `rand` first.
```go
rand.Seed(time.Now().Unix())
```

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/pkg/browser"
+)
+
+// InteractiveBrowserCredential enables authentication to Azure Active Directory using an interactive browser to log in.
+type InteractiveBrowserCredential struct {
+	client       *aadIdentityClient
+	tenantID     string // Gets the Azure Active Directory tenant (directory) ID of the service principal
+	clientID     string // Gets the client (application) ID of the service principal
+	clientSecret string // Gets the client secret that was generated for the App Registration used to authenticate the client.
```
catalinaperalta, line 21: I'm wondering if it's better to have clientSecret in an options struct for this credential? 

On a related, but more general note, shouldn't all of the credentials have an options struct to help for future additions?

sdk/azidentity/aad_identity_client.go

sdk/azidentity/interactive_browser_credential.go
jhendrixMSFT: If clientSecret is optional then yeah it should go in an options struct (probably fold TokenCredentialOptions into this same struct).

IIRC, we added TokenCredentialOptions as the escape-hatch for optional parameters.  However, if we think it makes more sense to have per-ctor options then can we do away with TokenCredentialOptions?  Having a bunch of nested optional param structs seems a bit overkill.

catalinaperalta: Yes we had decided to have TokenCredentialOptions as the escape-hatch...but thinking about it more I dont think it's the best solution. If I were to add clientSecret to TokenCredentialOptions then I think it would be confusing when these same options are used for other credentials that require clientSecret. So I'm leaning towards each credential having its own option type.

jhendrixMSFT: Yes I see your point.

I forgot how many options are in TokenCredentialOptions so perhaps it needs to remain, although it might be better to not anonymously embed it in other options structs (i.e. make it another nil-able field).

catalinaperalta: Done

sdk/azidentity/go.mod
```diff
@@ -5,5 +5,7 @@ go 1.13
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.11.0
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.3.0
+	github.com/Azure/azure-sdk-for-go/sdk/to v0.1.1
+	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
```
jhendrixMSFT, line 9: We will need to add this to the list of approved 3rd party dependencies.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/pkg/browser"
+)
+
+// InteractiveBrowserCredentialOptions can be used when providing additional credential information, such as a client secret.
+// Also use these options to modify the default pipeline behavior through the TokenCredentialOptions.
+type InteractiveBrowserCredentialOptions struct {
+	ClientSecret *string                 // Gets the client secret that was generated for the App Registration used to authenticate the client. Only applies for web apps.
```
jhendrixMSFT, line 20: Doc comments should go above the field in the same format as the outer doc comment, i.e. `// ClientSecret ...`.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/pkg/browser"
+)
+
+// InteractiveBrowserCredentialOptions can be used when providing additional credential information, such as a client secret.
+// Also use these options to modify the default pipeline behavior through the TokenCredentialOptions.
+type InteractiveBrowserCredentialOptions struct {
```
jhendrixMSFT, line 19: Presumably we want to add a `DefaultInteractiveBrowserCredentialOptions()` func like we did in `azcore`.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+	clientID     string  // Gets the client (application) ID of the service principal
+	clientSecret *string // Gets the client secret that was generated for the App Registration used to authenticate the client.
+}
+
+// NewInteractiveBrowserCredential constructs a new InteractiveBrowserCredential with the details needed to authenticate against Azure Active Directory through an interactive browser window.
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal.
+// clientID: The client (application) ID of the service principal.
+// clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// options: allow to configure the management of the requests sent to Azure Active Directory.
+func NewInteractiveBrowserCredential(tenantID string, clientID string, options *InteractiveBrowserCredentialOptions) (*InteractiveBrowserCredential, error) {
```
schaabs, line 37: `tenantId` and `clientId` should be moved to `options` as well, the are not required. `tenantId` should default to organizations, and `clientId` should default to the developer sign-on client id.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+var authCodeReceiver = func(tenantID string, clientID string, scopes []string) (*interactiveConfig, error) {
+	return interactiveBrowserLogin(tenantID, clientID, scopes)
+}
+
+// interactiveBrowserLogin opens an interactive browser with the specified tenant and client IDs provided then returns the authorization code
+// received or an error.
+func interactiveBrowserLogin(tenantID string, clientID string, scopes []string) (*interactiveConfig, error) {
+	const authURLFormat = "https://login.microsoftonline.com/%s/oauth2/v2.0/authorize?response_type=code&response_mode=query&client_id=%s&redirect_uri=%s&state=%s&scope=%s&prompt=select_account"
+	if tenantID == "" {
+		tenantID = "common"
```
schaabs, line 80: tenantID should default to "organizations" not "common"

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+var _ azcore.TokenCredential = (*InteractiveBrowserCredential)(nil)
+
+var authCodeReceiver = func(tenantID string, clientID string, scopes []string) (*interactiveConfig, error) {
+	return interactiveBrowserLogin(tenantID, clientID, scopes)
+}
+
+// interactiveBrowserLogin opens an interactive browser with the specified tenant and client IDs provided then returns the authorization code
+// received or an error.
+func interactiveBrowserLogin(tenantID string, clientID string, scopes []string) (*interactiveConfig, error) {
+	const authURLFormat = "https://login.microsoftonline.com/%s/oauth2/v2.0/authorize?response_type=code&response_mode=query&client_id=%s&redirect_uri=%s&state=%s&scope=%s&prompt=select_account"
```
schaabs, line 78: The authority host should be configurable, this only will work for public cloud.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,117 @@
...
+		const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+		buff := make([]byte, 20)
+		for i := range buff {
+			buff[i] = charset[rand.Intn(len(charset))]
+		}
+		return string(buff)
+	}()
+	// start local redirect server so login can call us back
+	rs := newServer()
+	redirectURL := rs.Start(state)
```
schaabs, line 97: The redirectURL needs to be configurable to support the case that a user is using their own App Registration.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: I dont think so in this case. The defaults are applied automatically if nil is specified for the options. I could add a comment saying so.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
jhendrixMSFT: The idea is that if a user wants to configure one piece, they call `DefaultXXXOptions()` to get a pre-populated options struct with default values then change the part(s) they care about.

catalinaperalta: So I've got a problem with this optional tenantID and optional clientID, because if we provide the defaults ("organizations" and the developer sign on client ID) if a user were to change only one, wouldnt it be broken?

sdk/azidentity/aad_identity_client.go
```diff
@@ -373,3 +381,65 @@ func (c *aadIdentityClient) createDeviceCodeNumberRequest(ctx context.Context, t
...
+	return c.authenticateAuthCode(ctx, tenantID, clientID, cfg.authCode, clientSecret, cfg.redirectURI, scopes)
+}
+
+// authenticateAuthCode requests an Access Token with the authorization code and returns the token or an error in case of authentication failure.
+// ctx: The current request context
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal
+// clientID: The client (application) ID of the service principal
+// clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// scopes: The scopes required for the token
+func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, redirectURI string, scopes []string) (*azcore.AccessToken, error) {
```
catalinaperalta, line 406: Separating this because I'm adding an AuthorizationCodeCredential once this is merged.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done, thanks for the info Scott!

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Also done

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,150 @@
...
+// NewInteractiveBrowserCredential constructs a new InteractiveBrowserCredential with the details needed to authenticate against Azure Active Directory through an interactive browser window.
+// tenantID: The Azure Active Directory tenant (directory) ID of the service principal.
+// clientID: The client (application) ID of the service principal.
+// clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// options: allow to configure the management of the requests sent to Azure Active Directory.
+func NewInteractiveBrowserCredential(options *InteractiveBrowserCredentialOptions) (*InteractiveBrowserCredential, error) {
+	var credentialOptions *TokenCredentialOptions
+	tenantID := "organizations"
+	clientID := developerSignOnClientID
+	redirectURI := ""
```
jhendrixMSFT, line 54: This, and the above defaults, should all go in a `DefaultInteractiveBrowserCredentialOptions()` method.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,150 @@
...
+
+// AuthenticationPolicy implements the azcore.Credential interface on InteractiveBrowserCredential and calls the Bearer Token policy
+// to get the bearer token.
+func (c *InteractiveBrowserCredential) AuthenticationPolicy(options azcore.AuthenticationPolicyOptions) azcore.Policy {
+	return newBearerTokenPolicy(c, options)
+}
+
+var _ azcore.TokenCredential = (*InteractiveBrowserCredential)(nil)
+
+var authCodeReceiver = func(authorityHost string, tenantID string, clientID string, redirectURI string, scopes []string) (*interactiveConfig, error) {
```
jhendrixMSFT, line 102: Add a comment indicating this was done so it could be modified for testing purposes.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,150 @@
...
+	return interactiveBrowserLogin(authorityHost, tenantID, clientID, redirectURI, scopes)
+}
+
+// interactiveBrowserLogin opens an interactive browser with the specified tenant and client IDs provided then returns the authorization code
+// received or an error.
+func interactiveBrowserLogin(authorityHost string, tenantID string, clientID string, redirectURL string, scopes []string) (*interactiveConfig, error) {
+	const authURLFormat = "%s/%s/oauth2/v2.0/authorize?response_type=code&response_mode=query&client_id=%s&redirect_uri=%s&state=%s&scope=%s&prompt=select_account"
+	if tenantID == "" {
+		tenantID = "organizations"
```
jhendrixMSFT, line 111: Didn't this already get set elsewhere?

jhendrixMSFT: Same comment applies to `clientID` below.

sdk/azidentity/interactive_browser_credential_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}, nil
+	}
+	_, err = cred.GetToken(context.Background(), azcore.TokenRequestOptions{Scopes: []string{scope}})
+	if err == nil {
+		t.Fatalf("Expected an error but did not receive one.")
+	}
+	var authFailed *AuthenticationFailedError
+	if !errors.As(err, &authFailed) {
+		t.Fatalf("Expected: AuthenticationFailedError, Received: %T", err)
+	} else {
```
jhendrixMSFT, line 61: You can remove this `else` clause as the `t.Fatal()` above will terminate the test.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,150 @@
...
+// clientID: The client (application) ID of the service principal.
+// clientSecret: Gets the client secret that was generated for the App Registration used to authenticate the client.
+// options: allow to configure the management of the requests sent to Azure Active Directory.
+func NewInteractiveBrowserCredential(options *InteractiveBrowserCredentialOptions) (*InteractiveBrowserCredential, error) {
+	var credentialOptions *TokenCredentialOptions
+	tenantID := "organizations"
+	clientID := developerSignOnClientID
+	redirectURI := ""
+	var clientSecret string
+	if options != nil {
```
jhendrixMSFT, line 56: Following the `DefaultXXXOptions()` pattern above, you can then make a copy of `options` and stash it in `InteractiveBrowserCredential` instead of doing all the following copying.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential_test.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: I'm not sure what you mean. I've simplified it now with the options struct, let me know if that's better?

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,135 @@
...
+	clientID string
+	// Gets the client secret that was generated for the App Registration used to authenticate the client.
+	clientSecret string
+	// The redirect URI used to request the authorization code. Must be the same URI that is configured for the App Registration.
+	redirectURI string
+}
+
+// DefaultInteractiveBrowserCredentialOptions use this function to get the default values for logging in with
+// an interactive browser window.
+func DefaultInteractiveBrowserCredentialOptions() *InteractiveBrowserCredentialOptions {
```
jhendrixMSFT, line 47: This should return by value as it can never fail (it's the pattern all the other `DefaultOptions()` use).

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,135 @@
...
+	tenantID, clientID, clientSecret, redirectURI := organizationsTenantID, developerSignOnClientID, "", ""
+	return &InteractiveBrowserCredentialOptions{
+		TenantID:     &tenantID,
+		ClientID:     &clientID,
+		ClientSecret: &clientSecret,
```
jhendrixMSFT, line 52: Can't this and `RedirectURI` remain `nil` for their default values?

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,135 @@
...
+	// The redirect URI used to request the authorization code. Must be the same URI that is configured for the App Registration.
+	redirectURI string
+}
+
+// DefaultInteractiveBrowserCredentialOptions use this function to get the default values for logging in with
+// an interactive browser window.
+func DefaultInteractiveBrowserCredentialOptions() *InteractiveBrowserCredentialOptions {
+	tenantID, clientID, clientSecret, redirectURI := organizationsTenantID, developerSignOnClientID, "", ""
+	return &InteractiveBrowserCredentialOptions{
+		TenantID:     &tenantID,
```
jhendrixMSFT, line 50: Is it ever legal for `TenantID` or `ClientID` to be `nil`/empty string?  If they always have to have _some_ value (which the above initialization alludes to) then they probably shouldn't be pointer-to-string.

sdk/azidentity/interactive_browser_credential.go
jhendrixMSFT: To clarify, you could simply have a copy of the `options` struct in `InteractiveBrowserCredential` instead of copying all the individual fields.  That way, if more options show up they are automatically included.

catalinaperalta: I think that's fine, I'll just point out that that means carrying around the TokenCredentialOptions as well...is it terrible? I dont think so, but we would be carrying around this extra data.

jhendrixMSFT: A fair point.  You could set the `Options` field to `nil` in that case.

I don't think it's a big deal either way, so if you want to leave it as-is that's fine by me.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done, changed it so they can be nil

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done, changed it to hold the options struct

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,128 @@
...
+		ClientSecret: nil,
+		RedirectURI:  nil,
+	}
+}
+
+// NewInteractiveBrowserCredential constructs a new InteractiveBrowserCredential with the details needed to authenticate against Azure Active Directory through an interactive browser window.
+// options: allow to configure the management of the requests sent to Azure Active Directory, pass in nil for default behavior.
+func NewInteractiveBrowserCredential(options *InteractiveBrowserCredentialOptions) (*InteractiveBrowserCredential, error) {
+	if options == nil {
+		*options = DefaultInteractiveBrowserCredentialOptions()
```
jhendrixMSFT, line 54: This will panic if `options` is `nil`.  You'll need to capture the return value in a local then take its address.

jhendrixMSFT: Probably need a test to cover this case.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -0,0 +1,128 @@
...
+		buff := make([]byte, 20)
+		for i := range buff {
+			buff[i] = charset[rand.Intn(len(charset))]
+		}
+		return string(buff)
+	}()
+	// start local redirect server so login can call us back
+	rs := newServer()
+	if redirectURL == nil {
+		*redirectURL = rs.Start(state)
```
jhendrixMSFT, line 107: Will panic if `nil`.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

jhendrixMSFT (approved):

PR #12641, opened 2020-10-01
* fix id missing for the consumption request in next paging request

hotsnow:
- [Y] The purpose of this PR is explained in this or a referenced issue.
- [N] The PR does not update generated files.
- [N] Tests are included and/or updated for code changes.
- [N] Updates to [CHANGELOG.md][] are included.
- [N] Apache v2 license headers are included in each file.
 

jhendrixMSFT:
Thanks for putting this together.

Note that this SDK is generated from [Azure swagger files](https://github.com/Azure/azure-rest-api-specs).  What this means is that any manually made changes to the generated code will be lost the next time the SDK is regenerated.  We're investigating ways to incorporate hand-written code with our generated code but have nothing ready at this time.

So while I would like to accept this, it will be lost in the next update.

PR #12697, opened 2020-10-09
Adding options struct to credential types

catalinaperalta:
This PR adds options structs to ClientCertificateCredential.


catalinaperalta:
I'm torn between adding an options bag for all credentials even the ones that dont need it at the moment or if that adds an additional (annoying?) layer when creating credentials. Thoughts?

jhendrixMSFT:
I feel your pain, it bothers me too.  However, if we ever want to add any options later on it means we either introduce a breaking change (unlikely), or a new constructor e.g. `NewFooThingWithNewOption()` which isn't so great either (although the Go community accepts this pattern, given the limitations of the language).

Given that our design guidelines state that we use this options bag, even if it's empty, I think adding it now gives us future-proofing (we are doing this in generated code).

jhendrixMSFT:
Personally, I would have preferred us to use variadic configuration for hand-written code as it composes well and lets us add new options.  But that ship has sailed.

catalinaperalta:
> Given that our design guidelines state that we use this options bag, even if it's empty, I think adding it now gives us future-proofing (we are doing this in generated code).

You're right, ok I'll add the other missing options structs for the other credentials. Agreed the variadic configurations would also have been very flexible for cases like these.

sdk/azidentity/client_certificate_credential.go
```diff
@@ -27,13 +27,28 @@ type ClientCertificateCredential struct {
...
+// password: The password required to decrypt the private key.  Pass nil if there is no password.
+// options: Manage the configuration of requests sent to Azure Active Directory.
+type ClientCertificateCredentialOptions struct {
+	// The password required to decrypt the private key.  Pass nil if there is no password.
+	Password *string
+	// Manage the configuration of requests sent to Azure Active Directory.
+	Options *TokenCredentialOptions
+}
+
+// DefaultClientCertificateCredentialOptions returns an instance of ClientCertificateCredentialOptions with the default values set.
```
jhendrixMSFT, line 40: "... returns an instance of ClientCertificateCredentialOptions initialized with default values."

This is the doc string I've been using across all of these `DefaultFooOptions()` funcs (comment applies to other funcs).

jhendrixMSFT (approved):
Few tweaks to doc strings.

sdk/azidentity/client_certificate_credential.go
catalinaperalta: Done

PR #12709, opened 2020-10-10
ResouceID -> ResourceID

tony-xia:


ArcturusZhang:
Hi @tony-xia thanks for contributing!

Unfortunately everything under `services` directory is auto-generated using autorest from the [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/blob/c42e6f0540dc41a1d442c103bbb6f0ba71acf95b/specification/applicationinsights/resource-manager/Microsoft.Insights/stable/2015-05-01/componentFeaturesAndPricing_API.json#L382), therefore any changes may be overridden next time we generate our SDK. 

And as for this typo, it first exists [here](https://github.com/Azure/azure-rest-api-specs/blob/c42e6f0540dc41a1d442c103bbb6f0ba71acf95b/specification/applicationinsights/resource-manager/Microsoft.Insights/stable/2015-05-01/componentFeaturesAndPricing_API.json#L401) and thus it is introduced in the SDK.

Since this could not be merged, I would close this PR and I hope you would not mind.

Could you please open an issue for this in the [azure-rest-api-specs repo](https://github.com/Azure/azure-rest-api-specs/issues)?

PR #12726, opened 2020-10-12
Add NewTLSHTTP2Server

catalinaperalta:
This PR adds a function for creating a TLS server with HTTP2 configuration and a method to access the http.Server configured for the test server.


catalinaperalta:
I'm open to just leaving NewTLSServer with the HTTP2 config. Let me know

sdk/internal/mock/mock.go
```diff
@@ -58,6 +60,27 @@ func NewTLSServer() (*Server, func()) {
 	return s, func() { s.srv.Close() }
 }
 
+// NewTLSHTTP2Server creates a new Server object with HTTP2 configuration.
+// The returned close func must be called when the server is no longer needed.
+// It will return nil for both the server and close func if it encountered an error
+// when configuring HTTP2 for the new TLS server.
+func NewTLSHTTP2Server() (*Server, func()) {
```
jhendrixMSFT, line 67: Instead of adding a new constructor, how about making the existing ones use variadic configuration?  It seems likely we'll want to add more configuration options in the future.

sdk/internal/mock/mock.go
```diff
@@ -58,6 +60,27 @@ func NewTLSServer() (*Server, func()) {
...
+	s := newServer()
+	s.srv = httptest.NewUnstartedServer(http.HandlerFunc(s.serveHTTP))
+	if err := http2.ConfigureServer(s.srv.Config, new(http2.Server)); err != nil {
+		return nil, nil
```
jhendrixMSFT, line 71: Since the constructors are no-fail, I think it would be better to `panic(err)` here.  This way the context of the failure isn't lost, plus returning a `nil` `Server` will panic later anyways.

sdk/internal/mock/mock.go
```diff
@@ -58,6 +60,27 @@ func NewTLSServer() (*Server, func()) {
...
+	}
+	s.srv.TLS = s.srv.Config.TLSConfig
```
jhendrixMSFT, line 73: Presumably this is where the magic happens?

sdk/internal/mock/mock.go
```diff
@@ -58,6 +60,27 @@ func NewTLSServer() (*Server, func()) {
...
+	s.srv.StartTLS()
+	return s, func() { s.srv.Close() }
+}
+
+// ServerConfig returns the server config. Please note this should not be
+// modified since Start or StartTLS has already been called on the server.
+func (s *Server) ServerConfig() *http.Server {
```
jhendrixMSFT, line 80: I assume this is exposed so tests can access it.  Curious which part(s) needed to fix the issue you were working on?

sdk/internal/mock/mock.go
catalinaperalta: Yeah makes sense, let me look into it.

sdk/internal/mock/mock.go
catalinaperalta: Yes, the test looks something like this
```
srv, close := mock.NewTLSHTTP2Server()
	defer close()
	tr := &http.Transport{TLSClientConfig: srv.ServerConfig().TLSConfig}
	if err := http2.ConfigureTransport(tr); err != nil {
		t.Fatalf("Failed to configure http2 transport: %v", err)
	}
	tr.TLSClientConfig.InsecureSkipVerify = true
	client := &http.Client{Transport: tr}
	srv.AppendResponse(mock.WithBody([]byte(accessTokenRespSuccess)))
	options := DefaultInteractiveBrowserCredentialOptions()
	options.Options = &TokenCredentialOptions{AuthorityHost: srv.URL(), HTTPClient: client}
```

catalinaperalta: I'm following the "proper" way to configure a new server, according to this: https://github.com/golang/go/issues/34939

sdk/internal/mock/mock.go
```diff
@@ -50,14 +50,27 @@ func NewServer() (*Server, func()) {
...
 }
 
-// NewTLSServer creates a new Server object.
+// NewTLSServer creates a new Server object and applies any additional configurations
+// provided on the httptest.Server.
 // The returned close func must be called when the server is no longer needed.
-func NewTLSServer() (*Server, func()) {
+// It will return nil for both the server and close func if it encountered an error
+// when configuring HTTP2 for the new TLS server.
+func NewTLSServer(config ...func(*httptest.Server)) (*Server, func()) {
```
jhendrixMSFT, line 58: This is a great start, but limits us to configuring `*httptest.Server` only.  Take a look at `ResponseOption`, if we apply that same pattern it provides more flexibility.

While you're here, could you please apply the same variadic configuration change to `NewServer()` for consistency.

catalinaperalta: Will add the variadic config to NewServer as well. 

As to the ResponseOption config, are you talking about providing functions that provide defaults and return an interface with a private method? I had wanted to have something that allows us to modify the Mock package server without providing a common default...like this implementation for http2 servers. So I could think of adding an interface with a public method that could then provide us with even more flexibility. Is that better? Or do you think this should be exactly the same as ResponseOption?

jhendrixMSFT: They wouldn't have to provide defaults, just access to the things they configure.  Something like this.
```go
func WithTLSConfig(cfg *tls.Config) ServerOption {
	return serverOption(func(s *Server) {
		s.srv.TLS = cfg
	})
}
```

catalinaperalta: Ah gotcha. Will finish implementing like that.

sdk/internal/mock/mock.go
catalinaperalta: Done

sdk/internal/mock/mock.go
catalinaperalta: Done

sdk/internal/mock/mock.go
catalinaperalta: Done in the tests

jhendrixMSFT (approved):

PR #12748, opened 2020-10-14
Azcore docs

jhendrixMSFT:
 
### DIFF
--- a/sdk/azcore/README.md
+++ b/sdk/azcore/README.md
@@ -1,0 +1,39 @@
+# Azure Core Client Library for Go
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/azcore)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore)
+[![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/go/go%20-%20azcore%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1843&branchName=master)
+[![Code Coverage](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)
+
+`azcore` provides a set of common interfaces and types for Go SDK client libraries.
+These libraries follow the [Azure SDK Design Guidelines for Go](https://azure.github.io/azure-sdk/golang_introduction.html).
+
+## Getting started
+
+This project uses [Go modules](https://github.com/golang/go/wiki/Modules) for versioning and dependency management.
+
+Typically, you will not need to explicitly install `azcore` as it will be installed as a client library dependency.
+To add the latest version to your `go.mod` file, execute the following command.
+
+```bash
+go get -u github.com/Azure/azure-sdk-for-go/sdk/azcore
+```
+
+General documentation and examples can be found on [pkg.go.dev](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore).
+
+## Contributing
+This project welcomes contributions and suggestions. Most contributions require
+you to agree to a Contributor License Agreement (CLA) declaring that you have
+the right to, and actually do, grant us the rights to use your contribution.
+For details, visit [https://cla.microsoft.com](https://cla.microsoft.com).
+
+When you submit a pull request, a CLA-bot will automatically determine whether
+you need to provide a CLA and decorate the PR appropriately (e.g., label,
+comment). Simply follow the instructions provided by the bot. You will only
+need to do this once across all repos using our CLA.
+
+This project has adopted the
+[Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
+For more information, see the
+[Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/)
+or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any
+additional questions or comments.
--- a/sdk/azcore/README.md
+++ b/sdk/azcore/README.md
@@ -1,10 +1,10 @@
-# Azure Core Client Library for Go
+# Azure Core Client Package for Go
 
 [![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/azcore)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore)
 [![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/go/go%20-%20azcore%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1843&branchName=master)
 [![Code Coverage](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)
 
-`azcore` provides a set of common interfaces and types for Go SDK client libraries.
+The `azcore` module provides a set of common interfaces and types for Go SDK client libraries.
 These libraries follow the [Azure SDK Design Guidelines for Go](https://azure.github.io/azure-sdk/golang_introduction.html).
 
 ## Getting started
--- a/sdk/azcore/README.md
+++ b/sdk/azcore/README.md
@@ -1,17 +1,17 @@
-# Azure Core Client Package for Go
+# Azure Core Client Module for Go
 
 [![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/azcore)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore)
 [![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/go/go%20-%20azcore%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1843&branchName=master)
 [![Code Coverage](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)
 
-The `azcore` module provides a set of common interfaces and types for Go SDK client libraries.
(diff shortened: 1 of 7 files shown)


sdk/azcore/README.md
```diff
@@ -0,0 +1,39 @@
+# Azure Core Client Library for Go
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/azcore)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore)
+[![Build Status](https://dev.azure.com/azure-sdk/public/_apis/build/status/go/go%20-%20azcore%20-%20ci?branchName=master)](https://dev.azure.com/azure-sdk/public/_build/latest?definitionId=1843&branchName=master)
+[![Code Coverage](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)](https://img.shields.io/azure-devops/coverage/azure-sdk/public/1843/master)
```
catalinaperalta, line 5: Just curious, how do you get this image?

sdk/azcore/core.go
```diff
@@ -14,7 +14,7 @@ import (
 // Request and react to the received Response.
 type Policy interface {
 	// Do applies the policy to the specified Request.  When implementing a Policy, mutate the
-	// request before calling req.Do() to move on to the next policy, and respond to the result
+	// request before calling req.Next() to move on to the next policy, and respond to the result
```
catalinaperalta, line 17: Needs period at the end.

sdk/azcore/README.md
jhendrixMSFT: I searched for "azure devops code coverage badge" and found https://shields.io/category/coverage

sdk/azcore/core.go
jhendrixMSFT: There's one more line to the sentence which ends with a period.

sdk/azcore/doc.go
```diff
@@ -5,160 +5,125 @@
...
-Package azcore implements an HTTP request/response middleware pipeline whose
-policy objects mutate an HTTP request's URL, query parameters, and/or headers before
-the request is sent over the wire.
+Package azcore implements an HTTP request/response middleware pipeline.
 
-Not all policy objects mutate an HTTP request; some policy objects simply impact the
-flow of requests/responses by performing operations such as logging, retry policies,
-timeouts, failure injection, and deserialization of response payloads.
+The middleware consists of three components.
+- One or more Policy objects.
```
jhendrixMSFT, line 11: Somewhat minor issue, I don't like calling these "objects" as it makes me think of OOP.  Open to suggestions.

sdk/azcore/README.md
catalinaperalta: I see, thanks for the info! :)

sdk/azcore/core.go
catalinaperalta: Ah yes...my eyes skipped straight over the next line.

sdk/azcore/doc.go
catalinaperalta: Instance(s)?

jhendrixMSFT: That could work.  Also thought of "implementation(s)".

sdk/azcore/example_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	body := strings.NewReader("this is seekable content to be uploaded")
+	err = req.SetBody(azcore.NopCloser(body), "text/plain")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func ExampleLogger_Should() {
+	const LogExpensiveThing azcore.LogClassification = "ExpensiveThing"
```
catalinaperalta, line 51: Not sure if we need more comments. Possibly some explaining this is how you create your own classification and then check if you should only log that?

sdk/azcore/doc.go
```diff
@@ -5,160 +5,125 @@
...
-   }
+   func (req *Request) SetBody(body ReadSeekCloser, contentType string) error
+
+A seekable stream is required so that upon retry, the retry Policy object can seek the stream
+back to the beginning before retrying the network request and re-uploading the body.
+
+Processing the Response
+
+When the HTTP response is received, the underlying *http.Response is wrapped in a Response type.
+The Response type contains various convenience methods, things like testing the HTTP response code
```
catalinaperalta, line 123: I'd remove "things".

sdk/azcore/example_test.go
jhendrixMSFT: Fixed

sdk/azcore/doc.go
jhendrixMSFT: Fixed

sdk/azcore/doc.go
jhendrixMSFT: Changed to instance, I like it.

sdk/azcore/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+`azcore` provides a set of common interfaces and types for Go SDK client libraries.
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

sdk/azcore/transport_default_http_client.go
```diff
@@ -31,9 +31,7 @@ func init() {
 	}
 }
 
-// DefaultHTTPClientTransport ...
+// DefaultHTTPClientTransport returns a shared Transport. It uses an http.Client.
```
catalinaperalta, line 34: Should this doc string be a little more like our other default func doc strings? Something like: "...returns a shared Transport initialized with default values in an http.Client."

catalinaperalta (approved):
LGTM, left one more comment

sdk/azcore/doc.go
```diff
@@ -5,160 +5,125 @@
...
-   func (p *myPolicy) Do(ctx context.Context, request pipeline.Request) (pipeline.Response, error) {
-      // TODO: Mutate/process the HTTP request here...
-      response, err := p.node.Do(ctx, request)	// Forward HTTP request to next Policy & get HTTP response
-      // TODO: Mutate/process the HTTP response here...
-      return response, err	// Return response/error to previous Policy
-   }
+Template for implementing a stateless Policy:
 
-Implementing the Factory Interface
+   func MyStatelessPolicy() Policy {
```
JeffreyRichter, line 35: Should this be NewMyStatelessPolicy()?

sdk/azcore/transport_default_http_client.go
jhendrixMSFT: You bring up a good point, and this makes me think that perhaps we should ditch the word "Default" altogether since it's different from `DefaultFooOptions()`.  Name it simply `HTTPClientTransport()` (no "New" prefix since this is shared).  What do you think?

sdk/azcore/doc.go
```diff
@@ -5,160 +5,125 @@
...
-   type myPolicyFactory struct {
-      // TODO: Add any configuration/setting fields if desired...
+         // return Response/error to previous Policy
+         return resp, err
+	   })
    }
 
-   func (f *myPolicyFactory) New(node pipeline.PolicyNode) Policy {
-      return &myPolicy{node: node} // TODO: Also initialize any configuration/setting fields here (if desired)...
+Template for implementing a stateful Policy:
```
JeffreyRichter, line 49: Do you want to add a ctor here like you have for the above?

sdk/azcore/doc.go
```diff
@@ -5,160 +5,125 @@
...
 
-Now that you have a pipeline.Pipeline object, you can create a pipeline.Request object (which is a simple
-wrapper around Go's standard http.Request object) and pass it to Pipeline's Do method along with passing a
-context.Context for cancelling the HTTP request (if desired).
+Implementing the Transport Interface
 
-   type Pipeline interface {
-      Do(ctx context.Context, methodFactory pipeline.Factory, request pipeline.Request) (pipeline.Response, error)
-   }
+The Transport interface is a specialized Policy.  Its responsibility is to send the HTTP request
```
JeffreyRichter, line 69: In C/C++ we have a TransportPolicy and its option is a Transport (which is not a policy). Our TransportPolicy calls the Transport and then reads the body (unless its a blob/file download). I like this approach and I think this is what NewPipline expect too. So, maybe the doc wording here just needs some cleanup with respect to terminology.

For example, you say "The Trasnport is always the last policy in the chain" but Transport is not a policy.

JeffreyRichter (approved):

sdk/azcore/transport_default_http_client.go
catalinaperalta: Hmmm...well I'm fine with default. But if you think this could cause customers confusion and they would want to change this then HTTPClientTransport would be ok. I do think Default describes this well. Maybe an addition needs to be made to the comment about when this is meant to be used? Or why this exists?

JeffreyRichter: If the customer doesn't specify a transport, then this IS the default transport, right?
I think this IS similar to our client options. In fact, we might want to add a TransportPolicyOptions struct and put the Transport as a field inside. Then we can get the DefaultTransportPolicyOptions and return the initialized structure with the field initialized to what this function returns. Then, we could actually delete this function.

This has 2 benefits: consistency with other policies and the ability to add fields to the transport policy without breaking in the future.

jhendrixMSFT: Correct, this is the default if you pass `nil` as the transport arg in `NewPipeline()`.

I feel like we should simply delete `DefaultHTTPClientTransport()`, so if you want the default behavior pass `nil`.  If you want a custom `Transport`, pass in something that implements that interface.

JeffreyRichter: OK

jhendrixMSFT: @catalinaperalta are you ok with this?

catalinaperalta: Yes that sounds good to me 👍

jhendrixMSFT: Fixed

sdk/azcore/doc.go
jhendrixMSFT: Fixed

sdk/azcore/doc.go
jhendrixMSFT: Fixed

sdk/azcore/doc.go
jhendrixMSFT: Fixed

PR #12832, opened 2020-10-16
Add Azure Arc managed identity environment

catalinaperalta:
This PR provides the necessary additions to support Azure Arc server managed identity environments.

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -148,7 +148,7 @@ func (c *managedIdentityClient) createAuthRequest(ctx context.Context, clientID 
 	case msiTypeAppServiceV20170901, msiTypeAppServiceV20190801:
 		return c.createAppServiceAuthRequest(ctx, clientID, scopes)
 	case msiTypeAzureArc:
-		// need to perform preliminary request to retreive the secret key provided by the IDENTITY_ENDPOINT
+		// need to perform preliminary request to retreive the secret key challenge provided by the HIMDS service
 		key, err := c.getAzureArcSecretKey(ctx, scopes)
 		if err != nil {
 			return nil, &AuthenticationFailedError{inner: err, msg: "Failed to retreive secret key from the identity endpoint."}
@@ -278,6 +278,8 @@ func (c *managedIdentityClient) getMSIType() (msiType, error) {
 			c.endpoint = endpointEnvVar
 			if header := os.Getenv(identityHeader); header != "" { // if BOTH the env vars IDENTITY_ENDPOINT and IDENTITY_HEADER are set the msiType is AppService
 				c.msiType = msiTypeAppServiceV20190801
+			} else if arcIMDS := os.Getenv(arcIMDSEndpoint); arcIMDS != "" {
+				c.msiType = msiTypeAzureArc
 			} else { // if ONLY the env var IDENTITY_ENDPOINT is set the msiType is Azure Functions
 				c.msiType = msiTypeUnavailable
 				return c.msiType, &CredentialUnavailableError{CredentialType: "Managed Identity Credential", Message: "This Managed Identity Environment is not supported yet"}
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -191,6 +203,56 @@ func (c *managedIdentityClient) createAppServiceAuthRequest(ctx context.Context,
...
+func (c *managedIdentityClient) getAzureArcSecretKey(ctx context.Context, scopes []string) (string, error) {
+	// create the request to retreive the secret key challenge provided by the HIMDS service
+	request, err := azcore.NewRequest(ctx, http.MethodGet, c.endpoint)
+	if err != nil {
+		return "", err
+	}
+	request.Header.Set(azcore.HeaderMetadata, "true")
+	q := request.URL.Query()
+	q.Add("api-version", azureArcAPIVersion)
+	q.Add("resource", strings.Join(scopes, " "))
```
jhendrixMSFT, line 215: Is this using `scopes` for something other than authentication scopes, or does Arc call this `resource`?

sdk/azidentity/managed_identity_client.go
```diff
@@ -191,6 +203,56 @@ func (c *managedIdentityClient) createAppServiceAuthRequest(ctx context.Context,
...
+	if err != nil {
+		return "", err
+	}
+	// the endpoint is expected to return a 401 with the WWW-Authenticte header set to the location
+	// of the secret key file. Any other status code indicates an error in the request.
+	if response.StatusCode != 401 {
+		return "", &AuthenticationFailedError{inner: newAADAuthenticationFailedError(response), msg: fmt.Sprintf("Expected a 401 Unauthorized response, received: %d", response.StatusCode)}
+	}
+	if header := response.Header.Get("WWW-Authenticate"); len(header) != 0 {
+		// the WWW-Authenticate header is expected in the following format: Basic realm=/some/file/path.key
```
jhendrixMSFT, line 228: I'd invert the above condition to keep all of this out of the `if` block body.

sdk/azidentity/managed_identity_client.go
```diff
@@ -191,6 +203,56 @@ func (c *managedIdentityClient) createAppServiceAuthRequest(ctx context.Context,
...
+		pos := strings.LastIndex(header, "=")
+		if pos == -1 {
+			return "", errors.New("Did not receive a value from WWW-Authenticate header")
```
jhendrixMSFT, line 231: Include the value of `header`.

sdk/azidentity/managed_identity_client.go
```diff
@@ -191,6 +203,56 @@ func (c *managedIdentityClient) createAppServiceAuthRequest(ctx context.Context,
...
+		}
+		key, err := ioutil.ReadFile(header[pos+1:])
+		if err != nil {
+			return "", fmt.Errorf("Could not read file contents: %w", err)
```
jhendrixMSFT, line 235: Include the file path?

sdk/azidentity/managed_identity_client.go
catalinaperalta: Arc calls it resource, also all the other Managed Identity credential scenarios call it resource and expect a resource. We have the code that trims the suffix in the AuthenticationPolicy code.

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go

sdk/azidentity/managed_identity_client.go
jhendrixMSFT: Yes I remember now.  We're using `TokenRequestOptions.Scopes` for something other than scopes for MSI.

At the very least, the parameter names should be changed from `scopes` to `resources` as they aren't scopes.

Second, if this is how we intend to implement and publicly support this, we should document `TokenRequestOptions.Scopes` that for MSI it isn't a slice of scopes but a slice of resources.

jhendrixMSFT: Which makes me wonder if we should rename `Scopes` to something more applicable, or just have a `Resources` field (I prefer the latter as it's explicit).

Also, does the suffix trimming in `AuthenticationPolicy()` work for all cases?  I recall that some scopes use "//.default".  The current algorithm will leave a trailing '/' in that case, is that ok/expected?  We had problems with confusion over this in track 1.

catalinaperalta: I can rename the scopes params in MSI to `resources` but I think I'll do it in a separate PR to do it for all of the methods that call it scopes. Also, I can add a comment explaining that for MSI this field will contain resources instead of scopes. I can also fix the case where we have `//.default`.

jhendrixMSFT: Be aware that some of them do have a trailing '/' while others do not.  See [environments](https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L107) in `go-autorest` for examples (I assume this is the same thing).  Are the scopes uniform in this way?  This seems brittle.

catalinaperalta: Hmmm that makes me think that removing only the `/.default` is on purpose. I can ask Scott about this, but it seems to me to account for the situation in which some still need to have a trailing `/`

jhendrixMSFT: It could be.  Confirmation would be great (also update the algorithm's comment explaining this).

catalinaperalta: Also, about my previous comment in AuthenticationPolicy we trim the suffix because we know we'll be most likely receiving a scope from the client lib.

catalinaperalta: Ok sounds good, how about I make all of those improvements to the scopes vs resources in a separate PR though? I can rename scopes to resources for this specific method in this one though.

jhendrixMSFT (approved):

PR #12882, opened 2020-10-20
Track2 updates

jhendrixMSFT:
Regenerate ARM packages with latest code generator.
Add back examples that were inadvertently deleted.
Updated examples based on codegen changes.


catalinaperalta:
Didn't we all decide to not include ARM Event hubs last week?

catalinaperalta:
Also, there was an azblob sample, we should probably keep it visible as well for future reference?

jhendrixMSFT:
Regarding event hubs, for shipping you are correct.  We won't actually be merging any of the content in this branch into master for shipping.

I forgot about storage, however I'm questioning the usefulness.  With the recent codegen changes for data-plane there are no longer any public operation group clients.  Given that the storage team already has samples for Go, I think we should focus on those and not ours.  What do you think?

catalinaperalta:
Ok yes that's fine by me if they will generate their own samples for Storage. I thought I heard Mayuri mention it today didn't want us to repeat work. Sounds good since they're already developing their own then.

catalinaperalta (approved):
LGTM

PR #12891, opened 2020-10-21
Adding manual-tests directory for managed identity e2e tests

catalinaperalta:
Adding manual_tests directory to store e2e tests for managed identity scenarios. 
Reorganize samples directory.


jhendrixMSFT:
Looks like we need CI for the `samples` directory.

sdk/samples/azidentity/manual-tests/webjobs/main.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: Why is this under a `webjobs` directory?  This seems like a pretty generic sample for retrieving a token via MSI.

catalinaperalta: I had originally thought the tests would branch off and become different at some point, but you're right. I'm changing it to managed_identity since that's all it does at the moment.

jhendrixMSFT (approved):

PR #12980, opened 2020-10-26
Adding CI scripts for samples

catalinaperalta:
Add scripts that only build and vet for samples.

### DIFF
--- a/eng/pipelines/templates/steps/build.yml
+++ b/eng/pipelines/templates/steps/build.yml
@@ -1,0 +1,32 @@
+parameters:
+  ServiceDirectory: ''
+  GoWorkspace: ''
+  Scope: 'sdk/...'
+  Image: ''
+  GoVersion: ''
+
+steps:
+
+  - pwsh: |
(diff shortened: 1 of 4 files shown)


catalinaperalta:
go vet fails because the examples arent named correctly. Looking into how I can rename them

eng/pipelines/templates/steps/build.yml
```diff
@@ -0,0 +1,32 @@
+parameters:
```
jhendrixMSFT, line 1: With the addition of this file I wonder if we can remove the build and vet steps from `build-test.yml` (probably rename it to simply `test.yml`), then apply both templates in `archetype-sdk-client.yml`.  This way we don't duplicate the content.

@scbedd for his input.

catalinaperalta: I did think about doing the same. I'll make the change if Scott thinks it's fine.

jhendrixMSFT (approved):

PR #13015, opened 2020-10-28
Addressing consistency review feedback

### LINKED ISSUE #13014: Address Client Certificate Credential Consistency Review feedback
Following changes need to be done, based on the review discussion:

- [x] Update Tenant ID Validation Message
- [ ] Add support for send certificate chain (SNI Auth) 
- [x]  Investigate credential behavior when scope is passed as nil (How does the credential handle it ? , what error does user see?). if its appropriately handled then no action is needed.
- [x] Investigate Credential Unavailable Exception scenarios (validate their use case and consistency across other credentials).
- [x] Rename clientCertificate param to certificate_path with the goal to add consistency with other languages.
- [x] Investigate possibility of flattenting/composing options bag.

catalinaperalta:
Updating tenantID validation error message and updating client certificate path parameter name. Fixes #13014

### DIFF
--- a/sdk/azidentity/authorization_code_credential.go
+++ b/sdk/azidentity/authorization_code_credential.go
@@ -42,7 +42,7 @@ func DefaultAuthorizationCodeCredentialOptions() AuthorizationCodeCredentialOpti
 // options: Manage the configuration of the requests sent to Azure Active Directory, they can also include a client secret for web app authentication.
 func NewAuthorizationCodeCredential(tenantID string, clientID string, authCode string, redirectURI string, options *AuthorizationCodeCredentialOptions) (*AuthorizationCodeCredential, error) {
 	if !validTenantID(tenantID) {
-		return nil, &CredentialUnavailableError{CredentialType: "Authorization Code Credential", Message: "Invalid tenantID provided. You can locate your tenantID by following the instructions listed here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names."}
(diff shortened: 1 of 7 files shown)


sdk/azidentity/authorization_code_credential.go
```diff
@@ -42,7 +42,7 @@ func DefaultAuthorizationCodeCredentialOptions() AuthorizationCodeCredentialOpti
 // options: Manage the configuration of the requests sent to Azure Active Directory, they can also include a client secret for web app authentication.
 func NewAuthorizationCodeCredential(tenantID string, clientID string, authCode string, redirectURI string, options *AuthorizationCodeCredentialOptions) (*AuthorizationCodeCredential, error) {
 	if !validTenantID(tenantID) {
-		return nil, &CredentialUnavailableError{CredentialType: "Authorization Code Credential", Message: "invalid tenant ID passed to credential"}
+		return nil, &CredentialUnavailableError{CredentialType: "Authorization Code Credential", Message: "Invalid tenantID provided. You can locate your tenantID by following the instructions listed here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names."}
```
jhendrixMSFT, line 45: Place message in a `const`.

jhendrixMSFT (approved):
Remove duplicate message, looks good.

sdk/azidentity/authorization_code_credential.go
catalinaperalta: Done

PR #13036, opened 2020-10-29
Flatten credential options

catalinaperalta:
Flatten all credential options into the main credential options structs. Addresses: #13014

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -53,12 +53,9 @@ type aadIdentityClient struct {
 // that are passed into it along with a default pipeline.
 // options: TokenCredentialOptions that can configure policies for the pipeline and the authority host that
 // will be used to retrieve tokens and authenticate
-func newAADIdentityClient(authorityHost string, options *pipelineOptions) (*aadIdentityClient, error) {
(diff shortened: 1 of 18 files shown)


sdk/azidentity/aad_identity_client.go
```diff
@@ -45,21 +45,20 @@ type interactiveConfig struct {
...
+	authorityHost string
+	pipeline      azcore.Pipeline
 }
 
 // newAADIdentityClient creates a new instance of the aadIdentityClient with the TokenCredentialOptions
 // that are passed into it along with a default pipeline.
 // options: TokenCredentialOptions that can configure policies for the pipeline and the authority host that
 // will be used to retrieve tokens and authenticate
-func newAADIdentityClient(options *TokenCredentialOptions) (*aadIdentityClient, error) {
+func newAADIdentityClient(authorityHost string, options *pipelineOptions) (*aadIdentityClient, error) {
```
jhendrixMSFT, line 56: Since this isn't public surface area, you could opt to made `options` by-value as we have complete control over the input.  Not sure how useful the change would be but it does appear to simplify the `nil` case below (i.e. that goes away).

jhendrixMSFT (approved):

sdk/azidentity/aad_identity_client.go

PR #13224, opened 2020-11-04
Adds armresources module with examples

jhendrixMSFT:
This adds a handful of examples, more are required.  However I wanted to get an early read on the pattern used in the examples before I copy/paste more that end up needing to be reworked.  E.g. should we have the helper `getDefaultClient()`, or is it more valuable to duplicate this code so each example is fully self-contained?  I'm leaning towards the latter as this is what consumers will see in `pkg.go.dev`.

I'm also wondering about the usage of a "default client" that is fed into operation group-specific clients.  I feel like this is a little confusing/clunky.


catalinaperalta:
I am more on the side of having each sample be self-contained because, as you mentioned, it will be shown more completely in pkg.go.dev. Also, it could make understanding and using each individual sample better for users, even if there is a lot of repetitive code.

sdk/arm/resources/2020-06-01/armresources/README.md
```diff
@@ -0,0 +1,48 @@
+# Azure Identity Client Module for Go
```
catalinaperalta, line 1: This is all mentioning azure identity.

sdk/arm/resources/2020-06-01/armresources/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+[Source code](https://github.com/Azure/azure-sdk-for-go/tree/master/sdk/arm/resources/2020-06-01/armresources)
+
+# Getting started
+
+## Install the package
+
+This project uses [Go modules](https://github.com/golang/go/wiki/Modules) for versioning and dependency management.
+
+Install the Azure Identity module:
```
catalinaperalta, line 15: Change Azure Identity

sdk/arm/resources/2020-06-01/armresources/README.md
jhendrixMSFT: Fixed.

sdk/arm/resources/2020-06-01/armresources/README.md
jhendrixMSFT: Fixed.

sdk/arm/resources/2020-06-01/armresources/README.md
```diff
@@ -0,0 +1,48 @@
+# Azure Resources Module for Go
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/arm/resources/2020-06-01/armresources)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/arm/resources/2020-06-01/armresources)
+
+The `armresources` module provides operations for working with resources and resource groups.
```
catalinaperalta, line 5: Should this say "for working with Azure resources and resource groups"?

jhendrixMSFT: We can.  For reference I copied this verbatim from the swagger.

sdk/arm/resources/2020-06-01/armresources/README.md
```diff
@@ -0,0 +1,48 @@
...
+Install the Azure Resources module:
+
+```sh
+go get -u github.com/Azure/azure-sdk-for-go/sdk/arm/resources/2020-06-01/armresources
+```
+
+## Prerequisites
+
+- an [Azure subscription](https://azure.microsoft.com/free/)
+- Go 1.13 or above
```
catalinaperalta, line 24: Should this be go1.14 and above now?

sdk/arm/resources/2020-06-01/armresources/examples_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/to"
+)
+
+func getDefaultClient() *armresources.Client {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		log.Fatalf("failed to obtain a credential: %v", err)
+	}
+	return armresources.NewDefaultClient(cred, nil)
+}
```
catalinaperalta, line 22: I'm thinking we should add this in each sample so users get the whole story about how to use the client and credentials.

jhendrixMSFT: Agreed, fixed.

sdk/arm/resources/2020-06-01/armresources/examples_test.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+}
+
+func ExampleResourceGroupsOperations_BeginDelete() {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		log.Fatalf("failed to obtain a credential: %v", err)
+	}
+	client := armresources.NewResourceGroupsClient(armresources.NewDefaultClient(cred, nil), "<subscription ID>")
```
jhendrixMSFT, line 66: I think this is a bit weird, i.e. we pass an `*armresources.Client` to another a `*armresources.ResourceGroupsClient`.  What if we renamed the former to `Connection`?  That's really what it represents as it consists of a credential, pipeline, and endpoint.

sdk/arm/resources/2020-06-01/armresources/README.md
catalinaperalta: I see. I think it's better with Azure in there, but if the swagger is like that then I dont feel strongly enough to push on it.

sdk/arm/resources/2020-06-01/armresources/examples_test.go
catalinaperalta: Yes, I've been staring at this because it feels off. I agree it feels clunky as is, I dont mind renaming the `*armresource.Client` to `Connection`

jhendrixMSFT: In fact, for ARM this makes me think that we could use a shared `Connection` type in `armcore` (needs some experimentation).

catalinaperalta: Oh I like that idea. It could come from ARM core. Depends on what implementation we go with, but I think it could work.

sdk/arm/resources/2020-06-01/armresources/README.md
jhendrixMSFT: I added it.

sdk/arm/resources/2020-06-01/armresources/README.md
jhendrixMSFT: I was thinking about that.  All the others are Go 1.13, and all of our build tags specify 1.13 as the minimum version.  I'm inclined to leave it as 1.13 for now and if we decide we want to bump the minimum to 1.14 then we do a repo-wide sweep.

catalinaperalta: Hmmm ok, we can discuss when to bump everything up to 1.14 later.

jhendrixMSFT: Also I believe that our code will work with 1.13, so while we don't have to support I don't feel that we have to go out of our way to prevent it from working.

sdk/arm/resources/ci.yml
```diff
@@ -0,0 +1,15 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
+trigger:
+  paths:
+    include:
+    - sdk/arm/resources/
+
+pr:
+  paths:
+    include:
+    - sdk/arm/resources/
```
catalinaperalta, line 10: Should this just be arm? Or will we be adding each module path here?

jhendrixMSFT: Each RP will have its own pipeline so that we don't have to build and verify the world when a subset changes (keep in mind there will be a lot of API versions per RP so it will add up).

catalinaperalta: Ok makes sense 👍

catalinaperalta (approved):
LGTM

nickzhums:
@jhendrixMSFT two points of feedback here
- besides samples we should have a high-level guide / quickstart of the SDK. (Including how to install, authenticate, use pattern). This can be done near the actual release date after we address all the things in the SDK
- For existing samples, I see currently the code samples located under each API version. Do we plan to update those as new versions are released every time? We also need to configure and structure those samples in such a way so that it can show up on **docs.microsoft.com**. An example is this repo: https://github.com/Azure-Samples/azure-samples-python-management

PR #13467, opened 2020-11-10
Fix profileBuilder

ArcturusZhang:
it turns out after applying https://github.com/Azure/autorest.go/pull/520, the profileBuilder might not work (check the results [here](https://github.com/Azure/azure-sdk-for-go/pull/13413)).
The tool will not automatically import the new uuid package, and therefore compile errors occur.

This PR fix the profileBuilder to let it scan the import statement and add them to the generated file when necessary.

### DIFF
--- a/tools/profileBuilder/model/aliasPackage.go
+++ b/tools/profileBuilder/model/aliasPackage.go
@@ -332,26 +332,19 @@ func (alias *AliasPackage) AddFunc(decl *ast.FuncDecl, originalFile *ast.File) e
 	alias.functions = append(alias.functions, decl)
 
 	for _, im := range imports {
-		if err := alias.addImport(im); err != nil {
-			return err
-		}
+		alias.addImport(im)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
What exactly is broken without this fix?  I looked at the linked PR, the only failure I see is the "broken link detection" which looks unrelated.

ArcturusZhang:
> What exactly is broken without this fix? I looked at the linked PR, the only failure I see is the "broken link detection" which looks unrelated.

I am not quite sure why this happens, but previously we rely on the programmatically invoked `goimports` to complete the import, such like the uuid package. But after we have switched the uuid to `github.com/gofrs/uuid`, it does not work anymore. Hence, it would not automatically import the `github.com/gofrs/uuid` package (in those are using uuid). I am not quite sure if this is related with the fact that the old uuid package always has an alias, ie, `uuid "github.com/satori/go.uuid"`. Do you have any clue of this?

PR #13495, opened 2020-11-11
Refactor various public surface area

jhendrixMSFT:
Made several option fields non-pointers as there is no semantic
difference between nil and their zero-value.
Added missing default options constructors.
Added doc comment to options structs explaining to use their
corresponding constructor funcs.
Unexported a few things that didn't need to be exported.
Renamed URI to URL in public surface area only.


jhendrixMSFT:
Note this did not remove the `redirectURL` parameter as I'm still waiting on an ack from the MSAL team.

sdk/azidentity/azidentity.go
```diff
@@ -134,9 +134,9 @@ type pipelineOptions struct {
 func setAuthorityHost(authorityHost string) (string, error) {
 	if authorityHost == "" {
 		authorityHost = AzurePublicCloud
-		if envAuthorityHost := os.Getenv("AZURE_AUTHORITY_HOST"); envAuthorityHost != "" {
-			authorityHost = envAuthorityHost
-		}
+	}
+	if envAuthorityHost := os.Getenv("AZURE_AUTHORITY_HOST"); envAuthorityHost != "" {
+		authorityHost = envAuthorityHost
```
catalinaperalta, line 139: Having this in the `authorityHost == ""` check was on purpose. This gives precedence to the secret passed in as a parameter in the ctor.

sdk/azidentity/azidentity_test.go
```diff
@@ -23,6 +23,13 @@ func Test_AuthorityHost_Parse(t *testing.T) {
 
 func Test_SetEnvAuthorityHost(t *testing.T) {
 	err := os.Setenv("AZURE_AUTHORITY_HOST", envHostString)
+	defer func() {
+		// Unset that host environment vairable to avoid other tests failed.
```
catalinaperalta, line 27: Nit: variable

catalinaperalta: It was my own misspelling but good to catch

sdk/azidentity/azidentity_test.go
```diff
@@ -23,6 +23,13 @@ func Test_AuthorityHost_Parse(t *testing.T) {
...
+		err = os.Unsetenv("AZURE_AUTHORITY_HOST")
+		if err != nil {
+			t.Fatalf("Unexpected error when unset environment vairable: %v", err)
```
catalinaperalta, line 30: Same nit here, I clearly had trouble with that word the day I wrote this 😂.

sdk/azidentity/azidentity_test.go
```diff
@@ -33,32 +40,27 @@ func Test_SetEnvAuthorityHost(t *testing.T) {
...
 	}
-	if authorityHost != customHostString {
-		t.Fatalf("Unexpected error when get host from environment vairable: %v", err)
-	}
-
-	// Unset that host environment vairable to avoid other tests failed.
-	err = os.Unsetenv("AZURE_AUTHORITY_HOST")
-	if err != nil {
-		t.Fatalf("Unexpected error when unset environment vairable: %v", err)
+	// env takes priority
```
catalinaperalta, line 61: Why?

sdk/azidentity/azidentity.go
jhendrixMSFT: If we leave the code like that then you can never override it via the env var.

sdk/azidentity/azidentity_test.go
jhendrixMSFT: It lets you change the value without having to crack the code.

sdk/azidentity/azidentity_test.go
jhendrixMSFT: Fixed

sdk/azidentity/azidentity_test.go
jhendrixMSFT: Fixed (I found a few more cases)

sdk/azidentity/azidentity.go
catalinaperalta: I left it like this because in many identity environments there are several environment variables that could possibly override the information that the user wants to employ for the credential. I am checking in with the identity crew to see if the other languages also have this, so far JS and Python both said yes that the params passed into the ctor take precendence over the environment vars. I'm asking Scott to get his thoughts on this. Will post back here once I get more information.

catalinaperalta: I'm open to possibly having this for `AUTHORITY_HOST` but other credential params might not be able to have the same order of precedence between ctor params and env vars.

jhendrixMSFT: On the one hand I can understand the params taking precedence in order to prevent them being inadvertently changed (if that was the original intent, a comment would have been helpful here; I saw the impl and assumed it was a bug).  Historically we've always had the env vars take precedence as a facility to let users change configuration on-the-fly.  I can see pros/cons for both approaches.  Adding @schaabs to weigh in.

catalinaperalta: Just heard from Scott offline, he said "values passed through code should always trump environment variables". I can add a comment above this stating that it was a conscious decision to value parameters passed in through the code than environment variables. 

Although, since this is a concept widely applied throughout `azidentity` I can go further and state this in the README to make sure users know what is happening.

catalinaperalta: We can also see if Scott has more to say on this thread, if you want to discuss further.

jhendrixMSFT: Agreed clear documentation is required as this is a deviation from how it works in track 1.

It also means that we can't set a default value for `AuthorityHost` in the `DefaultXXXOptions()` methods else the env var will always be ignored (more behavior that needs to be documented).

jhendrixMSFT: This has been resolved.  I added some comments to clarify the behavior.

catalinaperalta (approved):
LGTM

PR #13496, opened 2020-11-11
Add support for clouds other than public

jhendrixMSFT:
Renamed DefaultEndpoint to AzurePublicCloud.
Added const for the sovereign clouds.
Construct token scope based on the cloud.
Updated NewRPRegistrationPolicy() to take an endpoint param.

### DIFF
--- a/sdk/armcore/connection.go
+++ b/sdk/armcore/connection.go
@@ -67,7 +67,7 @@ func NewConnection(endpoint string, cred azcore.TokenCredential, options *Connec
 		azcore.NewTelemetryPolicy(&options.Telemetry),
 		NewRPRegistrationPolicy(endpoint, cred, &options.RegisterRPOptions),
 		azcore.NewRetryPolicy(&options.Retry),
-		cred.AuthenticationPolicy(azcore.AuthenticationPolicyOptions{Options: azcore.TokenRequestOptions{Scopes: []string{scope(endpoint)}}}),
+		cred.AuthenticationPolicy(azcore.AuthenticationPolicyOptions{Options: azcore.TokenRequestOptions{Scopes: []string{endpointToScope(endpoint)}}}),
 		azcore.NewLogPolicy(nil))
 	return NewConnectionWithPipeline(endpoint, p)
 }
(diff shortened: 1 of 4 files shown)


sdk/armcore/connection.go
```diff
@@ -78,3 +87,10 @@ func (c *Connection) Endpoint() string {
 func (c *Connection) Pipeline() azcore.Pipeline {
 	return c.p
 }
+
+func scope(endpoint string) string {
```
catalinaperalta, line 91: What happens if someone passes in an appropriate scope? With the `/.default` suffix? I think we need to check if the scope already ends with that so that we dont append it twice.

jhendrixMSFT: This converts the ARM endpoint to a scope.  I'll rename the func so it's clear.

catalinaperalta: One more question, what if we make the cloud endpoints some specific type like `CloudEndpoint`, or some other better name, and pass that in as the parameter to NewConnection(), so it looks like this: `NewConnection(endpoint CloudEndpoint, cred azcore.TokenCredential, options *ConnectionOptions)`?

I'm saying this because leaving it as a string could potentially leave it open for people to pass the endpoint in as a string and not use the consts, then that string could have the default suffix scope. I do think this will be a minority of cases, because hopefully most people would use the NewDefaultConnection() func. However, if someone is using NewConnection() for some reason I think we should have something, either the const type as the param or a comment indicating that they can pass in one of the exported consts for the endpoint. I would be fine with the doc comment instead of the param change. What do you think?

jhendrixMSFT: I thought about making it a `const`, the reason why I didn't is to support Azure stack which will have custom endpoints.  It would still work, it just requires casting the custom endpoint e.g. `CloudEndpoint("https://my.custom.endpoint.io")`.  Maybe the cast is ok (I'm starting to lean this way).  @JeffreyRichter what's you opinion on this?

If we change this to a typed `const` then we should do the same in `azidentity` for the authorization endpoints.

jhendrixMSFT: Per internal discussion we won't be changing this to a typed const.

catalinaperalta (approved):
LGTM

PR #13565, opened 2020-11-12
Remove RedirectURL parameter from auth code flow

jhendrixMSFT:
To align with the MSAL implementation which relies on the native client
redirect URL.

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -385,7 +385,7 @@ func (c *aadIdentityClient) authenticateInteractiveBrowser(ctx context.Context, 
 	if err != nil {
 		return nil, err
 	}
-	return c.authenticateAuthCode(ctx, tenantID, clientID, cfg.authCode, clientSecret, cfg.redirectURI, scopes)
+	return c.authenticateAuthCode(ctx, tenantID, clientID, cfg.authCode, clientSecret, scopes)
 }
 
 // authenticateAuthCode requests an Access Token with the authorization code and returns the token or an error in case of authentication failure.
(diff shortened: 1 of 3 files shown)


sdk/azidentity/aad_identity_client.go
```diff
@@ -415,14 +415,15 @@ func (c *aadIdentityClient) authenticateAuthCode(ctx context.Context, tenantID s
...
+func (c *aadIdentityClient) createAuthorizationCodeAuthRequest(ctx context.Context, tenantID string, clientID string, authCode string, clientSecret string, scopes []string) (*azcore.Request, error) {
 	data := url.Values{}
 	data.Set(qpGrantType, "authorization_code")
 	data.Set(qpClientID, clientID)
 	if clientSecret != "" {
 		data.Set(qpClientSecret, clientSecret) // only for web apps
 	}
-	data.Set(qpRedirectURI, redirectURI)
+	// similar to MSAL, we use a hard-coded redirect URI here as we never actually redirect to it
+	data.Set(qpRedirectURI, "https://login.microsoftonline.com/common/oauth2/nativeclient")
```
catalinaperalta, line 426: I'm just curious if you know about any app type that might need a custom redirect URI? Or if the MSAL team ever mentioned any? Not blocking on it, but just curious about this.

catalinaperalta (approved):
LGTM

sdk/azidentity/aad_identity_client.go
jhendrixMSFT: In our [conversation](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r517352378) the MSAL folks said that "MSAL stops the redirect" so there's no need to actually perform the redirect.

If we do end up needing this we can probably add it to the options struct given that it's optional from an MSAL perspective.

catalinaperalta: Ok that sounds like a good plan to me.

harikb:
This change is causing a problem for me.  The "redirectURI" parameter is needed in the flow of "Confidential client's Auth Code" flow as well.

If you look at your sample code https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/test/devapps/confidential_auth_code_sample.go 

redirectToURLConfidential in the sample needs to use an app specific redirect URL, because this is the one really sent to the browser. Later when 'code' is given to the site and we try to fetch an access token, that code no longer works because the above change hard-codes a particular  URL of 

`https://login.microsoftonline.com/common/oauth2/nativeclient`

@jhendrixMSFT  @catalinaperalta

PR #13605, opened 2020-11-13
Adding ARM compute module and samples

catalinaperalta:
Adding ARM compute module, docs and samples


sdk/arm/compute/2020-09-30/armcompute/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/azure-sdk-for-go/sdk/arm/compute/2020-09-30/armcompute
+
+go 1.13
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/armcore v0.5.1
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.13.4
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.6.0
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.5.0
```
jhendrixMSFT, line 9: What's using `internal`?  Can it be removed?

catalinaperalta: There are several fields that expect a pointer to type and I saw that the track1 samples similarly also referenced the track1 to package, so I decided to use it to have the variables in the struct where they belong. For instance look for:  `ExampleVirtualMachinesClient_BeginCreateOrUpdate()`

catalinaperalta: Although in that particular sample I can fix the vmName assignation since it doesnt need the `to` but the other fields would similarly have to be assigned earlier.

jhendrixMSFT: The `to` package is ok, but hopefully we don't need to reference `internal`.

catalinaperalta: Oh whoops...didnt pay attention to the package. internal is references because one of the samples needs a UUID.

catalinaperalta: Looking for it

catalinaperalta: Look at: ExampleVirtualMachineExtensionsClient_BeginCreateOrUpdate() 
Line 49

catalinaperalta: I can test the sample without the sequence version and see if it works without it

catalinaperalta: The operation fails without the `SequenceVersion` :/

jhendrixMSFT: I see.  Since this is a publicly consumable sample we can't reference the `internal` package as customers can't actually use it.  If we need to generate a UUID then we should use something like `github.com/gofrs/uuid`

jhendrixMSFT: Or, since this is an example and brevity is best maybe just hard-code the null UUID.

catalinaperalta: Ok I'll leave a string for users to replace 👍

catalinaperalta: Done

jhendrixMSFT (approved):

PR #13701, opened 2020-11-19
Bug fix: adding body back to response

catalinaperalta:
 
### DIFF
--- a/sdk/armcore/poller.go
+++ b/sdk/armcore/poller.go
@@ -458,7 +458,8 @@ func (pt *pollingTrackerBase) updateRawBody() error {
 		if seeker, ok := pt.resp.Body.(io.Seeker); ok {
 			_, err = seeker.Seek(0, io.SeekStart)
 			if err != nil {
-				return err
+				pt.Err = err
+				return pt.Err
 			}
 		} else {
 			// put the body back so it's available to other callers
--- a/sdk/armcore/poller.go
+++ b/sdk/armcore/poller.go
@@ -6,6 +6,7 @@
 package armcore
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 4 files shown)


sdk/armcore/poller.go
```diff
@@ -452,6 +454,16 @@ func (pt *pollingTrackerBase) updateRawBody() error {
 			pt.Err = err
 			return pt.Err
 		}
+		// seek back to the beginning of the body or reassign the information to the body
+		if seeker, ok := pt.resp.Body.(io.Seeker); ok {
+			_, err = seeker.Seek(0, io.SeekStart)
+			if err != nil {
+				return err
```
jhendrixMSFT, line 461: From the above code I believe `err` should be assigned to `pt.Err` before returning yes?

sdk/azcore/version.go
```diff
@@ -10,5 +10,5 @@ const (
 	UserAgent = "azcore/" + Version
 
 	// Version is the semantic version (see http://semver.org) of the pipeline package.
-	Version = "0.13.0"
+	Version = "0.13.3"
```
jhendrixMSFT, line 13: Since `v0.13.3` has already been tagged this won't get picked up.

I've been thinking about this problem in general.  At the very minimum, for PRs into `master` we should have a check to ensure that `version.go` is being updated and if it isn't to fail CI.

catalinaperalta: The check sounds like a good idea to me. I'll remove this change

jhendrixMSFT: In addition, we don't even inject `UserAgent` into the telemetry string :(  I will take a look at fixing this (it's missing for `armcore`, I haven't checked `azidentity`).

catalinaperalta: `azidentity` doesnt do it currently, if I recall correctly.

sdk/armcore/poller.go
catalinaperalta: Yes I think so. Done.

jhendrixMSFT (approved):

PR #13703, opened 2020-11-20
Adding Seek() method and updating payload()

catalinaperalta:
 
### DIFF
--- a/sdk/azcore/policy_body_download_test.go
+++ b/sdk/azcore/policy_body_download_test.go
@@ -303,32 +303,25 @@ func TestReadBodyAfterSeek(t *testing.T) {
 	if !ok {
 		t.Fatalf("unexpected body type: %t", resp.Body)
 	}
-	_, err = nb.Seek(0, io.SeekStart)
+	i, err := nb.Seek(0, io.SeekStart)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if nb.i != 0 {
+	if i != 0 {
 		t.Fatalf("did not seek correctly")
 	}
-	_, err = nb.Seek(5, io.SeekCurrent)
+	i, err = nb.Seek(5, io.SeekCurrent)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if nb.i != 5 {
+	if i != 5 {
 		t.Fatalf("did not seek correctly")
 	}
-	_, err = nb.Seek(5, io.SeekCurrent)
+	i, err = nb.Seek(5, io.SeekCurrent)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if nb.i != 10 {
+	if i != 10 {
 		t.Fatalf("did not seek correctly")
 	}
-	// payload, err = resp.payload()
-	// if err != nil {
-	// 	t.Fatalf("unexpected error: %v", err)
-	// }
-	// if string(payload) != message {
-	// 	t.Fatalf("unexpected response: %s", string(payload))
-	// }
 }
--- a/sdk/azcore/policy_body_download_test.go
+++ b/sdk/azcore/policy_body_download_test.go
@@ -7,6 +7,7 @@ package azcore
 
 import (
 	"context"
+	"io"
 	"net/http"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


sdk/azcore/response.go
```diff
@@ -26,15 +26,16 @@ type Response struct {
...
-		return nil
-	}
+func (r *Response) payload() ([]byte, error) {
 	// r.Body won't be a nopClosingBytesReader if downloading was skipped
 	if buf, ok := r.Body.(*nopClosingBytesReader); ok {
-		return buf.Bytes()
+		return buf.Bytes(), nil
 	}
-	return nil
+	bytesBody, err := ioutil.ReadAll(r.Body)
```
jhendrixMSFT, line 34: need to call `r.Body.Close()` after this line

sdk/azcore/response.go
```diff
@@ -26,15 +26,16 @@ type Response struct {
...
+	if err != nil {
+		return nil, err
+	}
+	return bytesBody, nil
```
jhendrixMSFT, line 38: I'm thinking that before we return, we should wrap the result in a `nopClosingBytesReader`.  Either way we should restore the `r.Body` so that customers can still read from it.

jhendrixMSFT: It also ties in with `removeBOM()`

sdk/azcore/response.go
```diff
@@ -120,12 +139,17 @@ func (r *Response) Drain() {
 }
 
 // removeBOM removes any byte-order mark prefix from the payload if present.
-func (r *Response) removeBOM() {
+func (r *Response) removeBOM() error {
```
jhendrixMSFT, line 142: I'd add a comment like `// NOTE: you MUST call payload() before calling this method` as the below type assertion could fail in some cases.

sdk/azcore/response.go
catalinaperalta: In your first comment, did you mean that we should change the return type for payload()?

sdk/azcore/response.go
catalinaperalta: Done. I didnt check the error for this, let me know if I should.

sdk/azcore/response.go
catalinaperalta: Done

catalinaperalta: Actually, looking at the code, is the comment true? We call payload before trimming anyways.

sdk/azcore/response.go
jhendrixMSFT: No need to check the error, calling `Close()` on an `io.ReadCloser` is one of the rare patterns where we don't need to check it.

sdk/azcore/response.go
jhendrixMSFT: Nope, you got it right.

sdk/azcore/response.go
jhendrixMSFT: Derp you are correct, comment is redundant.

sdk/azcore/response.go
```diff
@@ -84,12 +91,19 @@ func (r *Response) UnmarshalAsByteArray(v **[]byte, format Base64Encoding) error
 // UnmarshalAsJSON calls json.Unmarshal() to unmarshal the received payload into the value pointed to by v.
 // If no payload was received a RequestError is returned.  If json.Unmarshal fails a UnmarshalError is returned.
 func (r *Response) UnmarshalAsJSON(v interface{}) error {
```
jhendrixMSFT, line 93: Can you make sure we have a test that unmarshalls JSON (or XML, doesn't really matter) when the body download policy is disabled?  I suspect we don't test this today.

sdk/azcore/policy_body_download_test.go
```diff
@@ -234,10 +264,71 @@ func TestSkipBodyDownloadWith400(t *testing.T) {
...
+	}
+	nb, ok := resp.Body.(*nopClosingBytesReader)
+	if !ok {
+		t.Fatalf("unexpected body type: %t", resp.Body)
+	}
+	_, err = nb.Seek(0, io.SeekStart)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if nb.i != 0 {
```
jhendrixMSFT, line 310: Instead of checking `nb.i` here, you should check against the returned offset from the call to `Seek()` above.  Comment applies to all invocations of `Seek()`.

catalinaperalta: Done

sdk/azcore/response.go
catalinaperalta: Done

sdk/azcore/response.go
catalinaperalta: Done, removed the comment.

jhendrixMSFT (approved):

PR #13842, opened 2020-12-03
fix profile 2020-09-01 of compute

ArcturusZhang:
I put it into the wrong directory initially - we might need to release a v49.0.1 to release this change


ArcturusZhang:
It turns out that this might be a breaking change for the profiles and hence a breaking change for the SDK.
But this mistake in path would not affect the functionality of this profile, we might could postpone this to the end of this month releasing along with v50.0.0? 
What do you think?@jhendrixMSFT

jhendrixMSFT:
Agreed, technically it's a breaking change so wait for v50.

PR #13910, opened 2020-12-08
Generator: generate the changelog from a backup instead of playing with git

ArcturusZhang:
1. Now generate the changelog from a backup instead of playing with git
1. Make generated changelog more compact
1. Changelog now has a summary in this format: `Total ## breaking change(s), ## additive change(s).`


generate_options.json
```diff
@@ -8,8 +8,5 @@
     "--use-onever",
     "--preview-chk",
     "--version=V2"
-  ],
-  "afterScripts": [
-    "gofmt -w ./services/"
```
jhendrixMSFT, line 13: The formatting happens elsewhere yes?

jhendrixMSFT:
Few questions about this.

1. Why generate from a copy (backup)?
2. Do you have a sample of the compact changelog format?  What changed in it.

generate_options.json
ArcturusZhang: yeah, I just embedded this into the script - before we generate the changelog. And I remove this since we do not actually have a place that is using this right now

ArcturusZhang:
> Few questions about this.
> 
> 1. Why generate from a copy (backup)?
> 2. Do you have a sample of the compact changelog format?  What changed in it.

You could find the samples in this changelogs in v49.0.0 - which might need some tweaks
https://github.com/ArcturusZhang/azure-sdk-for-go/blob/d6435f714d5aa2c1d0b4b98c9538de7ec82c073b/services/eventgrid/2018-01-01/eventgrid/CHANGELOG.md
and 
https://github.com/ArcturusZhang/azure-sdk-for-go/blob/d6435f714d5aa2c1d0b4b98c9538de7ec82c073b/services/adhybridhealthservice/mgmt/2014-01-01/adhybridhealthservice/CHANGELOG.md

jhendrixMSFT:
Looking at the CHANGELOG linked for `adhybridhealthservice`, this looks suspicious.

Function NewAddsConfigurationPage signature has been changed from (func(context.Context, AddsConfiguration) (AddsConfiguration, error)) to (AddsConfiguration,func(context.Context, AddsConfiguration) (AddsConfiguration, error))

In the "changed to", note the extra "AddsConfiguration,", this looks broken to me.

ArcturusZhang:
> Looking at the CHANGELOG linked for `adhybridhealthservice`, this looks suspicious.
> 
> Function NewAddsConfigurationPage signature has been changed from (func(context.Context, AddsConfiguration) (AddsConfiguration, error)) to (AddsConfiguration,func(context.Context, AddsConfiguration) (AddsConfiguration, error))
> 
> In the "changed to", note the extra "AddsConfiguration,", this looks broken to me.

Well, originally, it looks like this:
```
### Siganature change
- `NewAddsConfigurationPage`
    - From
         `func(context.Context, AddsConfiguration) (AddsConfiguration, error)`
    - To
         `(AddsConfiguration,func(context.Context, AddsConfiguration) (AddsConfiguration, error))`
```
The context is that this function is the constructor of the pager struct, which has a parameter of type of a function. And we add a new parameter before that function-typed parameter.

jhendrixMSFT:
Oh yes you're correct.  Maybe we need to improve the format here a bit to make it more readable.  I like the `### Signature change` table in your comment, can we do that?

ArcturusZhang:
> Oh yes you're correct. Maybe we need to improve the format here a bit to make it more readable. I like the `### Signature change` table in your comment, can we do that?

Funny story, the table form is just the original format. Some context for this:
1. In this PR, I am changing the generator tool, which would generate the code as well as changelog in the SDK automation pipeline.
1. The generated changelog would be displayed in the comment of the sdk automation pipeline in a swagger PR - for the reviewer to intuitively have a knowledge of how this swagger would impact the downstream SDKs.
1. The changelog shown in the comments only have a very limited space (some certain amount of lines)

Therefore I have to do this change on the changelog to make it more compact (one change one line, like this signature change).
We could revert the format of changelog for the changelog in those SDK packages, but those changelogs are not generated by this tool (it is generated by an external tool that is not quite ready to put in this repo)
Therefore how about this plan, we could keep the original style of changelog for the changelogs in each package, but for the changelog generated and shown in the pipeline, we use a compact version (also suggestions are welcomed for this compact version)

jhendrixMSFT:
I see, thanks for the explanation.

Would it make sense to update the PR comment with a link to the generated changelog?  While I'm not against adding the changelog info to the PR itself, it seems like it could still overflow the limit.  What if the PR text contained a summary, e.g. `nn number of breaking changes`, `mm number of additions` etc with a link to the complete changelog?

ArcturusZhang:
> I see, thanks for the explanation.
> 
> Would it make sense to update the PR comment with a link to the generated changelog? While I'm not against adding the changelog info to the PR itself, it seems like it could still overflow the limit. What if the PR text contained a summary, e.g. `nn number of breaking changes`, `mm number of additions` etc with a link to the complete changelog?

The link for changelog is provided by the SDK automation pipeline. But the summary is an awesome idea, I will be working on that today. Once it overflows, the PR comment only shows the last a few lines, therefore the summary will always be there.

ArcturusZhang:
Hi @jhendrixMSFT I add one line of the summary at the end of the change log, could you please have a look?

jhendrixMSFT (approved):
I only reviewed the changes to `apidiff`.

PR #13956, opened 2020-12-11
Add logging options to all option types

jhendrixMSFT:
 
### DIFF
--- a/sdk/azidentity/authorization_code_credential.go
+++ b/sdk/azidentity/authorization_code_credential.go
@@ -33,6 +33,7 @@ func DefaultAuthorizationCodeCredentialOptions() AuthorizationCodeCredentialOpti
 	return AuthorizationCodeCredentialOptions{
 		Retry:     azcore.DefaultRetryOptions(),
 		Telemetry: azcore.DefaultTelemetryOptions(),
+		Logging:   azcore.DefaultLogOptions(),
 	}
 }
 
--- a/sdk/azidentity/authorization_code_credential.go
+++ b/sdk/azidentity/authorization_code_credential.go
@@ -24,6 +24,8 @@ type AuthorizationCodeCredentialOptions struct {
 	Retry azcore.RetryOptions
 	// Telemetry configures the built-in telemetry policy behavior
 	Telemetry azcore.TelemetryOptions
+	// Logging configures the built-in logging policy behavior.
+	Logging azcore.LogOptions
 }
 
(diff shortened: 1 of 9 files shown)


sdk/azidentity/client_certificate_credential.go
```diff
@@ -96,7 +98,7 @@ func NewClientCertificateCredential(tenantID string, clientID string, certificat
 	if err != nil {
 		return nil, err
 	}
-	c, err := newAADIdentityClient(authorityHost, pipelineOptions{HTTPClient: options.HTTPClient, Retry: options.Retry, Telemetry: options.Telemetry})
```
catalinaperalta, line 99: The defaults should be adding after line 41 in DefaultClientCertificateCredentialOptions()

sdk/azidentity/authorization_code_credential.go
```diff
@@ -64,7 +66,7 @@ func NewAuthorizationCodeCredential(tenantID string, clientID string, authCode s
 	if err != nil {
 		return nil, err
 	}
-	c, err := newAADIdentityClient(authorityHost, pipelineOptions{HTTPClient: options.HTTPClient, Retry: options.Retry, Telemetry: options.Telemetry})
+	c, err := newAADIdentityClient(authorityHost, pipelineOptions{HTTPClient: options.HTTPClient, Retry: options.Retry, Telemetry: options.Telemetry, Logging: options.Logging})
```
catalinaperalta, line 69: Same thing here. There defaults should be added after line 35 in the default options func.

sdk/azidentity/client_certificate_credential.go
jhendrixMSFT: Fixed

sdk/azidentity/authorization_code_credential.go
jhendrixMSFT: Fixed

jhendrixMSFT:
I noticed that there are no config options for `DefaultAzureCredential` which seems broken.  I can add them in this PR, let me know your thoughts.

catalinaperalta:
> I noticed that there are no config options for `DefaultAzureCredential` which seems broken. I can add them in this PR, let me know your thoughts.

That is by design, with `DefaultAzureCredential` we are literally saying you get all the defaults. I also believe this is the cross-language design for `DefaultAzureCredential`. We can bring in Scott if you want more info, but I do recall that was at least part of the reasoning, this question has come up before.

jhendrixMSFT:
Hmm I see.  We can wait for customer feedback to see how this works in practice.

catalinaperalta (approved):
LGTM

PR #13995, opened 2020-12-14
azidentity: Adding SubjectName/Issuer support

### LINKED ISSUE #11874: Support SubjectName/Issuer (SNI) certificate authentication for Service Principals
Support SubjectName/Issuer (SNI) certificate authentication for Service Principals

catalinaperalta:
Fixes #11874

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -120,8 +120,8 @@ func (c *aadIdentityClient) authenticate(ctx context.Context, tenantID string, c
 // clientID: The client (application) ID of the service principal
 // clientCertificatePath: The path to the client certificate PEM file
 // scopes: The scopes required for the token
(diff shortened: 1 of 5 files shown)


jhendrixMSFT (approved):
It looks good, do you have a way to test that the header is correct?

catalinaperalta:
> It looks good, do you have a way to test that the header is correct?

The short and I would say is not a perfect way to test it. But what I was thinking about is maybe comparing the number of certificates in the slice and possibly encoding the certificate in the test and comparing it to the value in the header? That could be a way to test it. Also, just an FYI, I did manually perform a live test against AAD with this feature set and successfully authenticated since I received an access token in response.

jhendrixMSFT:
Good idea on the test, and glad that the manual test worked.

catalinaperalta:
Added the test! Could only check up to the JWT's header, since both the payload and signature have random or unique characters generated in them.

jhendrixMSFT (approved):

PR #13996, opened 2020-12-14
azidentity: Adding doc.go

### LINKED ISSUE #12930: Add a doc.go file to azidentity
Add a doc.go file to azidentity.  It should contain explanations and examples for how to use the module.

catalinaperalta:
Fixes #12930

### DIFF
--- a/sdk/azidentity/doc.go
+++ b/sdk/azidentity/doc.go
@@ -22,7 +22,7 @@ The following credential types are included in this module:
 	- UsernamePasswordCredential
 
 By default, the recommendation is that users call NewDefaultAzureCredential() which will
-provide a default ChainedTokenCredential configuration conformed of:
+provide a default ChainedTokenCredential configuration composed of:
 	- EnvironmentCredential
 	- ManagedIdentityCredential
 	- AzureCLICredential
@@ -34,16 +34,15 @@ Example call to NewDefaultAzureCredential():
 	if err != nil {
 		// process error
 	}
-	client := armresources.NewResourcesClient(armcore.NewDefaultConnection(cred, nil), "<subscription ID>")
+	// pass credential in to an Azure SDK client
 
 Example call to NewDefaultAzureCredential() with options set:
 	// these options will make sure the AzureCLICredential will not be added to the credential chain
-	opts := DefaultAzureCredentialOptions{ExcludeAzureCLICredential: true}
-	cred, err := NewDefaultAzureCredential(&opts)
+	cred, err := NewDefaultAzureCredential(&DefaultAzureCredentialOptions{ExcludeAzureCLICredential: true})
 	if err != nil {
 		// process error
 	}
-	client := armresources.NewResourcesClient(armcore.NewDefaultConnection(cred, nil), "<subscription ID>")
+	// pass credential in to an Azure SDK client
 
 Additional configuration of each credential can be done through each credential's
 options type. These options can also be used to modify the default pipeline for each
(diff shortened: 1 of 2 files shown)


sdk/azidentity/doc.go
```diff
@@ -0,0 +1,89 @@
...
+	- ClientSecretCredential
+	- DefaultAzureCredential
+	- DeviceCodeCredential
+	- EnvironmentCredential
+	- InteractiveBrowserCredential
+	- ManagedIdentityCredential
+	- UsernamePasswordCredential
+
+By default, the recommendation is that users call NewDefaultAzureCredential() which will
+provide a default ChainedTokenCredential configuration conformed of:
```
jhendrixMSFT, line 25: "conformed" => "composed"

sdk/azidentity/doc.go
```diff
@@ -0,0 +1,89 @@
...
+Example call to NewDefaultAzureCredential():
+	cred, err := NewDefaultAzureCredential(nil) // pass in nil to get default behavior for this credential
+	if err != nil {
+		// process error
+	}
+	client := armresources.NewResourcesClient(armcore.NewDefaultConnection(cred, nil), "<subscription ID>")
+
+Example call to NewDefaultAzureCredential() with options set:
+	// these options will make sure the AzureCLICredential will not be added to the credential chain
+	opts := DefaultAzureCredentialOptions{ExcludeAzureCLICredential: true}
```
jhendrixMSFT, line 41: Nit: typically this would be expressed as a struct literal at the call site.

sdk/azidentity/doc.go
```diff
@@ -0,0 +1,89 @@
...
+	cred, err := azidentity.NewClientCertificateCredential("<tenant ID>", "<client ID>", "<certificate path>", &opts)
+	if err != nil {
+		// process error
+	}
+
+CREDENTIAL AUTHORITY HOSTS
+
+The default authority host for all credentials, except for ManagedIdentityCredential, is the
+AzurePublicCloud host. This value can be changed through the credential options or by specifying
+a different value in an environment variable called AZURE_AUTHORITY_HOST.
```
jhendrixMSFT, line 64: As I recall, there's an ordering issue here where if the value is explicitly set in the code the env var will never be honored.  Would be good to explain that here.

sdk/azidentity/doc.go
```diff
@@ -0,0 +1,89 @@
...
+	- AzureCLICredential
+Configuration options can be used to exclude any of the previous credentials from the
+DefaultAzureCredential implementation.
+
+Example call to NewDefaultAzureCredential():
+	cred, err := NewDefaultAzureCredential(nil) // pass in nil to get default behavior for this credential
+	if err != nil {
+		// process error
+	}
+	client := armresources.NewResourcesClient(armcore.NewDefaultConnection(cred, nil), "<subscription ID>")
```
jhendrixMSFT, line 37: I wonder if it would be better to simply say something like `// use cred ....` instead of something this specific.  I don't feel super-strong about it, so feel free to leave it if you think it's better as-is.

jhendrixMSFT (approved):
A few minor things.

sdk/azidentity/doc.go
catalinaperalta: Done

sdk/azidentity/doc.go
catalinaperalta: Done

sdk/azidentity/doc.go
catalinaperalta: Done

sdk/azidentity/doc.go
catalinaperalta: Done

catalinaperalta: I didnt want to leave it as just "Use the credential" because then people might think they have to call GetToken themselves, whereas we do that for them in the SDK clients. So I added a message that expresses that, for the common case.

jhendrixMSFT (approved):

PR #14013, opened 2020-12-16
Adding ADFS support

### LINKED ISSUE #11875: Add ADFS / Azure Stack Support
- On-prem active directory instance support
- Requested by CLI / PSH for .NET and Python

catalinaperalta:
Fixes #11875

### DIFF
--- a/sdk/azidentity/aad_identity_client.go
+++ b/sdk/azidentity/aad_identity_client.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 
@@ -17,8 +18,6 @@ import (
 
 const (
 	clientAssertionType = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
-	tokenEndpoint       = "/oauth2/v2.0/token"
-	authEndpoint        = "/oauth2/v2.0/authorize"
 )
 
 const (
@@ -194,7 +193,7 @@ func (c *aadIdentityClient) createRefreshTokenRequest(ctx context.Context, tenan
 	data.Set(qpScope, strings.Join(scopes, " "))
 	dataEncoded := data.Encode()
 	body := azcore.NopCloser(strings.NewReader(dataEncoded))
-	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(c.authorityHost, tenantID, tokenEndpoint))
+	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(c.authorityHost, tenantID, tokenEndpoint(oauthPath(tenantID))))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


sdk/azidentity/azidentity.go
```diff
@@ -209,3 +209,10 @@ func validTenantID(tenantID string) bool {
 	}
 	return match
 }
+
+func getIdentityEndpointSuffix(tenantID string) string {
+	if tenantID == "adfs" {
```
jhendrixMSFT, line 214: It's really this simple?

sdk/azidentity/azidentity.go
```diff
@@ -209,3 +209,10 @@ func validTenantID(tenantID string) bool {
...
+		return "/oauth2/token"
```
jhendrixMSFT, line 215: This should probably be a `const` next to `tokenEndpoint`.

sdk/azidentity/aad_identity_client.go
```diff
@@ -192,7 +192,7 @@ func (c *aadIdentityClient) createRefreshTokenRequest(ctx context.Context, tenan
 	data.Set(qpScope, strings.Join(scopes, " "))
 	dataEncoded := data.Encode()
 	body := azcore.NopCloser(strings.NewReader(dataEncoded))
-	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(c.authorityHost, tenantID, tokenEndpoint))
+	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(c.authorityHost, tenantID, getIdentityEndpointSuffix(tenantID)))
```
jhendrixMSFT, line 195: Do we need to provide a way for callers to specify a complete URL here (stack scenario)?

sdk/azidentity/azidentity.go
catalinaperalta: Yes, seems so. The only thing left to do is validate all of the credentials against an in-house ADFS instance. I've talked to someone on the identity crew, so I'll do that before moving forward with this PR.

sdk/azidentity/aad_identity_client.go
catalinaperalta: Not as far as I know yet. They will change the authority host, pass in the "adfs" tenant ID and we will automatically add the correct ADFS endpoint for them. Will get back to you after testing this against an ADFS instance in case something needs to change, but it seems this is how it works.

catalinaperalta:
Resolved old conversations as the code has changes a bit.

sdk/azidentity/aad_identity_client.go
catalinaperalta: I dont think there is a case were a fully custom URL needs to be passed in.

catalinaperalta:
This has been tested with an ADFS instance. Testing is only pending for Azure CLI credential: #14405

sdk/azidentity/azidentity.go
jhendrixMSFT: Do we need to add documentation for this behavior?  Seems hidden at present.

jhendrixMSFT: Do we need to add documentation for this behavior?  Seems hidden.

jhendrixMSFT (approved):

PR #14031, opened 2020-12-17
Exclude tests when parsing packages

jhendrixMSFT:
 
### DIFF
--- a/tools/internal/modinfo/modinfo.go
+++ b/tools/internal/modinfo/modinfo.go
@@ -127,7 +127,7 @@ func GetModuleInfo(baseline, staged string) (Provider, error) {
 	lhs, err := exports.Get(baseline)
 	if err != nil {
 		// if baseline has no content then this is a v1 package
-		if ei, ok := err.(exports.LoadPackageErrorInfo); !ok || ei.PkgCount() != 0 {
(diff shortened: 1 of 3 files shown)


tools/internal/modinfo/modinfo.go
```diff
@@ -127,7 +127,7 @@ func GetModuleInfo(baseline, staged string) (Provider, error) {
 	lhs, err := exports.Get(baseline)
 	if err != nil {
 		// if baseline has no content then this is a v1 package
-		if ei, ok := err.(exports.LoadPackageErrorInfo); !ok || ei.PkgCount() != 0 {
+		if ei, ok := err.(exports.LoadPackageErrorInfo); !ok || len(ei.Packages()) != 0 {
```
jhendrixMSFT, line 130: @ArcturusZhang is this comparison a bug?  Shouldn't it be `len(ei.Packages()) == 0`?

catalinaperalta (approved):

tools/internal/modinfo/modinfo.go
ArcturusZhang: This should not be a bug. 
The `exports.Get` function returns errors when there are no packages in the `baseline`. And here we are trying to identity if the baseline has no package, if so, we escape this error. But we will not escape the case that multiple packages were found in baseline.

PR #14052, opened 2020-12-23
Adding port option for interactive browser

catalinaperalta:
Users may need to set a specific port for their localhost server. Enable that option and also check that the authority host for interactive browser is in the right format for the string format it is going to be added into.

Fixes #14052.

### DIFF
--- a/sdk/azidentity/interactive_browser_credential.go
+++ b/sdk/azidentity/interactive_browser_credential.go
@@ -147,8 +147,8 @@ func interactiveBrowserLogin(authorityHost string, tenantID string, clientID str
 	}, nil
 }
 
+// if the authority host ends with a "/"" remove it and return
 func checkAuthHostFormat(s string) string {
-	// if the authority host ends with a / remove it and return
 	if string(s[len(s)-1]) == "/" {
 		return s[:len(s)-1]
 	}
--- a/sdk/azidentity/interactive_browser_credential.go
+++ b/sdk/azidentity/interactive_browser_credential.go
@@ -153,11 +153,3 @@ func interactiveBrowserLogin(authorityHost string, tenantID string, clientID str
 		redirectURI: redirectURL,
 	}, nil
 }
-
-// if the authority host ends with a "/"" remove it and return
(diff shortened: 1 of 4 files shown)


sdk/azidentity/interactive_browser_credential.go
```diff
@@ -144,3 +146,11 @@ func interactiveBrowserLogin(authorityHost string, tenantID string, clientID str
 		redirectURI: redirectURL,
 	}, nil
 }
+
+// if the authority host ends with a "/"" remove it and return
+func checkAuthHostFormat(s string) string {
```
jhendrixMSFT, line 151: There is `azcore.JoinPaths()` which will handle URL path concatenation and trim any extra `/` as required.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -26,6 +26,8 @@ type InteractiveBrowserCredentialOptions struct {
 	ClientSecret string
 	// The redirect URL used to request the authorization code. Must be the same URL that is configured for the App Registration.
 	RedirectURL string
+	// The localhost port for the local server that will be used to redirect back to
```
jhendrixMSFT, line 29: Based on the implementation, you should document that leaving this a zero will default to a random port number.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Fixed. I parsed the authority host string and then joined the paths.

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done.

sdk/azidentity/interactive_browser_credential.go
jhendrixMSFT: Can this func be removed now?

catalinaperalta: Yes, it's done now.

jhendrixMSFT (approved):

PR #14053, opened 2020-12-23
APIDIFF: Fixing bugs in content generation

catalinaperalta:
This PR adds necessary updates so that the content can be generated correctly for libraries, such as azcore, azidentity, armcore, etc. It has been tested with the previously mentioned modules.

### DIFF
--- a/tools/apidiff/exports/exports.go
+++ b/tools/apidiff/exports/exports.go
@@ -123,7 +123,7 @@ func (c *Content) addConst(pkg Package, g *ast.GenDecl) {
 				co.Type = "*ast.BinaryExpr"
 				v = pkg.getText(ce.X.Pos(), ce.Y.End())
 			} else {
-				panic("unhandled case for adding constant")
+				panic(fmt.Sprintf("unhandled case for adding constant: %s", pkg.getText(vs.Pos(), vs.End())))
 			}
 		}
 		// TODO should this also be removed?
--- a/tools/apidiff/exports/exports.go
+++ b/tools/apidiff/exports/exports.go
@@ -99,8 +99,16 @@ func (c *Content) addConst(pkg Package, g *ast.GenDecl) {
 		v := ""
 		// Type is nil for untyped consts
 		if vs.Type != nil {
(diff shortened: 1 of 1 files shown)


tools/apidiff/exports/exports.go
```diff
@@ -110,10 +118,15 @@ func (c *Content) addConst(pkg Package, g *ast.GenDecl) {
 				// const FooConst = FooType("value")
 				co.Type = pkg.getText(ce.Fun.Pos(), ce.Fun.End())
 				v = pkg.getText(ce.Args[0].Pos(), ce.Args[0].End())
+			} else if ce, ok := vs.Values[0].(*ast.BinaryExpr); ok {
+				// const FooConst = "value" + Bar
+				co.Type = "*ast.BinaryExpr"
```
jhendrixMSFT, line 123: Why do we need this?

tools/apidiff/exports/exports.go
```diff
@@ -110,10 +118,15 @@ func (c *Content) addConst(pkg Package, g *ast.GenDecl) {
...
+				v = pkg.getText(ce.X.Pos(), ce.Y.End())
 			} else {
-				panic(fmt.Sprintf("unhandled case for adding constant: %s", pkg.getText(vs.Pos(), vs.End())))
+				panic("unhandled case for adding constant")
```
jhendrixMSFT, line 126: Can we keep the text for the line in the panic message?

tools/apidiff/exports/exports.go
catalinaperalta: If I recall correctly, the Type field can't be empty so I put this arbitrary value in it. I could put the content of the const in it instead if you think that's better?

jhendrixMSFT: Oh yes I remember now.  It's used during comparison to see if the `const` type has changed.  I think what you have is fine.

tools/apidiff/exports/exports.go
catalinaperalta: Done

jhendrixMSFT (approved):

PR #14061, opened 2020-12-24
Refine the apidiff tool - make reorder of the function parameters a breaking change

ArcturusZhang:
Now the function parameter list will also include the parameter name, and the change of parameter name will also be considered as a signature change - aka a breaking change.
For example, we might have a function `CreateOrUpdate(ctx context.Context, resourceGroup string, name string, body Body)`, if we change it to `CreateOrUpdate(ctx context.Context, name string, resourceGroup string, body Body)`, it should be considered as a breaking change.

Also I would like to modify the codegen for track 1 to make it sort the parameters in the functions. Might using the following pattern:
Path parameter goes first, and then body and then query.
In path parameters, we also sort them in the order of they appear in the path (i.e. resourceGroupName appears first in the path, then it is in the first place of the parameter list).

### DIFF
--- a/tools/apidiff/delta/delta.go
+++ b/tools/apidiff/delta/delta.go
@@ -224,16 +224,20 @@ func GetFuncSigChanges(lhs, rhs exports.Content) map[string]FuncSig {
 			continue
 		}
 		sig := FuncSig{}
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
Per offline conversation, including the param name will incorrectly flag param renames as a breaking change.

PR #14109, opened 2021-01-08
Switching changelogs to track2 format

catalinaperalta:
This PR switches the content of the changelogs for the ARM modules to use the output from the apidiff tool.

### DIFF
--- a/sdk/arm/compute/2020-09-30/armcompute/CHANGELOG.md
+++ b/sdk/arm/compute/2020-09-30/armcompute/CHANGELOG.md
@@ -1,3 +1,5 @@
-Generated from https://github.com/Azure/azure-rest-api-specs/tree/c72af9b8dca625d3f1429a0bd913bad7b3879645
+# Release History
 
-Code generator @autorest/go@4.0.0-preview.9
+## v0.1.0 (released)
+
+
--- a/sdk/arm/keyvault/2019-09-01/armkeyvault/CHANGELOG.md
+++ b/sdk/arm/keyvault/2019-09-01/armkeyvault/CHANGELOG.md
@@ -1,3 +1,4 @@
-Generated from https://github.com/Azure/azure-rest-api-specs/tree/c72af9b8dca625d3f1429a0bd913bad7b3879645
+# Release History
 
-Code generator @autorest/go@4.0.0-preview.9
+## v0.1.0 (released)
+
--- a/sdk/arm/network/2020-07-01/armnetwork/CHANGELOG.md
+++ b/sdk/arm/network/2020-07-01/armnetwork/CHANGELOG.md
@@ -1,3 +1,4 @@
-Generated from https://github.com/Azure/azure-rest-api-specs/tree/c72af9b8dca625d3f1429a0bd913bad7b3879645
+# Release History
 
-Code generator @autorest/go@4.0.0-preview.9
+## v0.1.0 (released)
+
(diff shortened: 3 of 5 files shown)


ArcturusZhang:
Out of curiosity, why are these changelog files shown as "binary files" on Github?

catalinaperalta:
> Out of curiosity, why are these changelog files shown as "binary files" on Github?

Looking into this, I think there must be some non-ASCII character in the file. Going to try to see how they're being saved.

catalinaperalta:
> > Out of curiosity, why are these changelog files shown as "binary files" on Github?
> 
> Looking into this, I think there must be some non-ASCII/UTF8 character in the file. Going to try to see how they're being saved.

This has been fixed, the files where encoded with UTF16.

ArcturusZhang:
Hi @catalinaperalta @jhendrixMSFT I think it is a good idea that we keep the metadata of "how this package was generated" to contain the parameters that the code generator takes to completely reproduce this generation, which will help our further improvements of the release pipeline. For example, we might need to identify the rename of a package from the changes of readme tags. 
Also this information makes the generation deterministic, any one can regenerate this package and get exactly the same result.

jhendrixMSFT:
So I used to feel strongly about making the build-time metadata available for just such a reason.  However, for track 2 the POR is to "service the latest version".  In my mind, it means that we'd really never consume this metadata as if we needed to rebuild a package it would be from the latest swagger.  We should confirm that this is indeed the plan.

I have no objections to adding it, but if there are complexities adding it I won't be sad if it's missing.

catalinaperalta:
> I think it is a good idea that we keep the metadata of "how this package was generated" to contain the parameters that the code generator takes to completely reproduce this generation, which will help our further improvements of the release pipeline.

@ArcturusZhang is this already automated? How are you feeding this metadata to the tool? (I'm referring to the generator version and specs repo commit hash)

ArcturusZhang:
> > I think it is a good idea that we keep the metadata of "how this package was generated" to contain the parameters that the code generator takes to completely reproduce this generation, which will help our further improvements of the release pipeline.
> 
> @ArcturusZhang is this already automated? How are you feeding this metadata to the tool? (I'm referring to the generator version and specs repo commit hash)

I now have a tool that automatically generates the new releases for me (on my private repo), which will handle these metadata and put them in the changelog file.
But I have a feeling that we could do this by adding this to the code generator, for example by generating a go file full of comments with these information to keep this information unified.

ArcturusZhang:
> So I used to feel strongly about making the build-time metadata available for just such a reason. However, for track 2 the POR is to "service the latest version". In my mind, it means that we'd really never consume this metadata as if we needed to rebuild a package it would be from the latest swagger. We should confirm that this is indeed the plan.
> 
> I have no objections to adding it, but if there are complexities adding it I won't be sad if it's missing.

Clarify a bit for this: some service team has confirmed that they did not guarantee the swaggers on master branch are always shippable, this is also the reason why we no longer do full regeneration on our track 1 SDK.
To work around this, we need to record from which commit hash every package comes. If we have this, we could regenerate every package using new version of code generator and the same commit hash as the one it is generated from.
I have done a round of this in the major version release v50.0.0

jhendrixMSFT:
Which service team(s) said this?  The contract is that, barring bugs of course, all swaggers in the master branch are deemed shippable.  If a service team is explicitly breaching this contract we need to surface this with the arch board as that's a problem.

catalinaperalta:
Joel will be updating to this format as he regenerates all of the modules with the latest code generator.

PR #14124, opened 2021-01-11
Accept zero-value options for default values

jhendrixMSFT:
Specify zero-value options structs to accept default values.
Remove DefaultXxxOptions() methods.
Fixed up docs for RetryOptions.

### DIFF
--- a/sdk/azcore/policy_retry.go
+++ b/sdk/azcore/policy_retry.go
@@ -28,7 +28,7 @@ type RetryOptions struct {
 	MaxRetries int32
 
 	// TryTimeout indicates the maximum time allowed for any single try of an HTTP request.
-	// The default value is one minute.
+	// This is disabled by default.  Specify a value greater than zero to enable.
 	// NOTE: Setting this to a small value might cause premature HTTP request time-outs.
 	TryTimeout time.Duration
 
@@ -80,9 +80,6 @@ func (o *RetryOptions) init() {
 	}
 	if o.StatusCodes == nil {
 		o.StatusCodes = StatusCodesForRetry
-	}
-	if o.TryTimeout == 0 {
-		o.TryTimeout = 1 * time.Minute
 	}
 }
 
@@ -162,11 +159,15 @@ func (p *retryPolicy) Do(req *Request) (resp *Response, err error) {
 			return
 		}
 
-		// Set the per-try time for this particular retry operation and then Do the operation.
-		tryCtx, tryCancel := context.WithTimeout(req.Context(), options.TryTimeout)
-		clone := req.clone(tryCtx)
-		resp, err = clone.Next() // Make the request
-		tryCancel()
+		if options.TryTimeout == 0 {
+			resp, err = req.Next()
+		} else {
+			// Set the per-try time for this particular retry operation and then Do the operation.
+			tryCtx, tryCancel := context.WithTimeout(req.Context(), options.TryTimeout)
+			clone := req.clone(tryCtx)
+			resp, err = clone.Next() // Make the request
+			tryCancel()
+		}
 		if err == nil {
 			Log().Writef(LogRetryPolicy, "response %d", resp.StatusCode)
 		} else {
--- a/sdk/azcore/policy_retry.go
+++ b/sdk/azcore/policy_retry.go
@@ -39,7 +39,7 @@ type RetryOptions struct {
 
 	// MaxRetryDelay specifies the maximum delay allowed before retrying an operation.
 	// Typically the value is greater than or equal to the value specified in RetryDelay.
-	// The default Value is 120 seconds.  A value less than zero means ther is no cap.
+	// The default Value is 120 seconds.  A value less than zero means there is no cap.
 	MaxRetryDelay time.Duration
 
 	// StatusCodes specifies the HTTP status codes that indicate the operation should be retried.
--- a/sdk/azcore/policy_retry.go
+++ b/sdk/azcore/policy_retry.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"math"
 	"math/rand"
 	"net/http"
 	"time"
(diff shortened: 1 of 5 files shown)


sdk/azcore/policy_retry.go
```diff
@@ -19,27 +20,31 @@ const (
 )
 
 // RetryOptions configures the retry policy's behavior.
+// All zero-value fields will be initialized with their default values.
 type RetryOptions struct {
 	// MaxRetries specifies the maximum number of attempts a failed operation will be retried
-	// before producing an error.  A value of zero means one try and no retries.
+	// before producing an error.
+	// The default value is three.  A value less than zero means one try and no retries.
 	MaxRetries int32
```
JeffreyRichter, line 28: What if we called this MaxTries: 0=default, 1=1 try, 2=2 tries, 3=3 tries, etc.
This way there is no -1 -- it just bugs me.

sdk/azcore/policy_retry.go
```diff
@@ -55,14 +60,31 @@ var (
...
+	}
+	if o.MaxRetryDelay == 0 {
+		o.MaxRetryDelay = 120 * time.Second
+	} else if o.MaxRetryDelay < 0 {
+		// not really an unlimited cap, but sufficiently large enough to be considered as such
+		o.MaxRetryDelay = math.MaxInt64
+	}
+	if o.RetryDelay == 0 {
+		o.RetryDelay = 4 * time.Second
+	} else if o.RetryDelay < 0 {
```
JeffreyRichter, line 78: I'm OK with 0s being set to our desired values but I'm not on board with the error checking and other value changing.
I'm afraid, that we silently change a value and now unexpected behavior occurs and the customer has no way to figure it out unless they step through our code. I think bad values should just remain bad values; most likely something will fail, the customer will see the failure, see the value, and correct their code.

This is also hard to maintain, may be inconsistently applied, and may hurt versioning if we ever introduce some sentinel value to have some special meaning for some field.

sdk/azcore/policy_retry.go
jhendrixMSFT: We can do that, and that will address the most common cases.  However, we still have to do _something_ about a negative value since the type allows it even though it semantically makes no sense.  As an aside, we had a bug a few years ago around this exact thing in track 1; some incorrect math was leading to a negative try value, so the request was never even sent leading to a very hard-to-diagnose bug.  It would be preferable to prevent the possibility of such bugs from even occurring (isn't this part of the value-add of an SDK?).

sdk/azcore/policy_retry.go
jhendrixMSFT: This behavior is explicitly documented with this change, so I disagree that we're silently changing it.

> most likely something will fail, the customer will see the failure, see the value, and correct their code.

In trivial apps I can buy this.  However in larger, more complex apps it might not be that straight-forward.

jhendrixMSFT: The only place where we _do_ silently change the value is `TryTimeout` which I can remove (zero or negative values here yield the same result, a deadline exceeded error).

sdk/azcore/policy_retry.go
jhendrixMSFT: If we change it to `MaxTries` then we must return an error from `retryPolicy.Do()` if `MaxTries` is negative.

sdk/azcore/policy_retry.go
jhendrixMSFT: > This is also hard to maintain, may be inconsistently applied, and may hurt versioning if we ever introduce some sentinel value to have some special meaning for some field.

You do have a point here, and I contend that using the variadic approach alleviate these concerns since the values are changed _only_ when a customer explicitly calls a `WithOption()` method.

JeffreyRichter: What if a variadic function is called and passed bad values? Would you return an error or silently correct? Neither of these options is good in my opinion.

sdk/azcore/policy_retry.go
JeffreyRichter: This is the argument for having the XxxClient ctors return an error if any option does not pass validation.

sdk/azcore/policy_retry.go
jhendrixMSFT: We'd need to keep the same semantics as in this change, i.e. zero for default and less than zero for none.

sdk/azcore/policy_retry.go
jhendrixMSFT: We could, however I think it's a high tax to pay (making the ctors not no-fail) when we can design the API in a way to avoid this.

0 == default
> 0 == custom value
< 0 == none

I think this is very clear and assigns semantic meaning to the complete range of an `int`.  There is nothing to validate, all values are valid.

sdk/azcore/policy_retry.go
```diff
@@ -19,27 +20,31 @@ const (
...
-	// equal to or greater than RetryDelay.
+	// RetryDelay specifies the initial amount of delay to use before retrying an operation.
+	// The delay increases exponentially with each retry up to the maximum specified by MaxRetryDelay.
+	// The default value is four seconds.  A value less than zero means no delay between retries.
 	RetryDelay time.Duration
 
 	// MaxRetryDelay specifies the maximum delay allowed before retrying an operation.
-	// If you specify 0, then you must also specify 0 for RetryDelay.
+	// Typically the value is greater than or equal to the value specified in RetryDelay.
+	// The default Value is 120 seconds.  A value less than zero means ther is no cap.
```
catalinaperalta, line 42: Nit: there

catalinaperalta (approved):
LGTM

sdk/azcore/policy_retry.go
jhendrixMSFT: Fixed

jhendrixMSFT:
Added the change to make the per-try timeout opt-in.

JeffreyRichter (approved):

PR #14139, opened 2021-01-12
Switching the environment variable check for ManagedIdentityCredential

### LINKED ISSUE #12981: UnsupportedApiVersion using managed identity in Azure Functions
IDENTITY_ENDPOINT and IDENTITY_HEADER are set in Functions apps on Linux Consumption hosting plans. However, the endpoint returns an error for version 2019-08-01, for example: "(UnsupportedApiVersion) The HTTP resource that matches the request URI 'http://localhost:8081/msi/token?api-version=2019-08-01&resource=https://vault.azure.net' does not support the API version '2019-08-01'"

Lacking a way to tell whether it's in an affected environment, ManagedIdentityCredential should continue to use the prior API version, 2017-09-01, and its associated environment variables MSI_ENDPOINT and MSI_SECRET.

catalinaperalta:
This PR gives priority to checking the MSI_ENDPOINT and MSI_SECRET environment variables over the IDENTITY_ENDPOINT and IDENTITY_HEADER variables. This is because in Azure Functions on Linux consumptions hosts, the latest api-version is not yet supported and causes an error. 

Fixes #12981

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -274,7 +274,14 @@ func (c *managedIdentityClient) createCloudShellAuthRequest(ctx context.Context,
 
 func (c *managedIdentityClient) getMSIType() (msiType, error) {
 	if c.msiType == msiTypeUnknown { // if we haven't already determined the msiType
-		if endpointEnvVar := os.Getenv(identityEndpoint); endpointEnvVar != "" { // check for IDENTITY_ENDPOINT
+		if endpointEnvVar := os.Getenv(msiEndpoint); endpointEnvVar != "" { // if the env var MSI_ENDPOINT is set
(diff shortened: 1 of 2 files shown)


jhendrixMSFT (approved):
Is there any downside to using the older API version even when the newer one is available?

catalinaperalta:
> Is there any downside to using the older API version even when the newer one is available?

At the moment, I dont think so. The previous API version is still supported in all the MSI environments. However, I think at some point in the future we will have to give preference to the new identity header, also the fact that newer versions always (typically) come with improvements and security updates. The only environments that I think would be using the legacy API vs the  current APIs would be App Service and Azure Functions ones. So, for now, I think this is fine and this is the current MSI order followed by the other languages as well.

PR #14147, opened 2021-01-13
azidentity: Remove DefaultXxxOptions() funcs 

catalinaperalta:
This PR removes the DefaultXxxOptions() funcs from every azidentity credential type.

### DIFF
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -42,12 +42,13 @@ type AzureCLICredential struct {
 // NewAzureCLICredential constructs a new AzureCLICredential with the details needed to authenticate against Azure Active Directory
 // options: configure the management of the requests sent to Azure Active Directory.
 func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
-	if options == nil {
-		options = &AzureCLICredentialOptions{}
+	cp := AzureCLICredentialOptions{}
+	if options != nil {
+		cp = *options
 	}
-	options.init()
+	cp.init()
 	return &AzureCLICredential{
-		tokenProvider: options.TokenProvider,
+		tokenProvider: cp.TokenProvider,
 	}, nil
 }
 
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -27,9 +27,11 @@ type AzureCLICredentialOptions struct {
 	TokenProvider AzureCLITokenProvider
 }
 
-// DefaultAzureCLICredentialOptions returns an instance of AzureCLICredentialOptions initialized with default values.
-func DefaultAzureCLICredentialOptions() AzureCLICredentialOptions {
-	return AzureCLICredentialOptions{TokenProvider: defaultTokenProvider()}
+// init returns an instance of AzureCLICredentialOptions initialized with default values.
+func (o *AzureCLICredentialOptions) init() {
+	if o.TokenProvider == nil {
+		o.TokenProvider = defaultTokenProvider()
+	}
 }
 
 // AzureCLICredential enables authentication to Azure Active Directory using the Azure CLI command "az account get-access-token".
@@ -41,9 +43,9 @@ type AzureCLICredential struct {
 // options: configure the management of the requests sent to Azure Active Directory.
 func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
 	if options == nil {
-		def := DefaultAzureCLICredentialOptions()
-		options = &def
+		options = &AzureCLICredentialOptions{}
 	}
+	options.init()
 	return &AzureCLICredential{
 		tokenProvider: options.TokenProvider,
 	}, nil
--- a/sdk/azidentity/bearer_token_policy_test.go
+++ b/sdk/azidentity/bearer_token_policy_test.go
@@ -22,9 +22,10 @@ const (
 )
 
 func defaultTestPipeline(srv azcore.Transport, cred azcore.Credential, scope string) azcore.Pipeline {
-	retryOpts := azcore.DefaultRetryOptions()
-	retryOpts.MaxRetryDelay = 500 * time.Millisecond
-	retryOpts.RetryDelay = 50 * time.Millisecond
+	retryOpts := azcore.RetryOptions{
+		TryTimeout: 1 * time.Minute,
+		MaxRetries: 1,
+	}
 	return azcore.NewPipeline(
 		srv,
 		azcore.NewRetryPolicy(&retryOpts),
@@ -37,7 +38,7 @@ func TestBearerPolicy_SuccessGetToken(t *testing.T) {
 	defer close()
 	srv.AppendResponse(mock.WithBody([]byte(accessTokenRespSuccess)))
 	srv.AppendResponse(mock.WithStatusCode(http.StatusOK))
-	options := DefaultClientSecretCredentialOptions()
+	options := ClientSecretCredentialOptions{}
 	options.AuthorityHost = srv.URL()
 	options.HTTPClient = srv
 	cred, err := NewClientSecretCredential(tenantID, clientID, secret, &options)
@@ -64,7 +65,7 @@ func TestBearerPolicy_CredentialFailGetToken(t *testing.T) {
 	defer close()
 	srv.AppendResponse(mock.WithStatusCode(http.StatusUnauthorized))
 	srv.AppendResponse(mock.WithStatusCode(http.StatusOK))
-	options := DefaultClientSecretCredentialOptions()
+	options := ClientSecretCredentialOptions{}
 	options.AuthorityHost = srv.URL()
 	options.HTTPClient = srv
 	cred, err := NewClientSecretCredential(tenantID, clientID, wrongSecret, &options)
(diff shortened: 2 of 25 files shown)


catalinaperalta:
Leaving this as a draft meanwhile I figure out why this fails in Windows :/

sdk/azidentity/azure_cli_credential.go
```diff
@@ -41,9 +43,9 @@ type AzureCLICredential struct {
 // options: configure the management of the requests sent to Azure Active Directory.
 func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error) {
 	if options == nil {
-		def := DefaultAzureCLICredentialOptions()
-		options = &def
+		options = &AzureCLICredentialOptions{}
 	}
+	options.init()
```
jhendrixMSFT, line 48: You'll want to ensure that you `init()` a copy.  I suggest creating the `AzureCLICredential` type first then calling `init()` on its `options` field.

sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -22,9 +22,9 @@ const (
 )
 
 func defaultTestPipeline(srv azcore.Transport, cred azcore.Credential, scope string) azcore.Pipeline {
-	retryOpts := azcore.DefaultRetryOptions()
-	retryOpts.MaxRetryDelay = 500 * time.Millisecond
-	retryOpts.RetryDelay = 50 * time.Millisecond
+	retryOpts := azcore.RetryOptions{
+		TryTimeout: 1 * time.Minute,
```
jhendrixMSFT, line 26: Looks like the original values for `MaxRetryDelay` and `RetryDelay` were removed, was that intentional?

sdk/azidentity/device_code_credential.go
```diff
@@ -82,9 +83,9 @@ type DeviceCodeCredential struct {
 // options: Options used to configure the management of the requests sent to Azure Active Directory, please see DeviceCodeCredentialOptions for a description of each field.
 func NewDeviceCodeCredential(options *DeviceCodeCredentialOptions) (*DeviceCodeCredential, error) {
 	if options == nil {
-		temp := DefaultDeviceCodeCredentialOptions()
-		options = &temp
+		options = &DeviceCodeCredentialOptions{}
 	}
+	options.init()
```
jhendrixMSFT, line 88: Perform `init()` on a copy.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -66,9 +65,9 @@ type InteractiveBrowserCredential struct {
 // options: allow to configure the management of the requests sent to Azure Active Directory, pass in nil for default behavior.
 func NewInteractiveBrowserCredential(options *InteractiveBrowserCredentialOptions) (*InteractiveBrowserCredential, error) {
 	if options == nil {
-		temp := DefaultInteractiveBrowserCredentialOptions()
-		options = &temp
+		options = &InteractiveBrowserCredentialOptions{}
 	}
+	options.init()
```
jhendrixMSFT, line 70: Perform `init()` on a copy.

sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -174,15 +174,22 @@ func TestBearerPolicy_GetTokenFailsNoDeadlock(t *testing.T) {
 		HTTPClient:    srv,
 		AuthorityHost: srv.URL(),
 		Retry: azcore.RetryOptions{
-			// leaving TryTimeout at zero will trigger a deadline exceeded error causing GetToken() to fail
-			RetryDelay:    50 * time.Millisecond,
-			MaxRetryDelay: 500 * time.Millisecond,
-			MaxRetries:    3,
+			// leaving TryTimeout at as a very small value will trigger a deadline exceeded error causing GetToken() to fail
+			TryTimeout: 1 * time.Nanosecond,
+			MaxRetries: 0,
```
jhendrixMSFT, line 179: Missing original values for `RetryDelay` and `MaxRetryDelay`.  Also, the zero here is redundant (doesn't hurt anything of course).

sdk/azidentity/azure_cli_credential.go
catalinaperalta: The AzureCLICredential doesnt have the options struct on it. It has the TokenProvider directly on it, that's why I called the init before intializing it. I can create a copy of the options, initialize that and then assign it to the credential. Does that sounds like a good plan?

sdk/azidentity/device_code_credential.go
catalinaperalta: Same comment as AzureCLICredential. Let me know if the other solutions works!

sdk/azidentity/azure_cli_credential.go
jhendrixMSFT: Ah I see, I missed that.  Should it have the options in it?  E.g. if another option is added in the future.  If not your suggestion works fine.

jhendrixMSFT: Hmm yes I remember.  Let's not change it now, we can in the future if the need arises.

catalinaperalta: Done

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: Done..added them back in

sdk/azidentity/device_code_credential.go
catalinaperalta: Done

sdk/azidentity/interactive_browser_credential.go
catalinaperalta: Done

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: Done and fixed, it was supposed to be a -1 so it doesnt retry.

catalinaperalta:
Adding an issue for updating the doc comments after this gets merged: #14175.

sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -22,9 +22,11 @@ const (
 )
 
 func defaultTestPipeline(srv azcore.Transport, cred azcore.Credential, scope string) azcore.Pipeline {
-	retryOpts := azcore.DefaultRetryOptions()
-	retryOpts.MaxRetryDelay = 500 * time.Millisecond
-	retryOpts.RetryDelay = 50 * time.Millisecond
+	retryOpts := azcore.RetryOptions{
+		MaxRetryDelay: 500 * time.Millisecond,
+		RetryDelay:    50 * time.Millisecond,
+		TryTimeout:    1 * time.Minute,
```
jhendrixMSFT, line 28: This is no longer the default, is it required for tests to pass?

sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -174,15 +176,22 @@ func TestBearerPolicy_GetTokenFailsNoDeadlock(t *testing.T) {
 		HTTPClient:    srv,
 		AuthorityHost: srv.URL(),
 		Retry: azcore.RetryOptions{
-			// leaving TryTimeout at zero will trigger a deadline exceeded error causing GetToken() to fail
-			RetryDelay:    50 * time.Millisecond,
-			MaxRetryDelay: 500 * time.Millisecond,
-			MaxRetries:    3,
+			// leaving TryTimeout at as a very small value will trigger a deadline exceeded error causing GetToken() to fail
+			TryTimeout: 1 * time.Nanosecond,
+			MaxRetries: -1,
```
jhendrixMSFT, line 181: This will trigger no retries which invalidates the test.

catalinaperalta: Yes I noticed this. I'll reverse it, but I'm looking into it and I've made sure I have all of the same configurations as before and this test is still failing. Could something have changed down the pipeline? (I'm also thinking of skipping this test and opening an issue to resolve this in it's own PR)

sdk/azidentity/bearer_token_policy_test.go
```diff
@@ -174,15 +176,22 @@ func TestBearerPolicy_GetTokenFailsNoDeadlock(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Unable to create credential. Received: %v", err)
 	}
-	pipeline := defaultTestPipeline(srv, cred, scope)
+	pipeline := azcore.NewPipeline(
+		srv,
+		azcore.NewRetryPolicy(&azcore.RetryOptions{
+			MaxRetryDelay: 500 * time.Millisecond,
+			RetryDelay:    50 * time.Millisecond,
+			MaxRetries:    -1,
```
jhendrixMSFT, line 191: This also disables retries.  Can it be switched back to `defaultTestPipeline()`?

sdk/azidentity/bearer_token_policy_test.go
jhendrixMSFT: I think we need to understand what's happening with the failure.  The change to default values should have had zero impact (i.e. the old/new pattern should yield identical default values).

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: This is the same test as above. I'm looking into why this test is broken.

catalinaperalta: Oh, I also agree it was wrong to put MaxRetries at -1. I'll make sure the fix doesnt have that.

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: The problem is that now, instead of getting an error the test doesnt receive an error and gets the successful response.

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: Fixed this, left the values for MaxRetryDelay and RetryDelay, removed TryTimeout. Done.

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: Set TryTimeout as a negative value like you recommended. Done.

sdk/azidentity/bearer_token_policy_test.go
catalinaperalta: Left this using the defaultTestPipeline(). Done.

jhendrixMSFT (approved):

PR #14161, opened 2021-01-14
The output-folder for a profile now no longer reports errors

ArcturusZhang:
 
### DIFF
--- a/tools/generator/validate/validation.go
+++ b/tools/generator/validate/validation.go
@@ -77,7 +77,7 @@ func MgmtCheck(ctx *MetadataValidateContext, tag string, metadata model.Metadata
 			return err
 		}
 		if !mgmtOutputRegex.MatchString(rel) {
-			return fmt.Errorf("the output-folder of a management plane package '%s' must be in this pattern: '^services(/preview)?/[^/]+/mgmt/[^/]+/[^/]+$'", tag)
+			return fmt.Errorf("the output-folder of a management plane package '%s' is expected to have this pattern: 'services/(preview)?/{RPname}/mgmt/{packageVersion}/{namespace}'", tag)
 		}
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


tools/generator/validate/validation.go
```diff
@@ -77,7 +77,7 @@ func MgmtCheck(ctx *MetadataValidateContext, tag string, metadata model.Metadata
 			return err
 		}
 		if !mgmtOutputRegex.MatchString(rel) {
-			return fmt.Errorf("the output-folder of a management plane package '%s' must be in this pattern: '^services(/preview)?/[^/]+/mgmt/[^/]+/[^/]+$'", tag)
+			return fmt.Errorf("the output-folder of a management plane package '%s' is expected to have this pattern: 'services/(preview)?/{RPname}/mgmt/{packageVersion}/{namespace}'", tag)
```
yungezz, line 80: the [readme](https://github.com/Azure/azure-rest-api-specs/blob/7caa9127f88b5d8bbae431fcc2fc5872f6c1ee24/specification/compute/resource-manager/readme.go.md#tag-profile-hybrid-2020-09-01-and-go)  output folder is like `output-folder: $(go-sdk-folder)/profiles/2020-09-01/compute/mgmt/compute`, .../profiles/{pacakgeversion}/{RPname}/mgmt/{RPname} again, so service team need fix readme anyway right?

ArcturusZhang: We have some special exceptions for the output-folder (like the one reporting error in this case), which are only used for `profiles` and directly maintained by myself, therefore I did not mention that special case in this error message to keep this message brief.
You could check the change of the regex below, that is the main part of this fix. Previously this validation reports the output folder starting with `profiles` as an error, now it is included in the regex.

yungezz (approved):

PR #14164, opened 2021-01-14
Updating apidiff tool to use track2 format

catalinaperalta:
Updating the apidiff tool to use track2 format.

### DIFF
--- a/tools/apidiff/cmd/diff.go
+++ b/tools/apidiff/cmd/diff.go
@@ -23,12 +23,12 @@ import (
 )
 
 var diffCmd = &cobra.Command{
-	Use:   "diff <base export filepath> <target export filepath>",
+	Use:   "diff <base export filepath> <target export filepath> <release tag version> <release marker>",
 	Short: "Generate a diff report between the two export report files",
 	Long:  `The diff command consumes two JSON files with the export reports, and generates a diff report between them.`,
-	Args:  cobra.ExactArgs(2),
+	Args:  cobra.ExactArgs(4),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return diffCommand(args[0], args[1])
+		return diffCommand(args[0], args[1], args[2], args[3])
 	},
 }
 
@@ -37,7 +37,7 @@ func init() {
 	diffCmd.PersistentFlags().BoolVarP(&asMarkdown, "markdown", "m", false, "emits the report in markdown format")
 }
 
-func diffCommand(basePath, targetPath string) error {
+func diffCommand(basePath, targetPath, version, marker string) error {
 	base, err := ioutil.ReadFile(basePath)
 	if err != nil {
 		return fmt.Errorf("failed to read base export file %s: %+v", basePath, err)
@@ -57,7 +57,7 @@ func diffCommand(basePath, targetPath string) error {
 	r := getPkgsReport(baseExport, targetExport)
 
 	if asMarkdown {
-		fmt.Println(r.ToMarkdown())
+		fmt.Println(r.ToMarkdown(true, version, marker))
 	} else {
 		b, _ := json.Marshal(r)
 		fmt.Println(string(b))
--- a/tools/apidiff/cmd/diff.go
+++ b/tools/apidiff/cmd/diff.go
@@ -57,7 +57,7 @@ func diffCommand(basePath, targetPath, version string) error {
 	r := getPkgsReport(baseExport, targetExport)
 
 	if asMarkdown {
-		fmt.Println(r.ToMarkdown(true, version))
+		fmt.Println(r.ToMarkdown(version))
 	} else {
 		b, _ := json.Marshal(r)
 		fmt.Println(string(b))
--- a/tools/apidiff/cmd/diff.go
+++ b/tools/apidiff/cmd/diff.go
@@ -23,12 +23,12 @@ import (
 )
 
 var diffCmd = &cobra.Command{
-	Use:   "diff <base export filepath> <target export filepath> <release tag version> <release marker>",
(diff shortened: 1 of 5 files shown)


catalinaperalta:
@ArcturusZhang would these changes break anything on your end?

tools/apidiff/markdown/markdown.go
```diff
@@ -37,18 +37,34 @@ func (md *Writer) checkNL() {
...
+	md.sb.WriteString("### ")
 	md.sb.WriteString(h)
 	md.sb.WriteString("\n\n")
 }
 
 // WriteSubheader writes a sub-header to the markdown document
 func (md *Writer) WriteSubheader(sh string) {
 	md.checkNL()
-	md.sb.WriteString("### ")
+	md.sb.WriteString("#### ")
```
ArcturusZhang, line 67: This change might change the style of the changelog I am generating in the release tool and the generation script running in SDK automation pipeline. 
But I am OK I could change them accordingly

ArcturusZhang (commented):
Generally I am OK with these changes. 
Are you OK if I include the corresponding changes in the `generator` tool in this PR? I am using the `WriteHeader` function and I need to change them to `WriteTopLevelHeader` to keep the changelog style unchanged.

catalinaperalta:
> Generally I am OK with these changes.
> Are you OK if I include the corresponding changes in the `generator` tool in this PR? I am using the `WriteHeader` function and I need to change them to `WriteTopLevelHeader` to keep the changelog style unchanged.

Ok great! Thanks for the input. It's fine by me if you add the generator changes.

ArcturusZhang:
Hi @catalinaperalta there is a problem: this PR changes the output of `report.Package`. The content of CHANGELOG.md is generated using the `ToMarkdown` function. The change in this PR will add a `#` to all the headers generated by the `report.Package.ToMarkdown` which potentially affects all the CHANGELOG.md in every package.
Is there a way that we could make this configurable?

catalinaperalta:
> Hi @catalinaperalta there is a problem: this PR changes the output of `report.Package`. The content of CHANGELOG.md is generated using the `ToMarkdown` function. The change in this PR will add a `#` to all the headers generated by the `report.Package.ToMarkdown` which potentially affects all the CHANGELOG.md in every package.
> Is there a way that we could make this configurable?

@ArcturusZhang I'm not sure I'd want to fork based on this. I think if we adopt the new track2 format it would only require a one time formatting change to the generated changelogs and the other content would remain unchanged.

ArcturusZhang:
@catalinaperalta OK, I am fine with the changes - it only affects the changelog not the code itself. We can fix the issues when we have some

catalinaperalta:
@ArcturusZhang thank you! @jhendrixMSFT I think you can review this now. :)

tools/apidiff/cmd/diff.go
```diff
@@ -23,12 +23,12 @@ import (
 )
 
 var diffCmd = &cobra.Command{
-	Use:   "diff <base export filepath> <target export filepath>",
+	Use:   "diff <base export filepath> <target export filepath> <release tag version> <release marker>",
```
jhendrixMSFT, line 26: Add explanation for new params.

jhendrixMSFT: Remove "release marker" per our discussion.

tools/apidiff/markdown/markdown_test.go
```diff
@@ -1,22 +1,38 @@
...
 	md := Writer{}
+	md.WriteTitle("Title")
 	md.WriteHeader("Header1")
 	md.WriteSubheader("Sub-header1")
 	md.WriteLine("Foo")
 	md.WriteSubheader("Sub-header2")
 	md.WriteLine(getTable().String())
 	result := md.String()
-	expected := `## Header1
+	expected := `# Title
```
jhendrixMSFT, line 27: Missing WriteTopLevelHeader()

tools/apidiff/report/packages.go
```diff
@@ -159,17 +173,25 @@ func (r PkgsReport) IsEmpty() bool {
 }
 
 // ToMarkdown writes the report to string in the markdown form
-func (r *PkgsReport) ToMarkdown() string {
+func (r *PkgsReport) ToMarkdown(withHeader bool, version, marker string) string {
```
jhendrixMSFT, line 176: Why do we need the `withHeader` param?

catalinaperalta: I added this so that anybody creating a new command and who only wants the pure diff without the headers can do so by calling this with the bool set to false.

tools/apidiff/markdown/markdown_test.go
catalinaperalta: Done

tools/apidiff/cmd/diff.go
catalinaperalta: Done and done.

tools/apidiff/report/packages.go
catalinaperalta: I'll also add a comment about this to explain that

jhendrixMSFT: I see.  Looking at the code, it would appear that the `version` param is only used when `withHeader` is `true`.  Perhaps we can condense this to just `withHeader`, and if it's not the empty string write the header with the version.  What do you think?

catalinaperalta: Yeah, that's a good idea. I'll change it.

catalinaperalta: Done

jhendrixMSFT (approved):

PR #14185, opened 2021-01-20
PROPOSAL: Adding generator and specs repo info to APIDIFF output

catalinaperalta:
The following additions serve as a convenience when generating diffs between packages. The optional arguments would enable developers to output the diff with the azure-rest-api-specs repo commit hash and the Go generator version beneath the release header. 

Another solution for adding this information could be to simply use a script that runs through generated diff documents and appends the information to those files as needed.

### DIFF
--- a/tools/apidiff/cmd/changelog.go
+++ b/tools/apidiff/cmd/changelog.go
@@ -26,13 +26,17 @@ import (
 )
 
 var changelogCmd = &cobra.Command{
-	Use:   "changelog <package search dir> <base commit> <target commit> <release tag version>",
+	Use:   "changelog <package search dir> <base commit> <target commit> <release tag version> [<specs repo commit hash> <Go generator version>]",
 	Short: "Generates a CHANGELOG report in markdown format for the packages under the specified directory.",
 	Long: `The changelog command generates a CHANGELOG for all of the packages under the directory specified in <package dir>.
-A table for added, removed, updated, and breaking changes will be created as required.`,
+A table for added, removed, updated, and breaking changes will be created as required.
+Optional arguments for generated code include the azure-rest-api-specs commit hash that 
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
Hi @catalinaperalta This looks good to me, but I suppose we still need the relative path of the `readme.md` file and the corresponding tag to keep the metadata complete. 
I am using commit hash, readme file relative path, tag name, and the code generator version to track every package in their CHANGELOGs, so that we could identify it when there is a change of output-folder or rename of tag, etc.

catalinaperalta:
> but I suppose we still need the relative path of the readme.md file and the corresponding tag to keep the metadata complete

@ArcturusZhang we can add any additional metadata we want, only I see this adding more args to the command. I also think we should all agree on what additional information we want to output in the changelogs. @jhendrixMSFT do you also have an opinion on also adding the readme.md and tag to the changelog?

jhendrixMSFT:
We can also just append content to the file via the shell, e.g. `ver v1.2.3 commit abc >> changelog.md` or something.

ArcturusZhang:
@catalinaperalta @jhendrixMSFT I have been using this format in track 1 packages now, please take this file as an example: https://github.com/Azure/azure-sdk-for-go/blob/master/services/aad/mgmt/2017-04-01/aad/CHANGELOG.md

ArcturusZhang:
And for your information, I have written some utilities for writing and reading the generation metadata, please check it out here: https://github.com/Azure/azure-sdk-for-go/blob/2748f9c295d18b949cb1bfacb6bce08ad1d87501/tools/generator/autorest/generationMetadata.go#L22
Please feel free to reuse or refactor this to make it more general if necessary

catalinaperalta:
Per offline discussion, we agreed that adding the metadata doesn't really belong with the `changelog` command.

PR #14216, opened 2021-01-25
Ensure interactive auth flow can be cancelled

jhendrixMSFT:
Replaced WaitGroup with channel.
Flow context to WaitForCallback() and select on both channels.
Disabling PKCE should not be allowed.  The functionality was disabled
and the config option can be removed in the next breaking change.

### DIFF
--- a/sdk/azidentity/interactive_browser_credential.go
+++ b/sdk/azidentity/interactive_browser_credential.go
@@ -34,10 +34,6 @@ type InteractiveBrowserCredentialOptions struct {
 	// The localhost port for the local server that will be used to redirect back.
 	// By default, a random port number will be selected.
 	Port int
-	// Deprecated: Disables use of PKCE during authorization.
-	// The default is false.
-	// TODO: remove
-	DisablePKCE bool
 	// The host of the Azure Active Directory authority. The default is AzurePublicCloud.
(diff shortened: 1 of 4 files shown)


sdk/azidentity/interactive_browser_credential.go
```diff
@@ -34,8 +34,9 @@ type InteractiveBrowserCredentialOptions struct {
 	// The localhost port for the local server that will be used to redirect back.
 	// By default, a random port number will be selected.
 	Port int
-	// Disables use of PKCE during authorization.
+	// Deprecated: Disables use of PKCE during authorization.
 	// The default is false.
+	// TODO: remove
 	DisablePKCE bool
```
catalinaperalta, line 40: Do we need to leave this note since we're in preview?

jhendrixMSFT: We could remove it now, just we had a minor bump last week (I probably overthink it).

catalinaperalta: I mean I understand where you're coming from...but I feel like leaving it like this during preview we might accidentally forget to remove it at some point and then cause ourselves problems? If we leave it like this then I guess I'd say let's open an issue to track removing it?

jhendrixMSFT: Removed

catalinaperalta (approved):
LGTM

PR #14277, opened 2021-02-04
Add Pager and Poller interfaces to azcore

jhendrixMSFT:
 
### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -6,6 +6,7 @@
 package azcore
 
 import (
+	"context"
 	"errors"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Once we have generics we can move the remaining type-specific methods to these interfaces and remove them from codegen.

sdk/azcore/core.go
```diff
@@ -109,3 +110,38 @@ func (n nopCloser) Close() error {
...
+	// Poll fetches the latest state of the LRO.  It returns an HTTP response or error.
+	// If the LRO has completed successfully, the poller's state is update and the HTTP
+	// response is returned.
+	// If the LRO has completed with failure or was cancelled, the poller's state is
+	// updated and the error is returned.
+	// If the LRO has not reached a terminal state, the poller's state is updated and
+	// the latest HTTP response is returned.
+	// If Poll fails, the poller's state is unmodified and the error is returned.
+	// Calling Poll on an LRO that has reached a terminal state will return the final
+	// HTTP response or error.
```
catalinaperalta, line 131: Calling Poll may sometimes return the final response, but not always. There are several cases where FinalResponse needs to be called to get the true final response.

jhendrixMSFT: It should provide the final HTTP response of the LRO (terminal/non-terminal state).  You are correct though that it might not contain the final result (i.e. the response body of the modified object).

catalinaperalta (approved):
LGTM

sdk/azcore/core.go
```diff
@@ -109,3 +110,38 @@ func (n nopCloser) Close() error {
 func NopCloser(rs io.ReadSeeker) ReadSeekCloser {
 	return nopCloser{rs}
 }
+
+// Poller provides operations for checking the state of a long-running operation.
+// An LRO can be in either a non-terminal or terminal state.  A non-terminal state
+// indicates the LRO is still in progress.  A terminal state indicates the LRO has
+// completed successfully, failed, or was cancelled.
+type Poller interface {
```
tombuildsstuff, line 118: would it make sense to include the "Complete" Poller interface here too (e.g. this is the backing value for `GetX`, but rather for `GetXComplete`?)

jhendrixMSFT: Putting the [`PollUntilDone()`](https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/arm/compute/2020-09-30/armcompute/zz_generated_models.go#L63) method on the response envelope was an intentional decision.  The reasoning is to promote the "happy path" for the common case where a consumer can simply wait for the LRO to complete without needing a custom polling loop.  The other problem here is that `PollUntilDone()` returns a strongly-typed result so we can't place it inside `azcore` until we have generics (if we really wanted to, we'd either have to use an `interface{}` out param, or change pattern to be similar to track 1).

jhendrixMSFT: Generally speaking, in track 2 we're taking the approach of trying to promote the easy/common APIs, leaving the "complex" APIs/patterns under a level of indirection (if that makes sense).  This is also why we don't expose the "*createRequest" and "*handleResponse" methods in the default public surface area; they're considered an advanced scenario (still looking for a clean way to solve this one).

JeffreyRichter (approved):

PR #14376, opened 2021-03-03
Release track 2 msi module 2018-11-30

ArcturusZhang:
 

sdk/arm/msi/2018-11-30/armmsi/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/Azure/azure-sdk-for-go/sdk/arm/msi/2018-11-30/armmsi
```
jhendrixMSFT, line 1: The `go.mod` and `go.sum` files should go in the `./sdk/arm/msi` directory.  This will also cause the module name to change accordingly.

ArcturusZhang: I have moved the `go.mod` file

sdk/arm/msi/2018-11-30/armmsi/zz_generated_connection.go
```diff
@@ -7,4 +7,4 @@
 
 package armmsi
 
-const telemetryInfo = "azsdk-go-armmsi/v0.2.0"
+const telemetryInfo = "azsdk-go-armmsi/v0.3.0"
```
jhendrixMSFT, line 10: I think this can remain at `v0.1.0` since we haven't released any modules for this.

Also, I just realized that with our one-module-per-RP, we'll need to include the API version in the telemetry string.

ArcturusZhang: Ah, you are seeing the version number bumping because I am releasing this module in my forked and replace it in terraform for my testing. I will make a new PR for the actual release.

I suppose if we need to include api-version here, we might need to do it in autorest.go?

jhendrixMSFT: Unlikely as the code generator doesn't have this info.  It will likely need to be added when the module version is added.

PR #14437, opened 2021-03-18
Add helpers for arrays

jhendrixMSFT:
 
### DIFF
--- a/sdk/to/to.go
+++ b/sdk/to/to.go
@@ -42,8 +42,8 @@ func TimePtr(t time.Time) *time.Time {
 	return &t
 }
 
-// ArrayOfInt32Ptr returns an array of *int32 from the specified values.
-func ArrayOfInt32Ptr(vals ...int32) []*int32 {
+// Int32PtrArray returns an array of *int32 from the specified values.
+func Int32PtrArray(vals ...int32) []*int32 {
 	arr := make([]*int32, len(vals), len(vals))
 	for i := range vals {
 		arr[i] = Int32Ptr(vals[i])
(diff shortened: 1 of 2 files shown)


sdk/to/to.go
```diff
@@ -41,3 +41,66 @@ func StringPtr(s string) *string {
 func TimePtr(t time.Time) *time.Time {
 	return &t
 }
+
+// ArrayOfInt32Ptr returns an array of *int32 from the specified values.
```
jhendrixMSFT, line 45: SliceOf.... ?

catalinaperalta: I think for being idiomatic slice makes more sense, but I also know that this may be used with fields that have array in the name. I'd be ok with SliceOf

catalinaperalta (approved):
LGTM

sdk/to/to.go
jhendrixMSFT: You bring up a good point.  I think in our codegen we use the term "array".  If we wanted to use "slice" here, then IMO we should update the codegen to call it "slice" there too.  I will add though that swagger uses the term "array".

JeffreyRichter (approved):

sdk/to/to.go
jhendrixMSFT: In our codegen we call these `StringArray`, `Int32Array`, etc.  Perhaps we should rename these to `Int32PtrArray()` etc for consistency?  @JeffreyRichter

JeffreyRichter: I do like consistency, so yes, I think so.

jhendrixMSFT: Fixed

PR #14441, opened 2021-03-22
Initial generation of the aztables module

christothes:
Includes initial scaffolding for the `TableClient` including the `SharedKeyCredential`


sdk/tables/aztables/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Release History
+
+## 12.0.0-beta.1 (Unreleased)
```
catalinaperalta, line 3: Shouldnt this start at v0.1.0? Since it'll be a new module, or am I missing some context here?

christothes: Cross language we decided to synchronize on v12 as the starting point because it is the latest version when taking track1 into account across Comsos tables and storage tables

jhendrixMSFT: The short answer is that aligning versions across languages isn't really an option for Go.  Longer answer, major versions have significant implications in Go, please see [semantic import versioning](https://github.com/golang/go/wiki/Modules#semantic-import-versioning) for the gory details.

sdk/tables/aztables/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+Common uses of the Table service include:
+
+- Storing TBs of structured data capable of serving web scale applications
+- Storing datasets that don't require complex joins, foreign keys, or stored procedures and can be de-normalized for fast access
+- Quickly querying data using a clustered index
+- Accessing data using the OData protocol and LINQ filter expressions
+
+### Thread safety
+We guarantee that all client instance methods are thread-safe and independent of each other ([guideline](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-service-methods-thread-safety)). This ensures that the recommendation of reusing client instances is always safe, even across threads.
```
catalinaperalta, line 58: If we want to link to this, should it instead be a section we discuss and add to the Go guidelines? In order to consider and add Go-specific conventions. (NOTE: I havent clicked on the link itself yet)

sdk/tables/aztables/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Additional concepts
+<!-- CLIENT COMMON BAR -->
+[Client options]() |
+[Accessing the response]() |
+[Long-running operations]() |
+[Handling failures]() |
+[Diagnostics]() |
+[Mocking]() |
+[Client lifetime](https://devblogs.microsoft.com/azure-sdk/lifetime-management-and-thread-safety-guarantees-of-azure-sdk-net-clients/)
```
catalinaperalta, line 68: Same question as above.

sdk/tables/aztables/shared_policy_shared_key_credential.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// NewSharedKeyCredential creates an immutable SharedKeyCredential containing the
+// storage account's name and either its primary or secondary key.
+func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredential, error) {
```
jhendrixMSFT, line 24: This is the one copied from azblob yes?

sdk/tables/aztables/README.md
christothes: Thanks - Yes, this was just an initial copy/paste from .NET. I'll remove for now unless we have a go equivalent statement

sdk/tables/aztables/README.md
christothes: same answer as above 😄

sdk/tables/aztables/shared_policy_shared_key_credential.go
christothes: It's copied from there, but tables is slightly different in how it builds the stringToSign.

sdk/tables/aztables/tableClient.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// A TableClient represents a URL to an Azure Storage blob; the blob may be a block blob, append blob, or page blob.
+type TableClient struct {
+	client *tableClient
+	cred   SharedKeyCredential
+}
+
+// NewTableClient creates a TableClient object using the specified URL and request policy pipeline.
+func NewTableClient(serviceURL string, cred azcore.Credential, options *ClientOptions) (TableClient, error) {
```
catalinaperalta, line 20: I'm wondering if ClientOptions should be called TableClientOptions? Thoughts? Would there ever be any other client in this module? Is this what azblob did?

catalinaperalta: I just checked azblob, I see they left it as ClientOptions.

christothes: It was my intuition to name it `TableClientOptions` as well. I did copy this naming from azblob, but I also prefer the more specific name if `ClientOptions` is not a convention.

catalinaperalta: I'm not sure, I can see why it's just client options...since it'll probably only ever have options related to modifying the policies. So I'm unsure. I agree that my first instinct was to also go with `TableClientOptions`. I think this might be a better question for Jeff when he sees this.

sdk/tables/aztables/tableClient.go
```diff
@@ -0,0 +1,41 @@
...
+		return TableResponseResponse{}, err
+	} else {
+		return resp.(TableResponseResponse), err
+	}
+}
+
+// Delete
+func (t TableClient) Delete(ctx context.Context, name string) (TableDeleteResponse, error){
+	return t.client.Delete(ctx, name, nil)
+}
```
catalinaperalta, line 41: Nit: newline

sdk/tables/aztables/tableClient.go
jhendrixMSFT: Does `azblob` share the `ClientOptions` across all its clients?  If so, that's probably why they went with a generic name.

catalinaperalta:
General comment: there should be a newline at the end of the files

sdk/tables/aztables/tableClient.go
catalinaperalta: Yes I think so. I think the naming for the options will depend on how many other clients there will be and what options they'll have and if in the future we think that the options could vary between clients to a degree where we should have individually named options for client.

sdk/tables/aztables/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+### Delete table entities
+
+If we no longer need our new table entity, it can be deleted.
+
+
+## Troubleshooting
+
+When you interact with the Azure table library using the .NET SDK, errors returned by the service correspond to the same HTTP
```
catalinaperalta, line 127: I think any reference to .NET should be removed before merging. We can add the Go specific information later.

sdk/tables/aztables/tableClient.go
christothes: changing to to TableClientOptions for now - we can always revisit later if needed.

catalinaperalta (approved):
It's looking great. Did not take too close a look at tests yet, but I dont see why this cant be merged into your feature branch.

sdk/tables/aztables/shared_policy_shared_key_credential.go
```diff
@@ -0,0 +1,161 @@
...
+	accountKey  atomic.Value // []byte
+}
+
+// AccountName returns the Storage account's name.
+func (c *SharedKeyCredential) AccountName() string {
+	return c.accountName
+}
+
+// SetAccountKey replaces the existing account key with the specified account key.
+func (c *SharedKeyCredential) SetAccountKey(accountKey string) error {
```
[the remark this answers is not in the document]
christothes: yes, it can be changed for the purpose of key rotation of an existing client.

sdk/tables/aztables/tableClient.go
```diff
@@ -0,0 +1,39 @@
...
+	return TableClient{client: &tableClient{con}, cred: *c}, nil
+}
+
+// Create
+func (t TableClient) Create(ctx context.Context, name string) (TableResponseResponse, error) {
+	resp, err := t.client.Create(ctx, TableProperties{&name}, nil, nil)
+	if resp == nil {
+		return TableResponseResponse{}, err
+	} else {
+		return resp.(TableResponseResponse), err
```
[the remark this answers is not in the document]
christothes: It will, unless we expose the responseType option, which controls how much data is returned by the service.

sdk/tables/aztables/shared_policy_shared_key_credential.go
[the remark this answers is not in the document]
catalinaperalta: Ok sounds good then!

PR #14461, opened 2021-03-26
Add support for multi-part form data

jhendrixMSFT:
Added method WriteMultipartFormData() to Request.

### DIFF
--- a/sdk/azcore/request.go
+++ b/sdk/azcore/request.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime/multipart"
 	"net/http"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
This was adapted from [track 1](https://github.com/Azure/go-autorest/blob/master/autorest/preparer.go#L306) and has been tested against the local test server.

sdk/azcore/request.go
```diff
@@ -187,6 +188,44 @@ func (req *Request) SetBody(body ReadSeekCloser, contentType string) error {
 	return nil
 }
 
+// WriteMultipartFormData writes the specified keys/values as multi-part form
+// fields with the specified value.  File content must be specified as a ReadSeekCloser.
+// All other values are treated as string values.
+func (req *Request) WriteMultipartFormData(formData map[string]interface{}) error {
```
jhendrixMSFT, line 194: Maybe this should be `SetMultipartFormData()`?  Similar to `SetBody()`

catalinaperalta: For consistency, I would say Set is better.

catalinaperalta (approved):
LGTM! I think Set would be better to be consistent with SetBody.

JeffreyRichter (approved):

sdk/azcore/request.go
jhendrixMSFT: Agreed, changed.

PR #14491, opened 2021-04-05
Introduce internal/testFramework and test aztables with it

christothes:
# Summary

This PR introduces a new package in the `internal` module called `testframework`. It implements the following functionality:
- Allows for track2 clients to easily wire up to recorded tests
- Does sanitation of secrets and allows for total customization of what is sanitized
- Allows configuration of custom matching logic
- Allows test recordings to leverage recorded variables which are used by test recording playbacks
- Allows generation of recorded random values that are seeded with a recorded variable so that random generation is deterministic on playback. Includes support for:
  - alpha-numeric values
  - uuid
  - time.Time values based on recorded "now" value

The `aztables` module was used as a test bed for the `testframework`. The `testframework` changes can be extracted into a separate PR so that this can get to master, once reviewed and approved.


sdk/tables/aztables/tableClient_test.go
```diff
@@ -4,9 +4,115 @@
...
+
+func (s *tableClientLiveTests) SetUpTest(c *chk.C) {
+	// setup the test environment
+	recordedTestSetup(c, testKey(c, s.endpointType), s.endpointType, s.mode)
+}
+
+func (s *tableClientLiveTests) TearDownTest(c *chk.C) {
+	// teardown the test context
+	recordedTestTeardown(testKey(c, s.endpointType))
+}
```
christothes, line 105: example of how a test would leverage the recorded test framework 1/2

sdk/tables/aztables/zt_tableRecordedTests.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	client, err := NewTableServiceClient(uri, cred, &TableClientOptions{HTTPClient: recording, Retry: azcore.RetryOptions{MaxRetries: -1}})
+	c.Assert(err, chk.IsNil)
+	clientsMap[testName] = &testContext{client: client, recording: recording}
+}
+
+func recordedTestTeardown(key string) {
+	clientsMap[key].recording.Stop()
+}
```
christothes, line 83: example of how a test would leverage the recorded test framework 2/2

sdk/tables/aztables/zt_tableRecordedTests.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func getTestContext(key string) *testContext {
+	return clientsMap[key]
+}
+
+func getTableName(context *testContext, prefix ...string) string {
+	if len(prefix) == 0 {
+		return context.recording.GenerateAlphaNumericId(tableNamePrefix, 20, true)
```
christothes, line 108: example of using recorded random Id generation

sdk/tables/aztables/recordings/tableClientLiveTests(cosmos)/TestAddEntity-variables.yaml
```diff
@@ -0,0 +1,5 @@
+---
+COSMOS_TABLES_ENDPOINT_SUFFIX: cosmos.azure.com
+TABLES_COSMOS_ACCOUNT_NAME: chrisstablesprim
+TABLES_PRIMARY_COSMOS_ACCOUNT_KEY: Kg==
```
christothes, line 4: example: sanitized secret recorded variable

sdk/tables/aztables/recordings/tableClientLiveTests(cosmos)/TestAddEntity-variables.yaml
```diff
@@ -0,0 +1,5 @@
...
+randomSeed: "1617580856"
```
christothes, line 5: recorded random seed, for deterministic random Id generation.

sdk/tables/aztables/go.sum
```diff
@@ -19,5 +20,9 @@ golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/text v0.3.3 h1:cokOdA+Jmi5PJGXLlLllQSgYigAEfHXJAERHVMaCc2k=
 golang.org/x/text v0.3.3/go.mod h1:5Zoc/QRtKVWzQhOtBMvqHzDpF6irO9z98xDceosuGiQ=
 golang.org/x/tools v0.0.0-20180917221912-90fa682c2a6e/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c h1:Hei/4ADfdWqJk1ZMxUNpqntNwaWcugrBjAiHlqqRiVk=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c/go.mod h1:JHkPIbrfpd72SG/EVd6muEfDQjcINNoR0C8j2r3qZ4Q=
+gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v2 v2.4.0 h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
+gopkg.in/yaml.v2 v2.4.0/go.mod h1:RDklbk79AGWmwhnvt/jBztapEOGDOx6ZbXqjP6csGnQ=
```
catalinaperalta, line 28: You can run `go mod tidy` to clean up the go.mod and go.sum entries

sdk/internal/testframework/testcontext.go
```diff
@@ -0,0 +1,37 @@
...
+	fail Failer
+	log  Logger
+	name string
+}
+
+type Failer func(string)
+type Logger func(string)
+type Name func() string
+
+func NewTestContext(f Failer, l Logger, name Name) TestContext {
```
catalinaperalta, line 24: Nit: `name` to `n`

This doesnt really matter but I'd say for consistency it's better?

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,371 @@
...
+	return nil
+}
+
+// Gets an environment variable by name and panics if it is not found
+func GetRequiredEnv(name string) string {
+	env, ok := os.LookupEnv(name)
+	if ok {
+		return env
+	} else {
+		panic(envNotExistsError(name))
```
catalinaperalta, line 187: Panicking is a bit controversial. I know it's a feature of the language, but most people prefer not to panic. Although this will be an internal module, other Go SDK devs may use this so I would suggest following the other pattern and returning an error. So the return signature would look like `(string, error)`. 

You can also ask Joel and Jeff for their opinions on this.

christothes: I think I should just call the testContext failed here instead. That would just fail the test.

sdk/tables/aztables/tableClient_test.go
```diff
@@ -4,9 +4,126 @@
...
+
+// setup the test environment
+func (s *tableClientLiveTests) BeforeTest(suite string, test string) {
+	recordedTestSetup(s.T(), s.T().Name(), s.endpointType, s.mode)
+}
+
+// teardown the test context
+func (s *tableClientLiveTests) AfterTest(suite string, test string) {
+	recordedTestTeardown(s.T().Name())
+}
```
christothes, line 115: Example of how a test would leverage the recorded test framework (1/2)

sdk/tables/aztables/zt_tableRecordedTests.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.Nil(err)
+	clientsMap[testName] = &testContext{client: client, recording: recording}
+}
+
+func recordedTestTeardown(key string) {
+	context, ok := clientsMap[key]
+	if ok {
+		context.recording.Stop()
+	}
+}
```
christothes, line 91: example of how a test would leverage the recorded test framework (2/2)

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestCreateTable.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - azsdk-go-tables/<version> azsdk-go-tables/<version> azcore/v0.13.4 (go1.16.2;
+        Windows_NT)
+      X-Ms-Date:
+      - Sun, 04 Apr 2021 21:54:53 GMT
+      X-Ms-Version:
+      - "2019-02-02"
+    url: https://chrisstablesprim.table.cosmos.azure.com/Tables
+    method: POST
+  response:
+    body: '{"TableName":"gotablecwmvbx3xi23m7","odata.metadata":"https://chrisstablesprim.table.cosmos.azure.com/$metadata#Tables/@Element"}'
```
catalinaperalta, line 28: How do we sanitize the body? If we want to clean the url in odata.metadata?

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestQueryTable-variables.yaml
```diff
@@ -0,0 +1,5 @@
+---
+COSMOS_TABLES_ENDPOINT_SUFFIX: cosmos.azure.com
+TABLES_COSMOS_ACCOUNT_NAME: chrisstablesprim
+TABLES_PRIMARY_COSMOS_ACCOUNT_KEY: Kg==
+randomSeed: "1617573294"
```
catalinaperalta, line 5: So these are also sanitized when calling the default sanitizer?

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+	recPath, varPath := getFilePaths(c.Name())
+	rec, err := recorder.NewAsMode(recPath, modeMap[mode], nil)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// If the mode is set in the environment, let that override the requested mode
+	// This is to enable support for nightly live test pipelines
+	envMode := GetOptionalEnv(ModeEnvironmentVariableName, string(mode))
```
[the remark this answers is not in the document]
catalinaperalta: I saw that this is in the test setup, I see why it's exported.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+	recPath := "recordings/" + name
+	varPath := fmt.Sprintf("%s-variables.yaml", recPath)
+	return recPath, varPath
+}
+
+// Calls os.Create on the VariablesFile and creates it if it or the path does not exist
+// Callers must call Close on the result
+func (r *Recording) createVariablesFileIfNotExists() (*os.File, error) {
+	f, err := os.Create(r.VariablesFile)
+	if err != nil {
```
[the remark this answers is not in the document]
christothes: Possibly - although for any other error than file not found, I think we'd still ultimately fail to create it.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// Gets a recorded variable. If the variable is not found we return an error
+// variableType is optional and defaults to Default
+func (r *Recording) GetRecordedVariable(name string, variableType ...VariableType) (*string, error) {
+	var err error
+	result, ok := r.previousSessionVariables[name]
+	if !ok || r.Mode == Live {
+
+		result, err = GetRequiredEnv(name)
```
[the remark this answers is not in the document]
christothes: I considered not exporting it, but didn't initially because I thought it could be useful. But I don't feel strongly either way.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+		return nil, errors.New(error)
+	}
+	return resp, err
+}
+
+// Stops the recording and saves them, including any captured variables, to disk
+func (r *Recording) Stop() error {
+
+	r.recorder.Stop()
+	if r.Mode == Live {
```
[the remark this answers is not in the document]
christothes: It's intended to work as if the recording framework is not there. For example, to just experiment with tests against the live service without attempting to save recordings or read existing recordings.

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestCreateTable.yaml
christothes: Good catch - I need to add a `santizeBody` func to `recording_sanitizer`

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestQueryTable-variables.yaml
christothes: These are sanitized only if `GetRecordedVariable` is called with a `VariableType` argument that indicates it is a secret.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+		} else {
+			r.variables[name] = applyVariableOptions(result, variableType...)
+		}
+	}
+	return result, err
+}
+
+// Gets a recorded variable with a fallback default value
+// variableType is optional and defaults to Default
+func (r *Recording) GetOptionalRecordedVariable(name string, defaultValue string, variableType ...VariableType) string {
```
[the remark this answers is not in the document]
christothes: I think I was just trying to be clever here. Making it a pointer would also work, but I liked the idea of callers not having to specify it unless it was a secret value. This is probably not very idiomatic though.

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
catalinaperalta: My only thought about this is that we should return the first error that's return to not bury/hide it. I'll leave it up to you.

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
catalinaperalta: I think it's easier to have things unexported and go exporting them later as necessary. Although this is an internal package so it doesnt really matter as much, so whatever you think is better.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,373 @@
...
+// If variableType is not provided or Default, return result
+// If variableType is Secret_String, return SanitizedValue
+// If variableType isSecret_Base64String return SanitizedBase64Value
+func applyVariableOptions(val *string, variableType ...VariableType) *string {
+	var ret string
+	switch len(variableType) {
+	case 0:
+		return val
+	default:
+		switch vt := variableType[0]; vt {
```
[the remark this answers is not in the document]
christothes: Yes, it's just 1 or 0. This was just my attempt at being clever for an optional argument. I'd either doc it or change it to a * as mentioned in the other comment.

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
catalinaperalta: The variadic config is idiomatic, I just dont think it is in this case. 🤔 This popped out to me, because when I saw the variadic config I thought multiple transforms would be applied to the variable type. If you think it'll always remain a none or 1 choice then I'd say you could just make it a pointer BUT if at any point there could be more then this would be fine. We could discuss this in the meeting next week.

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
catalinaperalta: Ah I see that's a great feature!

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
christothes: As I've thought about it more, it probably makes mores sense a a required argument anyway, since it is important to be intentional for sanitization.

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestQueryTable-variables.yaml
catalinaperalta: Ah got it, thanks for the info :)

sdk/internal/testframework/recording.go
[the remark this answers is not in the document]
catalinaperalta: Yeah I think a pointer might be better 👍

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,414 @@
...
+
+	// wire up the sanitizer
+	DefaultSanitizer(rec)
+
+	return recording, err
+}
+
+// GetRecordedVariable returns a recorded variable. If the variable is not found we return an error
+// variableType determines how the recorded variable will be saved. Default indicates that the value should be saved without any sanitation.
+func (r *Recording) GetRecordedVariable(name string, variableType VariableType) (*string, error) {
```
jhendrixMSFT, line 114: The convention in Go is to return types by value by default, and in the case of an error you'd return its zero-value ("" here).  The exceptions for this are as follows.

- the object is intended to be singular in instance and thus shared, e.g. an `*http.Client`
- the object is mutable (i.e. has receivers that pass by reference)
- there is a semantic difference between the zero-value and `nil`
- the object is large, for some definition of large (there's some debate on this one)

Since this is a primitive type, only item 3 applies.  If there's no semantic difference here then you'll want to remove the `*`.  (aside: coming from .NET I know this is weird as strings are reference types, I struggled with this early on)

christothes: Thanks, makes sense. I tend to want to err on the side of moving pointers around rather than copying values, but the guidelines are very helpful!

sdk/internal/testframework/recording.go
```diff
@@ -111,19 +111,19 @@ func NewRecording(c TestContext, mode RecordMode) (*Recording, error) {
...
 
 		result, err = getRequiredEnv(name)
 		if err != nil {
 			r.c.Fail(err.Error())
-			return nil, err
+			return "", err
 		}
 		r.variables[name] = applyVariableOptions(result, variableType)
 	}
-	return result, err
```
catalinaperalta: Could this panic? (If result is nil)

sdk/tables/aztables/zc_table_pagers.go
```diff
@@ -29,6 +30,49 @@ type TableEntityQueryResponsePager interface {
 	Err() error
 }
 
+type StructEntityQueryResponsePager interface {
+	NextPage(context.Context) bool
+	PageResponse() StructQueryResponseResponse
+	Err() error
+}
+
+type StructQueryResponseResponse struct {
```
catalinaperalta, line 39: Why does this start with `Struct`?

catalinaperalta: Oh this isnt related to the testing framwork right?

christothes: right - It's some experimentation I was doing with customizing table query responses to serialize to customer provide structs rather than the default `map[string]interface{}`

catalinaperalta (approved):
Test framework looks good to me, will test with azidentity from your branch!

sdk/tables/aztables/zc_table_pagers.go
catalinaperalta: We should probably talk about naming for this later in the public surface area review.

catalinaperalta: Didnt take a hard look at it now.

PR #14492, opened 2021-04-06
Regenerate Blob SDK backend and (mostly) fix tests

adreed-msft:
This PR is not _entirely_ finished yet, hence why it is marked as a draft. A couple of issues regarding the generator need to be fixed, but some manual changes have been made to generated code to fix compiling errors.


RickWinter:
@adreed-msft Can this stale PR be closed?

PR #14494, opened 2021-04-06
Migrate license to MIT

qt-luigi:
Update content to MIT at a section of License on README.


ArcturusZhang (approved):
Hi @qt-luigi thank you for this PR! LGTM 👍

PR #14549, opened 2021-04-19
Track2: azcore log remove some duplicate code

magodo:
The PR removes some duplicate code in azcore log.go:

- Remove the check on `l.lst` in `Logger.Writef` as it is checked in `Logger.Should` already
- Reuse `Logger.Writef` in `Logger.Write` to DRY.

### DIFF
--- a/sdk/azcore/log.go
+++ b/sdk/azcore/log.go
@@ -77,16 +77,13 @@ func (l *Logger) Should(cls LogClassification) bool {
 // Write invokes the underlying Listener with the specified classification and message.
 // If the classification shouldn't be logged or there is no listener then Write does nothing.
 func (l *Logger) Write(cls LogClassification, message string) {
-	if l.lst == nil || !l.Should(cls) {
-		return
-	}
-	l.lst(cls, message)
+	l.Writef(cls, "%s", message)
 }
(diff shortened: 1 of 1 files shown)


sdk/azcore/log.go
```diff
@@ -77,16 +77,13 @@ func (l *Logger) Should(cls LogClassification) bool {
 // Write invokes the underlying Listener with the specified classification and message.
 // If the classification shouldn't be logged or there is no listener then Write does nothing.
 func (l *Logger) Write(cls LogClassification, message string) {
-	if l.lst == nil || !l.Should(cls) {
-		return
-	}
-	l.lst(cls, message)
+	l.Writef(cls, "%s", message)
```
jhendrixMSFT, line 80: The call to `l.lst()` in this case was intentional, to avoid interpreting any `%` characters as format specifiers (URL-encoded query params will contain them).

jhendrixMSFT (changes requested):

sdk/azcore/log.go
magodo: I've revert this.

jhendrixMSFT (approved):

PR #14572, opened 2021-04-22
add helper method for service error unmarshal failures

catalinaperalta:
 
### DIFF
--- a/sdk/azcore/response.go
+++ b/sdk/azcore/response.go
@@ -136,7 +136,10 @@ func (r *Response) UnmarshalError(unmarshalErr error) error {
 	if err != nil {
 		return err
 	}
-	return fmt.Errorf("raw service error: %s\n%w", payload, unmarshalErr)
+	if len(payload) == 0 {
+		return fmt.Errorf("empty error body\n%s", unmarshalErr)
+	}
+	return fmt.Errorf("raw service error: %s\n%s", payload, unmarshalErr)
 }
 
 // Drain reads the response body to completion then closes it.  The bytes read are discarded.
--- a/sdk/azcore/response.go
+++ b/sdk/azcore/response.go
@@ -129,6 +129,16 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
 	return err
 }
 
+// UnmarshalError will unmarshal the received service error payload along with the corresponding unmarshal error.
+// NOTE: this method is only meant to be called when handling a service error.
+func (r *Response) UnmarshalError(unmarshalErr error) error {
+	payload, err := r.payload()
+	if err != nil {
+		return err
+	}
+	return fmt.Errorf("raw service error: %s\n%w", payload, unmarshalErr)
+}
+
 // Drain reads the response body to completion then closes it.  The bytes read are discarded.
 func (r *Response) Drain() {
 	if r != nil && r.Body != nil {
--- a/sdk/azcore/response.go
+++ b/sdk/azcore/response.go
@@ -129,8 +129,7 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
 	return err
 }
 
-// UnmarshalError will combine the service error payload along with the corresponding unmarshal error into a new error
-// that includes information from both errors.
+// UnmarshalError will combine information from the service error payload along with the corresponding unmarshal error into a new error.
 // NOTE: this method is only meant to be called when handling a service error.
 func (r *Response) UnmarshalError(unmarshalErr error) error {
 	payload, err := r.payload()
--- a/sdk/azcore/response.go
+++ b/sdk/azcore/response.go
@@ -129,7 +129,8 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
 	return err
 }
 
-// UnmarshalError will unmarshal the received service error payload along with the corresponding unmarshal error.
+// UnmarshalError will combine the service error payload along with the corresponding unmarshal error into a new error
+// that includes information from both errors.
 // NOTE: this method is only meant to be called when handling a service error.
 func (r *Response) UnmarshalError(unmarshalErr error) error {
 	payload, err := r.payload()
--- a/sdk/azcore/response_test.go
+++ b/sdk/azcore/response_test.go
@@ -256,3 +256,30 @@ test error`
 		t.Fatal("unexpected value")
 	}
 }
+
+func TestResponseUnmarshalErrorEmptyResponseBody(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	srv.SetResponse(mock.WithStatusCode(400))
+	pl := NewPipeline(srv)
+	req, err := NewRequest(context.Background(), http.MethodGet, srv.URL())
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	resp, err := pl.Do(req)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


catalinaperalta:
I forced an unmarshal error to get the result of calling response.UnmarshalError(). Real world example: 
```
raw service error: {"code":"LibraryDoesNotExistInWorkspace","message":"The LibraryArtifact test does not exist."}
unmarshalling type azartifacts.CloudErrorAutoGenerated: json: Unmarshal(non-pointer azartifacts.CloudErrorAutoGenerated)
```

catalinaperalta:
generated error handler code should look like:
```
// getHandleError handles the Get error response.
func (client *LibraryClient) getHandleError(resp *azcore.Response) error {
	var err CloudErrorAutoGenerated
	if err := resp.UnmarshalAsJSON(err); err != nil {
		return resp.UnmarshalError(err)
	}
	return azcore.NewResponseError(&err, resp.Response)
}
```

sdk/azcore/response_test.go
```diff
@@ -7,6 +7,7 @@ package azcore
```
ArcturusZhang, line 7: I am curious that should we unify the package name of the test packages? I am seeing one of the test package `azcore_test` here: https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/azcore/example_test.go

catalinaperalta: Good point! Makes sense to make these also azcore_test. I can update this in a separate PR though.

jhendrixMSFT: So I believe the name `azcore_test` is intentional in this case.  When you use the package name `<package>_test` it means you don't have access to unexported content within `<package>`.  This is useful when writing examples for two reasons.

- the examples will have to use the fully qualified name, e.g. `azcore.Poller` which is nice for copy/paste
- it ensures you don't have access to unexported content

catalinaperalta: Gotcha, will leave as is.

catalinaperalta:
> generated error handler code should look like:
> 
> ```
> // getHandleError handles the Get error response.
> func (client *LibraryClient) getHandleError(resp *azcore.Response) error {
> 	var err CloudErrorAutoGenerated
> 	if err := resp.UnmarshalAsJSON(err); err != nil {
> 		return azcore.NewResponseError(resp.UnmarshalError(err), resp.Response)
> 	}
> 	return azcore.NewResponseError(&err, resp.Response)
> }
> ```
> 
> EDIT: actually changed this to also return the response error type, we can discuss this on the generator side

@jhendrixMSFT let me know if you think this is a good idea, I will wait to open the autorest PR until I hear your opinion on this.

sdk/azcore/response.go
```diff
@@ -129,6 +129,16 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
...
 }
 
+// UnmarshalError will unmarshal the received service error payload along with the corresponding unmarshal error.
+// NOTE: this method is only meant to be called when handling a service error.
+func (r *Response) UnmarshalError(unmarshalErr error) error {
+	payload, err := r.payload()
+	if err != nil {
+		return err
+	}
+	return fmt.Errorf("raw service error: %s\n%w", payload, unmarshalErr)
```
jhendrixMSFT, line 139: I wouldn't wrap `unmarshalErr` here since we don't need/want callers to test for explicit unmarshall errors.

catalinaperalta: Didnt we also want include the unmarshal error for ourselves?

sdk/azcore/response.go
```diff
@@ -129,6 +129,16 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
 	return err
 }
 
+// UnmarshalError will unmarshal the received service error payload along with the corresponding unmarshal error.
+// NOTE: this method is only meant to be called when handling a service error.
+func (r *Response) UnmarshalError(unmarshalErr error) error {
+	payload, err := r.payload()
```
jhendrixMSFT, line 135: I believe it's possible (although unlikely) that `payload` could be empty.  If this is the case, we have two options.

- add some default message "missing body" or whatever
- just return `unmarshalErr`

I _think_ that the JSON/XML unmarshallers will fail with a message indicating the first case ("unexpected end of blah blah").  If that's the case then let's do the second one.

sdk/azcore/response.go
jhendrixMSFT: We want to include it but not wrap it, so just a `%s` should do the trick.

jhendrixMSFT:
So this will be used like so right?

```go
func (client *LROsClient) delete202NoRetry204HandleError(resp *azcore.Response) error {
	var err CloudError
	if err := resp.UnmarshalAsJSON(&err); err != nil {
		return azcore.UnmarshalError(err)
	}
	return azcore.NewResponseError(&err, resp.Response)
}
```

sdk/azcore/response.go
catalinaperalta: I will check this and simply return the unmarshalErr if the other unmarshallers have a good message.

catalinaperalta:
> So this will be used like so right?
> 
> ```go
> func (client *LROsClient) delete202NoRetry204HandleError(resp *azcore.Response) error {
> 	var err CloudError
> 	if err := resp.UnmarshalAsJSON(&err); err != nil {
> 		return azcore.UnmarshalError(err)
> 	}
> 	return azcore.NewResponseError(&err, resp.Response)
> }
> ```

Well I initially had this, but then thought that we could wrap the error from azcore.UnmarshalError() in an azcore.NewResponseError() so that users can access the service error themselves and check for status codes, etc. What do you think?

jhendrixMSFT:
Yes you're correct, so it would be `returnazcore.NewResponseError( azcore.UnmarshalError(err), resp.Response)`

sdk/azcore/response.go
catalinaperalta: It doesnt seem that the other unmarshallers add a message if the payload is empty here's one of them as reference:
```
	payload, err := r.payload()
	if err != nil {
		return err
	}
	// TODO: verify early exit is correct
	if len(payload) == 0 {
		return nil
	}
	err = r.removeBOM()
	if err != nil {
		return err
	}
	err = json.Unmarshal(payload, v)
	if err != nil {
		err = fmt.Errorf("unmarshalling type %T: %s", v, err)
	}
	return err
}
```

catalinaperalta: I think it's still a good idea to just include the unmarshalErr even if the payload is empty, since the err we're sending is populated with the error from json.Ummarshal()

sdk/azcore/response.go
catalinaperalta: Done

sdk/azcore/response.go
catalinaperalta: I added empty error body to the message. Done?

jhendrixMSFT: Oh yes I forgot we have that empty body check.  We should follow up on that separately to see if that's really what we should be doing (I forget why I added that, might have been early on prototyping).

sdk/azcore/response.go
```diff
@@ -129,6 +129,19 @@ func (r *Response) UnmarshalAsXML(v interface{}) error {
 	return err
 }
 
+// UnmarshalError will unmarshal the received service error payload along with the corresponding unmarshal error.
```
jhendrixMSFT, line 132: Nit, I would reword this slightly this doesn't unmarshall stuff.

jhendrixMSFT (approved):
Small rewording of doc comment.

PR #14580, opened 2021-04-22
Support for basic LRO polling

jhendrixMSFT:
Added type LROPoller and supporting types for basic polling on long
running operations.
Fixed content type detection bug in logging.

### DIFF
--- a/sdk/azcore/poller.go
+++ b/sdk/azcore/poller.go
@@ -18,11 +18,6 @@ import (
 	"time"
 )
 
-const (
-	headerOperationLocation = "Operation-Location"
-	headerLocation          = "Location"
-)
-
 // ErrorUnmarshaller is the func to invoke when the endpoint returns an error response that requires unmarshalling.
 type ErrorUnmarshaller func(*Response) error
 
--- a/sdk/azcore/poller.go
+++ b/sdk/azcore/poller.go
@@ -224,7 +224,7 @@ func (l *LROPoller) PollUntilDone(ctx context.Context, freq time.Duration, respT
 			return nil, err
 		}
 		if l.Done() {
-			logPollUntilDoneExit(l.lro.State())
+			logPollUntilDoneExit(l.lro.Status())
 			if !l.lro.Succeeded() {
 				return nil, l.eu(&Response{resp})
 			}
@@ -252,7 +252,7 @@ type lroPoller interface {
 	Update(resp *Response) error
 	FinalGetURL(resp *Response) (string, error)
 	URL() string
-	State() string
+	Status() string
 	Succeeded() bool
 }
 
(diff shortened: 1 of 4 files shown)


sdk/azcore/poller.go
```diff
@@ -0,0 +1,453 @@
...
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	headerOperationLocation = "Operation-Location"
+	headerLocation          = "Location"
```
catalinaperalta, line 23: Although I see these are unexported, just asking if these should be organized along with the other headers in the headers.go file? Totally fine to leave it here though since they're just used for the pollers at the moment.

sdk/azcore/poller.go
```diff
@@ -0,0 +1,453 @@
...
+)
+
+// ErrorUnmarshaller is the func to invoke when the endpoint returns an error response that requires unmarshalling.
+type ErrorUnmarshaller func(*Response) error
+
+// NewLROPoller creates an LROPoller based on the provided initial response.
+// NOTE: this is only meant for internal use in generated code.
+func NewLROPoller(pollerType string, resp *Response, pl Pipeline, eu ErrorUnmarshaller) (*LROPoller, error) {
```
catalinaperalta, line 31: Should we document what pollerType should be??

sdk/azcore/poller.go
jhendrixMSFT: Mostly I think of headers.go as the location for all of the ones we export.  However, perhaps we should place the unexported ones here too, to reduce the potential to duplicate them.

sdk/azcore/poller.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (p *locPoller) Done() bool {
+	// a 202 means the operation is still in progress
+	// zero-value indicates the poller was rehydrated from a token
+	return p.status > 0 && p.status != http.StatusAccepted
+}
+
+func (p *locPoller) Succeeded() bool {
+	// any 2xx status code indicates success
+	return p.status >= 200 && p.status < 300
```
catalinaperalta, line 364: This is interesting, we dont have this in the ARM pollers do we??

sdk/azcore/poller.go
catalinaperalta: I like that idea...simply because I could see duplicating headers if not placed in that file. Of course these are unexported so up to you, but I do think it's better

catalinaperalta (approved):
Love this!

sdk/azcore/poller.go
jhendrixMSFT: I don't think ARM has such simple polling.  The logic for locPoller is that it will continue to poll on a 202 returned from the polling endpoint.  A non-202 response that's a 2xx indicates the operation succeeded.  All other status codes indicate it failed.

catalinaperalta: Oh cool. Good to know, thanks!

sdk/azcore/poller.go
jhendrixMSFT: I added a doc comment.  I also renamed the param to `pollerID` as I felt that was a more accurate description.

PR #14600, opened 2021-04-27
Testframework: various sanitizer fixes

catalinaperalta:
- Separate sanitizer for request bodies and response bodies (NOTE: I'm open to having just one body sanitizer with a field that specifies if it's meant to be used for the request, response or both. Let me know!)
- Assign default sanitizer to the recording sanitizer
- Export sanitizer on the recording to allow sanitizer customization

### DIFF
--- a/sdk/internal/testframework/recording.go
+++ b/sdk/internal/testframework/recording.go
@@ -33,7 +33,7 @@ type Recording struct {
 	recorder                 *recorder.Recorder
 	src                      rand.Source
 	now                      *time.Time
-	sanitizer                *RecordingSanitizer
+	Sanitizer                *RecordingSanitizer
 	c                        TestContext
 }
 
@@ -104,7 +104,7 @@ func NewRecording(c TestContext, mode RecordMode) (*Recording, error) {
 	rec.SetMatcher(recording.matchRequest)
 
 	// wire up the sanitizer
-	DefaultSanitizer(rec)
+	recording.Sanitizer = DefaultSanitizer(rec)
 
 	return recording, err
 }


catalinaperalta:
The request matcher is still breaking azidentity. Something is broken when comparing the headers of the recording and the request. I can try to look into this later this week.

christothes:
> The request matcher is still breaking with azidentity recordings. Something is broken when comparing the headers of the recording and the request. I can try to look into this in a few days.

My guess is that there is a header we need to ignore.

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -41,9 +42,14 @@ func (s *RecordingSanitizer) AddSanitizedHeaders(headers ...string) {
 	}
 }
 
-// AddBodysanitizer configures the supplied StringSanitizer to sanitize recording request and response bodies
-func (s *RecordingSanitizer) AddBodysanitizer(sanitizer StringSanitizer) {
-	s.bodySanitizer = sanitizer
+// AddBodysanitizerReq configures the supplied StringSanitizer to sanitize recording request body
+func (s *RecordingSanitizer) AddBodysanitizerReq(sanitizer StringSanitizer) {
```
christothes, line 46: Do you have a specific scenario in mind that requires different sanitization for requests and responses? If not, I'd stick with a single one for now.

catalinaperalta: For azidentity, the request for ClientSecretCredential for example includes client_id and client secret in the request body, meanwhile the response body returns the access token which needs to be sanitized, but I cant know what the token is going to be beforehand so I simply replaced the string to sanitize it...although I suppose I could combine all of this logic into one sanitizer 🤔 BUT the sanitizer gets a bit larger though and has to include logic for two separate bodies...so I thought it was cleaner to separate into two separate sanitizers. I foresee the request sanitizer growing to sanitize more elements such as usernames and passwords, etc. Meanwhile the response sanitizer could remain relatively smaller and simpler. Let me know what you think.

christothes: Are you concerned about perf, or complexity of the sanitization code?

I would expect the logic to be something like searching for indexes of property names we want to sanitize and parsing out the value to replace whatever it is with the sanitized value. So a generalized body sanitizer would look like:
if body contains client_secret , replace the value
if body contains bearerToken, replace the value, 
etc..

For requests or responses where we don't find the token we want to sanitize, there is no extra processing there.

catalinaperalta: I was mostly concerned with the complexity of the sanitization code, but I can revert these changes and leave them in one sanitizer for now. We can talk about it later if it gets too big in any case. 👍

christothes: Cool - yeah I think we should start simple, and address if it becomes unwieldy. I think having 2 may introduce more risk of using the wrong one and leaving something unsanitized.

christothes (approved):

Issue #14387, opened 2021-03-04
Proposal: Convert this repo to a Go module


jhendrixMSFT:
Please add a 👍 reaction if you support this proposal. If you do not, please leave a comment describing why.

Now that Go modules have solidified and are on by default in Go 1.16 we are considering how we migrate this repo to officially support them (this is for our track 1 content; track 2 already supports modules).

In keeping with our current release strategy and version, the natural option would be to place a `go.mod` file in the root of the repo.  However, given that we've already had dozens of major releases (`v52.3.0` at the time of this writing), it would mean that the next major version would introduce [semantic import versioning](https://github.com/golang/go/wiki/Modules#semantic-import-versioning).

In `go.mod`:
```
module github.com/Azure/azure-sdk-for-go/v53
```

This would require consumers to update their import statements starting with `v53` and beyond.

```
import github.com/Azure/azure-sdk-for-go/v53/services/foo/bar
```

While this "plays by the rules" it introduces a burden on consumers and we'd like feedback on the impact.

jhendrixMSFT:
Unfortunately this places a significant burden on consumers that use a large number of packages from this SDK.  This was mentioned [here](https://github.com/Azure/azure-sdk-for-go/pull/14626#discussion_r623995504) on the negative impact it would have on Terraform.

marchesir:
well at some point go may force the change, so not easy on anybody. i for 1 have an issue azure-sdk-for-go is required by terratest but where i work we use artifactory and as there is no go.mod it makes downloading hard as artifactory assumes .mod.

jhendrixMSFT:
I tend to agree.  It's a bitter pill but is likely inevitable.  In addition, I recall there was talk about tooling to simplify updating a major module version.

CC @tombuildsstuff

bcmills:
Regarding upgrade automation, I am aware of at least https://github.com/golang/go/issues/32816 and https://github.com/golang/go/issues/32014, as well as a couple of existing tools (https://github.com/icholy/gomajor and https://github.com/marwan-at-work/mod, but others may exist too).

bcmills:
I'm also planning to address module aliasing and renaming (golang/go#26904) for Go 1.18 (due out EoY this year), and it's possible that that will include a way for an existing module to leave a “fowarding address”.

(More design work is needed, but migrating from one module path to another — including from an unversioned path to a versioned one — is an important use-case that we're actively thinking about.)

jhendrixMSFT:
[...]

PR #14626, opened 2021-04-30
Release v54.0.0

ArcturusZhang:
 

go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/Azure/azure-sdk-for-go/v54
+
+go 1.16
```
jhendrixMSFT, line 3: This will prevent usage on older versions of Go.  For track 2 we've been using 1.13 as the min version.  Should do the same here.

go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/Azure/azure-sdk-for-go/v54
```
tombuildsstuff, line 1: presumably (to not break all existing consumers of the SDK) this wants to be:

```suggestion
module github.com/Azure/azure-sdk-for-go
```

else this'll break all existing consumers of the SDK?

jhendrixMSFT: It _must_ have a version suffix as we're at `v54` of the SDK.  This is what I was pointing out in https://github.com/Azure/azure-sdk-for-go/issues/14387, that migrating the SDK to a module has significant impact on our consumers.

tombuildsstuff: Unfortunately from our side that'd prevent us from upgrading to v54 and above of the SDK just due to the sheer number of things it'd conflict with. Whilst I can appreciate this is the "right" way of doing things going forward, unfortunately Go's module versioning is particularly painful for larger (and especially open-source) projects.

Since this SDK's gone through 53 major releases without it in Track1, to be honest I don't see the value of introducing such a large breaking change into Track1 - particularly when Track2 is unsuitable for some customers (and won't be for the foreseeable future).

From our side I guess we could work around this by republishing this under a temporary repository for the moment and keeping that up to date whilst we're still on Track1 - which would give us slightly more admin overhead (or more likely, a CI job) but would allow us to not conflict with _every_ pull request.

jhendrixMSFT: Let my circle back with my team about how we want to proceed here.

jhendrixMSFT: This has been removed for the `v54` release.

jhendrixMSFT (approved):

PR #14629, opened 2021-05-03
Add support for null map and slice

jhendrixMSFT:
Also export the `payload()` method on `Response`

### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -154,13 +154,19 @@ var nullables map[reflect.Type]interface{} = map[reflect.Type]interface{}{}
 // This is typically used in JSON-MERGE-PATCH operations to delete a value.
 func NullValue(v interface{}) interface{} {
 	t := reflect.TypeOf(v)
-	if t.Kind() != reflect.Ptr {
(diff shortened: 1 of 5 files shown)


sdk/azcore/core.go
```diff
@@ -173,13 +179,15 @@ func NullValue(v interface{}) interface{} {
 func IsNullValue(v interface{}) bool {
 	// see if our map has a sentinel object for this *T
 	t := reflect.TypeOf(v)
-	if t.Kind() != reflect.Ptr {
+	if k := t.Kind(); k != reflect.Ptr && k != reflect.Slice && k != reflect.Map {
```
catalinaperalta, line 182: We cant have nil slices?

jhendrixMSFT: We can.  This check is saying "if v isn't a pointer-to-type, and it's not a slice or a map, then it can't be nil so return false".

catalinaperalta: Oh, yeah definitely misread this. Thanks!

catalinaperalta (approved):

PR #14660, opened 2021-05-13
[SDK Automation] generator now implements its own logic to retrieve the tags from batch in readme.go.md

ArcturusZhang:
This PR contains the following changes:
1. introduces a go wrapper of the autorest execution
2. introduces a new function to retrieve the tags in batch section in `readme.go.md` to have more precise validation of the readme configurations as well as it gives the tool the functionality to reuse the options in package metadata to regenerate the package if the package has been generated before.


tools/generator/autorest/autorest.go
```diff
@@ -130,14 +130,14 @@ func (g *Generator) StderrPipe() (io.ReadCloser, error) {
...
+	Options model.Options
+	Message string
 }
 
 // Error ...
 func (e *GenerateError) Error() string {
-	arguments := make([]string, len(e.Arguments))
-	for i, o := range e.Arguments {
+	arguments := make([]string, len(e.Options.Arguments()))
+	for i, o := range e.Options.Arguments() {
```
chenjianfei2017, line 140: Is it better to use arguments defined above?
for i, o := range arguments {

ArcturusZhang: We are iterating over the `[]model.Option` here and accumulate them to a string slice. 
I allocate the slice `arguments` with a specific length for performance consideration (which matters very little)

chenjianfei2017:
LGTM

chenjianfei2017 (approved):

PR #14667, opened 2021-05-13
Enable user assigned identity for IMDS

catalinaperalta:


catalinaperalta:
checked using a user assigned identity on an Azure VM, it was successful :)

jhendrixMSFT (approved):

PR #14676, opened 2021-05-18
Progress on aztables

christothes:
Creating this PR for some early feedback on the API shape


sdk/tables/aztables/tableClient.go
```diff
@@ -7,34 +7,69 @@ import (
...
+func (t *TableClient) Query(queryOptions QueryOptions) TableEntityQueryResponsePager {
+	return &tableEntityQueryResponsePager{tableClient: t, queryOptions: &queryOptions, tableQueryOptions: &TableQueryEntitiesOptions{}}
 }
 
-// Create
-func (t TableClient) Create(ctx context.Context, name string) (TableResponseResponse, error) {
-	resp, err := t.client.Create(ctx, TableProperties{&name}, nil, nil)
-	if resp == nil {
-		return TableResponseResponse{}, err
+func (t *TableClient) QueryAsStruct(opt QueryOptions, s FromMapper) StructEntityQueryResponsePager {
```
christothes, line 46: The response type for this operation is mostly just a draft, but I'm curious about any feedback related to this idea for returning query results as a well defined interface struct rather than the default `map[string]interface{}`. This allows developers to use "strongly typed" models rather than the map.

sdk/tables/aztables/tableClient.go
```diff
@@ -7,34 +7,69 @@ import (
...
-		return resp.(TableResponseResponse), err
+		return nil, convertErr(err)
 	}
 }
 
-// Delete
-func (t TableClient) Delete(ctx context.Context, name string) (TableDeleteResponse, error) {
-	return t.client.Delete(ctx, name, nil)
+// AddEntity creates an entity from an arbitrary struct value.
+func (t *TableClient) AddEntity(ctx context.Context, entity interface{}) (*TableInsertEntityResponse, *runtime.ResponseError) {
```
christothes, line 63: This func is the analog for the QueryAsStruct method above. It allows adding of a "strongly typed" table entity struct rather than the default map shape.

sdk/internal/testframework/recording_sanitizer_test.go
```diff
@@ -154,4 +154,4 @@ func NewMockRoundTripper(server *mock.Server) *mockRoundTripper {
 
 func (m *mockRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	return m.server.Do(req)
-}
```
catalinaperalta, line 157: Nit: newline missing

sdk/tables/aztables/CHANGELOG.md
```diff
@@ -1,3 +1,3 @@
 # Release History
 
-## v0.1.0 (Unreleased)
+## 0.1.0 (Unreleased)
```
catalinaperalta, line 3: I think you should be leaving the v here to align with all of our other libs, we've left it this way for the ARM libs and also Joel has a PR open for releasing synapse data-plane that also includes the v.

sdk/tables/aztables/recordings/TestServiceClient_Cosmos/TestCreateTable.yaml
```diff
@@ -0,0 +1,75 @@
...
+      Authorization:
+      - sanitized
+      Content-Length:
+      - "36"
+      Content-Type:
+      - application/json
+      Dataserviceversion:
+      - "3.0"
+      User-Agent:
+      - azsdk-go-tables/<version> azsdk-go-tables/<version> azcore/v0.13.4 (go1.16.2;
```
catalinaperalta, line 19: The telemetry string in the generated code needs to be updated to replace `<version>` with `v0.1.0`.
Also just noticed shouldn't it have generated with `azsdk-go-aztables`? instead of just ending with `tables`

sdk/tables/aztables/tableClient.go
```diff
@@ -7,34 +7,69 @@ import (
 	"context"
```
catalinaperalta, line 7: FYI since this is for the UX study about how Go-ish our SDKs are 😝...the file names need to be fixed to be snake cased

sdk/tables/aztables/tableClient.go
```diff
@@ -7,34 +7,69 @@ import (
...
+	client  *tableClient
+	service *TableServiceClient
+	cred    SharedKeyCredential
+	name    string
 }
 
 // NewTableClient creates a TableClient object using the specified URL and request policy pipeline.
-func NewTableClient(serviceURL string, cred azcore.Credential, options *TableClientOptions) (*TableClient, error) {
-	con := newConnection(serviceURL, cred, options.getConnectionOptions())
+func NewTableClient(tableName string, serviceURL string, cred azcore.Credential, options *TableClientOptions) (*TableClient, error) {
```
catalinaperalta, line 22: Curious, since I heard you mention that azidentity cant be used with Tables....should the credential type here be different than an azcore.Credential interface then? Because azidentity does implement that. Could it lead to some confusion?

sdk/tables/aztables/table_pagers.go
```diff
@@ -0,0 +1,442 @@
...
+	NextPage(context.Context) bool
+
+	// Page returns the current TableQueryResponseResponse.
+	PageResponse() TableEntityQueryResponseResponse
+
+	// Err returns the last error encountered while paging.
+	Err() error
+}
+
+type StructEntityQueryResponsePager interface {
```
catalinaperalta, line 33: Why do you have the word "Struct" in this interface name?

catalinaperalta: Same question for the other types that include it.

sdk/tables/aztables/tableClient.go
christothes: The developer would know, typically. Because a table is schema-less it is possible to get back a result with jagged schema. But if that was expected, you'd use the map result query instead.

sdk/tables/aztables/tableClient.go
christothes: Good question. In other languages, the type is typically `TokenCredential` to indicate this. Based on the naming, I assumed this was not implied to be a token credential type. Is there not an equivalently named type in azcore?

sdk/tables/aztables/table_pagers.go
christothes: To differentiate it from the map version. The naming is really placeholder for now, but I'd be interested to hear any suggestions on how to differentiate these types by name.

sdk/tables/aztables/cosmos_patch_transform_policy.go
```diff
@@ -0,0 +1,27 @@
...
+package aztables
+
+import (
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// CosmosPatchTransformPolicy transforms PATCH requests into POST requests with the "X-HTTP-Method":"MERGE" header set.
+type CosmosPatchTransformPolicy struct{}
```
catalinaperalta, line 15: Wondering if this should have a ctor...like NewCosmosPatchTransformPolicy? To match the policies in azcore. This is probably a bigger discussion that may require both Jeff and Joel's input as well. In my opinion it's a bit odd to call ctors for all the other and not just for this one..

sdk/tables/aztables/table_client.go
```diff
@@ -0,0 +1,135 @@
...
+	Merge   TableUpdateMode = "merge"
+)
+
+// NewTableClient creates a TableClient object using the specified URL and request policy pipeline.
+func NewTableClient(tableName string, serviceURL string, cred azcore.Credential, options *TableClientOptions) (*TableClient, error) {
+	s, err := NewTableServiceClient(serviceURL, cred, options)
+	return s.GetTableClient(tableName), err
+}
+
+func (t *TableClient) Name() string {
```
catalinaperalta, line 35: If this is only needed for testing, I think this should be unexported. If customer's are going to need it, then you should probably just export the field on the struct, that is more Go-ish.

sdk/tables/aztables/table_transactional_batch.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+var defaultChangesetHeaders = map[string]string{
+	"Accept":       "application/json;odata=minimalmetadata",
+	"Content-Type": "application/json",
+	"Prefer":       "return-no-content",
+}
+
+// SubmitTransaction submits the table transactional batch according to the slice of TableTransactionActions provided.
+func (t *TableClient) SubmitTransaction(transactionActions []TableTransactionAction, tableSubmitTransactionOptions *TableSubmitTransactionOptions, ctx context.Context) (TableTransactionResponse, error) {
```
catalinaperalta, line 106: Context should be the first param as a convention.

sdk/tables/aztables/zc_tableConstants.go
```diff
@@ -0,0 +1,23 @@
...
+	OdataType     = "@odata.type"
+	EdmBinary     = "Edm.Binary"
+	EdmBoolean    = "Emd.Boolean"
+	EdmDateTime   = "Edm.DateTime"
+	EdmDouble     = "Edm.Double"
+	EdmGuid       = "Edm.Guid"
+	EdmInt32      = "Edm.Int32"
+	EdmInt64      = "Edm.Int64"
+	Edm           = "Edm.String"
+	ISO8601       = "2006-01-02T15:04:05.9999999Z"
```
catalinaperalta, line 22: Do all of these consts need to be exported? Seems like it'll pollute the exported space bit for intellisense.

catalinaperalta (approved):
I think you can merge this into your branch. I do think we need a small more informal API review with Jeff. There are a few things I can see that would be good to discuss like some naming (for instance finding a replacement for struct in the pager), some return types (like the interfaces), etc. Could be done in the APIView tool to only focus on public surface area for the UX study.

sdk/tables/aztables/table_client.go
christothes: It's a feature we got feedback on for the .NET client. I'll move it to a field.

sdk/tables/aztables/zc_tableConstants.go
christothes: good callout - none of these should be exported.

sdk/tables/aztables/cosmos_patch_transform_policy.go
christothes: I had the same thought - the only reason I didn't is that there is no config, so a standard instance with no arguments is all is needed. I'm fine to change it if we decided that it makes sense.

PR #14693, opened 2021-05-20
azidentity: move scope to resource conversion to GetToken in MSI cred

### LINKED ISSUE #14644: ManagedIdentityCredential doesn't translate scopes to resource
We are working on [support of managed identities in Grafana](https://github.com/grafana/grafana/pull/33807) and trying this library for token requests (both managed identity and client secret scenarios).

Looks like it's problematic to abstract over particular authentication mechanism using [TokenCredential](https://github.com/Azure/azure-sdk-for-go/blob/24b0cd86da44fcb99397bb8d83ac872ecfec5a79/sdk/azcore/credential.go#L37-L41) interface because of the known "scopes vs resource" problem.

If calling code provides a scope name, ClientSecretCredential works, ManagedIdentityCredential fails, if calling code provides a resource name then it's opposite ManagedIdentityCredential works but ClientSecretCredential fails.

Unfortunately, `strings.Join` here (and in other places) in **managed_identity_client.go** cannot work in principle:
https://github.com/Azure/azure-sdk-for-go/blob/24b0cd86da44fcb99397bb8d83ac872ecfec5a79/sdk/azidentity/managed_identity_client.go#L173

I saw mentions previously that AuthenticationPolicy supposedly has some trimming logic.

Maybe it makes sense to add this logic directly into `managedIdentityClient` since there's actual implementation of the protocol which is specifically doesn't support scopes?

catalinaperalta:
Fixes #14644

### DIFF
--- a/sdk/azidentity/managed_identity_credential.go
+++ b/sdk/azidentity/managed_identity_credential.go
@@ -5,7 +5,6 @@ package azidentity
 
 import (
 	"context"
-	"errors"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_credential.go
```diff
@@ -64,6 +65,14 @@ func NewManagedIdentityCredential(clientID string, options *ManagedIdentityCrede
 // scopes: The list of scopes for which the token will have access.
 // Returns an AccessToken which can be used to authenticate service client calls.
 func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts azcore.TokenRequestOptions) (*azcore.AccessToken, error) {
+	if opts.Scopes == nil {
+		return nil, errors.New("must specify a resource in order to authenticate")
```
jhendrixMSFT, line 69: Shouldn't this, and below, return a `CredentialUnavailableError`?

catalinaperalta: So I did think about this. I believe at one point we said that CredentialUnavailableError would only be called when the conditions to create the credential are not available (i.e. in the ctor). And AuthenticationFailedErrors would be exactly for when we attempt to authenticate but cannot do so. So I'm more tempted to go with AuthenticationFailedError since this is in the call to GetToken(). What do you think?

jhendrixMSFT: AuthenticationFailedError is fine (frankly either one would work, I think the important part is it's not a naked error).

catalinaperalta: Done, also added logging.

jhendrixMSFT (approved):

PR #14725, opened 2021-05-28
rename armmonitor to arminsights

ArcturusZhang:
Related PR: https://github.com/Azure/azure-rest-api-specs/pull/14588/files
I found that the monitor package previously had `insights` as its namespace, therefore I believe we might need to keep the same one in track 2, therefore I made this change.


sdk/monitor/arminsights/ci.yml
```diff
@@ -2,14 +2,14 @@
 trigger:
   paths:
     include:
-    - sdk/monitor/armmonitor
+    - sdk/monitor/arminsights
```
jhendrixMSFT, line 6: Can you confirm with the service team what the name should be?  I recall a few years ago they underwent some rebranding, I don't recall if we had actually fixed that in track 1 due to the breaking change.

jhendrixMSFT (changes requested):

jhendrixMSFT (approved):

Issue #14729, opened 2021-06-01
Track2 Poller crash


yupwei68:
### Bug Report

- import path of package in question, e.g. `.../services/compute/mgmt/2018-06-01/compute`
- SDK version e.g. `master`, `latest`, `18.1.0`
  - Specify the exact commit if possible; one way to get this is the REVISION
    column output by `dep status "github.com/Azure/azure-sdk-for-go`.
- output of `go version`

- What happened?
When trying track2, the plugin crash on poller.

- What did you expect or want to happen?
No crash

- How can we reproduce it?
Use track2 sdk built Terrform plugin to create a resource avs, and after 1.5 hrs, it crashed.

- Anything we should know about your environment.

It crashed at https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/armcore/poller.go#L198, looking alike it's a nil pointer problem.

When it sends request and receives no response, it crashed.

[DEBUG] Request to https://management.azure.com/subscriptions/.../resourceGroups/acctestRG-Vmware-yup/providers/Microsoft.AVS/privateClouds/acctest-PC-yup/operationstatuses/1be210d6-d357-42a0-bf7e-f8da4de108ea?api-version=2020-03-20 completed with no response

crash log:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe86f04]

goroutine 75 [running]:
github.com/Azure/azure-sdk-for-go/sdk/armcore.(*poller).Poll(0xc0010aafb0, 0x5edd618, 0xc0016ae180, 0xc0016e0200, 0x9, 0x10, 0x1, 0x1, 0x0)
        /home/weiyuping/work/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/Azure/azure-sdk-for-go/sdk/armcore/poller.go:198 +0xc4
github.com/Azure/azure-sdk-for-go/sdk/armcore.(*poller).PollUntilDone(0xc0010aafb0, 0x5edd618, 0xc0016ae180, 0x77359400, 0xc0016e0200, 0x9, 0x10, 0x4e53660, 0xc001180480, 0xc0003a80e0, ...)
        /home/weiyuping/work/src/github.com/terraform-providers/terraform-provider-azurerm/vendor/github.com/Azure/azure-sdk-for-go/sdk/armcore/poller.go:222 +0x17e
[...]

PR #14738, opened 2021-06-02
Return an error on nil response

jhendrixMSFT:
Properly handle the case where the HTTP client returns a nil response
and error.
Refactored poller.Poll() to simplify error propagation.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/14729

jhendrixMSFT:
There was a fix in `azcore` to return an error if the response and error were both `nil`.  It could be the case that the repro was using an older version of `azcore` without the fix.

PR #14741, opened 2021-06-03
azidentity: Adding Resource ID support in ManagedIdentityCredential

### LINKED ISSUE #12184: Add support for MSI resource ID
Similar to: https://github.com/Azure/go-autorest/pull/544

catalinaperalta:
Fixes #12184

Adding support for passing in a resource ID instead of a client ID for managed identities.

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -53,7 +53,7 @@ type managedIdentityClient struct {
 	imdsAvailableTimeoutMS time.Duration
 	msiType                msiType
 	endpoint               string
-	id                     IDKind
+	id                     ManagedIdentityIDKind
 }
 
 type wrappedNumber json.Number
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -45,10 +45,6 @@ const (
 	msiTypeAzureArc            msiType = 6
 )
 
-const (
-	qpResID string = "mi_res_id"
-)
-
 // managedIdentityClient provides the base for authenticating in managed identity environments
 // This type includes an azcore.Pipeline and TokenCredentialOptions.
 type managedIdentityClient struct {
@@ -168,7 +164,7 @@ func (c *managedIdentityClient) createAuthRequest(ctx context.Context, clientID 
 	}
 }
 
-func (c *managedIdentityClient) createIMDSAuthRequest(ctx context.Context, clientID string, scopes []string) (*azcore.Request, error) {
+func (c *managedIdentityClient) createIMDSAuthRequest(ctx context.Context, id string, scopes []string) (*azcore.Request, error) {
 	request, err := azcore.NewRequest(ctx, http.MethodGet, c.endpoint)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


sdk/azidentity/managed_identity_credential.go
```diff
@@ -11,9 +11,22 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
+type IDKind int
```
jhendrixMSFT, line 14: These will need doc comments.

sdk/azidentity/managed_identity_credential.go
```diff
@@ -54,10 +68,10 @@ func NewManagedIdentityCredential(clientID string, options *ManagedIdentityCrede
 	// Assign the msiType discovered onto the client
 	client.msiType = msiType
 	// check if no clientID is specified then check if it exists in an environment variable
-	if len(clientID) == 0 {
-		clientID = os.Getenv("AZURE_CLIENT_ID")
+	if len(id) == 0 {
+		id = os.Getenv("AZURE_CLIENT_ID")
```
jhendrixMSFT, line 72: Do we also need to support a resource ID env var?

sdk/azidentity/managed_identity_client.go
```diff
@@ -171,14 +173,16 @@ func (c *managedIdentityClient) createIMDSAuthRequest(ctx context.Context, clien
 	q := request.URL.Query()
 	q.Add("api-version", c.imdsAPIVersion)
 	q.Add("resource", strings.Join(scopes, " "))
-	if clientID != "" {
-		q.Add(qpClientID, clientID)
+	if c.id == ResourceID {
```
jhendrixMSFT, line 176: This logic should be its own func to avoid the duplication.  Might also be safer to implement as a `switch` with a `default` clause to guard against future values being added and the logic not properly updated.

sdk/azidentity/managed_identity_client.go
```diff
@@ -171,14 +173,16 @@ func (c *managedIdentityClient) createIMDSAuthRequest(ctx context.Context, clien
...
+		q.Add(string(qpResID), id)
```
jhendrixMSFT, line 177: Unnecessary type-cast?

sdk/azidentity/managed_identity_credential.go
catalinaperalta: Done

sdk/azidentity/managed_identity_credential.go
catalinaperalta: I do think we need to, but I wanted to wait to talk to someone like Jon Gallant who I think dealt with all of the env vars that we allow and naming before adding it. I opened an issue to track that: https://github.com/Azure/azure-sdk-for-go/issues/14742

sdk/azidentity/managed_identity_client.go
catalinaperalta: Yes it is! Thanks for catching that, it was a product of my previous implementation where the value of the resource ID was a different type. Done.

sdk/azidentity/managed_identity_client.go
catalinaperalta: I'm a little hesitant to abstract this into a func to avoid duplication because I dont know that all of the ID kinds are supported in every MSI environment. I checked that resource ID is supported for those that I added it to in this PR. Therefore this check is a per environment functionality that needs to be added. 
For instance App Service also supports `principal_id` which we may add to IDKind in the future, but azure VMs dont. Instead Azure VMs support `object_id`.

jhendrixMSFT: Makes sense, leave as-is then.

jhendrixMSFT (approved):

sdk/azidentity/managed_identity_credential.go
```diff
@@ -11,9 +11,25 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
+// IDKind is used to specify the type of identifier that is passed in for a user-assigned managed identity.
+type IDKind int
```
JeffreyRichter, line 15: I think IDKind might be too generic sounding. Perhaps ManagedIdentityIDKind or maybe just ManagedIdentityKind. I'm trying to draw a tight relationship that this is used specifically with the ManagedIdentityCredential type.

jhendrixMSFT: My vote is for `ManagedIdentityKind`

catalinaperalta: I was actually leaning towards `ManagedIdentityIDKind` because `ManagedIdentityKind` implies that this is saying what type of managed identity they're choosing instead of the type of identifier. Although I think it is very long. I'm trying to think of other possibilities too...maybe something related to the fact that this is something used with user assigned identities.

catalinaperalta: I think I'll take `ManagedIdentityIDKind`. I agree that it's better to leave it more general than my idea to specify something related to user assigned identities. 👍

catalinaperalta: Done.

JeffreyRichter (approved):

Issue #14730, opened 2021-06-01
Long running operation in creating a VM does not return a final state URL


ArcturusZhang:
### Bug Report

- import path of package in question
  `github.com/Azure/azure-sdk-for-go/sdk/compute/armcompute`
- SDK version 
  `v0.1.0`
- output of `go version`
  `go version go1.16 windows/amd64`

- What happened?
  The response returned by `PollUntilDone` function is basically empty without any information of this resource.
- What did you expect or want to happen?
  We should get the full parameters in the response of `PollUntilDone` function
- How can we reproduce it?
  Sample code here: https://github.com/ArcturusZhang/azure-sdk-for-go-track2-sample/blob/dbd77ed6bf9cbf516dc31f7cb754f0a4626a3537/main.go#L338
- Anything we should know about your environment.

Related issue: https://github.com/Azure/azure-sdk-for-go/issues/14665

jhendrixMSFT:
Fixed in `armcompute@v0.2.0`

PR #14752, opened 2021-06-07
LRO poller rewrite

jhendrixMSFT:
Simplified implementation of LRO pollers for ARM.
Public surface area has been slightly changed, making it identical to
the data-plane implementation.
The different polling mechanisms have been split into internal packages,
with an exported LROPoller that implements the overall polling algorithm.

### DIFF
--- a/sdk/armcore/internal/pollers/async/async.go
+++ b/sdk/armcore/internal/pollers/async/async.go
@@ -65,7 +65,7 @@ func New(resp *azcore.Response, finalState string, pollerID string) (*Poller, er
 			return nil, err
 		}*/
 		// for DELETE/PATCH/POST, provisioning state is optional
-		state = "InProgress"
+		state = pollers.StatusInProgress
 	} else if err != nil {
 		return nil, err
 	}
--- a/sdk/armcore/internal/pollers/async/async.go
+++ b/sdk/armcore/internal/pollers/async/async.go
@@ -57,10 +57,13 @@ func New(resp *azcore.Response, finalState string, pollerID string) (*Poller, er
 	// check for provisioning state
 	state, err := pollers.GetProvisioningState(resp)
 	if errors.Is(err, pollers.ErrNoBody) || errors.Is(err, pollers.ErrNoProvisioningState) {
-		if resp.Request.Method == http.MethodPut {
+		// NOTE: the ARM RPC spec explicitly states that for async PUT the initial response MUST
+		// contain a provisioning state.  to maintain compat with track 1 and other implementations
+		// we are explicitly relaxing this requirement.
+		/*if resp.Request.Method == http.MethodPut {
 			// initial response for a PUT requires a provisioning state
 			return nil, err
-		}
+		}*/
 		// for DELETE/PATCH/POST, provisioning state is optional
 		state = "InProgress"
 	} else if err != nil {
--- a/sdk/armcore/internal/pollers/async/async.go
+++ b/sdk/armcore/internal/pollers/async/async.go
@@ -87,13 +87,15 @@ func (p *Poller) FinalGetURL() string {
 		// for PATCH and PUT, the final GET is on the original resource URL
 		return p.OrigURL
 	} else if p.Method == http.MethodPost {
-		// for POST, we need to consult the final-state-via flag
-		if p.FinalState == finalStateLoc && p.LocURL != "" {
-			return p.LocURL
+		if p.FinalState == finalStateAsync {
+			return ""
 		} else if p.FinalState == finalStateOrig {
 			return p.OrigURL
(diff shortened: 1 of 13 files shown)


jhendrixMSFT:
Fixes
- https://github.com/Azure/azure-sdk-for-go/issues/14729
- https://github.com/Azure/azure-sdk-for-go/issues/14730

jhendrixMSFT:
@lmazuel if you have time could you please take a look at the contents of `aync.go`, `body.go`, and `loc.go` to see if anything looks suspicious?  I used the Python track 2 impl for inspiration.

jhendrixMSFT:
This is causing some test failures in the test server, I'm investigating.

jhendrixMSFT:
Autorest codegen tests are now passing and I've tested with `armcompute` and `armredis`.

sdk/armcore/poller.go
```diff
@@ -6,958 +6,307 @@
...
+			if err := delay(ctx, retryAfter); err != nil {
 				logPollUntilDoneExit(err)
 				return nil, err
 			}
 		}
 	}
 	// begin polling the endpoint until a terminal state is reached
 	for {
-		resp, err := p.Poll(ctx, pipeline)
+		resp, err := l.Poll(ctx)
```
catalinaperalta, line 237: this looks so much better!

catalinaperalta (commented):
I'm really liking this, it's so much easier to reason about and the division of the pollers into subpackages is great! Just left a few questions.

sdk/armcore/internal/pollers/async/async.go
```diff
@@ -0,0 +1,118 @@
...
+	// check for provisioning state
+	state, err := pollers.GetProvisioningState(resp)
+	if errors.Is(err, pollers.ErrNoBody) || errors.Is(err, pollers.ErrNoProvisioningState) {
+		// NOTE: the ARM RPC spec explicitly states that for async PUT the initial response MUST
+		// contain a provisioning state.  to maintain compat with track 1 and other implementations
+		// we are explicitly relaxing this requirement.
+		/*if resp.Request.Method == http.MethodPut {
+			// initial response for a PUT requires a provisioning state
+			return nil, err
+		}*/
```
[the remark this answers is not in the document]
jhendrixMSFT: Correct, we didn't enforce this in track 1 either (nor does python's implementation).  In fact, there's a test in the autorest LRO suite that covers this exact scenario.  I think it's likely to not be a problem in practice.  If we get here, it means we have an `Azure-AsyncOperation` header to poll for status on, so should be safe to assume the operation is in progress, we just make up the status message "InProgress" instead of what the endpoint should return.

sdk/armcore/internal/pollers/async/async.go
```diff
@@ -0,0 +1,118 @@
...
+
+// Poller is an LRO poller that uses the Azure-AsyncOperation pattern.
+type Poller struct {
+	Type       string `json:"type"`
+	AsyncURL   string `json:"asyncURL"`
+	LocURL     string `json:"locURL"`
+	OrigURL    string `json:"origURL"`
+	Method     string `json:"method"`
+	FinalState string `json:"finalState"`
+	CurState   string `json:"state"`
```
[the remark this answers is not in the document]
jhendrixMSFT: Added comments.

sdk/armcore/poller.go
```diff
@@ -6,958 +6,307 @@
...
+		return nil, errors.New("cannot return a final response from a poller in a non-terminal state")
 	}
-	if !resp.HasStatusCode(pollingCodes[:]...) {
-		return nil, p.pt.handleError(resp)
+	// if there's nothing to unmarshall into or no response body just return the final response
+	if respType == nil {
+		return l.resp.Response, nil
+	} else if l.resp.StatusCode == http.StatusNoContent || l.resp.ContentLength == 0 {
+		azcore.Log().Write(azcore.LogLongRunningOperation, "final response specifies a response type but no payload was received")
+		return l.resp.Response, nil
```
catalinaperalta, line 186: This isnt an error right?

catalinaperalta (approved):
Looks great! Left one last question just double checking.

sdk/armcore/poller.go
jhendrixMSFT: Correct.  I wanted to log this in case customers ask why a poller returns an empty `Widget`.  This would indicate a swagger or service issue (i.e. the swagger says the operation returns a `Widget` but the endpoint doesn't).

catalinaperalta: Ok yeah, it seemed strange to me that it wouldnt be an error if that's what's expected, but I get it being needed for some services that might have this issue.

PR #14767, opened 2021-06-10
Go recorded test framework

christothes:
Merging the testFramework module to master.


sdk/internal/testframework/request_matcher.go
```diff
@@ -0,0 +1,111 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"reflect"
+
+	"github.com/dnaeon/go-vcr/cassette"
+)
+
+type RequestMatcher struct {
```
jhendrixMSFT, line 18: How is this used, or is it vestigial and can be removed?

christothes: It's used to provide custom logic for how requests should be matched.

christothes: OK, yeah it has been a while since I have looked at this :) 
You're correct - I think in its current state, it only serves as the default implementation for matching logic. We exported the Sanitizer on `Recording` but never did the same for the RequestMatcher. I can do that now to allow for the extensibility, or we can let it evolve as needed. I'm fine to make the changes if we want to export it now.

jhendrixMSFT: My default statement on such things is to remove them, and add them if/when there's an actual need for it.  If @mohsha-msft can comment if `azblob` needs this functionality then let's keep it and ensure it works.

christothes: It's definitely needed, the question is whether or not we expose the extensibility.

sdk/internal/testframework/request_matcher.go
```diff
@@ -0,0 +1,111 @@
...
+	"X-Ms-Date":              nil,
+	"x-ms-date":              nil,
+	"x-ms-client-request-id": nil,
+	"User-Agent":             nil,
+	"Request-Id":             nil,
+	"traceparent":            nil,
+	"Authorization":          nil,
+}
+
+var recordingHeaderMissing = "Test recording headers do not match. Header '%s' is present in request but not in recording."
```
jhendrixMSFT, line 33: I think these all can be constants?

sdk/internal/testframework/request_matcher.go
christothes: I think I'll go ahead and implement the changes, as I am pretty sure we'll need them.

sdk/internal/testframework/request_matcher.go
christothes: they have tokens to indicate which specific value was mismatched.

jhendrixMSFT: They're just format strings though right?  They can be constants.

christothes: oh, yes.. sorry, I misunderstood :)

sdk/internal/testframework/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+In the snippet below, demonstrates an example test setup func in which we are initializing the `TestContext`
+with the methods that will be invoked when your recorded test needs to Log, Fail, get the Name of the test,
+or indicate that the test IsFailed.
+
+***Note**: an instance of TestContext should be initialized for each test.*
+
+```go
+// a map to store our created test contexts
+var clientsMap map[string]*testContext = make(map[string]*testContext)
```
mohsha-msft, line 25: QuickQuestion: This `testContext` will not be visible outside of the `testframework` package. Isn't this supposed to be a public one?

sdk/internal/testframework/README.md
```diff
@@ -0,0 +1,161 @@
...
+```go
+//func recordedTestSetup(t *testing.T, testName string, mode testframework.RecordMode) {
+//  <...>
+    // Set our client's HTTPClient to our recording instance.
+    // Optionally, we can also configure MaxRetries to -1 to avoid the default retry behavior.
+    client, err := NewTableServiceClient(uri, cred, &TableClientOptions{HTTPClient: recording, Retry: azcore.RetryOptions{MaxRetries: -1}})
+    assert.Nil(err)
+
+    // either return your client instance, or store it somewhere that your test can use it for test execution.
+    clientsMap[testName] = &testContext{client: client, recording: recording, context: &context}
```
mohsha-msft, line 83: It doesn't seem same as the one provided in `testcontext.go`. Any implementing test suite is required to create this one on their own?

```go
type testContext struct {
    client     *URL
    recording  *testframework.Recording
    context    *testframework.TestContext
}
```

sdk/internal/testframework/README.md
christothes: yes, it should be. In fact I'm not sure why there is an unexported one. I think I left it there by accident.

mohsha-msft: It's starting with small case so I don't think it'll be visible outside of `testframework` package.

sdk/internal/testframework/README.md
christothes: This is just a helper for keeping track of test state. Different tests may have different requirements as to what they want to store as state. However, I agree the name makes it confusing with the testframeowk.TestContext type.

sdk/internal/testframework/README.md
christothes: correct - we should only be referring to `testframework.TestContext` The local `testContext` is something else, which is just confusing because of the shared name.

mohsha-msft: Now it all makes sense. Thanks 😄

sdk/internal/testframework/README.md
```diff
@@ -0,0 +1,161 @@
...
+)
+
+type tableServiceClientLiveTests struct {
+    suite.Suite
+    mode         testframework.RecordMode
+}
+
+// Hookup to the testing framework
+func TestServiceClient_Storage(t *testing.T) {
+    storage := tableServiceClientLiveTests{endpointType: StorageEndpoint, mode: testframework.Playback /* change to Record to re-record tests */}
```
mohsha-msft, line 135: `endpointType` is not part of `tableServiceClientLiveTests` struct.

```go
type tableServiceClientLiveTests struct {    
    suite.Suite    
    mode         testframework.RecordMode
}
```

mohsha-msft, line 135: `endpointType` is not part of `tableServiceClientLiveTests` struct.

```go
type tableServiceClientLiveTests struct {    
    suite.Suite    
    mode         testframework.RecordMode
}
```

christothes: Good catch - I was trying to simplify the example, which was taken from the actual aztable package. I missed removing this reference to endpointType.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	// Try loading the recording if it already exists to hydrate the variables
+	err = recording.initVariables()
+	if err != nil {
+		return nil, err
+	}
+
+	// set the recorder Matcher
+	recording.Matcher = DefaultMatcher(c)
```
mohsha-msft, line 105: So far the test has been able to work perfectly. Appreciate your work @christothes. There are few questions I am curious about 

I see we're using `DefaultMatchers` here. While the default matchers work perfectly in most of the cases, there are few cases where I believe users may need to use custom matchers. So they should be provided options to supply their custom matchers. 

Here are the couple of scenarios which I found while working on integrating `testframework` in `azblob` SDK
1. For upload/download 1MB/2MB/4MB/8MB file to Azure Service (which is a common scenario), the recording file is storing body in request/response which is increasing the size of file and make it infeasible to keep in the repo. And depending on the use-case, I may not be interesting in comparing body content at all.
2. While dealing with Shared Access Signature (SAS, I want to omit the request URL and ignore it from comparisons because those URLs hold secret. I want to provide my custom headers to match apart from 

```go
var ignoredHeaders = map[string]*string{
	"Date":                   nil,
	"X-Ms-Date":              nil,
	"x-ms-date":              nil,
	"x-ms-client-request-id": nil,
	"User-Agent":             nil,
	"Request-Id":             nil,
	"traceparent":            nil,
	"Authorization":          nil,
}
```
supplied in `DefaultMatchers`

```go
matcher := &RequestMatcher{
		context:        testContext,
		IgnoredHeaders: ignoredHeaders,
	}
```

mohsha-msft: Also curious on your opinion on whether we can omit body from the recording as well

christothes: Some of the most recent commits on this PR added the ability to configure custom matchers.  For example, see the [SetBodyMatcher func here](https://github.com/Azure/azure-sdk-for-go/blob/921c67f7d8afeb91605c4ed912adc462da72636f/sdk/internal/testframework/request_matcher.go#L68). 

For setting custom headers to ignore for matching, you'd add items to the [IgnoredHeaders map](https://github.com/Azure/azure-sdk-for-go/blob/921c67f7d8afeb91605c4ed912adc462da72636f/sdk/internal/testframework/request_matcher.go#L20)

christothes: As far as eliminating the need to record the body content at all, I think this should be acheivable by setting a [custom BodySanitizer](https://github.com/Azure/azure-sdk-for-go/blob/921c67f7d8afeb91605c4ed912adc462da72636f/sdk/internal/testframework/recording_sanitizer.go#L45) where the "sanitization" is to simply return and empty object (or perhaps one that contains only a single "sanitized" property to make it clear that it was due to sanitization and not a truly empty body)

sdk/internal/testframework/request_matcher.go
```diff
@@ -0,0 +1,176 @@
...
+	m.methodMatcher = func(reqVal string, recVal string) bool {
+		isMatch := matcher(reqVal, recVal)
+		if !isMatch {
+			m.context.Log(fmt.Sprintf(methodMismatch, recVal, recVal))
+		}
+		return isMatch
+	}
+}
+
+func DefaultStringMatcher(s1 string, s2 string) bool {
```
jhendrixMSFT, line 98: Can this not be exported, or do we expect users of this package to need this?

sdk/internal/testframework/request_matcher.go
```diff
@@ -0,0 +1,176 @@
...
+	"io/ioutil"
+	"net/http"
+	"reflect"
+
+	"github.com/dnaeon/go-vcr/cassette"
+)
+
+type RequestMatcher struct {
+	context        TestContext
+	IgnoredHeaders map[string]*string
```
jhendrixMSFT, line 20: Does the value `*string` have actual usage, or is the idea to use this as a hash set?  If it's the latter, the convention is to use `struct{}`.

christothes: Thanks - I was wondering what the convention was.

sdk/internal/testframework/request_matcher.go
christothes: I don't believe it needs to be exported

sdk/internal/testframework/request_matcher.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	recordingHeaderMissing = "Test recording headers do not match. Header '%s' is present in request but not in recording."
+	requestHeaderMissing   = "Test recording headers do not match. Header '%s' is present in recording but not in request."
+	headerValuesMismatch   = "Test recording header '%s' does not match. request: %s, recording: %s"
+	methodMismatch         = "Test recording methods do not match. request: %s, recording: %s"
+	urlMismatch            = "Test recording URLs do not match. request: %s, recording: %s"
+	bodiesMismatch         = "Test recording bodies do not match.\nrequest: %s\nrecording: %s"
+)
+
+func DefaultMatcher(testContext TestContext) *RequestMatcher {
```
jhendrixMSFT, line 49: Can we also not export this?  I see there's an exported `Matcher` on the `Recording`, is the expectation that custom matching would be obtained by manipulating this field?

christothes: The `Matcher` field on `Recording is an instance of this type. The intended means for manipulating it would be:
- Modify `IgnoredHeaders`
- Set a custom matcher by calling `SetBodyMatcher`, `SetURLMatcher`, or `SetMethodMatcher`

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type StringSanitizer func(*string)
+
+const SanitizedValue string = "sanitized"
+const SanitizedBase64Value string = "Kg=="
+
+var sanitizedValueSlice = []string{SanitizedValue}
+
+func DefaultSanitizer(recorder *recorder.Recorder) *RecordingSanitizer {
```
jhendrixMSFT, line 29: Same question as with matcher.

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(i.Request.Body) > 0 {
+		s.sanitizeBodies(&i.Request.Body)
+	}
+	if len(i.Response.Body) > 0 {
+		s.sanitizeBodies(&i.Response.Body)
+	}
+	return nil
+}
+
+func DefaultStringSanitizer(s *string) {}
```
jhendrixMSFT, line 83: Unexport?

sdk/internal/testframework/README.md
```diff
@@ -0,0 +1,161 @@
+# Azure SDK for Go Recorded Test Framework
```
jhendrixMSFT, line 1: Would be good to mention that by default, the executed tests run in `record` mode if no recordings are available.  And to update recordings, simply delete them and rerun the tests.

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,416 @@
...
+
+	// wire up the sanitizer
+	recording.Sanitizer = DefaultSanitizer(rec)
+
+	return recording, err
+}
+
+// GetRecordedVariable returns a recorded variable. If the variable is not found we return an error
+// variableType determines how the recorded variable will be saved. Default indicates that the value should be saved without any sanitation.
+func (r *Recording) GetRecordedVariable(name string, variableType VariableType) (string, error) {
```
jhendrixMSFT, line 116: I think this should be renamed to something like `GetEnvVar()` and the doc comment updated appropriately.  It wasn't clear to me that this retrieved the value from the environment until I dug into the implementation.

sdk/internal/testframework/request_matcher.go
christothes: It would probably be more idiomatic to name this `NewRequestMatcher`. I initially named it this to convey that it is pre-configured with the default behavior and doesn't require calls to any of the `Set*Matcher` methods if you prefer the default behavior. I'll add a comment.

sdk/internal/testframework/README.md
christothes: It's mentioned in the comment on line 38

```go
//mode should be testframework.Playback. This will automatically record if no test recording is available and playback if it is.
recording, err := testframework.NewRecording(context, mode) 
```

jhendrixMSFT:
Here's an observation I had while testing this out.  I don't know what the fix looks like, perhaps it's just documentation/examples.

I wrote a test for `azidentity` that uses recordings, retrieving an access token for client secret credentials.  Using a `Secret_String` for the tenant ID is problematic as this is part of the request URL so there's no match during play-back.

Changing the tenant ID to `Default` works around the above (not sure this is how we'd want to check it in though).  Now, playback fails due to a mismatch of `Content-Length` headers.  I believe this is because the sanitized client ID and secrets are used in the request body which is different than the recording.

The recording contained the credentials as well as the access token which we'd obviously not want to commit, or even expose in a PR.

I suppose these problems might be unique to `azidentity` since it's directly testing endpoints with sensitive data, and will require heavy use of custom matchers etc.

christothes:
> I wrote a test for `azidentity` that uses recordings, retrieving an access token for client secret credentials. Using a `Secret_String` for the tenant ID is problematic as this is part of the request URL so there's no match during play-back.

For this scenario I think you'd want to set a custom URLSanitizer.

https://github.com/Azure/azure-sdk-for-go/blob/438d7e1df788ae0ce75e3f268b2465cc65a0c2d1/sdk/internal/testframework/recording_sanitizer.go#L52

> The recording contained the credentials as well as the access token which we'd obviously not want to commit, or even expose in a PR.

This should just require setting a custom body sanitizer that knows how to parse the token in the response and replace it with a sanitized value.

https://github.com/Azure/azure-sdk-for-go/blob/438d7e1df788ae0ce75e3f268b2465cc65a0c2d1/sdk/internal/testframework/recording_sanitizer.go#L47

> I suppose these problems might be unique to `azidentity` since it's directly testing endpoints with sensitive data, and will require heavy use of custom matchers etc.

Actually it should be custom sanitizers you need here. Although I don't think the sanitization is all that complex in these cases.

catalinaperalta:
> Actually it should be custom sanitizers you need here. Although I don't think the sanitization is all that complex in these cases.

I just wanted to chime in here since I also worked on getting some recordings set up for azidentity. Aside from the custom sanitizers, there needs to be a custom request matcher due to the problem Joel pointed out about the mismatched headers.

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type RecordingSanitizer struct {
+	recorder          *recorder.Recorder
+	headersToSanitize map[string]*string
+	urlSanitizer      StringSanitizer
+	bodySanitizer     StringSanitizer
+}
+
+type StringSanitizer func(*string)
```
jhendrixMSFT, line 22: Is there a reason why this is pointer-to-string, i.e. does the empty string have semantic importance?

christothes:
> > Actually it should be custom sanitizers you need here. Although I don't think the sanitization is all that complex in these cases.
> 
> I just wanted to chime in here since I also worked on getting some recordings set up for azidentity. Aside from the custom sanitizers, there needs to be a custom request matcher due to the problem Joel pointed out about the mismatched headers.

Could you give an example of where sanitization wouldn't do the right thing in terms of header matching?

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"net/http"
+
+	"github.com/dnaeon/go-vcr/cassette"
+	"github.com/dnaeon/go-vcr/recorder"
+)
+
+type RecordingSanitizer struct {
+	recorder          *recorder.Recorder
+	headersToSanitize map[string]*string
```
jhendrixMSFT, line 17: `map[string]struct{}`?

sdk/internal/testframework/recording_sanitizer.go
christothes: There may not be a good reason for this to be a pointer.

sdk/internal/testframework/recording_sanitizer.go
```diff
@@ -0,0 +1,85 @@
...
+	s.bodySanitizer = sanitizer
+}
+
+// AddUriSanitizer configures the supplied StringSanitizer to sanitize recording request and response URLs
+func (s *RecordingSanitizer) AddUrlSanitizer(sanitizer StringSanitizer) {
+	s.urlSanitizer = sanitizer
+}
+
+func (s *RecordingSanitizer) sanitizeHeaders(header http.Header) {
+	for headerName := range s.headersToSanitize {
```
jhendrixMSFT, line 57: While not necessarily important for testing, iterating over a map is fairly expensive in Go.

christothes: That's surprising - is there a more efficient way to do this?

jhendrixMSFT: If you're not using the map as a hash set, it might be better to just use a slice of headers to sanitize instead.  For this case I don't think it really matters, I just wanted to point out that this can be problematic in some cases (I'm acutely aware of this from profiling msal/go where this happens in a few places and it shows up fairly dominant).

christothes: I guess we aren't using it as a hash in this case, since we hash lookup in the request, but only iterate through the sanitized header names. Good callout!

catalinaperalta:
> Could you give an example of where sanitization wouldn't do the right thing in terms of header matching?

I suppose you could just sanitize all the headers. I just recall tracking this down to the matcher and for a header that azidentity doesnt particularly change...if I recall correctly.

christothes:
> > Could you give an example of where sanitization wouldn't do the right thing in terms of header matching?
> 
> I suppose you could just sanitize all the headers. I just recall tracking this down to the matcher and for a header that azidentity doesnt particularly change...if I recall correctly.

Hmm - no big deal, but if you happen to find the real example of this, I'd be curious in which scenario headers wouldn't match regardless if they were sanitized.

sdk/internal/testframework/recording_sanitizer.go
christothes: Thought about this more, I think it needs to be a pointer so that we can change the value passed in.

jhendrixMSFT: Ah I see, I didn't realize this is used as an out parameter.  A comment for this type would be good.

jhendrixMSFT: (Probably because out params are somewhat uncommon in Go)

sdk/internal/testframework/recording.go
```diff
@@ -0,0 +1,420 @@
+// +build go1.13
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package testframework
```
jhendrixMSFT, line 6: I've been struggling to come up with a better name.  Do you think there will be content in here other than recordings?

christothes: > Do you think there will be content in here other than recordings? It it internal code, so not super-critical.

Probably not.. Were you thinking something like `recordedtests` or `recordedtestfixture`?

jhendrixMSFT: I was think a bit more brief, like `recording`.  So code would look like `recording.NewRecording()`.  This would also facilitate removal of any `Recording` prefix on type names, e.g. `RecordingSanitizer` would simply become `Sanitizer` to avoid stuttering (i.e. avoid `recording.RecordingSanitizer`).

christothes: I like it!

sdk/internal/testframework/recording_test.go
```diff
@@ -0,0 +1,356 @@
+// +build go1.13
```
jhendrixMSFT, line 1: You'll also need to rename the directory to match that of the package name.

jhendrixMSFT (approved):

PR #14783, opened 2021-06-16
azidentity: add support for service fabric MSI environment

### LINKED ISSUE #11878: Azure.Identity Support for Managed Identity for Service Fabric Applications
A common challenge when building cloud applications is how to securely manage the credentials in your code for authenticating to various services without saving them locally on a developer workstation or in source control. Managed identities for Azure solve this problem for all your resources in Azure Active Directory (Azure AD) by providing them with automatically managed identities within Azure AD. You can use a service's identity to authenticate to any service that supports Azure AD authentication, including Key Vault, without any credentials stored in your code.

https://docs.microsoft.com/en-us/azure/service-fabric/concepts-managed-identity

https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/services-support-managed-identities#azure-service-fabric

catalinaperalta:
Fixes #11878

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -127,12 +127,12 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
 	switch v := value.ExpiresOn.(type) {
 	case int:
 		// service fabric is one of the MSI environments that returns an int
-		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
+		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Unix(int64(v), 0).UTC()}, nil
 	case float64:
-		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
+		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Unix(int64(v), 0).UTC()}, nil
 	case string:
 		if expiresOn, err := strconv.Atoi(v); err == nil {
-			return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(expiresOn)).UTC()}, nil
+			return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Unix(int64(expiresOn), 0).UTC()}, nil
 		}
 		// this is the case when expires_on is a time string
 		// this is the format of the string coming from the service
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -129,7 +129,6 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
 		// service fabric is one of the MSI environments that returns an int
 		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
 	case float64:
-		fmt.Println(v)
 		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
 	case string:
 		if expiresOn, err := strconv.Atoi(v); err == nil {
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -24,13 +24,15 @@ const (
 )
 
 const (
-	arcIMDSEndpoint    = "IMDS_ENDPOINT"
-	identityEndpoint   = "IDENTITY_ENDPOINT"
-	identityHeader     = "IDENTITY_HEADER"
-	msiEndpoint        = "MSI_ENDPOINT"
-	msiSecret          = "MSI_SECRET"
-	imdsAPIVersion     = "2018-02-01"
-	azureArcAPIVersion = "2019-08-15"
+	arcIMDSEndpoint          = "IMDS_ENDPOINT"
+	identityEndpoint         = "IDENTITY_ENDPOINT"
+	identityHeader           = "IDENTITY_HEADER"
+	identityServerThumbprint = "IDENTITY_SERVER_THUMBPRINT"
+	msiEndpoint              = "MSI_ENDPOINT"
+	msiSecret                = "MSI_SECRET"
+	imdsAPIVersion           = "2018-02-01"
+	azureArcAPIVersion       = "2019-08-15"
+	serviceFabricAPIVersion  = "2019-07-01-preview"
 )
 
 type msiType int
@@ -43,6 +45,7 @@ const (
 	msiTypeUnavailable         msiType = 4
 	msiTypeAppServiceV20190801 msiType = 5
 	msiTypeAzureArc            msiType = 6
+	msiTypeServiceFabric       msiType = 7
 )
 
 // managedIdentityClient provides the base for authenticating in managed identity environments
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -121,19 +124,29 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
...
-		return &azcore.AccessToken{Token: value.Token, ExpiresOn: eo}, nil
-	} else if expiresOn, err := time.Parse("1/2/2006 15:04:05 +00:00", value.ExpiresOn); err == nil { // the date string specified is for Linux OS
-		eo := expiresOn.UTC()
-		return &azcore.AccessToken{Token: value.Token, ExpiresOn: eo}, nil
-	} else {
-		return nil, err
+	switch v := value.ExpiresOn.(type) {
+	case int:
+		// service fabric is one of the MSI environments that returns an int
+		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
```
jhendrixMSFT, line 130: Is the value returned actually the number of seconds?

sdk/azidentity/managed_identity_client.go
```diff
@@ -121,19 +124,29 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
...
+	case float64:
+		return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(v)).UTC()}, nil
+	case string:
+		if expiresOn, err := strconv.Atoi(v); err == nil {
+			return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Now().Add(time.Second * time.Duration(expiresOn)).UTC()}, nil
```
jhendrixMSFT, line 135: Since we're touching this, can you please also fix this to properly handle Unix time stamp.  See https://github.com/Azure/go-autorest/issues/638 for details.

sdk/azidentity/managed_identity_client.go
catalinaperalta: That's what I read in the docs: https://docs.microsoft.com/en-us/azure/service-fabric/how-to-managed-identity-service-fabric-app-code#acquiring-an-access-token-using-rest-api

jhendrixMSFT: This is also a Unix timestamp, similar to the other issue I linked.  The key phrase is `represented as the number of seconds from "1970-01-01T0:0:0Z UTC"`

catalinaperalta: Ok yeah I'll read through the issue you linked and see what needs to change 👍

sdk/azidentity/managed_identity_client.go
catalinaperalta: Done...let me know if that's good...I normalized it all to UTC since that's what we were doing before.

catalinaperalta: Actually not done...I saw we're going the same thing for the timestamps....I'll get that fixed right now

catalinaperalta: Ok NO..haha just got mixed up for a second there...those are fine as they are. We're good to go now :P

catalinaperalta: Although when I was debugging in service fabric I saw that the value it returns seems to be something like this: 1.623886173e+09. It falls under the float case....do you know if the the time package handles this properly? Or is some additional parsing necessary?

jhendrixMSFT: I don't think this is a float, it's just a large number expressed in scientific notation.  Presumably it's the number of seconds from the Unix epoch, thus should be passed as the first arg to [`time.Unix()`](https://pkg.go.dev/time#Unix)

catalinaperalta: I agree that it's not a float, but in the switch case it falls into the float case....but I think you're saying it's fine as I put it then?

catalinaperalta: I'm referring to the int64() cast that time.Unix() expects

sdk/azidentity/managed_identity_client.go
```diff
@@ -109,7 +112,7 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
 		Token        string        `json:"access_token,omitempty"`
 		RefreshToken string        `json:"refresh_token,omitempty"`
 		ExpiresIn    wrappedNumber `json:"expires_in,omitempty"` // this field should always return the number of seconds for which a token is valid
-		ExpiresOn    string        `json:"expires_on,omitempty"` // the value returned in this field varies between a number and a date string
+		ExpiresOn    interface{}   `json:"expires_on,omitempty"` // the value returned in this field varies between a number and a date string
```
jhendrixMSFT, line 115: This should probably be a [`json.Number`](https://pkg.go.dev/encoding/json#Number).

sdk/azidentity/managed_identity_client.go
jhendrixMSFT: What is actually sent over the wire?  I suspect this is an artifact of modeling the field as an `interface{}`.

catalinaperalta: I expect it's a pretty large integer that is translated into scientific notation due to it's size. I can spin up the SF cluster again and see if I can get the raw response but it will take some time. I'll also use that to check how the large value is handled with json.Number.

catalinaperalta: So during my testing with json.Number I recalled why I didnt use it, it doesnt accept timestamps only numbers I believe.

catalinaperalta: I looked into this a little, after go 1.14 json.Number is checked a little more strictly during unmarshal...I think the right thing to do is unmarshal into the interface.

jhendrixMSFT: Ah yes, I do recall something about this being more strict now.

I suppose it's coming back as a `float64` due to this being JSON.

Does this come across the wire as a string?  I ask because regarding your question about losing any significant values, here's what `json.Number` does https://golang.org/src/encoding/json/decode.go?s=7375:7413#L189

catalinaperalta: The cluster keeps stalling right after the GetToken call so havent been able to access the raw value and it's a bit slow to restart the nodes, will try to restart it all again. 
It is definitely not coming back as a string, that's why I had to change the value from string to interface{}, since json.Unmarshal was returning an error about not being able to unmarshal a number into a string.

catalinaperalta: So it returns a value like this: `"expires_on":1624136297`.....that doesnt seem so big to me for it to be parsed into scientific notation....

catalinaperalta: Removed int from the switch case since the docs say that when unmarshalling into an interface json.Numbers will be unmarshalled into a float64.

sdk/azidentity/managed_identity_client.go
```diff
@@ -121,19 +124,26 @@ func (c *managedIdentityClient) createAccessToken(res *azcore.Response) (*azcore
...
+		// this is the case when expires_on is a time string
+		// this is the format of the string coming from the service
+		if expiresOn, err := time.Parse("1/2/2006 15:04:05 PM +00:00", v); err == nil { // the date string specified is for Windows OS
+			eo := expiresOn.UTC()
+			return &azcore.AccessToken{Token: value.Token, ExpiresOn: eo}, nil
+		} else if expiresOn, err := time.Parse("1/2/2006 15:04:05 +00:00", v); err == nil { // the date string specified is for Linux OS
+			eo := expiresOn.UTC()
+			return &azcore.AccessToken{Token: value.Token, ExpiresOn: eo}, nil
+		} else {
+			return nil, err
```
jhendrixMSFT, line 143: Can you please add one more test that handles the failure case?

catalinaperalta: Done

jhendrixMSFT (approved):

PR #14872, opened 2021-06-22
azidentity: Add environment variable for resource ID

### LINKED ISSUE #14742: Add support for a resource ID env var
Add an environment variable for the new resource ID support in mananged identity credential. Make sure it doesnt conflict with other known env vars.

catalinaperalta:
Fixes #14742

### DIFF
--- a/sdk/azidentity/managed_identity_credential.go
+++ b/sdk/azidentity/managed_identity_credential.go
@@ -72,7 +72,7 @@ func NewManagedIdentityCredential(id string, options *ManagedIdentityCredentialO
 	client.msiType = msiType
 	// check if no clientID is specified then check if it exists in an environment variable
 	if len(id) == 0 {
-		if options.ID == 1 {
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_credential.go
```diff
@@ -72,7 +72,11 @@ func NewManagedIdentityCredential(id string, options *ManagedIdentityCredentialO
 	client.msiType = msiType
 	// check if no clientID is specified then check if it exists in an environment variable
 	if len(id) == 0 {
-		id = os.Getenv("AZURE_CLIENT_ID")
+		if options.ID == 1 {
```
jhendrixMSFT, line 75: Don't we have constants instead of the magic `1` here?

catalinaperalta: Yes, done.

jhendrixMSFT (approved):

PR #14879, opened 2021-06-22
Add tenantID on TokenRequestOptions

catalinaperalta:
Necessary for multi-tenant auth scenarios.

### DIFF
--- a/sdk/azcore/credential.go
+++ b/sdk/azcore/credential.go
@@ -50,4 +50,6 @@ type AccessToken struct {
 type TokenRequestOptions struct {
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
+	// TenantID contains the tenant ID to use in a multi-tenant authentication scenario.
+	TenantID string
 }
--- a/sdk/azcore/credential.go
+++ b/sdk/azcore/credential.go
@@ -50,6 +50,7 @@ type AccessToken struct {
 type TokenRequestOptions struct {
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
-	// TenantID contains the tenant ID to use in a multi-tenant authentication scenario.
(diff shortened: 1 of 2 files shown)


sdk/azcore/credential.go
```diff
@@ -50,4 +50,6 @@ type AccessToken struct {
 type TokenRequestOptions struct {
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
+	// TenantID contains the tenant ID to use in a multi-tenant authentication scenario.
+	TenantID string
```
jhendrixMSFT, line 54: This isn't used at present, is that intentional?

Also, should the comment be updated to include a note that this replaces the tenant ID used to create the credential when set?

catalinaperalta: Yes, adding this first to tag this version and reference it in the multi-tenant work. If not the multi-tenant PR would fail tests since this field doesnt exist on the options. I'll update the comment.

jhendrixMSFT: But isn't this used in the bearer token policy which is in `azidentity`?

catalinaperalta: I dont follow....this field didnt exist before now. 🤔

catalinaperalta: For what it's worth this will also be used when adding support for multi-tenant auth challenges used in services like keyvault

catalinaperalta: Here it is in the .NET implementation: https://github.com/Azure/azure-sdk-for-net/blob/42d70236d406451f0323fa4de31ee1dab403d2aa/sdk/core/Azure.Core/src/TokenRequestContext.cs#L50

catalinaperalta: Resolving this since we discussed this offline.

jhendrixMSFT (approved):

PR #14889, opened 2021-06-23
Add bearer token policy to azcore

catalinaperalta:
Part of the work necessary for the multitenant authentication solution.


catalinaperalta:
This work is being completed in https://github.com/Azure/azure-sdk-for-go/pull/14929

PR #14890, opened 2021-06-23
azidentity: bearer token policy update

catalinaperalta:
Update credential code to use the token refresh policy in azcore. 

NOTE: this PR relies on first merging the token refresh  policy changes into azcore for tests to pass.


catalinaperalta:
@jhendrixMSFT this will be broken until we tag the changes in azcore...but you can see what the defaultTokenProcessor will do in azidentity.

catalinaperalta:
This was redesigned and added in a separate PR.

PR #14926, opened 2021-06-29
introduce pipelines to the modules in tools directory

ArcturusZhang:


benbp:
This may surface some issues down the line as the `sdk` directory is a big assumption across our tools/scripts and is hardcoded in many places (for example, the pipeline generator I just kicked off doesn't work since it only scans `sdk`). We may have to deal with some of this stuff on a case by case basis.

benbp (approved):

PR #14929, opened 2021-06-29
Multi-tenant authentication support

catalinaperalta:
Adding a policy in azcore that will allow customization for multi-tenant support in ARM.


sdk/azcore/credential.go
```diff
@@ -16,6 +16,9 @@ type AuthenticationPolicyOptions struct {
 	// the list of OAuth2 authentication scopes used when requesting a token.
 	// This field is ignored for other forms of authentication (e.g. shared key).
 	Options TokenRequestOptions
+	// TokenProcessor is used to provide a custom implementation for checking the expires_on
+	// time of the token, when tokens should be refreshed, and setting custom headers.
+	TokenProcessor BearerTokenProcessor
```
jhendrixMSFT, line 21: Can this be removed from here and passed directly to `NewBearerTokenPolicy()` instead?

catalinaperalta: Yeah I was torn about where to put it. I'm fine with moving it onto the ctor

catalinaperalta: Done...let me know how it's looking. The `defaultTokenProcessor` is what would exist in azidentity.

sdk/azcore/policy_bearer_token.go
```diff
@@ -0,0 +1,144 @@
...
+	cond *sync.Cond
+
+	// renewing indicates that the token is in the process of being refreshed
+	renewing bool
+
+	// header contains the authorization header value
+	header string
+
+	// expiresOn is when the token will expire
+	expiresOn map[string]time.Time
```
jhendrixMSFT, line 46: I think this, and `header` above should be moved into their specific policies.

catalinaperalta: I think moving expiresOn works well if we move it to the specific policy...but the header is used in the initial conditional on the refresh policy...so I think that one should stay as is. I'll work on removing the expiresOn from the policy.

catalinaperalta: Done

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,134 @@
...
+	options     TokenRequestOptions
+	implementer TokenRefreshProcessor
+}
+
+// NewTokenRefreshPolicy instantiates a thread safe token refresh policy.
+// Pass in the credential to use, along with a TokenProcessor that will be
+// used to provide a custom implementation for checking the expires_on
+// time of the token, when tokens should be refreshed, and setting custom headers.
+// Additionally specify options for the token to be retreived.
+func NewTokenRefreshPolicy(cred TokenCredential, p TokenRefreshProcessor, opts AuthenticationPolicyOptions) Policy {
```
catalinaperalta, line 58: I've renamed everything to TokenRefreshPolicy. Let me know if this is looking good enough to push the right changes to the ARM branch

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+)
+
+const (
+	bearerTokenPrefix = "Bearer "
+)
+
+// DO NOT USE. This is for internal SDK usage. The TokenRefreshProcessor is used to provide
+// custom implementations for the bearer token policy.
+type TokenRefreshProcessor interface {
```
jhendrixMSFT, line 20: How about `TokenRefresher`?

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,134 @@
...
+	// IsZeroOrExpired returns a bool if there is a tenant that needs to be initialized, it also
+	// returns a list of tenants to get a token for
+	IsZeroOrExpired(map[string]time.Time) (bool, []string)
+	// ShouldRefresh returns a bool is a token needs to be refreshed and the list of tenants that
+	// need to be refreshed
+	ShouldRefresh(map[string]time.Time) (bool, []string)
+	// GetToken performs the GetToken call and returns the header needed for authentication
+	GetToken(ctx context.Context, cred TokenCredential, tenants []string, opts TokenRequestOptions) (string, error)
```
jhendrixMSFT, line 28: How about we call this `Refresh()` to match `TokenRefresher`.  It's really not a singular token for the multi-tenant case anyways.

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,134 @@
...
+		implementer: p,
+	}
+}
+
+func (b *tokenRefreshPolicy) Do(req *Request) (*Response, error) {
+	if req.URL.Scheme != "https" {
+		// HTTPS must be used, otherwise the tokens are at the risk of being exposed
+		return nil, errors.New("token credentials require a URL using the HTTPS protocol scheme")
+	}
+	// create a "refresh window" before the token's real expiration date.
```
jhendrixMSFT, line 72: I think this comment no longer applies to this generic implementation?

sdk/azcore/policy_token_refresh.go
catalinaperalta: Yeah I like it...will change it

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,134 @@
...
+	// this allows callers to continue to use the old token while the
+	// refresh is in progress.
+	getToken, header, auxTenants := false, "", []string{}
```
jhendrixMSFT, line 75: Hoping we don't need `auxTenants` here, seems like that should be applicable to the multi-tenant policy only.

sdk/azcore/policy_token_refresh.go
catalinaperalta: Refresh sounds good to me...but maybe should it be `RefreshOrGet()`? Since it isnt always a refresh? First it initializes it

sdk/azcore/policy_token_refresh.go
catalinaperalta: `GetToken` takes in a slice of tenants to support the multi-tenant scenario....I could simply call the variable `tenants` or we could add some options type....but that will be another type to export in azcore..

sdk/azcore/policy_token_refresh.go
catalinaperalta: Done

sdk/azcore/policy_token_refresh.go
catalinaperalta: Removed.

catalinaperalta:
FYI moved the azidentity and armcore pieces since they will need to use the newly tagged version of azcore after this PR is merged.

sdk/azcore/policy_token_refresh.go
jhendrixMSFT: I think it likely depends on your perspective (FWIW we have a similar pattern in `go-autorest`).  Maybe it would be better to call the interface `TokenFetcher` and make this method `Fetch()`?

sdk/azcore/policy_token_refresh.go
jhendrixMSFT: Why does it even need to be here though, can't this be something that's part of the multi-tenant implementation?

sdk/azcore/policy_token_refresh.go
catalinaperalta: Yeah `TokenFetcher` and `Fetch` work...I'll update it. I do think all of these additions to public surface area should be approved by Jeff in a review eventually but this work for me.

sdk/azcore/policy_token_refresh.go
catalinaperalta: You're right, I was leaving it there to not carry the data on the processor in ARM...but that just puts the burden on everybody who uses this policy. Will remove it.

sdk/azcore/policy_token_refresh.go
catalinaperalta: Done

sdk/azcore/policy_token_refresh.go
catalinaperalta: Done

jhendrixMSFT:
Actually I think you could place this in the `internal` module.  That way we don't have to worry about the whole "public surface area" debate.  What do you think?

catalinaperalta:
> Actually I think you could place this in the internal module. That way we don't have to worry about the whole "public surface area" debate. What do you think?

I just have a question about what happens if the internal module depends on azcore? (Are there any circular references that would happen if so??)

jhendrixMSFT:
Hmm you're right, I hadn't thought about that.  While modules does fix this, it's a bit of a code smell and should be avoided.

If we wanted to move this to `internal`, we'd have to move `TokenCredential` and `TokenRequestOptions` to `internal` as well, then type-alias them in `azcore` (we did a similar thing in msal/go).  Other option is to just do what you're doing now.  Probably worth a discussion in scrum today.

catalinaperalta:
>If we wanted to move this to internal, we'd have to move TokenCredential and TokenRequestOptions to internal as well, then type-alias them in azcore (we did a similar thing in msal/go). Other option is to just do what you're doing now. Probably worth a discussion in scrum today.

I think that would work....but the problem is that we also use azcore request and response, as well as the authentication options and the policy interface...I think it would be a little more complicated than that.

jhendrixMSFT:
Agreed, it should remain in `azcore`.

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,127 @@
...
+	cond *sync.Cond
+
+	// renewing indicates that the token is in the process of being refreshed
+	renewing bool
+
+	// header contains the authorization header value
+	header string
+
+	// expiresOn is when the token will expire
+	expiresOn map[string]time.Time
```
jhendrixMSFT, line 43: Can't this also be removed and tracked in the implementing types?  We only pass it to `implementer`.

catalinaperalta: Yes done

catalinaperalta:
@JeffreyRichter could we get a quick review on the additions to public surface area please?

jhendrixMSFT:
@JeffreyRichter for context, this is making the refresh logic generic so it can be shared with single and multi-tenant authentication policies.

sdk/azcore/policy_token_refresh.go
```diff
@@ -0,0 +1,124 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	bearerTokenPrefix = "Bearer "
+)
+
+// DO NOT USE. This is for internal SDK usage. The TokenFetcher is used to provide
+// custom implementations for the bearer token policy.
```
jhendrixMSFT, line 19: "custom implementations for the token refresh policy."

catalinaperalta: Done

catalinaperalta:
Redesigned and updated in a separate PR.

PR #14960, opened 2021-07-01
[Tables] Add post-request validation of entity

seankane-msft:
Adding a check on error of create entity that generates a more helpful error if the user submits an entity without both `PartitionKey` and `RowKey`.

Taking the low hanging fruit to get started 😄


sdk/tables/aztable/errors.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package aztable
+
+import "errors"
+
+func checkEntityForPkRk(entity *map[string]interface{}, err error) error {
+	if _, ok := (*entity)["PartitionKey"]; !ok {
+		return errors.New("Entity must have a PartitionKey and RowKey")
```
christothes, line 10: Should we make this error a const in `zc_tableConstants.go`?

christothes (approved):

sdk/tables/aztable/table_client_test.go
```diff
@@ -458,6 +458,22 @@ func (s *tableClientLiveTests) TestBatchError() {
...
+
+	badEntity := &map[string]interface{}{
+		"Value":  10,
+		"String": "stringystring",
+	}
+
+	_, err := client.AddEntity(ctx, *badEntity)
+
+	assert.NotNil(err)
+	assert.Contains(err.Error(), "Entity must have a PartitionKey and RowKey")
```
christothes, line 474: the test can use the const now also.

christothes (approved):

PR #14961, opened 2021-07-01
[Tables] changing QueryOptions param to be a ptr

seankane-msft:


sdk/tables/aztable/table_client.go
```diff
@@ -55,14 +55,14 @@ func (t *TableClient) Delete(ctx context.Context) (TableDeleteResponse, error) {
 //
 // Query returns a Pager, which allows iteration through each page of results. Example:
 //
-// pager := client.Query(QueryOptions{})
+// pager := client.Query(&QueryOptions{})
```
catalinaperalta, line 58: I think this can be `nil` now.

sdk/tables/aztable/table_service_client.go
```diff
@@ -64,14 +64,14 @@ func (t *TableServiceClient) Delete(ctx context.Context, name string) (TableDele
 //
 // Query returns a Pager, which allows iteration through each page of results. Example:
 //
-// pager := client.Query(QueryOptions{})
+// pager := client.Query(&QueryOptions{})
```
catalinaperalta, line 67: Same `nil` change.

sdk/tables/aztable/zt_table_recorded_tests.go
```diff
@@ -97,7 +97,7 @@ func recordedTestTeardown(key string) {
 func cleanupTables(context *testContext, tables *[]string) {
 	c := context.client
 	if tables == nil {
-		pager := c.Query(QueryOptions{})
+		pager := c.Query(&QueryOptions{})
```
catalinaperalta, line 100: Switch to `nil`.

sdk/tables/aztable/table_client.go
```diff
@@ -55,14 +55,14 @@ func (t *TableClient) Delete(ctx context.Context) (TableDeleteResponse, error) {
...
 // for pager.NextPage(ctx) {
 //     resp = pager.PageResponse()
 //     fmt.sprintf("The page contains %i results", len(resp.TableEntityQueryResponse.Value))
 // }
 // err := pager.Err()
-func (t *TableClient) Query(queryOptions QueryOptions) TableEntityQueryResponsePager {
-	return &tableEntityQueryResponsePager{tableClient: t, queryOptions: &queryOptions, tableQueryOptions: &TableQueryEntitiesOptions{}}
+func (t *TableClient) Query(queryOptions *QueryOptions) TableEntityQueryResponsePager {
+	return &tableEntityQueryResponsePager{tableClient: t, queryOptions: queryOptions, tableQueryOptions: &TableQueryEntitiesOptions{}}
```
catalinaperalta, line 65: this is where you can put the `if queryOptions == nil` check.

catalinaperalta (commented):
Looking good, just switch to using `nil` when there are no options to be configured. Then in the code you can initialize an empty version of the options to be used to not cause a panic. 

Initialize the options internally like this in the ctor (example taken from azidentity): 
```
        if options == nil {
		options = &ClientSecretCredentialOptions{}
	}
```

PR #15003, opened 2021-07-08
[Tables] Add support for Entity struct

seankane-msft:
- [ ] Add `Entity` object for embedding in user entities
- [ ] Add method for serializing user entities into `map[string]interface{}`
- [ ] Add capability for deserializing user entities.


JeffreyRichter (approved):

sdk/tables/aztable/recordings/TestTableClient_Cosmos/TestAddAnnotatedEntity-variables.yaml
```diff
@@ -1,5 +1,5 @@
 ---
 COSMOS_TABLES_ENDPOINT_SUFFIX: cosmos.azure.com
-TABLES_COSMOS_ACCOUNT_NAME: chrisstablesprim
+TABLES_COSMOS_ACCOUNT_NAME: seankaneprim
 TABLES_PRIMARY_COSMOS_ACCOUNT_KEY: Kg==
-randomSeed: "1621634379"
```
richardpark-msft, line 5: Why does this end up changing, btw? Is it auto-generated as well?

seankane-msft: I had to re-record these tests

seankane-msft: The random seed used was different than when Chris originally ran it, I re-record by completely wiping the recordings files and then re-running tests.

christothes: Just for additional context (in case you didn't already know this), the randomSeed is a recorded value that is used by tests in playback mode so that any randomly generated recorded values such as names or UUIDs are generated consistently.

sdk/tables/aztable/byte_array_response.go
```diff
@@ -0,0 +1,129 @@
...
+	ClientRequestID *string
+
+	// ContentType contains the information returned from the Content-Type header response.
+	ContentType *string
+
+	// Date contains the information returned from the Date header response.
+	Date *time.Time
+
+	// ETag contains the information returned from the ETag header response.
+	ETag *string
```
christothes, line 24: for future - in some of the other languages ETag is a type in Core. I wonder if we should do that eventually here.

sdk/tables/aztable/byte_array_response.go
```diff
@@ -0,0 +1,129 @@
...
+package aztable
+
+import (
+	"encoding/json"
+	"net/http"
+	"time"
+)
+
+// ByteArrayResponse converts the MapOfInterfaceResponse.Value from a map[string]interface{} to a []byte
+type ByteArrayResponse struct {
```
christothes, line 13: Is this an interim step to do this as part of generated code, or do we plan to always do this conversion?

sdk/tables/aztable/entity.go
```diff
@@ -0,0 +1,186 @@
...
+			if propertyEdmTypeRawValue, ok := entity[propName+"@odata.type"]; ok {
+				if err = json.Unmarshal(propertyEdmTypeRawValue, &propertyEdmTypeValue); err != nil {
+					return
+				}
+			}
+
+			var propValue interface{} = nil
+			switch propertyEdmTypeValue {
+			case "": // "<property>@odata.type" doesn't exist, infer the EDM type from the JSON type
+				// Try to unmarshal this property value as an int32 first
```
christothes, line 97: I wonder if we should only try the int32 fallback if it deserialized as int64. My guess is that string will be more common.

sdk/tables/aztable/entity_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func (s *tableClientLiveTests) TestAddBasicEntity() {
```
christothes, line 14: Just curious if you benchmarked this against the current implementation.

sdk/tables/aztable/table_client.go
```diff
@@ -69,70 +70,97 @@ func (t *TableClient) Query(queryOptions *QueryOptions) TableEntityQueryResponse
...
 	return nil, errors.New("Invalid TableUpdateMode")
 }
 
 // UpsertEntity replaces the specified table entity if it exists or creates the entity if it does not exist.
 // If the entity exists and updateMode is Merge, the property values present in the specified entity will be merged with the existing entity rather than replaced.
 // The response type will be TableEntityMergeResponse if updateMode is Merge and TableEntityUpdateResponse if updateMode is Replace.
-func (t *TableClient) UpsertEntity(ctx context.Context, entity map[string]interface{}, updateMode TableUpdateMode) (interface{}, error) {
-	pk := entity[partitionKey].(string)
-	rk := entity[rowKey].(string)
+func (t *TableClient) UpsertEntity(ctx context.Context, entity []byte, updateMode TableUpdateMode) (interface{}, error) {
```
christothes, line 146: does the type change for entity mean that developers always have to marshal their entity before they pass it to us? Could we do it for them?

sdk/tables/aztable/table_client_test.go
```diff
@@ -46,435 +44,271 @@ func (s *tableClientLiveTests) TestServiceErrors() {
...
 
-	_, updateErr := client.UpsertEntity(ctx, mergeProperty, Replace)
-	require.Nil(updateErr)
+	// 2. Query for basic Entity
+	preMerge, err := client.GetEntity(ctx, entityToCreate.PartitionKey, entityToCreate.RowKey, nil)
+	require.NoError(err)
+
+	var unMarshalledPreMerge map[string]interface{}
+	err = json.Unmarshal(preMerge.Value, &unMarshalledPreMerge)
+	require.NoError(err)
```
christothes, line 178: Is how developers would have to handle entity responses, or is this just for testing purposes?

sdk/tables/aztable/byte_array_response.go
JeffreyRichter: Yes, I agree. azcore should have an etag type. @jhendrixMSFT

sdk/tables/aztable/entity.go
JeffreyRichter: The service sends int64 values as strings; not as numbers. If the @odata.type:Edm.Int64 is present then we have to convert a string to an int64 integer. If the Edm.Int64 is missing that an int64 will be a string. A number will either be int32 or float64. In other words, I think this is right as is.

sdk/tables/aztable/table_client.go
JeffreyRichter: I don't think it right to have asymmetrical behavior where customers MUST unmarshal themself and we marshal for them. I'm not aware of any Go package that does this so the symmetry is prefered.

sdk/tables/aztable/table_client_test.go
JeffreyRichter: There are MANY ways customers can handle entity responses. They unmarhsal to a map[string]interface or to a map[string]json.RawMessage, or to an EdmEntity (our own type) or to a structure that the customer defines. We enable all this flexibility by returning the raw JSON and then the customer decides how they want to deal with it.

sdk/tables/aztable/entity.go
christothes: My assumption was that this is to handle the default behavior of unmarshal, which makes all numbers Int64.

JeffreyRichter: json.Unmarshal converts all numbers to float64 by default.

christothes: right - that is what I assuming this logic was meant to deal with, so that we don't up-convert int32 values to Int64 on unmarshal.

sdk/tables/aztable/entity_test.go
seankane-msft: No I have not yet, but I will add an issue to add benchmarking in the future. [issue](https://github.com/Azure/azure-sdk-for-go/issues/15152).

sdk/tables/aztable/byte_array_response.go
seankane-msft: https://github.com/Azure/azure-sdk-for-go/issues/15153

sdk/tables/aztable/byte_array_response.go
seankane-msft: We always plan to do this conversion. This change gives the customer the abiltity to use the `json.Unmarshal` functionality on recieved entities.

Issue #15033, opened 2021-07-12
Make linter CI step fail builds when applicable


jhendrixMSFT:
The current track 2 CI linter doesn't fail the build if linting fails.  I believe it was originally implemented like this because the repo wasn't lint-free.  We need to fix up the current track 2 bits and enable the linter.

PR #15035, opened 2021-07-12
[EngSys] Stage fails if linter does not succeed

seankane-msft:
On the lint step of `Analyze`, if the exit code is not 0 this step will fail.


seankane-msft:
closes #15033

seankane-msft:
@jhendrixMSFT I was able to confirm that this change will fail a pipeline step if lint does not pass. Tested it on [tables here](https://github.com/Azure/azure-sdk-for-go/pull/15084).

jhendrixMSFT (approved):

PR #15038, opened 2021-07-13
Unexport HTTP headers to reduce public surface area

jhendrixMSFT:


richardpark-msft (approved):
Looks good - private and a much smaller subset. Can't ask for much more :)

seankane-msft (approved):

catalinaperalta (approved):
LGTM. Although when I saw this, I thought maybe this could be put in the internal module so that the other libs could still use them and avoid declaring the same consts in different mods. Not sure if there would be any problem with that though. Also maybe it's too much work just for the header consts and the the benefit is really small.

jhendrixMSFT:
It's not a bad idea, my only complaint is that it gets in the way.  i.e. when I'm doing some work in `azcore` I need to first add my constant(s) to `internal`, PR and commit, then update the dependency and move on.

Since this is all internal, if this does prove to be problematic in the future we can update it later without any impact to customers.

catalinaperalta:
I see your point. If it's going to slow things down, then it's better this way for now. `azidentity` will need to be updated to not depend on these exported headers then, when updating to the latest `azcore`.

jhendrixMSFT:
Agreed, and I think there will be a few more breaking changes from last week's review of `azcore` so we don't have to tag this right now.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/15047

PR #15076, opened 2021-07-15
[EngSys] adding coverage tools

seankane-msft:
Adds to coverage tests to fail if the code coverage is below a supplied number


chamons (commented):
Just wondering if there is a reason we're writing the tool in go? It seems to be searching for files named coverage.xml, running a regular expression over them, and parsing the line.

I feel like this could be down by a powershell script, but if we're extending it to be more complex in the future that makes a bit more sense.

seankane-msft:
> Just wondering if there is a reason we're writing the tool in go? It seems to be searching for files named coverage.xml, running a regular expression over them, and parsing the line.
> 
> I feel like this could be down by a powershell script, but if we're extending it to be more complex in the future that makes a bit more sense.

I am more familiar with using Go to do this than a Powershell script. In the future Rick and I discussed adding branch coverage and this tool could be expanded to do so.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -1,5 +1,6 @@
 parameters:
   ServiceDirectory: ''
+  CoverageGoal: 0.95
```
benbp, line 3: As the coverage go tool gains complexity/scope, I think it would be good to follow a convention of an SDK directory config file to store this type of info. As it is in this PR, you can't run the coverage locally to see if it will pass upstream without knowing to manually look up the CoverageGoal in ci.yml. It gets even harder if you want to run coverage for the whole repo (e.g. for cross-cutting changes).

benbp: My thinking in general for go is that we want to put as little context in the yaml as possible. In an ideal world the devops yaml would only function as a light orchestration wrapper and not contain any sdk-specific context.

seankane-msft: Is there an example of another repo that has a config file? Or to store this in a single config file and specifying the information per library?

benbp: I would say it's an open question, perhaps just a config.json file for easy interoperability? I know js uses the convention of putting custom key/value config data into the `package.json` file (e.g. for [samples configuration](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/eventhub/event-hubs/package.json#L84-L92)). I don't think there are as many examples in other repos.

If you think it's weird/unnecessary, I wouldn't disagree as right now there would be only tool that uses it. As we add more tooling that needs to have per-sdk behavior it could be helpful to have a common place for this kind of stuff that is amenable to both local and pipeline scripts.

seankane-msft: @benbp I added a config.json file in the `eng` directory. Right now it just has a list of packages and their code coverage goal but I think this will be a good way to expand any needs for it in the future.

eng/scripts/create_coverage.ps1
```diff
@@ -0,0 +1,17 @@
+#Requires -Version 7.0
+
+$coverageFiles = [Collections.Generic.List[String]]@()
+Get-Childitem -recurse -path $(SCOPE) -filter coverage.txt | foreach-object {
```
benbp: nit: I don't care much for the powershell casing style, but I'd pick one to be consistent (e.g. capitalize ForEach-Object). Also it's `Get-ChildItem`.

tools/internal/coverage/main.go
```diff
@@ -39,15 +50,45 @@ func FindCoverageFiles(p string) {
...
+
+// This supports doing a single package at a time. If this needs to be expanded in the future
+// this method will have to return a []*float64 for each packages goal
+func findCoverageGoal(covFiles []string, configData *CodeCoverage) *float64 {
+	var ret float64
+
+	for _, covFile := range covFiles {
+		for _, p := range configData.Packages {
+			if strings.Contains(covFile, p.Name) {
+				return &p.CoverageGoal
```
benbp, line 75: Is there a reason to use a pointer to a float here and not just a float? It looks like you de-reference it later on anyway.

eng/scripts/create_coverage.ps1
seankane-msft: I'll change to the consistent uppercase style, didn't notice this when I copied it over from the yml

tools/internal/coverage/main.go
seankane-msft: When I was passing it in as a flag the parsed flag was a pointer to a float so probably for consistency, but I guess in hindsight I'll say I was extra concerned about memory management 😆

richardpark-msft: Interestingly, I think the pointer and float64 end up being the same size in this case. So the pointer is actually less efficient (requires the pointer _and_ the float64!).

seankane-msft: I have been foiled again! @richardpark-msft

eng/pipelines/templates/jobs/archetype-sdk-client.yml
benbp: I like it!

tools/internal/coverage/main.go
benbp: I've found in some cases with the Azure SDKs (I forget why, perhaps because of auto-generated openAPI stuff and/or the flags as you say), there is a very specific reason that pointers needed to be used for basic types. But in all other cases, I think it's unnecessary.

benbp: Also I wouldn't be as concerned about memory management/pre-mature optimization for tools code as opposed to customer-facing code.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
benbp: If it's going to be in eng and contain all package info, it can probably be a dedicated file called `coverage.json` instead?

tools/internal/coverage/main.go
```diff
@@ -0,0 +1,137 @@
...
+	coverageValues := make([]float64, 0)
+	for _, coverageFile := range coverageFiles {
+		xmlFile, err := os.Open(coverageFile)
+		check(err)
+		defer xmlFile.Close()
+
+		byteValue, err := ioutil.ReadAll(xmlFile)
+		check(err)
+
+		re := regexp.MustCompile(`<coverage line-rate=\"\d.\d+\"`)
```
benbp, line 102: this `.` might actually match another digit, so if you want to match `.` an escape might be a little safer.

```suggestion
		re := regexp.MustCompile(`<coverage line-rate=\"\d\.\d+\"`)
```

benbp (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
seankane-msft: I went  with config.json instead because it may contain more info in the future other than coverage information.

Issue #15007, opened 2021-07-08
[Tables] List methods


seankane-msft:
- [ ] `ListTables`
- [ ] `ListEntities`

PR #15099, opened 2021-07-20
[Tables] change Query to List

seankane-msft:
Adds a `List` method to both the TableClient (for listing entities) and TableServiceClient (for listing tables)


sdk/tables/aztable/table_client.go
```diff
@@ -58,14 +58,35 @@ func (t *TableClient) Delete(ctx context.Context) (TableDeleteResponse, error) {
...
+//
+// List returns a Pager, which allows iteration through each page of results. Example:
+//
+// pager := client.List()
+// for pager.NextPage(ctx) {
+//     resp = pager.PageResponse()
+//     fmt.Fprintf("The page contains %i results", len(resp.TableEntityQueryResponse.Value))
+// }
+// err := pager.Err()
+func (t *TableClient) List() TableEntityQueryResponsePager {
```
christothes, line 84: What's the difference between `List` and `Query`?

sdk/tables/aztable/table_client_test.go
```diff
@@ -477,6 +478,43 @@ func (s *tableClientLiveTests) TestInvalidEntity() {
...
+			require.Equal(e["PartitionKey"].(string), "partition")
+			count += 1
+		}
+	}
+	err = pager.Err()
+	require.NoError(err)
+	require.Equal(5, count)
+}
+
+func insertNEntities(pk string, n int, client *TableClient) error {
```
christothes, line 503: nit: Should this go in `zt_table_recorded_tests` with all the other utility functions?

christothes (approved):

sdk/tables/aztable/table_client.go
seankane-msft: List is just an empty query. In Python we have both even though both can be done with just `Query`. I didn't see anything in the Go guidelines about not including a List so I have it here. I'll talk to Jeffrey about what our language guideline is.

seankane-msft:
closes #15007

sdk/tables/aztable/table_client.go
christothes: Ok cool. BTW, if we leave it, should the implementation just be:
```go
{
 return t.Query();
```

sdk/tables/aztable/table_service_client.go
```diff
@@ -68,13 +68,24 @@ func (t *TableServiceClient) Delete(ctx context.Context, name string) (TableDele
 // pager := client.Query(nil)
 // for pager.NextPage(ctx) {
 //     resp = pager.PageResponse()
-//     fmt.sprintf("The page contains %i results", len(resp.TableQueryResponse.Value))
+//     fmt.Fprintf("The page contains %i results", len(resp.TableQueryResponse.Value))
```
richardpark-msft, line 71: Thorough!

richardpark-msft: Although....shouldn't it just be fmt.Printf?

catalinaperalta: Butting in to your thread 😝 It should be `fmt.Sprintf(..)`

seankane-msft: I'm going to go with `fmt.Printf` because the goal is just to print to stdout the results.

richardpark-msft: @catalinaperalta  - is this because of the kate-a-line-UH thing? I have already apologized for that debacle.

catalinaperalta: That may have been a contributing factor...

sdk/tables/aztable/table_client.go
```diff
@@ -53,19 +53,20 @@ func (t *TableClient) Delete(ctx context.Context) (TableDeleteResponse, error) {
 // Top: The maximum number of entities that will be returned per page of results.
 // Note: This value does not limit the total number of results if NextPage is called on the returned Pager until it returns false.
 //
-// Query returns a Pager, which allows iteration through each page of results. Example:
+// List returns a Pager, which allows iteration through each page of results. Example:
 //
-// pager := client.Query(nil)
+// options := &ListOptions{Filter: to.StringPtr("PartitionKey eq 'pk001'"), Top: to.Int32Ptr(25), Select: to.StringPtr("PartitionKey,RowKey,Value,Price")}
```
JeffreyRichter, line 58: I think PK and RK are always returned no matter what and so you may want to document this and NOT show them in the doc comment.

sdk/tables/aztable/models.go
```diff
@@ -0,0 +1,29 @@
...
+	Format *OdataMetadataFormat
+	// Select expression using OData notation. Limits the columns on each record to just those requested, e.g. "$select=PolicyAssignmentId, ResourceId".
+	Select *string
+	// Maximum number of records to return.
+	Top *int32
+}
+
+func (l *ListOptions) toQueryOptions() *QueryOptions {
+	if l == nil {
+		return &QueryOptions{}
```
JeffreyRichter, line 20: Can we just get rid of QueryOptions entirely and just use ListOptions?
Is QueryOptions code generated?

JeffreyRichter (approved):

sdk/tables/aztable/models.go
seankane-msft: QueryOptions is generated code, in the future we can look at changing the name of QueryOptions post generation and get rid of the one in models.go

sdk/tables/aztable/table_client.go
seankane-msft: I dont believe this is true, [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/tables/azure-data-tables/tests/recordings/test_table_entity.test_get_entity_with_select.yaml#L108-L153) is a recording for the python suite that shows a response without the pk/rk. For reference, [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/tables/azure-data-tables/tests/test_table_entity.py#L618-L636) is the test as well.

JeffreyRichter: OK, maybe I'm wrong. See the request here where they are selecting the PK & RK: https://docs.microsoft.com/en-us/rest/api/storageservices/query-entities#sample-request

Issue #15008, opened 2021-07-08
[Tables] Authenticate via TokenCredential


seankane-msft:
`TableClient` and `TableServiceClient` should be able to use a `TokenCredential` from `azidentity`

PR #15101, opened 2021-07-20
[EngSys] Test proxy

seankane-msft:


seankane-msft:
closes #15008

sdk/internal/recording/recording.go
```diff
@@ -412,3 +417,275 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+// 	req.URL.Scheme = "https"
+// 	req.URL.Host = p.options.host
+// 	req.Host = p.options.host
+
+// 	req.Header.Set(UpstreamUriHeader, fmt.Sprintf("%v://%v", p.options.scheme, originalURLHost))
+// 	req.Header.Set(ModeHeader, recordMode)
+// 	req.Header.Set(recordingIdHeader, recordingId)
+
+// 	return req.Next()
+// }
```
seankane-msft, line 588: I would love to have this commented out section in azcore because it is going to be used by every test suite that runs against the test-proxy, however it creates a circular dependency. @jhendrixMSFT is there another way we could do this without creating a circular dependency between internal and azcore? We can break out functionality used strictly for testing as it's own library (but would have to require azcore to not use this functionality)

sdk/internal/recording/recording.go
```diff
@@ -412,3 +417,275 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+// 	}
+// }
+
+// func (f *FakeCredential) AuthenticationPolicy(azcore.AuthenticationPolicyOptions) azcore.Policy {
+// 	return azcore.PolicyFunc(func(req *azcore.Request) (*azcore.Response, error) {
+// 		authHeader := strings.Join([]string{"Authorization ", f.accountName, ":", f.accountKey}, "")
+// 		req.Request.Header.Set(azcore.HeaderAuthorization, authHeader)
+// 		return req.Next()
+// 	})
+// }
```
seankane-msft, line 648: Same as previous comment for fake credentials.

PR #15104, opened 2021-07-20
[Core] Rename AnonymousCredential

seankane-msft:
`AnonymousCredential()` -> `NewAnonymousCredential()`

### DIFF
--- a/sdk/azcore/policy_anonymous_credential.go
+++ b/sdk/azcore/policy_anonymous_credential.go
@@ -5,7 +5,7 @@
 
 package azcore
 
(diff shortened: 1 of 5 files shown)


sdk/azcore/policy_anonymous_credential.go
```diff
@@ -15,6 +15,6 @@ func anonCredPolicyFunc(req *Request) (*Response, error) {
 
 // AnonymousCredential is for use with HTTP(S) requests that read public resource
```
catalinaperalta, line 16:
```suggestion
// NewAnonymousCredential is for use with HTTP(S) requests that read public resource
```

chamons (approved):

PR #15137, opened 2021-07-23
[azidentity] Update code to use the latest azcore

catalinaperalta:
Update azidentity to depend on new changes in azcore v0.17.0.

### DIFF
--- a/sdk/azidentity/azidentity.go
+++ b/sdk/azidentity/azidentity.go
@@ -28,6 +28,16 @@ const (
 	defaultSuffix = "/.default"
 )
 
+const (
+	headerXmsDate                = "x-ms-date"
+	headerUserAgent              = "User-Agent"
+	headerURLEncoded             = "application/x-www-form-urlencoded"
+	headerAuthorization          = "Authorization"
+	headerAuxiliaryAuthorization = "x-ms-authorization-auxiliary"
+	headerMetadata               = "Metadata"
+	headerContentType            = "Content-Type"
+)
+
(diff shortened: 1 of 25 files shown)


ArcturusZhang:
Hi @catalinaperalta I did not notice this PR before I open mine https://github.com/Azure/azure-sdk-for-go/pull/15176

Comparing the content, we are missing some type assertions in this PR, like this: https://github.com/Azure/azure-sdk-for-go/blob/edb91f888461a893589c241c969eb9b5f6712b9a/sdk/azidentity/client_certificate_credential.go#L54
Could we also add these?

sdk/azidentity/azidentity.go
```diff
@@ -28,6 +28,15 @@ const (
...
 
+const (
+	headerXmsDate       = "x-ms-date"
+	headerUserAgent     = "User-Agent"
+	headerURLEncoded    = "application/x-www-form-urlencoded"
+	headerAuthorization = "Authorization"
+	headerMetadata      = "Metadata"
+	headerContentType   = "Content-Type"
+)
+
```
seankane-msft, line 39: is there a benefit to put these in internal? Will other packages also use these?

seankane-msft (approved):
LGTM :shipit:

sdk/azidentity/azidentity.go
catalinaperalta: I thought the same as you, but look at the thread in this PR: https://github.com/Azure/azure-sdk-for-go/pull/15038. There's a point that it might take too long to make changes to `azcore` if we need to add new headers and then release a new version of the `internal` module. 

Still I think there's an argument to be made for having some of the more common ones in the `internal` module if it makes it easier to handle for other SDK devs.

PR #15150, opened 2021-07-26
Update `armcompute` using latest code gen

ArcturusZhang:


sdk/compute/armcompute/zz_generated_connection.go
```diff
@@ -0,0 +1,82 @@
+// +build go1.13
```
ArcturusZhang, line 1: Why this file is generated? We have `azure-arm: true`

sdk/compute/armcompute/autorest.md
```diff
@@ -0,0 +1,9 @@
+### AutoRest Configuration
+
+> see https://aka.ms/autorest
+
+``` yaml
+require:
+- https://github.com/Azure/azure-rest-api-specs/blob/d7dc5fbb25ed5cc718f00afb568251294c8646b4/specification/compute/resource-manager/readme.md
+- https://github.com/Azure/azure-rest-api-specs/blob/d7dc5fbb25ed5cc718f00afb568251294c8646b4/specification/compute/resource-manager/readme.go.md
```
ArcturusZhang, line 8: I have to also require the `readme.go.md` otherwise the `directive` written in `readme.go.md` will not be included in this generation

sdk/compute/armcompute/zz_generated_constants.go
```diff
@@ -7,7 +7,7 @@
 
 package armcompute
 
-const telemetryInfo = "azsdk-go-armcompute/v0.2.0"
+const telemetryInfo = "azsdk-go-armcompute/<version>"
```
jhendrixMSFT, line 10: Bad version

sdk/compute/armcompute/build.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+// This file enables 'go generate' to regenerate this specific SDK
+//go:generate pwsh.exe ..\..\..\eng\scripts\build.ps1 -skipBuild -generate armcompute
```
jhendrixMSFT, line 5: Should be forward-slashes to work on other OS where PS is supported.

sdk/compute/armcompute/zz_generated_connection.go
jhendrixMSFT: It really seems like this setting is not being included.  Are you sure the new way of building is including the correct config files?

ArcturusZhang:
Hi @jhendrixMSFT could you please take a look at this PR again?

PR #15154, opened 2021-07-26
[identity] Remove nolints

### LINKED ISSUE #15110: [Identity] Remove golangci-lint skips
Remove the `// nolint` comments that were added by this [PR](https://github.com/Azure/azure-sdk-for-go/pull/15109).

seankane-msft:
closes #15110


catalinaperalta (approved):
LGTM...not sure why the linter is saying the page assignment is ineffectual

sdk/azidentity/interactive_browser_server.go
```diff
@@ -70,7 +72,10 @@ func (s *server) Start(reqState string, port int) string {
 			if s.err != nil {
 				page = failPage
 			}
-			w.Write([]byte(page)) // nolint:errcheck
+			_, err := w.Write([]byte(page))
+			if err != nil {
+				page = failPage
```
seankane-msft, line 77:
```suggestion
                 page = failPage //nolint:ineffassign
```

sdk/azidentity/interactive_browser_server.go
```diff
@@ -70,7 +72,10 @@ func (s *server) Start(reqState string, port int) string {
 			if s.err != nil {
 				page = failPage
 			}
-			w.Write([]byte(page)) // nolint:errcheck
+			_, err := w.Write([]byte(page))
+			if err != nil {
+				page = failPage //nolint:ineffassign
```
benbp, line 77: What is this doing? Seems like it was copied from above, but the above error handler is to set the failPage for writing. In case the actual write fails I'd assume you want to handle that too somehow (or ignore as before), as opposed to re-setting the variable which is then never used?

catalinaperalta: Agree with Ben about this actually. This assignment is ineffectual here.

catalinaperalta:
Are all of these changes still necessary?

PR #15162, opened 2021-07-27
[Internal] fixing lint issues

seankane-msft:


sdk/internal/ci.yml
```diff
@@ -3,12 +3,14 @@ trigger:
   paths:
     include:
     - sdk/internal/
+    - eng/
```
catalinaperalta, line 6: Why does this need to be added?

catalinaperalta (approved):

sdk/internal/ci.yml
seankane-msft: I'm doing this in a separate PR so I think this was a mental block adding it here, it will get added eventually so I'm going to keep it as is.

chamons (approved):

PR #15167, opened 2021-07-27
[EngSys] converting test and coverage steps into single script and step

seankane-msft:
- [ ] Convert steps into single powershell script
- [ ] Add more debugging information in output
- [ ] Allows devs to run the exact process locally via script

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -20,49 +20,53 @@ steps:
     env:
       GO111MODULE: 'on'
 
-  - ${{ if eq(parameters.RunTests, 'true') }}:
-    - pwsh: |
-        go get github.com/jstemmer/go-junit-report
-        go get github.com/axw/gocov/gocov
-        go get github.com/AlekSi/gocov-xml
-        go get github.com/matm/gocov-html
-        go get -u github.com/wadey/gocovmerge
-      displayName: "Install Coverage and Junit Dependencies"
-      workingDirectory: '${{parameters.GoWorkspace}}'
+  - pwsh: ../eng/scripts/run_tests.ps1
+    displayName: 'Run Tests and Create Coverage'
+    workingDirectory: '${{parameters.GoWorkspace}}'
 
-    - pwsh: |
-        $testDirs = (./eng/scripts/get_test_dirs.ps1 -serviceDir $(SCOPE))
-        foreach ($td in $testDirs) {
-          pushd $td
-          Write-Host "##[command]Executing go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic $td | go-junit-report -set-exit-code > report.xml"
-          go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic . | go-junit-report -set-exit-code > report.xml
-          # if no tests were actually run (e.g. examples) delete the coverage file so it's omitted from the coverage report
-          if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simplematch -quiet) {
-            Write-Host "##[command]Deleting empty coverage file"
-            rm coverage.txt
-          }
-        }
-      displayName: 'Run Tests'
-      workingDirectory: '${{parameters.GoWorkspace}}'
-      env:
-        GO111MODULE: 'on'
+  # - ${{ if eq(parameters.RunTests, 'true') }}:
+  #   - pwsh: |
+  #       go get github.com/jstemmer/go-junit-report
+  #       go get github.com/axw/gocov/gocov
+  #       go get github.com/AlekSi/gocov-xml
+  #       go get github.com/matm/gocov-html
+  #       go get -u github.com/wadey/gocovmerge
+  #     displayName: "Install Coverage and Junit Dependencies"
+  #     workingDirectory: '${{parameters.GoWorkspace}}'
 
-    - pwsh: ../eng/scripts/create_coverage.ps1
-      displayName: 'Generate Coverage XML'
-      workingDirectory: '${{parameters.GoWorkspace}}sdk'
+  #   - pwsh: |
+  #       $testDirs = (./eng/scripts/get_test_dirs.ps1 -serviceDir $(SCOPE))
+  #       foreach ($td in $testDirs) {
+  #         pushd $td
+  #         Write-Host "##[command]Executing go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic $td | go-junit-report -set-exit-code > report.xml"
+  #         go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic . | go-junit-report -set-exit-code > report.xml
+  #         # if no tests were actually run (e.g. examples) delete the coverage file so it's omitted from the coverage report
+  #         if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simplematch -quiet) {
+  #           Write-Host "##[command]Deleting empty coverage file"
+  #           rm coverage.txt
+  #         }
+  #       }
+  #     displayName: 'Run Tests'
+  #     workingDirectory: '${{parameters.GoWorkspace}}'
+  #     env:
+  #       GO111MODULE: 'on'
 
-    - task: PublishTestResults@2
-      condition: succeededOrFailed()
-      inputs:
-        testRunner: JUnit
-        testResultsFiles: '${{parameters.GoWorkspace}}sdk/**/report.xml'
-        testRunTitle: 'Go ${{ parameters.GoVersion }} on ${{ parameters.Image }}'
-        failTaskOnFailedTests: true
+    # - pwsh: ../eng/scripts/create_coverage.ps1
+    #   displayName: 'Generate Coverage XML'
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -20,49 +20,53 @@ steps:
...
-  - ${{ if eq(parameters.RunTests, 'true') }}:
-    - pwsh: |
-        go get github.com/jstemmer/go-junit-report
-        go get github.com/axw/gocov/gocov
-        go get github.com/AlekSi/gocov-xml
-        go get github.com/matm/gocov-html
-        go get -u github.com/wadey/gocovmerge
-      displayName: "Install Coverage and Junit Dependencies"
-      workingDirectory: '${{parameters.GoWorkspace}}'
+  - pwsh: ../eng/scripts/run_tests.ps1
```
benbp, line 23: I would make this an absolute path.

```suggestion
  - pwsh: /eng/scripts/run_tests.ps1
```

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,80 @@
...
+    Push-Location $td
+    $temp = Get-Location
+    Write-Host "Currently in $temp"
+    Write-Host "##[command] Executing go test -run ""^Test"" -race -v -coverprofile coverage.txt -covermode atomic $td | go-junit-report -set-exit-code > report.xml"
+    go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic . | go-junit-report -set-exit-code > report.xml
+    if (!$?) {
+        Write-Host "There was an error running the tests"
+        Exit $LASTEXITCODE
+    } else {
+        Write-Host "`nSuccessfully ran test suite at $temp`n"
```
benbp, line 39: I got annoyed with newline escapes and started doing this instead (up to you):

```suggestion
        Write-Host "" "Successfully ran test suite at $temp" ""
```

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,80 @@
+Param(
+    [string] $serviceDir
+)
+
+$cwd = Get-Location
+
+# 0. Find all test directories
+Write-Host "Finding test directories..."
+$testDirs = . $PSScriptRoot/get_test_dirs.ps1 -serviceDir sdk/$serviceDir
```
benbp, line 9: If you mean to execute the script, then you shouldn't also dot source it.

```suggestion
$testDirs = $PSScriptRoot/get_test_dirs.ps1 -serviceDir sdk/$serviceDir
```

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,80 @@
+Param(
+    [string] $serviceDir
+)
+
+$cwd = Get-Location
+
+# 0. Find all test directories
+Write-Host "Finding test directories..."
```
benbp, line 8:
```suggestion
Write-Host "Finding test directories in sdk/$serviceDir..."
```

eng/scripts/run_tests.ps1
seankane-msft: Doing this throws an error: `You must provide a value expression following the '/' operator.
```
ParserError: C:\Users\Sean Kane\go\src\github.com\Azure\azure-sdk-for-go\eng\scripts\run_tests.ps1:9
Line |
   9 |  $testDirs = $PSScriptRoot/get_test_dirs.ps1 -serviceDir sdk/$serviceD …
     |                            ~
     | You must provide a value expression following the '/' operator.
```

eng/pipelines/templates/steps/build-test.yml
seankane-msft: I added this change but it threw an error in DevOps [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1021652&view=logs&j=35205994-0183-5e87-7c1e-dd3d552c3593&t=8ccf51ca-6d04-5845-9c4d-7ec54a8f0a12&l=21).

benbp: Oh. I recommend invoking this via the [full powershell task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops) rather than calling the script via an inline `pwsh` command.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -20,9 +20,15 @@ steps:
...
 
-  - pwsh: ./eng/scripts/run_tests.ps1
-    displayName: 'Run Tests and Create Coverage'
-    workingDirectory: '${{parameters.GoWorkspace}}'
+  - task: PowerShell@2
+    inputs:
+      targetType: 'filePath'
+      filePath: '/eng/scripts/run_tests.ps1'
+      workingDirectory: '${{parameters.GoWorkspace}}'
+      arguments: '${{parameters.ServiceDirectory}}'
```
weshaggard, line 28: If you switch to the task directly you will need to set `pwsh: true` to run on powershell core if this runs on anything other than windows.

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,86 @@
...
+        Exit 0
+    }
+}
+
+Set-Location $cwd
+$temp = Get-Location
+Write-Host "Currently in $temp"
+
+# 2. Install coverage tools needed
+go get github.com/jstemmer/go-junit-report
```
benbp, line 60: I wonder if it's worth versioning these (e.g. making coverage its own go mod project directory)? That way we won't by surprised by any dependency changes with these tools that may affect the consistency of our coverage+test report generation.

weshaggard: We definitely try and pin versions of other tools to make sure we don't get broken by an unexpected update.

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+$runTests = $false
+# 0b. Verify there are test files with tests to run in at least one of these directories
+$testDirs | ForEach-Object {
+    Get-ChildItem -Path $_ -Filter *_test.go | ForEach-Object {
+        if (Select-String -path $_ -pattern 'Test' -SimpleMatch) {
+            $runTests = $true
+        }
+    }
+
+    if (!$runTests) {
```
benbp, line 22: This will fail immediately as the check runs for each dir.

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+
+# 0. Find all test directories
+Write-Host "Finding test directories in 'sdk/$serviceDir'"
+$testDirs = & $PSScriptRoot/get_test_dirs.ps1 -serviceDir sdk/$serviceDir
+# Issues here, not returning any objects
+Write-Host "Found test directories $testDirs"
+
+$runTests = $false
+# 0b. Verify there are test files with tests to run in at least one of these directories
+$testDirs | ForEach-Object {
```
benbp, line 15: Since you're not doing object construction, we try to have a convention using `foreach ()` instead, as it reads a little easier and avoids certain powershell gotchas with pipelining.

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+    Get-ChildItem -Path $_ -Filter *_test.go | ForEach-Object {
+        if (Select-String -path $_ -pattern 'Test' -SimpleMatch) {
```
benbp, line 17: I think `-pattern 'func Test` would be a more exact thing to match against.

seankane-msft: Some places use the `testify` suite capabilities and those tests don't have that pattern so for now I don't think that's the best way to do it

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -44,57 +41,39 @@ steps:
...
-        go get github.com/AlekSi/gocov-xml
-        go get github.com/matm/gocov-html
-        go get -u github.com/wadey/gocovmerge
-      displayName: "Install Coverage and Junit Dependencies"
-      workingDirectory: '${{parameters.GoWorkspace}}'
-
-    - template: /eng/common/testproxy/test-proxy-docker.yml
+  - template: /eng/common/testproxy/test-proxy-docker.yml
 
-    - pwsh: |
```
weshaggard, line 59: This change is going to clash with https://github.com/Azure/azure-sdk-for-go/pull/15560. Please have a look and we can figure out a good way to merge them. We might even want to remove the get_test_dirs.ps1 completely now and use the new `Get-AllPackageInfoFromRepo` function directly in your run_tests script.

eng/scripts/run_tests.ps1
```diff
@@ -0,0 +1,68 @@
...
+Set-Location $PSScriptRoot
+
+$coverageFiles = [Collections.Generic.List[String]]@()
+Get-ChildItem -recurse -path . -filter coverage.txt | ForEach-Object {
+  $covFile = $_.FullName
+  Write-Host "Adding $covFile to the list of code coverage files"
+  $coverageFiles.Add($covFile)
+}
+
+Pop-Location
```
benbp, line 68: Given that you're not popping location after all the calls to Push-Location, I think this will just drop you out at the most recent Push-Location directory.

PR #15182, opened 2021-07-28
[internal] Move logging functionality to `internal`

seankane-msft:
- [ ] This is step one two to move logger functionality to internal and leave only `SetClassification` and `SetListener` in `azcore`


catalinaperalta (approved):
🚢

PR #15187, opened 2021-07-29
[Tables] connection string parser

seankane-msft:
Adds connection string parsing for tables, allowing the customer to create a `TableClient` or `TableServiceClient` using a connection string with the methods `New...ClientFromConnectionString`


sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+func convertConnStrToMap(connStr string) (map[string]string, error) {
+	ret := make(map[string]string)
+	connStr = strings.TrimRight(connStr, ";")
+
+	splitString := strings.Split(connStr, ";")
+	if len(splitString) == 0 {
+		return ret, ErrConnectionString
+	}
+	for _, stringPart := range splitString {
+		parts := strings.Split(stringPart, "=")
```
richardpark-msft, line 41: This might not happen in Tables but in Service Bus we can have embedded '='s in the connection string.

My understanding is that's legal so you might want to say .SplitN(stringPart, "=", 2), which will only split _once_.

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+		if len(parts) != 2 {
+			return ret, ErrConnectionString
+		}
+		ret[parts[0]] = parts[1]
+	}
+	return ret, nil
+}
+
+func parseConnectionString(connStr string) (string, *azcore.Credential, error) {
```
richardpark-msft, line 50: Even though it's a private method it'd be nice to document it.

For instance, it's not immediately intuitive what that first 'string' is in the return value.

I don't know if it's frowned upon, but using named return values would also do the trick here. (@JeffreyRichter , @jhendrixMSFT?)

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+		return "", nil, ErrConnectionString
+	}
+
+	if accountName == "" || accountKey == "" {
+		// Try sharedaccesssignature
+		sharedAccessSignature, ok := connStrMap["sharedaccesssignature"]
+		if !ok {
+			return serviceURL, nil, ErrConnectionString
+		}
+		return sharedAccessSignature, nil, errors.New("there is not support for SharedAccessSignature yet")
```
richardpark-msft, line 74:
```suggestion
		return sharedAccessSignature, nil, errors.New("a SharedAccessSignature are not supported")
```

'yet' is kind of weird when it doesn't have a timeframe. There was a mild typo ('there is not support' should probably be 'there is no support') if you want to stick with the original text.

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+		return "", nil, ErrConnectionString
+	}
+	accountKey, ok := connStrMap["AccountKey"]
+	if !ok {
+		return "", nil, ErrConnectionString
+	}
+
+	if accountName == "" || accountKey == "" {
+		// Try sharedaccesssignature
+		sharedAccessSignature, ok := connStrMap["sharedaccesssignature"]
```
richardpark-msft, line 70: Are key names case-sensitive in connection strings?

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+	if !ok {
+		defaultProtocol = "https"
+	}
+
+	endpointSuffix, ok := connStrMap["EndpointSuffix"]
+	if !ok {
+		endpointSuffix = "core.windows.net"
+	}
+
+	tableEndpoint, ok := connStrMap["TableEndpoint"]
```
richardpark-msft, line 88: Looks like you can combine some of this and just set 'tableEndpoint' to either their custom one or the serviceURL you formed below (and return only once). Unless there's something else interesting you have planned where they'll diverge more.

sdk/tables/aztable/connection_string_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConnectionStringParser(t *testing.T) {
+	require := require.New(t)
+
+	connStr := "DefaultEndpointsProtocol=https;AccountName=dummyaccount;AccountKey=secretkeykey;EndpointSuffix=core.windows.net"
```
richardpark-msft, line 15: Do users get back a sensible error if they happened to paste the URL into 'AccountName' for some reason? Or even the full hostname (ie, accountname.core.windows.net)? Just curious what the service tells them.

sdk/tables/aztable/connection_string_test.go
```diff
@@ -0,0 +1,70 @@
...
+	serviceURL, cred, err := parseConnectionString(connStr)
+	require.NoError(err)
+	require.Equal(serviceURL, "https://dummyaccount.table.core.windows.net")
+	require.NotNil(cred)
+}
+
+func TestConnectionStringParserCustomDomain(t *testing.T) {
+	require := require.New(t)
+
+	connStr := "AccountName=dummyaccount;AccountKey=secretkeykey;TableEndpoint=www.mydomain.com;"
```
richardpark-msft, line 45: When they provide a table endpoint is it required that it has the https:// on it?

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+var ErrConnectionString = errors.New("connection string is either blank or malformed")
```
richardpark-msft, line 14: I find it helpful when the error message includes an example of the expected format, in case the user is just totally lost on what you're asking them for:

<for instance, you have this one like this in your code - I'd clean this one up so it doesn't say 'dummy' though>
"DefaultEndpointsProtocol=http;AccountName=dummyaccount;AccountKey=secretkeykey;EndpointSuffix=core.windows.net"

sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,101 @@
...
+
+func NewTableClientFromConnectionString(tableName string, connectionString string, options *TableClientOptions) (*TableClient, error) {
+	endpoint, credential, err := parseConnectionString(connectionString)
+	if err != nil {
+		return nil, err
+	}
+	return NewTableClient(tableName, endpoint, *credential, options)
+}
+
+func NewTableServiceClientFrommConnectionString(connectionString string, options *TableClientOptions) (*TableServiceClient, error) {
```
richardpark-msft, line 24:
```suggestion
func NewTableServiceClientFromConnectionString(connectionString string, options *TableClientOptions) (*TableServiceClient, error) {
```
Using too many 'm's chips away at our quarterly profits. Characters aren't free!

sdk/tables/aztable/connection_string_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func TestConnectionStringParserHTTP(t *testing.T) {
+	require := require.New(t)
+
+	connStr := "DefaultEndpointsProtocol=http;AccountName=dummyaccount;AccountKey=secretkeykey;EndpointSuffix=core.windows.net"
+	serviceURL, cred, err := parseConnectionString(connStr)
+	require.NoError(err)
+	require.Equal(serviceURL, "http://dummyaccount.table.core.windows.net")
+	require.NotNil(cred)
```
richardpark-msft, line 29: Why aren't we testing any of the fields in 'cred'?

richardpark-msft (changes requested):
We should test 'cred' for sure, since it's where the bulk of the connection string information actually goes.

Some other minor comments as well.

sdk/tables/aztable/connection_string.go
christothes: yes the sharedKey typically has "==" in it

sdk/tables/aztable/connection_string_test.go
christothes: only for SAS credential auth.

sdk/tables/aztable/connection_string_test.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
christothes, line 1: We need tests for azurite connection strings also, since they do some different things with the account name.

sdk/tables/aztable/connection_string.go
seankane-msft: Good point, although this is a personal placeholder and I intend on having SAS complete before the first preview release of tables goes out.

sdk/tables/aztable/connection_string.go
seankane-msft: No, this should be `SharedAccessSignature` instead, thank you for pointing this out.

sdk/tables/aztable/connection_string_test.go
seankane-msft: That's a good question, I would assume it will be an authentication error which from personal experience have been vague. I will make a note to investigate.

sdk/tables/aztable/connection_string_test.go
seankane-msft: The method returns an instance of `azcore.Credential` which is an interface that does not define any fields other than an `AuthenticationPolicy` method. I could change this to return an instance of SharedKeyCredential but in the future this could also return an instance of a SharedAccessSignatureCredential which puts me in a bit of a pickle.

richardpark-msft: I think it's okay for a unit test to use internal details to be more thorough, btw. So in this case it's perfectly fine (in a unit test in the same package) to do the cast to a concrete type.

So:

```go
value, ok := cred.(*SharedKeyCredential)
require.True(ok)  // ie, it's really of this type

// check fields of 'value' which is now of type *SharedKeyCredential
```

Issue #15180, opened 2021-07-28
UUID is pseudo random


pascaldekloe:
The UUIDs don't use a cryptographically randomised source. Not doing so breaks the specification.

RickWinter:
Thank you for reporting this issue.  We will make a fix available soon.

pascaldekloe:
The cryptographic random source can encounter errors, so you'd need to either change the name or use another UUID version.

pascaldekloe:
Thank you @RickWinter. The changes look good.

PR #15190, opened 2021-07-29
Update to crypto secure random number generation

RickWinter:


sdk/internal/uuid/uuid.go
```diff
@@ -6,45 +6,28 @@
...
 // A UUID representation compliant with specification in RFC 4122 document.
 type UUID [16]byte
 
 // New returns a new uuid using RFC 4122 algorithm.
 func New() UUID {
 	u := UUID{}
-	// Set all bits to randomly (or pseudo-randomly) chosen values.
-	// math/rand.Read() is no-fail so we omit any error checking.
+	// Set all bits to pseudo-random values.
 	// NOTE: this takes a process-wide lock
```
richardpark-msft, line 26: Do we know if the 'process-wide lock' bit is still true?

sdk/internal/uuid/uuid.go
```diff
@@ -6,45 +6,28 @@
...
 )
 
-func init() {
-	rand.Seed(time.Now().Unix())
-}
-
 // A UUID representation compliant with specification in RFC 4122 document.
 type UUID [16]byte
 
 // New returns a new uuid using RFC 4122 algorithm.
```
richardpark-msft, line 22: I'm not sure about convention but if a function can panic I think it should definitely be documented as that's a very rare occurrence. Maybe it's just taken as a given that though.

CC: @JeffreyRichter , @jhendrixMSFT for idiom check

richardpark-msft (commented):
The panic has me 'panic'king a bit because it feels like a big runtime change but I'm not sure how we treat that in the Go SDK.

I think the usage of this code is local enough that we could just inline it into azure-identity (I think this was the package we saw yesterday).

sdk/internal/uuid/uuid.go
seankane-msft: According to the [docs](https://pkg.go.dev/math/rand#Read), there is no mention of a process wide lock, I think this note can be removed in this case.

sdk/internal/uuid/uuid.go
seankane-msft: [Docs](https://pkg.go.dev/math/rand#Read) mention that this is always a nil error, I think we can ignore the error for now. You will likely need a `//nolint` comment, the linter might complain about not checking a returned error.

richardpark-msft: I think that doc is for math/rand (@RickWinter swapped to [crypto/rand](https://pkg.go.dev/crypto/rand))

seankane-msft: @richardpark-msft thanks for catching that the math/rand docs also make no mention of a long process, but I think we should be returning the error in this case.

RickWinter: I spoke with Jeff on this change and its our conclusion that the error cases should never be hit given how we use the API.  If they are hit then we would want to stop (panic) so users can debug an unexpected and unrecoverable error

JeffreyRichter (approved):

jhendrixMSFT:
Can I get some context on the reasoning for this change?  When we initially created this helper, we explicitly decided against using the crypto RNG as this was for creating UUIDs for things like request IDs (i.e. it was never meant to be a cryptographically secure UUID implementation).  Presumably this invariant was lost along the way and we're now trying to shore this up?  Assuming this is the case, are we sure that the impl as it stands meets that requirement?  Put another way, if we need a cryptographically secure UUID generation facility, should we be rolling our own or pick an existing impl?

RickWinter:
> Can I get some context on the reasoning for this change? When we initially created this helper, we explicitly decided against using the crypto RNG as this was for creating UUIDs for things like request IDs (i.e. it was never meant to be a cryptographically secure UUID implementation). Presumably this invariant was lost along the way and we're now trying to shore this up? Assuming this is the case, are we sure that the impl as it stands meets that requirement? Put another way, if we need a cryptographically secure UUID generation facility, should we be rolling our own or pick an existing impl?

The issue with the current impl is it will result in duplicate request ids if a client sends enough requests.  That results in a less than ideal debugging experience.  The goal of this change is to remove the likelihood we generate duplicate requiest ids

sdk/internal/uuid/uuid.go
jhendrixMSFT: It's not in the docs, you have to look at the impl.  For `crypto/rand` the comment might not be applicable.

jhendrixMSFT: For the Unix impl it also takes a lock.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/15180

sdk/internal/uuid/uuid.go
jhendrixMSFT: This has been fixed to not panic.  I also fixed the `Parse()` function as it was swallowing errors.

seankane-msft (approved):

PR #15196, opened 2021-07-30
[Tables] Architect Feedback

seankane-msft:
- [ ] Change Upsert to Insert, note that it does not fail if the entity already exists
- [ ] TableUpdateMode -> EntityUpdateMode, ReplaceEntity and MergeEntity
- [ ] Flow the delete options through


sdk/tables/aztable/table_service_client.go
```diff
@@ -50,8 +50,11 @@ func (t *TableServiceClient) CreateTable(ctx context.Context, name string) (Tabl
 }
 
 // Delete deletes a table by name.
-func (t *TableServiceClient) DeleteTable(ctx context.Context, name string) (TableDeleteResponse, error) {
-	return t.client.Delete(ctx, name, nil)
+func (t *TableServiceClient) DeleteTable(ctx context.Context, name string, options *TableDeleteOptions) (TableDeleteResponse, error) {
+	if options == nil {
```
JeffreyRichter, line 54: You shouldn't need this if because Delete will do the same thing.

JeffreyRichter (approved):

JeffreyRichter (approved):

PR #15207, opened 2021-08-03
[azcore] Add ETag type to Core

### LINKED ISSUE #15153: [Core] Add an `ETag` type

seankane-msft:
closes #15153 

Based off similar PR for [c++](https://github.com/Azure/azure-sdk-for-cpp/pull/1428)


seankane-msft:
@jhendrixMSFT @JeffreyRichter are there other methods/properties that need to be added for ETag?

JeffreyRichter:
See the c++ etag that Rick did. We spent a lot of time on that.

-- Jeffrey Richter<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.linkedin.com%2Fin%2Fjeffrichter%2F&data=04%7C01%7Cjeffreyr%40microsoft.com%7C22479ac85e5243cd5daf08d93164b830%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637595130922909659%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=0eSlDrJljhzGEmilUf5wTrxyhPVnAQ%2BksbnPiH4lJBs%3D&reserved=0> (see Architecting Distributed Cloud Apps<https://nam06.safelinks.protection.outlook.com/?url=http%3A%2F%2Faka.ms%2FRichterCloudApps&data=04%7C01%7Cjeffreyr%40microsoft.com%7C22479ac85e5243cd5daf08d93164b830%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637595130922919618%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=00QzMoRaDwCfjHo9hRw2h%2BO90PLM5mr9LRgALfzsXMg%3D&reserved=0> & Designing HTTP APIs<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D9Ng00IlBCtw%26list%3DPL9XzOCngAkqs4m0XdULJu_78nM3Ok3Q65%26index%3D1&data=04%7C01%7Cjeffreyr%40microsoft.com%7C22479ac85e5243cd5daf08d93164b830%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637595130922919618%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=7IX%2F2xnJDM6lR%2F26iR8c6Y9HJL2cKw%2B0cXTBd8oMQ2s%3D&reserved=0>)

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+package azcore
+
+import (
+	"strings"
+)
+
+// ETag is a property used for optimistic concurrency during updates
+// ETag is a validator based on https://tools.ietf.org/html/rfc7232#section-2.3.2
+type ETag struct {
+	value *string
```
JeffreyRichter, line 15: If a just allocate one of these value will be nil; Then, if I call Equals on it, a panic will happen.
So is nil a meaningful etag value or not?
I think we DO need the concept of a missing etag; nil could be this but then Equals needs to be fixed.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+
+// NewETag creates a new ETag struct
+func NewETag(value string) *ETag {
+	return &ETag{value: &value}
+}
+
+// ComparisonType specifies what type of comparison to use, Strong or Weak
+type ComparisonType string
+
+const (
```
JeffreyRichter, line 26: I suggest we get rid of these constants and have WeakEquals and StrongEquals methods instead.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+	Strong ComparisonType = "strong"
+	Weak   ComparisonType = "weak"
+)
+
+// Equals determines whether two entity-tags are equal. There are two types of comparison: Strong and Weak.
+// Strong comparison: two ETags are equivalent if both are not weak and their opaque-tags match character-by-character
+// Weak Comparison: two ETags are equivalent if their opaque-tags match character-by-character regardless of either or both being tagged as "weak"
+func (e ETag) Equals(right ETag, comparisonKind ComparisonType) bool {
```
JeffreyRichter, line 34: I'd rename 'right' as 'other' since equals is a symmetric operation.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+	// ETags are != if one value is null
```
JeffreyRichter, line 35: nil; not null

JeffreyRichter: I'm not sure what is happening here. Maybe you don't mean null/nil as a string in Go is never nil. A pointer to a string can be nil but you're not checking the pointer. Maybe you mean empty string instead of null/nil?
I'm not following the logic from the comments.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+	leftStart := e.getStart()
+	rightStart := right.getStart()
+
+	leftValue := (*e.value)[leftStart:]
+	rightValue := (*right.value)[rightStart:]
+
+	return leftValue == rightValue
+}
+
+func (e ETag) getStart() int {
```
JeffreyRichter, line 54: THis function is only called by Equals, right? If so, I'd define it inside equals itself:
getstart := func() int {
   if e.IsWeak() { 
      return 2 
   }
   return 0
 }

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+	if e.IsWeak() {
+		return 2
+	}
+	return 0
+}
+
+// String returns a string representation of an ETag
+func (e ETag) String() string {
+	return *e.value
```
JeffreyRichter, line 63: Will panic if value is nil

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// HasValue returns whether an ETag is present
+func (e ETag) HasValue() bool {
```
JeffreyRichter, line 67: So, this is good so you sort of have the concept of a non-existent etag. It's just not followed all the way through this code.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+	return e.value != nil
+}
+
+// IsWeak specifies whether the ETag is strong or weak.
+func (e ETag) IsWeak() bool {
+	return e.value != nil && len(*e.value) >= 4 && strings.HasPrefix(*e.value, "W/\"") && strings.HasSuffix(*e.value, "\"")
```
JeffreyRichter, line 73: I'd replace "e.value != nil" with a call to HasValue instead. Make the code more self-documenting/sustainable.

sdk/azcore/etag_test.go
```diff
@@ -0,0 +1,224 @@
...
+		t.Fatalf("ETag values are not equal")
+	}
+
+	e2 := NewETag("\"tag\"")
+	if e2.String() != "\"tag\"" {
+		t.Fatalf("ETag values are not equal")
+	}
+
+	e3 := NewETag("W/\"weakETag\"")
+	if e3.String() != "W/\"weakETag\"" {
```
richardpark-msft, line 24: No `testify/require` allowed in azcore?

sdk/azcore/etag.go
```diff
@@ -0,0 +1,80 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcore
+
+import (
+	"strings"
+)
+
+// ETag is a property used for optimistic concurrency during updates
```
richardpark-msft, line 12: Is there a need to have a struct at all if it only has a single value? Have we considered just saying a '*string' is the etag and just exporting a compare function that takes two *string's? 

Or do you have some more stuff planned here?

richardpark-msft (commented):
Just have some questions.

sdk/azcore/etag.go
JeffreyRichter: Good point, probably "type ETag *string" is good enough.

sdk/azcore/etag_test.go
seankane-msft: It's not currently there but I'm going to add it because this is too much repeated code for tests

sdk/azcore/etag.go
seankane-msft: I fixed this to add the logic that two `nil` `ETag`s will be equal, and they will be unequal if only one is `nil`

JeffreyRichter (approved):

sdk/azcore/core.go
```diff
@@ -202,4 +202,4 @@ func IsNullValue(v interface{}) bool {
 	}
 	// no sentinel object for this *t
 	return false
-}
+}
```
jhendrixMSFT, line 205: nit, missing EOL

sdk/azcore/etag.go
jhendrixMSFT: Is there a reason it's `*string` and not just `string`?  The former would make it awkward for by-value cases.

seankane-msft: @jhendrixMSFT this allows the ETag to be nil which allows a differentiation of nil ETag and an ETag of `""`

JeffreyRichter: Yes. It is valid (although very unlikely) to have an empty-string etag. So, we need a way to say that is actually no etag at all.

jhendrixMSFT: I see.  Is this common knowledge for ETag users, should we add a comment here explaining this?

JeffreyRichter: In the docs, we can reference this RFC: https://www.rfc-editor.org/rfc/rfc7232.html#section-2.3
The last example shows an empty string.

sdk/azcore/etag.go
```diff
@@ -0,0 +1,62 @@
...
+	bStart := getStart(b)
+
+	aVal := (*a)[aStart:]
+	bVal := (*b)[bStart:]
+
+	return aVal == bVal
+}
+
+// HasValue returns whether an ETag is present
+func HasValue(e ETag) bool {
```
jhendrixMSFT, line 49: I feel like `HasValue()` and `IsWeak()` should be methods on `ETag` (maybe `WeakEquals()` to but less sure on that one).  Was this discussed?

JeffreyRichter: Oh yes, you're right. These should be methods. I don't know how I missed this.

seankane-msft: Go doesn't allow methods on Types that are pointers, we need to have `type ETag struct` instead of `type ETag *string` to do that
https://golang.org/ref/spec#Method_declarations

jhendrixMSFT:
How do we plan to use this, hand-written code, generated code, both?

seankane-msft:
> How do we plan to use this, hand-written code, generated code, both?

@JeffreyRichter probably has more context on this than I do. In Tables this would be almost entirely on the hand-written side and we would dereference the pointer when sending to generated code so it still sees a pointer. The customer could use the comparison methods to determine whether to make a service call based on a received ETag in a previous operation.

JeffreyRichter:
I think we would use this in both generated and hand code. For example, it would be the data type for etag, If-match, & if-none-match headers.

sdk/azcore/etag.go
JeffreyRichter: I'd but the struct back then. The methods are better then functions.

jhendrixMSFT: Seems like extra complexity, I'd favor `type ETag string`

JeffreyRichter: Let's say a method takes an ETag value for the if-match header. How does the method know whether to add the header with the etag value or not add the header at all if ETag is a string?

sdk/azcore/etag.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Equals does a strong comparison of two ETags. Equals returns true when both
+// ETags are not weak and the values of the underlying strings are equal. If both ETags are "nil" they are considered equal
+func (e ETag) Equals(other ETag) bool {
+	return !e.IsWeak() && !other.IsWeak() && e == other
+}
+
+// WeakEquals does a weak compariosn of two ETags. Two ETags are equivalent if their opaque-tags match
+// character-by-character, regardless of either or both being tagged as "weak". If both ETags are "nil" they are considered equal
+func (e ETag) WeakEquals(other ETag) bool {
```
jhendrixMSFT, line 28: This doesn't handle the case of `nil` ETags per the comments.

Also, I'm leaning towards having the receivers passed by reference.  The reason being, if we are always defining ETags as `ETag *ETag` then calling this method will panic on a `nil` value.  If we make the receiver and parameter byref then we can perform the `nil` checks on the caller's behalf.  Thoughts?

JeffreyRichter: Fix spelling of "comparison". Remove doc sentence about "nil".
We don't want a pointer receiver because we want this to work like other fields in the struct where calling a method on nil panics - this is desired behavior.

jhendrixMSFT (approved):

PR #15216, opened 2021-08-05
[service-bus] Initial merge from azure-service-bus-go into the monorepo

richardpark-msft:
(CLOSED, moved to https://github.com/Azure/azure-sdk-for-go/pull/15244 to target a separate track 2 branch instead of 'main')

This is basically a verbatim merge (with `--allow-unrelated-histories`) of what's current in the azure-service-bus-go repo. I've moved it into a subfolder where it should reside, but have otherwise made no other source changes. 

For a short period of time we'll be working in both repos, but that should hopefully subside as we prep and release previews out of the monorepo.


sdk/servicebus/azservicebus/.gitignore
```diff
@@ -0,0 +1,31 @@
+# Binaries for programs and plugins
+*.exe
+*.dll
+*.so
+*.dylib
```
seankane-msft, line 5: You said this is a verbatim merge, but should we add the gitignore to the main gitignore at the project root?

sdk/servicebus/azservicebus/LICENSE
```diff
@@ -0,0 +1,21 @@
+    MIT License
```
seankane-msft, line 1: I'm no attorney but this license is slightly different than the license in azcore and we should probably be consistent.

seankane-msft (approved):

sdk/servicebus/azservicebus/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/Azure/azure-service-bus-go
```
jhendrixMSFT, line 1: This will need to be updated.

jhendrixMSFT:
Should this go into a branch instead of `main`?  It's how we've been doing other early track 2 modules.

sdk/servicebus/azservicebus/LICENSE
richardpark-msft: (I'm assuming this is the license you're talking about, btw)

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/LICENSE.txt

sdk/servicebus/azservicebus/go.mod
richardpark-msft: So this would be:

```
github.com/Azure/azure-sdk-for-go/sdk/service/servicebus/azservicebus
```

Is that the correct style for our modules? (oof, that's a long path)

seankane-msft: I believe `github.com/Azure/azure-sdk-for-go/sdk/servicebus/azservicebus` or `github.com/Azure/azure-sdk-for-go/sdk/azservicebus` could work too. I don't think you need the `/service/` portion.

sdk/servicebus/azservicebus/LICENSE
richardpark-msft: Updated so the contents are the same.

PR #15255, opened 2021-08-10
add clear output flag to build.ps1

ArcturusZhang:
Add a `clearOutput` flag in `build.ps1` flag to allow us first clear all the auto-generated files before we generate anything new.
The option of `autorest` `--clear-output-folder` is not acceptable because it will remove everything in this directory.

### DIFF
--- a/eng/scripts/build.ps1
+++ b/eng/scripts/build.ps1
@@ -1,5 +1,5 @@
 #Requires -Version 7.0
-param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]$skipBuild, [switch]$format, [switch]$tidy, [string]$config = "autorest.md", [string]$outputFolder)
+param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]$skipBuild, [switch]$clearOutput, [switch]$format, [switch]$tidy, [string]$config = "autorest.md", [string]$outputFolder)
 
 . $PSScriptRoot/meta_generation.ps1
 . $PSScriptRoot/get_module_dirs.ps1
@@ -9,6 +9,11 @@ function Process-Sdk ($path) {
     if ($clean) {
         Write-Host "##[command]Executing go clean -v ./... in " $path
         go clean -v ./...
+    }
+
+    if ($clearOutput) {
+        Write-Host "##[command]Cleaning auto-generated files in" $path
+        rm "zz_generated_*"
     }
 
     if ($generate) {
--- a/eng/scripts/build.ps1
+++ b/eng/scripts/build.ps1
@@ -5,20 +5,21 @@ param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]
 . $PSScriptRoot/get_module_dirs.ps1
 
 
-function Process-Sdk ($path) {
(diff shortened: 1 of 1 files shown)


eng/scripts/build.ps1
```diff
@@ -11,6 +11,11 @@ function Process-Sdk ($path) {
         go clean -v ./...
     }
 
+    if ($clearOutput) {
+        Write-Host "##[command]Cleaning auto-generated files in" $path
+        rm "zz_generated_*"
```
benbp, line 16: Only some of these commands actually seem to use `$path`, others (like this and `go clean` just use the current directory. I assume this is a bug?

ArcturusZhang: I actually do not know what the command `go clean` does here. I would defer @chamons to answer this and discuss whether we need that option.

I agree we should unify the usage of path here. But actually I do not like the way that we are passing the absolute path everywhere - how about we just remove the $path parameter from this function and make sure this function only takes effects in whatever the current directory is. The caller should ensure we are in the correct directory first, and then invoke this function, like this script does.

benbp: I'm realizing `$path` is probably more for being able to parse pipeline logs and understand which context things failed in, though that could probably also be achieved by logging directory changes when they happen.

I think the only consideration with absolute pathing is when you're running scripts to work against multiple project directories, error handling/control flow becomes a little simpler because you won't get dropped into subdirectories accidentally and you don't have to handle pushd/popd all over the place. Additionally if the scripts ever need to reference calls to other scripts/tools, then having one consistent script directory is more convenient (though I think just using `$PSScriptRoot` may solve that problem).

ArcturusZhang: The case here is that based on my understanding of how the go tools work (which might not be correct), some of them should either accept a relative directory (go vet for instance), or a full package identifier which is not necessary to be an absolute path. Therefore I prefer to use the relative path in the invocations of commands.
I added another variable to first get the absolute path of current directory once we go into the function so that we would have that in our verbose logging.

ArcturusZhang:
Hi @chamons @benbp could you please have a look again?
And @chamons could you please share something about the purpose of `go clean` in this script for me? What is this flag used for? It sounds kind of similar with what I am trying to introduce right now.

eng/scripts/build.ps1
```diff
@@ -1,19 +1,25 @@
 #Requires -Version 7.0
-param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]$skipBuild, [switch]$format, [switch]$tidy, [string]$config = "autorest.md", [string]$outputFolder)
+param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]$skipBuild, [switch]$clearOutput, [switch]$format, [switch]$tidy, [string]$config = "autorest.md", [string]$outputFolder)
```
benbp, line 2: Given this script is used in other contexts, a more descriptive parameter name like `cleanAutogeneratedFiles` or something might be better.

benbp (approved):

ArcturusZhang:
Thank you for @benbp 's review! I changed the flag name to `cleanGenerated` since this is shorter

PR #15256, opened 2021-08-10
[Tables] Add authentication via SAS credential

seankane-msft:


sdk/azcore/request.go
```diff
@@ -66,24 +66,34 @@ func (ov opValues) get(value interface{}) bool {
 	return ok
 }
 
+func stripQueryParams(path string) (string, string) {
```
JeffreyRichter, line 69: If this function is only ever called from inside JoinPaths, I'd move it in there. In fact, I probably wouldn't have it be a function and just move the code inside JoinPaths.

sdk/tables/aztable/sas_service.go
```diff
@@ -0,0 +1,156 @@
...
+	"bytes"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// TableSASSignatureValues is used to generate a Shared Access Signature (SAS) for an Azure Table instance.
+// For more information, see https://docs.microsoft.com/rest/api/storageservices/constructing-a-service-sas
+type TableSASSignatureValues struct {
+	Version           string      `param:"sv"`  // If not specified, this defaults to SASVersion
```
JeffreyRichter, line 16: Are these param tags actually used somewhere?

sdk/tables/aztable/sas_query_params.go
```diff
@@ -0,0 +1,292 @@
...
+	return
+}
+
+// https://docs.microsoft.com/en-us/rest/api/storageservices/constructing-a-service-sas
+
+// A SASQueryParameters object represents the components that make up an Azure Storage SAS' query parameters.
+// You parse a map of query parameters into its fields by calling NewSASQueryParameters(). You add the components
+// to a query parameter map by calling AddToValues().
+// NOTE: Changing any field requires computing a new SAS signature using a XxxSASSignatureValues type.
+type SASQueryParameters struct {
```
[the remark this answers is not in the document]
seankane-msft: snapshotTime should not be in there, but ipRange and protocol are used to [create the signature[(https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas#constructing-the-signature-string) and must remain (even if the user does not specify them).

PR #15259, opened 2021-08-10
[Internal] Add methods for configuring tests to work with test proxy

seankane-msft:
This PR adds the basic functionality for recording tests with the [testproxy](https://github.com/Azure/azure-sdk-tools/tree/main/tools/test-proxy) and a single sanitizer. The remaining sanitizers will be added in future PRs. This PR also downloads and starts the proxy as a background process running on ports 5000 and 5001.


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -4,67 +4,94 @@ parameters:
...
-    - pwsh: ../eng/scripts/create_coverage.ps1 ${{parameters.ServiceDirectory}}
-      displayName: 'Generate Coverage XML'
-      workingDirectory: '${{parameters.GoWorkspace}}sdk'
+      }
+      $(Build.SourcesDirectory)/eng/scripts/start-server.ps1 stop
+    displayName: 'Run Tests'
+    workingDirectory: '${{parameters.GoWorkspace}}'
+    env:
+      GO111MODULE: 'on'
+      AZURE_RECORD_MODE: 'playback'
```
benbp, line 67: It would be good to have this as a parameter that defaults to `playback` so that when we add live tests we can pass in `live` from the calling template.

benbp: But we can also just do that work when we get to it.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -4,67 +4,94 @@ parameters:
...
-    - task: PublishTestResults@2
-      condition: succeededOrFailed()
-      inputs:
-        testRunner: JUnit
-        testResultsFiles: '${{parameters.GoWorkspace}}sdk/**/report.xml'
-        testRunTitle: 'Go ${{ parameters.GoVersion }} on ${{ parameters.Image }}'
-        failTaskOnFailedTests: true
+  - pwsh: |
+      $coverageFiles = [Collections.Generic.List[String]]@()
+      Get-Childitem -recurse -path $(SCOPE) -filter coverage.txt | foreach-object {
```
benbp, line 71:
```suggestion
      Get-ChildItem -recurse -path $(SCOPE) -filter coverage.txt | ForEach-Object {
```

eng/scripts/start-server.ps1
```diff
@@ -0,0 +1,66 @@
+param(
```
benbp, line 1: I think a more descriptive name would be better here, e.g. `proxy-server.ps1` or something. Also calling it `start-server` makes it odd when you run `start-server.ps1 stop`

eng/pipelines/templates/steps/configure-proxy.yml
```diff
@@ -0,0 +1,18 @@
+parameters:
```
scbedd, line 1: There is a pretty excellent azure devops feature that I just was notified of. I'm going to see if I can use it.

seankane-msft: @scbedd do I still need this file or can I rely on the files in `eng/common/testproxy`?

scbedd: You should be able to delete this file now. I'm actually a bit disconcerted that your current setting of the PROXY_CERT variable isn't working properly.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -72,6 +70,7 @@ steps:
     displayName: 'Generate Coverage XML'
     workingDirectory: '${{parameters.GoWorkspace}}sdk'
 
+<<<<<<< HEAD
```
weshaggard: looks like a bad merge.

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
 	Live:     recorder.ModeDisabled,
 	Playback: recorder.ModeReplaying,
 }
+
+var recordMode, _ = os.LookupEnv("AZURE_RECORD_MODE")
```
richardpark-msft, line 442: If you're ignoring the second value I think you can just use os.Getenv()

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+		return "https://localhost:5001"
+	}
+	return "http://localhost:5000"
+}
+
+func getTestId(pathToRecordings string, t *testing.T) string {
+	return pathToRecordings + "/recordings/" + t.Name() + ".json"
+}
+
+func StartRecording(t *testing.T, pathToRecordings string, options *RecordingOptions) error {
```
richardpark-msft, line 492: Anything funky happen if the 'pathToRecordings' contains backslashes instead of forward slashes? Or would that be considered an error?

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+
+func (r RecordingOptions) HostScheme() string {
+	if r.UseHTTPS {
+		return "https://localhost:5001"
+	}
+	return "http://localhost:5000"
+}
+
+func getTestId(pathToRecordings string, t *testing.T) string {
+	return pathToRecordings + "/recordings/" + t.Name() + ".json"
```
richardpark-msft, line 489: Are we guaranteed that 'pathToRecordings' will never have a trailing slash?

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+
+func StartRecording(t *testing.T, pathToRecordings string, options *RecordingOptions) error {
+	if options == nil {
+		options = defaultOptions()
+	}
+	if recordMode == "" {
+		t.Log("AZURE_RECORD_MODE was not set, options are \"record\" or \"playback\". \nDefaulting to playback")
+		recordMode = "playback"
+	}
+	testId := getTestId(pathToRecordings, t)
```
richardpark-msft, line 500: Should probably validate that 'recordMode' is a valid value (ie, not just empty, but actually spelled correctly, etc..)

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+
+	url := fmt.Sprintf("%v/%v/start", options.HostScheme(), recordMode)
+
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("x-recording-file", testId)
```
richardpark-msft, line 509: Should x-recording-file be in a constant too? (noticed you did that with `IdHeader` below)

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+	Transport: tr,
+}
+
+type RecordingOptions struct {
+	UseHTTPS bool
+	Host     string
+	Scheme   string
+}
+
+func resetRecordingId() {
```
richardpark-msft, line 469: If I were to run tests in parallel (which I think is a good option for unit tests) would that break what you're doing?

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+}
+
+func StopRecording(t *testing.T, options *RecordingOptions) error {
+	defer resetRecordingId()
+
+	if options == nil {
+		options = defaultOptions()
+	}
+
+	url := fmt.Sprintf("%v/%v/stop", options.HostScheme(), recordMode)
```
richardpark-msft, line 526: Just checking here but are these all strings? If so, would it be more precise to use `%s`?

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+
+func LiveOnly(t *testing.T) {
+	if GetRecordMode() != modeRecording {
+		t.Skip("Live Test Only")
+	}
+}
+
+// Function for sleeping during a test for `duration` seconds. This method will only execute when
+// AZURE_RECORD_MODE = "record", if a test is running in playback this will be a noop.
+func Sleep(duration int) {
```
richardpark-msft, line 594: Instead of 'int' shouldn't it be a time.Duration?

richardpark-msft: What is the use case for this? Is it just to avoid polling?

sdk/internal/recording/recording.go
seankane-msft: This will change after every test so no it should not be a constant

sdk/internal/recording/recording.go
seankane-msft: It probably would. I'm not sure how to handle parallel tests right now and haven't tested it but I will add it as a todo. I agree it's something we should be able to handle.

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+func GetRecordingId() string {
+	return recordingId
+}
+
+func GetRecordMode() string {
+	return recordMode
+}
+
+func InPlayback() bool {
+	return GetRecordMode() == modePlayback
```
richardpark-msft, line 609: (NIT) If something is this uncomplicated it can be nice to just inline it, rather than adding a function over it.

In this case you've done a great job with having named constants, and the name of the function itself (GetRecordMode) is very clear.

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+
+func InRecord() bool {
+	return GetRecordMode() == modeRecording
+}
+
+func getRootCas() (*x509.CertPool, error) {
+	localFile, ok := os.LookupEnv("PROXY_CERT")
+
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
```
richardpark-msft, line 620: Might not be a huge deal but is there a specific error that you're ignoring here (like system cert pool not found)? Or is it literally _any_ error?

If there is a specific one it'd be good to just check that, and return the error otherwise (perhaps there's something dire going on).

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+	}
+
+	if !ok {
+		fmt.Println("Could not find path to proxy certificate, set the environment variable 'PROXY_CERT' to the location of your certificate")
+		return rootCAs, nil
+	}
+
+	cert, err := ioutil.ReadFile(localFile)
+	if err != nil {
+		fmt.Println("error opening cert file")
```
richardpark-msft, line 631: Do we need to fmt.Println here? Would fmt.Errorf() be more appropriate? (also, including the file name could be helpful in diagnosing errors).

sdk/internal/recording/recording.go
```diff
@@ -433,3 +438,221 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+		return nil, err
+	}
+
+	if ok := rootCAs.AppendCertsFromPEM(cert); !ok {
+		fmt.Println("No certs appended, using system certs only")
```
richardpark-msft, line 636: Should we call 'log' instead?

richardpark-msft: (or whatever our equivalent is here in testing code)

sdk/internal/recording/recording_test.go
```diff
@@ -60,6 +62,7 @@ func (s *recordingTests) TestStopDoesNotSaveVariablesWhenNoVariablesExist() {
 }
 
 func (s *recordingTests) TestRecordedVariables() {
+	s.T().Skipf("Skipping flaky test")
```
richardpark-msft, line 65: Do we have some idea of what the flakiness is?

sdk/internal/recording/recording_test.go
```diff
@@ -368,7 +372,148 @@ func (s *recordingTests) TestRecordRequestsAndFailMatchingForMissingRecording()
 }
 
 func (s *recordingTests) TearDownSuite() {
-	// cleanup test files
-	err := os.RemoveAll("recordings")
-	require.Nil(s.T(), err)
+	files, err := filepath.Glob("recordings/**/*.yaml")
```
richardpark-msft, line 375: Is there any way to just put the temp data into a single folder and os.RemoveAll() that? Not a huge deal, but would make cleaning up hand simpler as well.

sdk/internal/recording/recording_test.go
```diff
@@ -368,7 +372,148 @@ func (s *recordingTests) TestRecordRequestsAndFailMatchingForMissingRecording()
...
+	err = os.Setenv("PROXY_CERT", "not/a/path.crt")
+	require.NoError(t, err)
+	_, err = GetHTTPClient()
+	require.Error(t, err)
+
+	os.Setenv("PROXY_CERT", tempProxyCert)
+}
+
+func TestStopRecordingNoStart(t *testing.T) {
+	require := require.New(t)
```
richardpark-msft, line 500: That's neat!

sdk/internal/recording/recording.go
seankane-msft: Polling is one scenario. In Tables there are features that take up to 30 seconds to propagate and there's no need to wait in record mode for these scenarios.

sdk/internal/recording/recording.go
seankane-msft: All of this information is included in the error from `ioutil` which covers the error case nicely.

sdk/internal/recording/recording_test.go
seankane-msft: This is from the legacy recording framework and will hopefully be removed altogether in the near future

sdk/internal/recording/recording_test.go
seankane-msft: Saves a bit of code, sometimes I use it and sometimes I don't. No real pattern and I can't figure out if I like the option or not.

eng/config.json
```diff
@@ -10,7 +10,7 @@
         },
         {
             "Name": "internal",
-            "CoverageGoal": 0.90
+            "CoverageGoal": 0.83
```
RickWinter, line 13: We should not drop the coverage once its at 0.90.  Add new tests to keep the coverage above 90

RickWinter (changes requested):

sdk/internal/recording/recordings/TestUriSanitizer.json
```diff
@@ -0,0 +1,60 @@
...
+        "(function() { var conditionalSignInParams ={\u0022notifEnabled\u0022:false,\u0022notifFetchAsync\u0022:false}; BingAtWork.ConditionalSignIn \u0026\u0026 BingAtWork.ConditionalSignIn.bindToConditionalSignIn \u0026\u0026 BingAtWork.ConditionalSignIn.bindToConditionalSignIn(conditionalSignInParams); })();;(function() { var config ={\u0022url\u0022:\u0022https%3a%2f%2flogin.microsoftonline.com%2fcommon%2foauth2%2fauthorize%3fclient_id%3d9ea1ad79-fdb6-4f9a-8bc3-2b70f96e34c7%26response_type%3did_token%2bcode%26nonce%3d2ffee435-6f0f-4238-8eec-12b296d9638c%26redirect_uri%3dhttps%253a%252f%252fwww.bing.com%252forgid%252fidtoken%252fconditional%26scope%3dopenid%26response_mode%3dform_post%26msafed%3d0%26prompt%3dnone%26state%3d%257b%2522ig%2522%253a%252281D4A8D7A8BC437CA9971ACDE3401706%2522%257d\u0022,\u0022sandbox\u0022:\u0022allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts\u0022,\u0022currentEpoch\u0022:\u00221629906927000\u0022}; sj_evt.fire(\u0027ssoFrameExists\u0027, config); })();;\n",
+        "//]]\u003E\u003C/script\u003E\u003C/div\u003E\u003C/body\u003E\u003C/html\u003E"
+      ]
+    }
+  ],
+  "Variables": {}
+}
```
RickWinter, line 60:
```suggestion
}

```

sdk/internal/recording/recordings/TestStartStop.json
```diff
@@ -0,0 +1,60 @@
...
+        "(function() { var conditionalSignInParams ={\u0022notifEnabled\u0022:false,\u0022notifFetchAsync\u0022:false}; BingAtWork.ConditionalSignIn \u0026\u0026 BingAtWork.ConditionalSignIn.bindToConditionalSignIn \u0026\u0026 BingAtWork.ConditionalSignIn.bindToConditionalSignIn(conditionalSignInParams); })();;(function() { var config ={\u0022url\u0022:\u0022https%3a%2f%2flogin.microsoftonline.com%2fcommon%2foauth2%2fauthorize%3fclient_id%3d9ea1ad79-fdb6-4f9a-8bc3-2b70f96e34c7%26response_type%3did_token%2bcode%26nonce%3de7627bed-8863-43dd-b8c6-b71eab6a1684%26redirect_uri%3dhttps%253a%252f%252fwww.bing.com%252forgid%252fidtoken%252fconditional%26scope%3dopenid%26response_mode%3dform_post%26msafed%3d0%26prompt%3dnone%26state%3d%257b%2522ig%2522%253a%2522DB5D90E3764548E492F0279FC0C5B536%2522%257d\u0022,\u0022sandbox\u0022:\u0022allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts\u0022,\u0022currentEpoch\u0022:\u00221629906926000\u0022}; sj_evt.fire(\u0027ssoFrameExists\u0027, config); })();;\n",
+        "//]]\u003E\u003C/script\u003E\u003C/div\u003E\u003C/body\u003E\u003C/html\u003E"
+      ]
+    }
+  ],
+  "Variables": {}
+}
```
RickWinter, line 60:
```suggestion
}

```

sdk/internal/recording/recordings/TestUriSanitizer.json
seankane-msft: These are generated by the test-proxy so I am inclined to leave them as is. I'll reach out to scott B about what would go in here.

sdk/internal/recording/recording.go
richardpark-msft: I can see how the value changes, but the header name is a string constant, right?

eng/config.json
```diff
@@ -10,7 +10,7 @@
         },
         {
             "Name": "internal",
-            "CoverageGoal": 0.83
+            "CoverageGoal": 0.90
```
richardpark-msft: This here is a positive indicator. :)

sdk/internal/recording/recording.go
```diff
@@ -486,20 +482,22 @@ func (r RecordingOptions) HostScheme() string {
 }
 
 func getTestId(pathToRecordings string, t *testing.T) string {
+	if strings.HasSuffix(pathToRecordings, "/") {
```
richardpark-msft, line 485: Go has a nice module for the path joining - not a huge deal but will save you some trouble in the future: https://pkg.go.dev/path@go1.17#Join

sdk/internal/recording/recording.go
```diff
@@ -486,20 +482,22 @@ func (r RecordingOptions) HostScheme() string {
...
 
 func StartRecording(t *testing.T, pathToRecordings string, options *RecordingOptions) error {
 	if options == nil {
 		options = defaultOptions()
 	}
-	if recordMode == "" {
-		t.Log("AZURE_RECORD_MODE was not set, options are \"record\" or \"playback\". \nDefaulting to playback")
-		recordMode = "playback"
+	if !(recordMode == modeRecording || recordMode == modePlayback) {
+		return fmt.Errorf("AZURE_RECORD_MODE was not understood, options are \"record\" or \"playback\". Received: %v", recordMode)
```
richardpark-msft, line 496: No biggie, but could '%s' and use the actual constants in your error message since you're already Errorf'ing.

sdk/internal/recording/recording.go
```diff
@@ -512,13 +510,19 @@ func StartRecording(t *testing.T, pathToRecordings string, options *RecordingOpt
...
 		return err
 	}
-	recordingId = resp.Header.Get(IdHeader)
+	recId := resp.Header.Get(IdHeader)
+	if recId == "" {
+		b, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("Recording ID was not returned by the response. Response body: %s", b)
```
richardpark-msft, line 519: NIT: would it be helpful to say that the specific header was not found? Could aid in diagnostics if someone is trying to figure out what's wrong strictly by the error message.

sdk/internal/recording/recording_test.go
```diff
@@ -517,3 +516,45 @@ type RecordingFileStruct struct {
...
+	if GetRecordMode() == modePlayback {
+		t.Fatalf("Test should not run in playback")
+	}
+}
+
+func TestSleep(t *testing.T) {
+	start := time.Now()
+	Sleep(time.Second * 5)
+	duration := time.Since(start)
+	if GetRecordMode() == modePlayback {
```
richardpark-msft, line 531: This might not be an issue with Go's version of this, but my experience with 'timer' related functions is that they're not always as exact as you'd want.

Given that, being too exact with the time limit might make this test flaky.

I think what you really care about here is not the length of time (which honestly you can't screw up) and more that it definitely called sleep (and the time doesn't matter as much). So anything > 1 second, for instance, could be considered 'success'.

richardpark-msft (approved):
Just some minor things. 

The only one you might want to adjust is the timer test but I think this is good enough to checkin otherwise.

RickWinter (approved):

PR #15271, opened 2021-08-12
[tools/generator] add automation and release command for track2

tadelesh:


tools/generator/cmd/track2/common/fileProcessor.go
```diff
@@ -0,0 +1,222 @@
...
+		} else if changelog.Modified.HasAdditiveChanges() {
+			minor += 1
+			patch = 0
+		} else {
+			patch += 1
+		}
+	}
+
+	log.Printf("New version is: %d.%d.%d", major, minor, patch)
+	return fmt.Sprintf("%d.%d.%d", major, minor, patch), nil
```
ArcturusZhang, line 204: We actually have a dependency `github.com/MasterMind/semver` to deal semantic versions, we can leverage on that

tools/generator/cmd/track2/release/sdk.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: I believe this is not track 2 particular definitions. Could we move this to a more common place?

tools/generator/cmd/track2/release/spec.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: Same as above

tools/generator/cmd/track2/release/workTree.go
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ArcturusZhang, line 1: Same as above

ArcturusZhang (commented):
I left some comments, please take a look

tadelesh:
I have tried sdk automation in test branch. PR is in https://github.com/openapi-env-test/azure-rest-api-specs/pull/2967. @ArcturusZhang

swagger_to_sdk_config_track2.json
```diff
@@ -0,0 +1,28 @@
...
+  "initOptions": {
+    "initScript": {
+      "path": "sh ./initScriptTrack2.sh",
+      "stderr": {
+        "scriptWarning": false
+      }
+    }
+  },
+  "packageOptions": {
+    "breakingChangeLabel": "CI-BreakingChange-Go"
```
ArcturusZhang, line 26: We might need to add a new label for track 2
```suggestion
    "breakingChangeLabel": "CI-BreakingChange-Go-Track2"
```

eng/scripts/build.ps1
```diff
@@ -63,7 +63,7 @@ $root = Resolve-Path ($PSScriptRoot + "/../..")
 Set-Location $root
 $sdks = @{};
 
-foreach ($sdk in (Get-ModuleDirs 'sdk/...')) {
+foreach ($sdk in (Get-ModuleDirs 'sdk/')) {
```
ArcturusZhang, line 66: Why do we have to change this?

tadelesh: I think the function is to find all module dirs in sdk subfolders. So the three dots in path are not necessary and also Get-ChildItem with three dots path not work in linux environment (ci base image is ubuntu).

swagger_to_sdk_config_track2.json
```diff
@@ -0,0 +1,28 @@
+{
```
weshaggard, line 1: Lets try to put this under the eng folder instead of the root directory.

swagger_to_sdk_config_track2.json
```diff
@@ -0,0 +1,28 @@
...
+  "$schema": "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/documentation/sdkautomation/SwaggerToSdkConfigSchema.json",
+  "generateOptions": {
+    "generateScript": {
+      "path": "generator automation-track2",
```
weshaggard, line 5: We have been trying to avoid the name codename "track2" from many places in our repo and instead we have been trying to use the version of the code generator. Can we please try to come up with a naming convention that doesn't include "track2" in all the names of files and configurations?

tools/generator/cmd/root.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/Azure/azure-sdk-for-go/tools/generator/cmd/automation"
```
weshaggard, line 10: I know the tools repo already exists but as a future improvement it would be ideal if we could try to move these under the eng folder if possible.

swagger_to_sdk_config_track2.json
ArcturusZhang: How about `new-version` like the branch name in `azure-sdk-for-go-samples`?

tools/generator/cmd/root.go
ArcturusZhang: That is on our backlog to convert everything to powershell and move them to `eng` directory (track 1 related things excluded).
We are putting these here because we just want to implement this in the fastest way to make it running. We will translate this to powershell and move it to `eng` directory

swagger_to_sdk_config_track2.json
tadelesh: According to @ArcturusZhang , I've changed the name in script. Cause the files have been moved to eng folder, the filename is no need to user track2 suffix now.

initScriptTrack2.sh
```diff
@@ -0,0 +1,19 @@
+set -x
```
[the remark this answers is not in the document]
tadelesh: OK, I've move the script and config for sdk automation to eng folder.

swagger_to_sdk_config_track2.json
tadelesh: Same as above.

swagger_to_sdk_config_track2.json
weshaggard: While "new-version" is better then "track2" it is still a little shortsighted because what happens when we have the next new version? If we have a version number for the generator that would be preferred and that is what other languages are doing. If there isn't an explicit version perhaps we just call this v2 so that we have a version for it.

Also there looks to still be a few references to track2 in files, like the directory. Lets consider changing those to v2 instead.

tadelesh: v2 is a resonable name. I've changed all the track2 to v2.

tools/generator/go_mod_tidy_hack.go
```diff
@@ -1,3 +1,4 @@
+//go:build modhack
 // +build modhack
```
ArcturusZhang, line 2: @jhendrixMSFT could you please take a look at this change? Is this expected as in go 1.17?
And should we do the same for all other released SDK modules in their next release?

tools/generator/cmd/v2/common/fileProcessor.go
```diff
@@ -0,0 +1,252 @@
...
+
+const (
+	sdk_generated_file_prefix         = "zz_generated_"
+	autorest_md_swagger_url_prefix    = "- https://github.com/Azure/azure-rest-api-specs/blob/"
+	autorest_md_module_version_prefix = "module-version: "
+	swagger_md_module_name_prefix     = "module-name: "
+)
+
+var (
+	v2BeginRegex                   = regexp.MustCompile("^```\\s*yaml\\s*\\$\\(go\\)\\s*&&\\s*\\$\\(track2\\)")
```
weshaggard, line 29: I'm not entirely if this is everywhere in readmes already but if possible we should change this to v2 as well instead of track2.

tadelesh: For now, all readme in swagger use 'track2' to indicate the new sdk version config and also some other scripts use 'track2' to recognize this config. So I just reserve the 'track2' in regex expression and also add 'v2' string for future gradually convertion.

Gopkg.toml
```diff
@@ -19,7 +19,7 @@
 #  name = "github.com/x/y"
 #  version = "2.4.0"
 
-ignored = ["github.com/ahmetb/go-linq/*", "github.com/google/go-github/*"]
+ignored = ["github.com/ahmetb/go-linq/*", "github.com/google/go-github/*", "github.com/go-git/go-git/v5", "github.com/go-git/go-git/v5/*"]
```
ArcturusZhang, line 22: Maybe we could just simplify this to 
```suggestion
ignored = ["github.com/ahmetb/go-linq/*", "github.com/google/go-github/*", "github.com/go-git/go-git/*"]
```

tools/generator/cmd/v2/release/releaseCmd.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/Azure/azure-sdk-for-go/tools/generator/flags"
+	"github.com/Azure/azure-sdk-for-go/tools/generator/repo"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// Release command
+func Command() *cobra.Command {
+	releaseCmd := &cobra.Command{
+		Use:   "release-v2 <azure-sdk-for-go directory> <azure-rest-api-specs directory> <rp-name> <namespaceName>",
```
ArcturusZhang, line 22:
```suggestion
		Use:   "release-v2 <azure-sdk-for-go directory> <azure-rest-api-specs directory> <rp-name> [namespaceName]",
```
since namespaceName is optional

ArcturusZhang (commented):
some minor comments

ArcturusZhang (approved):

PR #15272, opened 2021-08-12
Add template script

ArcturusZhang:
This PR adds a template for mgmt plane packages and introduces a script to apply the template with the given parameters.

### DIFF
--- a/eng/template/CHANGELOG.md.tpl
+++ b/eng/template/CHANGELOG.md.tpl
@@ -1,0 +1,3 @@
+# Release History
+
+## v0.1.0 (released)
--- a/eng/template/README.md.tpl
+++ b/eng/template/README.md.tpl
@@ -1,0 +1,76 @@
+# Azure {{PackageTitle}} Module for Go
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/{{rpName}}/{{packageName}})](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/{{rpName}}/{{packageName}})
+
+The `{{packageName}}` module provides operations for working with Azure {{PackageTitle}}.
+
+[Source code](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/{{rpName}}/{{packageName}})
+
+# Getting started
+
+## Prerequisites
+
+- an [Azure subscription](https://azure.microsoft.com/free/)
+- Go 1.13 or above
+
+## Install the package
+
+This project uses [Go modules](https://github.com/golang/go/wiki/Modules) for versioning and dependency management.
+
+Install the Azure {{PackageTitle}} module:
+
+```sh
+go get github.com/Azure/azure-sdk-for-go/sdk/{{rpName}}/{{packageName}}
+```
+
+## Authorization
+
+When creating a client, you will need to provide a credential for authenticating with Azure {{PackageTitle}}.  The `azidentity` module provides facilities for various ways of authenticating with Azure including client/secret, certificate, managed identity, and more.
+
+```go
+cred, err := azidentity.NewDefaultAzureCredential(nil)
+```
+
+For more information on authentication, please see the documentation for `azidentity` at [pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity).
+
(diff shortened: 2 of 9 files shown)


eng/template/CHANGELOG.md.tpl
```diff
@@ -0,0 +1,3 @@
+# Release History
+
+## v0.1.0 (released)
```
weshaggard, line 3: Can you help me understand why this is 0.1.0? In all other languages we start at 1.0.0-beta.1 and unless there is a strong reason I'd suggest we do that in go as well.

eng/template/README.md.tpl
```diff
@@ -0,0 +1,76 @@
+# Azure {{PackageTitle}} Module for Go
+
+[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/{{rpName}}/{{packageName}})](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/{{rpName}}/{{packageName}})
```
weshaggard, line 3: Based on our other conversations this will always end up being a broken link for new libraries. Have we came up with a strategy for that?

eng/template/autorest.md.tpl
```diff
@@ -0,0 +1,10 @@
+### AutoRest Configuration
+
+> see https://aka.ms/autorest
+
+``` yaml
+require:
+- https://github.com/Azure/azure-rest-api-specs/blob/{{commitID}}/specification/{{rpName}}/resource-manager/readme.md
```
weshaggard, line 7: Will these links actually exist initially?

eng/template/ci.yml.tpl
```diff
@@ -0,0 +1,15 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
+trigger:
```
weshaggard, line 2: Lets be sure to add the branch names to these triggers as well.

weshaggard:
I see that this assumes mgmt libraries how much would it take to also add support for data-plane libraries as well? 

cc @RickWinter @jhendrixMSFT

eng/template/CHANGELOG.md.tpl
ArcturusZhang: This might be the golang way to represent a pre-release before v1.0.0 - but I would like to defer @jhendrixMSFT to provide some more details with more accuracy.

eng/template/README.md.tpl
ArcturusZhang: Unfortunately no. We do not really want to ship this without the link - this will be the link to package references

eng/template/autorest.md.tpl
ArcturusZhang: Yes, these must exist, otherwise the `autorest` command will complain it cannot get these files.

eng/template/CHANGELOG.md.tpl
weshaggard: While golang does support a "0.x.x" versioning scheme so does a lot of other languages but we have been trying to not use that in other places as it tends to make people believe we are a far away from anything they can depend on. golang does support the prerelease versioning scheme as well according to https://golang.org/doc/modules/version-numbers#pre-release so unless there is strong reason we should be consistent with our versioning polices and go with 1.0.0-beta.1 as the first version (https://azure.github.io/azure-sdk/policies_releases.html). @jhendrixMSFT I know we talked about this a little in the past but I'd like to revisit this given we are going to be having a large influx of new libraries.

eng/template/README.md.tpl
weshaggard: We at least need to pick a strategy for folks because this will break in CI for every new library we use it on.

ArcturusZhang: I once made a change to the ci.yml template to make sure we could go over every step in the build analysis job by moving the link verification to the last step). In this way, when we are onboarding a new RP, we can ensure that if the only failure in this PR is the link verification, and the RP is generated from a specific template, this PR should be fine and the invalid link will be good after the PR is merged and the tag is added. This is only current practical way to do service onboarding.

weshaggard: I suppose that is one option but it does mean that you will have to override any PR with changes to that file before it is release the first time. It isn't ideal and if we get the link wrong or things change/rename before we release we risk shipping a broken link here. 

I personally like leaving this commented out by default and after our first release we can uncomment it.

PR #15276, opened 2021-08-12
[Tables] Track2 tables

seankane-msft:


sdk/tables/aztable/connection_string.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// NewClientFromConnectionString creates a new Client struct from a connection string. The connection
+// string must contain either an account name and account key or an account name and a shared access signature.
+func NewClientFromConnectionString(connectionString string, tableName string, options *ClientOptions) (*Client, error) {
```
JeffreyRichter, line 15: I'd remove this function

sdk/tables/aztable/options.go
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JeffreyRichter, line 1: As a code hygiene issue, it is better to not have an options.go file at all. It is better to put the option structs near the methods that use them. The reason is because a change to the option struct means there is a change to the method and so they should be near each other.

sdk/tables/aztable/responses.go
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JeffreyRichter, line 1: Samt thing here: there should not be a response.go file; the response structures should with the methods that need them.

sdk/tables/aztable/table_pagers.go
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JeffreyRichter, line 1: Same thing here: there should not be a table_pagers.go; these constructs should be near the methods than require them.

JeffreyRichter (approved):

sdk/tables/aztable/connection_string.go
seankane-msft: Are we expecting a user to parse the connection string themself?

sdk/tables/aztable/table_pagers.go
seankane-msft: Will condense this, I did this to avoid bloat in the client and service_client files but I can see why this is better.

sdk/tables/aztable/connection_string.go
JeffreyRichter: No, keep the other NewServiceClientFromConnectionString function; just kill this one.
If they have a cxn str then they have to create the service client first and then they can call the method to get back a Client. We're not enabling anything with NewClientFromCxnStr that can't be done another way.

PR #15280, opened 2021-08-13
[service-bus] track 2 ServiceBusClient + ServiceBusProcessor

richardpark-msft:
The ServiceBusProcessor is one of the bigger features in Track 2 - it allows users to subscribe to a subscription or a queue and receive messages without having to worry about error handling. Errors are reported (also via a callback) but the processor internally takes care of retries, connections, etc... Users don't have to do anything more with errors. This greatly simplifies the getting started (and even long term) use cases for Service Bus.

This is just a first revision to lay some groundwork for how the processor will work, create a stress test [stress.go](https://github.com/Azure/azure-sdk-for-go/blob/47fd9f4ffc9da4d26b67f9650da86202351ec20a/sdk/servicebus/azservicebus/stress/stress.go) and a bunch of tests.

I've also included a [sample](https://github.com/Azure/azure-sdk-for-go/blob/47fd9f4ffc9da4d26b67f9650da86202351ec20a/sdk/servicebus/azservicebus/samples/processor/processor_sample.go)


sdk/servicebus/azservicebus/client.go
```diff
@@ -0,0 +1,103 @@
...
+	namespace *internal.Namespace
+	linksMu   *sync.Mutex
+	links     []interface {
+		Close(ctx context.Context) error
+	}
+}
+
+// ServiceBusClientOption is the type for an option that can configure ServiceBusClient.
+// For an example option, see `ServiceBusWithConnectionString`
+type ServiceBusClientOption func(client *ServiceBusClient) error
```
jhendrixMSFT, line 23: While I would prefer to use the variadic config pattern, it would be different than our other track 2 SDKs as well as our design guidelines.  I think we need to have a discussion about where we want to land on this.

sdk/servicebus/azservicebus/processor.go
```diff
@@ -0,0 +1,388 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/servicebus/azservicebus/internal"
+	"github.com/Azure/azure-sdk-for-go/sdk/servicebus/azservicebus/internal/utils"
+)
+
+var ErrReceiverClosed = errors.New("receiver has been closed and can no longer be used")
```
jhendrixMSFT, line 13: This should go in errors.go.

sdk/servicebus/azservicebus/samples/processor/processor_sample.go
```diff
@@ -0,0 +1,106 @@
...
+		// - a nil will cause us to Complete the message on the user's behalf.
+		// - a non-nil error will cause to Abandon the message. The error will also be forwarded
+		//   to their error handler below.
+		return nil
+	}, func(err error) {
+		// a customer can reasonably expect to see some errors here when the processor recovers
+		// from connection errors, or if some automated operations failed (like autocomplete
+		// settlement failure)
+
+		if err == context.Canceled {
```
jhendrixMSFT, line 79: You want to be using [`errors.Is()`](https://pkg.go.dev/errors#Is) to handle potential error wrapping.

jhendrixMSFT:
Conceptually, it sounds very similar to [`Receiver.Listen()`](https://github.com/Azure/azure-service-bus-go/blob/master/receiver.go#L214) in track 1.  Would this replace that functionality or is this slightly different?

richardpark-msft:
> Conceptually, it sounds very similar to [`Receiver.Listen()`](https://github.com/Azure/azure-service-bus-go/blob/master/receiver.go#L214) in track 1. Would this replace that functionality or is this slightly different?

It is similar in some ways (in that they're both callback-based) but it's intended to be a replacement. This first rev is just built on top because that's convenient, but there are TODOs in place to mark spots that I know are 'incorrect' to track 2, at least.

The big differences:
- There is only a single way to stop the listening (via `processor.Stop()`). Errors returned in the individual handle callbacks do not terminate the listen.
- The processor _never_ exits because of errors - it reports them (via the error callback) but it always restarts itself or maintains whatever it needs to continue listening. Users should, for the most part, be able to consider errors as cosmetic (ie: report via telemetry) but not fatal.
- This version would also allow credits to be added even if messages have not been settled. It's purely based on the callback completing.

Some other behavioral differences:
- Autocomplete isn't entirely new (there was a default disposition before) but it now matches the way that the other track 2 SDKs do it.
- MaxConcurrentCalls is distinct from prefetching. This will require some go-amqp changes that aren't yet done (but you and I have talked about).

There are probably more but this is just off the top of my head.

sdk/servicebus/azservicebus/client.go
richardpark-msft: Is the alternative just taking a `struct` as a parameter? I would have loved to do that (although the queue vs topic+subscription) might be a bit painful.

Who do we kick this discussion off with?

jhendrixMSFT: Yeah a struct.  https://azure.github.io/azure-sdk/golang_introduction.html#optional-parameters

We really need @JeffreyRichter to be here for this discussion.  You can leave it as-is for now, but this might change depending on how the conversation goes.

sdk/servicebus/azservicebus/processor.go
richardpark-msft: Moved.

sdk/servicebus/azservicebus/samples/processor/processor_sample.go
richardpark-msft: Done.

sdk/servicebus/azservicebus/client.go
richardpark-msft: Sounds good.

sdk/servicebus/azservicebus/processor.go
```diff
@@ -0,0 +1,395 @@
...
+//
+//   if err := processor.Start(messageHandler, errorHandler); err != nil {
+//     log.Fatalf("Processor failed to start: %s", err.Error())
+//   }
+//
+//   <- processor.Done()
+//
+// Any errors that occur (such as network disconnects, failures in handleMessage) will be
+// sent to your handleError function. The processor will retry and restart as needed -
+// no user invention is required.
```
seankane-msft, line 185: should this read "no user intervention is required"?

sdk/servicebus/azservicebus/processor.go
```diff
@@ -0,0 +1,395 @@
...
+func (p *Processor) DeferMessage(ctx context.Context, message *ReceivedMessage) error {
+	return message.legacyMessage.Defer(ctx)
+}
+
+func formatEntity(entity struct {
+	Subqueue     SubQueue
+	Queue        string
+	Topic        string
+	Subscription string
+}) (string, error) {
```
seankane-msft, line 349: Curiosity: why not have this struct defined outside of the method signature?

sdk/servicebus/azservicebus/receiver.go
```diff
@@ -0,0 +1,24 @@
+package azservicebus
+
+type ReceiveMode int
+
+const (
+	ReceiveModePeekLock         ReceiveMode = 0
+	ReceiveModeReceiveAndDelete ReceiveMode = 1
+)
```
seankane-msft, line 8: Do these need to have the "ReceiveMode" prefix? It makes the consts even longer and the second almost has stuttering

sdk/servicebus/azservicebus/samples/processor/processor_sample.go
```diff
@@ -0,0 +1,106 @@
+package main
```
seankane-msft, line 1: There is a samples directory inside of `sdk`, do we want to put samples there or are we moving to having samples in each module?

sdk/servicebus/azservicebus/processor.go
richardpark-msft: Yes, that makes more sense! :)

sdk/servicebus/azservicebus/processor.go
richardpark-msft: Just trying to avoid creating excess types when they're only used once, or in a really limited scope..

jhendrixMSFT (approved):

sdk/servicebus/azservicebus/receiver.go
richardpark-msft: I removed the 'ReceiveMode' part as you suggested. 

I'm not entirely sure if they're needed or not - it makes discovery slightly easier if you're just going off the package but perhaps it's not needed since the 'type' is there and they're grouped together in the source.

@jhendrixMSFT - is there a standard around this one?

sdk/servicebus/azservicebus/samples/processor/processor_sample.go
richardpark-msft: That's a great question! 

@jhendrixMSFT , @RickWinter - should I move these into `sdk/samples`? 

We had both kinds of samples in JS - the package samples would demo the package, and the repo level samples were more general and covered things like browser bundling, etc...

richardpark-msft:
@seankane-msft  - not ignoring your feedback, but I think the bulk of the code is decent enough to check in.

I think we have the larger question of where samples are intended to go. I can move my samples to wherever we decide to make the final spot.

PR #15291, opened 2021-08-13
Rename log constants

jhendrixMSFT:


sdk/internal/log/log.go
```diff
@@ -15,20 +15,20 @@ import (
 type Classification string
 
 const (
-	// Request entries contain information about HTTP requests.
+	// RequestClassification entries contain information about HTTP requests.
 	// This includes information like the URL, query parameters, and headers.
-	Request Classification = "Request"
+	RequestClassification Classification = "Request"
```
jhendrixMSFT, line 20: I'm really on the fence about this.  The names are pretty long now (I had to rename the LRO one to keep it reasonable).  Plus, I think the help docs are clear on the usage.

JeffreyRichter: How about rename
Classification to
Category   -- shorter
Group       -- ??  even shorter. For example: I want to see all log entries in the RequestGroup?
Or, I'm open to an even shorter word if you can think of one

jhendrixMSFT: Entry maybe?

jhendrixMSFT:
```go
type Entry string

const (
    RequestEntry Entry = "Request"
    ResponseEntry Entry = "Response"
)
```

Then in user code.
```go
log.SetEntries(log.RequestEntry, log.ResponseEntry)
```

jhendrixMSFT: Maybe that should be `log.SetEntriesFilter()` or something similar.

JeffreyRichter: I thinkk you have to make it plural (ex: RequestEntries) because it logs all entries.
But, what if we used "Events" instead: RequestEvents. It's common to "log events".
Then, log.EventTypes(log.RequestEvents, log.ResponseEvents)

JeffreyRichter (approved):

sdk/internal/log/log.go
jhendrixMSFT: I like events, works for me.  I'll make this change after I complete some of the other refactoring.

RickWinter (approved):

JeffreyRichter:
Sounds good to me.


  *   Jeffrey Richter<https://www.linkedin.com/in/jeffrichter/> (see Architecting Distributed Cloud Apps<http://aka.ms/RichterCloudApps> & Designing HTTP APIs<https://www.youtube.com/watch?v=9Ng00IlBCtw&list=PL9XzOCngAkqs4m0XdULJu_78nM3Ok3Q65&index=1>)

PR #15294, opened 2021-08-13
Remove Response type

jhendrixMSFT:
Just return an *http.Response instead.
Changed Response methods to functions.


JeffreyRichter (approved):
This code is OK but all the methods that operate on a response are now function-level instead of scoped to that type. 

Personally, I would have defined a private:
type azresponse http.Response 
and then kept all our METHODS on the azresponse type and then just converted the http.response to an azresponse to call a method, for example:
azresponse(resp).Drain()

jhendrixMSFT:
Agreed on the scoping.  I was envisioning that they'd move to the `runtime` package since these are mostly used by codegen.

JeffreyRichter:
That's fine then azresponse & its methods would be in runtime.
runtime.azresponse(resp).Drain()

PR #15324, opened 2021-08-19
Add generator Tools release and refresh

804873052:


RickWinter:
Is the a document describing the workflow that these tools follow?  Also I'm interested to learn more about why Go was picked as the language to write this automation.  The central EngSys team uses powershell for scripting and automation.

804873052:
This is migrating the remaining commands from the private tool to public repo (track1).

ArcturusZhang:
> Is there a document describing the tool workflow? Also I'm interested to learn more about why Go was picked as the language to write this automation. The central EngSys team uses powershell for scripting and automation.

These tools are only for track 1 SDK generation. Previously we have a specific private tool built for generating new releases for track 1 SDK to avoid too much manual work. We are trying to move that tool to this public place so that our customer could also try it if they want to generate their own private release of track 1 SDKs.

We are also building new tools for track 2 releases as well.
We understand we should be keeping everything in `eng` directory and write things in powershell, but there are things we do not find a powershell equivalent yet (for instance, generate changelog for the SDK packages). Therefore we are also implementing that in go in the `tools/generator` directory. But we will migrate that to powershell and `eng` directory after we get the prototype running.

ArcturusZhang (approved):
Please `goimports` all of the new files

PR #15325, opened 2021-08-19
Add template projet to verify automation changes

RickWinter:


ArcturusZhang:
[My PR](https://github.com/Azure/azure-sdk-for-go/pull/15272) also includes a template but for mgmt plane packages. I suppose we should keep the templates together and away from the real SDKs to reduce the confusions.
DotNet SDK keeps the template for data plane in the `sdk` directory, but template for mgmt plane in `eng` directory. But I suppose it is because the template for mgmt plane is not ready. It will make more sense if both of the templates can be found in the same directory.

Also I added a suffix for all the files in the template so that the template itself will not be indexed by `pkg.go.dev`. Do you think this is a concern?

sdk/template/aztemplate/ci.yml
```diff
@@ -0,0 +1,19 @@
...
+
+pr:
+  paths:
+    include:
+    - sdk/template/aztemplate/
+    - eng/
+
+
+stages:
+- template: ../../eng/pipelines/templates/jobs/archetype-sdk-client.yml
```
ArcturusZhang, line 16:
```suggestion
- template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
```

PR #15328, opened 2021-08-19
AsyncCache

### LINKED ISSUE #15327: AsycCache

ausfeldt:
PR for visibility. Still need to handle errors.

Closes #15327


sdk/cosmos/azcosmos/async_cache.go
```diff
@@ -0,0 +1,142 @@
...
+	fn       cacheValueTask
+	ch       <-chan interface{}
+}
+
+func newAsyncCache() *asyncCache {
+	return &asyncCache{}
+}
+
+func (ac *asyncCache) set(key interface{}, value interface{}) {
+	ac.values.Store(key, cacheValue{value: value})
```
ealsur, line 34: Does the sync.Map take care of concurrent sets for the same key?

sdk/cosmos/azcosmos/async_cache.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (ac *asyncCache) setAsync(key interface{}, singleValueInit cacheValueTask, ctx context.Context) error {
+	ch := ac.execCacheValueTask(singleValueInit)
```
ealsur, line 38: How does the flow work on concurrent calls to setAsync? Wouldn't both create a channel and both call Store?

ealsur (approved):
Nothing blocking, we will probably find out more with usage, just trying to understand how certain code paths work

sdk/cosmos/azcosmos/async_cache.go
```diff
@@ -0,0 +1,161 @@
...
+	_, err := ac.awaitCacheValue(key, ctx)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (ac *asyncCache) getValue(key interface{}) (interface{}, bool) {
```
ealsur, line 56: nit: the sets are set and setAsync, for consistency, get and getAsync

sdk/cosmos/azcosmos/async_cache.go
ausfeldt: sync.Map should take care of concurrent sets as it uses a RWMutex under the hood.

sdk/cosmos/azcosmos/async_cache.go
ausfeldt: set is synchronous and changed the name to reflect this.  set is used to initialize the cache. getAsync is used to asynchronously get the value from the cache and update the value depending on the obsolete value passed.

PR #15331, opened 2021-08-20
Prevent ManagedIdentityCredential mutating GetToken arguments

### LINKED ISSUE #15308: ManagedIdentityCredential should not do in-place mutation of Scopes
### Bug Report

`ManagedIdentityCredential` receives a list of scopes as input and internally converts them to a resource ID (#14693).

There's a problem that `ManagedIdentityCredential` is actually mutates the original array, which creates an unexpected side-effect for the calling code:

```go

var credential azcore.TokenCredential
var token *azcore.AccessToken
var err error

clientId := ""
scopes := []string { "htttps://management.azure.com/.default" }

if credential, err = azidentity.NewManagedIdentityCredential(clientId, nil); err != nil {
	return err
}

fmt.Printf("scopes: %s\n", scopes)
// scopes: [htttps://management.azure.com/.default]

if token, err = c.credential.GetToken(ctx, azcore.TokenRequestOptions{Scopes: scopes}); err != nil {
	return err
}

fmt.Printf("scopes: %s\n", scopes)
// scopes: [htttps://management.azure.com] <-- INPUT ARRAY HAS CHANGED
```

[This line](https://github.com/Azure/azure-sdk-for-go/blob/dcc9108ed8db94db491f10a08535e0b27a1a3087/sdk/azidentity/managed_identity_credential.go#L99) is responsible for in-place mutation of the input array.

Mutation should not happen since conversion from scopes to resource ID is implementation detail which should not affect calling code.

chlowell:
Closes #15308

### DIFF
--- a/sdk/azidentity/managed_identity_credential.go
+++ b/sdk/azidentity/managed_identity_credential.go
@@ -95,9 +95,9 @@ func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts azcore.To
 		addGetTokenFailureLogs("Managed Identity Credential", err, true)
 		return nil, err
 	}
-	// The following code will remove the /.default suffix from any scopes passed into the method since ManagedIdentityCredentials expect a resource string instead of a scope string
-	opts.Scopes[0] = strings.TrimSuffix(opts.Scopes[0], defaultSuffix)
-	tk, err := c.client.authenticate(ctx, c.id, opts.Scopes)
+	// managed identity endpoints require an AADv1 resource (i.e. token audience), not a v2 scope, so we remove "/.default" here
+	scopes := []string{strings.TrimSuffix(opts.Scopes[0], defaultSuffix)}
+	tk, err := c.client.authenticate(ctx, c.id, scopes)
 	if err != nil {
 		addGetTokenFailureLogs("Managed Identity Credential", err, true)
 		return nil, err
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* `ManagedIdentityCredential.GetToken` no longer mutates its `opts.Scopes`
 
 ### Other Changes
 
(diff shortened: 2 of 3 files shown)


sdk/azidentity/managed_identity_credential.go
```diff
@@ -95,9 +95,9 @@ func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts azcore.To
 		addGetTokenFailureLogs("Managed Identity Credential", err, true)
 		return nil, err
 	}
-	// The following code will remove the /.default suffix from any scopes passed into the method since ManagedIdentityCredentials expect a resource string instead of a scope string
-	opts.Scopes[0] = strings.TrimSuffix(opts.Scopes[0], defaultSuffix)
-	tk, err := c.client.authenticate(ctx, c.id, opts.Scopes)
+	// managed identity endpoints require an AADv1 resource (i.e. token audience), not a v2 scope, so we remove "/.default" here
+	scopes := []string{strings.TrimSuffix(opts.Scopes[0], defaultSuffix)}
```
chlowell, line 99: I think it's okay to allocate a slice here unconditionally because we'll normally need to trim the suffix but I'm curious, do we prefer to limit such allocations generally?

RickWinter: It seems totally reasonable here b/c we need to mutate the input.  @jhendrixMSFT, Is there design guidance surrounding this, should there be?

RickWinter (approved):

sdk/azidentity/managed_identity_credential.go
jhendrixMSFT: Depending on escape analysis, `scopes` might or might not be allocated on the heap.  Given that it's pretty small, I would guess that it's allocated on the stack given that it's passed to callers but doesn't escape the callee, however you'd have to profile to be sure.

Generally speaking, the language is opaque WRT when things are stack vs heap allocated as this can change as the compiler improves.  The general guidance is to write the code with the semantics you want, potentially changing it based on profiling (obviously there are exceptions but this is a good place to start).

jhendrixMSFT (approved):

PR #15346, opened 2021-08-20
add manual tag and release pipeline for track2

tadelesh:


tadelesh:
tested in personal repo
image
image

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,98 @@
+$Language = "go"
+$packagePattern = "go.mod"
+$RELEASE_TITLE_REGEX = "(?<releaseNoteTitle>^\#+\s+(?<version>v$([AzureEngSemanticVersion]::SEMVER_REGEX))(\s+(?<releaseStatus>\(.+\))))"
+$AUTOREST_VERSION_REGEX = "^module-version:\s+(?<version>$([AzureEngSemanticVersion]::SEMVER_REGEX))"
+
+function Get-go-PackageInfoFromPackageFile ($pkg, $workingDirectory)
```
chidozieononiwu, line 6: Looks like this method can also serve as `Get-go-PackageInfoFromRepo`, sicnce go only pulls from the repo.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,98 @@
...
+        ReleaseNotes   = $releaseNotes
+    }
+
+    return $resultObj
+}
+
+# get version from go config auterest.md
+function Get-Version ($pkgPath)
+{
+    $content = Get-Content(Join-Path $pkgPath "autorest.md")
```
chidozieononiwu, line 36: It ook s like this only applies to a handfull of management packages. Shoul there also be logic for retrieving version from version.go files?

eng/scripts/Language-Settings.ps1
tadelesh: Yes, Go release process has no package process and all package info can get directly from repo. But in order to reuse 'create-tag-and-git-release.ps1' script in eng sys  which use Get-go-PackageInfoFromPackageFile to get release info, I create this function by tradeoff.

eng/scripts/Language-Settings.ps1
tadelesh: I have changed the Get-Version method to support both data plane and mgmt plane. For mgmt plane of go track2, version can be obtained by autorest.md. For data plane, version can be obtained by version.go file. Maybe two version aquisition method can be unified if code gen add version.go to mgmt plane in the future. @jhendrixMSFT @ArcturusZhang

eng/pipelines/templates/jobs/archetype-go-release.yml
```diff
@@ -0,0 +1,33 @@
+parameters:
+  DependsOn: Build
+  ServiceDirectory: ''
+
+stages:
+  - ${{if and(eq(variables['Build.Reason'], 'Manual'), eq(variables['System.TeamProject'], 'internal'))}}:
```
benbp, line 6: I think this conditional is unnecessary since you're also including it in the template include for the archetype-sdk-client.yml file.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
...
+        ReleaseNotes   = $releaseNotes
+    }
+
+    return $resultObj
+}
+
+# get version from specific files (*constants.go, *version.go)
+function Get-Version ($pkgPath)
+{
+    # find any file with surfix
```
benbp, line 40:
```suggestion
    # find any file with suffix
```

benbp (approved):

eng/scripts/Language-Settings.ps1
weshaggard: We should definitely share and reuse this logic for both the packagefile and packageinfo.

@chidozieononiwu at some point I want to eliminate FromPackageFile function completely and either use PackageInfo or use the PackageInfo.json file we are generating as part of the build.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
...
+    $versionFiles = [Collections.Generic.List[String]]@()
```
weshaggard, line 41:
```suggestion
    $versionFiles = @()
```
We shouldn't need to create create a specific list type.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
...
+            Write-Error "Error parsing version."
+            Write-Error $_
+        }
+    }
+
+    Write-Host "Cannot find release version."
+    exit 1
+}
+
+# rewrite from artifact-metadata-parsing.ps1 as all 0.x.x version should be considered prerelease in go sdk
```
weshaggard, line 77: FYI I'm adding support for v0 versions in our core SemVer helper with https://github.com/Azure/azure-sdk-tools/pull/1972. If that is included do we still need to duplicate this entire function? Hopefully not.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
+$Language = "go"
+$packagePattern = "go.mod"
+# rewrite from ChangeLog-Operations.ps1 used in Get-ChangeLogEntriesFromContent for go uses vx.x.x as version number
+$RELEASE_TITLE_REGEX = "(?<releaseNoteTitle>^\#+\s+(?<version>v$([AzureEngSemanticVersion]::SEMVER_REGEX))(\s+(?<releaseStatus>\(.+\))))"
+# constants for go version fetch
+$GO_VERSION_REGEX = ".+\s*=\s*`".*v?(?<version>$([AzureEngSemanticVersion]::SEMVER_REGEX))`""
+$VERSION_FILE_SUFFIXS = "*constants.go", "*version.go"
```
weshaggard, line 7: Given we are only using this in the one place below I'd suggest moving it there and also I wouldn't capitalize the variable name in this case as that is more for constants.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
+$Language = "go"
+$packagePattern = "go.mod"
+# rewrite from ChangeLog-Operations.ps1 used in Get-ChangeLogEntriesFromContent for go uses vx.x.x as version number
+$RELEASE_TITLE_REGEX = "(?<releaseNoteTitle>^\#+\s+(?<version>v$([AzureEngSemanticVersion]::SEMVER_REGEX))(\s+(?<releaseStatus>\(.+\))))"
```
weshaggard, line 4: @chidozieononiwu we consider adding an option "v?" to our core regex instead of duplicating this. We do similar support for python in our Regex already which supports different prerelease labels.

eng/scripts/Language-Settings.ps1
```diff
@@ -0,0 +1,116 @@
...
+$GO_VERSION_REGEX = ".+\s*=\s*`".*v?(?<version>$([AzureEngSemanticVersion]::SEMVER_REGEX))`""
+$VERSION_FILE_SUFFIXS = "*constants.go", "*version.go"
+
+# rewrite from artifact-metadata-parsing.ps1 used in RetrievePackages for fetch go single module info
+function Get-go-PackageInfoFromPackageFile ($pkg, $workingDirectory)
+{
+    $workFolder = $pkg.Directory
+    $releaseNotes = ""
+    $namespaceName = $workFolder.Name
+    $rpName = $workFolder.Parent.Name
```
weshaggard, line 15: I think this will not work for some packages like azcore or azidentity which doesn't have a service folder currently

weshaggard: You might want to consider a regex similar to https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L148.

eng/scripts/Language-Settings.ps1
tadelesh: It will be great that semver helper can handle this. I'll remove it after your pr merged and automatic sync done.

eng/scripts/Language-Settings.ps1
tadelesh: I wonder whether there's already a definitely plan for this version compatibility change. Could we use the rewrite workaround before the eventually solution cause there are lots of release work for mgmt plane sdk refreshment recently.

eng/scripts/Language-Settings.ps1
weshaggard: My SemVer change is in.

eng/scripts/Language-Settings.ps1
weshaggard: There isn't a plan for this currently and I think what you have is an acceptable workaround for now.

weshaggard (approved):
If possible remove your CreateReleases function assuming it isn't needed after my SemVer change. Otherwise the changes look like a good starting point and we can tweak as we start to use it.

Issue #15057, opened 2021-07-13
NewResponseError should be internal


RickWinter:
### Feature Request

Customers do not need to use this method.  This should be in the package that SDK authors consume

jhendrixMSFT:
Fixed in `v0.19.0`

Issue #15055, opened 2021-07-13
RetryAfter move to codegen


RickWinter:
### Feature Request

Investigate if RetryAfter should live with CodeGen

jhendrixMSFT:
Fixed in `v0.19.0`

Issue #15053, opened 2021-07-13
Move Decode/Encode APIs


RickWinter:
### Feature Request

The decode/encode apis are for advanced scenarios.  Move to a less common place

jhendrixMSFT:
Fixed in `v0.19.0`

Issue #15052, opened 2021-07-13
JoinPaths API should be part of the codegen module


RickWinter:
### Feature Request

Move join path to CodeGen

jhendrixMSFT:
Fixed in `v0.19.0`

PR #15349, opened 2021-08-20
Refactor azcore

jhendrixMSFT:
See the CHANGELOG for details.


jhendrixMSFT:
Fixes:
- https://github.com/Azure/azure-sdk-for-go/issues/15018
- https://github.com/Azure/azure-sdk-for-go/issues/15048
- https://github.com/Azure/azure-sdk-for-go/issues/15050
- https://github.com/Azure/azure-sdk-for-go/issues/15051
- https://github.com/Azure/azure-sdk-for-go/issues/15052
- https://github.com/Azure/azure-sdk-for-go/issues/15053
- https://github.com/Azure/azure-sdk-for-go/issues/15055
- https://github.com/Azure/azure-sdk-for-go/issues/15057

jhendrixMSFT:
The accompanying codegen changes have been made and tested to help validate the changes.  In addition, some ARM and data-plane LRO scenarios have been tested as well.

sdk/azcore/CHANGELOG.md
```diff
@@ -1,13 +1,36 @@
...
 
+## v0.19.0
+
+### Breaking Changes
+* Split content out of `azcore` into various packages.  The intent is to separate content based on its usage (common, uncommon, SDK authors).
+  * `azcore` has all core functionality.
+  * `log` contains facilities for configuring in-box logging.
+  * `policy` is used for configuring pipeline options and creating custom pipeline policies.
+  * `runtime` contains various helpers used by SDK authors and generated content.
+  * `streaming` has helpers for streaming IO operations.
```
jhendrixMSFT, line 11: I'm not totally in love with the name `streaming`, so if anybody has a better idea please post them here.  `io` would really be ideal by that collides with the standard library.

sdk/azcore/example_test.go
```diff
@@ -39,12 +41,12 @@ func ExamplePipeline_Do() {
 }
 
 func ExampleRequest_SetBody() {
```
jhendrixMSFT, line 43: Besides the fact the linter is complaining, I wonder if we should remove this example (and the one for Do above).  Customers will never call either of these methods, it's purely codegen.

JeffreyRichter (approved):

sdk/azcore/example_test.go
jhendrixMSFT: Went ahead and removed these examples.

sdk/azcore/version.go
```diff
@@ -1,15 +0,0 @@
-//go:build go1.16
```
RickWinter, line 1: Keeping a version.go file for each back and sub-package is cleaner.  Makes it easy for our automation and for customers to quickly identify package versions.

jhendrixMSFT: The module versions as an atomic unit, not sub-packages.  The reason the version info moved to `internal/shared` is because it's used across multiple packages.

jhendrixMSFT: Also, it was a conscious decision to _not_ expose these values in the code.

sdk/azcore/CHANGELOG.md
```diff
@@ -1,13 +1,36 @@
...
+* The `*http.Request` in `runtime.Request` is no longer anonymously embedded.  Use the `Raw()` method to access it.
+* The `UserAgent` and `Version` constants have been made internal, `Module` and `Version` respectively.
+
+### Bug Fixes
+* Fixed an issue in the retry policy where the request body could be overwritten after a rewind.
+
+### Other Changes
+* Moved modules `armcore` and `to` content into `arm` and `to` packages respectively.
+  * The `Pipeline()` method on `armcore.Connection` has been replaced by `NewPipeline()` in `arm.Connection`.  It takes module and version parameters used by the telemetry policy.
+* Poller logic has been consolidated across ARM and core implementations.
```
RickWinter, line 24: Nit, Should the package be "pollers" vs something more generic related to polling/lros?

sdk/azcore/arm/poller.go
```diff
@@ -0,0 +1,83 @@
+//go:build go1.16
```
RickWinter, line 1: Should this be part of the pollers package?

sdk/azcore/internal/shared/constants.go
```diff
@@ -0,0 +1,34 @@
...
+	DefaultMaxRetries = 3
+)
+
+const (
+	// Module is the name of the calling module used in telemetry data.
+	Module = "azcore"
+
+	// Version is the semantic version (see http://semver.org) of this module.
+	Version = "v0.19.0"
+)
```
RickWinter, line 34: version.go at the root is a better place for this.  Customers may want to access the package / version information for their logging.

sdk/azcore/policy_anonymous_credential.go
```diff
@@ -6,13 +6,19 @@
```
RickWinter, line 6: Is it expected customers will use this?  If not we should place in the azcore/policy/... package

sdk/azcore/poller.go
```diff
@@ -7,442 +7,66 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/pollers/op"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
 // NewPoller creates a Poller based on the provided initial response.
 // pollerID - a unique identifier for an LRO, it's usually the client.Method string.
 // NOTE: this is only meant for internal use in generated code.
-func NewPoller(pollerID string, resp *http.Response, pl Pipeline, eu func(*http.Response) error) (*Poller, error) {
+func NewPoller(pollerID string, resp *http.Response, pl runtime.Pipeline, eu func(*http.Response) error) (*Poller, error) {
```
RickWinter, line 25: Does the customer call NewPoller?  If so I don't like that we expose "runtime" package to them.

sdk/azcore/to/to.go
```diff
@@ -0,0 +1,107 @@
+//go:build go1.16
+// +build go1.16
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
RickWinter, line 5: Add a comment into this file that this package will be deprecated once Go adds generics

sdk/azcore/version.go
RickWinter: I'm interested as to why we decided not to export.  That is different than other languages.

RickWinter (approved):

sdk/azcore/internal/shared/constants.go
jhendrixMSFT: We can't place this in the root as it's used across two different packages.

sdk/azcore/policy_anonymous_credential.go
jhendrixMSFT: Yes, anonymous credential is first-class for SDKs that support SAS authorization.

sdk/azcore/to/to.go
jhendrixMSFT: The package won't be deprecated, just all the per-type functions.  We'll end up with one generic `To[T any](t T) *T` function.

sdk/azcore/poller.go
jhendrixMSFT: I was thinking about this during the refactor (I meant to talk to Jeff about this yesterday but forgot).  We want to expose the core and ARM `Poller` types via their "root" packages `azcore` and `azcore/arm` respectively.  The usual convention is to place their constructors next to them, however it's true that we don't expect customer to call them.  However, if we were to move them to `runtime`, it gets a little weird as we now have two sets of constructors, so we'd have to rename the ARM ones to avoid name collisions.

@JeffreyRichter thoughts on this?

JeffreyRichter: Customers should not be newing up a Poller directly. Since a Poller requires an XxxClient (to poll against), either a customer will call an XxxClient method to create the new poller or we would have some package-level function that takes an xxxClient and returns the Poller.

sdk/azcore/arm/poller.go
jhendrixMSFT: The creation of ARM pollers has been moved to package `azcore/arm/runtime` to mimic what we do in core.

sdk/azcore/poller.go
jhendrixMSFT: This is done.  I also applied the same pattern to the creation of ARM pollers, so we have symmetry between the two and keeps the uncommon stuff out of the happy path.

ArcturusZhang (changes requested):
May I ask why we need to merge `armcore` into `azcore`? In other language's SDK, we usually have two different library of `Azure.Core` and `Azure.ResourceManager.Core`. This might introduce inconsistency comparing with other SDKs

jhendrixMSFT:
@ArcturusZhang we want to merge them to reduce the number of dependencies.  In addition, this should version better over time.  With the old model, we could find ourselves in a position where `armcore` uses a different major version than other libraries which could be problematic.

PR #15352, opened 2021-08-20
Cosmos DB - Adding errors on failed responses

### LINKED ISSUE #15351: Cosmos DB - error on failed responses
Currently the response from the pipeline in a case of an error just returns the error response with a failure status code. 

We should be exposing them as actual errors leveraging `azcore.NewResponseError`

ealsur:
This PR converts error responses (statusCode < 200 or statusCode >= 300 excluding 304) to proper errors, following the pattern provided by `azcore.NewResponseError`.

The error includes the raw response and attempts to deserialize the Cosmos DB error body and expose it in the `Error()` call.

Closes #15351

### DIFF
--- a/sdk/cosmos/azcosmos/cosmos_error.go
+++ b/sdk/cosmos/azcosmos/cosmos_error.go
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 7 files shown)


sdk/cosmos/azcosmos/cosmos_error.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// cosmosError is used as base error for any error response from the Cosmos service.
+type cosmosError struct {
+	body        string
+	code        string
+	rawResponse *http.Response
+}
+
+func (e *cosmosError) StatusCode() string {
```
jhendrixMSFT, line 20: This should return an `int`, the value being `rawResponse.StatusCode`

ausfeldt (approved):

PR #15370, opened 2021-08-24
bumping Build_Test to 1.16 and 1.17

seankane-msft:
Build_Test was running older versions of Go, bumping to be in line with `eng/pipelines/templates/jobs/archetype-sdk-client.yml` versions

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -96,7 +96,7 @@ jobs:
       workingDirectory: '$(sdkPath)'
       displayName: 'Format Check'
       failOnStderr: true
-      condition: succeededOrFailed()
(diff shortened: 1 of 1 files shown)


jhendrixMSFT (approved):

azure-pipelines.yml
```diff
@@ -55,40 +57,63 @@ jobs:
         go get -u golang.org/x/lint/golint
       workingDirectory: '$(sdkPath)'
       displayName: 'Install Dependencies'
+
     - script: go vet -v $(go list ./... $(go.list.filter))
       workingDirectory: '$(sdkPath)'
       displayName: 'Vet'
+      env:
+        GO111MODULE: 'off'
```
jhendrixMSFT, line 65: I believe you can add this to the `variables` section so you don't have to repeat it.

PR #15377, opened 2021-08-25
Return CredentialUnavailableError when IMDS has no assigned identity

### LINKED ISSUE #14214: Don't fail auth chain if one of the methods work
Hi!

I've been hitting the same "issue" over and over now with the new azidentity package. This should most likely be classified as a feature request and not a bug, since I do think it's by design.

When I develop something using the azidentity and test it locally, I usually test it with one of the methods configured (none of them excluded). Often Azure CLI or Environment - and if both are configured they usually both have the correct permissions / configuration.

The problem I'm hitting is when running in Azure DevOps or GitHub Actions is that they also have MSI - and that MSI usually doesn't have the permissions I need. Then I get this issue:
`Received an AuthenticationFailedError, there is an invalid credential in the chain.  details: invalid_request Identity not found`

It's quite easy to solve, I just make the Exclude option for each of them configurable and disable MSI in the pipelines.

It would be nice if we had a way to actually see that there are other ways available and use them instead.

As an example:
- There is an MSI, but won't be able to authentication
- There is an active Azure CLI session, that is working

If it fails the MSI but Azure CLI works - the chain should be seen as "succesful".

chlowell:
IMDS is always available to a VM in that it responds to requests but, when the VM has no assigned identity, it responds 400 to token requests: IMDS is available but managed identity authentication is not. ManagedIdentityCredential doesn't make this distinction today. It considers managed identity available whenever IMDS responds and treats the lack of an assigned identity as an authentication error, i.e. it returns ClientAuthenticationError. That error causes DefaultAzureCredential (ChainedTokenCredential) to halt when it should instead try the next credential in the chain. With this PR ManagedIdentityCredential does that and, after receiving a 400 from IMDS, doesn't attempt another token request. The motivation for the latter change is that there's no reason to believe another token request would succeed and, if it did succeed because someone assigned an identity while an application is running, that application could unexpectedly change its authenticated identity.

Closes #14214

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -60,6 +60,7 @@ type managedIdentityClient struct {
 	msiType              msiType
 	endpoint             string
 	id                   ManagedIdentityIDKind
+	unavailableMessage   string
 }
 
 type wrappedNumber json.Number
@@ -92,6 +93,10 @@ func newManagedIdentityClient(options *ManagedIdentityCredentialOptions) *manage
 // clientID: The client (application) ID of the service principal.
 // scopes: The scopes required for the token.
 func (c *managedIdentityClient) authenticate(ctx context.Context, clientID string, scopes []string) (*azcore.AccessToken, error) {
+	if len(c.unavailableMessage) > 0 {
+		return nil, &CredentialUnavailableError{credentialType: "Managed Identity Credential", message: c.unavailableMessage}
+	}
+
 	msg, err := c.createAuthRequest(ctx, clientID, scopes)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -103,6 +108,11 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, clientID strin
 		return c.createAccessToken(resp)
 	}
 
+	if c.msiType == msiTypeIMDS && resp.StatusCode == 400 {
```
jhendrixMSFT, line 111: What happens in the case of user-assigned identity auth where the user-assigned ID was incorrectly specified, or is that different from IMDS?

chlowell: IMDS responds the same when it can't access a user-assigned or system-assigned identity for any reason. "You specified the wrong ID" gets the same treatment as "managed identity is disabled for this VM". We could distinguish those cases here by assuming that if we have a value for `clientID`, the user intended to authenticate a user-assigned identity; then a 400 indicates misconfiguration of the application or the VM and, arguably, we should return ClientAuthenticationError.

I write "arguably" because we don't do that in other SDKs. I suspect that's an oversight, not by design--I'll have to think through possible consequences a little more. Probably this has just never come up. For a deployed application any managed identity error is typically fatal to authentication because the application is using either ManagedIdentityCredential alone or DefaultAzureCredential, whose remaining credentials integrate with dev tools.

PR #15412, opened 2021-08-30
Add Cosmos Account Region struct

jay-most:
The AccountRegion struct represents an Azure Cosmos DB database account in a specific region.

- [ X ] The purpose of this PR is explained in this or a referenced issue.
- [ X ] The PR does not update generated files.


sdk/cosmos/azcosmos/cosmos_account_region_properties.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+// The AccountRegion struct represents an Azure Cosmos DB database account in a specific region.
+type AccountRegion struct {
+	// Gets the name of the database account in the Azure Cosmos DB service.
+	Name string `json:"Constants.Properties.Name"`
```
ealsur, line 9: How does this json deserialization work? `json:....` is meant to map to the JSON payload attribute name.

sdk/cosmos/azcosmos/cosmos_account_region_properties.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+// The AccountRegion struct represents an Azure Cosmos DB database account in a specific region.
+type AccountRegion struct {
```
ealsur, line 7: Name-wise, we might want to align with the Java Track2 SDK: https://github.com/Azure/azure-sdk-for-java/blob/f8f3980d1e86d3b4983bf38477adc7e07ceab443/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/DatabaseAccountLocation.java#L11

ealsur (commented):
What is the purpose of this PR and how/where is this type used?

sdk/cosmos/azcosmos/cosmos_account_region_properties.go
ealsur: For example, if the Json is `{'name':'xxxxxx'}` then Go's Json comment would be `json: name`. See https://pkg.go.dev/encoding/json#Marshal

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -45,7 +46,7 @@ steps:
       foreach ($td in $testDirs) {
         pushd $td
         Write-Host "##[command]Executing go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic $td | go-junit-report -set-exit-code > report.xml"
-        go test -run "^Test" -race -v -coverprofile coverage.txt -covermode atomic . | go-junit-report -set-exit-code > report.xml
+        go test -run "^Test" -tags=${{parameters.Tags}} -race -v -coverprofile coverage.txt -covermode atomic . | go-junit-report -set-exit-code > report.xml
```
seankane-msft, line 49: When `Tags` is empty are there any issues with this command? Looks like this is only used for the emulator

Issue #15409, opened 2021-08-30
Move EngSys scripts and tools into the eng folder


RickWinter:
There are currently yaml files and various scripts at the root of the repo.  These should be moved into the Eng folder.

All of these moves should be coordinated with mgmt sdk to avoid disruption to their process.

- [ ] Move `azure-pipelines.yml` into `eng` folder
- [ ] Move `tools` folder under `eng`
- [ ] Move all `*.sh` files to `eng` folder
- [ ] Move all `*.json` files to `eng` folder

jhendrixMSFT:
Yeah `rungas.sh` can be deleted.

benbp:
> 1. Despite we already have a `swagger_to_sdk_config.json` in `eng` directory, but one of them is for track 1 SDK (the one in root directory), and the other is for track 2 SDK. Therefore we need to keep them both.
> 2. We have some external reference for these two configurations [here](https://github.com/Azure/azure-rest-api-specs/blob/51b37b069ecbb9d2fcd300eabd4b10b7911b7d7d/specificationRepositoryConfiguration.json#L7), we need to change those accordingly after we have done the refactor.

Looking at other repos, I realize this is not an uncommon pattern, so I don't see a need right now to delete the root swagger config for consistency reasons right now.

> 1. We move the `internal` module to `eng/tools` and change all the references accordingly. Replace it using relative path to make sure everything could work
> 2. We move all the tools to `eng/tools`. In this step we do not need to do any code change, just remove the replace clause and use the new internal module.

Sounds good. I have a PR up for step 1 at #15460.

PR #15453, opened 2021-09-02
Clean up/organize script and json files in top level directory

benbp:
Part of https://github.com/Azure/azure-sdk-for-go/issues/15409


weshaggard:
Places like https://github.com/Azure/azure-sdk-for-go/blob/main/swagger_to_sdk_config.json#L19 also need to be updated.

benbp:
> Places like https://github.com/Azure/azure-sdk-for-go/blob/main/swagger_to_sdk_config.json#L19 also need to be updated.

Thanks. Originally I had that file removed so I missed it when adding back.

swagger_to_sdk_config.json
```diff
@@ -16,7 +16,7 @@
   },
```
weshaggard, line 16: @tadelesh @ArcturusZhang as a potential further clean-up we can move this config file under eng directory as well we will just need to update the path at https://github.com/Azure/azure-rest-api-specs/blob/main/specificationRepositoryConfiguration.json#L5.

tadelesh: @benbp @weshaggard There is two versions of swagger_to_sdk_config.json. One under repo root (work with initScript.sh) is for track1 and the other under eng folder (work with eng/script/automation_init.sh) is for track2. We can aggregate and rename them, in the meantime, change the sdk automation config in swagger repo.

benbp: I had opted not to move this json file because most other repositories have one at the root, so I didn't see it was breaking convention like the other files.

weshaggard: I'm fine with punting on thus for now but I've definitely been trying to get the other languages to move those config files under the eng folder but I agree it isn't as critical currently and requires other changes outside of this repo to actually move.

PR #15459, opened 2021-09-02
[Tables] Add samples

seankane-msft:


sdk/samples/aztables/README.md
```diff
@@ -0,0 +1,145 @@
+---
+page_type: sample
+languages:
+  - go
+products:
+  - azure
+  - aztables
```
richardpark-msft, line 7: I think this 'product' is not just specific to Go, right? Do you know what the other languages are using for this tag?

richardpark-msft: (it could very well be 'aztables', I just wanted to check)

sdk/samples/aztables/README.md
```diff
@@ -0,0 +1,145 @@
...
+|[sample_create_client.go][create_client]|Instantiate a table client|Authorizing a `ServiceClient` object and `Client` object |
+|[sample_create_delete_table.go][create_delete_table]|Creating and deleting a table in a storage account |
+|[sample_insert_delete_entities.go][insert_delete_entities]|Inserting and deleting individual entities in a table |
+|[sample_query_tables.go][query_tables]|Querying tables in a storage account |
+|[sample_update_upsert_merge_entities.go][update_upsert_merge]| Updating, upserting, and merging entities |
+|[sample_batching.go][sample_batch]| Committing many requests in a single batch |
+
+
+### Prerequisites
+* Go 1.14 or later is required to use this package.
```
richardpark-msft, line 31: 1.16 is our minimum now, isn't it?

sdk/samples/aztables/sample_authentication.go
```diff
@@ -0,0 +1,87 @@
...
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	check(err)
+	client, err := aztables.NewServiceClient(serviceURL, cred, nil)
+	check(err)
+	return client
+}
+
+func AuthenticateWithSharedAccess() {
+	// TODO
```
richardpark-msft, line 52: Probably better to just not include this at all (since it's a sample) and just file an issue to fill it in.

sdk/samples/aztables/sample_batch.go
```diff
@@ -0,0 +1,111 @@
...
+			RowKey:       "rk002",
+		},
+		Price:       19.99,
+		Inventory:   15,
+		ProductName: "Calculators",
+		OnSale:      false,
+	}
+	marshalled, err = json.Marshal(entity2)
+	check(err)
+	batch[1] = aztables.TransactionAction{
```
richardpark-msft, line 70: Wouldn't a more typical pattern be to append to a slice?

sdk/samples/aztables/sample_batch.go
```diff
@@ -0,0 +1,111 @@
...
+		Entity: aztables.Entity{
+			PartitionKey: "pk001",
+			RowKey:       "rk003",
+		},
+		Price:       0.99,
+		Inventory:   150,
+		ProductName: "Pens",
+		OnSale:      true,
+	}
+	marshalled, err = json.Marshal(entity3)
```
richardpark-msft, line 85: I'm sure this API has already been through the wringer _but_ is there anything in our library that just automatically handles marshalling the entity? Or are these steps always required?

seankane-msft: These steps are always required

sdk/samples/aztables/sample_authentication.go
```diff
@@ -0,0 +1,82 @@
...
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	check(err)
+	client, err := aztables.NewServiceClient(serviceURL, cred, nil)
+	check(err)
+	return client
+}
+
+func main() {
+	AuthenticateWithTokenCredential()
```
richardpark-msft, line 52: We don't use the return result from either of these calls?

sdk/samples/aztables/sample_update_entities.go
```diff
@@ -0,0 +1,149 @@
...
+	if !ok {
+		panic("TABLES_PRIMARY_STORAGE_ACCOUNT_KEY could not be found")
+	}
+	serviceURL := fmt.Sprintf("https://%s.table.core.windows.net/%s", accountName, "myTable")
+	cred, err := aztables.NewSharedKeyCredential(accountName, accountKey)
+	check(err)
+	client, err := aztables.NewClient(serviceURL, cred, nil)
+	check(err)
+
+	// 1. First add an entity
```
richardpark-msft, line 123: Terminology wise looks like you're using 'insert' as the verb.

sdk/samples/aztables/sample_authentication.go
seankane-msft: I could probably fix these to not return anything at all, but the goal is to demonstrate how to do a specific action. I'm not as concerned with the return item, more so with the individual functions.

sdk/samples/aztables/sample_update_entities.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+		Price:       1.99,
+		Inventory:   35,
+		ProductName: "Pens",
+		OnSale:      true,
+	}
+	marshalled, err := json.Marshal(myEntity)
+	check(err)
+
+	_, err = client.InsertEntity(context.Background(), marshalled, &aztables.InsertEntityOptions{UpdateMode: aztables.MergeEntity})
```
richardpark-msft, line 137: Silly thing but it's interesting to have 'Insert' for the function name, then in the options the 'mode' is actually 'UpdateMode'.

Is this method actually an Upsert?

seankane-msft: It's actually an Upsert, but the REST API documents it as an Insert, so we kept the terminology inline with the rest docs

sdk/samples/aztables/sample_insert_delete_entities.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	serviceURL := fmt.Sprintf("https://%s.table.core.windows.net/%s", accountName, "myTable")
+
+	cred, err := aztables.NewSharedKeyCredential(accountName, accountKey)
+	check(err)
+	client, err := aztables.NewClient(serviceURL, cred, nil)
+	check(err)
+
+	anyETag := azcore.ETagAny
+	_, err = client.DeleteEntity(context.Background(), "pk001", "rk001", &aztables.DeleteEntityOptions{IfMatch: &anyETag})
```
richardpark-msft, line 98: Is it obvious where this API returns the etag if I wanted to do it with a specific one?

(might be useful to show how, since that's probably a common use case)

sdk/samples/aztables/sample_query_table.go
```diff
@@ -0,0 +1,50 @@
...
+		panic("TABLES_PRIMARY_STORAGE_ACCOUNT_KEY could not be found")
+	}
+	serviceURL := fmt.Sprintf("https://%s.table.core.windows.net/%s", accountName, "myTable")
+
+	cred, err := aztables.NewSharedKeyCredential(accountName, accountKey)
+	check(err)
+	client, err := aztables.NewClient(serviceURL, cred, nil)
+	check(err)
+
+	filter := fmt.Sprintf("PartitionKey eq '%v' or PartitionKey eq '%v'", "pk001", "pk002")
```
richardpark-msft, line 30: Is the expectation that they've inserted data using some other code that's not shown here?

sdk/samples/aztables/sample_insert_delete_entities.go
seankane-msft: If an API returns an ETag it is contained in the response object.

sdk/samples/aztables/sample_batch.go
```diff
@@ -0,0 +1,111 @@
...
+		OnSale:      false,
+	}
+	marshalled, err = json.Marshal(entity4)
+	check(err)
+	batch = append(batch, aztables.TransactionAction{
+		ActionType: aztables.Delete,
+		Entity:     marshalled,
+	})
+
+	_, err = client.SubmitTransaction(context.Background(), batch, nil)
```
richardpark-msft, line 109: You'll probably get some people copying code like this verbatim.

Is there anything interesting in the return from SubmitTransaction they might want to see?

seankane-msft: If they wanted to inspect each individual sub-response (batching is a series of sub requests contained in a single request) they could do so, otherwise it's unimportant. I do see how showing the response would be helpful

sdk/samples/aztables/sample_create_delete_table.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/aztables"
+)
+
+func CreateFromServiceClient() {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	check(err)
+	accountName, ok := os.LookupEnv("TABLES_STORAGE_ACCOUNT_NAME")
```
richardpark-msft, line 18: The boilerplate is starting to obscure what you're doing.

Can you consolidate and just have two functions (one for table client, one for service client) and just do the Create/Delete in each of those, rather than separating it out into 4?

sdk/samples/aztables/sample_query_table.go
seankane-msft: My goal was to demonstrate how to do a list operation and include a filter, it would work with the previous inserts and actually list a few entities

sdk/samples/aztables/sample_update_entities.go
```diff
@@ -0,0 +1,149 @@
...
+	check(err)
+
+	// 2. Update the entity object
+	myEntity.OnSale = false
+
+	// 3. Insert the entity on the service
```
richardpark-msft, line 143: It might be good to explain a bit of why you chose the UpdateMode that you used.

sdk/samples/aztables/README.md
```diff
@@ -0,0 +1,145 @@
...
+|`LessThan`|`lt`|
+|`LessThanOrEqual`|`le`|
+|`NotEqual`|`ne`|
+|`And`|`and`|
+|`Not`|`not`|
+|`Or`|`or`|
+
+### Example Filters
+
+#### Filter on `PartitionKey` and `RowKey`:
```
richardpark-msft, line 67: Do you use this information in one of the samples? (if not, we probably should prefer that over just having readme text).

seankane-msft: Not really, but querying specific EDM types is not intuitive so I like having this info here in the samples readme

sdk/samples/aztables/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/azure-sdk-for-go/sdk/samples/aztables
```
jhendrixMSFT, line 1: We don't plant to tag/release this module right?

seankane-msft: No

seankane-msft: But I want it to be a runnable module if customers were to copy paste the code

richardpark-msft (approved):
Approving samples, under mild duress.

PR #15460, opened 2021-09-02
Move tools/internal to eng/tools/internal

benbp:
This PR moves the tools/internal project to eng/tools/internal. This is the first of two steps to move all of tools to eng/tools. The second step will move the remaining projects over and update their references and module names.

FYI @ArcturusZhang 

Part of https://github.com/Azure/azure-sdk-for-go/issues/15409


weshaggard (approved):

ArcturusZhang (changes requested):
We need to fix the internal references in the `internal` package from `github.com/Azure/azure-sdk-for-go/tools/internal` to `github.com/Azure/azure-sdk-for-go/eng/tools/internal`.
You can search the string literal `github.com/Azure/azure-sdk-for-go/tools/internal` in the directory of `eng/tools/internal` and replace it with the correct package name

Also we might need to update the `azure-pipeline.yml` for track 1 to make sure it will ignore `eng` directory to avoid any unexpected errors.

benbp:
@ArcturusZhang good catch. I updated the package name to `eng/tools/internal`, which had the side effect of breaking compilation since the `internal` package was no longer in the same root directory as the `tools/*` packages that referenced it.

With the latest update, I moved all files over from `tools/` to `eng/tools/`, but kept the `replace` directive in go.mod so the PR will build (meaning the follow-on PR will just remove `replace`).

I also updated all pipelines ci.yml files and related scripts, though it seems these pipelines have not been created yet. Were you planning on doing that or would you like me to?

eng/tools/internal/ci.yml
```diff
@@ -2,14 +2,14 @@
 trigger:
   paths:
     include:
-      - tools/deprecate
+      - eng/tools/internal
```
benbp, line 5: Something funky is going on with the git mv/diff here as it's correlating the wrong ci.yml files. All the files have been moved correctly with the proper paths.

benbp:
Fixed a broken link, but the link verification step will still fail until merge into `main`.

benbp:
@ArcturusZhang do the commented commands in `generate.go` ([for example](https://github.com/Azure/azure-sdk-for-go/blob/main/profiles/2020-09-01/generate.go#L6)) need to be updated as well? Are these commands parsed and run?

ArcturusZhang:
> @ArcturusZhang do the commented commands in `generate.go` ([for example](https://github.com/Azure/azure-sdk-for-go/blob/main/profiles/2020-09-01/generate.go#L6)) need to be updated as well? Are these commands parsed and run?

I believe only the relative path of the tool `../../tools/profileBuilder/main.go` might need tweaks. The other parts are the parameters to the tool using the relative path of the current directory, therefore those do not need to change

benbp:
@ArcturusZhang ok, updated. Let me know if there are any steps you think would be good to do to test these changes. Also, is there a convenient time that I can merge this so that it has minimal impact to your workflow if issues come up?

azure-pipelines.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       sdkPath: '$(GOPATH)/src/github.com/$(build.repository.name)'
       GO111MODULE: 'off'
       IGNORE_BREAKING_CHANGES: true
-      go.list.filter: '| grep -v vendor | grep -v azure-sdk-for-go/sdk | grep -v azure-sdk-for-go/tools'
+      go.list.filter: '| grep -v vendor | grep -v azure-sdk-for-go/sdk | grep -v azure-sdk-for-go/eng/tools'
```
ArcturusZhang, line 34: how about we just exclude all the `eng`? Since we have some other scripts written go as well.
```suggestion
      go.list.filter: '| grep -v vendor | grep -v azure-sdk-for-go/sdk | grep -v azure-sdk-for-go/eng'
```

ArcturusZhang (commented):
I noticed that we are moving everything in tools all at once? I am OK with that, but unfortunately we have conflicts right now. Could you please resolve them?

benbp:
> I noticed that we are moving everything in tools all at once? I am OK with that, but unfortunately we have conflicts right now. Could you please resolve them?

Yep I've been having to rebase every once in a while. Should be fixed again now. It will be nice to merge this :)

See [my previous comment](https://github.com/Azure/azure-sdk-for-go/pull/15460#issuecomment-915443530) regarding moving everything all at once. We can't move `internal` out of the same directory parent without breaking compilation.

ArcturusZhang (approved):

PR #15456, opened 2021-09-02
Merge live and ci test stage templates

benbp:
This PR merges the live and ci stage templates into one unified pipeline file. It also updates aztables to run the live tests and work with the pipeline generators.

PR #15478, opened 2021-09-03
Live samples

seankane-msft:
Adds live samples tests to internal pipelines

This is dependent on #15456 

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -68,11 +68,12 @@ steps:
         PROXY_CERT: $(Build.SourcesDirectory)/eng/common/testproxy/dotnet-devcert.crt
         ${{ insert }}: ${{ parameters.EnvVars }}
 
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -68,6 +68,13 @@ steps:
         PROXY_CERT: $(Build.SourcesDirectory)/eng/common/testproxy/dotnet-devcert.crt
         ${{ insert }}: ${{ parameters.EnvVars }}
 
+    - ${{ if eq(variables['System.TeamProject'], 'internal')}}:
```
benbp, line 71: I prefer to avoid template expression conditionals like this when possible. I think you can achieve the same effect by passing in a `TestSamples` parameter through from `live.tests.yml` [here](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/jobs/live.tests.yml#L87).

seankane-msft:
Samples are going to live in the same directory as the client libraries to take advantage of godoc capabilities. Might revisit if there is a need to test them live.

Issue #15171, opened 2021-07-27
Manual credit management for batch receivers


richardpark-msft:
The credit allocation in the current AMQP library appears to use a 'rolling credit' allocation. Basically, you set a maximum credit size and it's maintained throughout the link.

This can work well for a streaming receiver, as it acts as a prefetch mechanism. But for batching receivers this makes it difficult to properly specify a fixed set of credits and ensure that no more than an initial set will be requested. You'd basically always be fighting against the "replenish" part of the credit allocation logic.

Other AMQP libraries (like rhea) and qpid, allow you to add credits manually. This is a simpler model and gives the caller (in this case, us with either Service Bus or Event Hubs) more control.

richardpark-msft:
This is the SB side of the story and is being handled in https://github.com/Azure/azure-sdk-for-go/pull/15480

richardpark-msft:
Handled as part of our go-amqp PR: https://github.com/Azure/go-amqp/pull/56

PR #15480, opened 2021-09-04
[service-bus] Adding in batch receiving using the Receiver and manual credit APIs

### LINKED ISSUE #15156: [service-bus] Batch receiving
Batch receiving is receiving messages with a finite limit. This differs from streaming in that you should be able to request 5 messages and get no more than 5 messages.

In JS they also drain the link after the receive so no excess messages or internal cache is required, which removes a whole class of potential problems.

Both of these are not implemented in the go-amqp library today. You are always forced to create a link with rolling credits turned on and cannot adjust it after the link has been created.

The typical use case for this kind of API is a set of workers that come up, process a message, and then die (for instance, workers in a build system).

The usage would be similar to this:

```go
// NOTE: context.WithTimeout() is probably more 'go' appropriate for the timeout.
messages, err := receiver.ReceiveMessages(ctx, 5, timeout)

if err != nil {
  // elided
}

for _, message := range messages {
  receiver.CompleteMessage(ctx, message)
}
```

As referenced in: https://github.com/Azure/azure-sdk-for-go/issues/15092
(AMQP work should be included as part of #14907)

richardpark-msft:
Adding in support for the 'batch receiving' APIs in Service Bus. This is pretty much the API surface we've used for all the other Service Bus track 2 SDKs.

This PR also adds in more live and unit tests to bring up coverage a bit. The changes were getting a bit large so I figured  we should review and commit.

Closes #15156, #15171


richardpark-msft:
(this work is in process, relies on this PR: https://github.com/Azure/go-amqp/pull/56, which is in review)

richardpark-msft:
(moved this out of draft a little hastily. Just a little cleanup - like, getting go-amqp released!)

sdk/servicebus/azservicebus/go.mod
```diff
@@ -10,11 +10,14 @@ require (
...
 	github.com/joho/godotenv v1.3.0
-	github.com/microsoft/ApplicationInsights-Go v0.4.4 // indirect
+	github.com/microsoft/ApplicationInsights-Go v0.4.4
 	github.com/mitchellh/mapstructure v1.3.3
 	github.com/stretchr/testify v1.6.1
 	golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd // indirect
 	nhooyr.io/websocket v1.8.6
 )
+
+replace github.com/Azure/go-amqp v0.13.11 => /home/codespace/src/_/amqp-credit/amqp
```
richardpark-msft, line 23: Just need to get go-amqp released with the latest manual credit fixes!

sdk/servicebus/azservicebus/client.go
```diff
@@ -24,23 +24,23 @@ type ServiceBusClient struct {
 
 // ServiceBusClientOption is the type for an option that can configure ServiceBusClient.
 // For an example option, see `ServiceBusWithConnectionString`
-type ServiceBusClientOption func(client *ServiceBusClient) error
+type ServiceBusClientOption func(client *Client) error
```
seankane-msft, line 27: Shouldn't these then remove ServiceBus prefix as well, ie `ClientOptions`?

sdk/servicebus/azservicebus/client.go
```diff
@@ -24,23 +24,23 @@ type ServiceBusClient struct {
...
 func ServiceBusWithConnectionString(connStr string) ServiceBusClientOption {
-	return func(client *ServiceBusClient) error {
+	return func(client *Client) error {
 		fn := internal.NamespaceWithConnectionString(connStr)
 		return fn(client.namespace)
 	}
 }
 
-// NewServiceBusClient creates a new ServiceBusClient.
+// NewClient creates a new ServiceBusClient.
```
seankane-msft, line 37:
```suggestion
// NewClient creates a new Client
```

sdk/servicebus/azservicebus/main_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/joho/godotenv"
+)
+
+func TestMain(m *testing.M) {
+	godotenv.Load()
```
seankane-msft, line 11: I saw in the other file you had to include the file, is that optional or required?

sdk/servicebus/azservicebus/client.go
```diff
@@ -24,23 +24,23 @@ type ServiceBusClient struct {
 
 // ServiceBusClientOption is the type for an option that can configure ServiceBusClient.
 // For an example option, see `ServiceBusWithConnectionString`
-type ServiceBusClientOption func(client *ServiceBusClient) error
+type ServiceBusClientOption func(client *Client) error
 
 // ServiceBusWithConnectionString configures a namespace with the information provided in a Service Bus connection string
 func ServiceBusWithConnectionString(connStr string) ServiceBusClientOption {
```
jhendrixMSFT, line 30: WithConnectionString

richardpark-msft: Fixed

sdk/servicebus/azservicebus/client.go
richardpark-msft: Perfect, updated!

sdk/servicebus/azservicebus/client.go
richardpark-msft: Indeed and renamed.

sdk/servicebus/azservicebus/go.mod
richardpark-msft: Updated to point to go-amqp v0.14.0

sdk/servicebus/azservicebus/main_test.go
richardpark-msft: Basically, if omitted it just assumes that the .env file is in the current folder.

With that said, I should test that it worked properly. I know the root folder one (this one) does.

sdk/servicebus/azservicebus/errors.go
```diff
@@ -4,3 +4,4 @@ import "errors"
 
 var ErrSenderClosed = errors.New("sender is closed and cannot be used")
 var ErrReceiverClosed = errors.New("receiver has been closed and can no longer be used")
+var ErrProcessorClosed = errors.New("processor has been closed and can no longer be used")
```
jhendrixMSFT, line 7: Worth a read https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

No need to make changes now but I think this is something we'll want to revisit.

sdk/servicebus/azservicebus/receiver.go
```diff
@@ -14,11 +24,306 @@ const (
...
+
+	// make sure we leave the link in a consistent state.
+	if err := r.receiver.DrainCredit(ctx); err != nil {
+		return nil, err
+	}
+
+	return messages, err
+}
+
+func (r *Receiver) CompleteMessage(ctx context.Context, message *ReceivedMessage) error {
```
jhendrixMSFT, line 228: Is this the track 2 pattern, i.e. place the message handling APIs on the receiver and not the message like in track 1?

sdk/servicebus/azservicebus/linkCommon.go
```diff
@@ -0,0 +1,36 @@
+package azservicebus
+
+import "context"
+
+type linkState struct {
```
jhendrixMSFT, line 5: Comment describing what this is used for?

sdk/servicebus/azservicebus/receiver.go
richardpark-msft: Yes. 

The big argument at the time is that we try not to use any 'Active Record' pattern with our models/data contracts.

sdk/servicebus/azservicebus/errors.go
richardpark-msft: Hm....that's an interesting one.

Thinking about this a bit more, the real reason these errors exist is because you've either abruptly closed the connection/link and were in the middle of an operation _or_ you closed something and tried to use it afterwards.

In both cases the link/processor isn't coming back. So maybe I could roll all these up and just say it's ErrClosed or similar. Because I think the default thought for most errors coming back from the clients is "I should retry this".

richardpark-msft: Hm....that's an interesting one.

Thinking about this a bit more, the real reason these errors exist is because you've either abruptly closed the connection/link and were in the middle of an operation _or_ you closed something and tried to use it afterwards.

In both cases the link/processor isn't coming back. So maybe I could roll all these up and just say it's ErrClosed or even just having an error type that's basically "FatalError" or similar? 

Because I think the default thought for most errors coming back from the clients is "I should retry this" so I just want to signal that somehow.

sdk/servicebus/azservicebus/linkCommon.go
richardpark-msft: Comment added.

sdk/servicebus/azservicebus/errors.go
jhendrixMSFT: As an example, for non-transient errors we've been using an interface-based pattern like [`NonRetriable()`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/errorinfo/errorinfo.go#L15).  This way, it doesn't matter what type of error is returned, if it implements the `NonRetriable` interface then we know it's non-transient.

I suspect that we'll want to move away from sentinel error types and use an interface-based approach.  This way the underlying error type doesn't matter and they can be extended over time.

richardpark-msft: Let me swap over now - there's no harm in it and it's better :)

richardpark-msft: Okay, latest commit does what I believe you were saying. Let me know if I got it.

sdk/servicebus/azservicebus/errors.go
```diff
@@ -1,7 +1,13 @@
 package azservicebus
 
-import "errors"
+import "fmt"
 
-var ErrSenderClosed = errors.New("sender is closed and cannot be used")
-var ErrReceiverClosed = errors.New("receiver has been closed and can no longer be used")
-var ErrProcessorClosed = errors.New("processor has been closed and can no longer be used")
+// implements `internal/errorinfo/NonRetriable`
+type ErrClosed struct {
```
jhendrixMSFT, line 6: I don't think this needs to be exported.

richardpark-msft: You are right. Unexported :)

sdk/servicebus/azservicebus/main_test.go
richardpark-msft: Okay, just tested it out a few ways and it's working the way I want it (in that you only have to have a single .env file, at the 'root' of the azservicebus package to run the tests.

There was some other code I got rid of that was attempting to load the .env file as well, so those are all gone and now it's pure TestMain().

sdk/servicebus/azservicebus/receiver.go
```diff
@@ -14,11 +24,306 @@ const (
 	SubQueueTransfer   = "transferDeadLetter"
 )
 
-type ServiceBusReceiver struct{}
+type Receiver struct {
```
jhendrixMSFT, line 27: Doc comment.

sdk/servicebus/azservicebus/receiver.go
```diff
@@ -14,11 +24,306 @@ const (
...
+	}
+
+	mu       sync.Mutex
+	ns       legacyNamespace
+	receiver internal.LegacyReceiver
+
+	linkState *linkState
+}
+
+type ReceiverOption func(receiver *Receiver) error
```
jhendrixMSFT, line 46: Doc comment

sdk/servicebus/azservicebus/receiver.go
```diff
@@ -14,11 +24,306 @@ const (
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	receiver.config.FullEntityPath = entityPath
+	return receiver, nil
+}
+
+type ReceiveOptions struct {
```
jhendrixMSFT, line 125: It's unfortunate that this has to be exported.  An option to get rid of this, although it's a bit more code, is to have `ReceiveOption` be an interface with an internal `apply()` method.  That way you don't have to export any of the dependent types.

```go
type ReceiveOption interface {
	apply(*receiveOption)
}

type receiveOption struct {
	maxWaitTime time.Duration
}

type receiveOptionFn func(*receiveOption)

func (r receiveOptionFn) apply(ro *receiveOption) {
	r(ro)
}

func ReceiveWithMaxWaitTime(max time.Duration) ReceiveOption {
	return receiveOptionFn(func(ro *receiveOption) {
		ro.maxWaitTime = max
	})
}
```

jhendrixMSFT: Also, and we don't have to change this now, but if we make `apply()` (or whatever impl we go with) return an `error` then it gives us future flexibility if we find ourselves in the situation where we want to validate a config option.

sdk/servicebus/azservicebus/receiver.go
jhendrixMSFT: Applies to other places with missing doc comments.

jhendrixMSFT (approved):
Some missing comments and a possible change on the variadic config pattern.

sdk/servicebus/azservicebus/errors.go
richardpark-msft: (Looks like we're good on this. Resolving)

sdk/servicebus/azservicebus/receiver.go
richardpark-msft: Seems sensible to me. I'm assuming we're going to have this same problem in autogenerated code. What pattern are we laying out there today (the apply version or the version I have above)?

sdk/servicebus/azservicebus/receiver.go
richardpark-msft: Many a doc comment has been added in my most recent push.

Also, since golint was so helpful I also fixed some other small things, like unnecessary else'ing, etc..

sdk/servicebus/azservicebus/receiver.go
richardpark-msft: Added, added, added!

PR #15519, opened 2021-09-09
failing if go build/vet do not succeed

seankane-msft:


tadelesh (commented):
Maybe `go mod tidy` and `go vet` also need to handle execution error just to be safe.

seankane-msft:
> Maybe `go mod tidy` and `go vet` also need to handle execution error just to be safe.

I'll add to `go mod tidy` but `go vet` is essentially the same as our linter and the linter allows for `//nolint` so I think it's better to default to the linter.

eng/scripts/build.ps1
```diff
@@ -10,6 +10,7 @@ function Process-Sdk () {
     if ($clean) {
         Write-Host "##[command]Executing go clean -v ./... in " $currentDirectory
         go clean -v ./...
+        if ($LASTEXITCODE) {exit $LASTEXITCODE}
```
weshaggard, line 13: nit:
```suggestion
        if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

weshaggard (approved):

tadelesh (approved):

PR #15523, opened 2021-09-09
Filter CI builds by service directory

benbp:
CI builds are currently not filtering by service, and are building all packages.


seankane-msft (approved):
Is the `-filter` arg already available on the build.ps1 script? If so, looks good to me.

RickWinter (approved):

benbp:
@seankane-msft yeah it's in there, it just was never added to the ci arguments.

PR #15531, opened 2021-09-10
Verify existence of doc.go file

seankane-msft:


sdk/internal/atomic/doc.go
```diff
@@ -0,0 +1,7 @@
+//go:build go1.16
+// +build go1.16
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package atomic
```
RickWinter, line 7:
```suggestion
package atomic

```

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -24,6 +24,16 @@ steps:
     failOnStderr: false
     workingDirectory: '${{parameters.GoWorkspace}}'
 
+  - pwsh: |
+      $modDirs = (./eng/scripts/get_module_dirs.ps1 -serviceDir $(SCOPE))
+      foreach ($md in $modDirs) {
+        pushd $md
+        Get-ChildItem doc.go
+        if ($LASTEXITCODE) { exit $LASTEXITCODE }
+      }
```
benbp, line 33: `$LASTEXITCODE` only applies to exit codes from external process calls. Powershell commands/functions are handled via `throw`/exceptions. You could also use a `Test-Path` conditional here.

```suggestion
      foreach ($md in $modDirs) {
        Get-ChildItem $md/doc.go
      }

File: eng/pipelines/templates/steps/analyze.yml [hunk 3]
```diff
@@ -24,6 +24,16 @@ steps:
     failOnStderr: false
     workingDirectory: '${{parameters.GoWorkspace}}'
 
+  - pwsh: |
+      $modDirs = (./eng/scripts/get_module_dirs.ps1 -serviceDir $(SCOPE))
```
benbp, line 28:
```suggestion
      $modDirs = ./eng/scripts/get_module_dirs.ps1 -serviceDir $(SCOPE)
```

benbp (approved):

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -24,6 +24,14 @@ steps:
     failOnStderr: false
     workingDirectory: '${{parameters.GoWorkspace}}'
 
+  - pwsh: |
+      $modDirs = ./eng/scripts/get_module_dirs.ps1 -serviceDir $(SCOPE)
+      foreach ($md in $modDirs) {
+        Get-ChildItem $md/doc.go
```
weshaggard, line 30: Is the idea that you want this to fail if it isn't present? If so perhaps we should do a Test-Path and give a nicer error message if it is missing.

PR #15537, opened 2021-09-10
Use a typedef for ARM endpoints

jhendrixMSFT:
This provides a clue for what values callers are supposed to pass.
Since the underlying type is a string, callers can still pass custom
endpoint strings using a cast.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Release History
 
-## 0.19.1 (Unreleased)
+## v0.20.0 (Unreleased)
+
+### Breaking Changes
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
I remembered that this came up in the API review of `azcore`.  Looks like we never opened an issue.

seankane-msft (approved):
Can you add a changelog entry before merging?

RickWinter (approved):

jhendrixMSFT:
@seankane-msft changelog has been updated.

jhendrixMSFT:
@chlowell do we still want to take this now that we'll be removing `arm.Connection`?  I _think_ the answer is yes as we'd use this typedef in the connection options?

chlowell:
Yes, we would only move the option, not remove it.

PR #15550, opened 2021-09-13
Cosmos DB - Re organize files into data

ealsur:
Move `sdk/cosmos` to `sdk/data/azcosmos`


weshaggard:
Moving the ci.yml file is going to break the pipeline so that will need to be updated once this change is ready/in.

ausfeldt (approved):

benbp:
@ausfeldt ping me when you merge I can update the internal+public pipelines.

benbp:
I think to be consistent this should be `data/azcosmos` not `data/cosmos/azcosmos`?

ealsur:
@benbp Based on https://azure.github.io/azure-sdk/golang_introduction.html#golang-packages, there could be an `armcosmos` package also for management operations, so the folder makes sense to group both up? @jhendrixMSFT Could you confirm?

sdk/data/cosmos/azcosmos/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/Azure/azure-sdk-for-go/sdk/cosmos/azcosmos
+module github.com/Azure/azure-sdk-for-go/sdk/data/cosmos/azcosmos
 
 go 1.13
```
seankane-msft, line 3: This should be go 1.16, which is our new min supported version

sdk/data/cosmos/azcosmos_emulator_tests/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/Azure/azure-sdk-for-go/sdk/data/cosmos/azcosmos_emulator_tests
+
+go 1.13
```
seankane-msft, line 3: Even though this is not going to be publicly released, this should also be go 1.16

ealsur: Done

sdk/data/cosmos/azcosmos/go.mod
ealsur: Done

benbp:
> @benbp We currently have 2 packages, one for black box testing `azcosmos_emulator_tests` I wanted to avoid dumping both inside `data`. @jhendrixMSFT Could you confirm?

I'm not sure what the best solution is here (i.e. if nesting go modules is to be avoided), but I think the convention is to have package names like `data/<track2 service name`, like there already is for `data/aztables`.

ealsur:
> I'm not sure what the best solution is here (i.e. if nesting go modules is to be avoided), but I think the convention is to have package names like `data/<track2 service name`, like there already is for `data/aztables`.

The problem with nesting is that `go test` would execute everything, both the UTs in the root package (azcosmos) and then the nested one (azcosmos_emulator_tests), and it would be fine except for the fact that the Cosmos DB emulator runs only on windows. So it would break in any Linux run. That's why in our YML we run them separately and Emulator tests run only on Windows.

Not sure how to go around this one.

weshaggard:
@ealsur is there no way to condition based on platform or some flag you can pass to go test? If so we should try doing that.

ealsur:
> @ealsur is there no way to condition based on platform or some flag you can pass to go test? If so we should try doing that.

We already discussed that and tried it, `go test` allows for flags to be passed yes but they are compilation flags. The go compilation flags make it so the development process is very hard. The files with the flags do not get compiled in any IDE and it disables any IDE Intellisense or autocompletion.

ealsur:
@weshaggard @benbp - Moved all the files into `sdk/data/azcosmos`, I'll see how to deal with Windows Emulator tests from the dev side

weshaggard (approved):

PR #15561, opened 2021-09-15
[service-bus] Allow sending of message batches

### LINKED ISSUE #15136: [service-bus] Message batching
Track 2 added in the ability to create a message batch - basically, it's a "append only" collection of messages which are encoded and sent as a single message. This can be a big efficiency boost for senders as it allows the combining of several small messages into one, eliminating some per-message send overhead which can add up.

Interfaces used for illustration - current policy for the Go SDK (from what I can see) is to use concrete types:

```go
type ServiceBusMessageBatch interface {
  AddMessage(msg *Message) bool
}

type interface ServiceBusSender {
  // other methods elided...

  // looks at the current link for the 'max batch size', creates a message batch
  // NOTE this does actually need to potentially initialize the link so it needs to take a context.
  CreateMessageBatch(ctx context.Context) (*ServiceBusMessageBatch, error)
}

// usage

batch, err := sender.CreateMessageBatch(context.TODO())

// NOTE: in other languages where we throw exceptions this check is less
// complicated. Perhaps we'd like to have a designated error (message too big) 
// so we just have the single return value instead of having `added` and `err`?
if added, err := batch.TryAdd(message); !added {
  // message failed to add - decisions to make:
  if err != nil {
      // If there was an internal encoding failure the best they can do is drop this message.
      return
  }

  // otherwise the only reason we failed here was because the message was too big to fit
  // into the existing batch. So we can:
  // a. Send the batch immediately and then start a new batch with this message
  // b. Discard this message

  // There is also the potential pitfall that the message is too large, even on it's own,
  // to fit into a batch. All of this is determinable using the .TryAdd() call and checking
  // it's return value.
}

sender.SendMessageBatch(context.TODO(), batch)

```

References:
- [batch.go](https://github.com/richardpark-msft/azure-sdk-for-go/blob/e68e55fb55079816240f0dab9991215d7d33885a/sdk/servicebus/batch.go#L106)

richardpark-msft:
Adding in sending of message batches with track 2.

Close #15136


sdk/servicebus/azservicebus/exampleSendingMessageBatch_test.go
```diff
@@ -0,0 +1,70 @@
+package azservicebus_test
```
seankane-msft, line 1: Nit: this file should probably be called example_sending_message_batch_test.go

sdk/servicebus/azservicebus/internal/sender.go
```diff
@@ -49,21 +49,27 @@ type (
...
 )
 
+// ie: `*internal.Sender`
+type LegacySender interface {
+	SendAMQPMessage(ctx context.Context, msg *amqp.Message) error
+	Close(ctx context.Context) error
+	MaxMessageSize() uint64
+}
+
+// NewLegacySender is just a wrapper for `NewSender` that returns an interface. Makes testing simpler.
```
seankane-msft, line 66: Is this also something we expect customers to use? Or is this strictly for testing?

richardpark-msft: Definitely just for me. It's in the `internal` package, so not usable outside of our module.

sdk/servicebus/azservicebus/exampleSendingMessageBatch_test.go
richardpark-msft: Ah, brand new to this, will rename!

jhendrixMSFT: This does have significance though.  The difference between `azservicebus` and `azservicebus_test` is that the latter will only let you use exported content (i.e. you can't directly access any internal types/funcs).  IMO this is mostly useful when writing examples as it ensures you don't inadvertently use non-public surface area, less so for unit tests.

sdk/servicebus/azservicebus/client.go
```diff
@@ -73,8 +73,8 @@ func (client *Client) NewProcessor(options ...ProcessorOption) (*Processor, erro
 }
 
 // NewSender creates a Sender, which allows you to send messages or schedule messages.
-func (client *Client) NewSender(queueOrTopic string, options ...SenderOption) (*Sender, error) {
-	sender, err := newSender(client.namespace, queueOrTopic, options...)
+func (client *Client) NewSender(queueOrTopic string) (*Sender, error) {
```
jhendrixMSFT, line 76: Why were the options removed?

sdk/servicebus/azservicebus/message.go
```diff
@@ -1,19 +1,108 @@
...
+	// TODO: is an empty session ID valid?
+	SessionID string
 }
 
 // ReceivedMessage represents a message received from Service Bus.
 type ReceivedMessage struct {
 	Message
-	LockToken      string
-	SequenceNumber int64
+	LockToken      *string
```
jhendrixMSFT, line 29: Is there a semantic difference between an empty string an `nil` here?  Similar question for `SequenceNumber` and `PartitionID` below.

sdk/servicebus/azservicebus/sender.go
```diff
@@ -16,31 +18,86 @@ type Sender struct {
...
+	return func(options *messageBatchOptions) error {
+		options.maxSizeInBytes = &maxSizeInBytes
+		return nil
+	}
+}
+
+// NewMessageBatch can be used to create a batch that contain multiple
+// messages. Sending a batch of messages is more efficient than sending the
+// messages one at a time.
+func (s *Sender) NewMessageBatch(ctx context.Context, options ...MessageBatchOption) (*MessageBatch, error) {
```
jhendrixMSFT, line 56: Conceptually, it seems weird that creating a message batch (which should just be a container of messages) needs a context.  Can we defer creating the sender until we are actually attempting to send?

richardpark-msft: It is a bit odd, but all the SDKs do this the same way.

The gist is that we need to create a link before we know the actual max message size (it comes with the initial attaching of the link itself). Without this the batching feature isn't useful since it can't properly check that the messages you're adding to the batch are properly sized at all.

So this is one case where we can't defer it.

sdk/servicebus/azservicebus/client.go
richardpark-msft: The only option that existed was one to set the sessionID for a Sender.

In track 2 there isn't a session association with a sender (it's expected to be set explicitly in each message that's sent).

Interestingly enough (if you follow the rabbit hole a bit more) there are some spots that accepted sender options but then didn't propagate them to the underlying call. So it was spotty at best (maybe on purpose...)

sdk/servicebus/azservicebus/exampleSendingMessageBatch_test.go
richardpark-msft: @jhendrixMSFT - agreed, package name was intentional.

I think @seankane-msft is only referring to the filename, however, unless I'm misunderstanding something.

sdk/servicebus/azservicebus/message.go
richardpark-msft: Let me check.

LockToken being an empty string is probably okay (I don't think it's a valid value in the range anyways).

Let me check on the other two. I'm thinking the zero values for them would be okay, in context, but want to see what we did with the other languages.

sdk/servicebus/azservicebus/exampleSendingMessageBatch_test.go
seankane-msft: I was referring to the filename. I can see why you would want to test from a separate package but its helpful to test internal methods too so I've generally avoided doing so.

jhendrixMSFT: Derp my bad, yes agreed the file name should be snake-cased, not camel-cased.

richardpark-msft: File renamed, and I'm glad you two were able to work out your differences.

sdk/servicebus/azservicebus/message.go
richardpark-msft: Allright:
- LockToken is now just a string. We should be okay in any context where it matters.
- PartitionID was not actually supposed to be exported. So I've removed it.
- SequenceNumber is now just a plain int64, not a pointer.

sdk/servicebus/azservicebus/message.go
```diff
@@ -1,12 +1,26 @@
...
+)
 
 // Message represents a sendable message, including only user-settable properties.
 // A message that has been received from Service Bus will be of type `ReceivedMessage`.
 type Message struct {
-	Body []byte
-	ID   string
+	Body                    []byte
+	ID                      string
+	PartitionKey            *string
```
jhendrixMSFT, line 19: Does this and `TransactionPartitionKey` need to be pointer-to-type?

sdk/servicebus/azservicebus/messageBatch.go
```diff
@@ -0,0 +1,119 @@
...
+		amqpMessage.Data = nil
+		mb.batchEnvelope = amqpMessage
+	}
+
+	mb.marshaledMessages = append(mb.marshaledMessages, bin)
+	return nil
+}
+
+// Size is the number of bytes in the message batch
+func (mb *MessageBatch) Size() int {
```
jhendrixMSFT, line 99: Does this and `Len()` methods need to be exported?  How do customers use them?

richardpark-msft: One immediate use case for Len(): if you're creating batches and, for instance, you fail to add a message to the batch _and_ Len() == 0 then that means your message is never going to fit (ie, it's too big to even be sent by itself).

Otherwise, they can be used by customers that might want to do their own optimization for when to send a batch (ie, by size in bytes, or by # of messages). Both of which I think are fairly common when it comes to this kind of thing.

(just to note: these are just standard track 2-isms. Nothing new here).

sdk/servicebus/azservicebus/message.go
richardpark-msft: nil is actually valid in partitionKey case.

I can't do transactions quite yet so I can't test the other but it follows they'd use the same rules.

jhendrixMSFT (approved):

PR #15574, opened 2021-09-15
Add type for credential authority hosts

chlowell:
Per our last azidentity API review, this adds a new type, `AzureAuthorityHost`, to give users a hint about expected values.


sdk/azidentity/azidentity.go
```diff
@@ -18,19 +18,23 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
+// The base URL for Azure Active Directory
+type AzureAuthorityHost string
```
jhendrixMSFT, line 22: Can we shorten this to `AuthorityHost`?  Having to specify `azidentity.AzureAuthorityHost` seems long and a bit redundant.

sdk/azidentity/azidentity.go
```diff
@@ -18,19 +18,23 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
+// The base URL for Azure Active Directory
```
jhendrixMSFT, line 21: The idiom for doc comments is the must start with the thing being doc'ed, e.g. `// AzureAuthorityHost is the base URL...`

sdk/azidentity/azidentity.go
```diff
@@ -148,22 +152,23 @@ type pipelineOptions struct {
...
-		authorityHost = AzurePublicCloud
-		// NOTE: we only allow overriding the authority host if no host was specified
+func setAuthorityHost(authorityHost AzureAuthorityHost) (string, error) {
+	host := string(authorityHost)
+	if host == "" {
 		if envAuthorityHost := os.Getenv("AZURE_AUTHORITY_HOST"); envAuthorityHost != "" {
-			authorityHost = envAuthorityHost
+			host = envAuthorityHost
+		} else {
+			host = string(AzurePublicCloud)
```
jhendrixMSFT, line 161: You could move this before the call to `os.Getenv()` to avoid the `else` clause.

sdk/azidentity/azidentity.go
chlowell: Sure, have done.

sdk/azidentity/azidentity.go
chlowell: That's how it was, I changed it because the unnecessary second assignment when the `if` evaluates true seemed strange to me. Is avoiding `else` like this typical Go style?

jhendrixMSFT (approved):

PR #15598, opened 2021-09-20
Example for `HTTPResponse` error

### LINKED ISSUE #15587: [Core] Add documentation of `azcore.HTTPResponse` interface
```golang
resp, err := Client.GetWidget(...)
var httpErr azcore.HTTPResponse
if errors.As(err, &httpErr) {
    // do something with httpErr.RawResponse()
}
```

seankane-msft:
Closes #15587

### DIFF
--- a/sdk/azcore/example_test.go
+++ b/sdk/azcore/example_test.go
@@ -8,11 +8,16 @@
 package azcore_test
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 )
 
 // false positive by linter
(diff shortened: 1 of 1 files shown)


jhendrixMSFT (approved):

sdk/azcore/example_test.go
```diff
@@ -58,3 +63,25 @@ func ExampleNullValue() {
...
+		if httpErr.RawResponse().StatusCode == http.StatusNotFound {
+			fmt.Printf("Repository could not be found: %v", httpErr.RawResponse())
+		} else if httpErr.RawResponse().StatusCode == http.StatusForbidden {
+			fmt.Printf("You do not have permission to access this repository: %v", httpErr.RawResponse())
+		} else {
+			// ...
+		}
+	}
+	// Do something with response
+	fmt.Println(ioutil.ReadAll(resp.Body))
```
richardpark-msft, line 86: Should we check that resp is not nil? Ie, are azcore.HTTPResponse's the only errors that can come back here?

jhendrixMSFT: So `HTTPResponse` is the interface used to get at the underlying `*http.Response` from an error returned by SDK code due to a failure HTTP status code.  In the context of this example, `Do()` will never return an `HTTPResponse` error because there's no HTTP status code to predicate success/failure.

PR #15603, opened 2021-09-21
[KeyVault] azsecrets

seankane-msft:
Basic API for key vault secrets. Putting this into a feature branch to get CI set things up.


sdk/keyvault/azsecrets/README.md
```diff
@@ -0,0 +1,307 @@
+<!-- autorest --use=@autorest/go@4.0.0-preview.27 https://github.com/Azure/azure-rest-api-specs/tree/1e2c9f3ec93078da8078389941531359e274f32a/specification/keyvault/data-plane --tag=package-7.2 --output-folder=internal --module-azsecrets --module-version=0.1.0 --openapi-type="data-plane" --security="AADToken" --security-scopes="https://vault.azure.net/.default" -->
```
seankane-msft, line 1: There are several Python references I need to fix in here

sdk/keyvault/azsecrets/ci.yml
```diff
@@ -0,0 +1,17 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
+trigger:
```
weshaggard, line 2: Lets please also add in the branch triggers. See recent updates to other ci.yml files where @benbp  added these.

sdk/keyvault/azsecrets/go.mod
```diff
@@ -0,0 +1,18 @@
+module github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets
+
+go 1.16
+
+replace github.com/Azure/azure-sdk-for-go/sdk/internal => ../../internal
```
jhendrixMSFT, line 5: Can this be removed?

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+// Client is the struct for interacting with a KeyVault Secrets instance
+type Client struct {
+	kvClient *internal.KeyVaultClient
+	vaultUrl string
+}
+
+// ClientOptions are the configurable options on a Client.
+type ClientOptions struct {
+	// HTTPClient sets the transport for making HTTP requests.
+	HTTPClient policy.Transporter
```
JeffreyRichter, line 29: Do we call this HTTPClient everywhere? Not Transport? @jhendrixMSFT

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+// SetSecretOptions contains the optional parameters for a Client.SetSecret operation
+type SetSecretOptions struct {
+	// Type of the secret value such as a password.
+	ContentType *string `json:"contentType,omitempty"`
+
+	// The secret management attributes.
+	SecretAttributes *SecretAttributes `json:"attributes,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
JeffreyRichter, line 136: Do we really want *string for the value type? Can't this be just string

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+	RawResponse *http.Response
+
+	// The secret management attributes.
+	Attributes *SecretAttributes `json:"attributes,omitempty"`
+
+	// The secret id.
+	ID *string `json:"id,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
JeffreyRichter, line 158: Can *string -> string?

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+// SecretUpdateParameters - The secret update parameters.
+type SecretProperties struct {
+	// Type of the secret value such as a password.
+	ContentType *string `json:"contentType,omitempty"`
+
+	// The secret management attributes.
+	SecretAttributes *SecretAttributes `json:"attributes,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
JeffreyRichter, line 443: Can *string -> string?

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+				},
+				RecoverableDays: i.Attributes.RecoverableDays,
+				RecoveryLevel:   deletionRecoveryLevelFromGenerated(*i.Attributes.RecoveryLevel).ToPtr(),
+			},
+		},
+	}
+}
+
+// RestoreSecretBackup restores a backed up secret, and all its versions, to a vault. This operation requires the secrets/restore permission.
+func (c *Client) RestoreSecretBackup(ctx context.Context, backup []byte, options *RestoreSecretBackupOptions) (RestoreSecretBackupResponse, error) {
```
JeffreyRichter, line 557: What is backup []byte?
This needs better documentation at least. I think the []byte comes from another operations response, right? Mention that operation in the docs here so the customer knows how to get this []byte.

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+// ListDeletedSecretsPage holds the data for a single page.
+type ListDeletedSecretsPage struct {
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+
+	// READ-ONLY; The URL to get the next set of deleted secrets.
+	NextLink *string `json:"nextLink,omitempty" azure:"ro"`
+
+	// READ-ONLY; A response message containing a list of the deleted secrets in the vault along with a link to the next page of deleted secrets
+	DeletedSecrets []*DeletedSecretItem `json:"value,omitempty" azure:"ro"`
```
JeffreyRichter, line 805: Can *DeletedSecretItem -> DeletedSecretItem?

sdk/keyvault/azsecrets/client.go
```diff
@@ -0,0 +1,1011 @@
...
+// The secret list result
+type ListSecretVersionsPage struct {
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+
+	// READ-ONLY; The URL to get the next set of secrets.
+	NextLink *string `json:"nextLink,omitempty" azure:"ro"`
+
+	// READ-ONLY; A response message containing a list of secrets in the key vault along with a link to the next page of secrets.
+	Secrets []*SecretItem `json:"value,omitempty" azure:"ro"`
```
JeffreyRichter, line 892: Kill the *?

sdk/keyvault/azsecrets/models.go
```diff
@@ -0,0 +1,189 @@
...
+	Attributes *SecretAttributes `json:"attributes,omitempty"`
+
+	// The content type of the secret.
+	ContentType *string `json:"contentType,omitempty"`
+
+	// The secret id.
+	ID *string `json:"id,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
JeffreyRichter, line 40: Kill the *?

sdk/keyvault/azsecrets/models.go
```diff
@@ -0,0 +1,189 @@
...
+	Attributes *SecretAttributes `json:"attributes,omitempty"`
+
+	// Type of the secret value such as a password.
+	ContentType *string `json:"contentType,omitempty"`
+
+	// Secret identifier.
+	ID *string `json:"id,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
JeffreyRichter, line 138: Kill the *?

sdk/keyvault/azsecrets/client.go
seankane-msft: this is a copy of what the generated code does, I can make this change in the public API I believe but it would add extra cycles

sdk/keyvault/azsecrets/client.go
jhendrixMSFT: Yes, I believe this came up before.  We can change the name to Transport.

sdk/keyvault/azsecrets/client.go
seankane-msft: In the go docs, there is also a sample that shows how to use the `BackupSecret` to get the backup []byte, and then use it directly in this call.

sdk/keyvault/azsecrets/client.go
seankane-msft: I think this should be a ptr for the same reason Jeff mentioned above.

sdk/keyvault/azsecrets/client.go
JeffreyRichter: I think "Tags *map[string]string" (* before map; not after ]) supports our nullable common denominator.
Setting Tags to null is how we'd delete the tags field for JSON patch (which is not supported for this particular scenario).
In this case, setting the value to null has no meaning which is why this looks wrong to me.
I'm also concerned about the round tripping: it might be useful to get Tags for 1 blob and then apply them to a new blob and so both the GET & PUT need a Tags map of the same type (map[string]string).

jhendrixMSFT: The pattern we have today is that to remove an item you would specify `fooMap["key"] = nil` which sends `{"fooMap": { "key": null } }` across the wire.  Doesn't tags behave this way?

JeffreyRichter: Yes, I see your point. No, Tags do not work like this. The Storage service doesn't support PATCH at all (and probably never will).
We could leave this as is as long as the round-tripping works easily. But, I do think the *string hurts usability for Tags since setting to null is not a thing.

jhendrixMSFT: I see.  Agreed that it hurts the usability then and should be removed in the public API.

sdk/keyvault/azsecrets/utils_test.go
```diff
@@ -87,9 +87,9 @@ func createClient(t *testing.T) (*Client, error) {
 
 	var cred azcore.TokenCredential
 	if recording.GetRecordMode() == "record" {
-		tenantId := lookupEnvVar("KEYVAULT_TENANT_ID")
-		clientId := lookupEnvVar("KEYVAULT_CLIENT_ID")
-		clientSecret := lookupEnvVar("KEYVAULT_CLIENT_SECRET")
+		tenantId := lookupEnvVar("KEYVAULT/AZSECRETS_TENANT_ID")
```
jhendrixMSFT, line 90: This is a weird format and doesn't work in PS.

JeffreyRichter (approved):
I'm approving because we've reviewed the API and I'm assuming the changes were made. I can't review this whole PR because all the files are new and there are no diffs.

sdk/keyvault/azsecrets/utils_test.go
seankane-msft: This passed the live pipeline, which I believe uses PS to deploy the resources. However, there is a bug in the deploy script for go that ben is aware of and this will go away in the future.

benbp: It passes in the live test pipeline because presumably the code works when `tenantId` equals an empty string. I'm going to update the deployment script to parse service directories with multiple segments.

benbp: Given that this variable reference does not work, you should change this string to the one that will work `AZSECRETS_TENANT_ID` in case you merge before my fix.

jhendrixMSFT (approved):

sdk/keyvault/azsecrets/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+```golang
+cred, err := azidentity.NewDefaultAzureCredential(nil)
+
+client, err := azsecrets.NewClient("https://my-key-vault.vault.azure.net/", cred, nil)
+```
+
+## Key concepts
+### Secret
+A secret consists of a secret value and its associated metadata and management information. This library handles secret values as strings, but Azure Key Vault doesn't store them as such. For more information about secrets and how Key Vault stores and manages them, see the [Key Vault documentation](https://docs.microsoft.com/azure/key-vault/general/about-keys-secrets-certificates).
```
richardpark-msft, line 113: > but Azure Key Vault doesn't store them as such

Just curious why this is - it seems a bit counter to how we normally expose APIs.

sdk/keyvault/azsecrets/README.md
```diff
@@ -0,0 +1,272 @@
...
+## Examples
+This section contains code snippets covering common tasks:
+* [Set a Secret](#set-a-secret "Set a Secret")
+* [Retrieve a Secret](#retrieve-a-secret "Retrieve a Secret")
+* [Update Secret metadata](#update-secret-metadata "Update Secret metadata")
+* [Delete a Secret](#delete-a-secret "Delete a Secret")
+* [List Secrets](#list-secrets "List Secrets")
+
+### Set a Secret
+[SetSecret](https://aka.ms/azsdk/go/keyvault) creates new secrets and changes the values of existing secrets. If no secret with the given name exists, `SetSecret` creates a new secret with that name and the given value. If the given name is in use, `SetSecret` creates a new version of that secret, with the given value.
```
richardpark-msft, line 126: aka.ms links are temporary or permanent this time around?

sdk/keyvault/azsecrets/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+```golang
+cred, err := azidentity.NewDefaultAzureCredential(nil)
+client, err := azsecrets.NewClient"https://my-key-vault.vault.azure.net/", cred, nil)
```
richardpark-msft, line 130:
```suggestion
client, err := azsecrets.NewClient("https://my-key-vault.vault.azure.net/", cred, nil)
```

richardpark-msft: I see a few spots where I think this typo was made.

sdk/keyvault/azsecrets/README.md
```diff
@@ -0,0 +1,272 @@
...
+          "sku": { "name": "standard" },
+          "tenantId": "...",
+          "vaultUri": "https://my-key-vault.vault.azure.net/"
+      },
+      "resourceGroup": "my-resource-group",
+      "type": "Microsoft.KeyVault/vaults"
+  }
+  ```
+
+  > The `"vaultUri"` property is the `vault_url` used by [azsecrets.NewClient][secret_client_docs]
```
richardpark-msft, line 59: I'm guessing the field is more like VaultURL, right? vault_url would be a bit more pythonic than I would expect.

sdk/keyvault/azsecrets/README.md
seankane-msft: Hoping for temporary!

PR #15611, opened 2021-09-22
[service-bus] Removing intermediate code, introducing abstractions for easier testing

richardpark-msft:
This is (unfortunately) a rather large PR. It's still "incomplete" but I think it's a good enough section of code to commit and I've basically done it _twice_ at this point trying to shrink it down. There is a whole slew of testing, etc, that I am planning on doing as follow on work to this, but it's meant to set a good foundation for everything to come.

Now, with that all said, here's a summary of the changes:
- Most of the internal code that used to handle interfacing with go-amqp is gone. Some of the code still remains because it's used for the ATOM link but I'll be replacing that soon enough with work that @chlowell is doing (hopefully I got the right Charles on that one)
- Namespace now holds onto an amqp.Client instance (and guards it with its life) - Namespace is basically the AMQP factory now, responsible for creating most everything to make unit testing simpler. There are some small exceptions but for the most part everything is an interface so we can mock it. 
- `AMQPLinks` exists to make life easier for associating a mgmt link to a sender/receiver, as well as tracking the lifetime of a session. It also hosts the Recover() function now, so we don't have to spread that out amongst every link-ish struct we've got in the tree. This also brings us closer to what .NET does - 1 mgmt link _per_ sender/receiver. This hasn't proven problematic for them and it simplifies things for us.
- retrier.go is a basic attempt at using the backoff.Backoff code that we put into eventhubs but making a simple iterative model, rather than requiring a callback. 

This is definitely a work in progress type of PR, but we're at the point now where we need to start jettisoning the old code. On the plus side I think my live test coverage **increased**.


richardpark-msft:
NOTE for reviewers: I'm not sure I understand why the overall azure-sdk-for-go build fails but I've not modified anything except for azservicebus and that CI build _does_ pass. To make that work I had to alter the common code coverage configuration file (which is in 'eng') which triggered a bunch of unrelated build pipelines.

sdk/messaging/azservicebus/go.mod
```diff
@@ -3,21 +3,20 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
 go 1.16
 
 require (
-	github.com/Azure/azure-amqp-common-go/v3 v3.1.0
+	github.com/Azure/azure-amqp-common-go/v3 v3.2.0
 	github.com/Azure/azure-sdk-for-go v51.1.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.7.0
 	github.com/Azure/go-amqp v0.15.0
 	github.com/Azure/go-autorest/autorest v0.11.18
```
jhendrixMSFT, line 10: These `go-autorest` dependencies will be gone at some point I assume.

sdk/messaging/azservicebus/internal/retrier.go
```diff
@@ -0,0 +1,100 @@
...
+	//    }
+	//
+	Try(ctx context.Context) bool
+}
+
+// Encapsulates a backoff policy, which allows you to configure the amount of
+// time in between retries as well as the maximum retries allowed (via MaxRetries)
+// NOTE: this should be copied by the caller as it is stateful.
+type BackoffRetrier struct {
+	Backoff    backoff.Backoff
```
jhendrixMSFT, line 52: I'd prefer if we didn't use anything from `backoff` as an exchange type.

sdk/messaging/azservicebus/sender.go
```diff
@@ -5,36 +5,36 @@ package azservicebus
...
-	legacySender internal.LegacySender
-
-	linkState *linkState
-
-	// for testing
-	nsCreateLegacySender func(ctx context.Context, entityPath string) (internal.LegacySender, error)
-}
+	// SendableMessage are sendable using Sender.SendMessage.
+	// Message, MessageBatch implement this interface.
+	SendableMessage interface {
```
jhendrixMSFT, line 28: Do we need to export this interface?  As a general rule, given the brittle nature of interfaces, we should strive to avoid exporting them.

sdk/messaging/azservicebus/sender.go
```diff
@@ -77,83 +73,75 @@ func (s *Sender) NewMessageBatch(ctx context.Context, options ...MessageBatchOpt
...
-		linkState: newLinkState(context.Background(), errClosed{link: "sender"}),
+		queueOrTopic: queueOrTopic,
 	}
 
-	sender.nsCreateLegacySender = ns.NewLegacySender
+	sender.links = ns.NewAMQPLinks(queueOrTopic, sender.createSenderLink)
 	return sender, nil
 }
+
+// handleAMQPError is called internally when an event has failed to send so we
```
jhendrixMSFT, line 118: Can this comment block be removed or is it waiting for something?

sdk/messaging/azservicebus/sender.go
```diff
@@ -57,18 +57,14 @@ func MessageBatchWithMaxSize(maxSizeInBytes int) func(options *messageBatchOptio
 // messages. Sending a batch of messages is more efficient than sending the
 // messages one at a time.
 func (s *Sender) NewMessageBatch(ctx context.Context, options ...MessageBatchOption) (*MessageBatch, error) {
-	if s.linkState.Closed() {
-		return nil, s.linkState.Err()
-	}
-
-	legacySender, err := s.createAmqpSender(ctx)
+	sender, _, _, _, err := s.links.Get(ctx)
```
jhendrixMSFT, line 60: The dogsled makes me think this method needs some refactoring (note, I haven't looked at it yet).

sdk/messaging/azservicebus/stress/stress.go
```diff
@@ -85,37 +85,42 @@ func runBasicSendAndReceiveTest() {
 	cleanupQueue := createQueue(telemetryClient, cs, queueName)
 	defer cleanupQueue()
 
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour*24*5)
```
jhendrixMSFT, line 88: Time intervals are usually written as "value * unit" so that it reads naturally, i.e. `5*24*time.Hour`

sdk/messaging/azservicebus/message.go
```diff
@@ -4,105 +4,331 @@
...
 
 import (
+	"context"
+	"fmt"
+	"time"
+
 	"github.com/Azure/azure-amqp-common-go/v3/uuid"
-	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
 	"github.com/Azure/go-amqp"
+	"github.com/Azure/go-autorest/autorest/to"
```
jhendrixMSFT, line 13: This should be removed.  If there is a large need you can use `azcore/to` instead.

sdk/messaging/azservicebus/message.go
```diff
@@ -4,105 +4,331 @@
...
 
-const (
-	annotationPartitionKey    = "x-opt-partition-key"
-	annotationViaPartitionKey = "x-opt-via-partition-key"
+type (
+	// ReceivedMessage is a received message from a Client.NewReceiver() or Client.NewProcessor().
+	ReceivedMessage struct {
+		Message
+
+		LockToken              *amqp.UUID
```
jhendrixMSFT, line 22: This exchange type, and potentially UUIDs in general, needs to be revisited.

sdk/messaging/azservicebus/message.go
```diff
@@ -4,105 +4,331 @@
...
 	// TODO: I don't think this should be strictly required. Need to
 	// look into why it won't send properly without one.
+	var messageID = m.ID
+
 	if messageID == "" {
 		uuid, err := uuid.NewV4()
 
-		if err != nil {
-			return nil, err
+		if err == nil {
```
jhendrixMSFT, line 116: What happens in the error case?

sdk/messaging/azservicebus/message_test.old
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: Does this file need to be here?

sdk/messaging/azservicebus/message.old
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: Does this file need to be here?

sdk/messaging/azservicebus/messageBatch.go
```diff
@@ -111,12 +104,16 @@ func (mb *MessageBatch) Len() int {
 
 // toAMQPMessage converts this batch into a sendable *amqp.Message
 // NOTE: not idempotent!
-func (mb *MessageBatch) toAMQPMessage() (*amqp.Message, error) {
+func (mb *MessageBatch) ToAMQPMessage() *amqp.Message {
```
jhendrixMSFT, line 107: Does this need to be exported?  Seems like an internal implementation detail.

sdk/messaging/azservicebus/messageBatch.go
```diff
@@ -111,12 +104,16 @@ func (mb *MessageBatch) Len() int {
...
-		mb.batchEnvelope.Data[idx] = bytes
-	}
-	return mb.batchEnvelope, nil
+	copy(mb.batchEnvelope.Data, mb.marshaledMessages)
+
+	return mb.batchEnvelope
+}
+
+// MessageType indicates the type of this message. Used for tracing.
+func (mb *MessageBatch) MessageType() string {
```
jhendrixMSFT, line 117: Unexport?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -196,120 +200,118 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, numMessages int, options
...
+	for {
+		// TODO: should we do retries here? Or should we just surface the error close to immediately?
+		amqpMessage, err := receiver.Receive(ctx)
+
+		if err != nil {
+			// at the moment all errors from Receive() are fatal (if they're not context cancellation errors)
+			// for simplicity, we'll just store whatever we got and return it later if nothing else goes bad.
+			fatalError = err
+
+			if err != context.Canceled && err != context.DeadlineExceeded {
```
jhendrixMSFT, line 235: Use `errors.Is()` for sentinel error checks.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -196,120 +200,118 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, numMessages int, options
...
-		}
-
-		return nil
-	}))
-
-	select {
-	case <-listenHandle.Done():
-		if listenHandle.Err() != allMessagesReceivedError {
-			err = listenHandle.Err()
+			go func() {
```
jhendrixMSFT, line 259: What is this doing?  Comment would be good.

sdk/messaging/azservicebus/processor.go
```diff
@@ -189,47 +181,64 @@ func newProcessor(ns legacyNamespace, options ...ProcessorOption) (*Processor, e
...
 				}
 
 				// we retry infinitely, but do it in the pattern they specify via their retryOptions for each "round" of retries.
-				retry := p.subscribe(ctx, receiver, p.config.ShouldAutoComplete, handleMessage, handleError)
+				err = p.subscribe(ctx, receiver, handleMessage, handleError)
 
-				if retry {
-					return true, amqpCommon.Retryable("")
+				if err != nil {
+					if isRetryableSubscribeError(err) {
```
jhendrixMSFT, line 230: I believe this line can be condensed into its parent?

sdk/messaging/azservicebus/go.mod
richardpark-msft: We do have an HTTP client in here (it's for the ATOM endpoint that lets you manage Service Bus resources). 

Would we still remove it then? I'm not familiar with the package deps for HTTP stacks in Go.

sdk/messaging/azservicebus/sender.go
richardpark-msft: This interface is used for input only (which I thought was okay).

sdk/messaging/azservicebus/sender.go
richardpark-msft: The easiest one would be to just return a struct, I think.

sdk/messaging/azservicebus/go.mod
jhendrixMSFT: Yep, this is all track 1 artifacts that need to go.

sdk/messaging/azservicebus/internal/retrier.go
richardpark-msft: Does it change things if this is only in the `internal` package?

sdk/messaging/azservicebus/go.mod
richardpark-msft: Okay, cool. Yeah, I'll see about removing this in a future PR. I might have to push till after the first beta.

sdk/messaging/azservicebus/internal/retrier.go
jhendrixMSFT: No, because we don't control the definition of `backoff.Backoff` which might change in the future.

sdk/messaging/azservicebus/go.mod
jhendrixMSFT: If you just need an HTTP client you can use `net/http.DefaultClient` from the standard library.

sdk/messaging/azservicebus/processor.go
richardpark-msft: I don't think so, only because there is a default 'err != nil' behavior too (ie, if it's not retryable we exit the retries altogether).

TBH, there are some issues with how errors are handled here that I am planning on going through in a recovery PR. I've learned quite a bit since I first wrote this. :)

sdk/messaging/azservicebus/sender.go
jhendrixMSFT: What types implement this?  It seems like the methods on the interface are really internal implementation details.

sdk/messaging/azservicebus/go.mod
```diff
@@ -3,21 +3,20 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
 go 1.16
 
 require (
-	github.com/Azure/azure-amqp-common-go/v3 v3.1.0
+	github.com/Azure/azure-amqp-common-go/v3 v3.2.0
```
jhendrixMSFT, line 6: Also, is there an issue tracking removal of this dependency?  I know we talked about moving this to an internal package to be shared with EH.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Comment added

sdk/messaging/azservicebus/go.mod
richardpark-msft: Created: https://github.com/Azure/azure-sdk-for-go/issues/15646

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Done.

I made an isCancelled() function and just went and replaced all the spots where I was actually doing this incorrectly.

sdk/messaging/azservicebus/sender.go
richardpark-msft: Gone. I can always get it back from history later if needed.

sdk/messaging/azservicebus/stress/stress.go
richardpark-msft: Done

sdk/messaging/azservicebus/message.old
richardpark-msft: No, deleted.

sdk/messaging/azservicebus/messageBatch.go
richardpark-msft: No, byproduct of a separate idea. Unexported.

sdk/messaging/azservicebus/messageBatch.go
richardpark-msft: Yes, unexported.

sdk/messaging/azservicebus/sender.go
richardpark-msft: Message and MessageBatch (and when it comes in - an external form of *amqp.Message).

The other option, which might be more appropriate, is to have separate send functions for each type instead.

So:

SendMessage
SendMessageBatch
SendAMQPMessage

Would separate functions be better?

sdk/messaging/azservicebus/message_test.old
richardpark-msft: No, deleted.

sdk/messaging/azservicebus/sender.go
richardpark-msft: I can change this in the next PR.

sdk/messaging/azservicebus/message.go
richardpark-msft: They just end up with an empty message ID.

There's no good default to give it here, unfortunately.

sdk/messaging/azservicebus/message.go
richardpark-msft: Removed all autorest/to usages, replaced with azcore/to for now.

sdk/messaging/azservicebus/sender.go
jhendrixMSFT: The pattern here is to leave the interface type as exported but to _not_ export the methods since these are intended to be implemented internally.  This means that customers can't implement this interface (we don't want them to) thus allows us to change/extend it in a non-breaking way.

sdk/messaging/azservicebus/processor.go
jhendrixMSFT: Ah you're right, I missed that.

sdk/messaging/azservicebus/internal/retrier.go
richardpark-msft: Made the entire `BackoffRetrier` an internal type instead.

sdk/messaging/azservicebus/message.go
richardpark-msft: To clarify - do you mean in this PR? Or do we already have a conversation happening?

sdk/messaging/azservicebus/sender.go
richardpark-msft: I think I've got it. Does this seem right then (it's what I've just pushed):

```go
SendableMessage interface {
  toAMQPMessage() *amqp.Message
  messageType() string
}
```

jhendrixMSFT: Yep!

sdk/messaging/azservicebus/message.go
jhendrixMSFT: How do callers use the `LockToken`, can we simply expose it as a `[16]byte`?

For some context.  In track 1 we exposed UUIDs using a thrid-party UUID type which was problematic for various reasons.  The conclusion we arrived at (for HTTP-based services) was to expose UUIDs in their native format (strings) so that callers can use their UUID-parsing package of choice.

jhendrixMSFT: This would also imply that I need to remove the `amqp.UUID` type.

richardpark-msft: Took your suggestion, switch it to [16]byte array.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -0,0 +1,168 @@
...
+	"errors"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
+	"github.com/Azure/go-amqp"
+)
+
+var errReceiveAndDeleteReceiver = errors.New("messages that are received in receiveAndDelete mode are not settlable")
+
+type messageSettler struct {
```
jhendrixMSFT, line 17: Should this go under `internal`, how are we making that distinction?

jhendrixMSFT: I'm not saying it should, just wondering how you're thinking about when things should/shouldn't.

jhendrixMSFT: Maybe there's a more general pattern to distill here and share across the team.

richardpark-msft: This is a goofy one.

Yes, I wanted to put under `internal` _but_ it references `Message`, which is in the public surface. So that caused a circular reference.

Is there a way to work around that aside from creating a common package between the two? I didn't want to do all of that in this PR, but I did have the same desire.

jhendrixMSFT: Ah I see your point.  I don't think there's a good way to achieve this separation outside of what you recommended, however I'm not sure if the work is really worth it given the narrow scope of this type (well, I assume the scope is narrow but maybe I'm mistaken).  As long as we follow our rule about only using "exported" fields/methods it's probably good enough (I wonder if there's a linter rule for this....).

sdk/messaging/azservicebus/processor.go
```diff
@@ -33,27 +34,21 @@ type processorConfig struct {
 
 // Processor is a push-based receiver for Service Bus.
 type Processor struct {
-	// configuration data that is read-only after the Processor has been created
-	config processorConfig
+	*messageSettler
```
jhendrixMSFT, line 37: Oh I just noticed this.  It means that this type is part of public surface area.  Assuming this was intentional, can you add a comment to `messageSettler` indicating that it's not really internal?

It would be ideal to avoid this if possible (mostly I worry about inadvertently introducing a breaking change to `messageSettler`).

richardpark-msft: I think I'll just hide it for real, instead of doing what I'm doing here. Principle of least surprise and all that.

richardpark-msft: Okay, done. messageSettler is no longer embedded.

sdk/messaging/azservicebus/messageSettler.go
richardpark-msft: For now I'll leave it. It might be nice to rework it a bit because the top level of this package is stuffed with files and it'd be nice if it was only private stuff.

jhendrixMSFT (approved):

PR #15623, opened 2021-09-23
fix: some compatible issue in build.ps1 after unify module discovery

tadelesh:

build.ps1 can not work after cleaning up the module discovery. This PR fix it and change all the fast track build script of existed RP and tooling.


eng/scripts/build.ps1
```diff
@@ -1,6 +1,7 @@
 #Requires -Version 7.0
 param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]$skipBuild, [switch]$cleanGenerated, [switch]$format, [switch]$tidy, [string]$config = "autorest.md", [string]$outputFolder)
 
+. $PSScriptRoot/meta_generation.ps1
```
ArcturusZhang, line 4: I have a question about this one: what does this used for?

tadelesh: It is used for old _meta.json config converting. As it is deprecated, I've removed related logic.

tools/generator/ci.yml
```diff
@@ -25,3 +25,4 @@ stages:
   - template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
     parameters:
       ServiceDirectory: '../tools/generator'
+      RunChangelogCheck: false
```
weshaggard, line 28: Things under the tools repo do they actually need a full pipeline? Do we release them? I suggest instead of making this specific to verify changelog we might need to have a parameter that indicates that this is not a shipping package and we turn off any thing related to shipping/releasing. There are other things like the release template that might fail for this package also. We could even consider looking at ServiceDirectory and enabling those features only for things under the sdk folder.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -1,13 +1,16 @@
...
-    type: string
+  - name: RunTests
+    type: boolean
     default: false
-  - name:  RunLiveTests
-    type: string
+  - name: RunLiveTests
+    type: boolean
     default: false
+  - name: RunChangelogCheck
```
weshaggard, line 11: We should consider a more general "NonShipping" parameter which would turn off validation, testing, releasing etc for things like the tools. Or we could also consider refactoring the existing templates to split out the pieces.

weshaggard (approved):
Added some minor feedback that you can consider addressing after merging.

tools/generator/ci.yml
tadelesh: I've changed the flag to 'NonShipping'. I think splitting the pipeline template for tooling and package shipping is a good idea.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
tadelesh: Same as above.

Issue #15610, opened 2021-09-22
[service-bus] Error handling needs work to be consistent and reliable


richardpark-msft:
As part of a recent refactor it's become more obvious that the error handling in Service Bus needs to be made more consistent. There are some good classification functions in the older code (which needs to be brought forward) and basically we just need to have enough information to say:

- This link is dead and needs to restart
- There is a retryable issue here (transient) and we should just try it again

This affects the Receiver and Sender a bit more since the Processor doesn't really make any distinctions about what's retryable vs not.

Tagging @jhendrixMSFT on this as we've had discussion about how we might provide some interfaces or similar to make the classification easier for client code.

richardpark-msft:
I'm taking a first stab at this for beta 1 (October release).

The big things:
- The processor type brings in built-in infinite retrying, which should be more durable than the previous APIs. 
- Our error classification has been improved - much thanks to the go-shuttle project (and @serbrech). I've added in some further checking and also allow for link and connection level recovery (we fallback to connection more than I'd like, but it's a first go).

As a first beta release bar I've got a stress test running, with network outages every minute for 5 seconds. This gives us some decent recovery testing, while also getting a good ballpark on our longevity.

richardpark-msft:
Some of this work has been incorporated by not requiring the user to intervene. We still report the error but the next call to Receiver.ReceiveMessages() will repair it.

We still want to add in some classification, as that will help customers if they want to decide to make that next call or not,.

richardpark-msft:
Note, another interesting discussion was here:
https://github.com/Azure/azure-service-bus-go/issues/96

When applicable, we should return a structured error. The equivalent of a 404 here would be useful.

ryepup:
My team is working on a servicebus integration, and running into a problem differentiating recoverable and unrecoverable errors.

Specifically we're writing a function to ensure a topic is defined, calling `GetTopic` to see if it exists, and a `CreateTopic` if needed. We're not seeing an official way to differentiate between recoverable "topic not found" errors and unrecoverable network errors.

```go
resp, err := client.GetTopic(ctx, "topic", &admin.GetTopicOptions{})
if err != nil {
  // err could be an unrecoverable auth/network/etc problem
  // or a recoverable 404 because the topic is not defined
}
```

[...]

PR #15629, opened 2021-09-24
[service-bus] Adding in PeekMessages, DeferMessage+ReceiveDeferredMessages support

richardpark-msft:
Adding in some of the critical receiver methods for deferring and retrieving deferred messages. Also, added in the track 2 compliant PeekMessages function.

There were also some bugs in how we were handling ReceiveMessages, which were found in some tests that I added.

This PR is built on top of a pending PR (https://github.com/Azure/azure-sdk-for-go/pull/15611) so you'll want to start reviewing this code from commit 0f3d40327b9339c2b6ff05d3eb1767faa848761e (my linkageddon commit) to see the relevant bits.

I've added in a temporary `replace` in the go.mod file so I can pick up this fix in go-amqp that hasn't yet released: https://github.com/Azure/go-amqp/pull/72).

### DIFF
--- a/sdk/messaging/azservicebus/go.mod
+++ b/sdk/messaging/azservicebus/go.mod
@@ -22,3 +22,6 @@ require (
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	nhooyr.io/websocket v1.8.6
 )
+
+// remove after go-amqp releases https://github.com/Azure/go-amqp/pull/72
+replace github.com/Azure/go-amqp v0.15.0 => github.com/Azure/go-amqp v0.15.1-0.20210923181113-8f9a02b39d60
--- a/sdk/messaging/azservicebus/go.mod
+++ b/sdk/messaging/azservicebus/go.mod
@@ -7,13 +7,12 @@ require (
 	github.com/Azure/azure-sdk-for-go v51.1.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.19.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.11.0
-	github.com/Azure/azure-sdk-for-go/sdk/internal v0.7.0
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.7.1
 	github.com/Azure/go-amqp v0.15.0
 	github.com/Azure/go-autorest/autorest v0.11.18
 	github.com/Azure/go-autorest/autorest/adal v0.9.13
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/devigned/tab v0.1.1
-	github.com/google/uuid v1.3.0
 	github.com/joho/godotenv v1.3.0
 	github.com/jpillora/backoff v1.0.0
 	github.com/microsoft/ApplicationInsights-Go v0.4.4
--- a/sdk/messaging/azservicebus/internal/amqp_test_utils.go
+++ b/sdk/messaging/azservicebus/internal/amqp_test_utils.go
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package internal
+
+import (
+	"context"
+	"fmt"
+)
+
+type fakeNS struct {
+	claimNegotiated int
+	MgmtClient      MgmtClient
+	Session         AMQPSessionCloser
+}
+
+type fakeAMQPSender struct {
+	closed int
+	AMQPSender
+}
+
+type fakeAMQPSession struct {
+	AMQPSessionCloser
+	closed int
+}
+
+type fakeMgmtClient struct {
+	MgmtClient
+	closed int
+}
+
+type FakeAMQPLinks struct {
+	AMQPLinks
+
+	// values to be returned for each `Get` call
+	Revision uint64
+	Receiver AMQPReceiver
+	Sender   AMQPSender
+	Mgmt     MgmtClient
+	Err      error
+}
+
+func (l FakeAMQPLinks) Get(ctx context.Context) (AMQPSender, AMQPReceiver, MgmtClient, uint64, error) {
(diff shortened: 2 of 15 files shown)


sdk/messaging/azservicebus/go.mod
```diff
@@ -3,22 +3,24 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
...
 	github.com/Azure/azure-sdk-for-go v51.1.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.7.0
 	github.com/Azure/go-amqp v0.15.0
 	github.com/Azure/go-autorest/autorest v0.11.18
 	github.com/Azure/go-autorest/autorest/adal v0.9.13
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/Azure/go-autorest/autorest/to v0.4.0
-	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
 	github.com/devigned/tab v0.1.1
+	github.com/google/uuid v1.3.0
```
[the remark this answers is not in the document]
richardpark-msft: Ooh, good idea. Let me check it out. monorepo has it's advantages.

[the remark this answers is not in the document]
richardpark-msft: I feel absolutely silly now too.

The only usage left after a purge I'd done before was this:

```go
uuid.UUID(expectedAMQPEncodedLockTokenGUID)
```

Which can just be:

```go
expectedAMQPEncodedLockTokenGUID
```

:facepalm

All gone now. :)

seankane-msft (approved):
LGTM :shipit:

sdk/messaging/azservicebus/internal/amqp_test_utils.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: If these artifacts are used solely for tests you should rename the file to amqp_utils_test.go so they're only available in tests.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -69,13 +70,10 @@ type ReceiverOption func(receiver *Receiver) error
 // for a queue or subscription.
 func ReceiverWithSubQueue(subQueue SubQueue) ReceiverOption {
 	return func(receiver *Receiver) error {
-		switch subQueue {
-		case SubQueueDeadLetter:
-		case SubQueueTransfer:
-		case "":
+		if subQueue == SubQueueDeadLetter || subQueue == SubQueueTransfer {
```
jhendrixMSFT, line 73: I see this logic was duplicated elsewhere.  Consider moving it to some internal validation func or maybe a method on SubQueue depending on where it's needed.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -270,8 +267,35 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, maxMessages int, options
 		}
 	}
 
+	if fatalError != nil {
+		if isCancelled(fatalError) {
+			select {
+			case <-userCtx.Done():
+				// if _we_ cancelled the context it'll just because the normal timeouts
+				// fired, which isn't an error. So we only care if the user did it.
+				break
```
jhendrixMSFT, line 276: Shouldn't this be a `return`?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -400,5 +541,11 @@ func createLinkOptions(mode ReceiveMode, entityPath string) []amqp.LinkOption {
 }
 
 func isCancelled(err error) bool {
-	return errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)
+	if errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded) ||
+		err.Error() == "context canceled" { // go-amqp is returning this when I cancel
```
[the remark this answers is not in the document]
richardpark-msft: I really want to test this in a more controlled way because it sure seems like I should have.

I've got this issue to really take a look at error classification:
https://github.com/Azure/azure-sdk-for-go/issues/15610

sdk/messaging/azservicebus/internal/amqp_test_utils.go
richardpark-msft: I think I'm getting bit again by this whole separate-package bit again. 

These are also used in the main package's tests and the _test files in `internal` don't get included in that case.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Funnily enough, I actually created a struct specifically for this kind of logic and never took advantage of this. Updated.

sdk/messaging/azservicebus/internal/amqp_test_utils.go
jhendrixMSFT: Ah yes good point.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -358,15 +493,25 @@ func (e *entity) String() (string, error) {
...
-	} else if e.Subqueue == SubQueueTransfer {
+	} else if e.subqueue == SubQueueTransfer {
 		entityPath += "/$Transfer/$DeadLetterQueue"
 	}
 
 	return entityPath, nil
 }
 
+func (e *entity) SetSubQueue(subQueue SubQueue) error {
+	if subQueue == SubQueueDeadLetter || subQueue == SubQueueTransfer {
```
jhendrixMSFT, line 506: Centralize?

richardpark-msft: I think you caught this in between reviews.

This is where I centralized it. The two ReceiverWithSubQueue and ProcessorWithSubQueue methods just call into here now. So there's not 3, just 1.

PR #15630, opened 2021-09-24
fix: release pipeline condition problem

tadelesh:


eng/scripts/Language-Settings.ps1
```diff
@@ -23,8 +23,8 @@ function Get-GoModuleVersionInfo($modPath)
     }
   }
 
-  LogWarning "Unable to find version for $modPath"
-  return $null
+  LogWarning "Unable to find version for $modPath, considered to be onboard rp package"
```
weshaggard, line 26: This is very specific to the mgmt libraries. Can we not just fix the missing versioning instead trying to fake the version?

weshaggard (changes requested):
The condition change looks correct but I don't like the change to module discovery. We should instead add a version file to the module to fix any modules that are missing.

tadelesh:
> The condition change looks correct but I don't like the change to module discovery. We should instead add a version file to the module to fix any modules that are missing.

When onboarding a new RP, the package folder won't contain any go code but some config for codegen. So if we use build.ps1 to do the code generation, module discovery will skip the package to be onboard. I agree with you that changing module discovery functionality is not a good implementation to handle such onboard situation. Instead, I add some onboard logic to build.ps1.

eng/scripts/build.ps1
```diff
@@ -3,70 +3,91 @@ param([string]$filter, [switch]$clean, [switch]$vet, [switch]$generate, [switch]
...
+    foreach ($sdk in $sdks)
+    {
         Push-Location $sdk.DirectoryPath
         Process-Sdk
         Pop-Location
     }
+
+    if ($sdks.Count -eq 0 -and $filter)
+    {
+        Write-Host "Cannot find go module under $filter, considered to be an onboard RP package"
```
weshaggard, line 84: Is this going to work in the general case? What if someone passes an invalid filter? Should we at least check to see if the location exists and if not then error in a different way instead of trying to process?

weshaggard: Or are there any other flags that you are passing that might indicate that you are in an on-boarding process?

weshaggard (approved):
The change looks reasonable but I do think we might need to tweak the logic to determine we are in on-boarding mode instead of other error modes.

eng/scripts/build.ps1
tadelesh: Add config existing test.

eng/scripts/build.ps1
```diff
@@ -9,75 +9,96 @@ param(
...
     $sdks = Get-AllPackageInfoFromRepo $filter
 
-    foreach ($sdk in $sdks) {
+    foreach ($sdk in $sdks)
+    {
         Push-Location $sdk.DirectoryPath
         Process-Sdk
         Pop-Location
     }
+
```
weshaggard, line 92: future clean-up I would capture `Join-Path $RepoRoot "sdk" $filter` once instead of duplicating it 3 times.

Issue #15632, opened 2021-09-24
InteractiveBrowserCredentialOptions.RedirectURL doesn't work


chlowell:
`InteractiveBrowserCredential.GetToken()` starts a redirect server only when `InteractiveBrowserCredentialOptions.RedirectURL` has no value:
https://github.com/Azure/azure-sdk-for-go/blob/6ff991256abe3ac3c5b5c43a977f34f230782a02/sdk/azidentity/interactive_browser_credential.go#L128-L132

Acquiring a token is therefore impossible when the option has a value. Fixing this is more complex than simply ensuring the redirect server starts, because the server doesn't work for all valid values of `RedirectURL`, e.g. `https://127.0.0.1`. However, MSAL for Go supports `RedirectURL`, so this should be easy to fix as part of #14920.

chlowell:
Switching to MSAL's implementation will fix this for the common case. However, an application that has registered `https://localhost` but not `http://localhost` for a redirect URI (which is unusual) won't be able to authenticate users until https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/281 is fixed.

PR #15634, opened 2021-09-24
Remove ClientSecret and Port from browser credential options

chlowell:
Removes a couple fields from InteractiveBrowserCredentialOptions:
- ClientSecret: not part of this credential's mainline scenario. We don't support it in other languages; neither does MSAL for Go.
- Port: redundant given we have RedirectURL, which is what we really need. Also not supported by MSAL for Go.

Speaking of RedirectURL, while doing this I discovered it doesn't work. I don't want to remove it, because we do need it, but neither do I want to fix it, because adopting MSAL entails deleting the code that needs fixing. So I instead documented that it doesn't work, and opened #15632 to track fixing it.

### DIFF
--- a/sdk/azidentity/interactive_browser_credential.go
+++ b/sdk/azidentity/interactive_browser_credential.go
@@ -19,24 +19,18 @@ import (
 	"github.com/pkg/browser"
 )
 
-// InteractiveBrowserCredentialOptions can be used when providing additional credential information, such as a client secret.
+// InteractiveBrowserCredentialOptions provides optional configuration.
 // Use these options to modify the default pipeline behavior if necessary.
 // All zero-value fields will be initialized with their default values. Please note, that both the TenantID or ClientID fields should
 // changed together if default values are not desired.
 type InteractiveBrowserCredentialOptions struct {
-	// The Azure Active Directory tenant (directory) ID of the service principal.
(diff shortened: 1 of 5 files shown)


sdk/azidentity/interactive_browser_credential.go
```diff
@@ -125,10 +119,7 @@ func interactiveBrowserLogin(ctx context.Context, authorityHost string, opts *In
 	if err != nil {
 		return nil, err
 	}
-	redirectURL := opts.RedirectURL
-	if redirectURL == "" {
-		redirectURL = rs.Start(state.String(), opts.Port)
-	}
+	redirectURL := rs.Start(state.String())
```
jhendrixMSFT, line 122: We really always need to run this local server?  This part was always unclear to me.

chlowell: Yes. The basic flow is:
1. open a browser to an AAD sign in page, providing a redirect URL in the query
1. user authenticates
1. AAD redirects to the given URL with an authorization code in the query
1. application redeems the authorization code for an access token

This flow was designed for web apps. To use it in local development, we need something local to receive the redirect and expose the auth code to our credential.

jhendrixMSFT (approved):

PR #15639, opened 2021-09-24
[service-bus] Azure Identity support

### LINKED ISSUE #15542: [service-bus] Adjust negotiateClaim's call to be calculated based on token expiration
Currently our negotiateClaim refresh takes place every 15 minutes (this is just baked into the code, it's not calculated). 

Other track 2 SDKs do a calculation. For instance, this is JS's:

For SAS tokens:
```typescript
this._tokenTimeout = (3600 - 900) * 1000;
```

For AAD based tokens:
```typescript
this._tokenTimeout = tokenObject.expiresOnTimestamp - Date.now() - 2 * 60 * 1000;
```

(C# has a similar calculation)

We should do something similar since these have been proven to be practical.

### LINKED ISSUE #15638: ServiceBus client should accept Azure Identity as an authentication mechanism
ServiceBusClient accepts a connection string, but should also accept a TokenCredential, like the ones from Azure Identity

richardpark-msft:
This PR allows you to pass in a azure identity TokenCredential and authenticate to Service Bus.

I also took this opportunity to fix our refresh calculation interval to be dynamic when we use TokenCredentials since they have a built in expiration date. JS had some nice "wiggle-room" calculations which I've also incorporated.

Note for reviewers: this PR is built on top of https://github.com/Azure/azure-sdk-for-go/pull/15611. You'll want to review the code after this commit: https://github.com/Azure/azure-sdk-for-go/commit/0f3d40327b9339c2b6ff05d3eb1767faa848761e to see the additions this PR is bringing in.

Fixes #15542
Fixes #15638

### DIFF
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -27,7 +27,7 @@ type Client struct {
 
 type clientConfig struct {
 	connectionString string
-	tokenCredential  azcore.TokenCredential
+	credential       azcore.TokenCredential
 	// the Service Bus namespace name (ex: myservicebus.servicebus.windows.net)
 	fullyQualifiedNamespace string
 }
@@ -39,10 +39,10 @@ type ClientOption func(client *Client) error
 // NewClient creates a new Client for a Service Bus namespace, using a TokenCredential.
 // A Client allows you create receivers (for queues or subscriptions) and senders (for queues and topics).
 // fullyQualifiedNamespace is the Service Bus namespace name (ex: myservicebus.servicebus.windows.net)
-// tokenCredential is one of the credentials in the `github.com/Azure/azure-sdk-for-go/sdk/azidentity` package.
-func NewClient(fullyQualifiedNamespace string, tokenCredential azcore.TokenCredential, options ...ClientOption) (*Client, error) {
+// credential is one of the credentials in the `github.com/Azure/azure-sdk-for-go/sdk/azidentity` package.
+func NewClient(fullyQualifiedNamespace string, credential azcore.TokenCredential, options ...ClientOption) (*Client, error) {
 	return newClientImpl(clientConfig{
-		tokenCredential:         tokenCredential,
+		credential:              credential,
 		fullyQualifiedNamespace: fullyQualifiedNamespace,
 	}, options...)
 }
@@ -73,10 +73,10 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
 
 	if client.config.connectionString != "" {
 		nsOptions = append(nsOptions, internal.NamespaceWithConnectionString(client.config.connectionString))
-	} else if client.config.tokenCredential != nil {
+	} else if client.config.credential != nil {
 		option := internal.NamespacesWithTokenCredential(
 			client.config.fullyQualifiedNamespace,
-			client.config.tokenCredential)
+			client.config.credential)
 
 		nsOptions = append(nsOptions, option)
 	} else {
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -5,9 +5,11 @@ package azservicebus
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
 	"github.com/devigned/tab"
 )
@@ -15,9 +17,7 @@ import (
 // Client provides methods to create Sender, Receiver and Processor
 // instances to send and receive messages from Service Bus.
 type Client struct {
-	config struct {
-		connectionString string
-	}
+	config    clientConfig
 	namespace *internal.Namespace
 	linksMu   *sync.Mutex
 	links     []interface {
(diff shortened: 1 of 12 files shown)


sdk/messaging/azservicebus/client.go
```diff
@@ -29,14 +36,25 @@ type Client struct {
...
 	return func(client *Client) error {
 		client.config.connectionString = connStr
 		return nil
 	}
 }
 
+// WithTokenCredential configures a namespace to connect using a TokenCredential.
+// fullyQualifiedNamespace is the Service Bus namespace name (ex: myservicebus.servicebus.windows.net)
+// tokenCredential is one of the credentials in the `github.com/Azure/azure-sdk-for-go/sdk/azidentity` package.
+func WithTokenCredential(fullyQualifiedNamespace string, tokenCredential azcore.TokenCredential) ClientOption {
```
chlowell, line 50: Seems the parameter name could be shorter without creating confusion:
```suggestion
func WithTokenCredential(fullyQualifiedNamespace string, credential azcore.TokenCredential) ClientOption {
```

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -303,62 +308,74 @@ func (ns *Namespace) Recover(ctx context.Context) (*amqp.Client, error) {
...
+				amqpClient, err := nsGetAMQPClientImpl(ctx)
+
+				if err != nil {
+					span.Logger().Error(err)
+					return time.Time{}, err
+				}
+
+				token, expiration, err := ns.TokenProvider.GetTokenAsTokenProvider(audience)
+
+				if err != nil {
```
chlowell, line 343: What happens when refreshing the token fails when the token hasn't yet expired?

sdk/messaging/azservicebus/client.go
richardpark-msft: Done

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: I rebased so this might have moved your comment to the wrong line.

I think you're asking me whether the `expiration` there is used if there is a failure. We don't use it - we only use if it we succeeded.

There is a retry loop around the negotiate itself, so we do have some fallback logic there. There is also a "pause 1 minute and then try 5 times, per 5 seconds" check later in the function.

sdk/messaging/azservicebus/client.go
```diff
@@ -54,7 +69,21 @@ func NewClient(options ...ClientOption) (*Client, error) {
 	}
 
 	var err error
-	client.namespace, err = internal.NewNamespace(internal.NamespaceWithConnectionString(client.config.connectionString))
+	var nsOptions []internal.NamespaceOption
+
+	if client.config.connectionString != "" {
```
jhendrixMSFT, line 74: I'd move these validations into their respective constructors.

sdk/messaging/azservicebus/internal/token_provider.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/Azure/azure-amqp-common-go/v3/auth"
+	"github.com/Azure/azure-amqp-common-go/v3/sas"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// TokenProvider handles access tokens and expiration calculation for SAS
+// keys (via connection strings) or TokenCredentials from Azure Identity.
+type TokenProvider struct {
```
jhendrixMSFT, line 19: Does this type need to be exported?  Looks like you can only create it internally.

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -169,10 +165,19 @@ func NamespaceWithAzureEnvironment(namespaceName, environmentName string) Namesp
...
 
-// NamespaceWithTokenProvider sets the token provider on the namespace
-func NamespaceWithTokenProvider(provider auth.TokenProvider) NamespaceOption {
+// NamespacesWithTokenCredential sets the token provider on the namespace
+func NamespacesWithTokenCredential(namespaceName string, tokenCredential azcore.TokenCredential) NamespaceOption {
 	return func(ns *Namespace) error {
-		ns.TokenProvider = provider
+		ns.TokenProvider = newTokenProviderWithTokenCredential(tokenCredential)
+
+		parts := strings.SplitN(namespaceName, ".", 2)
```
jhendrixMSFT, line 173: Not sure what the format is supposed to be.  Should this go in the doc comment for this constructor?

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -371,7 +388,13 @@ func (ns *Namespace) negotiateClaimImpl(ctx context.Context,
...
 				return
 			case <-time.After(renewalDuration):
-				_ = refreshClaim() // logging will report the error for now
+				expiresOn, err = refreshClaim() // logging will report the error for now
+
+				if err != nil {
+					expiresOn = time.Now().Add(time.Minute) // check again in a minute
+				}
+
+				renewalDuration = time.Until(expiresOn)
```
jhendrixMSFT, line 397: Does this use the token's expiration time as-is, or do we adjust it to make the window smaller so that we attempt to refresh before the token is actually expired?

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: Ah, offline discussion clarified the question somewhat.

Service Bus authenticates differently than other HTTP stacks in that the token is not passed for each link's request, but instead done as a whole for the connection, out of band from the individual requests. It's described here if you're interested:

https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-protocol-guide#claims-based-authorization

So essentially this code is just requesting/updating a token from the TokenProvider and notifying Service Bus about it. All previous or current tokens are just local to this loop, there's no stored token state in our code.

sdk/messaging/azservicebus/client.go
richardpark-msft: Done

sdk/messaging/azservicebus/internal/token_provider.go
richardpark-msft: Done.

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: Updated. I realized after you wrote this that I probably shouldn't allow just a barename without a suffix.

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: Oof, yeah, you're right I need to do that.

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -37,7 +34,7 @@ type (
 	Namespace struct {
 		Name          string
 		Suffix        string
-		TokenProvider auth.TokenProvider
+		TokenProvider *tokenProvider
```
jhendrixMSFT, line 37: This also should not be exported then?

richardpark-msft: Haha, independently came to the same conclusion. Fixed.

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -442,16 +436,35 @@ func (ns *Namespace) getUserAgent() string {
...
+	const max = 49 * 24 * time.Hour
+	const clockDrift = 10 * time.Minute
+
+	var refreshDuration = expirationTime.Sub(currentTime) - clockDrift
+
+	if refreshDuration < min {
+		return min
+	} else if refreshDuration > max {
+		return max
+	} else {
```
jhendrixMSFT, line 467: Nit, you can drop the `else` and just `return refreshDuration`

richardpark-msft: Done

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: Ah, oops, I focused on the wrong thing there.

TokenProvider is used in some tests external from the internal package.

jhendrixMSFT (approved):

PR #15653, opened 2021-09-28
Fix release tool

tadelesh:


ArcturusZhang (approved):

eng/tools/generator/template/rpName/packageName/ci.yml.tpl
```diff
@@ -8,7 +8,7 @@ trigger:
       - release/*
   paths:
     include:
-    - sdk/{{rpName}}/{{packageName}}/
+    - sdk/resourcemanager/{{rpName}}/{{packageName}}/
```
weshaggard, line 11: @ArcturusZhang @tadelesh Is this an approved convention change? It will break a number of pieces of automation. 

FYI @RickWinter @jhendrixMSFT

weshaggard: for example https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L164 which is the automatic detection release go packages that gets added to the csv and release-notes. With this change that will break and before we go and fix things I want to be sure this is what we want and we should document the path conventions.

weshaggard: @RickWinter forwarded me the email with the context so I'll make adjustments to some of the automation to support this new convention.

tadelesh: Thanks, Wes. I'm about to send the mail to ask for your help with release detection script change.

Issue #3080, opened 2018-10-19
[Go] Provide sdk api for using resource ids


praveenvaltix:
### Service Package Request

Hi,

I have a question on how resources are linked and exposed in Azure through the REST APIs/Go SDK. From my understanding every resource has a name and a resource ID. What I am noticing is that the Get APIs take the resource name while the references within a resource may refer to other resources through an ID. 

As an example, the compute.virtualmachines.Get() returns a VirtualMachine which has references to network interfaces and the network references to IDs. But there is no way to get a resource based on the ID. The only way to get the resources than is to scrape and save all the resource names/IDs and use APIs to retrieve.

Is my understanding correct or wrong?

Thanks,
Praveen.

praveenvaltix:
I got some ideas after looking at https://gist.github.com/vladbarosan/fb2528754cbd97df51ca11fe7be27d2f

This is written by @vladbarosan

praveenvaltix:
Thanks @vladbarosan for confirming.

I guess my follow up question is, to reduce boiler plate code for us developers, there should ideally be a Get APIs which take the resource id instead of the name. 

See below for e.g -

		resource, err := ParseResourceID(nicID)
		if err != nil {
			log.Printf("Error parsing resource id %v", err)
			return nil, err
		}

		// get nic info
		nicInfo, err := nicClient.Get(
			context.TODO(),
			p.ResourceGroup,
			resource.resourceName,
			"",
		)
		if err != nil {
			log.Printf("Error getting nic info %v", err)
			return nil, err
		}

If instead nicclient.Get() took an ID, we could have avoided the "ParseResource". 

Thanks for confirming though.

vladbarosan:
Thanks @praveenvaltix for the feedback. We are considering various improvements for our next iteration and this is one of them.

tombuildsstuff:
@vladbarosan given that Resource ID's _shouldn't_ be case-sensitive (but are in a bunch of places) - I feel like it'd be more helpful to expose a function which parses and returns the items needed for the Get call as a struct - rather than just accepting an ID (which, could fail given the casing is incorrect)? FWIW we have support for this within Terraform as [the `parseAzureResourceID` function](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/helpers/azure/resourceid.go#L14-L103) - but it's not ideal (given not all ID's contain a Subscription ID, for example)

ArcturusZhang:
Staled issue, closing

[...]

PR #15654, opened 2021-09-28
[ARMCore] resource identifiers

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/3080

This PR introduces the `ResourceID` and `ResourceType` which provides utilities of manipulating resource IDs.
Code logic translated from C# SDK: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/ResourceIdentifier.cs and https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/ResourceType.cs


sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+	resourceGroupName string
+	location          string
+	resourceType      ResourceType
+	name              string
+	isChild           bool
+
+	stringValue *string
+}
+
+func (id ResourceIdentifier) Parent() *ResourceIdentifier {
```
jhendrixMSFT, line 51: I understand this was translated from C# but we should remove the getters and just export the necessary fields.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+	parent            *ResourceIdentifier
+	subscriptionId    string
+	provider          string
+	resourceGroupName string
+	location          string
+	resourceType      ResourceType
+	name              string
+	isChild           bool
+
+	stringValue *string
```
jhendrixMSFT, line 48: Does this really need to be a pointer or is the empty string good enough.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+const (
+	providersKey             = "providers"
+	subscriptionsKey         = "subscriptions"
+	resourceGroupsLowerKey   = "resourcegroups"
+	locationsKey             = "locations"
+	builtInResourceNamespace = "Microsoft.Resources"
+)
+
+var (
+	RootResourceIdentifier = &ResourceIdentifier{
```
jhendrixMSFT, line 25: Do these really need to be exported?

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func newResourceIdentifierWithProvider(parent *ResourceIdentifier, providerNamespace, resourceTypeName, resourceName string) *ResourceIdentifier {
+	id := &ResourceIdentifier{}
+	id.init(parent, NewResourceType(providerNamespace, resourceTypeName), resourceName, false)
+	return id
+}
+
+func chooseResourceType(resourceTypeName string, parent *ResourceIdentifier) ResourceType {
+	switch strings.ToLower(resourceTypeName) {
```
jhendrixMSFT, line 98: You want [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold) instead of all these calls to `strings.ToLower`

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+		id.parent = parent
+	}
+	id.isChild = isChild
+	id.resourceType = resourceType
+	id.name = name
+}
+
+// ParseResourceIdentifier parses a string to an instance of ResourceIdentifier
+func ParseResourceIdentifier(id string) (*ResourceIdentifier, error) {
+	if len(id) == 0 {
```
jhendrixMSFT, line 144: Seems like we could use a regex to validate the input and pull out the interesting parts.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+		if len(s) == 0 {
+			continue
+		}
+		r = append(r, s)
+	}
+
+	return r
+}
+
+type ResourceType struct {
```
jhendrixMSFT, line 248: Missing doc comment.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,315 @@
...
+func NewResourceType(providerNamespace, name string) ResourceType {
+	return ResourceType{
+		namespace:   providerNamespace,
+		t:           name,
+		types:       splitStringAndOmitEmpty(name, "/"),
+		stringValue: fmt.Sprintf("%s/%s", providerNamespace, name),
+	}
+}
+
+func NewResourceTypeFromParent(parent ResourceType, childType string) ResourceType {
```
jhendrixMSFT, line 281: I wonder if this should be a method on `ResourceType`, did you try it to see how it works?

jhendrixMSFT (changes requested):
It would also help to have an examples_test.go file that demonstrates how to use this.

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: Make sense. And serialized resource identifier can never be empty, therefore I changed this to `string`

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: I was thinking that we need to make this struct read-only and can only be initialized using the `ParseResourceIdentifier` function to make sure all the fields in this struct will be handled by our own logic

jhendrixMSFT: This starts looking like C# IMO.

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: Yeah - this is designed to be the root level parent of the resource identifiers instead of `nil`.
And this is also a resource identifier that you cannot get from `ParseResourceIdentifier`

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: docs have been added

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: To be a method will also work perfectly. Let me change this

ArcturusZhang: I have changed this one to a method `AppendChild`, please take a look

sdk/azcore/arm/resource_identifier.go
jhendrixMSFT: I see your point, we can leave as-is for now.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,321 @@
...
+	isChild           bool
+
+	stringValue string
+}
+
+func (id ResourceIdentifier) Parent() *ResourceIdentifier {
+	return id.parent
+}
+
+func (id ResourceIdentifier) SubscriptionId() string {
```
jhendrixMSFT, line 58: Since the `Parse()` method returns `ResourceIdentifier` by reference, you should make the receivers also by reference.

ArcturusZhang: I have made the receivers of ResourceIdentifier and ResourceType pointers

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,238 @@
...
+	RootResourceIdentifier = &ResourceIdentifier{
+		parent:       nil,
+		resourceType: TenantResourceType,
+		name:         "",
+	}
+)
+
+// ResourceIdentifier represents a resource ID such as `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRg`
+// Don't create this type directly, use ParseResourceID() instead
+type ResourceIdentifier struct {
```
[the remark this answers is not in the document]
ArcturusZhang: Done

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,238 @@
...
+	parent            *ResourceIdentifier
```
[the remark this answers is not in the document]
ArcturusZhang: Nope, I removed the pointer here

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,238 @@
...
+	} else {
+		id.parent = parent
+	}
+	id.isChild = isChild
+	id.resourceType = resourceType
+	id.name = name
+}
+
+// ParseResourceIdentifier parses a string to an instance of ResourceIdentifier
+func ParseResourceIdentifier(id string) (*ResourceIdentifier, error) {
```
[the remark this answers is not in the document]
ArcturusZhang: Done

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+
+// IsParentOf returns true when the receiver is the parent resource type of the child.
+func (t *ResourceType) IsParentOf(child *ResourceType) bool {
+	if child == nil {
+		return false
+	}
+	if !strings.EqualFold(t.Namespace, child.Namespace) {
+		return false
+	}
+	var types = splitStringAndOmitEmpty(t.Type, "/")
```
jhendrixMSFT, line 50: Is this splitting really necessary, couldn't you just do `strings.EqualFold(t.Type, child.Type)`?

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+func (t *ResourceType) LastType() string {
+	return t.Types[len(t.Types)-1]
+}
+
+func (t *ResourceType) String() string {
+	return t.stringValue
+}
+
+// IsParentOf returns true when the receiver is the parent resource type of the child.
+func (t *ResourceType) IsParentOf(child *ResourceType) bool {
```
jhendrixMSFT, line 43: I don't see a test for this, and I don't see this being used anywhere.  Can it be removed?  If not, please add a test.

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+// ResourceType represents an Azure resource type, e.g. "Microsoft.Network/virtualNetworks/subnets"
+type ResourceType struct {
+	Namespace string
+	Type      string
+	Types     []string
+
+	stringValue string
+}
+
+func (t *ResourceType) LastType() string {
```
jhendrixMSFT, line 34: I think this method can be removed now yes?  If you need a helper to avoid constantly writing `t.Types[len(t.Types)-1]` you can make an unexported func.

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ParseResourceType parses the ResourceType from a resource type string (e.g. Microsoft.Network/virtualNetworks/subsets)
+// or a resource identifier string (e.g. /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRg/providers/Microsoft.Network/virtualNetworks/vnet/subnets/mySubnet)
+func ParseResourceType(resourceIdOrType string) (ResourceType, error) {
+	// split the path into segments
+	parts := splitStringAndOmitEmpty(resourceIdOrType, "/")
+
+	// There must be at least a namespace and type name
+	if len(parts) < 1 {
+		return ResourceType{}, fmt.Errorf("invalid resource id or type: %s", resourceIdOrType)
```
jhendrixMSFT, line 88: "ID"

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+		return NewResourceType(builtInResourceNamespace, parts[0]), nil
+	} else if strings.Contains(parts[0], ".") {
+		// Handle resource types (Microsoft.Compute/virtualMachines, Microsoft.Network/virtualNetworks/subnets)
+		// it is a full type name
+		return NewResourceType(parts[0], strings.Join(parts[1:], "/")), nil
+	} else {
+		// Check if ResourceID
+		id, err := ParseResourceID(resourceIdOrType)
+		if err != nil {
+			return ResourceType{}, fmt.Errorf("invalid resource id: %s", resourceIdOrType)
```
jhendrixMSFT, line 103: "ID"

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,107 @@
...
+func NewResourceType(providerNamespace, typeName string) ResourceType {
+	return ResourceType{
+		Namespace:   providerNamespace,
+		Type:        typeName,
+		Types:       splitStringAndOmitEmpty(typeName, "/"),
+		stringValue: fmt.Sprintf("%s/%s", providerNamespace, typeName),
+	}
+}
+
+// AppendChild initiate an instance using the receiver ResourceType as parent and append childType to it.
```
jhendrixMSFT, line 75: AppendChild creates a ResourceType instance using the receiver as the parent with childType appended to it.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,210 @@
...
+func (id ResourceID) String() string {
+	if len(id.stringValue) > 0 {
+		return id.stringValue
+	}
+
+	id.stringValue = id.resourceString()
+	return id.stringValue
+}
+
+func (id ResourceID) resourceString() string {
```
jhendrixMSFT, line 180: Since this is only used from `String()`  you can merge this into that.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,210 @@
...
+		ResourceType: TenantResourceType,
+		Name:         "",
+	}
+)
+
+// ResourceID represents a resource ID such as `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRg`
+// Don't create this type directly, use ParseResourceID() instead
+type ResourceID struct {
+	Parent            *ResourceID
+	SubscriptionId    string
```
jhendrixMSFT, line 35: SubscriptionID

sdk/azcore/arm/resource_identifier.go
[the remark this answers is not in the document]
jhendrixMSFT: What is the error, you can't just dereference it?  Is it because parent can be `nil`?

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,210 @@
...
+	ResourceGroupName string
+	Location          string
+	ResourceType      ResourceType
+	Name              string
+
+	isChild     bool
+	stringValue string
+}
+
+func newResourceID(parent *ResourceID, resourceTypeName string, resourceName string) *ResourceID {
```
jhendrixMSFT, line 46: A minor thing, I try to keep the unexported code after all the exported code to make the public stuff easier to find.

sdk/azcore/arm/resource_type.go
ArcturusZhang: This is testing whether a type is parent of the other, therefore we do not test if they are equal. But we could test one to be prefix of the other if we do not use splitting.

sdk/azcore/arm/resource_type.go
ArcturusZhang: Tests added

sdk/azcore/arm/resource_type.go
ArcturusZhang: It turns out that we already split them while constructing them. Therefore I removed the split here, and directly use `child.Types`

sdk/azcore/arm/resource_type.go
ArcturusZhang: Removed

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: I will move these unexported helpers at the end of this file

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: Done

jhendrixMSFT:
I cleaned up the formatting and tweaked a few comments.  Please be sure to pull before making further commits.

sdk/azcore/arm/resource_type.go
```diff
@@ -0,0 +1,115 @@
...
+// ParseResourceType parses the ResourceType from a resource type string (e.g. Microsoft.Network/virtualNetworks/subsets)
+// or a resource identifier string.
+// e.g. /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRg/providers/Microsoft.Network/virtualNetworks/vnet/subnets/mySubnet)
+func ParseResourceType(resourceIDOrType string) (ResourceType, error) {
+	// split the path into segments
+	parts := splitStringAndOmitEmpty(resourceIDOrType, "/")
+
+	// There must be at least a namespace and type name
+	if len(parts) < 1 {
+		return ResourceType{}, fmt.Errorf("invalid resource ID or type: %s", resourceIDOrType)
```
jhendrixMSFT, line 92: I'm unclear how this code-path is hit.  Passing "foo" to this function works, i.e. `parts == 1`.  Can you please clarify?

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,207 @@
...
+	id.ResourceType = resourceType
+	id.Name = name
+}
+
+func appendNext(parent *ResourceID, parts []string, id string) (*ResourceID, error) {
+	if len(parts) == 0 {
+		return parent, nil
+	}
+
+	if len(parts) == 1 {
```
jhendrixMSFT, line 163: There's no code coverage for this branch, can you please ensure it's completely covered?  There's a lot of logic in here.

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return appendNext(RootResourceID, parts, id)
+}
+
+func (id *ResourceID) String() string {
+	if len(id.stringValue) > 0 {
+		return id.stringValue
+	}
+
+	if id.Parent == nil {
```
jhendrixMSFT, line 74: Is this correct?

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: I will add some test case for this.

ArcturusZhang: Added

sdk/azcore/arm/resource_type.go
ArcturusZhang: If you pass `/` or empty string to it, it will throw error here

sdk/azcore/arm/resource_identifier.go
ArcturusZhang: Yes, every string concatenation below has the prefix of `/`, therefore we do not need return anything if we hit our end when parent is nil.

ArcturusZhang:
Hey @jhendrixMSFT I was told we still have some edge cases when we hit the case that we might `subscriptions` in other cases - such like `/subscriptions/17fecd63-33d8-4e43-ac6f-0aafa111b38d/resourceGroups/myRg/providers/Microsoft.ApiManagement/service/myservicename/subscriptions/mysubscription` (this is a real ID from apimanagement.
I will continue to add some more test cases like this and will let you know when I am done

ArcturusZhang:
Hi @jhendrixMSFT I just fixed the `subscriptions` related issue and added a new test case. Please take a look again. Thank you!

sdk/azcore/arm/resource_identifier.go
```diff
@@ -0,0 +1,206 @@
...
+		Parent:       nil,
+		ResourceType: TenantResourceType,
+		Name:         "",
+	}
+)
+
+// ResourceID represents a resource ID such as `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myRg`.
+// Don't create this type directly, use ParseResourceID instead.
+type ResourceID struct {
+	Parent            *ResourceID
```
jhendrixMSFT, line 34: Can you please add doc comments to these fields as well?  Similar to what you did for `ResourceType`

ArcturusZhang:
Hey @jhendrixMSFT could you please take a look at my recent changes?

jhendrixMSFT (approved):

PR #15660, opened 2021-09-28
Remove azidentity.AADAuthenticationFailedError

chlowell:
This removes `AADAuthenticationFailedError` so AAD response structure isn't part of our API, and because we don't need it; `AuthenticationFailedError` can encompass errors arising specifically from AAD auth failures. In practical terms, this change makes getting low-level details such as a correlation ID a bit more work. I think that's okay because this change also makes AAD's detailed error description part of `AuthenticationFailedError.Error()`, so the most important information is the most visible, and exposes the underlying `*http.Response` via `AuthenticationFailedError.RawResponse()`, so no information is lost.

### DIFF
--- a/sdk/azidentity/device_code_credential.go
+++ b/sdk/azidentity/device_code_credential.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
@@ -158,7 +159,8 @@ func (c *DeviceCodeCredential) GetToken(ctx context.Context, opts policy.TokenRe
 		}
 		// if there is an error, check for an AADAuthenticationFailedError in order to check the status for token retrieval
 		// if the error is not an AADAuthenticationFailedError, then fail here since something unexpected occurred
-		if authRespErr := (*AADAuthenticationFailedError)(nil); errors.As(err, &authRespErr) && authRespErr.Message == "authorization_pending" {
+		var authFailed *AuthenticationFailedError
+		if errors.As(err, &authFailed) && strings.Contains(authFailed.msg, "authorization_pending") {
 			// wait for the interval specified from the initial device code endpoint and then poll for the token again
 			time.Sleep(time.Duration(dc.Interval) * time.Second)
 		} else {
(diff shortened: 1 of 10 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -173,7 +173,7 @@ func (c *managedIdentityClient) createAuthRequest(ctx context.Context, clientID
 		// need to perform preliminary request to retreive the secret key challenge provided by the HIMDS service
 		key, err := c.getAzureArcSecretKey(ctx, scopes)
 		if err != nil {
-			return nil, &AuthenticationFailedError{inner: err, msg: "Failed to retreive secret key from the identity endpoint."}
+			return nil, &AuthenticationFailedError{msg: "Failed to retreive secret key from the identity endpoint."}
```
jhendrixMSFT, line 176: Nit: Idiom for error messages is to not start with a capital letter and not end with punctuation (the linter will flag this, but likely can't see through this case).

sdk/azidentity/device_code_credential.go
```diff
@@ -158,7 +159,8 @@ func (c *DeviceCodeCredential) GetToken(ctx context.Context, opts policy.TokenRe
 		}
 		// if there is an error, check for an AADAuthenticationFailedError in order to check the status for token retrieval
 		// if the error is not an AADAuthenticationFailedError, then fail here since something unexpected occurred
-		if authRespErr := (*AADAuthenticationFailedError)(nil); errors.As(err, &authRespErr) && authRespErr.Message == "authorization_pending" {
+		var authFailed *AuthenticationFailedError
+		if errors.As(err, &authFailed) && strings.Contains(authFailed.msg, "authorization_pending") {
```
jhendrixMSFT, line 163: This makes me a bit nervous, but given the change, I don't have a better solution.

jhendrixMSFT (approved):

sdk/azidentity/device_code_credential.go
jhendrixMSFT: What if we kept the original error type but just didn't export it?  That way we can use these fields internally.

chlowell: The solution is: adopt MSAL 😉

jhendrixMSFT: Oh yes of course.

PR #15669, opened 2021-09-29
Cosmos DB - Enabling emulator tests

ealsur:
Add Emulator tests into the CI to execute skipped tests.

The tests can only run on Windows and were being skipped if the environment variable was not set.

This PR adds a new step that runs only on Windows and sets the environment variable to enable running the emulator tests.

### DIFF
--- a/sdk/data/azcosmos/ci.yml
+++ b/sdk/data/azcosmos/ci.yml
@@ -36,7 +36,7 @@ stages:
         Windows_Go115:
           pool.name: azsdk-pool-mms-win-2019-general
           image.name: MMS2019
-          go.version: '1.16'
+          go.version: '1.16.7'
         Windows_Go114:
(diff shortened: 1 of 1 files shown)


benbp (approved):

sdk/data/azcosmos/ci.yml
```diff
@@ -24,4 +24,57 @@ stages:
...
+- stage: Emulator
+  displayName: 'Cosmos Emulator'
+  jobs:
+  - job: DownloadAndRunCosmosEmulator
+    displayName: Download and run Cosmos Emulator
+
+    strategy:
+      matrix:
+        Windows_Go116:
+          pool.name: azsdk-pool-mms-win-2019-general
```
weshaggard, line 37: @benbp at some point we should normalize the name of the pool and image in the go repo. We have common templates and scripts that expect this to be named OSVmImage for example.

PR #15670, opened 2021-09-29
Cosmos DB Get Started Example for Go

jay-most:
Cosmos DB Get Started example for Go. This example shows you how to get started using the Azure Cosmos DB SDK for Go. NewCosmosClient creates a new instance of CosmosClient with the specified values. It uses the default pipeline configuration.

Examples include:

1. Creating a database
2. Creating a container
3. Update container properties
4. Item CRUD
5. Optimistic Concurrency
6. Etag PreConditionFail

- [x ] The purpose of this PR is explained in this or a referenced issue.
- [ x] The PR does not update generated files.
- [ x] Tests are included and/or updated for code changes.


sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
```
ealsur, line 4: Based on other examples: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/example_test.go, https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/example_test.go#L4, name the file `example_test.go` and change the package to `azcosmos_test`

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// All operations for Go operate on a context.Context, allowing you to control cancellation/timeout.
+	ctx := context.Background()
+
+	// ===== 1. Creating a database =====
+
+	databaseName := CosmosDatabaseProperties{Id: "databaseName"}
+	database, err := client.CreateDatabase(ctx, databaseName, nil, nil)
+	log.Panic(err)
```
ealsur, line 34: Question: Does the `log.Panic` do nothing if the `err` is `nil`?

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// This example showcases several common operations to help you get started, such as:
+	// ===== 2. Creating a container =====
+
+	properties := CosmosContainerProperties{
+		Id: "aContainer",
+		PartitionKeyDefinition: PartitionKeyDefinition{
+			Paths: []string{"/id"},
+		},
+		IndexingPolicy: &IndexingPolicy{
```
ealsur, line 44: Let's start with a simple case of not defining the IndexingPolicy (should use the service default)

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	database, err := client.CreateDatabase(ctx, databaseName, nil, nil)
+	log.Panic(err)
+
+	// This example showcases several common operations to help you get started, such as:
+	// ===== 2. Creating a container =====
+
+	properties := CosmosContainerProperties{
+		Id: "aContainer",
+		PartitionKeyDefinition: PartitionKeyDefinition{
+			Paths: []string{"/id"},
```
ealsur, line 42: Using the `/id` as the partition key has brought some confusion on documentation, can we use something else, like `/myPartitionKey`?

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Read the manual throughput property
+	throughputResponse, _ := container.ReadThroughput(ctx, nil)
+	_, _ = throughputResponse.ThroughputProperties.ManualThroughput()
+
+	// Replace manual throughput property
+	newScale := NewManualThroughputProperties(500)
+	_, _ = container.ReplaceThroughput(ctx, *newScale, nil)
+
+	// Container autoscale throughput
```
ealsur, line 90:
```suggestion
	// Migrate from manual throughput to autoscale
```

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	// Items in an Azure Cosmos container are uniquely identified by their id and partition key value.
+	pk, _ := NewPartitionKey("1")
+
+	item := map[string]string{
+		"id":    "1",
+		"value": "2",
+	}
+
+	// Create item.
+	itemResponse, _ := container.CreateItem(ctx, pk, item, nil)
```
ealsur, line 110: How is this working if you deleted the container in line 97?

ealsur: Add error checking

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Read item.
+	itemResponse, _ = container.ReadItem(ctx, pk, "1", nil)
```
ealsur, line 113: add error checking

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	resp, _ = container.Delete(ctx, nil)
+
+	// ===== 3. Item CRUD =====
+
+	// Items in an Azure Cosmos container are uniquely identified by their id and partition key value.
+	pk, _ := NewPartitionKey("1")
+
+	item := map[string]string{
+		"id":    "1",
+		"value": "2",
```
ealsur, line 106: If you do change the partition key definition, this map would need to contain the partition key property also

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Item response body
+	var itemResponseBody map[string]interface{}
+	err = json.Unmarshal(itemResponse.Value, &itemResponseBody)
+	log.Panic(err)
+
+	// ===== 4. Optimistic Concurrency =====
+
+	// Azure Cosmos DB supports optimistic concurrency control to prevent lost updates or deletes and detection of conflicting operations.
+	itemResponse, _ = container.UpsertItem(ctx, pk, item, &CosmosItemRequestOptions{EnableContentResponseOnWrite: true})
```
ealsur, line 123: `&CosmosItemRequestOptions{EnableContentResponseOnWrite: true}` is not needed

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Read item.
+	itemResponse, _ = container.ReadItem(ctx, pk, "1", nil)
+
+	// Item response body
+	var itemResponseBody map[string]interface{}
+	err = json.Unmarshal(itemResponse.Value, &itemResponseBody)
+	log.Panic(err)
+
+	// ===== 4. Optimistic Concurrency =====
```
ealsur, line 120: This is not Optimistic Concurrency, what you are doing in this section is Session consistency guarantee

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Azure Cosmos DB supports optimistic concurrency control to prevent lost updates or deletes and detection of conflicting operations.
+	itemResponse, _ = container.UpsertItem(ctx, pk, item, &CosmosItemRequestOptions{EnableContentResponseOnWrite: true})
+
+	itemSessionToken := itemResponse.SessionToken()
+
+	itemResponse, _ = container.ReadItem(ctx, pk, "1", &CosmosItemRequestOptions{SessionToken: itemSessionToken})
+
+	// ===== 5. Etag PreConditionFail =====
```
ealsur, line 129: This is Optimistic Concurrency

sdk/data/azcosmos/cosmos_example_test.go
```diff
@@ -0,0 +1,136 @@
...
+	itemEtag := itemResponse.ETag()
+
+	itemResponse, _ = container.ReadItem(ctx, pk, "1", &CosmosItemRequestOptions{IfMatchEtag: &itemEtag})
```
ealsur, line 132: The gist of Optimistic Concurrency is that you do:
1. ReadItem
2. Get Etag from read response
3. Modify the document value, change anything
4. Call ReplaceItem, passing IfMatchEtag
5. Check if the response was a failure and the error was StatusCode 412
6. If it was 412, repeat point 1, 2, 3, 4 until it succeeds

jay-most: I'll add

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
```
ealsur, line 4:
```suggestion
package azcosmos_test
```

ealsur: You will need to import ""github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos" and prefix all SDK apis with "azcosmos.{API}", for example "azcosmos.NewCosmosClient"

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// Replace container properties
+	resp, _ = container.Replace(ctx, updatedProperties, nil)
+
+	// Read the manual throughput property
+	throughputResponse, _ := container.ReadThroughput(ctx, nil)
+	_, _ = throughputResponse.ThroughputProperties.ManualThroughput()
+
+	// Replace manual throughput property
+	newScale := NewManualThroughputProperties(500)
+	_, _ = container.ReplaceThroughput(ctx, *newScale, nil)
```
ealsur, line 87: let's for the sake of correctness always check the err property, if people copy/paste from here, we rather have them do it the right thing

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// Migrate from manual throughput to autoscale
+	_, _ = throughputResponse.ThroughputProperties.AutoscaleMaxThroughput()
+
+	newScale = NewAutoscaleThroughputProperties(10000)
+	_, _ = container.ReplaceThroughput(ctx, *newScale, nil)
+
+	// ===== 4. Item CRUD =====
+
+	// Items in an Azure Cosmos container are uniquely identified by their id and partition key value.
+	// pk, _ := NewPartitionKey("myPartitionKey")
```
ealsur, line 98: This should be uncommented?

The value of the PK must be the value of the property, for example:

```
pk, err := NewPartitionKey("aPKValue")
if (err != nil) {
    log.Fatal(err)
}

item := map[string]string{
		"id":    "1",
		"value": "2",
                "myPartitionKey": "aPKValue"
}

itemResponse, err := container.CreateItem(ctx, pk, item, nil)
```

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// Create item.
+	itemResponse, err := container.CreateItem(ctx, nil, item, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Read item.
+	itemResponse, err = container.ReadItem(ctx, nil, "1", nil)
```
ealsur, line 112: missing the PK parameter on the read, reads need both id and PK values

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// Item response body
+	var itemResponseBody map[string]interface{}
+	err = json.Unmarshal(itemResponse.Value, &itemResponseBody)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// ===== 5. Session consistency =====
+
+	itemResponse, err = container.UpsertItem(ctx, nil, item, nil)
```
ealsur, line 126: Same here: missing the PK parameter. PK parameter is required.

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	itemEtag := itemResponse.ETag()
+	item = map[string]string{
+		"id":    "1",
+		"value": "newValue",
+	}
+
+	// Check the item response status code. If an error is imitted and the response code is 412 then retry operation.
+	for itemResponse.RawResponse.StatusCode == 412 {
```
ealsur, line 152: Based on this logic, the for will never execute. `itemResponse` is coming from the ReadItem with an HTTP 200 response

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// Azure Cosmos DB supports optimistic concurrency control to prevent lost updates or deletes and detection of conflicting operations.
+	itemResponse, err = container.ReadItem(ctx, nil, "1", nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	itemEtag := itemResponse.ETag()
+	item = map[string]string{
+		"id":    "1",
+		"value": "newValue",
```
ealsur, line 148: missing "myPartitionKey" property

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// ===== 1. Creating a database =====
+
+	databaseName := azcosmos.CosmosDatabaseProperties{Id: "databaseName"}
+	database, err := client.CreateDatabase(ctx, databaseName, nil, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// This example showcases several common operations to help you get started, such as:
```
ealsur, line 42: This can be move up before the `1.` right? Would be a bit more sense?

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Replace manual throughput property
+	newScale := azcosmos.NewManualThroughputProperties(500)
+	_, err = container.ReplaceThroughput(ctx, *newScale, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Migrate from manual throughput to autoscale
+	_, err = throughputResponse.ThroughputProperties.AutoscaleMaxThroughput()
```
ealsur, line 103: This is the other way around right? Read the AutoScaleMaxThroughput after the replace that changes to autoscale?

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Replace container properties
+	resp, err = container.Replace(ctx, updatedProperties, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Read the manual throughput property
+	throughputResponse, _ := container.ReadThroughput(ctx, nil)
+	_, err = throughputResponse.ThroughputProperties.ManualThroughput()
```
ealsur, line 90: Can we actually show the printing of the value for both this and the read of the AutoscaleMaxThroughput? Otherwise it's not really showing any usage of the value.

```
manualThroughput, err := throughputResponse.ThroughputProperties.ManualThroughput()
if err != nil {
    log.Fatal(err)
}
fmt.Printf("Container is provisioned with %v RU/s", manualThroughput)
```

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// Replace container properties
+	resp, err = container.Replace(ctx, updatedProperties, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Read the manual throughput property
+	throughputResponse, _ := container.ReadThroughput(ctx, nil)
```
ealsur, line 89: missing error check

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+	_, err = throughputResponse.ThroughputProperties.AutoscaleMaxThroughput()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	newScale = azcosmos.NewAutoscaleThroughputProperties(10000)
+	_, err = container.ReplaceThroughput(ctx, *newScale, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
```
ealsur, line 112: After the Replace, we should be doing another ReadThroughput, and then getting the `throughputResponse.ThroughputProperties.AutoscaleMaxThroughput`

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Item response body
+	var itemResponseBody map[string]interface{}
+	err = json.Unmarshal(itemResponse.Value, &itemResponseBody)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
ealsur, line 142: missing Delete?

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+	item = map[string]string{
+		"id":             "1",
+		"value":          "newValue",
+		"myPartitionKey": "newPartitionKey",
+	}
+
+	// Check the item response status code. If an error is imitted and the response code is 412 then retry operation.
+	err = retryOptimisticConcurrencyExample(numberRetry, 1000*time.Millisecond, func() (int, error) {
+		itemResponse, err = container.ReadItem(ctx, pk, "1", nil)
+		return itemResponse.RawResponse.StatusCode, err
```
ealsur, line 169: Any failure (like 412) is surfaced in the `err`, not in the itemResponse.

Basically you need to check if there is any err and if so, attempt to extract the statusCode to see if it's 412:

```
var httpErr azcore.HTTPResponse
if errors.As(err, &httpErr) && httpErr.RawResponse().StatusCode == 412 {
    // needs to be retried
}
```

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+	itemResponse, err = container.ReadItem(ctx, pk, "1", &azcosmos.CosmosItemRequestOptions{SessionToken: itemSessionToken})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// ===== 6. Optimistic Concurrency Etag PreConditionFail =====
+
+	// Azure Cosmos DB supports optimistic concurrency control to prevent lost updates or deletes and detection of conflicting operations.
+	numberRetry := 3
+	item = map[string]string{
```
ealsur, line 160: This variable is not being used in the below example

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,201 @@
...
+		"id":             "1",
+		"value":          "newValue",
+		"myPartitionKey": "newPartitionKey",
+	}
+
+	// Check the item response status code. If an error is imitted and the response code is 412 then retry operation.
+	err = retryOptimisticConcurrencyExample(numberRetry, 1000*time.Millisecond, func() (int, error) {
+		itemResponse, err = container.ReadItem(ctx, pk, "1", nil)
```
ealsur, line 168: After the read, you need to extract the ETag from the response, then use the Etag on a Replace and check for a replace (example in C# https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L676-L772, it does not have the loop, but the logic of Read+Replace is there).

```
itemResponse, err = container.ReadItem(ctx, pk, "1", nil)
if err != nil {
		log.Fatal(err)
	}

var itemResponseBody map[string]interface{}
	err = json.Unmarshal(itemResponse.Value, &itemResponseBody)
	if err != nil {
		log.Fatal(err)
	}

// Change some value
itemResponseBody["value"] = "newValue"

etag := itemResponse.ETag()

// Replace with Etag
itemResponse, err = container.ReplaceItem(ctx, pk, "1", itemResponseBody, &CosmosItemRequestOptions{IfMatchEtag: &etag})

then check if there is an err and its statuscode == 412
```

sdk/data/azcosmos/example_test.go
ealsur: the `err` variable is not being populated, it should be:
`throughputResponse, err := container.ReadThroughput(ctx, nil)`

sdk/data/azcosmos/example_test.go
```diff
@@ -0,0 +1,223 @@
...
+	// Replace manual throughput property
+
+	newScale := azcosmos.NewManualThroughputProperties(500)
+	_, err = container.ReplaceThroughput(ctx, *newScale, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Migrate from manual throughput to autoscale
+	autoscaleMaxThroughputResponse, err := throughputResponse.ThroughputProperties.AutoscaleMaxThroughput()
```
ealsur, line 112: This should happen after the next Replace. At this point the container is still provisioned with normal throughput (not autoscale).

PR #15677, opened 2021-09-30
Remove NewAuthenticationPolicy from credentials

chlowell:
Also the default bearer token policy, which now lives in azcore.


sdk/azidentity/go.mod
```diff
@@ -2,10 +2,14 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
 go 1.14
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../azcore
```
jhendrixMSFT, line 5: Presumably this is pending the new `azcore` release?

chlowell: Yes. The published `azcore.TokenCredential` embeds `azcore.Credential`, leading to this PR breaking our interface checks. Is there a better mechanism for working around this sort of thing?

jhendrixMSFT: Given it's temporal it's fine as it unblocks the PR.  The only risk is if there was some other breaking change in `azcore` then you could be broken.  You can specify a branch/commit if you don't want to run the risk of being broken, but I think that's unlikely.

jhendrixMSFT (approved):

Issue #15646, opened 2021-09-27
[amqp-common] Migrate Azure/azure-amqp-common-go into monorepo


richardpark-msft:
Service Bus (track 0 and track 2) as well as the GA'd Event Hubs package both depend on this external dependency.

This is purely an internally used package (unlike `go-amqp`) so moving this into the monorepo (azure-sdk-for-go) makes sense.

For SB, which is in the monorepo, this isn't a problem. However, we are probably stymied from doing this immediately because of Event Hubs, which is in an external repo and wouldn't be able to use internal packages in azure-sdk-for-go (yet).

PR #15681, opened 2021-10-01
[service-bus] Improving recovery and error handling

richardpark-msft:
Adding in a _lot_ more error handling code and retries in areas that were shown to be weak during Service Bus stress testing.

The biggest issues were mostly:
- Some errors weren't accounted for at all (for instance, credentials expiring seemed to fall through)
- Some errors are getting stringized (or aren't proper types) so there's some string matching code in the errors.go. Shouldn't be permanent, but we can recognize things fairly well with what's in there now.
- Some spots didn't have retries but really needed them. For safety, there are also some bugs we're working around where some frames appear to be arriving out of order in go-amqp. Things like that just result in a connection reset, which should recover it.

Also, mopped up some other problems:
- Fixed the naming of the `NewReceiver/NewProcessor` functions so they have ForQueue/ForSubscription in them
- Moved over a version of the stderrTracer from event-hubs. It's not _great_, but it's functional enough to work with the stress.go. It could use some work though, it's a bit hacked together.

### DIFF
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -19,13 +19,17 @@ import (
 // Client provides methods to create Sender, Receiver and Processor
 // instances to send and receive messages from Service Bus.
 type Client struct {
-	config      clientConfig
-	namespace   *internal.Namespace
+	config    clientConfig
+	namespace interface {
+		// used internally by `Client`
+		internal.NamespaceWithNewAMQPLinks
+		// for child clients
+		internal.NamespaceForAMQPLinks
+		internal.NamespaceForMgmtClient
+	}
 	linksMu     *sync.Mutex
 	linkCounter uint64
-	links       map[uint64]interface {
-		Close(ctx context.Context) error
-	}
+	links       map[uint64]internal.Closeable
 }
 
 type clientConfig struct {
@@ -84,9 +88,7 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
 	client := &Client{
 		linksMu: &sync.Mutex{},
 		config:  config,
-		links: map[uint64]interface {
-			Close(ctx context.Context) error
-		}{},
+		links:   map[uint64]internal.Closeable{},
 	}
 
 	for _, opt := range options {
@@ -194,9 +196,7 @@ func (client *Client) NewSender(queueOrTopic string) (*Sender, error) {
 func (client *Client) Close(ctx context.Context) error {
 	var lastError error
 
-	var links []interface {
-		Close(ctx context.Context) error
-	}
+	var links []internal.Closeable
 
 	client.linksMu.Lock()
 
@@ -219,9 +219,7 @@ func (client *Client) Close(ctx context.Context) error {
 	return nil
 }
 
-func (client *Client) addCloseable(id uint64, closeable interface {
-	Close(ctx context.Context) error
-}) {
+func (client *Client) addCloseable(id uint64, closeable internal.Closeable) {
 	client.linksMu.Lock()
 	client.links[id] = closeable
 	client.linksMu.Unlock()
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -118,7 +118,7 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
 
 // NewProcessor creates a Processor for a queue.
 func (client *Client) NewProcessorForQueue(queue string, options ...ProcessorOption) (*Processor, error) {
-	options = append(options, ProcessorWithQueue(queue))
+	options = append(options, processorWithQueue(queue))
 	id, cleanupOnClose := client.getCleanupForCloseable()
 
 	processor, err := newProcessor(client.namespace, cleanupOnClose, options...)
@@ -133,7 +133,7 @@ func (client *Client) NewProcessorForQueue(queue string, options ...ProcessorOpt
 
 // NewProcessor creates a Processor for a subscription.
 func (client *Client) NewProcessorForSubscription(topic string, subscription string, options ...ProcessorOption) (*Processor, error) {
-	options = append(options, ProcessorWithSubscription(topic, subscription))
+	options = append(options, processorWithSubscription(topic, subscription))
 	id, cleanupOnClose := client.getCleanupForCloseable()
 
 	processor, err := newProcessor(client.namespace, cleanupOnClose, options...)
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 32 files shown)


sdk/messaging/azservicebus/client.go
```diff
@@ -111,72 +116,123 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
...
-
 	if lastError != nil {
 		return fmt.Errorf("errors while closing links: %w", lastError)
 	}
 	return nil
 }
+
+func (client *Client) addCloseable(id uint64, closeable interface {
+	Close(ctx context.Context) error
+}) {
```
seankane-msft, line 224: nit: You use the `closeable` interface you have defined here in other spots. Why not define this interface as a `closeable` to reduce code?

sdk/messaging/azservicebus/client.go
```diff
@@ -111,72 +116,123 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
 	return client, err
 }
 
-// NewProcessor creates a Processor, which allows you to receive messages from ServiceBus.
-func (client *Client) NewProcessor(options ...ProcessorOption) (*Processor, error) {
-	processor, err := newProcessor(client.namespace, options...)
+// NewProcessor creates a Processor for a queue.
+func (client *Client) NewProcessorForQueue(queue string, options ...ProcessorOption) (*Processor, error) {
+	options = append(options, ProcessorWithQueue(queue))
```
jhendrixMSFT, line 121: Do we still need this option even though queue is a param now?

sdk/messaging/azservicebus/client.go
```diff
@@ -111,72 +116,123 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
...
 func (client *Client) Close(ctx context.Context) error {
 	var lastError error
 
+	var links []interface {
+		Close(ctx context.Context) error
+	}
+
 	client.linksMu.Lock()
 
 	for _, link := range client.links {
```
jhendrixMSFT, line 203: This can be simplified to `copy(links, client.links)`.  End result is the same.

sdk/messaging/azservicebus/client.go
```diff
@@ -111,72 +116,123 @@ func newClientImpl(config clientConfig, options ...ClientOption) (*Client, error
...
+func (client *Client) addCloseable(id uint64, closeable interface {
+	Close(ctx context.Context) error
+}) {
+	client.linksMu.Lock()
+	client.links[id] = closeable
+	client.linksMu.Unlock()
+}
+
+func (client *Client) getCleanupForCloseable() (uint64, func()) {
+	id := atomic.AddUint64(&client.linkCounter, 1)
```
jhendrixMSFT, line 231: I don't see `linkCounter` being used elsewhere.

sdk/messaging/azservicebus/internal/amqpInterfaces.go
```diff
@@ -5,7 +5,9 @@ package internal
 
 import (
 	"context"
+	"time"
 
+	"github.com/Azure/azure-amqp-common-go/v3/rpc"
```
jhendrixMSFT, line 10: This will go away eventually right?

sdk/messaging/azservicebus/stress/stress.go
```diff
@@ -24,9 +26,15 @@ import (
 // | summarize Sum=sum(valueCount) by bin(timestamp, 30s), name
 // | render timechart
 
-var messagesSent int32
-var messagesReceived int32
-var exceptions int32
+type Stats struct {
```
jhendrixMSFT, line 29: This is all currently part of public surface area, is that intentional?

richardpark-msft: No, unexported

sdk/messaging/azservicebus/client.go
richardpark-msft: In this case `client.links` is a map. Am I missing a way to make that work?

sdk/messaging/azservicebus/client.go
richardpark-msft: I still have the option because, internally, it's still the easiest way for me to deal with it without having to expand to two functions everywhere. Might refactor later.

I did unexport it though, that was unintentional.

sdk/messaging/azservicebus/internal/amqpInterfaces.go
richardpark-msft: Issue tracking this is here: https://github.com/Azure/azure-sdk-for-go/issues/15646

Is there urgency around this one though? I originally thought we could do this at any time.

sdk/messaging/azservicebus/client.go
richardpark-msft: Yes, it's just used to provide a unique ID for inserting into our `client.links` map. I build a closure with the returned ID so we basically only deal with it through return values after this.

sdk/messaging/azservicebus/client.go
richardpark-msft: Done.

sdk/messaging/azservicebus/client.go
jhendrixMSFT: Ah I missed that.

sdk/messaging/azservicebus/internal/amqpInterfaces.go
jhendrixMSFT: No just making sure it will happen.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -24,70 +39,79 @@ func (s *messageSettler) useManagementLink(m *ReceivedMessage, linkRevision uint
...
-	}
+	for retrier.Try(ctx) {
+		var receiver internal.AMQPReceiver
+		var mgmt internal.MgmtClient
+		var linkRevision uint64
+
+		_, receiver, mgmt, linkRevision, lastErr = s.links.Get(ctx)
+
+		if lastErr != nil {
+			_ = s.links.RecoverIfNeeded(ctx, lastErr)
```
jhendrixMSFT, line 58: If we always treat RecoverIfNeeded as no-fail perhaps it would be better to have it not return an error (I know that feels weird).

richardpark-msft: Yeah, I was thinking that too. I have post-beta-1 plans where I do want to make usage of it more, but not yet in this one.

sdk/messaging/azservicebus/messageBatch.go
```diff
@@ -34,29 +34,18 @@ func newMessageBatch(maxBytes int) *MessageBatch {
...
-
-func (e errMessageTooLarge) NonRetriable()    {}
-func (e errMessageTooLarge) MessageTooLarge() {}
-
 // Add adds a message to the batch if the message will not exceed the max size of the batch
-// If the message is too large, an error of type 'ErrMessageTooLarge' will be returned.
-func (mb *MessageBatch) Add(m *Message) error {
+// This function will return:
+// (true, nil) if the message was added.
+// (false, nil) if the message was too large to fit into the batch.
```
jhendrixMSFT, line 40: If it's just that the batch is full I would reword this "if the batch is full". The current wording makes it read like if the message were smaller it would fit.

jhendrixMSFT (approved):

PR #15690, opened 2021-10-02
[service-bus] Additional recovery code

richardpark-msft:
After doing some more stress testing it's clear that our timeouts are far too aggressive and should be lengthened. They were using _nanoseconds_ by default, which was very incorrect.

There were also some other fixes:
- We pass in a link revision when recovering to avoid over-recovering the link
- There was a spot in the processor where recovery needed to happen and it wasn't. That's been fixed as well.


richardpark-msft:
@jhendrixMSFT - stress testing this newer set of changes now. It's holding up well and provides a much better experience.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -118,22 +118,42 @@ func (links *amqpLinks) ManagementPath() string {
...
+			tab.StringAttribute("revisions", fmt.Sprintf("ours(%d), theirs(%d)", ourLinkRevision, *theirLinkRevision)),
+		)
+		return nil
+	}
+
 	links.mu.Lock()
 	defer links.mu.Unlock()
 
-	links.revision++
+	if theirLinkRevision != nil && ourLinkRevision > *theirLinkRevision {
```
jhendrixMSFT, line 148: This duplicated bit, is this just double-checked locking pattern?

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -118,22 +118,42 @@ func (links *amqpLinks) ManagementPath() string {
 
 // recoverLink will recycle all associated links (mgmt, receiver, sender and session)
 // and recreate them using the link.linkCreator function.
-func (links *amqpLinks) recoverLink(ctx context.Context) error {
+func (links *amqpLinks) recoverLink(ctx context.Context, theirLinkRevision *uint64) error {
```
jhendrixMSFT, line 121: This is a pointer to disambiguate zero from no revision?

jhendrixMSFT (approved):

sdk/messaging/azservicebus/internal/amqpLinks.go
richardpark-msft: Yeah, just to make it more obvious.

I actually do start the link revision from 1 to avoid this but I felt like it was just too subtle in the source code.

sdk/messaging/azservicebus/internal/amqpLinks.go
richardpark-msft: Yeah, check once in the cheap 'read lock' and then check again in the expensive 'write lock'.

PR #15700, opened 2021-10-05
Refactor aztables constructors

chlowell:
Basically removing `azcore.Credential` and adding constructors with more explicit credential parameter types. Allows the module to function with upcoming versions of `azcore` and `azidentity`.


sdk/data/aztables/go.mod
```diff
@@ -4,10 +4,12 @@ go 1.16
 
 replace github.com/Azure/azure-sdk-for-go/sdk/internal => ../../internal
```
jhendrixMSFT, line 5: Can this be removed now?

sdk/data/aztables/go.mod
```diff
@@ -4,10 +4,12 @@ go 1.16
...
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../../azcore
```
jhendrixMSFT, line 7: What is the upcoming azcore release waiting for?

PR #15703, opened 2021-10-05
Update CODEOWNERS

RickWinter:


.github/CODEOWNERS
```diff
@@ -28,7 +28,13 @@
 /sdk/storage/            @rickwinter @jhendrixMSFT
 
 # PRLabel: %Cosmos
-/sdk/data/azcosmos/        @ealsur @ausfeldt 
+/sdk/data/azcosmos/      @ealsur @ausfeldt 
+
+# PRLabel: %Tables
+/sdk/data/aztables/      @seankane-msft
+
+# PRLabel: %Tables
```
seankane-msft, line 36: I don't know what it should be, but this PRLabel probably should not be Tables

RickWinter, line 36:
```suggestion
# PRLabel: %Mgmt
```

PR #15705, opened 2021-10-05
Refactor azcosmos client constructor

chlowell:
The next version of azcore doesn't contain `azcore.Credential`, so client constructors must be explicit about the credential types they accept. Pretty straightforward here because there's only one credential type. I renamed the constructor NewCosmosClient -> NewClientWithSharedKey to align with upcoming guidelines for the SDK (the "good" name, NewClient, would go to a constructor taking `azcore.TokenCredential`, i.e. an Azure AD credential).

### DIFF
--- a/sdk/data/azcosmos/README.md
+++ b/sdk/data/azcosmos/README.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-This client library enables client applications to connect to Azure Cosmos via the SQL API. Azure Cosmos is a globally distributed, multi-model database service. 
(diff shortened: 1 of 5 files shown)


ealsur:
@chlowell The recommendation then is to have `NewClient` for Azure AD scenarios and `NewClientWithSharedKey` for the other?

chlowell:
Yes.

ealsur (approved):

ealsur:
@chlowell Can you update the README examples please? Then we can merge this in

ealsur (approved):

PR #15708, opened 2021-10-05
[service-bus] Separating code coverage changes out

richardpark-msft:
Talking with Joel and Ben - the track 1 CI pipeline is triggering for the changes I've got in my main PR. 

This is unintentional, and I believe the only reason it's running at all is because the config.json file changed. Testing that theory now.

### DIFF
--- a/eng/config.json
+++ b/eng/config.json
@@ -27,6 +27,10 @@
         {
             "Name": "azsecrets",
             "CoverageGoal": 0.76
+        },
+        {
+            "Name": "messaging",
+            "CoverageGoal": 0.10
         }
     ]
 }


jhendrixMSFT (approved):

jhendrixMSFT:
Ideally we should fix the track 1 CI to skip building all track 2 content.

richardpark-msft:
> Ideally we should fix the track 1 CI to skip building all track 2 content.

@benbp, is this similar to what you were suggesting to me earlier?

jhendrixMSFT:
On a side-note, is there a way we can avoid triggering track 1 CI when updating the code coverage config file?  It's specific to track 2 and the track 1 CI takes forever without adding any value.

PR #15713, opened 2021-10-05
adding license, adding step to check for license

seankane-msft:

### DIFF
--- a/sdk/keyvault/azsecrets/CHANGELOG.md
+++ b/sdk/keyvault/azsecrets/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Release History
 
+## 0.1.1 (2021-10-06)
+* Adds the MIT License for redistribution
+
 ## 0.1.0 (2021-10-05)
 * This is the initial release of the `azsecrets` library


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -43,6 +43,14 @@ steps:
     failOnStderr: true
     workingDirectory: 'sdk/${{parameters.ServiceDirectory}}'
 
+  - script: |
+      echo scanning copyright headers in $(pwd)
+      grep -L -r --include LICENSE -P "Copyright (\d{4}|\(c\)) Microsoft" . | tee >&2
```
weshaggard, line 48: Is this needed in every module? Is the license in the root of the repo enough?

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -43,6 +43,14 @@ steps:
...
+    displayName: 'License Check'
+    condition: succeededOrFailed()
+    failOnStderr: true
```
weshaggard, line 51: Will grep set the correct exit code?

eng/pipelines/templates/steps/analyze.yml
seankane-msft: Yes I believe every module needs to have a copy of the license.

eng/pipelines/templates/steps/analyze.yml
seankane-msft: I believe so, I used the same command as the `Copyright Header Check` step above and have seen this one fail as expected.

eng/pipelines/templates/steps/analyze.yml
seankane-msft: Hmmm, thinking about this again I think it only needs to be in the root of each module

weshaggard (approved):

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -43,6 +43,14 @@ steps:
     failOnStderr: true
     workingDirectory: 'sdk/${{parameters.ServiceDirectory}}'
 
+  - script: |
+      echo scanning copyright headers in $(pwd)
```
jhendrixMSFT, line 47: `echo ensuring $(pwd)/LICENSE file exists` ?

jhendrixMSFT (approved):

RickWinter (approved):

eng/pipelines/templates/steps/analyze.yml
benbp: Adding `./` breaks the grep command so it doesn't work, because the `--include` value is supposed to be a glob, which it then searches recursively for in `.`.

```
# Example with LICENSE.txt in net repo
⇉ ⇉ ⇉ grep -L -r --include LICENSE.txt -P "donotmatch" .  | tee >&2
./samples/CloudClipboard/CloudClipboard/wwwroot/lib/jquery-validation-unobtrusive/LICENSE.txt
./samples/CloudClipboard/CloudClipboard/wwwroot/lib/jquery/LICENSE.txt
./samples/AppSecretsConfig/wwwroot/lib/jquery-validation-unobtrusive/LICENSE.txt
./samples/AppSecretsConfig/wwwroot/lib/jquery/LICENSE.txt
./LICENSE.txt

# Should succeed but fails
⇉ ⇉ ⇉ grep -L -r --include ./LICENSE.txt -P "Copyright (\d{4}|\(c\)) Microsoft" .
⇉ ⇉ ⇉ echo $?
1
```

I think the reason this may be succeeding is because `pipefail` is not set, so piping the command to `tee` masks the exit code.

PR #15717, opened 2021-10-05
[azservicebus] Getting live tests working

### LINKED ISSUE #15714: [service-bus] Live testing
Live testing needs to be setup - mostly just translating the older .tf file that had the resources into a standard ARM template.

richardpark-msft:
Grabbed the test-resources.json from the JS SDK, made a small adjustment to load up the right credential type.

Fixes #15714


richardpark-msft:
I didn't use bicep @benbp, which might be a strike against me.

benbp (approved):

PR #15722, opened 2021-10-06
Refactor azblob constructors for azcore 0.20.0

chlowell:
The next version of azcore doesn't contain `azcore.Credential`, so client constructors must be explicit about the credential types they accept. The names here align with upcoming guidelines for the SDK: shortest name, `New*Client`, goes to `azcore.TokenCredential`; other constructors add a With* suffix for the credential type; constructors for anonymous authentication are "...WithNoCredential".


sdk/storage/azblob/go.mod
```diff
@@ -2,11 +2,16 @@ module github.com/Azure/azure-sdk-for-go/sdk/storage/azblob
 
 go 1.16
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../../azcore
```
jhendrixMSFT, line 5: Remove and replace with `v0.20.0`

sdk/storage/azblob/zz_generated_connection.go
```diff
@@ -51,7 +51,15 @@ func newConnection(endpoint string, cred azcore.Credential, options *connectionO
 	policies = append(policies, options.PerCallPolicies...)
 	policies = append(policies, runtime.NewRetryPolicy(&options.Retry))
 	policies = append(policies, options.PerRetryPolicies...)
-	policies = append(policies, cred.NewAuthenticationPolicy(runtime.AuthenticationOptions{TokenRequest: policy.TokenRequestOptions{}}))
+
+	switch c := cred.(type) {
```
jhendrixMSFT, line 55: Seems like a `default` clause it appropriate here, although that would mean this func would need to return the resultant error.

jhendrixMSFT (approved):

seankane-msft (approved):

adreed-msft (approved):
lgtm

sdk/storage/azblob/zc_append_blob_client.go
```diff
@@ -5,17 +5,40 @@ package azblob
...
 
 type AppendBlobClient struct {
 	BlobClient
 	client *appendBlobClient
 }
 
-func NewAppendBlobClient(blobURL string, cred azcore.Credential, options *ClientOptions) (AppendBlobClient, error) {
-	con := newConnection(blobURL, cred, options.getConnectionOptions())
+// NewAppendBlobClient creates an AppendBlobClient with the specified URL, Azure AD credential, and options.
+func NewAppendBlobClient(blobURL string, cred azcore.TokenCredential, options *ClientOptions) (AppendBlobClient, error) {
```
mohsha-msft, line 20: Can we have just one `NewAppendBlobClient` interface?
```golang
func NewAppendBlobClient(blobURL string, options *CreateAppendBlobClientOptions) {
   credType, creds, clientOptions, err := options.format()
  // based on cred type, we can create clients
}
```

Inside zm_append_blob_request_options.go,
```golang
type CreateAppendBlobClientOptions struct {
     credential
    clientOptions
}

@zezha-msft , what do you think?

mohsha-msft: Same for all the clients. We should have a uniform interface method instead of multiple methods. That was the whole reason to introduce option bags in parameters

zezha-msft: I don't love multiple constructors either but credential is a mandatory field, so I don't think it should fit inside the option bag.

zezha-msft (approved):
lgtm, could you please run the tests live just once to double check?

mohsha-msft (approved):

PR #15864, opened 2021-10-20
[azservicebus] Creating a channel-based receiving function

richardpark-msft:
In a comment thread @JeffreyRichter mentioned that the Processor is not necessarily idiomatic Go and that a channel might be a better fit. 

Looking at this more I had about 90% of that code already written so I repackaged it and exposed it as a user primitive instead. We lose one feature (autocomplete) but we can still add in auto-lock renewal and we put the customer in charge of how they want to handle the messages, which seems like a win.

I've also got some other sundry little fixes in there, but I think this should provide a good start for discussion and would be okay to merge in, provided the design makes sense to everyone.


richardpark-msft:
API view for this changeset:
https://apiview.dev/Assemblies/Review/8016d8f02db04f08a4a9db346cd16918

CC: @JeffreyRichter

PR #15958, opened 2021-10-27
UploadStreamToBlockBlob: only require body be io.Reader

serprex:
This function only takes an io.Reader in azure-storage-blob-go,
& wal-g abstracts over multiple upload mechanisms which all operate on io.Reader

### DIFF
--- a/sdk/storage/azblob/chunkwriting.go
+++ b/sdk/storage/azblob/chunkwriting.go
@@ -33,7 +33,7 @@ type blockWriter interface {
 // well, 4 MiB or 8 MiB, and auto-scale to as many goroutines within the memory limit. This gives a single dial to tweak and we can
 // choose a max value for the memory setting based on internal transfers within Azure (which will give us the maximum throughput model).
 // We can even provide a utility to dial this number in for customer networks to optimize their copies.
-func copyFromReader(ctx context.Context, from io.ReadSeekCloser, to blockWriter, o UploadStreamToBlockBlobOptions) (BlockBlobCommitBlockListResponse, error) {
+func copyFromReader(ctx context.Context, from io.Reader, to blockWriter, o UploadStreamToBlockBlobOptions) (BlockBlobCommitBlockListResponse, error) {
 	if err := o.defaults(); err != nil {
 		return BlockBlobCommitBlockListResponse{}, err
 	}
--- a/sdk/storage/azblob/highlevel.go
+++ b/sdk/storage/azblob/highlevel.go
@@ -598,7 +598,7 @@ func (u *UploadStreamToBlockBlobOptions) getCommitBlockListOptions() *CommitBloc
 
 // UploadStreamToBlockBlob copies the file held in io.Reader to the Blob at blockBlobClient.
 // A Context deadline or cancellation will cause this to error.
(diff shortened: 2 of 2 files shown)


sdk/storage/azblob/chunkwriting.go
```diff
@@ -33,7 +33,7 @@ type blockWriter interface {
 // well, 4 MiB or 8 MiB, and auto-scale to as many goroutines within the memory limit. This gives a single dial to tweak and we can
 // choose a max value for the memory setting based on internal transfers within Azure (which will give us the maximum throughput model).
 // We can even provide a utility to dial this number in for customer networks to optimize their copies.
-func copyFromReader(ctx context.Context, from io.ReadSeekCloser, to blockWriter, o UploadStreamToBlockBlobOptions) (BlockBlobCommitBlockListResponse, error) {
+func copyFromReader(ctx context.Context, from io.Reader, to blockWriter, o UploadStreamToBlockBlobOptions) (BlockBlobCommitBlockListResponse, error) {
```
mohsha-msft, line 36: We're moving to use io.ReadSeekCloser across all the SDKs. You can write simple struct like [here](https://github.com/Azure/azure-sdk-for-go/blob/dfdc424c554cfb6e45a53e259c5c215bb4e1d3a6/sdk/storage/azblob/internal/zc_shared.go#L27)

mohsha-msft (changes requested):
Track2 (New SDK) was introduced
1. To provide more stable SDKs to avoid (or minimize) breaking changes for the end-user.
2. To introduce big design changes and decouple components such as logging, authentication, etc.

We've kept the similar architecture and use case across all new SDKs. So, I'll have to decline to this PR. Can you tell me why it should be changed?

sdk/storage/azblob/chunkwriting.go
serprex: Right. I'd implemented that in my PR to workaround for now, but then if this interface does change to use these unused methods that'll introduce a bug rather than a compile time error. It also makes the interface more confusing to use because then I can't tell from the function signature whether the reader will be kept open or not

But alright, if there's some design ideology at play I'll have to accept it

jhendrixMSFT: For context, we require a seekable stream to facilitate our retry logic.  If the PUT request fails, we can rewind the stream to try again.

@mohsha-msft I'm confused in this case.  The API takes a `ReadSeekCloser` but doesn't actually require a `Seeker` or a `Closer`.  Is it used elsewhere (maybe via a type-assertion)?

zezha-msft: @serprex sorry, just to clarify, this is a private method right? How are you leveraging it exactly?

jhendrixMSFT: It's called from `BlockBlobClient.UploadStreamToBlockBlob()` which also takes a `ReadSeekCloser`

aksdb: I think this requirement is hard to meet in services.
Often I stream data right from a client through my service to blob storage. The HTTP connection I am feeding from can't be a `Seeker`. If azblob wants a retry mechanism, I think it should implement this internally with some kind of (configurable?) buffer. The burden should not be on the caller.

mohsha-msft:
Running CI.

mohsha-msft (approved):

PR #15971, opened 2021-10-28
Keys challenge

seankane-msft:
Adds Challenge Auth Policy for KV


sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return resp, err
+}
+
+// parses Tenant ID from auth challenge
+// https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000
+func (k KeyVaultChallengePolicy) parseTenant(url string) *string {
+	parts := strings.Split(url, "/")
+	tenant := parts[len(parts)-1]
```
chlowell, line 112: This is anxiety talking but I would take the first path segment, not the last, so that this code still works when there's more to the path (not that I expect there to be; and yes, I think it's safer to assume the tenant ID remains the first segment)

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,189 @@
...
+	if resp.StatusCode == 401 {
+		// Check for a new auth policy
+		err := k.findScopeAndTenant(resp)
+
+		// Error parsing challenge, doomed to fail. Return
+		if err != nil {
+			return nil, err
+		}
+
+		tk, err := k.mainResource.GetResource(as)
```
chlowell, line 89: Could this return the token Key Vault just rejected?

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,189 @@
...
+		k.tenantID = nil
+		return errors.New("response has no WWW-Authenticate header for challenge authentication")
+	}
+
+	// Strip down to auth and resource
+	// Format is "Bearer authorization=\"<site>\" resource=\"<site>\""
+	authHeader = strings.ReplaceAll(authHeader, "Bearer ", "")
+
+	parts := strings.Split(authHeader, " ")
+	if len(parts) != 2 {
```
chlowell, line 131: The challenge could contain additional parameters, so I suggest removing this check. If we get authorization and resource or scope parameters, we can proceed, ignoring any others.

sdk/keyvault/internal/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal
+
+go 1.17
```
chlowell, line 3: 1.16?

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,189 @@
...
+	return &tenant
+}
+
+// sets the k.scope and k.tenantID from the WWW-Authenticate header
+func (k *KeyVaultChallengePolicy) findScopeAndTenant(resp *http.Response) error {
+	authHeader := resp.Header.Get("WWW-Authenticate")
+	if authHeader == "" && k.scope == nil && k.tenantID == nil {
+		// No WWW-Authenticate and we don't already know the scope and tenatnID
+		k.scope = nil
+		k.tenantID = nil
```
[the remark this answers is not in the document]
seankane-msft: This takes care of the situation of no `WWW-Authenticate` on the very first request. This method is also called in the scenario where the vault changes tenants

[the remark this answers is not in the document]
chlowell: But why assign nil to these fields in a block that only executes when those fields have nil values? You effectively have this:
```go
if A == nil && B == nil {
    A = nil
    B = nil
```

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if k.scope == nil || k.tenantID == nil {
+		// First request, get both to get the token
+		challengeReq, err := k.getChallengeRequest(req)
+		if err != nil {
+			return nil, err
+		}
+
+		challengeResp, err := k.transport.Do(challengeReq)
```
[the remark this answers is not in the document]
seankane-msft: What's the best way to do that? Should I pass a reference to the pipeline to call all additional policies? If I call `req.Next()` I assume that will entirely run the pipeline so that could be acceptable too right?

sdk/keyvault/internal/challenge_policy.go
seankane-msft: It can in which case either the tenant is no longer correct or the scope is invalid. At this point we have no way to authenticate the request and should fail.

sdk/keyvault/internal/challenge_policy.go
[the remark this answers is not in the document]
seankane-msft: Poor timing, this was removed but my comment was made before I pushed the commits. I believe were on the same page here now?

sdk/keyvault/internal/challenge_policy.go
chlowell: When we get a 401 we know the cached token is invalid, so we should discard it to prevent the client from using it again.

seankane-msft: Taken care of, I instantiate a new ExpiringResource in that situation (if that's the wrong way to handle it let me know).

seankane-msft: Update: Jeff and I fixed this by adding a `Reset` to the `ExpiringResource`

sdk/keyvault/internal/challenge_policy.go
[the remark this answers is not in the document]
chlowell: Not the entire pipeline--that would recurse into this policy. `req.Next()` is the right idea. We know `req` has been through every policy ahead of this one, now we want to send it through all the following policies. One possible gotcha is that `req.Next()` treats the policy list like a queue, popping the next policy off with each call, so be sure that calling `req.next()` here doesn't affect the policy list when you call it again below.

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,195 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if token, ok := tk.(*azcore.AccessToken); ok {
+			req.Raw().Header.Set(
+				headerAuthorization,
+				fmt.Sprintf("%s %s", bearerHeader, token.Token),
+			)
+		}
```
chlowell, line 105: Missing an `else` here. If we didn't get a token, there's no reason to send another request.

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,195 @@
...
+		err := k.findScopeAndTenant(resp)
+
+		// Error parsing challenge, doomed to fail. Return
+		if err != nil {
+			return nil, err
+		}
+
+		tk, err := k.mainResource.GetResource(as)
+		if err != nil {
+			return nil, err
```
chlowell, line 97: We have a response from Key Vault which might be useful in debugging. Should we return it here?

sdk/keyvault/internal/challenge_policy.go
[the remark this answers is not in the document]
seankane-msft: the challenge request should pass through the entire pipeline or the original request should pass through? The original request does but if we want the challenge request to do so then I should add a copy of the clients pipeline to this struct and use that pipeline to do any modifications of this request. Does that sound correct?

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,195 @@
...
+func parseTenant(url string) *string {
+	parts := strings.Split(url, "/")
+	tenant := parts[3]
+	return &tenant
+}
+
+// sets the k.scope and k.tenantID from the WWW-Authenticate header
+func (k *KeyVaultChallengePolicy) findScopeAndTenant(resp *http.Response) error {
+	authHeader := resp.Header.Get("WWW-Authenticate")
+	if authHeader == "" && k.scope == nil && k.tenantID == nil {
```
chlowell, line 128:
```suggestion
	if authHeader == "" {
```
Is there any reason for this function to continue when the response doesn't contain a challenge?

sdk/keyvault/internal/expiring_resource_test.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func TestNewExpiringResourceError(t *testing.T) {
+	er := NewExpiringResource(func(state interface{}) (newResource interface{}, newExpiration time.Time, err error) {
+		return "", time.Time{}, errors.New("failed")
+	})
+	res, err := er.GetResource("stale")
+	require.Error(t, err)
+	require.Equal(t, "", res)
+}
```
chlowell, line 48: missing newline

sdk/keyvault/azkeys/utils_test.go
```diff
@@ -26,30 +26,97 @@ import (
...
+
+		tenantID := os.Getenv("AZKEYS_TENANT_ID")
+		err = recording.AddHeaderRegexSanitizer("WWW-Authenticate", "00000000-0000-0000-0000-000000000000", tenantID, nil)
+		if err != nil {
+			panic(err)
+		}
+
+		mhsmURL, ok := os.LookupEnv("AZURE_MANAGEDHSM_URL")
+		if ok {
+			fmt.Println("Did not find managed HSM url, skipping those tests")
```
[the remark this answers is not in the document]
chlowell: New problem: `if !ok` you enable MHSM tests anyway

sdk/keyvault/azkeys/azcrypto/crypto_client.go
```diff
@@ -97,21 +97,21 @@ func NewClient(keyURL string, credential azcore.TokenCredential, options *Client
...
 		genOptions.PerRetryPolicies,
-		shared.NewKeyVaultChallengePolicy(credential, genOptions.Transport),
+		shared.NewKeyVaultChallengePolicy(credential),
 	)
 	conn := generated.NewConnection(genOptions)
 
 	vaultURL, err := parseVaultURL(keyURL)
 	if err != nil {
-		return &Client{}, err
+		return Client{}, err
```
chlowell, line 106: The guidelines insist on constructors returning clients by reference, what's the reason to return by value here? One benefit of returning a pointer is you can do this to prevent users calling methods on an uninitialized client:
```suggestion
		return nil, err
```

seankane-msft: Good catch, the example shows a return by pointer which is what I first glanced at. Will update

chlowell: Still returning structs here

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -0,0 +1,203 @@
...
+	if requestErr != nil {
+		return nil, requestErr
+	}
+
+	// If it fails and has a 401, try it with a new token
+	if resp.StatusCode == 401 {
+		// Force a new token
+		k.mainResource.Reset()
+
+		// Check for a new auth policy
```
chlowell, line 86: That's not what the line of code does. Personally, I find it clear enough, no need for a comment

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -102,18 +99,14 @@ func (k *KeyVaultChallengePolicy) Do(req *policy.Request) (*http.Response, error
...
 			// tk is not an azcore.AccessToken type, something went wrong and we should return the 401 and accompanying error
 			return resp, requestErr
 		}
 
-		resp, err = http.DefaultClient.Do(req.Raw())
-		if err != nil {
-			// A second request failed, return error
-			return nil, err
-		}
+		return req.Next()
```
chlowell, line 109: This does what we want even though we called `req.Next()` already?

sdk/keyvault/internal/challenge_policy_test.go
```diff
@@ -6,13 +6,45 @@
...
+	p.findScopeAndTenant(&resp)
+	if *p.scope != scope {
+		t.Fatalf("scope was not properly parsed, got %s, expected %s", *p.scope, scope)
+	}
+
+	if *p.tenantID != fakeTenant {
+		t.Fatalf("tenant ID was not properly parsed, got %s, expected %s", *p.tenantID, fakeTenant)
+	}
+
+	resp.Header.Set("WWW-Authenticate", "Bearer authorization=\"https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000\", resource=\"https://managedhsm.azure.net\" scope=\"https://vault.azure.net/.default\"")
```
chlowell, line 40: Suggest being more thorough here. `scope` and `resource` can be in any order, with arbitrary other (comma-separated) parameters mixed in

sdk/keyvault/internal/challenge_policy.go
seankane-msft: Looking at the source it will throw an error, I believe we can use `http.DefaultClient` here, the request has gone through all the policies and this block only updates the authorization header.

sdk/keyvault/internal/challenge_policy_test.go
```diff
@@ -38,12 +38,37 @@ func TestFindScopeAndTenant(t *testing.T) {
...
 	resp.Header.Set("WWW-Authenticate", "Bearer authorization=\"https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000\", resource=\"https://managedhsm.azure.net\" scope=\"https://vault.azure.net/.default\"")
+	p.findScopeAndTenant(&resp)
+	if *p.scope != scope {
+		t.Fatalf("scope was not properly parsed, got %s, expected %s", *p.scope, scope)
+	}
+	if *p.tenantID != fakeTenant {
+		t.Fatalf("tenant ID was not properly parsed, got %s, expected %s", *p.tenantID, fakeTenant)
+	}
 
+	resp.Header.Set("WWW-Authenticate", "Bearer resource=\"https://managedhsm.azure.net\" scope=\"https://vault.azure.net/.default\", authorization=\"https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000\"")
```
chlowell, line 49: Suggest building challenges with e.g. `fmt.Sprintf` to make clear which parameter has the expected value

sdk/keyvault/internal/challenge_policy_test.go
```diff
@@ -38,12 +38,37 @@ func TestFindScopeAndTenant(t *testing.T) {
...
+	resp.Header.Set("WWW-Authenticate", "Bearer authorization=\"https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000\", unimportantkey=\"unimportantvalue\" resource=\"https://vault.azure.net/.default\"")
+	p.findScopeAndTenant(&resp)
+	if *p.scope != "https://vault.azure.net/.default" {
+		t.Fatalf("scope was not properly parsed, got %s, expected %s", *p.scope, "https://vault.azure.net/.default")
+	}
+	if *p.tenantID != fakeTenant {
+		t.Fatalf("tenant ID was not properly parsed, got %s, expected %s", *p.tenantID, fakeTenant)
+	}
 
+	resp.Header.Set("WWW-Authenticate", "Bearer   authorization=\"https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000\",    unimportantkey=\"unimportantvalue\"   resource=\"https://vault.azure.net/.default\"    fakekey=\"fakevalue\"			")
```
chlowell, line 67: Another interesting case is `scope` without `resource`

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -135,27 +139,24 @@ func (k *KeyVaultChallengePolicy) findScopeAndTenant(resp *http.Response) error
...
-		url := subParts[1]
-
-		url = strings.ReplaceAll(url, "\"", "")
-		vals = append(vals, url)
 	}
 
-	if !strings.HasSuffix(vals[1], "/.default") {
-		vals[1] += "/.default"
+	k.tenantID = parseTenant(vals["authorization"])
+	if scope, ok := vals["resource"]; ok {
```
chlowell, line 151: If the challenge has a `scope` parameter, we should prefer it

sdk/keyvault/internal/challenge_policy.go
chlowell: Not all policies modify request state. For example, we want this request to pass through the logging policy again.

seankane-msft: I'm not sure how to pass this request through the pipeline then. The policies attached to the original request will be emptied now and the pipeline is a private member of the `Connection` struct in the generated code. Should this policy keep a copy of an entire pipeline minus this challenge policy for this request?

PR #15980, opened 2021-10-29
[azservicebus] Make ReceiveMessages overall timeout is controlled by ctx

### LINKED ISSUE #15981: Remove the timeout parameter for ReceiveMessages() in favor of just letting the user use the passed in context
Using the context for this kind of timeout is perfectly idiomatic, and doesn't compromise the functions intention at all.

richardpark-msft:
This PR changes the ReceiveMessages function so the overall timeout is controlled via the user's 'ctx' parameter, rather than having a separately specified 'MaxWaitTime' variable.

(This is based a previous API review)

Fixes #15981

### DIFF
--- a/sdk/messaging/azservicebus/example_receiver_test.go
+++ b/sdk/messaging/azservicebus/example_receiver_test.go
@@ -56,8 +56,8 @@ func ExampleClient_NewReceiverForSubscription_deadLetterQueue() {
 }
 
 func ExampleReceiver_ReceiveMessages() {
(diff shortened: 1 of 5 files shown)


jhendrixMSFT (approved):

richardpark-msft:
I think I scheduled my live test too soon so it cancelled the basic CI job. 

Live tests passed, btw. I'm just triggering the normal tests using an empty commit.

Issue #15925, opened 2021-10-25
[azservicebus] Some more API changes


richardpark-msft:
Some more API cleanup:

- [x] #15981
- [x] Prefix the lock mode constants with ReceiveMode (mentioned by @seankane-msft)
- [x] Unembed the Receiver in SessionReceiver if to remove the extraneous RenewMessageLock() function

richardpark-msft:
All of this work is done as of the 3 linked PRs.

PR #15985, opened 2021-10-30
[azservicebus] Session API work (session state support, removing RenewMessageLock), etc...

### LINKED ISSUE #15773: Add Set/GetSessionState support to SessionReceiver

richardpark-msft:
- Adding in ability to save/restore session state (just an arbitrary bunch of bytes)
- Prefix the ReceiveMode constants with `ReceiveMode`
- Removing the unneeded RenewMessageLock() function from the session receiver.

Fixes #15773, #15925

### DIFF
--- a/sdk/messaging/azservicebus/internal/mgmt.go
+++ b/sdk/messaging/azservicebus/internal/mgmt.go
@@ -58,6 +58,9 @@ type MgmtClient interface {
 
 	RenewLocks(ctx context.Context, linkName string, lockTokens []amqp.UUID) ([]time.Time, error)
 	RenewSessionLock(ctx context.Context, sessionID string) (time.Time, error)
+
+	GetSessionState(ctx context.Context, sessionID string) ([]byte, error)
+	SetSessionState(ctx context.Context, sessionID string, state []byte) error
 }
 
 func newMgmtClient(ctx context.Context, links AMQPLinks, ns NamespaceForMgmtClient) (MgmtClient, error) {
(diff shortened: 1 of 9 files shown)


sdk/messaging/azservicebus/internal/mgmt.go
```diff
@@ -58,6 +58,9 @@ type MgmtClient interface {
 
 	RenewLocks(ctx context.Context, linkName string, lockTokens []amqp.UUID) ([]time.Time, error)
 	RenewSessionLock(ctx context.Context, sessionID string) (time.Time, error)
+
+	GetSessionState(ctx context.Context, sessionID string) ([]byte, error)
+	SetSessionState(ctx context.Context, sessionID string, state []byte) error
 }
 
 func newMgmtClient(ctx context.Context, links AMQPLinks, ns NamespaceForMgmtClient) (MgmtClient, error) {
```
seankane-msft, line 66: Curiosity: Why have the `MgmtClient` interface and the `mgmtClient` struct separately? Is there another type that implements the MgmtClient interface?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -172,6 +172,159 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, maxMessages int, options
...
+
+		receivedMessages = append(receivedMessages, receivedMsg)
+	}
+
+	return receivedMessages, nil
+}
+
+// PeekOptions contains options for the `Receiver.PeekMessages`
+// function.
+type PeekOptions struct {
```
seankane-msft, line 203: Nit: shouldn't this be `PeekMessagesOptions` per [guidelines](https://azure.github.io/azure-sdk/golang_introduction.html#optional-parameters). Are there multiple types of Peeks or is the only Peek PeekMessages?

seankane-msft (approved):
Two small comments, otherwise LGTM

sdk/messaging/azservicebus/internal/mgmt.go
richardpark-msft: No, single implementor. The reason for it is for unit testing.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: I think you're right, I'll rename.

richardpark-msft: Renamed. Good catch!

PR #16040, opened 2021-11-01
adding RouteURL method to internal/recording

seankane-msft:


sdk/internal/recording/recording.go
```diff
@@ -473,11 +473,21 @@ func defaultOptions() *RecordingOptions {
 	}
 }
 
-func (r RecordingOptions) hostScheme() string {
+func (r RecordingOptions) Host() string {
```
chlowell, line 476: In `http://localhost:5000`, "http" is the scheme and "localhost" the host ("localhost:5000" is the authority)

chlowell (commented):
What's the scenario motivating this?

sdk/internal/recording/recording.go
seankane-msft: Getting the proxy URL ended up being a lot of repeated code, I think it's better to have a method `ProxyURL` that returns the full URL to the proxy instead of separate URLs for the Host and Scheme.

seankane-msft:
> What's the scenario motivating this?

I made some changes and now the code reduction is evident by looking at `sdk/data/aztables/proxy_test.go`. The RouteURL method would need to be in each sdk, this puts it in one place and abstracts away more of the implementation details.

sdk/internal/recording/recording.go
```diff
@@ -473,11 +473,35 @@ func defaultOptions() *RecordingOptions {
...
+		rawReq.URL.Host = r.hostAndPort()
+		rawReq.Host = r.hostAndPort()
+
+		rawReq.Header.Set(UpstreamURIHeader, fmt.Sprintf("%v://%v", r.scheme(), originalURLHost))
+		rawReq.Header.Set(ModeHeader, GetRecordMode())
+		rawReq.Header.Set(IDHeader, GetRecordingId(t))
+	}
+}
+
+func (r RecordingOptions) hostAndPort() string {
```
chlowell, line 489: I see `net/url` defines `URL.Host` as optionally including a port. That doesn't align with the URL schema but it admits a reasonable argument for us to align with the standard library here:
```suggestion
func (r RecordingOptions) host() string {
```

sdk/data/aztables/proxy_test.go
```diff
@@ -84,16 +84,17 @@ func NewRecordingPolicy(t *testing.T, o *recording.RecordingOptions) policy.Poli
...
-		req.Raw().URL.Scheme = p.Scheme()
-		req.Raw().URL.Host = p.Host()
-		req.Raw().Host = p.Host()
-
-		req.Raw().Header.Set(recording.UpstreamURIHeader, fmt.Sprintf("%v://%v", p.Scheme(), originalURLHost))
-		req.Raw().Header.Set(recording.ModeHeader, recording.GetRecordMode())
-		req.Raw().Header.Set(recording.IDHeader, recording.GetRecordingId(p.t))
-	}
+	// if recording.GetRecordMode() != "live" && !recording.IsLiveOnly(p.t) {
+	p.options.RouteURL(p.t, req.Raw())
```
chlowell, line 88: Why do client modules need to call this? Naively, I would expect a recording policy implemented in a common internal module to replace the URL.

sdk/data/aztables/proxy_test.go
```diff
@@ -84,16 +84,17 @@ func NewRecordingPolicy(t *testing.T, o *recording.RecordingOptions) policy.Poli
...
+	// originalURLHost := req.Raw().URL.Host
+	// req.Raw().URL.Scheme = p.Scheme()
+	// req.Raw().URL.Host = p.Host()
+	// req.Raw().Host = p.Host()
+
+	// req.Raw().Header.Set(recording.UpstreamURIHeader, fmt.Sprintf("%v://%v", p.Scheme(), originalURLHost))
```
chlowell, line 94: Do you mean to keep all these comments?

sdk/internal/recording/recording.go
```diff
@@ -473,11 +473,35 @@ func defaultOptions() *RecordingOptions {
 	}
 }
 
-func (r RecordingOptions) hostScheme() string {
+func (r RecordingOptions) RouteURL(t *testing.T, rawReq *http.Request) {
```
chlowell, line 476: To me, "route" suggests finding a handler, not modifying the request. ReplaceURL?

seankane-msft: That's a reasonable thought, ReplaceURL isn't obvious what is being replaced. RouteURLToProxy? It's longer but it's more clear whats hapening

chlowell: I still shy away from "route", perhaps just because of past web dev experience, but setting that aside it doesn't suggest modifying the request. What's changing is the authority so, some verb + authority? ReplaceAuthority? SubstituteAuthority?

sdk/internal/recording/recording.go
```diff
@@ -473,11 +473,35 @@ func defaultOptions() *RecordingOptions {
...
+func (r RecordingOptions) scheme() string {
 	if r.UseHTTPS {
-		return "https://localhost:5001"
+		return "https"
 	}
-	return "http://localhost:5000"
+	return "http"
+}
+
+func (r RecordingOptions) hostScheme() string {
```
chlowell, line 503: Nitpicking: this is often called the "base URL" in practice--everything before the path. I don't know that there's a proper term for it in the URL schema.

sdk/internal/recording/recording.go
seankane-msft: Changed to ReplaceAuthority which I believe is more clear

sdk/internal/CHANGELOG.md
```diff
@@ -3,10 +3,12 @@
 ## 0.8.2 (Unreleased)
 
 ### Features Added
+* Adding `RecordingOptions.RouteURL` to handle routing requests to the test proxy
 
 ### Breaking Changes
 
 ### Bugs Fixed
+* When running in `LiveMode` the sanitizers will not send requests.
```
chlowell, line 11: Is this a reasonable summary for users of the recording package?
```suggestion
* Adding recording sanitizers has no effect when running in `LiveMode`
```

chlowell (approved):

PR #16044, opened 2021-11-01
[azservicebus] AdminClient can now create topics and subscriptions

### LINKED ISSUE #15090: [service-bus] ServiceBusAdministrationClient
Track 2 SDKs should have a top level client that manages communication with the Service Bus ATOM endpoint (basically - a similar endpoint to what you can accomplish with ARM, but using a Service Bus connection string).

Work:
- [ ] Adding in Create/Delete/Get support for queues: https://github.com/Azure/azure-sdk-for-go/pull/15790
- [ ] GetQueueRuntimeProperties, ListQueues, UpdateQueues
- [ ] Subscriptions
- [ ] #16045
- [ ] Topics
- [ ] GetNamespaceProperties

Issues that I've seen that we should address as part of this work:
- [#229: Blank error returned when using a key without Manage permission](https://github.com/Azure/azure-service-bus-go/issues/229)

Future considerations:
- Generate using autorest and the swagger file for the endpoint: [link](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/sb_dataplane_namespace/specification/servicebus/data-plane/servicebus-swagger.json). @chlowell gave this a shot but it looks like our autorest impl for Go isn't quite up to the task yet. For now we'll ship using the existing EntityManager, with a unified surface in `AdminClient`.

richardpark-msft:
Adds in support for creating topics and subscriptions using the AdminClient.

There were also some bits of feedback cleaned up in here - all the request/response objects should be proper per our guidelines.

Fixes #15090


sdk/messaging/azservicebus/admin_client_subscription.go
```diff
@@ -0,0 +1,532 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &SubscriptionProperties{
+		Name:                             subscriptionName,
+		RequiresSession:                  desc.RequiresSession,
+		DeadLetteringOnMessageExpiration: desc.DeadLetteringOnMessageExpiration,
+		EnableDeadLetteringOnFilterEvaluationExceptions: desc.DeadLetteringOnFilterEvaluationExceptions,
```
chlowell, line 387: Nitpicking: needs a fmt

sdk/messaging/azservicebus/internal/atom/models.go
```diff
@@ -0,0 +1,230 @@
...
+	subscriptionContent struct {
+		XMLName                 xml.Name                `xml:"content"`
+		Type                    string                  `xml:"type,attr"`
+		SubscriptionDescription SubscriptionDescription `xml:"SubscriptionDescription"`
+	}
+)
+
+type (
+	// Entity is represents the most basic form of an Azure Service Bus entity.
+	Entity struct {
```
chlowell, line 226: Does this need to be in a separate `type` declaration because it's referenced in the one above?

sdk/messaging/azservicebus/sample.env
```diff
@@ -1,5 +1,6 @@
 # These values should be filled out when testing.
 SERVICEBUS_CONNECTION_STRING=<service bus connection string>
+SERVICEBUS_CONNECTION_STRING_NO_MANAGE=<service bus connection string with listen and send permissions only>
```
chlowell, line 3: This file is a good idea, thanks for maintaining it

richardpark-msft: Carried this over from the JS team. I think technically there should be an .env file for samples (ie, users) and one for developers of the package itself. We can figure that out over time.

sdk/messaging/azservicebus/admin_client_subscription.go
richardpark-msft: Fixed.

sdk/messaging/azservicebus/admin_client_queue.go
```diff
@@ -5,33 +5,340 @@ package azservicebus
...
+
+// AddQueue creates a queue using defaults for all options.
+func (ac *AdminClient) AddQueue(ctx context.Context, queueName string) (*AddQueueResponse, error) {
+	return ac.AddQueueWithProperties(ctx, &QueueProperties{
+		Name: queueName,
+	})
+}
+
+// CreateQueue creates a queue with configurable properties.
+func (ac *AdminClient) AddQueueWithProperties(ctx context.Context, properties *QueueProperties) (*AddQueueResponse, error) {
```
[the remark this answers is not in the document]
richardpark-msft: I had a discussion about a similar thing with @jhendrixMSFT about this before: [(link)](https://github.com/Azure/azure-event-hubs-go/pull/226#discussion_r671504002)

The gist was that it's okay if the nil passes through and it panics _but_ the critical part is it needs to be close to the point of origin. This one should fit that bill since we panic immediately because we do actually try to use properties.

So I'm going to leave this one with the acceptance that this is the Go way (even though it was counter to my own instincts at the time).

sdk/messaging/azservicebus/internal/atom/models.go
richardpark-msft: You're so charitable. Just a silly oversight (now corrected).

chlowell (approved):

PR #16067, opened 2021-11-03
[chunkwriting] Return original buffer to the pool

serprex:
When the buffer isn't filled to capacity a new slice is created that's
smaller than the original. In that case the smaller slice is returned to
the pool which prevents the rest of the capacity from being used again.

The solution is to pass the original slice through and attach the
length. This allows the original slice to be returned to the pool once
the operation is complete.

This change also simplifies the `sendChunk` method, ensuring that the
buffer is returned to the TransferManager even when no bytes were read
from the reader.

Copied from https://github.com/Azure/azure-storage-blob-go/pull/284

### DIFF
--- a/sdk/storage/azblob/chunkwriting.go
+++ b/sdk/storage/azblob/chunkwriting.go
@@ -112,6 +112,7 @@ type copier struct {
 type copierChunk struct {
 	buffer []byte
 	id     string
+	length int
 }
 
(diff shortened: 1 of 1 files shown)


serprex:
@mohsha-msft CI was failing earlier, seems unrelated, since this was also the case with my other io.Reader PR, after rebasing it's resolved

mohsha-msft (approved):

mohsha-msft:
Running CI.

Issue #15330, opened 2021-08-20
Provide log sanitizer functionality


jhendrixMSFT:
We have this in track 1.5.

https://github.com/Azure/azure-pipeline-go/blob/master/pipeline/core.go#L261

jhendrixMSFT:
Fixed in pending `azcore` release (post `v0.20.0`).

PR #16076, opened 2021-11-03
Add allow-lists for headers and query parameters

jhendrixMSFT:
LogOptions now contains allow-lists for headers and query parameters.
Any header and/or query parameter not in the allow lists, combined with
the default set, will have its value written as REDACTED to the log.

### DIFF
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -59,6 +59,7 @@ type LogOptions struct {
 
 	// AllowedHeaders is the slice of headers to log with their values intact.
 	// All headers not in the slice will have their values REDACTED.
+	// Applies to request and response headers.
 	AllowedHeaders []string
 
 	// AllowedQueryParams is the slice of query parameters to log with their values intact.
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -56,6 +56,14 @@ type LogOptions struct {
 	// The default value is false.
 	// NOTE: enabling this can lead to disclosure of sensitive information, use with care.
 	IncludeBody bool
+
+	// AllowedHeaders is the slice of headers to log with their values intact.
+	// All headers not in the slice will have their values REDACTED.
+	AllowedHeaders []string
+
+	// AllowedQueryParams is the slice of query parameters to log with their values intact.
+	// All query parameters not in the slice will have their values REDACTED.
+	AllowedQueryParams []string
 }
 
 // RetryOptions configures the retry policy's behavior.
--- a/sdk/azcore/runtime/policy_logging.go
+++ b/sdk/azcore/runtime/policy_logging.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/15330

sdk/azcore/runtime/policy_logging.go
```diff
@@ -30,7 +33,49 @@ func NewLogPolicy(o *policy.LogOptions) policy.Policy {
 	if o == nil {
 		o = &policy.LogOptions{}
 	}
-	return &logPolicy{options: *o}
+	// construct default hash set of allowed headers
+	allowedHeaders := map[string]struct{}{
```
jhendrixMSFT, line 37: This set was taken from the [.NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L21) implementation.

sdk/azcore/runtime/policy_logging.go
```diff
@@ -30,7 +33,49 @@ func NewLogPolicy(o *policy.LogOptions) policy.Policy {
...
+		"x-ms-client-request-id":        {},
+		"x-ms-return-client-request-id": {},
+	}
+	// add any caller-specified allowed headers to the set
+	for _, ah := range o.AllowedHeaders {
+		allowedHeaders[strings.ToLower(ah)] = struct{}{}
+	}
+	// now do the same thing for query params
+	allowedQP := map[string]struct{}{
+		"api-version": {},
```
jhendrixMSFT, line 69: I added this as I think it's valuable (.NET doesn't specify any).  Let me know if there are other query params we know are safe to add.

sdk/azcore/runtime/response.go
```diff
@@ -146,42 +143,3 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
 		return fmt.Errorf("unrecognized byte array format: %d", format)
 	}
 }
-
```
jhendrixMSFT, line 149: This content was moved to policy_logging.go

JeffreyRichter (approved):
This looks great. One potential issue...
The customer create a LogOptions and sets some headers/QPs. This then get passed to the client ctor which wants to add more headers/QPs - this is a problem because the ctor should NOT modify the customer's Options instance. The ctor would have to make a deep copy of itself, then add the service-specific headers/QPs and then pass the copy down to NewPipeline.

We had a similar problem with customer & client lib policies and we made NewPipeline take 2 extra []Policy parameters. 

Maybe we need a client-only struct with extra fields in it (for future proofing) that customers don't know about but client lib code can use to augment the customer's client options without mutating it in any way.

sdk/azcore/policy/policy.go
```diff
@@ -56,6 +56,14 @@ type LogOptions struct {
 	// The default value is false.
 	// NOTE: enabling this can lead to disclosure of sensitive information, use with care.
 	IncludeBody bool
+
+	// AllowedHeaders is the slice of headers to log with their values intact.
+	// All headers not in the slice will have their values REDACTED.
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed.

jhendrixMSFT:
This implementation doesn't mutate the customer-provided slices.

chlowell (approved):

PR #16082, opened 2021-11-04
Test case  track2

Alancere:


sdk/resourcemanager/compute/armcompute/utils_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (f *fakeCredential) GetToken(ctx context.Context, options policy.TokenRequestOptions) (*azcore.AccessToken, error) {
+	return &azcore.AccessToken{
+		Token:     "faketoken",
+		ExpiresOn: time.Date(2040, time.January, 1, 1, 1, 1, 1, time.UTC),
+	}, nil
+}
+
+func createResourceGroup(t *testing.T,cred azcore.TokenCredential,option *arm.ClientOptions,subscriptionID,name,location string) (*armresources.ResourceGroup,func()) {
```
seankane-msft, line 122: My guess for why this is failing is that you don't actually need to run this when you're in playback, instead returning valid values for the playback test. You can add a snippet to the top of this function
``` 
    if recording.GetRecordMode() == recording.PlaybackMode {
        return &rg.ResourceGroup{/*fill in here */}, func() {}
    }
```

PR #16089, opened 2021-11-04
Remove AzureGermany Endpoint

chlowell:
Azure Germany closed (all subscriptions disabled) on 10/29: https://www.microsoft.com/en-us/cloud-platform/germany-cloud-regions


jhendrixMSFT (approved):

jhendrixMSFT:
I believe we have this in `azcore/arm` too.

PR #16091, opened 2021-11-04
Add options struct for NewPipeline

chlowell:
Following from https://github.com/Azure/azure-sdk-for-go/pull/16076#pullrequestreview-798084846, this adds a new struct for pipeline and policy options intended for use by SDK developers. Pipeline constructors merge these options with the application's `azcore.ClientOptions` without mutating the latter.

### DIFF
--- a/sdk/azcore/arm/runtime/pipeline.go
+++ b/sdk/azcore/arm/runtime/pipeline.go
@@ -12,13 +12,12 @@ import (
 	armpolicy "github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/pipeline"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
-	azpolicy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 )
 
 // NewPipeline creates a pipeline from connection options.
 // The telemetry policy, when enabled, will use the specified module and version info.
-func NewPipeline(module, version string, cred azcore.TokenCredential, options *arm.ClientOptions) pipeline.Pipeline {
+func NewPipeline(module, version string, cred azcore.TokenCredential, sdkOpts azruntime.SDKOptions, options *arm.ClientOptions) pipeline.Pipeline {
 	if options == nil {
 		options = &arm.ClientOptions{}
 	}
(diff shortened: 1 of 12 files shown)


sdk/azcore/runtime/pipeline.go
```diff
@@ -11,26 +11,34 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 )
 
+// SDKOptions contains Pipeline options for SDK developers
+type SDKOptions struct {
```
jhendrixMSFT, line 15: Shouldn't this be `PipelineOptions`?

sdk/azcore/runtime/pipeline.go
```diff
@@ -11,26 +11,34 @@ import (
...
+	AllowedHeaders, AllowedQueryParameters []string
```
jhendrixMSFT, line 16: Why are these duplicated from `LogOptions`, what's the general pattern here?

JeffreyRichter (approved):

sdk/azcore/runtime/pipeline.go
chlowell: This is to allow an SDK developer to augment the `ClientOptions` provided by the application without mutating it. The SDK developer could do that by cloning the application's `ClientOptions` and augmenting the clone--we could make that easier with a helper--but I believe @JeffreyRichter wants the flexibility to grow this new struct without having to extend `ClientOptions` as well.

JeffreyRichter: Yes, plus we never want to mutate the csutomer's ClientOptions because they use it to create multiple clients.

jhendrixMSFT: I'm onboard with having a struct here so we can expand the `NewPipeline` funcs, but we should change the name to match our other patterns.

I don't think we need to duplicate any options fields here though to achieve the goal of not mutating caller-provided state.

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -12,30 +12,28 @@ import (
...
 	ep := options.Host
 	if len(ep) == 0 {
 		ep = arm.AzurePublicCloud
 	}
-	perCallPolicies := []azpolicy.Policy{}
+	authPolicy := NewBearerTokenPolicy(cred, &armpolicy.BearerTokenOptions{
+		Scopes:           []string{shared.EndpointToScope(string(ep))},
+		AuxiliaryTenants: options.AuxiliaryTenants,
+	})
+	o := plOpts
```
jhendrixMSFT, line 32: This is unnecessary as a copy was already made at the function call-site.

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -12,30 +12,28 @@ import (
...
+	o.PerRetry = append(o.PerRetry, authPolicy)
```
jhendrixMSFT, line 33: Depending on the capacity of `PerRetry` this can modify the underlying array.  I put together an example the demonstrates this [here](https://play.golang.org/p/KMu97L8xuVw).

To avoid potentially mutating the backing array of a slice, we should just create new arrays that merge the inputs together.

sdk/azcore/runtime/pipeline.go
```diff
@@ -11,26 +11,39 @@ import (
...
 // perRetry: additional policies to invoke once per request and once per retry of that request
-func NewPipeline(module, version string, perCall, perRetry []policy.Policy, options *policy.ClientOptions) Pipeline {
-	if options == nil {
-		options = &policy.ClientOptions{}
+func NewPipeline(module, version string, plOpts PipelineOptions, options *policy.ClientOptions) Pipeline {
+	cp := policy.ClientOptions{}
+	if options != nil {
+		cp = *options
+	}
+	if len(plOpts.AllowedHeaders) > 0 {
```
jhendrixMSFT, line 29: We need to do this same thing for `AllowedQueryParameters` yes?

jhendrixMSFT (approved):

PR #16102, opened 2021-11-05
[Tables] Updating Documentation

### LINKED ISSUE #16041: Using continuation token to retrieve records from Azure Storage Table fails with status 415 Unsupported Media Type and error.data.message contains AtomFormatNotSupported
### Bug Report
 
- import path of package in question, e.g. `.../services/compute/mgmt/2018-06-01/compute`
"github.com/Azure/azure-sdk-for-go/storage"

- SDK version 
github.com/Azure/azure-sdk-for-go v59.0.0
f92de54510280b0eba246b54113abcdbc65fc239

same behaviour with v52.5 as well.

- output of `go version`
go version go1.16 linux/amd64

 
- What happened?
Querying a table using the continuationToken provided in the first call fails.

The table should return the records regardless if it's the first call (no continuationToken) or on subsequent calls (with continuationToken)

- How can we reproduce it?

```
func QueryTableByContinuationToken(azureStorageAccountName string, azureStorageAccountKey string, tableName string, continuationToken string) (*storage.EntityQueryResult, error) {

	table, err := getTableReference(azureStorageAccountName, azureStorageAccountKey, tableName)
	if err != nil {
		return nil, errors.Wrap(err, "failed to get azure storage table reference")
	}

	result, err := table.QueryEntities(30, storage.FullMetadata, &storage.QueryOptions{Top: 0})
	if err != nil {
		return nil, errors.Wrap(err, "failed to query azure storage table")
	}

	result.QueryNextLink.NextLink = &continuationToken

	return result.NextResults(nil)
}
```

calling the previous method passing the continuationToken that looks like this:

`"https://<account>.table.core.windows.net/<table>?%24filter=PartitionKey+eq+%272021-11-01%27+and+Timestamp+le+datetime%272021-11-01T23%3A00%3A00Z%27&%24top=50&NextPartitionKey=1%2116%21MjAyMS0xMS0wMQ--&NextRowKey=1%2116%21MTY6MjI6NDEuMzk3&timeout=30"`

 returns a nil result with an error struct of type github.com/Azure/azure-sdk-for-go/storage.AzureStorageServiceError and the following content:

`{Code: "", Message: "Response body could no be unmarshaled: invalid character '<' looking for beginning of value. Body: <?xml version=\"1.0\" encoding=\"utf-8\"?><error xmlns=\"http://schemas.microsoft.com/ado/2007/08/dataservices/metadata\"><code>AtomFormatNotSupported</code><message xml:lang=\"en-US\">Atom format is not supported.\nRequestId:8e2d22cb-e002-001b-3e54-cf3ce9000000\nTime:2021-11-01T19:16:39.5611825Z</message></error>.", AuthenticationErrorDetail: "", QueryParameterName: "", QueryParameterValue: "", Reason: "", Lang: "", StatusCode: 415, RequestID: "8e2d22cb-e002-001b-3e54-cf3ce9000000", Date: "Mon, 01 Nov 2021 19:16:39 GMT", APIVersion: "2018-03-28"}`

Ubuntu 20.04

seankane-msft:
closes #16041 

### DIFF
--- a/sdk/data/aztables/README.md
+++ b/sdk/data/aztables/README.md
@@ -40,7 +40,7 @@ Once you have the account URL, it can be used to create the service client:
 ```golang
 cred, err := aztables.NewSharedKeyCredential("myAccountName", "myAccountKey")
 handle(err)
-serviceClient, err := aztables.NewServiceClient("https://<myAccountName>.table.core.windows.net/", cred, nil)
+serviceClient, err := aztables.NewServiceClientWithSharedKeyCredential("https://<myAccountName>.table.core.windows.net/", cred, nil)
 handle(err)
 ```
 
@@ -194,7 +194,7 @@ myEntity := aztables.EDMEntity{
 marshalled, err := json.Marshal(myEntity)
 handle(err)
 
-resp, err := client.AddEntity(context.Background(), marshalled, nil)
+resp, err := client.AddEntity(context.TODO(), marshalled, nil)
 handle(err)
 ```
 
@@ -215,7 +215,7 @@ options := &ListEntitiesOptions{
 }
 
 pager := client.List(options) // pass in "nil" if you want to list all entities
-for pager.NextPage(context.Background()) {
+for pager.NextPage(context.TODO()) {
     resp := pager.PageResponse()
     fmt.Printf("Received: %v entities\n", len(resp.Entities))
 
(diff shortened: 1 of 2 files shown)


sdk/data/aztables/README.md
```diff
@@ -40,7 +40,7 @@ Once you have the account URL, it can be used to create the service client:
 ```golang
 cred, err := aztables.NewSharedKeyCredential("myAccountName", "myAccountKey")
 handle(err)
-serviceClient, err := aztables.NewServiceClient("https://<myAccountName>.table.core.windows.net/", cred, nil)
+serviceClient, err := aztables.NewServiceClientWithSharedKeyCredential("https://<myAccountName>.table.core.windows.net/", cred, nil)
```
JeffreyRichter, line 43: @chlowell : Do we want to clean up the "shared key credential" story? Specifically, do we want a NameKeyCredential in azcore which customers use here which would rename this func to NewServiceWithNamedKeyCredential.

And then other client libs can use azcore's NamedKeyCredential (with each lib providing its own policy for how it should work for tht service).

JeffreyRichter (approved):

PR #16106, opened 2021-11-05
Update azidentity V1 adapter sample

chlowell:
Updates the sample to use the latest version of the adapter. Also has the examples list subscriptions instead of creating a resource group, and deletes a redundant one.


sdk/samples/azidentity/SDKV1Adapter/example_SDKV1_test.go
```diff
@@ -5,139 +5,70 @@ package azidentitysamples
...
-				Scopes: []string{"https://management.azure.com/.default"}}})
-	if err != nil {
-		panic("failed to get credential")
-	}
-	// assign the authorizer to your client's authorizer
-	groupsClient.Authorizer = a
-	// perform an operation with the complete client
-	list, err := groupsClient.ListComplete(context.Background(), "", nil)
+	// This scope is correct for Azure Resource Manager in Azure Public Cloud. Other clouds require a different scope.
+	a := azidext.NewTokenCredentialAdapter(cred, []string{"https://management.azure.com//.default"})
```
seankane-msft, line 57: should it be `https://management.azure.com/.default`

seankane-msft (approved):

sdk/samples/azidentity/SDKV1Adapter/example_SDKV1_test.go
chlowell: That might work as well, but no. The v1 resource was `https://management.azure.com/` so the equivalent v2 scope is `https://management.azure.com//.default` ([docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#trailing-slash-and-default))

Issue #15963, opened 2021-10-27
Make maxWaitTimeAfterFirstMessage configurable for Servicebus Receive function


peterzeller:
### Feature Request

The field `maxWaitTimeAfterFirstMessage` in https://github.com/Azure/azure-sdk-for-go/blob/47db7e2935fc81ededd911e1b268c8729d5af38d/sdk/messaging/azservicebus/receiver.go#L150 should be made public so that I can use batch receives with a latency below 1 second.

richardpark-msft:
Let me see what the team thinks, but I agree it'd be useful.

richardpark-msft:
This spurred some interesting discussion in our team, so thank you for submitting it!

In the immediate term, I've changed the the default wait time after first message to 20ms, rather than 1 second. We've got some good practical evidence (given that this is the value that the .NET stack uses) that this should work better in practice.

This will be released in the next beta (coming out in the next few days).

Longer term, we're still discussing if the underlying algorithm so we won't be exposing the setting publicly.

richardpark-msft:
(moving to backlog as this hasn't resurfaced as a customer issue and we're just matching the existing timeout that appears to work well in .net)

richardpark-msft:
Closing - we ended up making this flexible so you can configure it fully in #22154

PR #16110, opened 2021-11-07
[azservicebus] Final API changes for beta.3

richardpark-msft:
Lots of renames and changes based on the API reviews:

Admin API:
- All `Response` structs have been broken down into `Response+Result`
- QueueName, TopicName, SubscriptionName are no longer present in the Properties structs as that doesn't reflect the API itself.
- <Entity>Exists functions have been removed
- Missing fields have been added (UserMetadata, etc..)

Sender and Receiver API:
- SendMessages removed in favor of making SendMessageBatch more intuitive (by simplfying the .AddMessage API)
- Several renames suggested in review
- ReceiveMessages has been tuned to match the .NET limits (which has worked well in practice). This is partly addressing #15963, as our default limit was far higher than needed.

Missing:
- AuthorizationRules, SubscriptionRules (coming in a future PR but requires a bit more thought)


sdk/messaging/azservicebus/internal/mgmt.go
```diff
@@ -49,7 +49,7 @@ type (
 
 type MgmtClient interface {
 	Close(ctx context.Context) error
-	SendDisposition(ctx context.Context, lockToken *amqp.UUID, state Disposition) error
+	SendDisposition(ctx context.Context, lockToken *amqp.UUID, state Disposition, propertiesToModify *map[string]interface{}) error
```
jhendrixMSFT, line 52: I know this is internal but why is it a pointer?

richardpark-msft: Well, now I just feel silly. I'll remove (clearly map[] can be nil!). Fixed.

seankane-msft (commented):
Small comments

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -0,0 +1,92 @@
...
+	Name           string
+}
+
+type GetNamespacePropertiesResult struct {
+	NamespaceProperties
+}
+
+type GetNamespacePropertiesResponse struct {
+	GetNamespacePropertiesResult
+	RawResult *http.Response
```
chlowell, line 64:
```suggestion
	RawResponse *http.Response
```

sdk/messaging/azservicebus/admin/main_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"testing"
+
+	"github.com/joho/godotenv"
+)
+
+func TestMain(m *testing.M) {
+	err := godotenv.Load("../.env")
+
+	if err != nil {
+		log.Printf("Failed to load env file, no live tests will run: %s", err.Error())
```
chlowell, line 18: Will this work in our DevOps pipeline?

sdk/messaging/azservicebus/client.go
```diff
@@ -71,6 +71,10 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	}, options)
 }
 
+// Next overloads (ie, credential sticks with the client)
```
chlowell, line 74: Do you still need this?

sdk/messaging/azservicebus/message.go
```diff
@@ -14,53 +15,81 @@ import (
...
+	MessageID string
 
-		LockToken              [16]byte
-		DeliveryCount          uint32
-		LockedUntil            *time.Time // `mapstructure:"x-opt-locked-until"`
-		SequenceNumber         *int64     // :"x-opt-sequence-number"`
-		EnqueuedSequenceNumber *int64     // :"x-opt-enqueue-sequence-number"`
-		EnqueuedTime           *time.Time // :"x-opt-enqueued-time"`
-		DeadLetterSource       *string    // :"x-opt-deadletter-source"`
+	ContentType      string
```
chlowell, line 22: Is there a functional reason not to embed `Message`?

sdk/messaging/azservicebus/messageBatch.go
```diff
@@ -56,48 +85,34 @@ func (mb *MessageBatch) Add(m SendableMessage) (bool, error) {
...
-	mb.batchEnvelope.Data = make([][]byte, len(mb.marshaledMessages))
-	mb.batchEnvelope.Format = batchMessageFormat
+// createBatchEnvelope makes a copy of the properties of the message, minus any
+// payload fields (like Data, Value or (eventually) Sequence). The data field will be
+// filled in with all the messages when the batch is completed.
+func createBatchEnvelope(am *amqp.Message) *amqp.Message {
+	newAMQPMessage := *am
 
-	copy(mb.batchEnvelope.Data, mb.marshaledMessages)
+	// TODO: we appear to be missing the .Sequence section
```
chlowell, line 113: Author of the above comment seems to have additional insight here :suspect:

sdk/messaging/azservicebus/message.go
richardpark-msft: Yes, and it's all just to make it so you have to use `ReceivedMessage.Body() ([]byte, error)`:

(longer explanation):

It is possible for a user to send an AMQP message directly (ie, bypassing any of the convenience encoding). We don't have this exported in our SDK but all the other track 2 SDKs for Service Bus do (via AMQPAnnotatedMessage).

If they do this the 'Body' field can be misleading - it either won't contain a value (because the payload has been encoded differently than expected) or we could try to put something in there but then the user wouldn't have any way of knowing that something is wrong and it'd be an approximation.

sdk/messaging/azservicebus/client.go
richardpark-msft: Yeah, those'll be coming in soon enough.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -15,8 +15,8 @@ var errReceiveAndDeleteReceiver = errors.New("messages that are received in rece
 
 type settler interface {
 	CompleteMessage(ctx context.Context, message *ReceivedMessage) error
-	AbandonMessage(ctx context.Context, message *ReceivedMessage) error
-	DeferMessage(ctx context.Context, message *ReceivedMessage) error
+	AbandonMessage(ctx context.Context, message *ReceivedMessage, options *AbandonMessageOptions) error
+	DeferMessage(ctx context.Context, message *ReceivedMessage, options *DeferMessageOptions) error
 	DeadLetterMessage(ctx context.Context, message *ReceivedMessage, options *DeadLetterOptions) error
```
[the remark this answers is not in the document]
richardpark-msft: I can't but we always deliver these messages as pointers, so forcing it back would be some unnecessary copying (ie, they'd be doing it, by force, on _every_ message).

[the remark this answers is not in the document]
JeffreyRichter: I think the methods that return ReceivedMessage should return them by value; not by reference.
We have this as a [guideline](https://azure.github.io/azure-sdk/golang_introduction.html#golang-response-logical-entity) . It's called out explicitly (and probably should) but the example shows the client returns are by value.

I think our Receiver type is a client eventhough it is not suffixed with "Client".

sdk/messaging/azservicebus/messageBatch.go
richardpark-msft: Ha, I wish that guy would work here. ;) 

Removed the comment and filed https://github.com/Azure/go-amqp/issues/94, as it's something we need to discuss for full AMQP support. It looks like (based on the TODO comment in the go-amqp source) that they're actually squishing sequence into [][]Data so technically my comment is "incorrect".

sdk/messaging/azservicebus/admin/admin_client.go
richardpark-msft: Fixed

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -0,0 +1,92 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/atom"
+)
+
+// Client allows you to administer resources in a Service Bus Namespace.
+// For example, you can create queues, enabling capabilities like partitioning, duplicate detection, etc..
+// NOTE: For sending and receiving messages you'll need to use the `Client` type instead.
```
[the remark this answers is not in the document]
richardpark-msft: Recursive comments are probably not a good idea. Fixed with a better termination clause. :)

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -0,0 +1,92 @@
...
+type Client struct {
+	em *atom.EntityManager
+}
+
+type AdminClientOptions struct {
+	// for future expansion
+}
+
+// NewClientFromConnectionString creates an Client authenticating using a connection string.
```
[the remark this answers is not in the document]
richardpark-msft: Fixed.

sdk/messaging/azservicebus/admin/main_test.go
richardpark-msft: It just errors out harmlessly in DevOps.

richardpark-msft: This is a "local dev" affordance.

seankane-msft (approved):

sdk/messaging/azservicebus/admin/main_test.go
chlowell: I mean, do the live tests run in DevOps?

sdk/messaging/azservicebus/admin/admin_client_subscription.go
```diff
@@ -50,6 +46,9 @@ type SubscriptionProperties struct {
 	// Status is the current status of the queue.
 	Status *EntityStatus
 
+	// AutoDeleteOnIdle is the idle interval after which the subscription is automatically deleted.
+	AutoDeleteOnIdle *time.Duration
```
chlowell, line 50: I don't mean to start a naming struggle but to me this sounds like a bool

sdk/messaging/azservicebus/receiver.go
```diff
@@ -369,38 +363,33 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
-	}()
+	}
 
 	// Receive until the drain completes, at which point it'll cancel
 	// our context.
-	// NOTE: That's a gap here where we need to be able to drain _only_ the internally cached messages
-	// in the receiver. Filed as https://github.com/Azure/go-amqp/issues/71
 	for {
-		am, err := receiver.Receive(receiveCtx)
+		am, err := receiver.Prefetched(context.Background())
```
chlowell, line 386: Should this use the `ctx` parameter?

sdk/messaging/azservicebus/sender.go
```diff
@@ -74,12 +66,13 @@ func (s *Sender) SendMessage(ctx context.Context, message SendableMessage) error
 	}
 
 	return sender.Send(ctx, message.toAMQPMessage())
+	// func (s *Sender) SendAMQPAnnotatedMessage(ctx context.Context, message *AMQPAnnotatedMessage) error {
```
chlowell, line 69: 🗑️

chlowell (approved):

sdk/messaging/azservicebus/receiver.go
richardpark-msft: No, but I should put a comment on there.

Basically, when the user cancels this function we still have some cleanup we need to do or else we'll have bleed over from call to call (the link is basically a buffer of messages). 

So this is a non-interruptible loop that ensures that any messages sitting in the internal message buffer are out of our hands and into the result.

sdk/messaging/azservicebus/sender.go
richardpark-msft: Done.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Comment added. Let me know if it helps clarify. I also renamed the ctx parameter.

sdk/messaging/azservicebus/admin/admin_client_subscription.go
richardpark-msft: I agree, the name's a bit odd. For better or worse it's this way in the other track 2 SDKs, which are that way because they match the ARM layer.

sdk/messaging/azservicebus/admin/admin_client_queue.go
```diff
@@ -271,39 +273,49 @@ type QueuePropertiesPager interface {
...
+type QueueItem struct {
+	QueueName string
+	QueueProperties
 }
 
 // ListQueues lists queues.
-func (ac *AdminClient) ListQueues(options *ListQueuesOptions) QueuePropertiesPager {
-	var pageSize int
-	var skip int
+func (ac *Client) ListQueues(options *ListQueuesOptions) QueueItemPager {
```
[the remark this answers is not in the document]
richardpark-msft: I agree with you here. However, that interface was taken from the guidelines:

https://azure.github.io/azure-sdk/golang_introduction.html#golang-pagination-pagers-interface-page

I didn't even think to question it because I remembered our discussion about making things easier to mock but I can remove it if we don't think the guideline is right.

sdk/messaging/azservicebus/message.go
```diff
@@ -14,53 +15,81 @@ import (
...
-		ApplicationProperties map[string]interface{}
-		Format                uint32
+	// deferred indicates we received it using ReceiveDeferredMessages. These messages
+	// will still go through the normal Receiver.Settle functions but internally will
+	// always be settled with the management link.
+	deferred bool
+}
+
+// Body returns the body for this received message.
+// If the body not compatible with ReceivedMessage this function will return an error.
```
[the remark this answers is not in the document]
richardpark-msft: I'll clarify it when I actually add in AMQP support which I'm doing now. I didn't want to add too much info here until then.

What I will eventually write is:

The body (in an AMQP message) can be in three separate spots - in the Value section, the Data section or in a Sequence section. Each section has a possible value that will not cleanly encode to a single []byte array.

Data -> can contain _multiple_ byte arrays
Value -> can contain AMQP primitives, etc..
Sequence -> can be a sequence that's not bytes

sdk/messaging/azservicebus/sender.go
```diff
@@ -53,18 +45,18 @@ func (s *Sender) NewMessageBatch(ctx context.Context, options *MessageBatchOptio
 		return nil, err
 	}
 
-	maxBytes := int(sender.MaxMessageSize())
+	maxBytes := int32(sender.MaxMessageSize())
```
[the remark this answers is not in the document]
richardpark-msft: Let me check what the actual native type is. It might be that I need to go the other way, even if, in reality, the domain doesn't support a crazy large max message size.

sdk/messaging/azservicebus/sender.go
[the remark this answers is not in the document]
richardpark-msft: Luckily removed. @chlowell made me throw it in the🗑️

sdk/messaging/azservicebus/sender.go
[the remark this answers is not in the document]
richardpark-msft: So this is an AMQP field (straight from the attach frame) and it's typed to be a ulong, not a long.

* http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#:~:text=See%20subsection%202.6.7.-,max-message-size,-the%20maximum%20message
* http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-ulong

So we should probably follow suit there, but int32 is definitely wrong.

sdk/messaging/azservicebus/admin/admin_client_queue.go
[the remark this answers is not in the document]
richardpark-msft: Removed in https://github.com/Azure/azure-sdk-for-go/pull/16141

sdk/messaging/azservicebus/admin/main_test.go
richardpark-msft: Oh, yes they do.

sdk/messaging/azservicebus/messageSettler.go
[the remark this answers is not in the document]
richardpark-msft: @JeffreyRichter : There's a very common and unavoidable use case where the user will need to pass this message _back_ into one of our APIs to settle it. 

It seems like returning by value would require them to eat an efficiency loss of copying for no real gain.

Can I get an exemption here? :)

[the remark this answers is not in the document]
JeffreyRichter: That's fine. leave it as a pointer.

PR #16111, opened 2021-11-08
Code snippet generator tool and scripts

ArcturusZhang:
This PR introduces the `snippet-generator` tool, which replaces the code snippet in the markdown files from the real code snippet in our test cases surrounded by magic comments `// Snippet: name` and `// EndSnippet`


weshaggard:
We should look at trying to use our existing tool https://github.com/Azure/azure-sdk-tools/tree/main/tools/snippet-generator, I realize that one might be C# specific right now but it is worth trying to share and generalize it across languages so we don't have a ton of one-off tools for this.

If we for some reason we do still need a tool specific to go we should still consider putting that in the azure-sdk-tools repo.

ArcturusZhang:
> We should look at trying to use our existing tool https://github.com/Azure/azure-sdk-tools/tree/main/tools/snippet-generator, I realize that one might be C# specific right now but it is worth trying to share and generalize it across languages so we don't have a ton of one-off tools for this.
> 
> If we for some reason we do still need a tool specific to go we should still consider putting that in the azure-sdk-tools repo.

Yes, the tool was DotNet specific therefore we cannot directly use that.
And the biggest concern is that would pipelines be guaranteed to have the proper environment to invoke dotnet tools? For instance, in our Go SDK pipelines, I could be sure that inside our agent, it is fine to install a tool written in Go. But I am not sure if the agent has dotnet installed (and if it is not proper to install dotnet in an agent in GO SDK)

In the meantime, let me try make some modification to that tool to generalize it to serve more language.
I would like to keep this PR here - if I managed to generalize that tool, we could use this PR to introduce the script that consumes the updated Snippet-Generator.

ArcturusZhang:
> We should look at trying to use our existing tool https://github.com/Azure/azure-sdk-tools/tree/main/tools/snippet-generator, I realize that one might be C# specific right now but it is worth trying to share and generalize it across languages so we don't have a ton of one-off tools for this.
> 
> If we for some reason we do still need a tool specific to go we should still consider putting that in the azure-sdk-tools repo.

Well, unfortunately the `snippet-generator` in `azure-sdk-tools` is using `CSharpSyntaxTree` which makes it impossible to generalize to other languages: https://github.com/Azure/azure-sdk-tools/blob/1d16c6a27a475a42c1b0506463920be36813ea1f/tools/snippet-generator/Azure.Sdk.Tools.SnippetGenerator/DirectoryProcessor.cs#L237

weshaggard:
>And the biggest concern is that would pipelines be guaranteed to have the proper environment to invoke dotnet tools? For instance, in our Go SDK pipelines, I could be sure that inside our agent, it is fine to install a tool written in Go. But I am not sure if the agent has dotnet installed (and if it is not proper to install dotnet in an agent in GO SDK)

Our agents already have dotnet tools installed and there are a number of scripts/tools in the go engineering system that already depend on dotnet and I think that is a reasonable dependency. Just because the repo is primarily about go does not mean it cannot or should not depend on other tools. We need to be very careful about duplicating all our tools and helpers in go when we already have a set of shared tools implemented in a common language (PS, TS, dotnet, python whatever). 

We should only use language specific tools where it makes the most sense to do so. For example most languages have a parser for their particular language but not necessarily others so in that case it might make sense to use that specific language. 

>Well, unfortunately the snippet-generator in azure-sdk-tools is using CSharpSyntaxTree which makes it impossible to generalize to other languages:

While the reading/writing of the language specific files might be language specific the parsing and handling of all the MD files should be common and if we can it would be ideal to share as much as possible. 

I have not personally dug deep on this particular tool so I don't have a specific recommendation but my primary point is we need to stop duplicating all these tools in every language as that simply doesn't scale from a maintenance point of view and it also leads to a lot of inconsistencies in handling common things which we want to avoid. We should look to reuse, with potentially slight modifications, before we jump to reimplementing from scratch.

ArcturusZhang:
> > And the biggest concern is that would pipelines be guaranteed to have the proper environment to invoke dotnet tools? For instance, in our Go SDK pipelines, I could be sure that inside our agent, it is fine to install a tool written in Go. But I am not sure if the agent has dotnet installed (and if it is not proper to install dotnet in an agent in GO SDK)
> 
> Our agents already have dotnet tools installed and there are a number of scripts/tools in the go engineering system that already depend on dotnet and I think that is a reasonable dependency. Just because the repo is primarily about go does not mean it cannot or should not depend on other tools. We need to be very careful about duplicating all our tools and helpers in go when we already have a set of shared tools implemented in a common language (PS, TS, dotnet, python whatever).
> 
> We should only use language specific tools where it makes the most sense to do so. For example most languages have a parser for their particular language but not necessarily others so in that case it might make sense to use that specific language.
> 
> > Well, unfortunately the snippet-generator in azure-sdk-tools is using CSharpSyntaxTree which makes it impossible to generalize to other languages:
> 
> While the reading/writing of the language specific files might be language specific the parsing and handling of all the MD files should be common and if we can it would be ideal to share as much as possible.
> 
> I have not personally dug deep on this particular tool so I don't have a specific recommendation but my primary point is we need to stop duplicating all these tools in every language as that simply doesn't scale from a maintenance point of view and it also leads to a lot of inconsistencies in handling common things which we want to avoid. We should look to reuse, with potentially slight modifications, before we jump to reimplementing from scratch.

Make sense. Digging deeper into the tool, the CSharpSyntaxTree related things are meant to format the code - therefore I could try to generalize the tool to make sure it could be used on Go SDK

Issue #16086, opened 2021-11-04
InsertMerge/InsertReplace fails if record doesn't exist already in aztables


kierxn:
### Bug Report

- Package in question: `github.com/Azure/azure-sdk-for-go/sdk/data/aztables`

- SDK version e.g. `master`, `latest`, `18.1.0`: `github.com/Azure/azure-sdk-for-go/sdk/data/aztables v0.3.0`
- output of `go version`: `go version go1.17.2 windows/amd64`

- What happened?
Trying to batch insert records with `aztables.InsertMerge`/`aztables.InsertReplace` will always return a `Code: ResourceNotFound, Message: The specified resource does not exist.` error.

- What did you expect or want to happen?
I would expect this to insert my record or replace/merge if it exists already.

- How can we reproduce it?
Worth noting that this fails with both InsertMerge and InsertReplace.
```Go
func main() {
	service, _ := aztables.NewServiceClientFromConnectionString(os.Getenv("CONN_STR"), nil)
	client := service.NewClient(os.Getenv("TABLE_NAME"))

	batch := make([]aztables.TransactionAction, 0)

	entity1 := aztables.EDMEntity{
		Entity: aztables.Entity{
			PartitionKey: "p1",
			RowKey:       "r1",
		},
	}
	marshalled, _ := json.Marshal(entity1)

	batch = append(batch, aztables.TransactionAction{
		ActionType: aztables.InsertMerge,
		Entity: marshalled,
	})

	resp, err := client.SubmitTransaction(context.Background(), batch, nil)

	if err != nil {
		panic(err)
	}
	fmt.Println(resp)
}
```

- Anything we should know about your environment.
N/A

seankane-msft:
Hi @kierxn apologies on the bug and thank you for reporting it. I fixed the bug in #16114 just now, barring approval from the team this should go out tomorrow and fix your issue. If you'd like to review the PR I added a test case for this scenario to confirm that it is the same as your issue

seankane-msft:
This has been fixed in `v0.4.0` which is now available in a beta release [here](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/aztables).

PR #16114, opened 2021-11-08
[Tables] Fixes Transactional Batch for insert merge/replace entity.

seankane-msft:
Fixes customer bug #16086


richardpark-msft:
Might want to get in the habit of removing all the PR description boilerplate otherwise it looks like you didn't fill it out :)

richardpark-msft:
Similarly, having a PR title with some more detail would be good as well. (still reviewing)

sdk/data/aztables/transactional_batch.go
```diff
@@ -301,20 +301,24 @@ func (t *Client) generateEntitySubset(transactionAction *TransactionAction, writ
...
-		transactionAction.IfMatch = &star
-	}
+	// if transactionAction.IfMatch == nil {
+	// 	star := azcore.ETagAny
+	// 	transactionAction.IfMatch = &star
+	// }
 
 	switch transactionAction.ActionType {
 	case Delete:
+		ifMatch := "*"
```
richardpark-msft, line 311: We have that fancy azcore.ETagAny constant for this, right?

sdk/data/aztables/transactional_batch.go
```diff
@@ -301,20 +301,24 @@ func (t *Client) generateEntitySubset(transactionAction *TransactionAction, writ
 		return fmt.Errorf("entity properties must contain a %s property", rowKey)
 	}
 	// Consider empty ETags as '*'
-	if transactionAction.IfMatch == nil {
-		star := azcore.ETagAny
-		transactionAction.IfMatch = &star
-	}
+	// if transactionAction.IfMatch == nil {
```
RickWinter, line 304: Remove the commented code

richardpark-msft (commented):
Some questions.

sdk/data/aztables/client.go
```diff
@@ -528,10 +528,26 @@ func (t *Client) UpdateEntity(ctx context.Context, entity []byte, options *Updat
...
 	case MergeEntity:
-		resp, err := t.client.MergeEntity(ctx, generated.Enum1Three0, t.name, partKey, rowkey, options.toGeneratedMergeEntity(mapEntity), &generated.QueryOptions{})
+		resp, err := t.client.MergeEntity(
+			ctx,
+			generated.Enum1Three0,
+			t.name,
+			partKey,
+			rowkey,
+			options.toGeneratedMergeEntity(mapEntity),
+			&generated.QueryOptions{},
```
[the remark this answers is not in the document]
seankane-msft: yes unfortunately you have to pass in a pointer to a defined struct.

sdk/data/aztables/transactional_batch.go
seankane-msft: Yes. The generated code takes a string, not the azcore.ETag type so this was slightly simpler.

sdk/data/aztables/transactional_batch.go
```diff
@@ -346,7 +348,7 @@ func (t *Client) generateEntitySubset(transactionAction *TransactionAction, writ
 			entity[partitionKey].(string),
 			entity[rowKey].(string),
 			opts,
-			qo,
+			&generated.QueryOptions{},
```
chlowell, line 351: Why does `InsertMerge` need different `QueryOptions`?

chlowell (approved):

sdk/data/aztables/transactional_batch.go
seankane-msft: The `qo` passed an `ETag` of `*`, which returns a ResourceNotFoundError if the entity does not already exist. Passing the empty struct will not send that header at all which creates the entity if it does not exist, which is the desired behavior.

PR #16120, opened 2021-11-09
Mock test for mgmt plane pipeline

tadelesh:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -121,6 +121,29 @@ stages:
         displayName: "Show Docker Logs"
         condition: succeededOrFailed()
 
+    - ${{ if startsWith(parameters.ServiceDirectory, 'resourcemanager') }}:
+      - job: MgmtMockTest
+        displayName: Mock Test For Mgmt Plane
+        variables:
+          - template: ../variables/globals.yml
```
weshaggard, line 128:
```suggestion
          - template: /eng/pipelines/templates/variables/globals.yml
```
Better for these templates to use full path so that they don't break when things move around.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -121,6 +121,29 @@ stages:
         displayName: "Show Docker Logs"
         condition: succeededOrFailed()
 
+    - ${{ if startsWith(parameters.ServiceDirectory, 'resourcemanager') }}:
+      - job: MgmtMockTest
```
weshaggard, line 125: It is probably worth creating a separate job template for this and having that imported based on the condition. Maybe move all of into mgmt-mock-test.yml and make that job template instead of step template.

tadelesh: Good idea. I've reorg the files. BTW, it seems some template under templates/jobs belong to stages. Shall we need to move it properly at some point.

weshaggard: It is just a convention so it isn't necessary to move exactly but it is a good idea to keep the templates split into the correct folder.

tadelesh:
@weshaggard Could you have to approve this PR? We'll enable the 'Mock Test For Mgmt Plane' as a optional job which can give the mock test coverage report but won't block the release when failure by any problem. I've tried with some test. You can see the success [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1194863&view=results) and [failure](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1194864&view=results) here.

changlong-liu (approved):
LTGM

weshaggard (approved):

PR #16125, opened 2021-11-09
[KeyVault] Update azsecrets authentication policy

seankane-msft:
* Update client authentication policy to use the `internal.NewKeyVaultChallengePolicy`
* Add recordings for new auth policy

### DIFF
--- a/sdk/keyvault/azsecrets/CHANGELOG.md
+++ b/sdk/keyvault/azsecrets/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.2.1 (Unreleased)
 
 ### Features Added
-* Uses a challenge authentication policy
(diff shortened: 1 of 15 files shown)


sdk/keyvault/azsecrets/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.2.1 (Unreleased)
 
 ### Features Added
+* Uses a challenge authentication policy
```
chlowell, line 6: I think this would end up in customer-facing release notes. From a customer's perspective the feature is something like 
```suggestion
* Clients can now connect to Key Vaults in any cloud
```

chlowell (approved):

PR #16127, opened 2021-11-09
Add error hint for common version const format mistake

benbp:


seankane-msft (approved):

RickWinter (approved):

weshaggard (approved):
No issue with adding this but I will say I given our loose model and lack of consistency of this version info there are likely other errors lurking in our future as well. For example I'm waiting for us to return the wrong version because we find multiple version matches in the files.

benbp:
> No issue with adding this but I will say I given our loose model and lack of consistency of this version info there are likely other errors lurking in our future as well. For example I'm waiting for us to return the wrong version because we find multiple version matches in the files.

Yeah I kind of expect this to involve into a compendium of common format errors and a corresponding hint to fix each one.

PR #16150, opened 2021-11-11
[Core] bumping internal to 0.8.2

seankane-msft:
Bumps version of internal for latest changes. This allows all our client/mgmt plane packages to use the latest versions of internal as well.

### DIFF
--- a/sdk/azcore/go.mod
+++ b/sdk/azcore/go.mod
@@ -1,7 +1,7 @@
 module github.com/Azure/azure-sdk-for-go/sdk/azcore
 
 require (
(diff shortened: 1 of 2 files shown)


chlowell:
Forgive a newbie gopher question, but is this necessary? I would expect Go to consider 0.8.2 compatible with 0.8.1, so given a requirement for each, Go should choose 0.8.2 without complaint, and that should be okay so long as 0.8.2 actually is compatible.

jhendrixMSFT:
Your hypothesis is correct.  This is explained in the [version selection(https://github.com/golang/go/wiki/Modules#version-selection) section of the modules wiki.

PR #16151, opened 2021-11-11
[Internal] defer to PlaybackMode, try to find proxy cert in eng/common

seankane-msft:
- [ ] If `AZURE_RECORD_MODE` environment is not set, defer to Playback
- [ ] If the `PROXY_CERT` environment is not found, try to find it in the `eng/common/testproxy/dotnet-devcert.crt` based on the GOPATH.


sdk/internal/recording/recording.go
```diff
@@ -624,11 +626,37 @@ func GetRecordingId(t *testing.T) string {
...
+	goPath, ok := os.LookupEnv("GOPATH")
+	if !ok {
+		goPath, err = os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+	}
+
+	goPath = filepath.Join(goPath, "src", "github.com", "Azure", "azure-sdk-for-go", "eng", "common", "testproxy", "dotnet-devcert.crt")
+	return goPath, nil
```
chlowell, line 652: If I'm reading this right, when `goPath` doesn't exist, we return the error from `ioutil.ReadFile` directly. I wonder we should add context to that message so the reader doesn't wonder what that file read was for.

sdk/internal/recording/recording.go
```diff
@@ -624,11 +626,37 @@ func GetRecordingId(t *testing.T) string {
...
+	fileLocation, ok := os.LookupEnv("PROXY_CERT")
+	if ok {
+		return fileLocation, nil
+	}
+
+	// Couldn't be found, lets try to find it from the GOPATH location
+	var err error
+	goPath, ok := os.LookupEnv("GOPATH")
+	if !ok {
+		goPath, err = os.UserHomeDir()
```
chlowell, line 645: Is this a conventional GOPATH? If not, I think it's okay to require devs to set GOPATH or PROXY_CERT.

sdk/internal/recording/recording.go
```diff
@@ -624,11 +626,37 @@ func GetRecordingId(t *testing.T) string {
 }
 
 func GetRecordMode() string {
+	if !(recordMode == RecordingMode || recordMode == PlaybackMode || recordMode == LiveMode) {
+		recordMode = PlaybackMode
```
chlowell, line 630: What do you think of doing this in `init`? Then you get the desired default behavior and don't have to consider calling patterns around this package.

sdk/internal/recording/recording.go
seankane-msft: Yes, if it is not set Go defaults the home directory: https://golang.org/doc/gopath_code#GOPATH

sdk/internal/recording/recording.go
seankane-msft: We only return the error if the `os.UserHomeDir()` cannot be found, which is if the `$HOME` envvar is not set

sdk/internal/recording/recording.go
```diff
@@ -439,7 +439,37 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+	}
+
+	localFile, err := findProxyCertLocation()
+	if err != nil {
+		log.Println("Could not find the PROXY_CERT environment variable and was unable to locate the path in eng/common")
+	}
+
+	rootCAs, err = x509.SystemCertPool()
+	if err != nil && strings.Contains(err.Error(), "system root pool is not available on Windows") {
```
chlowell, line 455: Suggest doing this conditionally depending on OS rather than string matching the error message (more robust that way)

sdk/internal/recording/recording.go
```diff
@@ -438,7 +439,37 @@ var modeMap = map[RecordMode]recorder.Mode{
...
+		log.Println("could not create a system cert pool")
+	}
+
+	cert, err := ioutil.ReadFile(localFile)
+	if err != nil {
+		log.Printf("could not read file set in PROXY_CERT variable at %s.\n", localFile)
+	}
+
+	if ok := rootCAs.AppendCertsFromPEM(cert); !ok {
+		log.Println("no certs appended, using system certs only")
```
chlowell, line 467: What happens when `cert` is already in the pool?

sdk/internal/recording/recording.go
```diff
@@ -438,7 +439,37 @@ var modeMap = map[RecordMode]recorder.Mode{
 	Playback: recorder.ModeReplaying,
 }
 
-var recordMode = os.Getenv("AZURE_RECORD_MODE")
+func init() {
+	recordMode = os.Getenv("AZURE_RECORD_MODE")
+	if !(recordMode == RecordingMode || recordMode == PlaybackMode || recordMode == LiveMode) {
+		log.Printf("AZURE_RECORD_MODE was not understood, options are %s, %s, or %s Received: %v. Defaulting to playback", RecordingMode, PlaybackMode, LiveMode, recordMode)
```
chlowell, line 445: I think we should set a default only when AZURE_RECORD_MODE has no value. When it has an unexpected one we know the user wants specific behavior but not what behavior. Erroring out so the user can clarify seems better to me than proceeding.

sdk/internal/recording/recording.go
seankane-msft: I can't find anything on duplicated certificates so I'm not sure what to assume here. I would assume having the same certificate twice is fine?

chlowell: That's how I think it *should* work but don't know how it actually works, so I was asking out of curiosity. I think it's fine to start with this and adjust if necessary.

sdk/internal/recording/recording.go
chlowell: Now that you've added `init()`, `recordMode` will have a value when this method executes

sdk/internal/recording/recording.go
```diff
@@ -624,45 +659,34 @@ func GetRecordingId(t *testing.T) string {
...
-	if ok := rootCAs.AppendCertsFromPEM(cert); !ok {
-		t.Log("No certs appended, using system certs only")
+		goPath, err = os.UserHomeDir()
+		if err != nil {
+			return "", fmt.Errorf("could not find the home directory using `os.UserHomeDir()`, %s", err.Error())
+		}
 	}
 
-	return rootCAs, nil
+	goPath = filepath.Join(goPath, "src", "github.com", "Azure", "azure-sdk-for-go", "eng", "common", "testproxy", "dotnet-devcert.crt")
```
chlowell, line 684: Forgive the newbie gopher question but would this path typically exist in the workspace of an SDK developer? It doesn't exist in mine, and now I wonder whether I've taken an unconventional approach in setting up my workspace

jhendrixMSFT: I don't think we can make assumptions about the path, and given that GOPATH is going away, shouldn't rely on it either.

seankane-msft: When is GOPATH slated to be removed @jhendrixMSFT ?

jhendrixMSFT: I don't know when, and I suppose it's likely to be here from some time.

With the introduction of modules, there's no longer a need to explicitly set it or place one's source code inside it.  For example, my GOPATH from `go env`.

```
set GOPATH=C:\Users\jhendrix\go
```

I don't place any of my source code in this directory, it's all under various subdirectories in `c:\git`.

seankane-msft: That's unfortunate for this scenario. I think the best case is to now remove all this and only search in the environment variable, agreed @jhendrixMSFT ?

jhendrixMSFT: If this is needed to find files within the tree you could execute this to get the repo root.
```
git rev-parse --show-toplevel
```

Returns something like `C:/git/Azure/go-amqp` or whatever.

I'll also say this is why having test assets under `<package>/testdata` makes sense as you can just use a relative path (I get this scenario is a bit odd).

seankane-msft: I'll try using the git command to do so, if that doesn't work well I'll remove this bit of code. We would have to copy the cert to every package/testdata folder which is more work than asking everyone to set the environment variable

sdk/internal/recording/recording.go
```diff
@@ -627,42 +664,23 @@ func GetRecordMode() string {
...
+	if ok {
+		return fileLocation, nil
 	}
 
-	cert, err := ioutil.ReadFile(localFile)
+	out, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
 	if err != nil {
-
-		return nil, err
+		log.Print("Could not find PROXY_CERT environment vaiable or toplevel of git repository, please set PROXY_CERT environment variable to location of certificate found in eng/common/testproxy/dotnet-devcert.crt")
```
chlowell, line 675:
```suggestion
		log.Print("Could not find PROXY_CERT environment variable or toplevel of git repository, please set PROXY_CERT to location of certificate found in eng/common/testproxy/dotnet-devcert.crt")
```

chlowell (approved):

PR #16207, opened 2021-11-17
feat: add check release to auto trigger internal release job

tadelesh:

As there are too many manual trigger works for SDK release, I added a release check stage before the release pipeline. The stage will check whether there is a new release and to trigger or skip release stage then.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-go-release.yml
+++ b/eng/pipelines/templates/jobs/archetype-go-release.yml
@@ -4,67 +4,86 @@ parameters:
   ServiceDirectory: ''
 
 stages:
-  - ${{if and(eq(variables['Build.Reason'], 'Manual'), eq(variables['System.TeamProject'], 'internal'))}}:
-    - stage: Release
-      displayName: 'Release: ${{ parameters.ServiceDirectory }}'
-      dependsOn: ${{ parameters.DependsOn }}
-      condition: and(succeeded(), ne(variables['SetDevVersion'], 'true'), ne(variables['Skip.Release'], 'true'), ne(variables['Build.Repository.Name'], 'Azure/azure-sdk-for-go-pr'))
-      jobs:
-        - deployment: TagRepository
-          displayName: "Create release tag"
-          condition: and(succeeded(), ne(variables['Skip.TagRepository'], 'true'))
-          environment: github
+  - stage: CheckRelease
+    displayName: 'Check Release: ${{ parameters.ServiceDirectory }}'
+    dependsOn: ${{ parameters.DependsOn }}
+    condition: and(succeeded(), ne(variables['SetDevVersion'], 'true'), ne(variables['Skip.Release'], 'true'), ne(variables['Build.Repository.Name'], 'Azure/azure-sdk-for-go-pr'))
+    jobs:
+      - job: CheckReleaseJob
+        displayName: "Check whether need to release"
+        steps:
+          - task: PowerShell@2
+            name: Verify
+            inputs:
+              filePath: ./eng/scripts/Verify-NeedToRelease.ps1
+              arguments: >
+                -PackageName 'sdk/${{ parameters.ServiceDirectory }}'
+                -ServiceDirectory '${{ parameters.ServiceDirectory }}'
+                -repoId Azure/azure-sdk-for-go
+                -workingDirectory $(System.DefaultWorkingDirectory)
+              pwsh: true
+            env:
+              GH_TOKEN: $(azuresdk-github-pat)
+  - stage: Release
+    displayName: 'Release: ${{ parameters.ServiceDirectory }}'
+    dependsOn: CheckRelease
+    condition: and(succeeded(), eq(dependencies.CheckRelease.outputs['CheckReleaseJob.Verify.NeedToRelease'], 'true'))
+    jobs: 
+      - deployment: TagRepository
+        displayName: "Create release tag"
+        condition: and(succeeded(), eq(variables['NeedToRelease'], 'true'), ne(variables['Skip.TagRepository'], 'true'))
+        environment: github
 
-          pool:
-            name: azsdk-pool-mms-ubuntu-2004-general
-            vmImage: MMSUbuntu20.04
+        pool:
+          name: azsdk-pool-mms-ubuntu-2004-general
+          vmImage: MMSUbuntu20.04
 
-          strategy:
-            runOnce:
-              deploy:
-                steps:
-                  - checkout: self
-                  - template: /eng/common/pipelines/templates/steps/retain-run.yml
-                  - template: /eng/common/pipelines/templates/steps/verify-changelog.yml
-                    parameters:
-                      PackageName: 'sdk/${{ parameters.ServiceDirectory }}'
-                      ForRelease: true
-                  - task: PowerShell@2
-                    displayName: 'Verify no replace directives in go.mod file'
-                    inputs:
-                      targetType: 'filePath'
-                      filePath: ./eng/scripts/validate_go_mod.ps1
-                      arguments: '${{ parameters.ServiceDirectory }}'
-                      pwsh: true
-                  - template: /eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
-                    parameters:
-                      ArtifactLocation: $(Build.SourcesDirectory)/sdk/${{ parameters.ServiceDirectory }}
-                      ReleaseSha: $(Build.SourceVersion)
-                      RepoId: Azure/azure-sdk-for-go
-                      WorkingDirectory: $(System.DefaultWorkingDirectory)
+        strategy:
+          runOnce:
+            deploy:
+              steps:
+                - checkout: self
+                - template: /eng/common/pipelines/templates/steps/retain-run.yml
+                - template: /eng/common/pipelines/templates/steps/verify-changelog.yml
+                  parameters:
+                    PackageName: 'sdk/${{ parameters.ServiceDirectory }}'
+                    ForRelease: true
+                - task: PowerShell@2
+                  displayName: 'Verify no replace directives in go.mod file'
+                  inputs:
+                    targetType: 'filePath'
+                    filePath: ./eng/scripts/validate_go_mod.ps1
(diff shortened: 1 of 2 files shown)


eng/scripts/Verify-NeedToRelease.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+$apiUrl = "https://api.github.com/repos/$repoId"
+Write-Host "Using API URL $apiUrl"
+
+# VERIFY CHANGELOG
+$PackageProp = Get-PkgProperties -PackageName $PackageName -ServiceDirectory $ServiceDirectory
+$changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $PackageProp.ChangeLogPath
+$changeLogEntry = $changeLogEntries[$PackageProp.Version]
+
+if (!$changeLogEntry)
```
weshaggard, line 17: Why do we want to skip releasing if there is no changelog? Why wouldn't we just error?

weshaggard:
@tadelesh can you give me more context on what we are trying to fix here? If we don't want to release a given package we should opt it out of release as opposed to ignoring what looks like could be real errors in other packages we do want to release.

eng/scripts/Verify-NeedToRelease.ps1
tadelesh: This check release step is just used to check whether next release stage need to trigger to avoid manual trigger step. Every release validation will be done in the next release stage.

tadelesh:
> @tadelesh can you give me more context on what we are trying to fix here? If we don't want to release a given package we should opt it out of release as opposed to ignoring what looks like could be real errors in other packages we do want to release.

I need to trigger an internal release pipeline manually if I need to do the SDK release for now. Then after the build stage in pipeline pass, I need to do another manual approval to continue the release stage. As we release SDK with separate RP packages, there are so many manual clicks. So I want to setup a new stage to trigger the internal release pipeline automatically. The trigger condition is: if a package is under development with the changelog in UNRELEASED status, then no need to trigger; if a package's latest version is ready with proper changelog, and such version has not released with tag, then to trigger. So the new added `check release` stage and script will do some check with the changelog and tag, then next `release` stage can continue or skip automatically according to the check result. This stage is just a helper before the release stage and will not influence any existed release validation.

weshaggard:
If the goal is to cut down on manual approvals, I don't think this is the correct approach. With the current changes it will actually potentially skip out on packages we want to ship but have errors that weren't discovered in other ways. I'm actually unsure how your current changes even eliminate the manual steps of clicking the approve button on the packages you want to release. 

Just in case you aren't aware the thing that causes the manual approval check is the `environment: github` part of the deployment stage. If we want to reduce the number of approvals we would need to make some adjustments to that logic. We might be able to potentially setup a special pipeline or maybe a pipeline variable/parameter that someone could set to automatically do the release after build. However we need to be careful with doing that as only some folks have access to approve releases so whatever we setup will need to use the same SG for protecting the auto-release logic. If you want to setup some time with me next week I'd be happy to talk through some options.

tadelesh:
> If the goal is to cut down on manual approvals, I don't think this is the correct approach. With the current changes it will actually potentially skip out on packages we want to ship but have errors that weren't discovered in other ways. I'm actually unsure how your current changes even eliminate the manual steps of clicking the approve button on the packages you want to release.
> 
> Just in case you aren't aware the thing that causes the manual approval check is the `environment: github` part of the deployment stage. If we want to reduce the number of approvals we would need to make some adjustments to that logic. We might be able to potentially setup a special pipeline or maybe a pipeline variable/parameter that someone could set to automatically do the release after build. However we need to be careful with doing that as only some folks have access to approve releases so whatever we setup will need to use the same SG for protecting the auto-release logic. If you want to setup some time with me next week I'd be happy to talk through some options.

I think you misunderstand what I've done. When I want to do a release for some SDK. I need to find the SDK's specific internal pipeline in the ado and trigger it manually. Then the pipeline will do build and test first. After that, I need to do the approval to continue the pipeline. So there are TWO manual work here (trigger and approval). What I've done is to cut the first manual trigger work. Any following release step including the approval step will not change.

weshaggard:
OK I think I see. You removed ` ${{if and(eq(variables['Build.Reason'], 'Manual'), eq(variables['System.TeamProject'], 'internal'))}}:` That means any CI triggered build you can go in an approve the release for it. There are a couple issues with this approach:
- You will need to keep the internal check here as the public builds will not have the necessary permissions to run the follow-up steps.
- With this approach if no one decides to release the package then the pipeline will be in a state where the approval will timeout and then fail so the builds will all look like they have failed. 

We will definitely have to fix the first issue but we might be able to live with the second issue but we would need to test to see what that looks like.

@benbp @seankane-msft what do you think about this release process for go packages?

eng/scripts/Verify-NeedToRelease.ps1
```diff
@@ -0,0 +1,42 @@
...
+{
+  Write-Host "Changelog not existed for package: $PackageName, version: $($PackageProp.Version)."
```
weshaggard, line 19:
```suggestion
  Write-Host "Changelog does not exist for package: $PackageName, version: $($PackageProp.Version)."
```

eng/scripts/Verify-NeedToRelease.ps1
```diff
@@ -0,0 +1,42 @@
...
+  Write-Output "##vso[task.setvariable variable=NeedToRelease;isOutput=true]false"
+  return
+}
+
+if ([System.String]::IsNullOrEmpty($changeLogEntry.ReleaseStatus) -or $changeLogEntry.ReleaseStatus -eq $CHANGELOG_UNRELEASED_STATUS)
+{
+  Write-Host "Changelog not in release status for package: $PackageName, version: $($PackageProp.Version)."
```
weshaggard, line 26:
```suggestion
  Write-Host "Changelog is not in release status for package: $PackageName, version: $($PackageProp.Version)."
```

tadelesh:
> OK I think I see. You removed ` ${{if and(eq(variables['Build.Reason'], 'Manual'), eq(variables['System.TeamProject'], 'internal'))}}:` That means any CI triggered build you can go in an approve the release for it. There are a couple issues with this approach:
> 
> * You will need to keep the internal check here as the public builds will not have the necessary permissions to run the follow-up steps.
> * With this approach if no one decides to release the package then the pipeline will be in a state where the approval will timeout and then fail so the builds will all look like they have failed.
> 
> We will definitely have to fix the first issue but we might be able to live with the second issue but we would need to test to see what that looks like.
> 
> @benbp @seankane-msft what do you think about this release process for go packages?

For the first issue, there is an internal check in the outside template (archetype-sdk-client.yml: ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))}}). I removed the inner check (archetype-go-release.yml: ${{if and(eq(variables['Build.Reason'], 'Manual'), eq(variables['System.TeamProject'], 'internal'))}}) for deduplication. Btw, I've fixed the log grammar. AFAIK, mgmt. plane of JS SDK has the same requirement.

eng/scripts/Verify-NeedToRelease.ps1
```diff
@@ -0,0 +1,42 @@
...
+Write-Host "Using API URL $apiUrl"
+
+# VERIFY CHANGELOG
+$PackageProp = Get-PkgProperties -PackageName $PackageName -ServiceDirectory $ServiceDirectory
+$changeLogEntries = Get-ChangeLogEntries -ChangeLogLocation $PackageProp.ChangeLogPath
+$changeLogEntry = $changeLogEntries[$PackageProp.Version]
+
+if (!$changeLogEntry)
+{
+  Write-Host "Changelog does not existed for package: $PackageName, version: $($PackageProp.Version)."
```
weshaggard, line 19:
```suggestion
  Write-Host "Changelog does not exist for package: $PackageName, version: $($PackageProp.Version)."
```

weshaggard (approved):
I'm good with these changes I would like to get someone like @seankane-msft to also examine the changes.

tadelesh:
> I'm good with these changes I would like to get someone like @seankane-msft to also examine the changes.

Thanks @weshaggard . @benbp @seankane-msft Could you help to have a review?

seankane-msft (approved):

seankane-msft:
@tadelesh I'm not sure this does what we want it to do. I tried to release `internal` with [this](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1220229&view=logs&j=f4752989-c568-5678-80f2-0a67848bfa4d&t=d624614a-b70e-58c1-fab8-e0fa21bc7cc7) internal pipeline, but the "Create release tag" step was never run. Can you take a look at why this is happening?

eng/pipelines/templates/jobs/archetype-go-release.yml
```diff
@@ -4,67 +4,86 @@ parameters:
...
+            env:
+              GH_TOKEN: $(azuresdk-github-pat)
+  - stage: Release
+    displayName: 'Release: ${{ parameters.ServiceDirectory }}'
+    dependsOn: CheckRelease
+    condition: and(succeeded(), eq(dependencies.CheckRelease.outputs['CheckReleaseJob.Verify.NeedToRelease'], 'true'))
+    jobs: 
+      - deployment: TagRepository
+        displayName: "Create release tag"
+        condition: and(succeeded(), eq(variables['NeedToRelease'], 'true'), ne(variables['Skip.TagRepository'], 'true'))
```
weshaggard, line 34: Based on https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#use-outputs-in-a-different-stage

I suspect you need to get this value from the outputs similar to the other usage. Actually, do we need the condition here as well as on the stage? Perhaps this one can be removed. 

I think this is likely the reason for the issue that @seankane-msft is seeing.

tadelesh: Yes, this job condition is wrong and also redundant. I've removed it by this [PR](https://github.com/Azure/azure-sdk-for-go/pull/16355). Thanks @weshaggard @seankane-msft for your validation and correction.

PR #16208, opened 2021-11-17
[Internal] updating docs and adding samples for internal package

### LINKED ISSUE #16200: The code snippets in README of `recording` package is out-of-date
Since we have gotten some breaking changes in the `recording` package, the code snippets in [README.md](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/recording/README.md) is out-of-date. Could we have an update on that?

And maybe we need to use other mechanism to ensure that they are always up-to-date, like using an external tool, or using example files.

seankane-msft:
Fixes #16200 by updating documentation and adding example_test.go file for any future API surface changes.


sdk/internal/recording/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+package recording_test
+
+import (
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
+)
+
+func ExampleStart() {
```
ArcturusZhang, line 15: Could we also introduce an end-to-end example that shows how to use this framework?

PR #16210, opened 2021-11-18
[azservicebus] Making a separate version of amqp-common as the `internal` package for messaging

### LINKED ISSUE #15646: [amqp-common] Migrate Azure/azure-amqp-common-go into monorepo

richardpark-msft:
The messaging SDKs like Service Bus (and eventually EventHubs) share common code for authenticating using cbs/claims, and various things related to AMQP. This was all bundled up into a package located at github.com/Azure/azure-amqp-common-go. 

We don't actually need to do that anymore for SB, and it's desirable to move more code into the monorepo.

This is a pretty minimal copy - just what's latest plus some changes to make it build as an `internal` package.

Fixes #15646


sdk/messaging/internal/version.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+package internal
+
+const (
+	// Version is the semantic version of the library
+	Version = "3.1.1"
```
jhendrixMSFT, line 7: This should be `0.1.0` yes?

sdk/messaging/internal/uuid/uuid.go
```diff
@@ -0,0 +1,72 @@
+package uuid
```
jhendrixMSFT, line 1: We already have a `uuid` package in the core `internal` module.

jhendrixMSFT (changes requested):
Also need a LICENSE.txt file and doc.go

sdk/messaging/internal/uuid/uuid.go
richardpark-msft: Done.

sdk/messaging/internal/version.go
richardpark-msft: Yes

richardpark-msft:
> 
> 
> Also need a LICENSE.txt file and doc.go

Done :)

jhendrixMSFT (approved):

PR #16221, opened 2021-11-18
Update transport

seankane-msft:

### DIFF
--- a/sdk/internal/recording/recording.go
+++ b/sdk/internal/recording/recording.go
@@ -655,6 +655,30 @@ func getRootCas(t *testing.T) (*x509.CertPool, error) {
 	return rootCAs, nil
 }
 
+type customHttpClient struct {
+	defaultClient *http.Client
+	options       RecordingOptions
+	t             *testing.T
+}
+
+func (c customHttpClient) Do(req *http.Request) (*http.Response, error) {
+	c.options.ReplaceAuthority(c.t, req)
+	return c.defaultClient.Do(req)
+}
+
+func GetRoutingHTTPClient(t *testing.T, options *RecordingOptions) (*customHttpClient, error) {
+	c, err := GetHTTPClient(t)
+	if err != nil {
+		return nil, err
+	}
+
+	return &customHttpClient{
+		defaultClient: c,
+		options:       *options,
+		t:             t,
+	}, nil
+}
+
 func GetHTTPClient(t *testing.T) (*http.Client, error) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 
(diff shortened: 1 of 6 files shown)


sdk/internal/recording/recording_test.go
```diff
@@ -433,6 +433,38 @@ func TestStartStop(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+
+	require.NotNil(t, GetRecordingId(t))
+
+	err = Stop(t, nil)
+	require.NoError(t, err)
+
+	// Make sure the file is there
+	jsonFile, err := os.Open("./testdata/recordings/TestStartStopRoutingClient.json")
```
richardpark-msft, line 463: It looks like we're checking in this file so what is this test really showing?

Should the `testdata/recordings/TestStartStopRoutingClient.json` not be checked in? Or can you check that the file has actually been updated?

sdk/internal/recording/recording.go
```diff
@@ -655,6 +655,33 @@ func getRootCas(t *testing.T) (*x509.CertPool, error) {
...
+	options       RecordingOptions
+	t             *testing.T
+}
+
+func (c RecordingHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	c.options.ReplaceAuthority(c.t, req)
+	return c.defaultClient.Do(req)
+}
+
+func GetRoutingHTTPClient(t *testing.T, options *RecordingOptions) (*RecordingHTTPClient, error) {
```
richardpark-msft, line 669: Is this basically `NewRecordingHTTPClient`? It looks like it fits that pattern (I don't see 'Routing' anywhere in here so it feels like that nomenclature is outdated)

richardpark-msft (changes requested):
Curious if the recordings file should be checked in, and maybe a name change depending on what your intention was.

sdk/internal/recording/recording_test.go
seankane-msft: Updated to check that the URLs are changed appropriately.

sdk/internal/recording/recording.go
seankane-msft: Changed the name

sdk/internal/recording/recording_test.go
richardpark-msft: The checking part looks good.

How do you prove the file was written by your test and wasn't the older checked-in file?

seankane-msft: Removed the recording file from the git history. Now each test will create a new file and delete it in clean up.

richardpark-msft (approved):
Looks good!

PR #16226, opened 2021-11-19
Adding Smoketests to nightly runs

### LINKED ISSUE #16078: Update smoke tests to auto-generate dependency checks for all t2 SDKs
The goal for smoke tests is to catch and block on dependency issues before release.

Delete the legacy smoke tests for each language (applies to net, python, java).
Pull in all track 2 packages to a single project and install/compile (nightly versions, GA/beta versions).
Run the project and instantiate clients from each package, where useful (for detecting lazy loaded runtime dependency issues).

seankane-msft:
Closes #16078


seankane-msft:
@benbp and I's original plan was to create a `go.mod` file with all packages and a replace directive to point to the local version and run `go mod tidy` to verify the `go.mod` file is valid. Looking at the description of the [`go mod tidy`](https://golang.org/ref/mod#go-mod-tidy) command, I'm not confident this will work and we might have to include source code for each package and run `go build` to verify it.

tadelesh:
> @benbp and I's original plan was to create a `go.mod` file with all packages and a replace directive to point to the local version and run `go mod tidy` to verify the `go.mod` file is valid. Looking at the description of the [go mod tidy](https://golang.org/ref/mod#go-mod-tidy) command, I'm not confident this will work and we might have to include source code for each package and run `go build` to verify it.

If we just want to do the dependency check for all the SDK packages, I think it is no need to add all the example files. Just as my [PR](https://github.com/Azure/azure-sdk-for-go/pull/16265#issuecomment-978779801) implemented, you can use `_` to hold a unused import in the source file so all lib will be built. For the common packages like azcore or azidentity, we can not do the replacement in go.mod cause when breaking changes happen, all other reference packages need to do upgrade first.

seankane-msft:
> > @benbp and I's original plan was to create a `go.mod` file with all packages and a replace directive to point to the local version and run `go mod tidy` to verify the `go.mod` file is valid. Looking at the description of the [go mod tidy](https://golang.org/ref/mod#go-mod-tidy) command, I'm not confident this will work and we might have to include source code for each package and run `go build` to verify it.
> 
> If we just want to do the dependency check for all the SDK packages, I think it is no need to add all the example files. Just as my [PR](https://github.com/Azure/azure-sdk-for-go/pull/16265#issuecomment-978779801) implemented, you can use `_` to hold a unused import in the source file so all lib will be built. For the common packages like azcore or azidentity, we can not do the replacement in go.mod cause when breaking changes happen, all other reference packages need to do upgrade first. @seankane-msft it seems we can merge our PR into one. What do you think?

@tadelesh Yes I think we are doing the same work here in two different ways. You make a good point about azcore/azidentity, those two packages will have to be left off. 

Is the work you're doing replacing all packages with the local version or is it only checking one version? I think we want to do both but put the checks in different places. The first check should go in the definition [here](https://github.com/Azure/azure-sdk-for-go/blob/main/azure-pipelines.yml) and another that should go in the [Analyze step here](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/steps/analyze.yml). What are your thoughts?

ckairen:
> 

Yes, we are planning on checking one-version for the [release pipeline] (https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/jobs/archetype-sdk-client.yml), and testing on per-package, and the all-package-local-version one for the nightly run. But instead of [azure-pipeline] (https://github.com/Azure/azure-sdk-for-go/blob/main/azure-pipelines.yml), which is meant for track 1, we'll do that in [build-test] (https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/steps/build-test.yml),

tadelesh:
> > > @benbp and I's original plan was to create a `go.mod` file with all packages and a replace directive to point to the local version and run `go mod tidy` to verify the `go.mod` file is valid. Looking at the description of the [go mod tidy](https://golang.org/ref/mod#go-mod-tidy) command, I'm not confident this will work and we might have to include source code for each package and run `go build` to verify it.
> > 
> > 
> > If we just want to do the dependency check for all the SDK packages, I think it is no need to add all the example files. Just as my [PR](https://github.com/Azure/azure-sdk-for-go/pull/16265#issuecomment-978779801) implemented, you can use `_` to hold a unused import in the source file so all lib will be built. For the common packages like azcore or azidentity, we can not do the replacement in go.mod cause when breaking changes happen, all other reference packages need to do upgrade first. @seankane-msft it seems we can merge our PR into one. What do you think?
> 
> @tadelesh Yes I think we are doing the same work here in two different ways. You make a good point about azcore/azidentity, those two packages will have to be left off.
> 
> Is the work you're doing replacing all packages with the local version or is it only checking one version? I think we want to do both but put the checks in different places. The first check should go in the definition [here](https://github.com/Azure/azure-sdk-for-go/blob/main/azure-pipelines.yml) and another that should go in the [Analyze step here](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/steps/analyze.yml). What are your thoughts?

I'll do it in track2 single package's ci and replace with local path in go.mod for this package only. I've moved it to `analyze` stage for my PR. For each package, I think a nightly build task which will check the latest code is compatible with other packages' version is useful. But it might be no need to ensure all the packages' latest code is compatible with each other. Instead, I think check whether all packages have a version which can be compatible used with each other is more useful. It can be done with no replacements in go.mod.

eng/tools/smoketests/main.go
```diff
@@ -352,6 +350,76 @@ func BuildMainFile(root string) error {
...
+		}
+		return nil
+	})
+	return err
+}
+
+func searchFile(path string) error {
+	var envVarRegex *regexp.Regexp
+	if runtime.GOOS == "windows" {
+		envVarRegex = regexp.MustCompile(`(?m)os.LookupEnv(.*)\r\n`)
```
benbp, line 369: Is the newline match here necessary? I could see it excluding potential future scenarios like:

```
if val, ok = os.LookupEnv("foo"); !ok {
...
```

Why not match inside the parentheses+quotes instead?

```suggestion
		envVarRegex = regexp.MustCompile(`(?m)os.LookupEnv\("(.*)"\)`)
```

eng/tools/smoketests/main.go
```diff
@@ -352,6 +350,76 @@ func BuildMainFile(root string) error {
...
+	} else {
+		envVarRegex = regexp.MustCompile(`(?m)os.Getenv(.*)\n`)
+	}
+
+	Getenvs := envVarRegex.FindAllString(stringData, -1)
+	envVars = append(envVars, trimGetenvs(Getenvs)...)
+	return nil
+}
+
+func trimLookupEnvs(values []string) []string {
```
benbp, line 393: This function seems unnecessary if you match the regex more exactly.

benbp (approved):

PR #16229, opened 2021-11-19
[azservicebus] Start trimming out track 1 dependencies

### LINKED ISSUE #15979: [azservicebus] Remove dependency on track 1
I haven't dug into why we're depending on track 1 but we need to remove it as it should be unneeded for track 2.

It's possible this is dependent on removing some of our older dependencies. It doesn't look like I'm using it directly.

richardpark-msft:
- Removing dependency on the external azure-amqp-common-go, using the `messaging/internal` package instead
- Remove use of autorest/date

Fixes #15979


sdk/messaging/azservicebus/go.mod
```diff
@@ -3,15 +3,14 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
 go 1.16
 
 require (
-	github.com/Azure/azure-amqp-common-go/v3 v3.2.1
 	github.com/Azure/azure-sdk-for-go v51.1.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.20.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.12.0
-	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.1
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.2
+	github.com/Azure/azure-sdk-for-go/sdk/messaging/internal v0.0.0-20211118043544-abfa4f313a9c
```
jhendrixMSFT, line 10: When will this get tagged?

sdk/messaging/azservicebus/internal/test/test_helpers.go
```diff
@@ -11,7 +11,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/Azure/azure-amqp-common-go/v3/conn"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/internal/conn"
 	rm "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources"
```
jhendrixMSFT, line 15: FYI another track 1 dependency (see also next line).

sdk/messaging/azservicebus/internal/atom/models.go
```diff
@@ -256,10 +281,24 @@ type (
...
+
+func StringToTime(timeStr string) time.Time {
+	if timeStr == "" {
+		return time.Time{}
+	}
+
+	parsedTime, err := time.Parse(time.RFC3339, timeStr)
+
+	if err != nil {
+		return time.Time{}
```
jhendrixMSFT, line 300: This seems a bit dubious as it's ambiguous with the empty string.

sdk/messaging/azservicebus/internal/atom/models.go
```diff
@@ -54,9 +52,9 @@ type (
 		MessageCount                        *int64        `xml:"MessageCount,omitempty"`                        // MessageCount - The number of messages in the queue.
 		IsAnonymousAccessible               *bool         `xml:"IsAnonymousAccessible,omitempty"`
 		Status                              *EntityStatus `xml:"Status,omitempty"`
-		AccessedAt                          *date.Time    `xml:"AccessedAt,omitempty"`
-		CreatedAt                           *date.Time    `xml:"CreatedAt,omitempty"`
-		UpdatedAt                           *date.Time    `xml:"UpdatedAt,omitempty"`
+		AccessedAt                          string        `xml:"AccessedAt,omitempty"`
```
jhendrixMSFT, line 55: Does `*time.Time` not work here?  I thought the default format is RFC3339 which is what `StringToTime()` is doing.

sdk/messaging/azservicebus/internal/test/test_helpers.go
richardpark-msft: Got rid of it now (we're actually not using it!) and that removed a lot of the old stuff.

sdk/messaging/azservicebus/internal/atom/models.go
richardpark-msft: It's one of the required fields but, since it's just a read only time field, I figured we didn't have to kill the entire deserialization for it.

I have some live tests to ensure we don't get time.IsZero(), ,however.

sdk/messaging/azservicebus/go.mod
```diff
@@ -3,17 +3,12 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
...
 require (
-	github.com/Azure/azure-amqp-common-go/v3 v3.2.1
-	github.com/Azure/azure-sdk-for-go v51.1.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.20.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.12.0
-	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.1
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.2
+	github.com/Azure/azure-sdk-for-go/sdk/messaging/internal v0.0.0-20211118043544-abfa4f313a9c
 	github.com/Azure/go-amqp v0.16.4
 	github.com/Azure/go-autorest/autorest v0.11.18
```
jhendrixMSFT, line 11: This is also a track 1 dependency.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/15979

sdk/messaging/azservicebus/go.mod
richardpark-msft: (offline discussion had)

We're going to leave this untagged for now. There are some benefits that seem to apply to this package. Mainly, since this code is not intended to really be used externally (it's just going to be some shared code between eventhubs and servicebus) we don't need to have it show up in pkg.go.dev.

Sharp edge: when we make changes in internal we'll need to push out new SB and EH packages, but this fits well with how those packages will work, so again, probably okay.

sdk/messaging/azservicebus/internal/atom/models.go
richardpark-msft: So I was under the assumption that something reasonable would happen as well. Prior to my change we were actually using the date.Time type and that had a custom unmarshaller. Using just plain time.Time worked for some of the entities, but failed for others.

The only thing that consistently worked as explicitly specifying the format, which seemed easier to do this way (since it's only deserializing, not serializing).

sdk/messaging/azservicebus/go.mod
richardpark-msft: Oh, it's GONE now! Look at that beautiful and svelte go.mod!

sdk/messaging/azservicebus/internal/atom/models.go
jhendrixMSFT: Do you know which ones it didn't work on?  That leads me to believe that some cases aren't in RFC3339 format which also means the `StringToTime` will silently fail.

sdk/messaging/azservicebus/internal/atom/models.go
richardpark-msft: Fixing. I couldn't take the dubiousness.

sdk/messaging/azservicebus/internal/atom/models.go
richardpark-msft: There is _one_ that's not parseable: `0001-01-01T00:00:00`. 

It comes back in a context where we don't even use it, so it's doubly annoying. I've modified the code to accomodate it (and return errors where appropriate).

richardpark-msft: (also, there's no longer a silent time.Zero's)

jhendrixMSFT (approved):

PR #16239, opened 2021-11-22
[Core] Handle non-HTTP errors in example_test.go

seankane-msft:

### DIFF
--- a/sdk/azcore/example_test.go
+++ b/sdk/azcore/example_test.go
@@ -81,7 +81,11 @@ func ExampleHTTPResponse() {
 		} else {
 			// ...
 		}
+	} else if err != nil {
+		// Handle non-HTTP error
+		fmt.Printf("There was a non-HTTP error that occurred: %s", err.Error())
 	}
+	
 	// Do something with response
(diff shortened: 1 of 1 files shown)


sdk/azcore/example_test.go
```diff
@@ -81,7 +81,11 @@ func ExampleHTTPResponse() {
 		} else {
 			// ...
 		}
+	} else if err != nil {
+		// Handle non-HTTP error
+		fmt.Printf("There was a non-HTTP error that occurred: %s", err.Error())
```
richardpark-msft, line 86: You'll want to avoid that fall-through - in the cases we were brainstorming I'm guessing 'resp' is nil.

sdk/azcore/example_test.go
```diff
@@ -81,7 +81,11 @@ func ExampleHTTPResponse() {
 		} else {
 			// ...
 		}
+	} else if err != nil {
```
jhendrixMSFT, line 84: The condition `if errors.As(err, &httpErr)` will always return false.

Issue #15758, opened 2021-10-08
ServiceBus stress tests


RickWinter:

richardpark-msft:
I think we can meaningfully say this is complete - we have several stress tests, executing regularly and adhoc. I can file future issues if we decide to enhance it further, and some stress work is also coming in naturally as part of other bug fixes.

PR #16241, opened 2021-11-22
[azservicebus] Stress test updates (mostly)

richardpark-msft:
Quite a few updates, mostly just trying to make the stress tests test more areas around reliablity, and be easier to write. We now have:
    
- infinite_send_and_receive: 5 days of constant sending and receiving, with 4 receivers on separate connections.
- constant_detachment: makes sure our links properly recover. There is still a bug that needs to be fixed for it to be 100% reliable (where amqp-common was eating and retrying on obviously fatal errors)
- finite_send_and_receive: sends 50000 messages, and receives them, making sure nothing is dropped
- long_running_renew_lock: receives a message and does nothing but renew the lock, every 5 seconds, and then complete it after 3 hours.
- Adding in a test case I saw in an issue: https://github.com/Azure/azure-service-bus-go/issues/126

Also, adding in some extra tools:
- StreamingBatch, similar to an idea we're kicking around for simplifying batch sending.
- A series of useful development CLI commands:
  - delete queues/topics by regex
  - constant update a queue to force detach logic
  - creating an auto-expiring test queue

Part of #15758


sdk/messaging/azservicebus/internal/stress/shared/streaming_batch.go
```diff
@@ -0,0 +1,110 @@
...
+
+	log.Printf("Sending message batch")
+	if err := sb.sender.SendMessageBatch(ctx, sb.currentBatch); err != nil {
+		return err
+	}
+
+	sb.stats.AddSent(sb.currentBatch.NumMessages())
+
+	// throttle a teeny bit.
+	time.Sleep(time.Second)
```
seankane-msft, line 79: Curiosity: why this sleep here?

richardpark-msft: It's to avoid throttling. 

I'm working on some changes that'll come in afterwards to make this more robust in the Sender itself.

richardpark-msft: (afterwards == separate PR btw)

seankane-msft (approved):
LGTM

PR #16243, opened 2021-11-23
Adopt msal for non-managed identity credentials

### LINKED ISSUE #14920: Convert `azidentity` to use msal/go
Switch the innards to use msal/go instead of our hand-written implementation.

### LINKED ISSUE #15632: InteractiveBrowserCredentialOptions.RedirectURL doesn't work

### LINKED ISSUE #15973: Add live test recordings
For the current live tests, and also add new tests for managed identity.

### LINKED ISSUE #16118: Remove azcore.ClientOptions copying
After #16091, it isn't necessary for credential constructors to copy `azcore.ClientOptions` parameters, because `runtime.NewPipeline` protects them from mutation.

chlowell:
Removes aad_identity_client.go and reimplements the credentials that were using it with MSAL clients. Adds recorded tests for these credentials so we don't lose all our test coverage. Closes #14920, closes #15632, closes #15973, closes #16118


sdk/azidentity/go.mod
```diff
@@ -2,12 +2,13 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
 go 1.16
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../azcore
```
jhendrixMSFT, line 5: Do we need this?

sdk/azidentity/go.mod
```diff
@@ -2,12 +2,13 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
...
+
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.20.0
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.2
+	github.com/AzureAD/microsoft-authentication-library-for-go v0.3.1
```
jhendrixMSFT, line 10: v0.4.0 is the latest preview that supports OBO

sdk/azidentity/go.mod
chlowell: Yes, until we release a version with `runtime.PipelineOptions`

jhendrixMSFT: Ah I see.  So I assume we won't be releasing this azidentity change until that time.

chlowell: That's my plan. I thought, why write code that's already out of date vs. main when there's no urgency to release azidentity.

sdk/azidentity/interactive_browser_credential_test.go
```diff
@@ -5,15 +5,16 @@ package azidentity
...
 	"testing"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
-	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
-	"golang.org/x/net/http2"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
 )
 
+var runManualBrowserTests = os.Getenv("AZIDENTITY_RUN_MANUAL_BROWSER_TESTS") != ""
```
seankane-msft, line 16: If you're checking to make sure the value is set you could use 
`var _, runManualBrowserTests = os.LookupEnv("...")`

seankane-msft (approved):

sdk/azidentity/azidentity.go
```diff
@@ -28,30 +42,13 @@ const (
...
-	headerUserAgent   = "User-Agent"
-	headerURLEncoded  = "application/x-www-form-urlencoded"
-	headerMetadata    = "Metadata"
-	headerContentType = "Content-Type"
-)
-
-const tenantIDValidationErr = "Invalid tenantID provided. You can locate your tenantID by following the instructions listed here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names."
-
 var (
 	successStatusCodes = [2]int{
```
jhendrixMSFT, line 46: Is this still needed?

sdk/azidentity/azidentity.go
```diff
@@ -80,15 +77,60 @@ func validTenantID(tenantID string) bool {
...
-// oauthPath returns the oauth path for AAD or ADFS
-func oauthPath(tenantID string) string {
-	if tenantID == "adfs" {
-		return "/oauth2"
+func (p pipelineAdapter) CloseIdleConnections() {
+	// do nothing
+}
+
+func (p pipelineAdapter) Do(r *http.Request) (*http.Response, error) {
+	req, err := runtime.NewRequest(r.Context(), r.Method, r.URL.String())
```
jhendrixMSFT, line 94: It's unfortunate that we have to do this (I feel like I had to do this elsewhere too).  Maybe we need a `runtime.WrapRequest` that creates a `runtime.Request` from an existing `http.Request`.

sdk/azidentity/authorization_code_credential.go
```diff
@@ -44,32 +47,70 @@ func NewAuthorizationCodeCredential(tenantID string, clientID string, authCode s
...
+	return &AuthorizationCodeCredential{authCode: authCode, clientSecret: options.ClientSecret, redirectURI: redirectURL, pca: c}, nil
 }
 
 // GetToken obtains a token from Azure Active Directory by redeeming the authorization code. This method is called automatically by Azure SDK clients.
 // ctx: Context controlling the request lifetime.
 // opts: Options for the token request, in particular the desired scope of the access token.
 func (c *AuthorizationCodeCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
-	tk, err := c.client.authenticateAuthCode(ctx, c.tenantID, c.clientID, c.authCode, c.clientSecret, "", c.redirectURI, opts.Scopes)
+	if c.cca != nil {
+		ar, err := c.cca.AcquireTokenSilent(ctx, opts.Scopes, confidential.WithSilentAccount(c.account))
```
jhendrixMSFT, line 86: I bet you could place this common code in a local func.
```go
acquireToken := func(func acquireSilent(/*params*/), func acquireByAuthCode(/*params*/)) (msal.AuthenticationResult, error) {
// stuff....
}
var ar msal.AuthenticationResult
var err error
if c.cca != nil {
    ar, err = acquireToken(c.cca.AcquireTokenSilent, c.cca.AcquireTokenByAuthCode)
}
// etc...
```

jhendrixMSFT (approved):

sdk/azidentity/authorization_code_credential.go
chlowell: The sticky part is that the MSAL methods have incompatible options params, e.g. `public.AcquireTokenSilent` requires `public.AcquireTokenSilentOption`, so a function which might call `confidential.AcquireTokenSilent` instead must know the difference. I guess I could hide the difference behind another type but imagine that isn't worth the cost in legibility. What do you think of something like this:
```go
var acquireTokenSilent func() (base.AuthResult, error)
if c.cca != nil {
    acquireTokenSilent = func() (base.AuthResult, error) {
        return c.cca.AcquireTokenSilent(ctx, opts.Scopes, confidential.WithSilentAccount(c.account))
    }
} else {
    acquireTokenSilent = func() (base.AuthResult, error) {
        return c.pca.AcquireTokenSilent(ctx, opts.Scopes, public.WithSilentAccount(c.account))
    }
}
```
Still repetitive but less so, and allows sharing the business logic at least.

jhendrixMSFT: Ah good point, I didn't look super close.  What you have is fine.

PR #16256, opened 2021-11-23
[Internal] Adding BodilessMatcher for use in recording

seankane-msft:
* Adds the bodiless matcher which ignores the body when matching a request. @chlowell already uses this in the azidentity tests so this makes his life a bit easier
* Renames `ResetSanitizers` to `ResetProxy`. This method resets sanitizers, transforms, and matchers


sdk/internal/recording/matchers_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddBodilessMatcher(t *testing.T) {
+	err := Start(t, packagePath, nil)
+	require.NoError(t, err)
+
+	err = AddBodilessMatcher(t, nil)
+	require.NoError(t, err)
```
chlowell, line 20: Should this also test that the matcher actually is set?

sdk/internal/recording/matchers.go
```diff
@@ -0,0 +1,39 @@
...
+func AddBodilessMatcher(t *testing.T, options *MatcherOptions) error {
+	if recordMode != PlaybackMode {
+		return nil
+	}
+	req, err := http.NewRequest("POST", "http://localhost:5000/Admin/SetMatcher", http.NoBody)
+	if err != nil {
+		panic(err)
+	}
+	req.Header["x-abstraction-identifier"] = []string{"BodilessMatcher"}
+	req.Header["x-recording-id"] = []string{GetRecordingId(t)}
```
chlowell, line 29: Does this mean BodilessMatcher applies only to this recording?

chlowell (commented):
Thanks for making my life a bit easier! Is there a way to reset the matcher to the default without restarting the proxy (I haven't found one)?

sdk/internal/recording/matchers.go
seankane-msft: Yes, going to upgrade so you can pass in `nil` for the `t` argument and will apply to all (which I think is a less common use case).

seankane-msft:
> Thanks for making my life a bit easier! Is there a way to reset the matcher to the default without restarting the proxy (I haven't found one)?

If you pass in the `*testing.T` instance in this case it will only apply for a single test. Does that cover your use case?

sdk/internal/recording/matchers_test.go
seankane-msft: Done

chlowell:
It does, though inefficiently. azidentity needs BodilessMatcher for every test, so ideally I would set it once at the start of the run and restore the default matcher at the end to avoid false positives for other packages' tests.

seankane-msft:
Would something similar to this work
```golang
func TestMain(m *testing.M) {
    err := AddBodilessMatcher(nil, nil)
    handle(err)
    // Add any session sanitizers

    exitval := m.Run()

    // This is improperly named and actually resets everything.
    err = recording.ResetSanitizers(nil)
    handle(err)

    os.Exit(exitVal)
}
```

chlowell:
It would! I had assumed `ResetSanitizers` reset only sanitizers. Probably ought to rename that.

sdk/internal/recording/matchers_test.go
```diff
@@ -0,0 +1,97 @@
...
+	jsonFile, err := os.Open(fmt.Sprintf("./testdata/recordings/%s.json", t.Name()))
+	require.NoError(t, err)
+	defer jsonFile.Close()
+
+	var data RecordingFileStruct
+	byteValue, err := ioutil.ReadAll(jsonFile)
+	require.NoError(t, err)
+	err = json.Unmarshal(byteValue, &data)
+	require.NoError(t, err)
+	require.Equal(t, data.Entries[0].RequestBody, "")
```
chlowell, line 55: I feel like I'm missing something here; doesn't this just verify the body regex sanitizer worked? Don't you need to send a second request, identical to the first save for its content, to verify the bodiless matcher is set?

seankane-msft: The CI runs in playback so that's sort of the check here. However, you're correct this would be a more thorough way to test the matcher works.

sdk/internal/recording/sanitizer.go
```diff
@@ -401,8 +401,8 @@ func AddURISubscriptionIDSanitizer(value string, options *RecordingOptions) erro
 	return handleProxyResponse(client.Do(req))
 }
 
-// ResetSanitizers deletes any existing sanitizers from the test proxy.
-func ResetSanitizers(options *RecordingOptions) error {
+// ResetProxy deletes any existing sanitizers, matchers, and transforms from the current test proxy session
```
chlowell, line 404: Just a suggestion, trying to capture that resetting means going back to default settings:
```suggestion
// ResetProxy restores the proxy's default sanitizers, matchers, and transforms
```

sdk/internal/CHANGELOG.md
```diff
@@ -3,10 +3,11 @@
 ## 0.8.3 (Unreleased)
 
 ### Features Added
-
+* Adds the `AddBodilessMatcher` method [#16256](https://github.com/Azure/azure-sdk-for-go/pull/16256)
 ### Breaking Changes
 
 ### Bugs Fixed
+* Renames `ResetSanitizers` to `ResetProxy`
```
chlowell, line 10: This is a breaking change

sdk/internal/recording/matchers.go
```diff
@@ -0,0 +1,42 @@
...
+	"testing"
+)
+
+// Optional parameters for the AddBodilessMatcher operation
+type MatcherOptions struct {
+}
+
+// AddBodilessMatcher adjusts the "match" operation to exclude the body when matching a request to a recording's entries.
+// Pass in `nil` for `t` if you want the bodiless matcher to apply everywhere
+func AddBodilessMatcher(t *testing.T, options *MatcherOptions) error {
```
chlowell, line 21: What do you think of `SetBodilessMatcher`? The rationale being that "set" is more correct semantically because the proxy only uses one matcher at a time; this method doesn't add to a collection but instead replaces the current matcher

chlowell (approved):

PR #16265, opened 2021-11-24
feat: add dependency check step to build pipeline

tadelesh:

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -8,6 +8,16 @@ parameters:
 
 steps:
 
+  - task: Powershell@2
+    displayName: 'Dependency Check'
+    env:
+      GO111MODULE: 'on'
+    inputs:
+      targetType: filePath
+      pwsh: true
+      filePath: eng/scripts/Invoke-DependencyCheck.ps1
+      arguments: '${{ parameters.ServiceDirectory }}'
+      
   - script: |
       curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${{parameters.LintVersion}}
       golangci-lint --version
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -16,7 +16,7 @@ steps:
       targetType: filePath
       pwsh: true
       filePath: eng/scripts/Invoke-DependencyCheck.ps1
-      arguments: '${{ parameters.ServiceDirectory }}'
+      arguments: 'sdk/${{ parameters.ServiceDirectory }}'
       
   - script: |
       curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${{parameters.LintVersion}}
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -36,16 +36,6 @@ steps:
       filePath: eng/scripts/build.ps1
       arguments: -vet -skipBuild -filter '${{ parameters.ServiceDirectory }}'
 
-  - task: Powershell@2
-    displayName: 'Dependency Check'
-    env:
-      GO111MODULE: 'on'
-    inputs:
-      targetType: filePath
-      pwsh: true
-      filePath: eng/scripts/Invoke-DependencyCheck.ps1
-      arguments: '${{ parameters.ServiceDirectory }}'
-
   - pwsh: |
       go install github.com/jstemmer/go-junit-report@v0.9.1
       go install github.com/axw/gocov/gocov@v1.0.0
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -36,6 +36,16 @@ steps:
       filePath: eng/scripts/build.ps1
       arguments: -vet -skipBuild -filter '${{ parameters.ServiceDirectory }}'
 
+  - task: Powershell@2
+    displayName: 'Dependency Check'
+    env:
+      GO111MODULE: 'on'
+    inputs:
+      targetType: filePath
+      pwsh: true
+      filePath: eng/scripts/Invoke-DependencyCheck.ps1
+      arguments: '${{ parameters.ServiceDirectory }}'
+
   - pwsh: |
       go install github.com/jstemmer/go-junit-report@v0.9.1
       go install github.com/axw/gocov/gocov@v1.0.0
--- a/eng/scripts/Invoke-DependencyCheck.ps1
+++ b/eng/scripts/Invoke-DependencyCheck.ps1
@@ -8,24 +8,6 @@ $ignoreCheck = "sdk/synapse/azartifacts", "sdk/template/aztemplate", "sdk/azcore
 
 $sdks = Get-AllPackageInfoFromRepo
 
-$sdkRoot = Join-Path $RepoRoot "sdk"
-
-# Get all existed packages
-$packagesImport = ""
-$newPackage = ""
-foreach ($sdk in $sdks)
-{
-    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
-    {
-        continue
-    }
-    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $sdkRoot $newPackageDirectory)).Path)
-    {
-        $newPackage = $sdk.Name
-    }
-    $packagesImport = $packagesImport + "`t_ `"github.com/Azure/azure-sdk-for-go/$($sdk.Name)`"`n"
-}
-
 ## Create depcheck module
 $workingPath = Join-Path $RepoRoot "sdk" "depcheck"
 if (Test-Path -Path $workingPath)
@@ -40,6 +22,23 @@ Write-Host "##[command]Executing go mod init in " $workingPath
 go mod init github.com/Azure/azure-sdk-for-go/sdk/depcheck
 if ($LASTEXITCODE) { exit $LASTEXITCODE }
 
+# Get all existed packages
+$packagesImport = ""
+foreach ($sdk in $sdks)
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $RepoRoot $newPackageDirectory)).Path)
+    {
+        ## Add replace for new package
+        $modPath = Join-Path $RepoRoot "sdk" "depcheck" "go.mod"
+        Add-Content $modPath "`nreplace github.com/Azure/azure-sdk-for-go/$($sdk.Name) => ../../$($sdk.Name)`n"
+    }
+    $packagesImport = $packagesImport + "`t_ `"github.com/Azure/azure-sdk-for-go/$($sdk.Name)`"`n"
+}
+
 ## Add main.go
 $mainPath = Join-Path $RepoRoot "sdk" "depcheck" "main.go"
 New-Item -Path $mainPath -ItemType File -Value '' -Force
@@ -52,13 +51,6 @@ $packagesImport
 func main() {
 }
 "
-
-## Add replace for new package
-if ($newPackage -ne "")
-{
-    $modPath = Join-Path $RepoRoot "sdk" "depcheck" "go.mod"
-    Add-Content $modPath "`nreplace github.com/Azure/azure-sdk-for-go/$newPackage => ../../$newPackage`n"
-}
 
 ## Run go mod tidy
 Write-Host "##[command]Executing go mod tidy in " $workingPath
--- a/eng/scripts/Invoke-DependencyCheck.ps1
+++ b/eng/scripts/Invoke-DependencyCheck.ps1
@@ -4,7 +4,7 @@ Param(
 
 . (Join-Path $PSScriptRoot .. common scripts common.ps1)
 
-$ignoreCheck = "sdk/synapse/azartifacts", "sdk/template/aztemplate", "sdk/azcore", "sdk/azidentity"
+$ignoreCheck = ,"sdk/synapse/azartifacts"
 
 $sdks = Get-AllPackageInfoFromRepo
 
--- a/eng/scripts/Invoke-DependencyCheck.ps1
+++ b/eng/scripts/Invoke-DependencyCheck.ps1
@@ -38,7 +38,7 @@ foreach ($sdk in $sdks)
     {
         continue
     }
-    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $RepoRoot $newPackageDirectory)).Path)
+    if ($sdk.DirectoryPath.EndsWith($PackageDirectory))
     {
         ## Add replace for new package
         $modPath = Join-Path $RepoRoot "sdk" "depcheck" "go.mod"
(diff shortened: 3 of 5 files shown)


tadelesh:
Incompatible dependency with different packages will cause mod auto-resolving failure when customer import multiple v2 packages including data plane and mgmt. plane SDK. This script added to build process will import all the v2 modules and try to resolve dependency as well as build and vet to ensure any update will not breaking the package compatible.

jhendrixMSFT:
This shouldn't be a problem post-GA.  Seems expensive to build the world for each PR.

tadelesh:
> This shouldn't be a problem post-GA. Seems expensive to build the world for each PR.

According to ci log, it will take about 3.5mins in average. It seems expensive indeed, but I think maybe it is valuable before GA as we may have multiple breaking changes. I don't know whether there is a more economic way to do such check, maybe we could just use it temporary and keep it as a tool after GA.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -36,6 +36,16 @@ steps:
...
+  - task: Powershell@2
+    displayName: 'Dependency Check'
+    env:
+      GO111MODULE: 'on'
+    inputs:
+      targetType: filePath
+      pwsh: true
+      filePath: eng/scripts/Invoke-DependencyCheck.ps1
+      arguments: '${{ parameters.ServiceDirectory }}'
+
```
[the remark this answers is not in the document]
tadelesh: Moved. It will save 3/4 times.

seankane-msft:
@tadelesh in the tests run against azkeys, the `go build` command fails with exit code 1 but without any details on what the error is. Does this work locally for you?

tadelesh:
> @tadelesh in the tests run against azkeys, the `go build` command fails with exit code 1 but without any details on what the error is. Does this work locally for you?

I've removed all the `-x` flag in `go build` cmd to give a more readable log. You can get the error for `azkeys` for now.

seankane-msft:
> I've removed all the `-x` flag in `go build` cmd to give a more readable log. @seankane-msft You can get the error for `azkeys` for now.

Thanks for removing that flag, the output is much clearer now. I have a fix for that once I can release changes to internal. For now that should not be blocking on this PR.

tadelesh:
> > I've removed all the `-x` flag in `go build` cmd to give a more readable log. @seankane-msft You can get the error for `azkeys` for now.
> 
> Thanks for removing that flag, the output is much clearer now. I have a fix for that once I can release changes to internal. For now that should not be blocking on this PR.

@seankane-msft If this PR merge first, will it be conflict with the 'nightly build' change you are doing?

@benbp @weshaggard Could you help to have a review?

seankane-msft:
> > > I've removed all the `-x` flag in `go build` cmd to give a more readable log. @seankane-msft You can get the error for `azkeys` for now.
> > 
> > 
> > Thanks for removing that flag, the output is much clearer now. I have a fix for that once I can release changes to internal. For now that should not be blocking on this PR.
> 
> @seankane-msft If this PR merge first, will it be conflict with the 'nightly build' change you are doing?

I don't believe so.

seankane-msft (approved):
LGTM

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -8,6 +8,16 @@ parameters:
...
 
+  - task: Powershell@2
+    displayName: 'Dependency Check'
+    env:
+      GO111MODULE: 'on'
+    inputs:
+      targetType: filePath
+      pwsh: true
+      filePath: eng/scripts/Invoke-DependencyCheck.ps1
+      arguments: '${{ parameters.ServiceDirectory }}'
```
weshaggard, line 19: For consistency should we pass in the `sdk/` part of this as well?
```suggestion
      arguments: 'sdk/${{ parameters.ServiceDirectory }}'
```

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
+Param(
+    [string] $newPackageDirectory
+)
+
+. (Join-Path $PSScriptRoot .. common scripts common.ps1)
+
+$ignoreCheck = "sdk/synapse/azartifacts", "sdk/template/aztemplate", "sdk/azcore", "sdk/azidentity"
```
weshaggard, line 7: Why do we want to ignore these? Is this temporary and expected to be fixed later? If so can we file an issue and add a comment here linking to that issue.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+func main() {
+}
+"
+
+## Add replace for new package
+if ($newPackage -ne "")
+{
+    $modPath = Join-Path $RepoRoot "sdk" "depcheck" "go.mod"
+    Add-Content $modPath "`nreplace github.com/Azure/azure-sdk-for-go/$newPackage => ../../$newPackage`n"
```
weshaggard, line 60: Instead of doing this replace here why not handle it in the loop above? If it is the package of question add direct import there.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+$newPackage = ""
+foreach ($sdk in $sdks)
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $sdkRoot $newPackageDirectory)).Path)
+    {
+        $newPackage = $sdk.Name
```
weshaggard, line 24: I would just make the special relative link import here instead of doing it later below.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+# Get all existed packages
+$packagesImport = ""
+$newPackage = ""
+foreach ($sdk in $sdks)
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $sdkRoot $newPackageDirectory)).Path)
```
weshaggard, line 22: Do we really need to resolve the paths? Why not just compare the Name?

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+    [string] $newPackageDirectory
+)
+
+. (Join-Path $PSScriptRoot .. common scripts common.ps1)
+
+$ignoreCheck = "sdk/synapse/azartifacts", "sdk/template/aztemplate", "sdk/azcore", "sdk/azidentity"
+
+$sdks = Get-AllPackageInfoFromRepo
+
+$sdkRoot = Join-Path $RepoRoot "sdk"
```
weshaggard, line 11: I would avoid this and just pass the "sdk" as part of the parameter and avoid needing to resolve full paths.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $sdkRoot $newPackageDirectory)).Path)
+    {
+        $newPackage = $sdk.Name
+    }
+    $packagesImport = $packagesImport + "`t_ `"github.com/Azure/azure-sdk-for-go/$($sdk.Name)`"`n"
```
weshaggard, line 26: Do we need to specify versions of these dependencies, or are we fine with taking the latest version?

eng/scripts/Invoke-DependencyCheck.ps1
benbp: Regardless I'd rather not hardcode skips into our scripts. It's easy to completely forget about them. I think it would also be better to have these as sdk-level opt-outs, rather than in a centralized place, e.g. passing them down through the `ci.yml` files as parameters in the respective service directories.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: The loop is for the `import` statement in `main.go` file. This is for the replacement of package in `go.mod` file.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Fixed.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Changed.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Changed.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: For single package, the script will let `go mod tidy` to try to find all compatible version except using `replace` to use the latest code of THE package which is under ci check.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: I've removed most of ignore package here except one: sdk/synapse/azartifacts. This package is a deprecated package that  code has not deleted yet. So I need to ignore this package in the dependency check.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: As path separator may be different for `$sdk.DirectoryPath` and `$newPackageDirectory`. Resolve path is just for safety.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,69 @@
...
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $RepoRoot $newPackageDirectory)).Path)
+    {
+        ## Add replace for new package
+        $modPath = Join-Path $RepoRoot "sdk" "depcheck" "go.mod"
+        Add-Content $modPath "`nreplace github.com/Azure/azure-sdk-for-go/$($sdk.Name) => ../../$($sdk.Name)`n"
```
weshaggard, line 37: Shouldn't this be an append to $packageImports? Also should there be an else on this if statement?

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,69 @@
...
+
+# Get all existed packages
+$packagesImport = ""
+foreach ($sdk in $sdks)
+{
+    if ($sdk.Name -like "*internal*" -or $sdk.Name -in $ignoreCheck)
+    {
+        continue
+    }
+    if ((Resolve-Path $sdk.DirectoryPath).Path -eq (Resolve-Path (Join-Path $RepoRoot $newPackageDirectory)).Path)
```
weshaggard, line 33: The slashes shouldn't be an issue when comparing the $sdk.name $newPackageDirectory as they should match exactly or if you are worried you can just normalize them by replacing them.

eng/scripts/Invoke-DependencyCheck.ps1
weshaggard: Nevermind I guess the imports are going into main.go and this is going into go.mod.

eng/scripts/Invoke-DependencyCheck.ps1
weshaggard: We might consider some other way of filtering this out because I expect this ignore list will be left here forever now even after that project is removed. At any rate this isn't a blocker but it also isn't the ideal way of excluding things.

eng/scripts/Invoke-DependencyCheck.ps1
weshaggard: I think comparing sdk.Name to newPackageDirectory should always work but if not then perhaps doing a simple replace of "\\" with "/" at the top of this file should work. It will be a lot more efficient then resolving the paths every time.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: I've changed the implementation for detecting deprecated packages. The deprecated packages in go will have `retract` section in `go.mod` file. So I'll skip the package by whether the package version is in the `retract` section.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Fixed.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Fixed.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+# Find whether latest version is in the `retract` section in `go.mod` to judge whether the package is temporarily deprecated. 
+function IsPackageDeprecated($sdk)
+{
+    $RETRACT_SECTION_REGEX = "retract\s*((?<retract>(.|\s)*))"
+    $modContent = Get-Content (Join-Path $sdk.DirectoryPath 'go.mod') -Raw
+    if ($modContent -match $RETRACT_SECTION_REGEX) {
+        return $($matches["retract"]).Indexof($sdk.Version) -ne -1
+    }
+}
```
weshaggard, line 31: Should we explicitly return false here? So we don't get some strange null behavior?

weshaggard (approved):
A couple comments but looks good.

eng/scripts/Invoke-DependencyCheck.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+# Get all existed packages
+$packagesImport = ""
+foreach ($sdk in $sdks)
+{
+    if ($sdk.Name -like "*internal*" -or (IsPackageDeprecated $sdk))
+    {
+        continue
+    }
+    if ($sdk.DirectoryPath.EndsWith($PackageDirectory))
```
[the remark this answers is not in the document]
tadelesh: Do you mean I can use $sdk.Name directly? I've changed the code.

eng/scripts/Invoke-DependencyCheck.ps1
tadelesh: Fixed.

eng/scripts/Invoke-DependencyCheck.ps1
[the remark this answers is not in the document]
weshaggard: Yes that is what I meant.

jhendrixMSFT:
I'm unclear why this was merged, especially when we a) have breaking changes that haven't been released yet and b) have a few more on the way.  I thought this was going to be enabled post-GA?

benbp:
The check enforcer should not have been overridden here (at the very least without a comment on justification) when the pipeline failure was a direct result of the change.

tadelesh:
> 

Got it. This check is a chicken-and-egg conundrum. So I just bypassed it if any breaking change is intentional. It's good to be set as an optional before GA.

tadelesh:
> The check enforcer should not have been overridden here (at the very least without a comment on justification) when the pipeline failure was a direct result of the change.

Got it. This check is a chicken-and-egg conundrum. So I just bypassed it if any breaking change is intentional. It's good to be set as an optional before GA.

PR #16267, opened 2021-11-24
Refactor IMDS discovery to remove probing, stop caching failures

### LINKED ISSUE #15396: Refactor managedIdentityClient.getMSIType calling pattern
From https://github.com/Azure/azure-sdk-for-go/pull/15377#discussion_r696809162: `managedIdentityClient.createAuthRequest` has an error case for calls when `msiType == msiTypeUnknown`, presumably because a caller can in principle call `createAuthRequest` without first calling `getMSIType`. In practice this doesn't occur, and the case isn't meaningful because the client has a default msiType (IMDS). Refactoring such that `msiType != msiTypeUnknown` when creating a request would eliminate this case and simplify error messaging.

### LINKED ISSUE #16116: NewManagedIdentityCredential shouldn't send an HTTP request
`NewManagedIdentityCredential` probes IMDS for availability when the hosting environment doesn't contain configuration for any other managed identity implementation. This entails sending an HTTP request, which is problematic for a couple reasons, e.g. `NewManagedIdentityCredential` doesn't have a `Context` parameter. Fix would be to defer the probe until `GetToken()`.

chlowell:
Here's a few changes to behavior around IMDS managed identity to bring azidentity into alignment with the .NET SDK:
- A failed IMDS token request no longer stops a credential from requesting tokens in the future. All auth failures are now retriable without constructing new credentials or clients.
- ManagedIdentityCredential no longer probes IMDS to determine whether it's available before requesting a token from it. Transient latency spikes and connection failures don't disable the credential anymore.
  - To minimize the impact of timeouts when IMDS really isn't available, initial requests from DefaultAzureCredential have a 1 second timeout. This timeout is removed when IMDS responds, and the credential uses its default configuration thereafter.

Closes #15396, closes #16116

### DIFF
--- a/sdk/azidentity/logging.go
+++ b/sdk/azidentity/logging.go
@@ -5,7 +5,6 @@ package azidentity
 
 import (
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
@@ -19,30 +18,6 @@ import (
 // used when obtaining credentials and the type of credential used.
 const EventAuthentication log.Event = "Authentication"
 
-// log environment variables that can be used for credential types
-func logEnvVars() {
-	if !log.Should(EventAuthentication) {
-		return
-	}
-	// Log available environment variables
-	envVars := []string{}
-	if envCheck := os.Getenv("AZURE_TENANT_ID"); len(envCheck) > 0 {
-		envVars = append(envVars, "AZURE_TENANT_ID")
-	}
-	if envCheck := os.Getenv("AZURE_CLIENT_ID"); len(envCheck) > 0 {
-		envVars = append(envVars, "AZURE_CLIENT_ID")
-	}
-	if envCheck := os.Getenv("AZURE_CLIENT_SECRET"); len(envCheck) > 0 {
-		envVars = append(envVars, "AZURE_CLIENT_SECRET")
-	}
-	if envCheck := os.Getenv(azureAuthorityHost); len(envCheck) > 0 {
-		envVars = append(envVars, azureAuthorityHost)
-	}
-	if len(envVars) > 0 {
-		log.Writef(EventAuthentication, "Azure Identity => Found the following environment variables:\n\t%s", strings.Join(envVars, ", "))
-	}
-}
-
 func logGetTokenSuccess(cred azcore.TokenCredential, opts policy.TokenRequestOptions) {
 	if !log.Should(EventAuthentication) {
 		return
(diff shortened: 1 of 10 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -111,36 +105,60 @@ func setRetryOptionDefaults(o *policy.RetryOptions) {
...
+			c.msiType = msiTypeCloudShell
+		}
+	} else if endpoint, ok := os.LookupEnv(identityEndpoint); ok {
+		if _, ok := os.LookupEnv(identityHeader); ok {
+			if _, ok := os.LookupEnv(identityServerThumbprint); ok {
+				env = "Service Fabric"
+				c.endpoint = endpoint
+				c.msiType = msiTypeServiceFabric
+			}
+		} else if _, ok := os.LookupEnv(arcIMDSEndpoint); ok {
```
jhendrixMSFT, line 136: What happens if this and the previous clause are false?

sdk/azidentity/managed_identity_credential.go
```diff
@@ -75,27 +75,20 @@ func NewManagedIdentityCredential(options *ManagedIdentityCredentialOptions) (*M
...
 // opts: Options for the token request, in particular the desired scope of the access token.
 func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
-	if opts.Scopes == nil {
-		err := errors.New("must specify a resource in order to authenticate")
-		addGetTokenFailureLogs("Managed Identity Credential", err, true)
-		return nil, err
-	}
-	if len(opts.Scopes) != 1 {
-		err := errors.New("can only specify one resource to authenticate with ManagedIdentityCredential")
+	if opts.Scopes == nil || len(opts.Scopes) != 1 {
```
jhendrixMSFT, line 90: The `nil` check can be removed.  `len()` on a `nil` slice is zero.

sdk/azidentity/logging.go
```diff
@@ -19,30 +18,6 @@ import (
 // used when obtaining credentials and the type of credential used.
 const EventAuthentication log.Event = "Authentication"
 
-// log environment variables that can be used for credential types
-func logEnvVars() {
```
jhendrixMSFT, line 23: Why was this and the below logging removed?  Is it captured elsewhere?

sdk/azidentity/managed_identity_client.go
chlowell: You get the default, IMDS, whenever another implementation isn't identified.

sdk/azidentity/logging.go
chlowell: I removed this function because it was called only in the context of managed identity, and none of these variables apply to managed identity. `NewEnvironmentCredential` logs similar information. I removed `logMSIEnv` because `newManagedIdentityClient` now logs the same information.

jhendrixMSFT (approved):
How confident are you about the existing test coverage?  Catalina used to run a variety of manual tests given the finicky nature of MSI/IMDS.  Do you think that's needed here, or are the recordings sufficient?

chlowell:
In any case I think we need to manually test all the platforms before 1.0. I updated the recordings for IMDS as part of this PR but didn't try any other platforms. We won't be able to record on all platforms and the test suite has gaps, e.g. no tests for Azure Arc (#16640).

Issue #14750, opened 2021-06-07
Table Storage - Batch Insert only inserts one record when using Azurite


davidandradeduarte:
### Bug Report

- Package and file in question `storage/table_batch.go`
- SDK version: `55.1.0`
- Go version: `1.16.4 darwin/amd64`
- Azurite version: `mcr.microsoft.com/azure-storage/azurite@latest`

### What happened?
Trying to insert multiple records into a table storage using batch mode doesn't work as expected and only inserts one.
**This only happens when using Azurite**.

### What did you expect or want to happen?
When running `t.ExecuteBatch` I'd expect it to insert all the entities added to the batch using `t.InsertEntity`.
I've tried with a different SDK (.NET) and it works.

### How can we reproduce it?
https://github.com/davidandradeduarte/azurite-insert-batch-bugs#batch-insert-with-go-sdk-using-azurite-only-inserts-one-record
This repository reproduces the issue. You just need to have Azurite running on default ports.
You can use the dotnet-sdk version included to test alongside and see that it works as expected.

ArcturusZhang:
Hi @davidandradeduarte thanks for this issue!

Hi @jhendrixMSFT could you please take a look?

davidandradeduarte:
hi 👋 
@ArcturusZhang, @jhendrixMSFT any update on this?

jhendrixMSFT:
Sorry for the delay.

The tables functionality in the `storage` package has been deprecated in favor of our new [`aztables`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/aztables) module.  Can you please see if this resolves your issue?

davidandradeduarte:
Thanks, I'll have a look!

davidandradeduarte:
Having a hard time trying to make it work.

Following this guide:
https://devblogs.microsoft.com/azure-sdk/announcing-the-preview-release-for-azure-tables-for-go/

I've updated the code as follows:
https://github.com/davidandradeduarte/azurite-insert-batch-bugs/blob/main/go-sdk/14750-aztables-fix/main.go

But fails with a _400 - Bad Request_ when I call `client.Create` (https://github.com/davidandradeduarte/azurite-insert-batch-bugs/blob/9767b15c5c5095df8a82af439d9cbab2102fc656/go-sdk/14750-aztables-fix/main.go#L21).

Azurite response
[...]

PR #16320, opened 2021-11-29
[Tables] Adding azurite test for Tables

seankane-msft:
Addresses customer reported issue #14750 . Adds a single test for running the Tables service against Azurite.


jhendrixMSFT (approved):
Do we need to add this to the docs anywhere?

davidandradeduarte (approved):
👍

PR #16326, opened 2021-11-30
[azservicebus] Updating readme with AAD/Service Principal related information

richardpark-msft:
Based off feedback from @mikekistler


mikekistler (approved):
Looks good! 👍

PR #16353, opened 2021-11-30
Fix messaging/internal version string format

benbp:


seankane-msft (approved):

richardpark-msft (approved):
I can only trust that this is what's supposed to happen.

Is this also what should be reported via our user agent strings?

benbp:
@richardpark-msft I do not know re: user agent strings.

For the version string, this is the format all the automation expects in order to determine the version of the package. It gets reported as a non-blocking error in the verify changelog step (since it checks every package version string), but should fail downstream. I'm not quite sure how this made it through, but we should think about ways to validate so that it can't.

richardpark-msft:
> 

Time to move that validation into the CI pipelines? There's no way to manually track all the stuff we need to do at this point.

benbp:
It looks like it already is, and it will report the error (e.g. https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1220098&view=results), but it won't fail the pipeline. I think we may want to update the verify changelog step to only fail for the package in question, though I don't understand why it scans all in the first place or if there's context behind that ([backing code](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L29)). @weshaggard @chidozieononiwu?

weshaggard:
@benbp The pipeline you linked to is the servicebus pipeline not the messaging/internal pipeline. The error you are seeing in the logs is actually just information about which packages were discovered and the messaging/internal one failed to find the version which was the check you added recently for diagnosing this. The only changelog that gets verified there is the servicebus one, even though all the changelogs get read during the package discovery. 

We look to be missing a pipeline for this new package which, if one was created and ran it would have failed.

weshaggard:
Looks like the prepare-pipelines check let this PR merge https://github.com/Azure/azure-sdk-for-go/pull/16210 even though the real pipeline didn't get created and ran. That seems like a whole in our process.

benbp:
@weshaggard the prepare pipelines problem is more likely related to the issues we had with the `internal` naming and not supporting multi-segment pipeline names. I fixed this for the keyvault/internal, but the timing of this PR may have been slightly before that.

benbp:
@weshaggard @richardpark-msft I updated this PR to change the internal ci file name to `ci.messaging.yml` which will generate a non-overlapping named pipeline `go - internal - messaging`.

I'm not sure how we want to try to catch these issues. Perhaps running prepare-pipelines for new files should fail when it finds pre-existing pipelines, but only when run from PR? The edge cases here are a little tricky.

PR #16392, opened 2021-12-02
[azidentity] Making ChainedTokenCredential re-use the first successful credential

### LINKED ISSUE #16268: ChainedTokenCredential prefers its first successful constituent
An instance of ChainedTokenCredential should iterate over its constituents only until one provides a token, and request tokens only from that credential thereafter. This is the default behavior of DefaultAzureCredential in other SDKs, and because azidentity is still unstable, we have the flexibility to make it the default behavior of ChainedTokenCredential as well. This will make the credential more because applications typically want to use a chain to support deploying to multiple environments without writing conditional logic to choose a credential type, not to support falling back to a different identity when one fails to authenticate. Still, let's also add a bool to ChainedTokenCredentialOptions so applications can opt in to the latter behavior.

sadasant:
This PR makes it so instances of `ChainedTokenCredential` will now re-use the first successful credential on `GetToken` calls.

Fixes #16268

Feedback always appreciated!


sdk/azidentity/chained_token_credential.go
```diff
@@ -19,7 +19,8 @@ type ChainedTokenCredentialOptions struct {
 
 // ChainedTokenCredential is a chain of credentials that enables fallback behavior when a credential can't authenticate.
 type ChainedTokenCredential struct {
-	sources []azcore.TokenCredential
+	sources              []azcore.TokenCredential
+	successfulCredential *azcore.TokenCredential
```
seankane-msft, line 23: I don't think this has to be an explicit pointer, interfaces are pointers under the hood.

sdk/azidentity/chained_token_credential.go
```diff
@@ -44,24 +45,30 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
-		} else if err != nil {
-			var authFailed AuthenticationFailedError
-			if errors.As(err, &authFailed) {
-				err = fmt.Errorf("Authentication failed:\n%s\n%s"+createChainedErrorMessage(errList), err)
-				authErr := newAuthenticationFailedError(err, authFailed.RawResponse())
-				return nil, authErr
+
+	if c.successfulCredential != nil {
+		successfulCredential := *c.successfulCredential
+		token, err = successfulCredential.GetToken(ctx, opts)
```
seankane-msft, line 51: Once you remove the pointer I think you can do `token, err = c.succesfulCredential.GetToken(ctx, opts)`

seankane-msft (commented):
Small comments

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -153,6 +153,99 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+	if tk.ExpiresOn.IsZero() {
+		t.Fatalf("Received an incorrect time in the response")
+	}
+	if cred.successfulCredential == nil {
+		t.Fatalf("The successful credential pointer was not assigned")
+	}
+	tk2, err2 := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{scope}})
+	if err2 != nil {
+		t.Fatalf("Received an error when attempting to get a token but expected none. Error: %v", err2)
+	}
```
sadasant, line 240: I could skip a lot of this error handling if we assumed that all of the behaviors unrelated to the main goal of this specific test have been already tested. Then, I could only check for errors on these last checks, which are the ones that matter the most to me. Would that make sense? Or should I keep every error check?

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -153,6 +153,99 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+	if tk.Token != tokenValue {
+		t.Fatalf("Received an incorrect access token")
+	}
+	if tk.ExpiresOn.IsZero() {
+		t.Fatalf("Received an incorrect time in the response")
+	}
+	if cred.successfulCredential == nil {
+		t.Fatalf("The successful credential pointer was not assigned")
+	}
```
sadasant, line 236: Is there a recommended way to check if the EnvironmentCredential was ever called? this would make this test more robust.

sdk/azidentity/chained_token_credential_test.go
seankane-msft: It's better to check every error, more idiomatic

sdk/azidentity/chained_token_credential.go
```diff
@@ -78,4 +85,14 @@ func createChainedErrorMessage(errList []CredentialUnavailableError) string {
...
 
+func formatError(err error, errList []CredentialUnavailableError) error {
+	var authFailed AuthenticationFailedError
+	if errors.As(err, &authFailed) {
+		err = fmt.Errorf("Authentication failed:\n%s\n%s"+createChainedErrorMessage(errList), err)
+		authErr := newAuthenticationFailedError(err, authFailed.RawResponse())
+		return authErr
+	}
+	return err
+}
```
sadasant, line 96: I abstracted this away so I could re-use it. There are many ways to do this! So please give me feedback. Part of the goal of this PR is to understand how to do things for the Go SDK.

sdk/azidentity/chained_token_credential.go
sadasant, line 23: This is really the simplest approach I could think of 😊

sadasant: ok ok! thank you!

sdk/azidentity/chained_token_credential.go
```diff
@@ -44,24 +45,30 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
-				return nil, authErr
+
+	if c.successfulCredential != nil {
+		successfulCredential := c.successfulCredential
+		token, err = successfulCredential.GetToken(ctx, opts)
+		if err != nil {
+			return nil, formatError(err, errList)
+		}
+		return token, nil
+	} else {
```
jhendrixMSFT, line 55: The `else` is redundant and can be removed.

sdk/azidentity/chained_token_credential.go
sadasant: Correct! Thank you!

sdk/azidentity/chained_token_credential.go
sadasant: True! Thank you!

sdk/azidentity/chained_token_credential_test.go
seankane-msft: type poli struct{ count int }

func (p *poli) Do(req *policy.Request) (*http.Response, error) {
	p.count += 1
	return req.Next()
}
func TestSomething(t *testing.T) {
	p := &poli{count: 1}
	envCred, err := NewEnvironmentCredential(&EnvironmentCredentialOptions{
			PerCallPolicies: []pipeline.Policy{p},
		},
	})
}

sdk/azidentity/chained_token_credential.go
jhendrixMSFT: Given that this is specific to ChainedTokenCredential, and it's only used in one place, I don't think there's much value in splitting this out.  At the very least, it should be a method on ChainedTokenCredential due to it specificity.

sadasant: That makes sense! — what about a function within the `GetToken` method? `formatError := func …`? would that be ok? @jhendrixMSFT

sdk/azidentity/chained_token_credential_test.go
sadasant: ohh cool!! I will do that! thank you!

sdk/azidentity/chained_token_credential.go
jhendrixMSFT: Sure, but is it used in more than one place?  If the goal is to reduce duplicated code that's great, but I don't see it in this PR.

jhendrixMSFT: Oh I see it now.  Yeah use a local func.

sadasant: on my way!

sdk/azidentity/chained_token_credential_test.go
sadasant: Done! Please review again 🙏

seankane-msft (approved):
LGTM, thanks @sadasant

sdk/azidentity/chained_token_credential.go
```diff
@@ -44,24 +45,38 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+		var authFailed AuthenticationFailedError
+		if errors.As(err, &authFailed) {
+			err = fmt.Errorf("Authentication failed:\n%s\n%s"+createChainedErrorMessage(errList), err)
+			authErr := newAuthenticationFailedError(err, authFailed.RawResponse())
+			return authErr
+		}
+		return err
+	}
+
+	if c.successfulCredential != nil {
```
jhendrixMSFT, line 59: The linked issue mentions adding a bool to the options so callers can opt in to the old behavior.  Will that also be added?

sadasant: thank you! I missed it

sadasant: @jhendrixMSFT let’s talk about names! I’m terrible with names. Some ideas: `doNotSaveSuccessfulCredential` ? `alwaysTryAllSources`? uhh… `alwaysTestSources`? `ignoreSuccessfulCredential`? (help)

jhendrixMSFT: Hmm maybe `EnableAllSources`?  I feel like `Sources` should be in the name given that the parameter name is `sources` when creating the ChainedTokenCredential.

CC @JeffreyRichter

sdk/azidentity/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Instances of `ChainedTokenCredential` will now skip looping through the list of source credentials and re-use the first successful credential on subsequent calls to `GetToken`.
```
jhendrixMSFT, line 9: The docs for `NewChainedTokenCredential` should be updated to describe this behavior.

sadasant: Good point!

sdk/azidentity/chained_token_credential.go
sadasant: `EnableAllSources` definitely implies that the default is that not all sources are enabled 🤔

sadasant: What about `RetryAllSources`?

jhendrixMSFT: I like that.  Let's start with that and we can refine it.

sadasant: On my way!

sdk/azidentity/chained_token_credential.go
```diff
@@ -14,15 +14,22 @@ import (
...
 // ChainedTokenCredentialOptions contains optional parameters for ChainedTokenCredential.
 type ChainedTokenCredentialOptions struct {
-	// placeholder for future options
+	// If true, it will not assume the first successful credential should be always used.
+	RetryAllSources bool
 }
 
 // ChainedTokenCredential is a chain of credentials that enables fallback behavior when a credential can't authenticate.
+// By default, this credential will assume that the first successful credential should be the only credential used on future requests.
+// When `retryAllSources` is true, it will always try to get a token with every credential available on the `sources` array.
```
jhendrixMSFT, line 23: Since `retryAllSources` isn't exported, we don't need to publicly doc it (plus it's doc'ed in the options).

sdk/azidentity/chained_token_credential.go
```diff
@@ -36,32 +43,52 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+			err = fmt.Errorf("Authentication failed:\n%s\n%s"+createChainedErrorMessage(errList), err)
+			authErr := newAuthenticationFailedError(err, authFailed.RawResponse())
+			return authErr
+		}
+		return err
+	}
+
+	if c.successfulCredential != nil {
+		token, err = c.successfulCredential.GetToken(ctx, opts)
+		if err != nil {
```
jhendrixMSFT, line 71: I think this block needs to be rewritten to handle the `retryAllSources == true` case yes?

sadasant: I don’t think so! As long as `successfulCredential` is never set, this part will never be used. In my current approach I’m not setting `successfulCredential` when `retryAllSources` is set, see here: https://github.com/Azure/azure-sdk-for-go/pull/16392/files#diff-d1472b3315b6cb3473bb118e734a540711129ee1b51d4f3f89b5eca7fd2c4295R87

sadasant: Of course, this could be changed so that `successfulCredential` is always set, but the conditional on line 69 is assumed to be false if `retryAllSources` is set. What makes the most sense?

jhendrixMSFT: I get it now.  The way I was thinking about this is that we'd always use `successfulCredential`, and if it were to suddenly start failing on future requests, we'd fall back to try the others if `retryAllSources` is `true`.  The way you have it now preserves the old behavior which is probably better anyways in case anybody depens on that.

sadasant: Doing the change!

sadasant: Pushed an update here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/a1bbdd1629549e699082252aefd84aa5376c3f83

sdk/azidentity/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Breaking Changes
 
+* Instances of `ChainedTokenCredential` will now skip looping through the list of source credentials and re-use the first successful credential on subsequent calls to `GetToken`.
+  * If the `RetryAllSources` option is set to true, it will not assume the first successful credential should be always used.
```
sadasant, line 10:
```suggestion
  * If the `RetryAllSources` option is set to true, it will not assume the first successful credential should be always used. It will continue to try all of the originally provided credentials each time the `GetToken` method is called.
```

sdk/azidentity/chained_token_credential.go
```diff
@@ -14,15 +14,20 @@ import (
...
 type ChainedTokenCredential struct {
-	sources []azcore.TokenCredential
+	sources              []azcore.TokenCredential
+	successfulCredential azcore.TokenCredential
+	retryAllSources      bool
 }
 
 // NewChainedTokenCredential creates a ChainedTokenCredential.
+// By default, this credential will assume that the first successful credential should be the only credential used on future requests.
+// If the `RetryAllSources` option is set to true, it will always try to get a token using all of the available credentials.
```
sadasant, line 30:
```suggestion
// If the `RetryAllSources` option is set to true, it will always try to get a token using all of the originally provided credentials.
```

sdk/azidentity/chained_token_credential.go
```diff
@@ -14,15 +14,20 @@ import (
 
 // ChainedTokenCredentialOptions contains optional parameters for ChainedTokenCredential.
 type ChainedTokenCredentialOptions struct {
-	// placeholder for future options
+	// If true, it will not assume the first successful credential should be always used.
+	RetryAllSources bool
```
RickWinter, line 18: Is this really a retry of all sources?  It still stops at the first succesfull match.  Would `RetrySources` make more sense

sadasant: Good thinking! I like RetrySources!

sadasant: Pushed the change here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/e288440b96d230c971216758a8bc5881d9241141

sdk/azidentity/chained_token_credential.go
```diff
@@ -14,15 +14,20 @@ import (
...
 // ChainedTokenCredential is a chain of credentials that enables fallback behavior when a credential can't authenticate.
 type ChainedTokenCredential struct {
-	sources []azcore.TokenCredential
+	sources              []azcore.TokenCredential
+	successfulCredential azcore.TokenCredential
+	retrySources         bool
 }
 
 // NewChainedTokenCredential creates a ChainedTokenCredential.
+// By default, this credential will assume that the first successful credential should be the only credential used on future requests.
```
chlowell, line 29: This would be better on the type declaration (rendered docs are [here](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ChainedTokenCredential)).

sdk/azidentity/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Breaking Changes
 
+* Instances of `ChainedTokenCredential` will now skip looping through the list of source credentials and re-use the first successful credential on subsequent calls to `GetToken`.
```
chlowell, line 9: Worth mentioning `DefaultAzureCredential` here--this PR makes the same change to it. The difference is that the behavior is configurable for `ChainedTokenCredential`.

sdk/azidentity/chained_token_credential.go
```diff
@@ -36,32 +41,50 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+			authErr := newAuthenticationFailedError(err, authFailed.RawResponse())
+			return authErr
+		}
+		return err
+	}
+
+	if c.successfulCredential != nil && !c.retrySources {
+		token, err = c.successfulCredential.GetToken(ctx, opts)
+		if err != nil {
+			return nil, formatError(err)
```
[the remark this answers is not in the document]
sadasant: Good point! thank you!

sdk/azidentity/CHANGELOG.md
sadasant: I’m assuming you meant to add `RetrySources` to the `DefaultAzureCredential`, is that correct? I added it here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/e2d1201e28ba824b2b640a5eda0b1be1c5120fbc

sdk/azidentity/chained_token_credential.go
sadasant: I moved it here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/e2d1201e28ba824b2b640a5eda0b1be1c5120fbc let me know how it looks!

sdk/azidentity/chained_token_credential.go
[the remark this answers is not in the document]
sadasant: Done here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/e2d1201e28ba824b2b640a5eda0b1be1c5120fbc

sdk/azidentity/CHANGELOG.md
chlowell: What I meant is that this entry should also name `DefaultAzureCredential` because its default behavior is changing in the same way. `DefaultAzureCredential` uses `ChainedTokenCredential`, so changing the latter's default behavior changes the former's as well. No problem, that's what we want, I just think it deserves a mention here.

We don't want `DefaultAzureCredential` to have configuration for this behavior though. It should always remember the first successful credential and use that exclusively.

sadasant: Oh sorry! I got that wrong. I’ll undo the change 👍

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -153,6 +153,220 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+	secretCountPolicy := &TestCountPolicy{}
+	environmentCountPolicy := &TestCountPolicy{}
+
+	srv.AppendResponse(mock.WithBody([]byte(accessTokenRespSuccess)))
+	srv.AppendResponse(mock.WithBody([]byte(accessTokenRespSuccess)))
+	options := ClientSecretCredentialOptions{}
+	options.AuthorityHost = AuthorityHost(srv.URL())
+	options.PerCallPolicies = []policy.Policy{secretCountPolicy}
+	options.Transport = srv
+	secCred, err := NewClientSecretCredential(tenantID, clientID, secret, &options)
```
chlowell, line 230: Suggest using fake credentials instead of real ones, to simplify the test and keep it focused on ChainedTokenCredential's behavior

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -153,6 +153,220 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+	if err != nil {
+		t.Fatalf("Received an error when attempting to get a token but expected none")
+	}
+	if tk.Token != tokenValue {
+		t.Fatalf("Received an incorrect access token")
+	}
+	if tk.ExpiresOn.IsZero() {
+		t.Fatalf("Received an incorrect time in the response")
+	}
+	if cred.successfulCredential == nil {
```
chlowell, line 258: You don't need to test implementation details like this--counting GetToken calls tells you whether ChainedTokenCredential behaved correctly--and doing so makes the test more fragile.

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -153,6 +153,220 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
 	}
 }
 
+func TestChainedTokenCredential_ChecksThatSuccessfulCredentialIsSet(t *testing.T) {
```
chlowell, line 156: Does this test anything the case below does not?

sadasant: No, just trying to have smaller tests. I’ll remove this one.

sdk/azidentity/chained_token_credential_test.go
sadasant: How does this look? https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/20197a94527b5dc38f8d0270df23c0fccff47e68

sdk/azidentity/chained_token_credential_test.go
sadasant: I’m confused about what you’re requesting. I like the idea of checking that this property is set, and counting getToken calls. In any case, please review this new approach! https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/20197a94527b5dc38f8d0270df23c0fccff47e68 It should be better than the previous one 🙂

sadasant:
Is this CI error related to this PR? https://dev.azure.com/azure-sdk/public/_build/results?buildId=1242645&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=bdeefc16-b669-5ebd-ad94-a2c19ade53b0

jhendrixMSFT:
It's due to a recent check that was added in https://github.com/Azure/azure-sdk-for-go/pull/16265.  I will turn this off for now.

jhendrixMSFT:
This has been fixed, please update/rebase to `main` and you should be good to go.

sdk/azidentity/chained_token_credential_test.go
chlowell: What we want to test is the credential's behavior: does it call its sources' GetToken methods correctly? The count of calls to each source's GetToken method is a direct observation of that. The value of `successfulCredential` is part of that behavior's implementation, not an observation of it. Checking the value of `successfulCredential` like this is validating whether the behavior is implemented the way you expect. The implementation isn't important, and we don't want to bake it into the test. So, I suggest not using `successfulCredential` in these tests.

sadasant: Why is the implementation not important? I wonder what scenarios you’re considering. At a minimum it should have the credential that succeeds 🤔

chlowell: If we change the implementation such that `successfulCredential` no longer exists, this test won't compile. If the credential still behaves correctly, is the broken test helpful?

sdk/azidentity/chained_token_credential.go
```diff
@@ -36,32 +44,49 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+	return &ChainedTokenCredential{sources: cp, retrySources: options.RetrySources}, nil
 }
 
 // GetToken calls GetToken on the chained credentials in turn, stopping when one returns a token. This method is called automatically by Azure SDK clients.
 // ctx: Context controlling the request lifetime.
 // opts: Options for the token request, in particular the desired scope of the access token.
 func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (token *azcore.AccessToken, err error) {
 	var errList []CredentialUnavailableError
+
+	formatError := func(err error) error {
```
chlowell, line 59: This is called once, may as well inline it

sadasant: Done here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/e14544a86e1c0e5344a0efe86e158af843dd46f1

sdk/azidentity/chained_token_credential_test.go
chlowell: > Even then, these tests would be easy to fix.
> 
It's still work. If we remove `successfulCredential` and the credential still behaves correctly, why would you want the test to break?

> In any case, part of the proposal of this PR is to expose the `successfulCredential` to the public. If that’s not part of the goal, we could take a different direction.
>
That's not the goal or the current state (`successfulCredential` isn't exported).

sadasant: oh I see your point! I’ll remove it from these tests.

sadasant: Updated here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/a19c2f27dcbdb3efe444e80cdf7bc59829f0a577 let me know if I missed anything!

sdk/azidentity/chained_token_credential.go
```diff
@@ -14,12 +14,20 @@ import (
 
 // ChainedTokenCredentialOptions contains optional parameters for ChainedTokenCredential.
 type ChainedTokenCredentialOptions struct {
-	// placeholder for future options
+	// RetrySources configures how the credential uses its sources.
+	// When true, the credential will always request a token from each source in turn,
+	// stopping when one provides a token. When false, the credential requests a token
+	// only from the source that previously retrieved a token--it never again tries the sources which failed.
+	RetrySources bool
```
chlowell, line 21: I doubt we'll find a perfect name for this, but I wonder whether we can come up with a more self-explanatory one. It controls whether `GetToken` always iterates over the sources, which has me thinking along the lines of `AlwaysIterate`, `AlwaysStartWithFirstSource`, or `AlwaysStartAtBeginning`. It would be nice if this option paired with the (opposite) option in other SDKs, so something like DoNot/PreferFirstSuccessfulSource or Forget/RememberFirstSuccessfulSource, but that's probably too high a bar. I'm just trying to start a brainstorm here.

sdk/azidentity/chained_token_credential.go
```diff
@@ -36,14 +44,25 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
 // opts: Options for the token request, in particular the desired scope of the access token.
 func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (token *azcore.AccessToken, err error) {
 	var errList []CredentialUnavailableError
+
+	if c.successfulCredential != nil && !c.retrySources {
+		token, err = c.successfulCredential.GetToken(ctx, opts)
+		if err != nil {
+			return nil, err
+		}
+		return token, nil
```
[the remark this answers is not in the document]
sadasant: oh thanks!

sdk/azidentity/chained_token_credential.go
sadasant: What about `cacheSuccessfulCredential` ?

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -140,3 +140,118 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+// TestCredential response
+type TestCredentialResponse struct {
+	token *azcore.AccessToken
+	err   error
+}
+
+// Credential used for testing
+type TestCredential struct {
+	getTokenCalls int
+	responses     []TestCredentialResponse
```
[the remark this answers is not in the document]
sadasant: Limiting it to just one repeated response makes it way less useful I think 🤔

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -140,3 +140,118 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+func (c *TestCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (token *azcore.AccessToken, err error) {
+	index := c.getTokenCalls
+	c.getTokenCalls += 1
+	response := c.responses[index]
+	return response.token, response.err
+}
+
+func TestChainedTokenCredential_RepeatedGetTokenWithSuccessfulCredential(t *testing.T) {
+	failedCredential := &TestCredential{responses: []TestCredentialResponse{
+		{err: newCredentialUnavailableError("MockCredential", "Mocking a credential unavailable error")},
```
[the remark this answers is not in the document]
sadasant: I’ll remove the unavailableCredential test 🤔

sdk/azidentity/chained_token_credential_test.go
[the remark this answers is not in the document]
chlowell: In principle, sure, but in fact that's the only way it's used 😜

sdk/azidentity/chained_token_credential_test.go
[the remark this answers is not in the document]
sadasant: Updated here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/2080f9635cba20b79f2fc07929a6dd0f9e108ac4

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -140,3 +140,118 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
...
+	if tk.ExpiresOn.IsZero() {
+		t.Fatalf("Received an incorrect time in the response")
+	}
+	if failedCredential.getTokenCalls != 1 {
+		t.Fatalf("The failed credential getToken should have been called once")
+	}
+	if successfulCredential.getTokenCalls != 1 {
+		t.Fatalf("The successful credential getToken should have been called once")
+	}
+	tk2, err2 := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{liveTestScope}})
```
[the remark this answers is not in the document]
sadasant: I don’t think a loop is the best approach for a test that has only two steps. What about abstracting out some of the checks into a function? I’ve pushed that idea here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/2080f9635cba20b79f2fc07929a6dd0f9e108ac4

sdk/azidentity/chained_token_credential_test.go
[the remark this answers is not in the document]
sadasant: Making GetToken return a static response seems to indicate that developers can alter that static response to change the way GetToken is answering which may end up being worse later.

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -140,3 +140,118 @@ func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *test
 		t.Fatalf("Received an incorrect time in the response")
 	}
 }
+
+// TestCredential response
+type TestCredentialResponse struct {
```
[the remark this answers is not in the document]
sadasant: Updated here: https://github.com/Azure/azure-sdk-for-go/pull/16392/commits/2080f9635cba20b79f2fc07929a6dd0f9e108ac4

sdk/azidentity/chained_token_credential.go
jhendrixMSFT: It would be good to name it in such a way that its zero-value (false) is the preferred behavior.  Calling it `CacheSuccessfulCredential` would mean we'd need to set it to `true` by default.  I think `RetrySources` is reasonable, however we could rename it before release.

jhendrixMSFT (approved):
Let's wait for @chlowell to approve and merge.

sdk/azidentity/chained_token_credential_test.go
```diff
@@ -113,30 +113,103 @@ func TestChainedTokenCredential_GetTokenFail(t *testing.T) {
...
-func TestChainedTokenCredential_GetTokenWithUnavailableCredentialInChain(t *testing.T) {
-	secCred, err := NewClientSecretCredential(fakeTenantID, fakeClientID, secret, nil)
+func (c *TestCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (token *azcore.AccessToken, err error) {
+	index := c.getTokenCalls
+	c.getTokenCalls += 1
+	response := c.responses[index]
+	return response.token, response.err
+}
+
+func testGoodGetTokenResponse(t *testing.T, token *azcore.AccessToken, err error) {
```
chlowell, line 135: This is almost the same as `testGetTokenSuccess` in `live_test.go`

chlowell (approved):

sdk/azidentity/chained_token_credential_test.go
sadasant: (We talked over teams and we’ll keep this how it is in this PR at the moment)

PR #16393, opened 2021-12-02
[messaging] The parsed connection doesn't have a .Host field anymore

richardpark-msft:
Changed this a bit ago, but @benbp got our CI pipeline running and found that the tests didn't work.

Technically these things are tested in azservicebus, but we should have these tests working anyways.


seankane-msft (approved):
So I take it `parsed.Namespace` == `"amqps://"` + namespace then?

benbp (approved):

richardpark-msft:
> So I take it `parsed.Namespace` == `"amqps://"` + namespace then?

The hostname is basically "protocol independent" since you should be able to use amqps or https (for websockets). So we were always having to do some manipulation anyways. Seemed better to just strip it off and no longer pretend that it mattered when we were getting it from a connection string.

richardpark-msft:
Check enforcer doesn't seem to be able to correlate the CI build with the PR but it passed:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1229912&view=results

I'm going to override check-enforcer.

PR #16396, opened 2021-12-03
[azservicebus] Invert the dependency for websockets so we don't directly depend on a 3rd party package

### LINKED ISSUE #16201: [azservicebus] Expose websocket support
We have some code in the azservicebus package that's not exposed, which will allow for websocket support.

Currently we're not exposing it but before we do I'd like to consider exposing an interface to allow the customer to inject their own websocket provider. We can still provide the 5 lines of code to do the link up themselves, but this would remove a required dependency in our library.

richardpark-msft:
We were previously depending on a 3rd party package for websockets, but our interaction with it was pretty minimal. I've reduced this down to just a single function call and allow a customer to pass in that function so we can invert the dependency and not directly require a specific version of the websockets library, provided it understands how to create a net.Conn.

I've moved the code we used to have in the library into an example instead, so it's still a simple operation for a customer to join it up with a package of their choice.

Fixes #16201

### DIFF
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -20,7 +20,7 @@ import (
 // Client provides methods to create Sender and Receiver
 // instances to send and receive messages from Service Bus.
 type Client struct {
-	config    clientCreds
+	creds     clientCreds
 	namespace interface {
 		// used internally by `Client`
 		internal.NamespaceWithNewAMQPLinks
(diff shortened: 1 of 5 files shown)


sdk/messaging/azservicebus/client.go
```diff
@@ -37,6 +38,10 @@ type Client struct {
 type ClientOptions struct {
 	// TLSConfig configures a client with a custom *tls.Config.
 	TLSConfig *tls.Config
+
+	// NewWebSocketConn is a function that can create a net.Conn for use with websockets.
+	// For an example, see ExampleNewClient_usingWebsockets() function in example_client_test.go.
+	NewWebSocketConn func(ctx context.Context, wssHost string) (net.Conn, error)
```
jhendrixMSFT, line 44: I wonder if this can be abstracted away a bit further so that we can avoid having to specify any params here (will all WS clients need just a context and host?).  What if this was just `WebSocketConn net.Conn` so callers have to create this ahead of time?

richardpark-msft: > (will all WS clients need just a context and host?). 

That's a good question. I only have the one websocket library we use today as the example. One thing I can do is put all the parameters into a struct so we can expand it in the future if needed.

>  What if this was just WebSocketConn net.Conn so callers have to create this ahead of time?

The built-in recovery would need to be able to reconnect (depending on what's needed) so we can't rely on a single instance created beforehand.

Also, requiring them to instantiate it up-front would (potentialy) break the ability for them to use this in the same 'lazy initialized' way as the non-websocket path.

jhendrixMSFT: That's a good point.

I thought about using a struct but it seems a bit clunky, maybe that's not too bad.

Where does the value for `wssHost` come from?

richardpark-msft: It's formed internally by our code. Technically a user could probably figure it out, so maybe we could omit it but we'd probably still want to pass in an parameters-bag anyways. So it seems like a sensible thing to give them.

The example shows how they'd use it and I think it makes life simpler for them.

richardpark-msft: Updated to pass in an 'args' instead.

sdk/messaging/azservicebus/client.go
```diff
@@ -75,34 +80,21 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
...
-		return nil
-	}
-
-	client.config.tlsConfig = options.TLSConfig
-	return nil
+	credential              azcore.TokenCredential
 }
 
-func newClientImpl(config clientConfig, options *ClientOptions) (*Client, error) {
+func newClientImpl(config clientCreds, options *ClientOptions) (*Client, error) {
```
RickWinter, line 94: Is there a reason the client param is named `config` when it's now a cred?

richardpark-msft: Fixed (it just lagged behind my type rename)

jhendrixMSFT (approved):

PR #16529, opened 2021-12-08
[Tables] Address Doc Bug Bash comments

### LINKED ISSUE #16494: [Tables] Old Storage package doesn't link to aztables
As part of the docs bug bash, I attempted to find the Tables SDK from the root of the repo. After searching for "tables" and finding a Table service client in the [old Storage package](https://github.com/Azure/azure-sdk-for-go/tree/main/storage) I learned that the package is going to be deprecated. Since there's no link to the `aztables` package, though, it made me think that the Table service client must be part of one of the three packages linked, rather than in a separate Tables SDK.

It might be helpful, if possible, to clear up the story of where new storage packages are and include `aztables` in the readme of the old package.

### LINKED ISSUE #16496: [Tables] Ask users to log into Azure CLI before running commands
We tell users that they can run an Azure CLI command in the [Create the client](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/data/aztables#create-the-client) section of the Tables readme, but we don't specify that `az login` needs to be run first to avoid an auth error. It's a minor detail, but new users might not know how to resolve the issue.

seankane-msft:
closes #16496 

* specify to login to az cli
* Added a section on authenticating with shared key that was missing

### DIFF
--- a/sdk/data/aztables/README.md
+++ b/sdk/data/aztables/README.md
@@ -32,6 +32,8 @@ The Azure Tables library allows you to interact with two types of resources:
 Interaction with these resources starts with an instance of a [client](#clients). To create a client object, you will need the account's table service endpoint URL and a credential that allows you to access the account. The `endpoint` can be found on the page for your storage account in the [Azure Portal][azure_portal_account_url] under the "Access Keys" section or by running the following Azure CLI command:
 
 ```bash
+# Log in to Azure CLI first, this opens a browser window
+az login
 # Get the table service URL for the account
 az storage account show -n mystorageaccount -g MyResourceGroup --query "primaryEndpoints.table"
 ```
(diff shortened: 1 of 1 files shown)


sdk/data/aztables/README.md
```diff
@@ -58,6 +60,14 @@ To use an account [shared key][azure_shared_key] (aka account key or access key)
...
 ```
 
+```golang
+cred, err := aztables.NewSharedKeyCredential("<accountName>", "<accountKey>")
+handle(err)
+client, err := aztables.NewServiceClientWithSharedKey(serviceURL, cred, nil)
+handle(err)
+```
+
+##### Creating the client with an AAD credential
```
jhendrixMSFT, line 70: I believe we want to place the ADD cred before the shared key as we try to promote AAD over shared key.

seankane-msft: Fixed

jhendrixMSFT (approved):

mccoyp (approved):
This should cover https://github.com/Azure/azure-sdk-for-go/issues/16496 👍

PR #16530, opened 2021-12-08
[azservicebus] Updating go-amqp, fixing several bugs

### LINKED ISSUE #15163: [service-bus] Sending and receiving messages using sessions
Similar to #15093 but with sessions.

Work:
- [x] AcceptSessionFor(Queue|Subscription)(<session-id>) - accepts a session from a session-based entity with a specific ID. This method is _not_ lazy - the link is opened in this call. 
- [x] AcceptNextSession(Queue|Subscription) - accepts the next available session from a session-based entity.
- [x] The Go Sender currently associates a session ID with the sender itself, which none of the other track 2 senders do. This should be removed.
- [x] ~~#16079~~ (moved to it's own issue)
- [x] #15773

Notes:
- [ ] An empty session ID should be valid for AcceptSession and for sending messages: https://github.com/Azure/azure-sdk-for-net/pull/24570

### LINKED ISSUE #16340: azservicebus receiveMessages never finishes when network is lost, gets stuck in drainLink
This bug is regarding the azservicebus data plane API.
I'm using `github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v0.3.1`

I am having a problem with the following code:

```go
	var notification types.FtpUploadNotification
	ctx, cancel := context.WithTimeout(context.TODO(), 20 * time.Second)
	defer cancel()

	messages, err := servicebus_handler.ReceiveMessages(ctx, 1, nil)
	if err != nil {
		return notification, err
	} else if len(messages) == 1 {
		body, err := messages[0].Body()
		if err == nil {
			json.Unmarshal(body, &notification)
			err := servicebus_handler.CompleteMessage(ctx, messages[0])
			if err != nil {
				panic(err)
			}
		}
		return notification, nil
	} else {
		return notification, errors.New("no message received")
	}
```

When during the timeout of 20 sec a network interruption happens the code is stuck in the following part of `receiver.go` in `receiver.DrainCredit(context.Background())`(Line 376).

```go
func (r *Receiver) drainLink(ctxForLoggingOnly context.Context, receiver internal.AMQPReceiver, messages []*ReceivedMessage) ([]*ReceivedMessage, error) {
	// start the drain asynchronously. Note that we ignore the user's context at this point
	// since draining makes sure we don't get messages when nobody is receiving.
	if err := receiver.DrainCredit(context.Background()); err != nil {
		tab.For(ctxForLoggingOnly).Debug(fmt.Sprintf("Draining of credit failed. link will be closed and will re-open on next receive: %s", err.Error()))

		// if the drain fails we just close the link so it'll re-open at the next receive.
		if err := r.amqpLinks.Close(context.Background(), false); err != nil {
			tab.For(ctxForLoggingOnly).Debug(fmt.Sprintf("Failed to close links on ReceiveMessages cleanup. Not fatal: %s", err.Error()))
		}
	}
```

The parameter `ctxForLoggingOnly` is containing the 20 sec timeout but is not used in the `receiver.DrainCredit(context.Background())` call. When I replace the `context.Background())` with `ctxForLoggingOnly` it is respecting the timeout and tries to reconnect in the outer code.

Can i somehow archive it in another way in the calling code?

Thanks in advance,
Dominic

richardpark-msft:
Fixes two issues:
- go-amqp v0.17.0 update handles cancelling drain when the link detaches
- and now we can also handle empty session IDs! Test added and reflected in the public API contract.

Fixes #16340
Fixes #15163


sdk/messaging/azservicebus/message.go
```diff
@@ -119,29 +119,34 @@ func (m *Message) toAMQPMessage() *amqp.Message {
...
 	// TODO: I don't think this should be strictly required. Need to
 	// look into why it won't send properly without one.
-	var messageID = m.MessageID
+	var messageID string
 
-	if messageID == "" {
+	if m.MessageID == nil || *m.MessageID == "" {
 		uuid, err := uuid.New()
 
 		if err == nil {
```
jhendrixMSFT, line 127: What happens if `err != nil`?

richardpark-msft: I actually think I can get rid of this workaround entirely now. I'll try it out real quick.

richardpark-msft: Ah, I was! There were just a few little spots, but it works great. The ID is still auto-generated (we're not doing it anymore so it's just happening on the Service Bus side).

There was one spot I was concerned about (the RPC Link) but we're auto-generating IDs in there separately so we should be good.

jhendrixMSFT (approved):

PR #16535, opened 2021-12-08
[KeyVault] improving test code coverage

seankane-msft:
Improving test code coverage and upgrading to internal@v0.8.3

### DIFF
--- a/sdk/keyvault/azsecrets/models.go
+++ b/sdk/keyvault/azsecrets/models.go
@@ -88,9 +88,13 @@ type Attributes struct {
 }
 
 func (s Attributes) toGenerated() *internal.SecretAttributes {
+	var recLevel *internal.DeletionRecoveryLevel
+	if s.RecoveryLevel != nil {
+		recLevel = s.RecoveryLevel.toGenerated().ToPtr()
+	}
 	return &internal.SecretAttributes{
 		RecoverableDays: s.RecoverableDays,
-		RecoveryLevel:   s.RecoveryLevel.toGenerated().ToPtr(),
+		RecoveryLevel:   recLevel,
 		Attributes: internal.Attributes{
 			Enabled:   s.Enabled,
 			Expires:   s.Expires,
--- a/sdk/keyvault/azsecrets/client_test.go
+++ b/sdk/keyvault/azsecrets/client_test.go
@@ -270,7 +270,7 @@ func TestUpdateSecretProperties(t *testing.T) {
 	client, err := createClient(t)
 	require.NoError(t, err)
 
-	secret, err := createRandomName(t, "secret")
+	secret, err := createRandomName(t, "secret1" + t.Name())
 	require.NoError(t, err)
 	value, err := createRandomName(t, "value")
 	require.NoError(t, err)
@@ -284,10 +284,17 @@ func TestUpdateSecretProperties(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, *getResp.Value, value)
 
+	expires := time.Now().Add(48 * time.Hour)
+	nb := time.Now().Add(-24 * time.Hour)
 	params := Properties{
 		ContentType: to.StringPtr("password"),
 		Tags: map[string]string{
 			"Tag1": "TagVal1",
+		},
+		SecretAttributes: &Attributes{
+			Enabled: to.BoolPtr(true),
+			Expires: &expires,
+			NotBefore: &nb,
 		},
 	}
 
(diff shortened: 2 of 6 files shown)


sdk/keyvault/azsecrets/models.go
```diff
@@ -88,9 +88,13 @@ type Attributes struct {
 }
 
 func (s Attributes) toGenerated() *internal.SecretAttributes {
+	var recLevel *internal.DeletionRecoveryLevel
+	if s.RecoveryLevel != nil {
```
jhendrixMSFT, line 92: Is this another case where the `toGenerated()` can be simplified?

sdk/keyvault/azsecrets/client.go
```diff
@@ -271,11 +271,8 @@ func (s *startDeleteSecretPoller) Poll(ctx context.Context) (*http.Response, err
 		// Service recognizes DeletedSecret, operation is done
 		s.lastResponse = resp
 		return resp.RawResponse, nil
-	} else if err != nil {
```
jhendrixMSFT, line 274: Why was this removed?  Looks like the error is now swallowed.

sdk/keyvault/azkeys/client.go
```diff
@@ -619,11 +601,9 @@ func (s *startDeleteKeyPoller) Poll(ctx context.Context) (*http.Response, error)
 		// Service recognizes DeletedKey, operation is done
 		s.lastResponse = resp
 		return resp.RawResponse, nil
-	} else if err != nil {
```
jhendrixMSFT, line 622: Another swallowed error?

sdk/keyvault/azkeys/client.go
```diff
@@ -56,7 +55,7 @@ func NewClient(vaultUrl string, credential azcore.TokenCredential, options *Clie
 
 	genOptions.PerRetryPolicies = append(
 		genOptions.PerRetryPolicies,
-		shared.NewKeyVaultChallengePolicy(credential, runtime.NewPipeline("azkeys", "0.1.0", nil, nil, genOptions)),
```
jhendrixMSFT, line 59: I see the pending 0.1.1 release of the internal module requires this pipeline parameter.  What's the plan for releasing that?

sdk/keyvault/azsecrets/models.go
seankane-msft: This is the same as the other PR, fixed after merging conflicts

sdk/keyvault/azkeys/client.go
seankane-msft: The internal module will be released before `azsecrets`/`azkeys`, then the change will be made in both keys and secrets

sdk/keyvault/azsecrets/client.go
seankane-msft: The aboe error checked if `err == nil` so the return statement after the else was deadcode. I removed the else statement. I need to add a check to make sure the error is because of an expected error (this should return a 409 does not exist while polling), thanks for catching that.

seankane-msft: Fixed

sdk/keyvault/azkeys/client.go
seankane-msft: Fixed

sdk/keyvault/azsecrets/client.go
```diff
@@ -271,11 +271,15 @@ func (s *startDeleteSecretPoller) Poll(ctx context.Context) (*http.Response, err
...
-		return s.deleteResponse.RawResponse, nil
 	}
-	s.lastResponse = resp
-	return resp.RawResponse, nil
+	var httpResponseErr azcore.HTTPResponse
+	if errors.As(err, &httpResponseErr) {
+		if httpResponseErr.RawResponse().StatusCode == 404 {
```
jhendrixMSFT, line 277: `http.StatusNotFound`

sdk/keyvault/azkeys/client_test.go
```diff
@@ -547,21 +639,21 @@ func TestReleaseKey(t *testing.T) {
 			req, err := http.NewRequest("GET", fmt.Sprintf("%s/generate-test-token", attestationURL), nil)
 			require.NoError(t, err)
 
-			resp, err := http.DefaultClient.Do(req)
-			require.NoError(t, err)
-			require.Equal(t, resp.StatusCode, http.StatusOK)
-			defer resp.Body.Close()
+			_, err = http.DefaultClient.Do(req)
+			require.Error(t, err)
+			// require.Equal(t, resp.StatusCode, http.StatusOK)
```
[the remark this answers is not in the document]
seankane-msft: This is waiting on the next api version to go into preview, once it does the uncommented portions will be ready.

sdk/keyvault/azsecrets/client.go
seankane-msft: Done

PR #16536, opened 2021-12-08
[azservicebus] Making the message batch goroutine safe.

richardpark-msft:


sdk/messaging/azservicebus/message_batch.go
```diff
@@ -16,6 +17,8 @@ var ErrMessageTooLarge = errors.New("the message could not be added because it i
 type (
 	// MessageBatch represents a batch of messages to send to Service Bus in a single message
 	MessageBatch struct {
+		mu sync.RWMutex
```
jhendrixMSFT, line 20: Is it common for batches to be concurrently used?

richardpark-msft: Don't know if it's common, but we've made these concurrency safe in languages: C#, JS (by default).

I've got a question out to the Java folks since it appears theirs is not, but it might just be a lack of Java understanding on my part.

richardpark-msft: TBH, I don't know. I think we've had a policy of making sure these are thread-safe in other languages (Java, C#, JS (well, you know by default)).

It basically functions as a write-only collection, so thread/goroutine safety makes sense to me.

jhendrixMSFT (approved):

PR #16615, opened 2021-12-10
Disable the Dependency Check step until GA

jhendrixMSFT:

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -10,6 +10,7 @@ steps:
 
   - task: Powershell@2
     displayName: 'Dependency Check'
+    condition: false
     env:
       GO111MODULE: 'on'
     inputs:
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@benbp if there's a better way to disable this step please let me know.

benbp:
@jhendrixMSFT this will work fine as a temporary disable method. FYI @tadelesh @seankane-msft.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -10,6 +10,7 @@ steps:
 
   - task: Powershell@2
     displayName: 'Dependency Check'
+    condition: false
```
weshaggard, line 13:
```suggestion
    continueOnError: true
```

You could also just continue on error so the step runs but doesn't fail the build.

jhendrixMSFT: Good idea, fixed.

benbp (approved):

PR #16657, opened 2021-12-15
Smoke test hook

ckairen:
Adding in the hook for smoke test scripts in the pipeline


eng/pipelines/templates/jobs/new.smoke.tests.yml
```diff
@@ -0,0 +1,32 @@
...
+    type: boolean
+    default: true
+  - name: TimeoutInMinutes
+    type: number
+    default: 60
+
+jobs:
+  - job: SmokeTest
+    steps:
+      - ${{ if eq(parameters.Daily, true) }}:
```
benbp, line 15: Is it worth separating these vs. calling a single script with a parameter `-Daily`?

eng/pipelines/templates/jobs/new.smoke.tests.yml
```diff
@@ -0,0 +1,32 @@
...
+        - task: PowerShell@2
+          displayName: 'Run Nightly SmokeTests'
+          inputs:
+            targetType: 'filePath'
+            filePath: eng/scripts/Smoke_Tests_Nightly.ps1
```
benbp, line 20: If we go with this approach, I'd call this script daily.

eng/pipelines/templates/jobs/new.smoke.tests.yml
seankane-msft: I think putting the daily parameter into the script that runs the smoke tests is worthwhile, I can make that change

seankane-msft: I have it set up so passing in `-serviceDirectory .` will run the Daily version, and passing in a serviceDirectory (ie `data/aztables`) will run smoke tests only for that package. Thoughts on this approach?

benbp: I think eventually we'll want to have the option to run the GA/beta validation against ALL packages rather than just a single service directory, so having the explicit boolean makes it a little clearer what the behavior is going to be and easier to update in the future.

seankane-msft:
@ckairen I merged my PR (#16735), can you update this PR by pulling in the new `main` branch?

Issue #16108, opened 2021-11-05
Add an `ErrorCode()` method to azcore.HTTPResponse


mikekistler:
### Feature Request

The [Azure API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors) require that all error responses should contain an `error` object that contains a string `code` and string `message` fields.  The `code` field is what we want customers to use when checking for specific errors; the `message` field is just an English description of `code` but is not part of the "API contract".

To make it easier for users to find the error code, there should be a method on `azcore.HTTPResponse` extracts the code from the response body (or from the `x-ms-error-code` header if it is available). This logic will be a bit complicated and we don't want to force users to have to write it themselves (leading to 1000s of implementations).

When the response does not contain a `code` field (contrary to the guidelines, but certainly happens), the method should simply return an empty string.

This method will enable an error checking pattern like
```
	_, err := container.Create(ctx, nil)
	var responseErr azcore.HTTPResponse
	if errors.as(err, &responseErr) {
		if responseErr.ErrorCode() == "ContainerAlreadyExists") {
			// Ignore - the container exists so processing can continue
		} else {
			fmt.Printf("Error: %s", err.Error())
			panic("Cannot access the container")
		}
	} else {
		fmt.Printf("Error: %s", err.Error())
		panic("Cannot access the container")
	}
```

We should make sure there is good documentation for this pattern, since it is not something that users can find through intellisense.  In particular, the following docs should be updated to describe this new method:

- The example code shown in the [docs for `azcore.HTTPResponse`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore#HTTPResponse)

- As another "Common Usage Pattern" in [this doc](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-core-concepts). Arguably it should be the first one since error handling is potentially applicable to every service method call, whereas the other methods only apply in specific circumstances.

jhendrixMSFT:
Fixed in the pending v0.21.0 release.

PR #16659, opened 2021-12-15
Refactor error responses

jhendrixMSFT:
SDK APIs will now return an *azcore.ResponseError when a request is made
to the service and the service returns a non-success HTTP status code.
ResponseError provides the HTTP status code, error code when available,
and the raw *http.Response.
The Error() method provides pretty-printing of the error, see the tests
for example output.
Since user-defined errors are no longer returned, the LRO pollers no
longer need a callback to create them so this has been removed.
Payload() and NopClosingBytesReader have been moved to the internal,
shared package so ResponseError can use them.
Removed some unnecessary error unmarshallers in the register RP policy.

### DIFF
--- a/sdk/azcore/errors.go
+++ b/sdk/azcore/errors.go
@@ -8,12 +8,9 @@ package azcore
 
 import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
-	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
 // ResponseError is returned when a request is made to a service and
 // the service returns a non-success HTTP status code.
 // Use errors.As() to access this type in the error chain.
 type ResponseError = shared.ResponseError
-
-var _ errorinfo.NonRetriable = (*ResponseError)(nil)
--- a/sdk/azcore/errors.go
+++ b/sdk/azcore/errors.go
@@ -7,20 +7,13 @@
 package azcore
 
 import (
-	"net/http"
-
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
-// HTTPResponse provides access to an HTTP response when available.
-// Errors returned from failed API calls will implement this interface.
-// Use errors.As() to access this interface in the error chain.
-// If there was no HTTP response then this interface will be omitted
-// from any error in the chain.
-type HTTPResponse interface {
-	RawResponse() *http.Response
-}
+// ResponseError is returned when a request is made to a service and
+// the service returns a non-success HTTP status code.
+// Use errors.As() to access this type in the error chain.
+type ResponseError = shared.ResponseError
 
-var _ HTTPResponse = (*shared.ResponseError)(nil)
-var _ errorinfo.NonRetriable = (*shared.ResponseError)(nil)
+var _ errorinfo.NonRetriable = (*ResponseError)(nil)
--- a/sdk/azcore/internal/shared/response_error.go
+++ b/sdk/azcore/internal/shared/response_error.go
@@ -24,22 +24,16 @@ func NewResponseError(resp *http.Response) error {
 	// prefer the error code in the response header
 	if ec := resp.Header.Get("x-ms-error-code"); ec != "" {
 		respErr.ErrorCode = ec
-	}
-
-	// write the request method and URL with response status code
-	msg := &bytes.Buffer{}
-	fmt.Fprintf(msg, "%s %s://%s%s\n", resp.Request.Method, resp.Request.URL.Scheme, resp.Request.URL.Host, resp.Request.URL.Path)
-	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
-	fmt.Fprintf(msg, "RESPONSE %d: %s\n", resp.StatusCode, resp.Status)
-
-	body, err := Payload(resp)
-	if err != nil {
-		body = []byte(fmt.Sprintf("Error reading response body: %v", err))
-		goto Epilog
+		return respErr
 	}
 
 	// if we didn't get x-ms-error-code, check in the response body
-	if respErr.ErrorCode == "" && len(body) > 0 {
+	body, err := Payload(resp)
+	if err != nil {
+		return err
+	}
+
+	if len(body) > 0 {
 		if code := extractErrorCodeJSON(body); code != "" {
 			respErr.ErrorCode = code
 		} else if code := extractErrorCodeXML(body); code != "" {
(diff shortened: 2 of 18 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/16108

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
...
+		RawResponse: resp,
+	}
+
+	// prefer the error code in the response header
+	if ec := resp.Header.Get("x-ms-error-code"); ec != "" {
+		respErr.ErrorCode = ec
+	}
+
+	// write the request method and URL with response status code
+	msg := &bytes.Buffer{}
```
JeffreyRichter, line 30: I think it would be better to do all this work in the Error method because you ALWAYS do it when an error occurs here but only do it if Error() is called by the customer. This code creates more heap objects and takes CPU time that is really only needed if Error() is called by the customer's code.

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	// first submatch is the entire thing, second one is the captured error code
+	return res[1]
+}
+
+// ResponseError is returned when a request is made to a service and
+// the service returns a non-success HTTP status code.
+// Use errors.As() to access this type in the error chain.
+type ResponseError struct {
+	msg string
```
JeffreyRichter, line 118: I would get rid of this field and just have Error() return the string instead.

sdk/azcore/errors.go
```diff
@@ -7,20 +7,13 @@
...
-// Use errors.As() to access this interface in the error chain.
-// If there was no HTTP response then this interface will be omitted
-// from any error in the chain.
-type HTTPResponse interface {
-	RawResponse() *http.Response
-}
+// ResponseError is returned when a request is made to a service and
+// the service returns a non-success HTTP status code.
+// Use errors.As() to access this type in the error chain.
+type ResponseError = shared.ResponseError
```
jhendrixMSFT, line 17: Is the name ok?

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
+//go:build go1.16
```
jhendrixMSFT, line 1: This file is the core of the change.

sdk/azcore/internal/shared/shared.go
```diff
@@ -132,6 +103,81 @@ func HasStatusCode(resp *http.Response, statusCodes ...int) bool {
 	return false
 }
 
+// Payload reads and returns the response body or an error.
```
jhendrixMSFT, line 106: All the additions here are preexisting content that's been moved (no changes).

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
...
+// ResponseError is returned when a request is made to a service and
+// the service returns a non-success HTTP status code.
+// Use errors.As() to access this type in the error chain.
+type ResponseError struct {
+	msg string
+
+	// ErrorCode is the error code returned by the resource provider if available.
+	ErrorCode string
+
+	// StatusCode is the HTTP status code.
```
JeffreyRichter, line 123: Add to the doc comment a reference to the Status* constants here: https://pkg.go.dev/net/http#pkg-constants ??
I say this because customers should use these contants in their code.

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
...
+	StatusCode int
+
+	// RawResponse is the underlying HTTP response.
+	RawResponse *http.Response
+}
+
+// Error implements the error interface for type ResponseError.
+func (e *ResponseError) Error() string {
+	return e.msg
```
JeffreyRichter, line 132: Move code to calculate the string here

sdk/azcore/internal/shared/response_error.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(res) != 2 {
+		return ""
+	}
+	// first submatch is the entire thing, second one is the captured error code
+	return res[1]
+}
+
+// ResponseError is returned when a request is made to a service and
+// the service returns a non-success HTTP status code.
+// Use errors.As() to access this type in the error chain.
```
JeffreyRichter, line 116: Do customers use ResponseError struct or an interface with errors.As()?
I thought the struct was going to be private and an interface was public.

JeffreyRichter (approved):

sdk/azcore/internal/shared/response_error.go
jhendrixMSFT: While I agree, the case that makes this tricky is when the error code is in the body, not the `x-ms-error-code` header.  We don't want `Error()` to populate the `ErrorCode` post-construction.  So if we have to do this work in the constructor, we end up doing almost the same work in `Error()` as well (more garbage etc).

This keeps things simple, and I think it's highly likely that `Error()` will be invoked.  If this really becomes a problem, we can change it later.

sdk/azcore/internal/shared/response_error.go
jhendrixMSFT: If we made this an interface, it wouldn't be versioning resilient.  While we could somewhat work around this by making it "not publicly implementable" by way of an unexported method, I don't think that really buys us anything and just adds an extra layer of indirection.

My mental model for error types is use concrete types for returning data (status code, error code, etc), and use interfaces for behavior (e.g. our NonRetriable() marker method for idempotent errors).  This allows you to combine them, e.g. `ResponseError` is also `NonRetriable()` and avoids the need for any error chaining.

sdk/azcore/internal/shared/response_error.go
jhendrixMSFT: Fixed

sdk/azcore/internal/shared/response_error.go
jhendrixMSFT: Fixed

sdk/azcore/internal/shared/response_error.go
jhendrixMSFT: Fixed

sdk/azcore/internal/shared/response_error_test.go
```diff
@@ -0,0 +1,385 @@
...
+		Body:       &readFailer{},
+		Request: &http.Request{
+			Method: http.MethodGet,
+			URL:    fakeURL,
+		},
+	})
+	_, ok := err.(*ResponseError)
+	if ok {
+		t.Fatalf("unexpected error type %T", err)
+	}
```
seankane-msft, line 273: Are you expecting this `err` not to be converted to a `ResponseError`? Should line 271 read `if !ok {` ?

seankane-msft (approved):
Noticed one (potential) issue

sdk/azcore/internal/shared/response_error_test.go
jhendrixMSFT: It used to, but after discussing with Jeff yesterday it's changed.  In the case where reading from the HTTP response body fails, we return that error instead.

jhendrixMSFT: The reason being, the read really shouldn't fail.  If it does, it's a bug so we don't want to mask it behind a `ResponseError`.

PR #16662, opened 2021-12-15
[Performance] Add performance testing framework

seankane-msft:
This PR introduces a performance testing framework for Go. It includes sample performance tests for `azblob`, `aztables` and `azkeys`.


eng/tools/azgoperf/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
jhendrixMSFT, line 2: Why isn't this MIT?

seankane-msft: This was the default by cobra's cli. Working on cleaning a few things up still, this is on that list.

RickWinter (commented):
Should we put all the go tools into a `go` folder instead of putting go in the name of all the tools.
Example:
`eng/tools/go/azperf/`

seankane-msft:
> Should we put all the go tools into a `go` folder instead of putting go in the name of all the tools? Example: `eng/tools/go/azperf/`

This is the only tool with `go` in the name, I think it's simpler to rename to `azperf`

vhvb1989 (approved):
nit. Can you add a gate to CI to build the perf fw and current tests?

What would help to avoid regressions .

It can be in another PR/issue

eng/tools/azperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+## Default Command Options
+
+| Flag | Short Flag | Default Value | Variable Name | Description |
+| -----| ---------- | ------------- | ------------- | ----------- |
+| `--duration` | `-d` | 10 seconds | internal.Duration (`int`) | How long to run an individual performance test |
+| `--iterations` | `-i` | 3 | internal.Iterations (`int`) | How many iterations of a performance test to run |
+| `--testproxy` | `-x` | N/A | internal.TestProxy (`string`) | Whether to run a test against a test proxy. If you want to run against `https` specify with `--testproxy https`, likewise for `http`. If you want to run normally omit this flag |
+| `--warmup` | `-w` | 3 seconds| internal.WarmUp (`int`) | How long to allow the connection to warm up. |
+| `--timeout` | `-t` | 10 seconds| internal.TimeoutSeconds (`int`) | How long to allow an operation to block. |
+
```
mikeharder, line 13: I'm not very familiar with how they work in go, but "parallelism" and "sync vs async" are important concepts in the other languages perf frameworks.

For "sync vs async", other languages like .NET, Java, and Python expose both sync and async APIs in our SDK clients, and `Run()` vs `RunAsync()` methods on the `PerfTest` base class.  This allows us to test both flavor of APIs, compare their performance, etc.  JS is async-only.

The "parallel" parameter controls how many instances of the test class will be executed in parallel.  For example, if "parallel=4", the perf framework will:

1. Create 4 instances of the test class
2. Call `GlobalSetup()` on the first instance
3. Call `Setup()` on each instance (in parallel)
4. Call `Run/RunAsync()` on each instance in a loop (in parallel)
5. Call `Cleanup()` on each instance (in parallel)
6. Call `GlobalCleanup()` on the first instance

mikeharder (changes requested):
Added high-level comments.  Will add more detailed feedback after these are resolved.

mikeharder:
Please port the following sample tests to this PR.  These samples are used to test the basic functionality and performance of the perf framework:

https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-devtools/src/azure_devtools/perfstress_tests/system_perfstress/no_op_test.py

https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-devtools/src/azure_devtools/perfstress_tests/system_perfstress/sleep_test.py

eng/tools/azperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+First, make sure you have go version 1.17 or later installed. You can install go from the [go.dev](https://go.dev/doc/install) site.
+
+Navigate to the `eng/tools/azperf` folder and run `go mod download` to download all the requirements for the performance framework.
+
+Build the CLI with the command:
+```pwsh
+go build .
+```
+
+Now you will have a single executable `azperf.exe` on Windows, or `azperf` on Mac/Linux, which can be used to run all performance tests.
```
mikeharder, line 98: I'm worried about the scalability of having all the perf tests in a single executable.  Most languages create one perf executable per SDK, so there would be one executable for Storage Blob perf tests, one for KeyVault perf tests, etc.  Would this make sense for go?

eng/tools/azperf/README.md
seankane-msft: Go is a bit different than the other languages. There are no explicit async APIs in our clients. Async programming is done via [goroutines](https://gobyexample.com/goroutines), but this is done entirely on the client side. I can add parallel testing for this by creating the `goroutines` manually but this will just test the sync APIs. I think for the first iteration of the performance framework, doing this without `goroutines` is ideal and I can add them later if they are necessary.

eng/tools/azperf/README.md
```diff
@@ -0,0 +1,120 @@
...
+        },
+    }
+
+    client, err := azkeys.NewClient("<my-vault-url>", cred, options)
+    if err != nil {return err}
+    k.client = client
+}
+```
+
+Before you can use the proxy in playback mode, you have to generate live recordings. To do this, call `recording.Start` before the test run and `recording.Stop` after the test run.
```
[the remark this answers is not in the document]
seankane-msft: Added

eng/tools/azperf/README.md
seankane-msft: I'm not sure but I will implement and see what it looks like. I'll start out with just one example from storage.

seankane-msft:
> nit. Can you add a gate to CI to build the perf fw and current tests?
> 
> What would help to avoid regressions .
> 
> It can be in another PR/issue

I moved this into our `internal` module so the CI is now there. I don't have anything added but it's there

eng/tools/azperf/README.md
seankane-msft: I've migrated over to an executable per SDK. I think it's much cleaner this way. Thanks for pointing this out Mike.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,270 @@
...
+	// Get the results from the channels
+	for idx, channel := range channels {
+		result := <-channel
+		if err != nil {
+			return err
+		}
+		printIteration(idx, result.timeInSeconds, result.count)
+
+		totalCount += result.count
+		totalTime += result.timeInSeconds
```
mikeharder, line 190: Two problems with this calculation:

1. You are adding the times across all parallel instances, when instead you want to be (approximately) averaging them.  If 4 parallel threads run for 10 seconds each, the total time should be 10 seconds, not 40.

2. Dividing the total number of operations by the average time across all threads also incorrect, for a more subtle mathematical reason.  Instead, you need to sum the operations per second for each thread.

Say thread1 completes 10 operations in 10 seconds, and thread2 completes 10 operations in 20 seconds (the last operation stalled for 10 seconds).  What should be the average throughput?

The correct answer is thread1 completed 1 operation per second, and thread2 completed 0.5 operations per second (10 / 20), for a sum of 1.5 operations per second across all threads.

The incorrect answer is to divide the total operations (20) by the average time (15) for a result of 1.333 operations per second.

This is an extreme example, and in practice the threads should complete at about the same time, but for correctness the total throughput needs to be calculated this way.

You should be able to copy the logic from Python:

https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-devtools/src/azure_devtools/perfstress_tests/_perf_stress_runner.py#L204-L213

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,270 @@
...
+	var channels []chan runResult
+
+	for idx := 0; idx < Parallel; idx++ {
+		wg.Add(1)
+		c := make(chan runResult, 1) // Create a buffered channel
+		channels = append(channels, c)
+		go runTest(p, c)
+	}
+
+	wg.Wait()
```
mikeharder, line 179: An important feature in the other languages is reporting intermediate results every second while the test is running.  Example output from .NET:

```
=== Test ===
    Current             Total          Average       CPU    WorkingSet    PrivateMemory
 25,195,915        25,195,915    25,119,342.75    25.15%        68.74M          161.21M
 24,333,189        49,529,104    24,603,047.21    25.82%        68.74M          161.33M
 23,544,658        73,073,762    24,433,749.18    26.07%        68.82M          161.47M

=== Results ===
Completed 73,073,762 operations in a weighted-average of 2.99s (24,433,749.18 ops/s, 0.000 s/op, 25.67% CPU)
```

Benefits:

1. Shows consistency of throughput over time, which can help identify insufficient warmup time, resource leaks, service throttling, etc.
2. Shows results in real-time while the test is running, without needing to wait the full duration.

sdk/internal/perf/testdata/perf/no_op_perf.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import "context"
+
+type noOpPerfTest struct{}
```
mikeharder, line 5: I compared the no-op test results across go and .NET.  Go is about 15% slower, but it should be plenty fast for our scenarios:

image

sdk/storage/azblob/testdata/perf/download_blob.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	m.containerClient = containerClient
+	_, err = m.containerClient.Create(context.Background(), nil)
+	if err != nil {
+		return err
+	}
+
+	m.blobClient = m.containerClient.NewBlockBlobClient(m.blobName)
+
+	m.data = "This is all placeholder random data for now. This is all placeholder random data for now. This is all placeholder random data for now."
```
mikeharder, line 47: To get a fair comparison between go and other languages, the upload/download tests need to add a `size` parameter for the size of the blob in bytes, and the listblobs test needs to add a `count` parameter for the number of blobs to list.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/spf13/pflag"
+)
+
+var (
+	debug          bool
+	Duration       int
+	TimeoutSeconds int
+	TestProxy      string
+	WarmUp         int
```
mikeharder, line 21: I don't think `WarmUp` is used yet.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,270 @@
...
+
+	wg.Wait()
+
+	// Get the results from the channels
+	for idx, channel := range channels {
+		result := <-channel
+		if err != nil {
+			return err
+		}
+		printIteration(idx, result.timeInSeconds, result.count)
```
mikeharder, line 187: Other frameworks only print the total across all threads, not results per thread.

sdk/storage/azblob/testdata/perf/download_blob.go
```diff
@@ -0,0 +1,74 @@
...
+	_, err := m.containerClient.Delete(context.Background(), nil)
+	return err
+}
+
+func (m *downloadPerfTest) Setup(ctx context.Context) error {
+	return nil
+}
+
+func (m *downloadPerfTest) Run(ctx context.Context) error {
+	_, err := m.blobClient.Download(ctx, nil)
```
mikeharder, line 64: This test never completes for me and hangs the perf test framework.

mikeharder (changes requested):

sdk/internal/perf/perf.go
seankane-msft: Thanks for the explanation, I've copied the Python impl.

sdk/internal/perf/perf.go
seankane-msft: Removed

sdk/internal/perf/perf.go
seankane-msft: This is simple when only running a single thread, when it's run on multiple threads, is the `Total` column the total run over all threads and the `Current` is the value for a single thread?

I am working on CPU, WorkingSet, and PrivateMemory but I might add it in a later PR.

sdk/internal/perf/testdata/perf/no_op_perf.go
seankane-msft: I have updated numbers for this, they're quite a bit higher now so I am verifying they are correct.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,270 @@
...
+func runPerfTest(p PerfTest) error {
+	err := runGlobalSetup(p)
+	if err != nil {
+		return err
+	}
+
+	totalCount := 0
+	totalTime := 0.0
+
+	runSetup(p)
```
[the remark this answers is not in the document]
seankane-msft: Fixed

sdk/internal/perf/perf.go
mikeharder: All values are aggregated across all threads.  The threads need to either update shared variables (in a thread-safe way), or the main thread needs to query the threads and aggregate the values.

`Current` is the number of operations completed in the last second, `Total` is the cumulative operations completed since the run was started, and `Average` is `Total / Elapsed Time`.

sdk/internal/perf/testdata/perf/no_op_perf.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/perf"
+)
+
+type noOpPerfTest struct {
+	perf.PerfTestOptions
+}
+
+func (n *noOpPerfTest) GlobalSetup(ctx context.Context) error {
```
mikeharder, line 13: It would be nice if tests could omit all these methods if they don't need to add any functionality.  But I know go doesn't have inheritance like the other languages, so if you think this is the best way to implement this in go, it's fine with me.

sdk/internal/perf/testdata/perf/sleep_perf.go
```diff
@@ -0,0 +1,56 @@
...
+func (s *sleepPerfTest) GlobalSetup(ctx context.Context) error {
+	return nil
+}
+
+func (s *sleepPerfTest) GlobalCleanup(ctx context.Context) error {
+	return nil
+}
+
+func (s *sleepPerfTest) Setup(ctx context.Context) error {
+	s.secondsPerOp = int(math.Pow(2.0, float64(s.ParallelIndex)))
```
mikeharder, line 25: Could this be moved to the ctor, or must it live in `Setup()`?  Is `ParallelIndex` available in the ctor?

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// All ensuing requests go through in Playback mode
+		setRecordingMode("playback")
+		err = start(p.GetMetadata().Name, nil)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	if WarmUp > 0 {
```
mikeharder, line 108: The per-second results should be printed during both warmup and duration, in separate sections of the output.  Example from .NET:

```
=== Warmup ===
    Current             Total          Average       CPU    WorkingSet    PrivateMemory
 19,516,504        19,516,504    19,451,361.00     8.39%        35.18M           60.59M
 22,341,875        41,858,379    20,682,285.66     8.70%        35.76M           60.71M
 21,822,398        63,680,777    21,272,532.56     8.32%        35.80M           60.71M

=== Results ===
Completed 63,680,777 operations in a weighted-average of 2.99s (21,272,532.56 ops/s, 0.000 s/op, 8.46% CPU)

=== Test ===
    Current             Total          Average       CPU    WorkingSet    PrivateMemory
 22,128,980        22,128,980    21,837,353.99     8.35%        35.88M           60.68M
 23,162,566        45,291,546    22,275,325.98     8.30%        35.90M           60.71M
 21,860,969        67,152,515    22,286,358.33     8.37%        35.97M           60.72M

=== Results ===
Completed 67,152,515 operations in a weighted-average of 3.01s (22,286,358.33 ops/s, 0.000 s/op, 8.33% CPU)
```

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+	perSecondCount := make([]int, 0)
+	w := tabwriter.NewWriter(os.Stdout, 16, 8, 1, ' ', tabwriter.AlignRight|tabwriter.Debug)
+	for time.Since(timeStart).Seconds() < float64(Duration) {
+		err := p.Run(context.Background())
+		if err != nil {
+			c <- runResult{err: err}
+		}
+		totalCount += 1
+
+		// Every second (roughly) we print out an update
```
mikeharder, line 130: The per-second updates need to be aggregated across all the parallel instances, rather than printing one line per parallel instance.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+		if time.Since(timeStart).Seconds() > float64(lastPrint) {
+			thisCount := totalCount - sumInts(perSecondCount)
+			perSecondCount = append(perSecondCount, thisCount)
+			_, err = fmt.Fprintf(
+				w,
+				"%s\t%s\t%.2f\t\n",
+				commaIze(thisCount),
+				commaIze(totalCount),
+				float64(sumInts(perSecondCount))/time.Since(timeStart).Seconds(),
```
mikeharder, line 139: The "average" column should also be formatted with commas

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+			return err
+		}
+
+		// Create a thread for running a single test
+		wg.Add(1)
+		c := make(chan runResult, 1) // Create a buffered channel
+		channels = append(channels, c)
+		go runTest(perfTest, c)
+	}
+
```
mikeharder, line 229: I'm seeing an issue at the end of the duration, where Linux hangs and Windows throws with this error:

```
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [semacquire]:
sync.runtime_Semacquire(0x0)
        C:/Program Files/Go/src/runtime/sema.go:56 +0x25
sync.(*WaitGroup).Wait(0x6a3a00)
        C:/Program Files/Go/src/sync/waitgroup.go:130 +0x71
github.com/Azure/azure-sdk-for-go/sdk/internal/perf.runPerfTest(0x706908)
        D:/Git/go/sdk/internal/perf/perf.go:230 +0x557
github.com/Azure/azure-sdk-for-go/sdk/internal/perf.Run({0xc000113f60, 0x0, 0x2})
        D:/Git/go/sdk/internal/perf/perf.go:325 +0x358
main.main()
        D:/Git/go/sdk/internal/perf/testdata/perf/main.go:6 +0x45

goroutine 8 [chan send]:
github.com/Azure/azure-sdk-for-go/sdk/internal/perf.runTest({0x74ffd8, 0xc00015a6c0}, 0x0)
        D:/Git/go/sdk/internal/perf/perf.go:157 +0xbc5
created by github.com/Azure/azure-sdk-for-go/sdk/internal/perf.runPerfTest
        D:/Git/go/sdk/internal/perf/perf.go:227 +0x16c

goroutine 9 [chan send]:
github.com/Azure/azure-sdk-for-go/sdk/internal/perf.runTest({0x74ffd8, 0xc00015a720}, 0x0)
        D:/Git/go/sdk/internal/perf/perf.go:157 +0xbc5
created by github.com/Azure/azure-sdk-for-go/sdk/internal/perf.runPerfTest
        D:/Git/go/sdk/internal/perf/perf.go:227 +0x16c
exit status 2
```

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+	for _, pTest := range perfTests {
+		err = runCleanup(pTest)
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	err = runGlobalCleanup(p(nil))
+
+	fmt.Println("\n=== Results ===")
```
mikeharder, line 255: It's nice to print results before cleanup, so the user doesn't need to wait if cleanup is slow.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+func RegisterArguments(f func()) {
+	f()
+	registerCalled = true
+}
+
+// Run runs all individual tests
+func Run(perfTests []NewPerfTest) {
+	// Start with adding all of our arguments
+	pflag.IntVarP(&Duration, "duration", "d", 10, "Duration of the test in seconds. Default is 10.")
+	pflag.StringVarP(&TestProxy, "test-proxies", "x", "", "whether to target http or https proxy (default is neither)")
```
mikeharder, line 296: The `test-proxies` parameter needs to accept a list of URIs separated by `;`.  The proxy URLs should be assigned round-robin to the parallel test instances.  So for `--parallel 5 --test-proxies http://proxy1;http://proxy2`, test instances 1, 3, and 5 should be assigned proxy1, and test instances 2 and 4 should be assigned proxy2.  This feature is used for scenarios where multiple test proxies are required to bottleneck the client.

https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-devtools/src/azure_devtools/perfstress_tests/_perf_stress_runner.py#L106

sdk/internal/perf/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	} else if i >= 10 {
+		return fmt.Sprintf("0%d", i)
+	} else if i > 0 {
+		return fmt.Sprintf("00%d", i)
+	}
+	return "000"
+}
+
+// Convert an integer to a string with comma separation in the format 1234 -> 1,234
+func commaIze(i int) string {
```
mikeharder, line 30: You might be able to use the `message` package instead of writing this yourself:

https://pkg.go.dev/golang.org/x/text/message?utm_source=godoc

sdk/internal/perf/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if copy == 0 {
+		return ret
+	}
+
+	return fmt.Sprintf("%d,%s", copy, ret)
+}
+
+func min(a, b int) int {
```
mikeharder, line 52: Could you use this instead?

https://pkg.go.dev/math#Min

mikeharder (dismissed):

sdk/internal/perf/utils.go
seankane-msft: Unfortunately that Min is for floats, after generics hopefully that will be fixed to be a generic type.

sdk/internal/perf/testdata/perf/sleep_perf.go
seankane-msft: Yes it is, moving now.

sdk/internal/perf/utils.go
seankane-msft: Using this package, but this also forces users of client libraries to download it because every package uses `internal`. Something I don't know how we tackle yet.

sdk/internal/perf/perf.go
```diff
@@ -0,0 +1,329 @@
...
+	pflag.IntVarP(&WarmUp, "warmup", "w", 5, "Duration of warmup in seconds. Default is 5.")
+	pflag.IntVarP(&Parallel, "parallel", "p", 1, "Degree of parallelism to run with. Default is 1.")
```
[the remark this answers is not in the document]
seankane-msft: Added.

sdk/internal/perf/testdata/perf/no_op_perf.go
seankane-msft: Unfortunately I'm not sure of a way to do that. Adding a nil method is pretty simple here. When I do a presentation with joel/jeff I'll ask for input.

sdk/internal/perf/perf.go
seankane-msft: Fixed

sdk/internal/perf/perf.go
seankane-msft: Added

sdk/internal/perf/perf.go
seankane-msft: Resolved

sdk/storage/azblob/testdata/perf/download_blob.go
seankane-msft: Added

mikeharder (changes requested):
Committing pending comments, more to come

PR #16721, opened 2021-12-29
Updated azblob README sample code

kimprice:
The container name should be all lowercase. Currently, running the sample code will cause an error that there are invalid characters in the container name.


mikekistler (approved):
Looks good! 👍

mohsha-msft (approved):

PR #16770, opened 2022-01-06
Rename armmanagedapplications ci.yml files to avoid pipeline name definition collisions

benbp:
The pipeline generation tool will use the parent directory name for generating pipeline names. This causes issues in the
go repo as there are more nested directories that have duplicate names. The fix for this is to add a "variant" to the
ci.yml file naming, which will cause that variant to be appended to the pipeline name (e.g. `go -
armmanagedapplications` becomes `go - armmanagedapplications - solutions`).


weshaggard (approved):

benbp:
FYI @tadelesh

tadelesh (approved):

Issue #16640, opened 2021-12-14
Add offline managed identity tests


chlowell:
i.e. validate the client creates the correct request for each platform:
- [x] App Service (2017-09-01)
- [x] Arc #16771
- [x] Cloud Shell #16771
- [x] IMDS
- [x] Service Fabric #16267

PR #16771, opened 2022-01-06
Add offline test for Azure Arc managed identity

chlowell:
Part of #16640

### DIFF
--- a/sdk/azidentity/managed_identity_credential_test.go
+++ b/sdk/azidentity/managed_identity_credential_test.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -57,10 +56,6 @@ func (m *mockIMDS) Do(req *http.Request) (*http.Response, error) {
 }
 
 func TestManagedIdentityCredential_AzureArc(t *testing.T) {
-	if runtime.GOOS == "windows" {
-		// https://go-review.googlesource.com/c/go/+/371296
-		t.Skip("working around T.Tempdir bug fixed in 1.18")
-	}
 	file, err := os.Create(filepath.Join(t.TempDir(), "arc.key"))
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_credential_test.go
```diff
@@ -54,80 +55,102 @@ func (m *mockIMDS) Do(req *http.Request) (*http.Response, error) {
 	panic("no more responses")
 }
 
-func TestManagedIdentityCredential_GetTokenInAzureArcLive(t *testing.T) {
-	if len(os.Getenv(arcIMDSEndpoint)) == 0 {
-		t.Skip()
-	}
-	msiCred, err := NewManagedIdentityCredential(nil)
+func TestManagedIdentityCredential_AzureArc(t *testing.T) {
+	file, err := os.Create(filepath.Join(t.TempDir(), "arc.key"))
```
jhendrixMSFT, line 59: I think you're missing a `defer file.Close()`

sdk/azidentity/managed_identity_credential_test.go
```diff
@@ -54,80 +55,102 @@ func (m *mockIMDS) Do(req *http.Request) (*http.Response, error) {
...
 	}
-	_, err = msiCred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{msiScope}})
+	tk, err := msiCred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{msiScope}})
 	if err != nil {
-		t.Fatalf("Received an error when attempting to retrieve a token")
+		t.Fatal(err)
 	}
-}
-
-func TestManagedIdentityCredential_GetTokenInCloudShellMockFail(t *testing.T) {
```
jhendrixMSFT, line 104: This test case isn't needed?

jhendrixMSFT (approved):
Few minor questions but looks good.

sdk/azidentity/managed_identity_credential_test.go
chlowell: Thanks, I'll add that. Not sure it will solve the failure on Windows. Apparently `testing` cleanup of temp dirs is flaky: https://go-review.googlesource.com/c/go/+/371296. That's fixed for 1.18 so I'll just skip this test on Windows for now.

jhendrixMSFT: Might be worth it to see if closing the file fixes this first.  If not then yeah just skip on Windows.

PR #16777, opened 2022-01-07
Align authentication errors with azcore.ResponseError

### LINKED ISSUE #16736: Make AuthenticationFailedError consistent with other client errors
Credentials return `AuthenticationFailedError` when authentication fails. That type has a `RawResponse()` method which returns any response the credential received. Naively, it seems `AuthenticationFailedError` should instead wrap [`azcore.ResponseError`](https://github.com/Azure/azure-sdk-for-go/blob/6264636169251df0ea48912167b417c2a9c108ce/sdk/azcore/internal/shared/response_error.go#L90), to enable consistent error handling. Otherwise the API compels applications to handle authentication failures differently than other errors:
```go
err := client.Fail()
var res *http.Response
var authErr azidentity.AuthenticationFailedError
if errors.As(err, &authErr) {
	res = authErr.RawResponse()
}
var respErr azcore.ResponseError
if errors.As(err, &respErr) {
	res = respErr.RawResponse
}
```

chlowell:
Closes #16736. `AuthenticationFailedError.Error()` now returns a string like this when a response is available:
```
POST https://login.microsoftonline.com/3c631bb7-a9f7-4343-a5ba-a6159135f1fc/oauth2/v2.0/token
RESPONSE 401 Unauthorized
{
  "error": "invalid_client",
  "error_description": "AADSTS7000215: Invalid client secret is provided.\r\nTrace ID: ...",
  "error_codes": [
    7000215
  ],
  "timestamp": "2022-01-07 19:44:10Z",
  "trace_id": "b968067a-6427-402d-9fed-4df6c8f67a02",
  "correlation_id": "375a304a-f733-40aa-9668-a87cf0e3adcb",
  "error_uri": "https://login.microsoftonline.com/error?code=7000215"
}
```


sdk/azidentity/errors.go
```diff
@@ -4,82 +4,88 @@
...
-type CredentialUnavailableError interface {
-	errorinfo.NonRetriable
-	credentialUnavailable()
-}
+var _ errorinfo.NonRetriable = (*AuthenticationFailedError)(nil)
 
-type credentialUnavailableError struct {
+// CredentialUnavailableError indicates a credential can't attempt
+// authentication because it lacks required data or state.
+type CredentialUnavailableError struct {
```
jhendrixMSFT, line 73: Does this actually need to be exported?  I thought that for this case the error is purely informative, and callers don't need to programmatically interact with it.

chlowell: I left it as such to enable applications to distinguish "I can't try to authenticate" from "authentication failed" but I imagine that distinction isn't important to most applications. I'll unexport it now, and export it later if someone has a need.

jhendrixMSFT (approved):

PR #16785, opened 2022-01-10
[KeyVault] Add support for OKP key type

### LINKED ISSUE #16775: [KeyVault] - Add support for OKP key type
...and the EdDSA signature algorithm, when Key Vault supports them.

seankane-msft:
Closes #16775 
- [ ] Adds `CreateOKPKey` function
- [ ] Adds relevant types


sdk/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
 # Release History
 
-## 0.1.1 (Unreleased)
+## 0.2.0 (Unreleased)
 
 ### Features Added
+* Adds the `CreateOKPKey` method for creating OKP public keys
+* Adds the `Ed25519` `JSONWebKeyCurveName`
+*Adds the `OKP` and `OKPHSM` `JSONWebKeyType`s
```
heaths, line 8: 1. Did the "JSONWeb" prefix already ship? None of the languages use that, just `KeyCurveName` and `KeyType`.
2. Missing a space required for a bulleted list on line 8.

sdk/keyvault/azkeys/client.go
```diff
@@ -335,6 +335,67 @@ func (c *Client) CreateRSAKey(ctx context.Context, name string, options *CreateR
...
+	HardwareProtected bool
+
+	// The key size in bits. For example: 2048, 3072, or 4096 for RSA.
+	KeySize *int32 `json:"key_size,omitempty"`
+
+	// The public exponent for a RSA key.
+	PublicExponent *int32 `json:"public_exponent,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]*string `json:"tags,omitempty"`
```
heaths, line 349: Why a reference for the map value? Just so values can be `nil`? I don't think this is actually possible service-side.

heaths (approved):

sdk/keyvault/azkeys/CHANGELOG.md
seankane-msft: Yes a first beta was released. The name can be changed to align with the other Tier 1 languages.

heaths: Aligning with other languages would be ideal if no good idiomatic reason exists to differ.

sdk/keyvault/azkeys/client.go
seankane-msft: Had a discussion with @jhendrixMSFT about this, we're changing all `Tags map[string]*string` to `Tags map[string]string`, the nil value is not valid.

sdk/keyvault/azkeys/models.go
```diff
@@ -397,3 +398,22 @@ type LifetimeActionsTrigger struct {
...
+}
+
+func convertToGeneratedMap(m map[string]string) map[string]*string {
+	ret := map[string]*string{}
+	for k, v := range m {
+		ret[k] = to.StringPtr(v)
+	}
+
+	return ret
+}
```
heaths, line 432: Is this maybe something the code generator should do, or even just not use a `map[string]*string`? I'm a bit confused why we'd even define generate `Tags` as a map of string pointer values given I doubt any of them accept missing values. Testing against a resource group and a Key Vault, tags could have an empty value, but even via JSON that comes back as an empty string. Shouldn't that empty string be reflected in Go as well, or have you universally chosen to represent it as `nil`?

```json
  "tags": {
    "test": ""
  },
```

I would say: do whatever is idiomatic for Go, but in my experience, it's been more common to see empty strings than `nil` for strings.

/cc @JeffreyRichter

jhendrixMSFT: There is a general problem that some maps (notably in ARM) use a `nil` value as the means to indicate "delete this key/value pair".  Unfortunately, not all specs are properly annotated with `x-nullable` to indicate this.  There's also the consideration of versioning tolerance if we were to support `x-nullable` in the future (i.e. should flipping this be a breaking change).

At any rate, we're looking at ways to improve this scenario as the common case has to pay the pointer-to-type tax which is unfortunate.

sdk/keyvault/azkeys/client.go
seankane-msft: This was taken care of in #16858

heaths:
@seankane-msft - I know this is draft, but also keep in mind this should not target 7.3-preview (or 7.3) now. We'll postpone it to 7.4-preview.

sdk/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,10 +1,16 @@
...
-## 0.2.1 (Unreleased)
+## 0.3.0 (Unreleased)
 
 ### Features Added
+* Adds the `CreateOKPKey` method for creating OKP public keys
+* Adds the `Ed25519` `KeyCurveName`
+* Adds the `OKP` and `OKPHSM` `KeyType`s
 
 ### Breaking Changes
+* `NewClient` returns an instance of a `Client`, instead of a `*Client`
```
chlowell, line 11: What's motivating this? I thought we'd aligned on returning pointers, which is still [the guideline](https://azure.github.io/azure-sdk/golang_introduction.html#service-client-constructors).

jhendrixMSFT: Correct, we should be returning clients by reference.

sdk/keyvault/azkeys/client.go
```diff
@@ -287,6 +282,16 @@ type CreateRSAKeyOptions struct {
 
 	// Application specific metadata in the form of key-value pairs.
 	Tags map[string]string `json:"tags,omitempty"`
+
+	// Elliptic curve name. For valid values, see KeyCurveName.
+	Curve *KeyCurveName `json:"crv,omitempty"`
```
heaths, line 292: `CurveName`

sdk/keyvault/azkeys/client.go
```diff
@@ -287,6 +282,16 @@ type CreateRSAKeyOptions struct {
...
+
+	// The attributes of a key managed by the key vault service.
+	KeyAttributes *KeyAttributes  `json:"attributes,omitempty"`
```
heaths, line 295: We flatten this in .NET, but if that's not an ideal option for Go this should be `Properties *KeyProperties`, but perhaps that's better covered in your other PR you'd then need to merge into this topic branch eventually.

PR #16787, opened 2022-01-10
Update azcosmos with the latest azcore

jhendrixMSFT:

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.2.0 (2022-1-13)
+## 0.2.0 (2022-01-13)
 
 ### Features Added
 * Failed API calls will now return an `*azcore.ResponseError` type.
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Release History
 
-## 0.1.1 (Unreleased)
+## 0.2.0 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
+* Updated to latest `azcore`. Public surface area is unchanged.  However, the `azcore.HTTPResponse` interface has been removed.
 
 ### Bugs Fixed
 
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -1,15 +1,12 @@
 # Release History
 
-## 0.2.0 (Unreleased)
+## 0.2.0 (2022-1-13)
 
 ### Features Added
+* Failed API calls will now return an `*azcore.ResponseError` type.
 
 ### Breaking Changes
 * Updated to latest `azcore`. Public surface area is unchanged.  However, the `azcore.HTTPResponse` interface has been removed.
-
-### Bugs Fixed
-
-### Other Changes
 
 ## 0.1.0 (2021-11-09)
 * This is the initial preview release of the `azcosmos` library
--- a/sdk/data/azcosmos/cosmos_error.go
+++ b/sdk/data/azcosmos/cosmos_error.go
@@ -5,25 +5,11 @@ package azcosmos
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 )
-
-// CosmosError is used as base error for any error response from the Cosmos service.
-type CosmosError struct {
-	body string
-
-	// ErrorCode is the error code returned by the service if available.
-	ErrorCode string
-
-	// StatusCode is the HTTP status code.
-	StatusCode int
-
-	// RawResponse is the underlying HTTP response.
-	RawResponse *http.Response
-}
 
 func newCosmosError(response *http.Response) error {
 	bytesRead, err := azruntime.Payload(response)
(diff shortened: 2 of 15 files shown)


sdk/data/azcosmos/example_test.go
```diff
@@ -501,7 +501,7 @@ func ExampleContainerClient_ReplaceItem_optimisticConcurrency() {
 		// Replace with Etag
 		etag := itemResponse.ETag
 		itemResponse, err = container.ReplaceItem(context.Background(), pk, id, marshalledReplace, &azcosmos.ItemOptions{IfMatchEtag: &etag})
-		var httpErr azcore.HTTPResponse
+		var httpErr interface{ RawResponse() *http.Response }
```
jhendrixMSFT, line 504: Given that your error type already exposes the raw *http.Response via a field, I think this can just go away.  Previously, we exposed this interface on error types so that callers had a way to get the underlying response if they cared about it.

ealsur: Is our error type still correct in terms of the fields it exposes related to the recommendations? Or should we do it differently?

ealsur (approved):
LGTM, had only a single comment related to the error response.

sdk/data/azcosmos/example_test.go
jhendrixMSFT: Our contract for error types is that they must expose the following bits of data.
- the error code, usually from `x-ms-error-code` header or maybe some field in the response body
- the HTTP status code
- the raw response

Anything else should be contextual to help diagnose the error and come from the `Error()` method.  For our code-generated SDKs, we have [`azcore.ResponseError`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/shared/response_error.go#L90-L99) for this.

Is `cosmosError` the common error type that's returned for failed API calls?

ealsur: Yes, for failed API calls we return `cosmosError` which right now exposes (https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/cosmos_error.go):

* `func (e *cosmosError) StatusCode() int`
* `func (e *cosmosError) ErrorCode() string`
* `func (e *cosmosError) RawResponse() *http.Response`

jhendrixMSFT: I see.  What's the public type that customers use to access this info?

jhendrixMSFT: Take a look at my proposed changes to `CosmosError`.

ealsur: RIght now there was no public type, as far as I remember, our earlier discussion was that as long as our internal type worked with the proposed azcore interface, it would be good enough. There is no more azcore error interface now?

ealsur: I don't mind making CosmosError public as long as it's fine from the API design perspective and follows the azcore guidelines 😄

jhendrixMSFT: @JeffreyRichter this should align with our current thinking for error handling.  Any concerns?

sdk/data/azcosmos/cosmos_error.go
```diff
@@ -11,26 +11,18 @@ import (
...
 )
 
-// cosmosError is used as base error for any error response from the Cosmos service.
-type cosmosError struct {
-	body        string
-	code        string
-	rawResponse *http.Response
-}
+// CosmosError is used as base error for any error response from the Cosmos service.
+type CosmosError struct {
```
JeffreyRichter, line 15: Should we get rid of this and just use azcore.ResponeError?

sdk/data/azcosmos/cosmos_error.go
```diff
@@ -39,16 +31,17 @@ func newCosmosError(response *http.Response) error {
 		return err
 	}
 
-	cError := cosmosError{
-		rawResponse: response,
+	cError := CosmosError{
```
JeffreyRichter, line 34: This could create an azcore.ResponseError instead.

JeffreyRichter (approved):

sdk/data/azcosmos/example_test.go
JeffreyRichter: I think CosmosError should be removed entirely and the azcore.ResponseError shoud be used instead.

jhendrixMSFT: We could do that.  However, I want to point out that having `CosmosError` would allow for any Cosmos-specific error info to be exposed in the future.  How likely is this to happen, and more generally, do we want to support this scenario?

jhendrixMSFT: Another option would be to have `CosmosError` wrap an `azcore.ResponseError` instead of reimplementing it.  That way we have consistent handling while keeping the extensibility for Cosmos-specific content.

jhendrixMSFT: I'll also add, that if we go with the wrapping technique, it means that we could add `CosmosError` later if/when needed.

sdk/data/azcosmos/cosmos_error.go
ealsur: I don't mind using azcore.ResponseError, is that a new interface/type in the new azcore version?

sdk/data/azcosmos/example_test.go
jhendrixMSFT: Fixed.

sdk/data/azcosmos/cosmos_error.go
jhendrixMSFT: Fixed

sdk/data/azcosmos/cosmos_error.go
jhendrixMSFT: Yeah it's something we added in `v0.21.0`.  I've updated the PR with this change.

sdk/data/azcosmos/CHANGELOG.md
```diff
@@ -1,14 +1,12 @@
 # Release History
 
-## 0.1.1 (Unreleased)
+## 0.2.0 (2022-1-13)
```
seankane-msft, line 3:
```suggestion
## 0.2.0 (2022-01-13)
```
Not sure if you need the `01`, but better safe than sorry.

PR #16789, opened 2022-01-10
More resilient updating of ExpiringResource

### LINKED ISSUE #10567: Add TokenRefresh Offset functionality
Specs: https://gist.github.com/jianghaolu/c56bd46dfa96dfff9bd7e5df8aca58e7

chlowell:
Today, `ExpiringResource` attempts to update its value when the resource is within 2 minutes of expiring. Call an update "eager" when it occurs within that 2 minute window, i.e. while the current value is still valid. This PR makes a couple changes to improve resilience around eager updates:
1. When an eager update fails, `ExpiringResource.GetResource()` now returns the current value--which is still valid--and no error
2. `ExpiringResource` waits at least 30 seconds between eager updates, to reduce the chance of hitting a rate limit
    - this limit doesn't apply when the resource has expired or hasn't been acquired for the first time

I also enlarged the window for eager updates to 5 minutes, to align with other SDKs.

Closes #10567

### DIFF
--- a/sdk/azcore/internal/shared/expiring_resource.go
+++ b/sdk/azcore/internal/shared/expiring_resource.go
@@ -14,7 +14,7 @@ import (
 // AcquireResource abstracts a method for refreshing an expiring resource.
 type AcquireResource func(state interface{}) (newResource interface{}, newExpiration time.Time, err error)
 
-// ExpiringResource is a temporal resource (usually a credential), that requires periodic refreshing.
+// ExpiringResource is a temporal resource (usually a credential) that requires periodic refreshing.
 type ExpiringResource struct {
 	// cond is used to synchronize access to the shared resource embodied by the remaining fields
 	cond *sync.Cond
@@ -27,6 +27,9 @@ type ExpiringResource struct {
 
 	// expiration indicates when the shared resource expires; it is 0 if the resource was never acquired
 	expiration time.Time
+
+	// lastAcquired indicates when a thread/goroutine last attempted to acquire/update the resource
+	lastAcquired time.Time
 
 	// acquireResource is the callback function that actually acquires the resource
 	acquireResource AcquireResource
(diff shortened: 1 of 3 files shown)


sdk/azcore/internal/shared/expiring_resource.go
```diff
@@ -43,13 +46,15 @@ func (er *ExpiringResource) GetResource(state interface{}) (interface{}, error)
...
+	const window = 5 * time.Minute   // This example updates the resource 5 minutes prior to expiration
+	const backoff = 30 * time.Second // Minimum wait time between eager update attempts
 
-	now, acquire, resource := time.Now(), false, er.resource
+	now, acquire, expired, resource := time.Now(), false, false, er.resource
 	// acquire exclusive lock
 	er.cond.L.Lock()
 	for {
-		if er.expiration.IsZero() || er.expiration.Before(now) {
+		expired = er.expiration.IsZero() || er.expiration.Before(now)
```
seankane-msft, line 56: Should the second conditional add the `window` from above to trigger a re-acquire? It won't trigger if the expiration is within 5 minutes of the expiration time (I believe).

sdk/azcore/internal/shared/expiring_resource.go
```diff
@@ -28,6 +28,9 @@ type ExpiringResource struct {
 	// expiration indicates when the shared resource expires; it is 0 if the resource was never acquired
 	expiration time.Time
 
+	// lastAcquired indicates when a thread/goroutine last attempted to acquire/update the resource
+	lastAcquired time.Time
```
jhendrixMSFT, line 32: Nit, from the usage and comment I think this would be better named `lastAttempt`

sdk/azcore/internal/shared/expiring_resource.go
```diff
@@ -59,8 +64,9 @@ func (er *ExpiringResource) GetResource(state interface{}) (interface{}, error)
 			// Getting here means that this thread/goroutine will wait for the updated resource
 		} else if er.expiration.Add(-window).Before(now) {
 			// The resource is valid but is expiring within the time window
-			if !er.acquiring {
-				// If another thread/goroutine is not acquiring/renewing the resource, this thread/goroutine will do it
+			if !er.acquiring && er.lastAcquired.Add(backoff).Before(now) {
```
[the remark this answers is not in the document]
chlowell: The test fails because the callback returns `time.Now()` as the value's expiration time. When clock resolution is low or the runtime is fast (apparently one of these is the case on Windows), the second `GetResource()`  gets the same value from `time.Now()`, finds the current time equal to the expiration time, and doesn't refresh because the last attempt was less than 30 seconds prior. I think it's a test bug--if the test intends its first resource value to be expired on arrival, it should return a past expiration time for that value.

As for whether `GetResource` should refresh in this case anyway, I think it shouldn't. The STS counts all requests equally. The intent of the 30 second wait is to avoid hitting a rate limit with an unnecessary request. In this case the resource hasn't expired, so a refresh is unnecessary.

sdk/azcore/internal/shared/expiring_resource.go
seankane-msft: Yes that does answer my question, thank you!

jhendrixMSFT (approved):

PR #16790, opened 2022-01-11
[azservicebus] Fixing recovery bugs in the Sender and Management

### LINKED ISSUE #16695: amqp[:]link:detach-forced errors (servicebus)
Not sure if it's bug, but we're getting these message when we try to send a message:

> link detached, reason: *Error{Condition: amqp:link:detach-forced, Description: The link 'G6:73468206:bfuTLxfCTv-vJErFRs2RKNeKGfRKB74MuRl7L47653nVdIPUP4uSOw' is force detached. Code: publisher(link12860924). Details: AmqpMessagePublisher.IdleTimerExpired: Idle timeout: 00:10:00., Info: map[]}

or

> link detached, reason: *Error{Condition: amqp:link:detach-forced, Description: The link '15CpdZ60BgL-L4Tls0se1k2_PAn4vk8OX7wH_vfQDag_MQnkF8ikGg' is force detached. Code: RenewToken. Details: Unauthorized access. 'Send' claim(s) are required to perform this operation. 

Initially the messages are sent, but after a while we see one of these errors..

If it isn't a bug, how should handle this? Or try reconnecting?

I'm using these packages:

> github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v0.3.2
> github.com/Azure/azure-event-hubs-go/v3 v3.3.17

In the past ( using `github.com/Azure/azure-service-bus-go v0.11.3`) there was a way to recover:

```
        err := q.sender.Send(ctx, &msg)
        if err != nil {
                if err = q.sender.Recover(ctx); err != nil {
                        fmt.Printf("azure.queue.sender: can't recover sender %T - %s", err, err.Error())
                        return "", err
                }
        
                q.sender.Send(ctx, &msg)
                //...
        }
```

Any ideas would help. Thank you

richardpark-msft:
Fixing sender and management bugs with connection/link recovery

The management link was incorrectly using a RWMutex (using RLock in an area that most definitely was a 'write' and needed to be guarded differently). After removing that one RLock it just ended up being a Mutex, so switched that back. This bug was leading to some panic's in the recovery testing I'm doing on another branch.

The sender wasn't doing recovery. Originally it was going to happen at a lower level, but that didn't materialize so the Sender couldn't handle detaches, etc.. This has been fixed, using the simpler Retry mechanism that I've got in another branch as well. This isn't plumbed all the way through yet, but once it is we can eliminate our 3rd party dependency on the Backoff lib.

Fixes #16695


sdk/messaging/azservicebus/session_receiver_test.go
```diff
@@ -132,7 +132,7 @@ func TestSessionReceiver_acceptSessionButAlreadyLocked(t *testing.T) {
 	// You can address a session by name which makes lock contention possible (unlike
 	// messages where the lock token is not a predefined value)
 	receiver, err = client.AcceptSessionForQueue(ctx, queueName, "session-1", nil)
-	require.True(t, internal.IsSessionLockedError(err))
+	require.EqualValues(t, internal.RecoveryKindFatal, internal.ToSBE(context.Background(), err).RecoveryKind)
```
jhendrixMSFT, line 135: At present, the RecoveryKindFatal isn't publicly available.  How does this get exposed up to the caller?

sdk/messaging/azservicebus/sender.go
```diff
@@ -39,33 +40,65 @@ type MessageBatchOptions struct {
...
-		return nil, err
-	}
+	err := internal.Retry(ctx, "send", func(ctx context.Context, args *internal.RetryFnArgs) error {
+		if args.LastErr != nil {
+			if err := s.links.RecoverIfNeeded(ctx, lastRevision, args.LastErr); err != nil {
+				return err
+			}
+		}
+
+		sender, _, _, lr, err := s.links.Get(ctx)
```
jhendrixMSFT, line 53: I assume this is the thing we're retrying as it does network I/O

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -208,6 +198,59 @@ func shouldRecreateConnection(ctxForLogging context.Context, err error) bool {
...
+type ServiceBusError struct {
+	inner        error
+	RecoveryKind recoveryKind
+}
+
+func (sbe *ServiceBusError) String() string {
+	return sbe.inner.Error()
+}
+
+func (sbe *ServiceBusError) AsError() error {
```
jhendrixMSFT, line 217: This should be `Unwrap()` so that it seemlessly works with `errors.As()`, see [docs](https://pkg.go.dev/errors).

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -208,6 +198,59 @@ func shouldRecreateConnection(ctxForLogging context.Context, err error) bool {
...
+	}
+
+	return sbe
+}
+
+func isPermanentNetError(err error) bool {
+	var netErr net.Error
+
+	if errors.As(err, &netErr) {
+		temp := netErr.Temporary()
```
jhendrixMSFT, line 242: This is unreliable and should be avoided.  E.g. a DNS hiccup would return true here but it's really transient.

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -208,6 +198,59 @@ func shouldRecreateConnection(ctxForLogging context.Context, err error) bool {
...
+		timeout := netErr.Timeout()
+		return !temp && !timeout
+	}
+
+	return false
+}
+
+func isEOF(err error) bool {
```
jhendrixMSFT, line 250: Nit but does this wrapper help?

sdk/messaging/azservicebus/internal/errors.go
jhendrixMSFT: In fact I'd yank this entire func.

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -229,17 +272,89 @@ func IsDrainingError(err error) bool {
...
-	if !errors.As(err, &amqpError) {
-		return false
+func GetRecoveryKind(ctxForLogging context.Context, err error) recoveryKind {
+	if IsCancelError(err) {
+		return RecoveryKindFatal
+	}
+
+	var netErr net.Error
+
+	if errors.As(err, &netErr) {
```
jhendrixMSFT, line 305: Remove

sdk/messaging/azservicebus/session_receiver_test.go
richardpark-msft: There isn't a way for external people to use this. It's purely for internal classification.

I think this properly falls into the case of "the best you can do is log" since the lock itself is taken by some other process/session and you can't stop them.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: I think I named this badly - it's not an `error` (although it was for a little bit). It's just...my wrapper around error that's not intended to be passed directly. Can you think of a less misleading name?

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: The fall-through will basically mark this as RecoveryKindFatal, however. It seems like I should possibly classify it as RecoveryKindNone, indicating a retry would make sense.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: I have a note that I'm not even sure if I need this at all. I can inline it (it's even only used in one spot!)

sdk/messaging/azservicebus/sender.go
richardpark-msft: Yes, we could minimize the block but I mostly just like to make it simpler to read through.

sdk/messaging/azservicebus/internal/errors.go
jhendrixMSFT: This goes back to that conversation with Jeff about retries.  The TL;DR is that unless you know for sure that an error is terminal, you should consider it as transient (this only works when there's a retry cap, which you have).

sdk/messaging/azservicebus/internal/errors.go
jhendrixMSFT: I get it now, I missed that ServiceBusError isn't an actual error.  Maybe ErrorInfo or something like that.

richardpark-msft: (I'll rename this in the other PR to make my merge less gross)

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: Inlined.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: Done.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: Removed. Basically, just doing what you noted in another spot - treat any net.Error as transient.

jhendrixMSFT: I thought `isPermanentNetError` was going away?

sdk/messaging/azservicebus/internal/errors.go
jhendrixMSFT: I still don't think there's a need to check for `net.Error`.  My pseudo-code algorithm in my head is "if err is not in this list of things we know to be terminal then retry".

sdk/messaging/azservicebus/internal/errors.go
jhendrixMSFT: OH maybe it's going away in the second PR?

richardpark-msft:
(just updating to the actual published azcore and azidentity, rather than their commit hashes)

sdk/messaging/azservicebus/internal/retrier.go
```diff
@@ -130,3 +133,136 @@ func (rp *backoffRetrier) Try(ctx context.Context) bool {
 		return false
 	}
 }
+
+type RetryFnArgs struct {
+	I int32
+	// LastErr is the returned error from the previous loop.
+	// If you have potentially expensive
```
chlowell, line 140: Missing some comment here?

chlowell (approved):

sdk/messaging/azservicebus/internal/retrier.go
richardpark-msft: (not ignoring this, will come back in the next PR and take a look)

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: (Yeah, I'm deleting it in the next PR).

PR #16810, opened 2022-01-12
[KeyVault] Keys - disabling broken tests

seankane-msft:
The service disabled the Key Rotation endpoints while they investigate a service side bug. I am disabling these tests (and lowering code coverage), so I can release the library. When the service fixes the bug and enables the endpoints again the tests will be enabled (and coverage raised).

### DIFF
--- a/eng/config.json
+++ b/eng/config.json
@@ -18,7 +18,7 @@
         },
         {
             "Name": "keyvault/azkeys",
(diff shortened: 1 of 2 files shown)


richardpark-msft (approved):
Seems fine, I assume you're coming back to this soon anyways and it's just the tests.

PR #16815, opened 2022-01-12
Adding manifest generation to build

scbedd:
Related to Azure/azure-sdk-tools#2273

Thanks @seankane-msft and @richardpark-msft for the help here!


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -66,6 +66,15 @@ steps:
...
       condition: succeededOrFailed()
 
+  - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
+    displayName: 'Generate BOM'
+    inputs:
+      BuildDropPath: ${{ parameters.ServiceDirectory }}
+
+  - publish: ${{ parameters.ServiceDirectory }}/_manifest
+    displayName: 'Publish BOM Manifest'
+    artifact: 'manifest'
```
weshaggard, line 76: Should we be consistent and call all the uploads manifest for the other languages as well?

weshaggard (approved):

eng/pipelines/templates/steps/build-test.yml
scbedd: Yessir. I'm actually adjusting these in the original PRs to match.

eng/pipelines/templates/jobs/archetype-sdk-client-samples.yml
```diff
@@ -14,18 +14,22 @@ stages:
             pool.name: azsdk-pool-mms-ubuntu-2004-general
             image.name: MMSUbuntu20.04
             go.version: '1.17'
+            generate.bom: false
```
weshaggard, line 17: Given this is the default value is it worth copying everywhere?

scbedd: You have a good point. I'll try it.

PR #16817, opened 2022-01-12
Test-Proxy Runs using tool

scbedd:
...rather than docker.

@seankane-msft 

I have no explanation (and can't get one short term) about why the test proxy refuses to properly start on our CI machines. I want to say that there is something wrong with the actual instances of the machines given the irregularity with which this appears.


seankane-msft (approved):

scbedd:
Additionally, I have run this manually against the problematic [azkeys build.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1291537&view=results)

scbedd:
Given I have signoff from @seankane-msft , and have also run the previously problematic azkeys build, I'm going to merge this.

PR #16831, opened 2022-01-14
[azservicebus] Fixing several reliability/recovery bugs

### LINKED ISSUE #15170: [service-bus] Connection/network recovery

### LINKED ISSUE #15610: [service-bus] Error handling needs work to be consistent and reliable

### LINKED ISSUE #15642: [service-bus] Unify all the various retry mechanisms
We've got a mixture of amqp-common Retry and the nice Backoff package that Joel used in EventHubs.

I added in a retrier.go that blends together Backoff with a limited retry count that has a simpler flow since it's just used in a normal for loop, rather than a callback. 

I need to do a couple of things:
- Remove usages of amqp-common Retry, replace with retrier
- Establish some reasonable defaults (not necessarily uniform per control)
- Take the policy as a parameter for some internal APIs

### LINKED ISSUE #15644: [service-bus] amqp-common's retryableRPC method returns string errors
The retryableRPC() function returns errors as strings. 

There's enough information in the underlying RPC call that should give us some more programatically actionable fields in the error. We should fix it so it does just that, rather than forcing us to parse strings to figure out what the error is.

### LINKED ISSUE #16088: [azservicebus] amqp-common's link.RetryableRPC() retries an entity does not exist
This can lead to a series of retries that will not work. It'd be nice to be able to pass in an error handling strategy per RPC call so we can provide better error handling when we know what the context of the call is and just fail early.

The underlying error is:

Code: 404
Description: "The messaging entity 'sb://<mysb>.servicebus.windows.net/samples' could not be found. To know more visit https://aka.ms/sbResourceMgrExceptions.  TrackingId:888b355f-038e-42e1-ab5d-cf1d0bfa20d1_G53, SystemTracker:riparkdev2.servicebus.windows.net:samples, Timestamp:2021-11-04T17:42:29"

richardpark-msft:
Fixing several reliability and recovery bugs that were affecting sending and receiving. This PR finally exports retry options so callers can configure how long and how many retries they want.

The big changes:
* Retry configuration options are now exposed through the ClientOptions for both the Admin client and the Service Bus client.
* amqpLinks.go has it's own Retry mechanism, based on the azcore retry algorithm. This does exponential backoff, jitter and is configurable using Client.RetryOptions. This retry mechanism cleans up several areas that were doing the BackoffRetrier before.
* MgmtClient is gone, and we just use rpcLink now. There were some coordination issues that just weren't working properly, so the rpc link code has been moved into azservicebus for now (it was always an internal type and remains so). 
* Sender/Receiver both use the better amqplinks.Retry() and their code has been structured to make that easier to read (hopefully!)

Bug fixes:
* Fixed a bug where the proper error wasn't being broadcast when the connection failed in rpc, preventing proper connection recovery.
* Fixed a bug where links wouldn't properly recreate if the connection recovery happened in the claim negotiation.

Fixes #15170 (Connection/network recovery)
Fixes #15610 (Error handling needs work to be consistent and reliable)
Fixes #15642 (Unify all the various retry mechanisms)

These were fixed by more or less sidestepping retryableRPC() mechanism:
Fixes #15644 (amqp-common's retryableRPC method returns string errors)
Fixes #16088 (amqp-common's link.RetryableRPC() retries an entity does not exist)


sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -21,8 +22,12 @@ type Client struct {
 	em atom.EntityManager
 }
 
+type RetryOptions = utils.RetryOptions
```
jhendrixMSFT, line 25: Missing doc comment.  I just copy the one from the aliased type.  It's a shame VSCode doesn't propagate the comment for you if it's missing from the outer type.

sdk/messaging/azservicebus/go.mod
```diff
@@ -10,7 +10,6 @@ require (
 	github.com/Azure/go-amqp v0.17.0
 	github.com/devigned/tab v0.1.1
 	github.com/joho/godotenv v1.3.0
-	github.com/jpillora/backoff v1.0.0
```
jhendrixMSFT, line 13: Missing `go mod tidy`?  I don't see a `go.sum` update.

sdk/messaging/azservicebus/internal/atom/errors.go
```diff
@@ -6,10 +6,15 @@ package atom
...
 	"fmt"
 	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-func NewResponseError(inner error, resp *http.Response) error {
-	return ResponseError{inner, resp}
+func NewResponseError(resp *http.Response) error {
+	return &azcore.ResponseError{
```
jhendrixMSFT, line 14: Just confirming there's no useful error code to return?

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -38,7 +43,13 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
...
 // NewClient creates a Client authenticating using a TokenCredential.
 func NewClient(fullyQualifiedNamespace string, tokenCredential azcore.TokenCredential, options *ClientOptions) (*Client, error) {
-	em, err := atom.NewEntityManager(fullyQualifiedNamespace, tokenCredential, internal.Version)
+	var retryOptions utils.RetryOptions
+
+	if options != nil && options.RetryOptions != nil {
+		retryOptions = *options.RetryOptions
+	}
+
+	em, err := atom.NewEntityManager(fullyQualifiedNamespace, tokenCredential, internal.Version, retryOptions)
```
richardpark-msft, line 52: @jhendrixMSFT - just to note, this is a really temporary situation. The ATOM client is still, mostly, just using the old code from the original SB package. I want to move it over to look more like the generated clients so I can just use the same policy_retry, etc...

sdk/messaging/azservicebus/internal/atom/errors.go
richardpark-msft: From what I could see the codes are just free form string messages. So 'no', but they will show up in the .Error() since they're in the body.

sdk/messaging/azservicebus/internal/utils/retrier.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
richardpark-msft, line 1: @jhendrixMSFT: This is just moved. I had to do a little bit of package splitting to make it so I could use the entityManager in the amqplinks_test.go (otherwise there were a few spots where circular references were happening).

This is basically the same as the alg that's in azcore right now for retry, minus HTTP status code checking. It's also internal for now, in case we decide to do some consolidation later.

sdk/messaging/azservicebus/internal/mgmt.go
```diff
@@ -34,171 +31,39 @@ const (
...
-
-	ScheduleMessages(ctx context.Context, enqueueTime time.Time, messages ...*amqp.Message) ([]int64, error)
-	CancelScheduled(ctx context.Context, seq ...int64) error
-
-	RenewLocks(ctx context.Context, linkName string, lockTokens []amqp.UUID) ([]time.Time, error)
-	RenewSessionLock(ctx context.Context, sessionID string) (time.Time, error)
-
-	GetSessionState(ctx context.Context, sessionID string) ([]byte, error)
-	SetSessionState(ctx context.Context, sessionID string, state []byte) error
+type mgmtError struct {
```
richardpark-msft, line 34: @jhendrixMSFT: this whole file went through some refactoring. Basically, the mgmtClient used to be a struct and it held onto an RPCLink, which it would protect using a lock. 

Looking at it more, it just seemed like overkill and I already had all this kind of locking code in amqpLinks.go. So I basically just made each function standalone here, and made it take an amqpLinks instead.

I also modified amqpLinks to return an RPCLink instead of a MgmtClient. This just makes things simpler since it means link management remains in amqplinks.go and isn't spread around anymore.

sdk/messaging/azservicebus/internal/mgmt.go
```diff
@@ -306,7 +164,7 @@ func (mc *mgmtClient) PeekMessages(ctx context.Context, fromSequenceNumber int64
 		msg.ApplicationProperties["server-timeout"] = uint(time.Until(deadline) / time.Millisecond)
 	}
 
-	rsp, err := mc.doRPCWithRetry(ctx, msg, 5, 5*time.Second)
+	rsp, err := doRPC(ctx, "peek", rpcLink, msg)
```
richardpark-msft, line 167: @jhendrixMSFT: most of the changes in this file are going to look like this - removing the call to the implicit retries and relying on the caller to initiate and run the retries.

This is mostly to blunt issues where you have nested retries, so you end up with nxn attempts. So, for the most part, only the clients do the retries.

sdk/messaging/azservicebus/client.go
```diff
@@ -14,23 +14,30 @@ import (
...
+	// is just to make it the first value in the struct
+	// See:
+	//   Godoc: https://pkg.go.dev/sync/atomic#pkg-note-BUG
+	//   PR: https://github.com/Azure/azure-sdk-for-go/pull/16847
 	linkCounter uint64
-	links       map[uint64]internal.Closeable
-	creds       clientCreds
-	namespace   interface {
+
+	linksMu   *sync.Mutex
```
jhendrixMSFT, line 31: The ordering of these changed, was that intentional?

sdk/messaging/azservicebus/internal/rpc.go
```diff
@@ -0,0 +1,445 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
richardpark-msft, line 1: @jhendrixMSFT: this is just a copy of the code from messaging/internal/rpc.go with changes to:
- Remove the retryableRPC function, since retries are handled by callers, not here.
- Fixing a small bug in where we were always returning a detached error. Now we broadcast the actual last error we got, which helps with recovery.

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -215,26 +211,21 @@ func (ns *Namespace) NewAMQPSession(ctx context.Context) (AMQPSessionCloser, uin
 	return session, clientRevision, err
 }
 
-// NewMgmtClient creates a new management client with the internally cached *amqp.Client.
```
richardpark-msft, line 218: @jhendrixMSFT - no longer needed since we centered around using the rpcLink in amqpLinks, rather than using ManagementClient.

sdk/messaging/azservicebus/client.go
richardpark-msft: Just reordering for it be logical and to keep the mutex near the stuff it actually protects (it's not perfect, but the alignment requirement makes it less easy)

sdk/messaging/azservicebus/receiver.go
```diff
@@ -333,118 +348,114 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
 	defaultTimeAfterFirstMessage := 20 * time.Millisecond
 
 	if r.receiveMode == ReceiveModeReceiveAndDelete {
 		defaultTimeAfterFirstMessage = time.Second
 	}
 
-	messages, err := r.getMessages(ctx, receiver, maxMessages, defaultTimeAfterFirstMessage)
+	var linksWithID *internal.LinksWithID
+
+	err := r.amqpLinks.Retry(ctx, "receiveMessages", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
```
richardpark-msft, line 364: @jhendrixMSFT: the changes in here are a bit more meaty than other spots. 

Basically, I just wanted to make sure that the retry() loop existed around the entire receive operation. Before this it was only around certain spots, which made it harder to just read through (I think). I also inlined some code to make it more obvious what is happening.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -46,40 +51,39 @@ type AMQPLinks interface {
...
 	// the AMQP session for either the 'sender' or 'receiver' link
 	session AMQPSessionCloser
 
 	// these are populated by your `createLinkFunc` when you construct
 	// the amqpLinks
-	sender   AMQPSenderCloser
-	receiver AMQPReceiverCloser
-
-	// last connection revision seen by this links instance.
-	clientRevision uint64
```
richardpark-msft, line 78: These two "revisions" (which are basically just IDs) are in the LinkID struct now.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,238 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
 
 	if closedPermanently {
 		span.AddAttributes(tab.StringAttribute("outcome", "was_closed_permanently"))
-		return errClosedPermanently{}
+		return ErrNonRetriable{Message: "Link has been closed permanently"}
 	}
 
-	if theirLinkRevision != nil && ourLinkRevision > *theirLinkRevision {
+	// cheap check before we do the lock
+	if ourLinkRevision.Link != theirLinkRevision.Link {
```
richardpark-msft, line 140: I'll do a naming-rename pass in another PR since this one is already big enough.

The basics are that I was previously calling this a revision, which implied some things (for instance, that you could meaningfully compare them). In reality it's just a unique ID, generated without any fanfare, and should just be treated as opaque.

Since it is an incrementing ID, however, it's useful for me for debugging purposes since it makes it simpler to tell how often something has had to be recreated.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,238 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
-	if shouldRecreateLink(origErr) {
-		span.AddAttributes(
-			tab.StringAttribute("recovery_kind", "link"),
-			tab.StringAttribute("error", origErr.Error()),
-			tab.StringAttribute("error_type", fmt.Sprintf("%T", origErr)))
+	sbe := GetSBErrInfo(origErr)
 
-		if err := links.recoverLink(ctx, &linksRevision); err != nil {
-			span.AddAttributes(tab.StringAttribute("recoveryFailure", err.Error()))
+	if sbe.RecoveryKind == RecoveryKindLink {
```
richardpark-msft, line 189: Using 'sbe' here instead of the old technique made it simpler to group all of the classification logic into a single spot. It's still all the same logic though, just made a bit simpler by returning constants.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,238 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
-		if err := links.recoverConnection(ctx); err != nil {
-			span.Logger().Error(fmt.Errorf("failed to recreate connection: %w", err))
-			return err
-		}
-
-		// unconditionally recover the link if the connection died.
-		if err := links.recoverLink(ctx, nil); err != nil {
-			span.Logger().Error(fmt.Errorf("failed to recover links after connection restarted: %w", err))
+	} else if sbe.RecoveryKind == RecoveryKindConn {
+		if err := links.recoverConnection(ctx, theirID); err != nil {
```
richardpark-msft, line 198: recoverConnection was changed so that it recovers the connection _and_ the links. Before they were separate, which meant that we weren't always recreating the links after the connection had been recreated. This was just wasteful, so now they're one operation.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,238 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
-	err := links.ns.Recover(ctx, clientRevision)
+	created, err := links.ns.Recover(ctx, uint64(theirID.Conn))
 
 	if err != nil {
-		tab.For(ctx).Error(fmt.Errorf("Recover connection failure: %w", err))
+		log.Writef(EventConn, "Recover connection failure: %s", err)
 		return err
 	}
 
+	if created || theirID.Link == links.id.Link {
```
richardpark-msft, line 224: Basically, recovery here is:
- If we recreated the connection we _definitely_ need to recreate our link
- If the link we got an error on matches our current link we need to recreate our link. (if the ID _didn't_ match then that indicates we already recovered, so they just need to re-get the link).

sdk/messaging/azservicebus/admin/admin_client.go
richardpark-msft: Added doc comment. I'll remove this in the next PR when I get this HTTP client consistent with the actual azcore-way.

sdk/messaging/azservicebus/go.mod
richardpark-msft: Tidy done and uploaded. Removes the backoff package entirely now.

sdk/messaging/azservicebus/internal/amqpLinks.go
jhendrixMSFT: This explanation would probably be a good code comment here.

richardpark-msft: Yeah, was thinking the same thing when I wrote it out. :) Will add.

richardpark-msft: Added some more comments on this and the function it calls.

sdk/messaging/azservicebus/internal/utils/retrier.go
jhendrixMSFT: Presumably we'll need this, or something similar for EH?  Makes me think that we might want an `amqpcore` or something so that we can have a pipeline/policies that handle `*amqp.Message` types.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
```diff
@@ -159,8 +159,9 @@ function DeployStressPackage(
     $dockerFilePath = "$($pkg.Directory)/Dockerfile"
     if ($pushImages -and (Test-Path $dockerFilePath)) {
         Write-Host "Building and pushing stress test docker image '$imageTag'"
-        $dockerFile = Get-ChildItem $dockerFilePath
-        Run docker build -t $imageTag -f $dockerFile.FullName $dockerFile.DirectoryName
+        $dockerFile = Get-ChildItem "$($pkg.Directory)/Dockerfile"
```
benbp, line 162: This will need to be removed. @ckairen do you think you would be able to get to https://github.com/Azure/azure-sdk-tools/issues/2569 this week?

benbp: I think it's just a matter of copying these changes and adding some docs for the new chart annotation fields located [here](https://github.com/Azure/azure-sdk-for-go/pull/16831/files#diff-fa372fad5b3dbf6bab1bb08336692324876f4c04553e99e28c4ba381dd1e48dbR9-R10).

sdk/messaging/azservicebus/internal/stress/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
+name: go-sb-stress
+description: Stress tests for Go
+version: 0.1.1
+appVersion: v0.1
+annotations:
+  stressTest: 'true'  # enable auto-discovery of this test via `find-all-stress-packages.ps1`
+  namespace: 'go-sb-stress'
```
benbp, line 8: I had been thinking that namespaces would be better to mirror the language repo for organization purposes?

```suggestion
  namespace: 'golang'
```

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
richardpark-msft: Yeah, it's in here so what's in the branch can deploy, but it's not meant to be permanent.

benbp: Yeah only reason I bring it up is our tooling will prevent you from merging due to changes in eng/common.

benbp:
Stress test config changes look good, with some minor nits.

sdk/messaging/azservicebus/internal/stress/Chart.yaml
richardpark-msft: Done, but just used 'go'.

sdk/messaging/azservicebus/internal/atom/entity_manager.go
```diff
@@ -238,61 +243,90 @@ func (em *entityManager) Delete(ctx context.Context, entityPath string, mw ...Mi
...
+		return false
 	}
 
-	if resp != nil {
-		return nil, NewResponseError(err, resp)
+	var respErr *azcore.ResponseError
+
+	// TODO: this is very much temporary. We need to move this over to the azcore HTTP stack.
+	if errors.As(err, &respErr) {
+		if respErr.StatusCode == http.StatusRequestTimeout || // 408
```
seankane-msft, line 319: Are only these specific error codes fatal or is it anything in the 4XX and 5XX range? If the latter is true you could do: 
```
if errors.As(err, &respErr) {
     return respErr.StatusCode > 400
}
```

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
ckairen: Ok, I'll rush this this week

sdk/messaging/azservicebus/internal/utils/retrier.go
richardpark-msft: I think that's a great idea. amqpLinks as well.

(although I'm waiting to do that until absolutely necessary since it'll slow down development of azservicebus at this moment).

sdk/messaging/azservicebus/go.mod
```diff
@@ -10,7 +10,6 @@ require (
 	github.com/Azure/go-amqp v0.17.0
 	github.com/devigned/tab v0.1.1
 	github.com/joho/godotenv v1.3.0
```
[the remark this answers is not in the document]
richardpark-msft: This is a dev dep _only_ (ie, it's used for the tests). There's no distinction in here about it, but we don't use it in production code at all.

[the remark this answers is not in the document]
jhendrixMSFT: It might be worth grouping test-only dependencies into a separate `require` clause with a preceding comment that says these are for tests only.  I believe that `go mod tidy` and other tools will preserve this structure (we should verify this before making the change).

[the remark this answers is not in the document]
richardpark-msft: That's a great idea, I can at least try that.

I do wish there was a way to make it enforcing, but we'll go with this at least.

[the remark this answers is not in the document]
richardpark-msft: It works! I updated it and split things a bit nicer.

sdk/messaging/azservicebus/internal/atom/entity_manager.go
richardpark-msft: It's these specific ones, but this list just matches what's in azcore right now for the 'default retryable status codes'.

Some codes (like unauthorized, for instance) would not be retryable.

richardpark-msft:
> @benbp : Stress test config changes look good, with some minor nits.

I didn't realize that the changing of a .ps1 in `eng` would result in the system failing me out. I've reverted it, but I'll remember this for a long time.

benbp:
> I didn't realize that the changing of a .ps1 in `eng` would result in the system failing me out. I've reverted it, but I'll remember this for a long time.

@richardpark-msft because `eng/common` contents get synced from azure-sdk-tools, we can't allow downstream changes otherwise chaos ensues trying to make automatic patches.

richardpark-msft:
> > I didn't realize that the changing of a .ps1 in `eng` would result in the system failing me out. I've reverted it, but I'll remember this for a long time.
> 
> @richardpark-msft because `eng/common` contents get synced from azure-sdk-tools, we can't allow downstream changes otherwise chaos ensues trying to make automatic patches.

It's strict, but fair.

sdk/messaging/azservicebus/internal/rpc.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	var statusCode int
+	statusCodeCandidates := []string{statusCodeKey, altStatusCodeKey}
+	for i := range statusCodeCandidates {
+		if rawStatusCode, ok := res.ApplicationProperties[statusCodeCandidates[i]]; ok {
+			if cast, ok := rawStatusCode.(int32); ok {
+				statusCode = int(cast)
+				break
+			} else {
```
jhendrixMSFT, line 258: Nit; break makes the else unnecessary.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -74,72 +77,81 @@ type ReceiverOptions struct {
...
+	} else {
+		if err := checkReceiverMode(options.ReceiveMode); err != nil {
+			return err
+		}
 
-	if err := checkReceiverMode(options.ReceiveMode); err != nil {
-		return err
+		receiver.receiveMode = options.ReceiveMode
+
+		if err := entity.SetSubQueue(options.SubQueue); err != nil {
```
jhendrixMSFT, line 97: Why was this moved to this `else` clause?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -516,3 +527,11 @@ func createLinkOptions(mode ReceiveMode, entityPath string) []amqp.LinkOption {
 
 	return opts
 }
+
+func checkReceiverMode(receiveMode ReceiveMode) error {
+	if receiveMode == ReceiveModePeekLock || receiveMode == ReceiveModeReceiveAndDelete {
+		return nil
+	} else {
```
jhendrixMSFT, line 534: Nit, unnecessary `else` (the linter should be catching these, need to check).

sdk/messaging/azservicebus/sender.go
```diff
@@ -77,55 +63,23 @@ func (s *Sender) NewMessageBatch(ctx context.Context, options *MessageBatchOptio
...
-		}
-
-		sender, _, _, lr, err := s.links.Get(ctx)
-
-		if err != nil {
-			return err
-		}
-
-		lastRevision = lr
+	return s.links.Retry(ctx, "SendMessage", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
```
jhendrixMSFT, line 66: Why doesn't this retry loop need to `ResetAttempts()` like receiver?

sdk/messaging/azservicebus/sender.go
```diff
@@ -184,14 +140,21 @@ func (sender *Sender) createSenderLink(ctx context.Context, session internal.AMQ
...
 
-func newSender(ns internal.NamespaceWithNewAMQPLinks, queueOrTopic string, cleanupOnClose func()) (*Sender, error) {
+type newSenderArgs struct {
+	ns             internal.NamespaceWithNewAMQPLinks
+	queueOrTopic   string
+	cleanupOnClose func()
+	retryOptions   RetryOptions
+}
+
+func newSender(args newSenderArgs, retryOptions RetryOptions) (*Sender, error) {
```
jhendrixMSFT, line 150: Redundant `retryOptions` arg?

sdk/messaging/azservicebus/client.go
```diff
@@ -45,8 +52,16 @@ type ClientOptions struct {
...
 	NewWebSocketConn func(ctx context.Context, args NewWebSocketConnArgs) (net.Conn, error)
+
+	// RetryOptions controls how often operations are retried from this client and any
+	// Receivers and Senders created from this client.
+	RetryOptions RetryOptions
 }
 
+// RetryOptions controls how often operations are retried from this client and any
+// Receivers and Senders created from this client.
+type RetryOptions utils.RetryOptions
```
jhendrixMSFT, line 63: `type RetryOptions = utils.RetryOptions`

sdk/messaging/azservicebus/internal/atom/entity_manager.go
```diff
@@ -238,61 +243,90 @@ func (em *entityManager) Delete(ctx context.Context, entityPath string, mw ...Mi
...
+		return nil, err
+	}
+
+	return finalResp, nil
+}
+
+func isFatalHTTPError(err error) bool {
+	var netErr net.Error
+
+	if errors.As(err, &netErr) {
```
jhendrixMSFT, line 311: I think this can go away.

sdk/messaging/azservicebus/internal/atom/entity_manager.go
```diff
@@ -238,61 +243,90 @@ func (em *entityManager) Delete(ctx context.Context, entityPath string, mw ...Mi
...
+	}, isFatalHTTPError, em.retryOptions)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return finalResp, nil
+}
+
+func isFatalHTTPError(err error) bool {
```
jhendrixMSFT, line 308: How is this used, to determine if something should retry (I'm guessing based on checking of HTTP status codes)?

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -345,7 +161,106 @@ func GetRecoveryKind(ctxForLogging context.Context, err error) recoveryKind {
...
 }
+
+type (
+	// ErrMissingField indicates that an expected property was missing from an AMQP message. This should only be
+	// encountered when there is an error with this library, or the server has altered its behavior unexpectedly.
+	ErrMissingField string
+
+	// ErrMalformedMessage indicates that a message was expected in the form of []byte was not a []byte. This is likely
+	// a bug and should be reported.
+	ErrMalformedMessage string
```
jhendrixMSFT, line 194: I know this is internal, but do we really need all these error types?  i.e. do we need to detect and handle them separately?

sdk/messaging/azservicebus/sender.go
jhendrixMSFT: I get it now.

sdk/messaging/azservicebus/internal/atom/entity_manager.go
richardpark-msft: Yeah. This code isn't long for this world - I just want to throw it away and use the retry policy stuff in azcore.

sdk/messaging/azservicebus/internal/atom/entity_manager.go
richardpark-msft: For better or worse my default in this function is to assume an error is fatal. I think we talked about making it so network errors are just always treated as retryable (so that's all this is doing).

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: I don't think there are any great spots where these would be used. Filed https://github.com/Azure/azure-sdk-for-go/issues/16896 to remove them.

eng/common/scripts/stress-testing/stress-test-deployment-lib.ps1
ckairen: https://github.com/Azure/azure-sdk-tools/pull/2599 Here's the PR :)

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Fixed.

sdk/messaging/azservicebus/internal/rpc.go
richardpark-msft: Fixed.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Some of the code that applied receiver options was outside of the call so I merged that into here (which means that we always need to calculate and set the entity path).

sdk/messaging/azservicebus/sender.go
richardpark-msft: Fixed this - it is redundant, but only because I shouldn't have put retryOptions in the args!

sdk/messaging/azservicebus/client.go
richardpark-msft: Fixed.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -333,118 +348,114 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
-		amqpMessage, err := receiver.Receive(ctx)
+	if err != nil && !internal.IsCancelError(err) {
+		return ret(err)
+	}
 
-		if err != nil {
-			if internal.IsCancelError(err) {
-				return messages, nil
+	if needsDrain {
+		// start the drain asynchronously. Note that we ignore the user's context at this point
```
jhendrixMSFT, line 430: Are we sure it's safe to ignore the user's context?  Can this take too long?

sdk/messaging/azservicebus/internal/rpc.go
jhendrixMSFT: Why do we need a copy of the code, is the other one going away?

sdk/messaging/azservicebus/receiver.go
richardpark-msft: All we're really waiting for here is an ACK from the service so it should be fast.

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -249,47 +240,46 @@ func (ns *Namespace) Close(ctx context.Context) error {
...
 	span.AddAttributes(
-		tab.Int64Attribute("revision", int64(ns.clientRevision)),
-		tab.Int64Attribute("requested", int64(clientRevision)))
+		tab.Int64Attribute("connID", int64(ns.connID)),
+		tab.Int64Attribute("theirConnID", int64(theirConnID)))
 
-	if ns.clientRevision > clientRevision {
-		span.Logger().Info(fmt.Sprintf("Skipping recovery, already recovered: %d vs %d", ns.clientRevision, clientRevision))
+	if ns.connID != theirConnID {
+		log.Writef(EventConn, "Skipping connection recovery, already recovered: %d vs %d", ns.connID, theirConnID)
```
jhendrixMSFT, line 259: Are we switching away from logging this info in spans?

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -373,15 +354,38 @@ func (ns *Namespace) startNegotiateClaimRenewer(ctx context.Context,
...
+
+					if err == nil {
+						break
+					}
+
+					// if we fail our retries _and_ we've exceeded the window where our token would have
+					// been good we can just stop.
+					if time.Since(expiresOn) <= 0 {
+						log.Writef(EventAuth, "[%s] token has expired, stopping refresh loop", entityPath)
+						break TokenRefreshLoop
```
jhendrixMSFT, line 387: Nit, Couldn't you just `return` here?

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,243 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
+func NewAMQPLinks(ns NamespaceForAMQPLinks, entityPath string, createLink CreateLinkFunc) AMQPLinks {
+	l := &AMQPLinksImpl{
 		entityPath:        entityPath,
 		managementPath:    fmt.Sprintf("%s/$management", entityPath),
 		audience:          ns.GetEntityAudience(entityPath),
 		createLink:        createLink,
-		baseRetrier:       baseRetrier,
 		closedPermanently: false,
-		revision:          1,
+		id:                LinkID{},
```
jhendrixMSFT, line 111: Nit: this is redundant as `id` will already be zero-initialized.

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -97,240 +101,243 @@ type CreateLinkFunc func(ctx context.Context, session AMQPSession) (AMQPSenderCl
...
-	if shouldRecreateLink(origErr) {
-		span.AddAttributes(
-			tab.StringAttribute("recovery_kind", "link"),
-			tab.StringAttribute("error", origErr.Error()),
-			tab.StringAttribute("error_type", fmt.Sprintf("%T", origErr)))
+	sbe := GetSBErrInfo(origErr)
 
-		if err := links.recoverLink(ctx, &linksRevision); err != nil {
-			span.AddAttributes(tab.StringAttribute("recoveryFailure", err.Error()))
+	if sbe.RecoveryKind == RecoveryKindLink {
```
jhendrixMSFT, line 189: Are there only the two values for `RecoveryKind`?  Wondering if this should be a `switch` with a `default` for an unhandled case (not saying you have to make this change).

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: For now, I think this fits better with what we're doing. 

I have a preliminary sketch from Jeff that outlines tracing and we could consider adding some of this data as attributes at that point.

sdk/messaging/azservicebus/internal/amqpLinks.go
richardpark-msft: There are 4 values today and I believe all of them need to exist:

1. Link -> link has to be recreated (connection is fine)
2. Conn -> connection has to be recreated (and thus, also the links)
3. Fatal -> no amount of anything will resurrect this (for instance, a user closing the client explicitly)
4. None -> can retry, nothing needs to be recreated.

We might be able to do something here but I'm not sure if the `default`/`else` case would help _although_ I could see us optimizing this in the future.

sdk/messaging/azservicebus/internal/rpc.go
jhendrixMSFT: I think GH ate my reply as I don't see it on the main PR page.

jhendrixMSFT (approved):
Still a few open questions but we can move this along.

richardpark-msft:
Merging and will continue some more improvements in a much more incremental PR.

Thanks @jhendrixMSFT , I promise to NEVER* do that to you again.

*(probably)

PR #16851, opened 2022-01-18
Changes for AZURE_CLIENT_SEND_CERTIFICATE_CHAIN

christothes:
This PR adds a new option to `internal/mock` which allows for remote addresses to be routed to the mock server. This change allows for more mock functional identity tests.


sdk/internal/mock/mock.go
```diff
@@ -123,7 +127,20 @@ func (s *Server) Do(req *http.Request) (*http.Response, error) {
 		resp := s.getResponse()
 		return nil, resp.err
 	}
-	resp, err := s.srv.Client().Do(req)
+	var err error
+	var resp *http.Response
+	if s.routeAllRequestsToMockServer /*&& strings.Contains(req.URL.Host, "login")*/ {
```
jhendrixMSFT, line 132: Nit, remove trailing comment.

sdk/internal/mock/mock.go
```diff
@@ -123,7 +127,20 @@ func (s *Server) Do(req *http.Request) (*http.Response, error) {
...
+		originalURL := req.URL
+		mockUrl, _ := url.Parse(req.URL.String())
```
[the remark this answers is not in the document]
christothes: OK, I just wasn't sure if that would be just copying the reference rather than creating a new instance with the values copied.

sdk/azidentity/azidentity_test.go
```diff
@@ -27,6 +27,92 @@ const (
 	customHostString = "https://custommock.com/"
 )
 
+func getTenantDiscoveryResponse(url string) string {
+	return `{` +
```
chlowell, line 31: Backtick syntax allows multiline literals, so you don't have to concatenate every line

sdk/azidentity/environment_credential.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
+const (
+	envVarSendCertChain = "AZURE_CLIENT_SEND_CERTIFICATE_CHAIN"
+)
```
chlowell, line 19: Nitpicking:
```suggestion
const envVarSendCertChain = "AZURE_CLIENT_SEND_CERTIFICATE_CHAIN"
```

sdk/azidentity/environment_credential_test.go
```diff
@@ -173,6 +179,50 @@ func TestEnvironmentCredential_UsernamePasswordSet(t *testing.T) {
...
+func TestEnvironmentCredential_UsernamePasswordSet_withCertificateChain_mock(t *testing.T) {
+	resetEnvironmentVarsForTest()
+
+	validateReq := func(req *http.Request) bool {
+		body, err := ioutil.ReadAll(req.Body)
+		if err == nil {
+			bodystr := string(body)
+			kvps := strings.Split(bodystr, "&")
+			assertion := strings.Split(kvps[0], "=")
+			token, _ := jwt.Parse(assertion[1], nil) //  func(token *jwt.Token) (interface{}, error) { return []byte(""), nil})
```
chlowell, line 191: Vestigial comment?

sdk/azidentity/azidentity_test.go
```diff
@@ -27,6 +27,92 @@ const (
 	customHostString = "https://custommock.com/"
 )
 
+func getTenantDiscoveryResponse(url string) string {
```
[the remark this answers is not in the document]
christothes: Initially I thought I would need to do that, but it wasn't required. I'll remove the param for now.

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -82,6 +87,44 @@ func TestClientCertificateCredential_GetTokenSuccess_withCertificateChain(t *tes
...
 
+func TestClientCertificateCredential_GetTokenSuccess_withCertificateChain_mock(t *testing.T) {
+
+	validateReq := func(req *http.Request) bool {
+		body, err := ioutil.ReadAll(req.Body)
+		if err == nil {
+			bodystr := string(body)
+			kvps := strings.Split(bodystr, "&")
+			assertion := strings.Split(kvps[0], "=")
+			token, _ := jwt.Parse(assertion[1], nil)
```
seankane-msft: Does this line return an error as the second returned object? If so, it's probably best to check this error here

sdk/internal/mock/mock.go
```diff
@@ -123,7 +127,20 @@ func (s *Server) Do(req *http.Request) (*http.Response, error) {
 		resp := s.getResponse()
 		return nil, resp.err
 	}
-	resp, err := s.srv.Client().Do(req)
+	var err error
+	var resp *http.Response
+	if s.routeAllRequestsToMockServer {
+		originalURL := req.URL
+		mockUrl := *req.URL
+		srvUrl, _ := url.Parse(s.srv.URL)
```
seankane-msft, line 135: Unlikely, but if there is an error here we should return the error and an empty `&http.Response{}`

sdk/azidentity/client_certificate_credential_test.go
christothes: yes, but in the end the test will fail either way, I believe.

sdk/internal/mock/mock.go
christothes: I'm not sure it is worth handling this extreme edge case.

sdk/azidentity/client_certificate_credential_test.go
seankane-msft: It will, but if there is an error on parsing, it will fail because of a nil pointer exception whereas this will help with diagnosability

sdk/internal/mock/mock.go
chlowell: Handling unlikely errors is the quintessence of Go programming 🤣 

But shouldn't we return `nil` instead of `&http.Response{}`?

sdk/azidentity/azidentity_test.go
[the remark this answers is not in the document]
chlowell: Might as well be a constant then?

sdk/internal/mock/mock.go
christothes: Fair enough 😄

sdk/azidentity/client_certificate_credential_test.go
christothes: actually, in this case we return an error even though the jwt parse is successful because we are passing nil for the validation function. I think we could check for JWT being null here instead though.

sdk/azidentity/azidentity_test.go
```diff
@@ -23,10 +28,114 @@ const (
...
+		":method": "POST",
+		":path": "/fake-tenant/oauth2/v2.0/token",
+		":scheme": "https",
+		"Accept-Encoding": "gzip",
+		"Content-Length": "2",
+		"Content-Type": "application/x-www-form-urlencoded; charset=utf-8",
+		"User-Agent": "azsdk-go-azidentity/v0.12.1 azsdk-go-azcore/v0.21.0 (go1.16.7; linux)"
+		},
+		"RequestBody": {},
+		"StatusCode": 400,`
```
chlowell, line 115: Got the next request from the recording dangling here
```suggestion
```

sdk/azidentity/azidentity_test.go
```diff
@@ -23,10 +28,114 @@ const (
...
 )
 
+func validateJWTRequestContainsHeader(t *testing.T, headerName string) mock.ResponsePredicate {
+	return func(req *http.Request) bool {
+		body, err := ioutil.ReadAll(req.Body)
+		if err == nil {
```
chlowell, line 121: Nitpicking: `err != nil` would be more idiomatic (handler immediately follows the call that returned the error; happy path less indented)

sdk/azidentity/azidentity_test.go
```diff
@@ -23,10 +28,114 @@ const (
 
 // constants for this file
 const (
-	envHostString    = "https://mock.com/"
-	customHostString = "https://custommock.com/"
+	envHostString              = "https://mock.com/"
+	customHostString           = "https://custommock.com/"
+	getTenantDiscoveryResponse = `{
```
chlowell, line 33: I'd probably put this and the validation function in `client_certificate_credential_test.go`. This file's more about high-level tests and broadly useful helpers
```suggestion
	tenantDiscoveryResponse = `{
```

chlowell (approved):
Thanks for doing this! I hope you enjoyed writing some Go :hamster:

PR #16857, opened 2022-01-19
[KeyVault] Add Immutable Property

seankane-msft:
Regenerate to the latest 7.3-preview swagger and add `KeyReleasePolicy` to the `Client.UpdateKeyProperties` method.

This change hasn't been rolled out to services, this will stay in draft until that is done.


seankane-msft:
[apiview link](https://apiview.dev/Assemblies/Review/2f0778c6eac240dc94ab53b8ae44926a?diffRevisionId=424ff7d1a79a434ea0410f4dc245f3b5&doc=False&diffOnly=False)

sdk/keyvault/azkeys/client.go
```diff
@@ -284,15 +284,35 @@ type CreateRSAKeyOptions struct {
...
+	KeyAttributes *KeyAttributes         `json:"attributes,omitempty"`
+	KeyOps        []*JSONWebKeyOperation `json:"key_ops,omitempty"`
+
+	// The policy rules under which the key can be exported.
+	ReleasePolicy *KeyReleasePolicy `json:"release_policy,omitempty"`
 }
 
 // convert CreateRSAKeyOptions to generated.KeyCreateParameters
 func (c CreateRSAKeyOptions) toKeyCreateParameters(k KeyType) generated.KeyCreateParameters {
+	var keyOps []*generated.JSONWebKeyOperation
```
JeffreyRichter, line 301: It's not a big deal in this case but you can make this code more efficient by pre-allocating the entire array:
var keyOps []*generated.JSONWebKeyOperation = make([]*generated.JSONWebKeyOperation, 0, len(c.KeyOps))

This is a great pattern to know and use whenever possible as it reduces allocations and improves performance by reducing GCs and by avoiding copying when append needs to grow the underlying array.

sdk/keyvault/azkeys/client.go
```diff
@@ -284,15 +284,35 @@ type CreateRSAKeyOptions struct {
...
+	if c.KeyOps != nil {
```
JeffreyRichter, line 302: I don't think you need this if. I think the for loop is skipped if c.KeyOps is nil

JeffreyRichter (approved):

sdk/keyvault/azkeys/client.go
seankane-msft: I don't use a `make` statement because if `c.KeyOps` is nil, we don't want to send an empty struct in the body of the request. This bypasses that. I can use `make` instead of append inside of the `if` statement

sdk/keyvault/azkeys/client.go
```diff
@@ -98,8 +98,11 @@ func (c *CreateKeyOptions) toKeyCreateParameters(keyType KeyType) generated.KeyC
 	}
 
 	var ops []*generated.JSONWebKeyOperation
-	for _, o := range c.KeyOps {
-		ops = append(ops, (*generated.JSONWebKeyOperation)(o))
+	if c.KeyOps != nil {
+		ops = make([]*generated.JSONWebKeyOperation, len(c.KeyOps))
```
heaths, line 102: The name the other SDKs use is just `KeyOperations`.

sdk/keyvault/azkeys/client.go
```diff
@@ -284,15 +287,36 @@ type CreateRSAKeyOptions struct {
 
 	// Application specific metadata in the form of key-value pairs.
 	Tags map[string]string `json:"tags,omitempty"`
+
+	// Elliptic curve name. For valid values, see JsonWebKeyCurveName.
+	Curve *JSONWebKeyCurveName `json:"crv,omitempty"`
```
heaths, line 292: Just `KeyCurveName` is what the other SDKs use.

sdk/keyvault/azkeys/client.go
```diff
@@ -284,15 +287,36 @@ type CreateRSAKeyOptions struct {
...
 // convert CreateRSAKeyOptions to generated.KeyCreateParameters
 func (c CreateRSAKeyOptions) toKeyCreateParameters(k KeyType) generated.KeyCreateParameters {
+	var keyOps []*generated.JSONWebKeyOperation
+	if c.KeyOps != nil {
+		keyOps = make([]*generated.JSONWebKeyOperation, len(c.KeyOps))
+		for i, k := range c.KeyOps {
+			keyOps[i] = (*generated.JSONWebKeyOperation)(k)
+		}
+	}
 	return generated.KeyCreateParameters{
```
heaths, line 311: Given the method is private, can I assume `generated.KeyCreateParameters` is not exposed? We generally follow `<MethodName>Options` for public inputs.

sdk/keyvault/azkeys/models.go
```diff
@@ -32,18 +32,38 @@ type Attributes struct {
 
 // KeyAttributes - The attributes of a key managed by the key vault service.
 type KeyAttributes struct {
-	Attributes
-	// READ-ONLY; softDelete data retention days.
+	// Determines whether the object is enabled.
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Expiry date in UTC.
+	Expires *time.Time `json:"exp,omitempty"`
```
heaths, line 39: Other languages use `ExpiresOn`.

sdk/keyvault/azkeys/models.go
```diff
@@ -32,18 +32,38 @@ type Attributes struct {
...
+
+	// Indicates if the private key can be exported.
+	Exportable *bool `json:"exportable,omitempty"`
+
+	// Not before date in UTC.
+	NotBefore *time.Time `json:"nbf,omitempty"`
+
+	// READ-ONLY; Creation time in UTC.
+	Created *time.Time `json:"created,omitempty" azure:"ro"`
```
heaths, line 48: `CreatedOn`

sdk/keyvault/azkeys/models.go
```diff
@@ -32,18 +32,38 @@ type Attributes struct {
...
 
-	// READ-ONLY; Reflects the deletion recovery level currently in effect for keys in the current vault. If it contains 'Purgeable' the key can be permanently
-	// deleted by a privileged user; otherwise, only the system
+	// READ-ONLY; Reflects the deletion recovery level currently in effect for keys in the current vault. If it contains 'Purgeable'
+	// the key can be permanently deleted by a privileged user; otherwise, only the system
 	// can purge the key, at the end of the retention interval.
 	RecoveryLevel *DeletionRecoveryLevel `json:"recoveryLevel,omitempty" azure:"ro"`
+
+	// READ-ONLY; Last updated time in UTC.
+	Updated *time.Time `json:"updated,omitempty" azure:"ro"`
```
heaths, line 59: `UpdatedOn`

sdk/keyvault/azkeys/models.go
```diff
@@ -306,16 +325,33 @@ type KeyReleasePolicy struct {
 	ContentType *string `json:"contentType,omitempty"`
 
 	// Blob encoding the policy rules under which the key can be released.
-	Data []byte `json:"data,omitempty"`
+	EncodedPolicy []byte `json:"data,omitempty"`
```
heaths, line 328: This needs to be base64url-encoded, as does any other `[]byte` in the API (that we encoded, anyway; PEM input will still be base64-encoded but we don't mutate that). The REST model is therefore a string.

heaths (changes requested):

sdk/keyvault/azkeys/models.go
jhendrixMSFT: This does get base64-URL encoded before sending over the wire, see [here](https://github.com/Azure/azure-sdk-for-go/blob/0f8d8fdf242f8905f8396f34ba6cdc1d9b182d94/sdk/keyvault/azkeys/internal/generated/models.go#L895).

sdk/keyvault/azkeys/client.go
seankane-msft: Your assumption is correct, anything in the `generated` subpackage is not exposed. This method has a `CreateRSAKeyOptions` struct.

sdk/keyvault/azkeys/client.go
seankane-msft: Updated

sdk/keyvault/azkeys/client.go
seankane-msft: Updated

sdk/keyvault/azkeys/models.go
seankane-msft: Updated

sdk/keyvault/azkeys/models.go
seankane-msft: Updated

sdk/keyvault/azkeys/models.go
seankane-msft: Updated

sdk/keyvault/azkeys/CHANGELOG.md
```diff
@@ -3,12 +3,15 @@
...
-* Adds the `ReleasePolicy` parameter to the `UpdateKeyPropertiesOptions` struct
+* Adds the `ReleasePolicy` parameter to the `UpdateKeyPropertiesOptions` struct.
 * Adds the `Immutable` boolean to the `KeyReleasePolicy` model.
 
 ### Breaking Changes
 * Changed the `Data` to `EncodedPolicy` on the `KeyReleasePolicy` struct.
-* Changed the `Tags` properties from `map[string]*string` to `map[string]string`
+* Changed the `Tags` properties from `map[string]*string` to `map[string]string`.
+* Changed the `Updated`, `Created`, and `Expires` properties to `UpdatedOn`, `CreatedOn`, and `ExpiresOn`.
+* Renamed `JSONWebKeyOperation` to `KeyOperations`.
```
heaths, line 13: Just singular i.e., `KeyOperation`: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Keys/src/KeyOperation.cs

heaths (approved):
Signing off to unblock, but please be sure to use `KeyOperation` instead of `KeyOperations` (should be singular).

sdk/keyvault/azkeys/client.go
```diff
@@ -73,7 +73,7 @@ type CreateKeyOptions struct {
 
 	// The attributes of a key managed by the key vault service.
 	KeyAttributes *KeyAttributes   `json:"attributes,omitempty"`
-	KeyOps        []*KeyOperations `json:"key_ops,omitempty"`
+	KeyOps        []*KeyOperation `json:"key_ops,omitempty"`
```
heaths, line 76: The only place we exposed "KeyOps" was in the `JSONWebKey` class (struct). Everywhere else it was `KeyOperations`.

heaths: In hindsight this was a mistake, IMO, but we already shipped like that. I think the original idea is for the JWK to more closely match its underlying JSON structure.

heaths (approved):
One slight (unfortunate) change, but otherwise LGTM.

sdk/keyvault/azkeys/client.go
seankane-msft: I'm slightly confused by this statement, are you recommending anywhere that exposes a `[]*KeyOperations` the field should be called `KeyOperations`, and in the `JSONWebKey` struct it should be called `KeyOps` ?

seankane-msft: Checked out the apiview for .NET, understood now. Changes have been made.

PR #16859, opened 2022-01-19
[KeyVault] Fix issue with PATCH method

seankane-msft:
I noticed this issue while fixing #16858 , on PATCH methods, properties would be erased if they are not explicitly set. This fixes that by setting generated values as `nil` if they are not explicitly specified.

### DIFF
--- a/sdk/keyvault/azsecrets/CHANGELOG.md
+++ b/sdk/keyvault/azsecrets/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Release History
 
-## 0.4.1 (Unreleased)
+## 0.5.0 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixes a bug where `UpdateSecretProperties` will delete properties that are not explicitly set each time.
 
 ### Other Changes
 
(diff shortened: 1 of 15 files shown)


sdk/keyvault/azsecrets/CHANGELOG.md
```diff
@@ -1,12 +1,13 @@
 # Release History
 
-## 0.4.1 (Unreleased)
+## 0.5.0 (Unreleased)
```
jhendrixMSFT, line 3: For preview, we've been trying to follow the pattern that breaking changes bump the minor version.  This doesn't look to be breaking, at least for compile time.  Is this being considered breaking for runtime?

seankane-msft: In runtime I think so, the request body will change with this code change.

jhendrixMSFT: I see.  Then should the comment be moved to the `Breaking Changes` category?  It almost makes me think that we could benefit from a `Runtime Changes` category to differentiate between compile-time and runtime breaking changes (maybe we just note this is runtime only in the comment).

seankane-msft: Agree that it should be under breaking changes. I don't think we'll have a lot of changes that fall under a "Breaking Change but only at runtime" and for now should just note it in the changelog entry.

jhendrixMSFT (approved):

PR #16921, opened 2022-01-26
reverting to proxy tool

seankane-msft:
Reverting to proxy tool instead of test proxy docker image


sdk/azidentity/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.13.1 (Unreleased)
 
 ### Features Added
+* Force pipelines
```
chlowell, line 6: Can you expand on this? (I have no idea what it means)

seankane-msft: I'm just touching azidentity to force the identity pipelines to run. I want to make sure I don't incidentally break your pipelines

chlowell: Ah, I see. You can also trigger arbitrary pipelines by commenting e.g. "/azp run go - azidentity"

sdk/azidentity/live_test.go
```diff
@@ -89,21 +88,10 @@ func init() {
...
 		}
-		req.Header["x-abstraction-identifier"] = []string{"BodilessMatcher"}
-		res, err := http.DefaultClient.Do(req)
-		if err != nil {
-			panic(err)
-		}
-		if res.StatusCode != http.StatusOK {
-			log.Panicf("failed to enable BodilessMatcher: %v", res)
-		}
-		// TODO: reset matcher
```
chlowell, line 106: Is this unnecessary now? My concern is that playing back azidentity tests leaves body matching disabled. If I then run, say, Key Vault tests, and don't restart the proxy first, body matching is still disabled, and tests that should fail because of the bug I'm adding may pass instead.

seankane-msft: In KV I have a `ResetProxy` at the beginning of `TestMain` that way I'm always running from a fresh proxy. Do you want me to add that to identity as well? It's already there on RecordingMode, adding to PlaybackMode makes sense as well.

chlowell: Well, it depends on what all `ResetProxy` does. I want the proxy to be in its default state before and after azidentity tests. I don't want azidentity tests to disable the default OAuth response sanitizer; that creates a risk of recording secrets for another package.

sdk/azidentity/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
 go 1.16
 
+replace github.com/Azure/azure-sdk-for-go/sdk/internal => ../internal
```
scbedd, line 5: Does this stick around until this merges to main?

scbedd (approved):

sdk/azidentity/go.mod
seankane-msft: Yes, I'll have to do a release of internal first then they'll all be removed before the next release cycle.

sdk/azidentity/live_test.go
seankane-msft: `ResetProxy` returns the proxy to it's default state, I believe this is what you want.

chlowell: In that case I think this function should call `ResetProxy` before and after every recording or playback run.

seankane-msft: Yes, that's why I have at the beginning of [`TestMain`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/keyvault/azkeys/utils_test.go#L35-L38) in the key vault test libraries.

seankane-msft: I added the reset after test runs as well.

chlowell: I'm confused. It looks like you're enabling the bodiless matcher after the run? I want to restore the proxy's default settings i.e. to enable the default matcher and any default sanitizers. I want the proxy in its default state at the start and end of `TestMain`, when not in live mode. I don't want to do anything with the proxy in live mode.

Also, looking at the code, I see that past me believed `ResetProxy` removes default sanitizers--if the OAuth sanitizer is a default, that must be correct because otherwise these recordings wouldn't work. Is there an easy way to restore proxy defaults?

chlowell (changes requested):

sdk/azidentity/live_test.go
seankane-msft: If you call `ResetProxy`, the OAuth sanitizer will be on. Confirmed by @scbedd . ResetProxy restores to defaults.

scbedd: Yep! Hitting Admin/Reset gets directed to [here](https://github.com/Azure/azure-sdk-tools/blob/55f1b2f5eb63e1bd373be66e685c6fe9d61e0698/tools/test-proxy/Azure.Sdk.Tools.TestProxy/RecordingHandler.cs#L437)

The default constructor (which is called for our singleton instance of RecordingHandler at server startup) calls SetDefaultExtensions(). 

But...is oauth sanitizer a default?

chlowell: > If you call `ResetProxy`, the OAuth sanitizer will be on. Confirmed by @scbedd . ResetProxy restores to defaults.

Great. About the block after `m.Run()` then--why enable the bodiless matcher there?

> But...is oauth sanitizer a default?

I hope so. Otherwise, the proxy would record authentication secrets by default.

seankane-msft: After the m.Run should be a Reset, I think that was a copy-pasta error

chlowell (approved):

PR #16925, opened 2022-01-26
[azservicebus] Handle 500 as a retryable code (no recovery needed)

richardpark-msft:
Handle 500 as a retryable code (no link/connection recovery needed)

This can happen if, for instance, your link gets detached in the middle of an operation. Typically you'll get the 500, and then subsequent retries will work or reveal the underlying cause (ie, link detached, etc..).

### DIFF
--- a/sdk/messaging/azservicebus/internal/errors.go
+++ b/sdk/messaging/azservicebus/internal/errors.go
@@ -175,7 +175,10 @@ func GetRecoveryKind(err error) recoveryKind {
 		}
 
 		// simple timeouts
-		if me.Resp.Code == 408 || me.Resp.Code == 503 {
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (approved):

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -175,7 +175,13 @@ func GetRecoveryKind(err error) recoveryKind {
 		}
 
 		// simple timeouts
```
RickWinter, line 177: Should we have verbose logging for the 500?

richardpark-msft: It happens naturally as part of current logging with the Retry() code, along with any other error.

PR #16928, opened 2022-01-27
[azservicebus] Use the DefaultAzureCredential.

richardpark-msft:
Make it so we don't need to create and destroy SPs for local dev box testing _and_ it'll work just fine in CI since the environment credential is included.


jhendrixMSFT (approved):

RickWinter (commented):
Should we use ChainedTokenCredential to be explicit in the credentials we look for?

richardpark-msft:
just rebasing now. Looks like there were two issues, both of them just CI config related:
- Bicep template wasn't properly formatting the SERVICEBUS_ENDPOINT variable. I'll file a bug that we should catch this kind of obvious incorrect formatting (or simply strip it out ourselves).
- I didn't have the "environment var renaming" YAML in my ci.yml.

PR #16932, opened 2022-01-27
Avoid retries on terminal failures in challenge policy

jhendrixMSFT:
If the challenge request fails, return the response to the caller as it
won't contain an authentication header.
Make hard errors non retriable.

### DIFF
--- a/sdk/keyvault/internal/challenge_policy.go
+++ b/sdk/keyvault/internal/challenge_policy.go
@@ -19,6 +19,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
 const headerAuthorization = "Authorization"
@@ -57,6 +58,10 @@ func (k *KeyVaultChallengePolicy) Do(req *policy.Request) (*http.Response, error
 			return nil, err
 		}
 
+		if resp.StatusCode > 299 {
+			// the request failed, don't try any further
+			return resp, nil
+		}
 		err = k.findScopeAndTenant(resp)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


sdk/keyvault/internal/challenge_policy.go
```diff
@@ -57,6 +58,10 @@ func (k *KeyVaultChallengePolicy) Do(req *policy.Request) (*http.Response, error
 			return nil, err
 		}
 
+		if resp.StatusCode > 299 {
```
jhendrixMSFT, line 61: Does this look reasonable to you?

seankane-msft: I think this should go after the `err = k.findScopeAndTenant(resp)` line. The first response should be a 404 with the authorization in the `WWW-Authenticate` header.

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -57,6 +58,10 @@ func (k *KeyVaultChallengePolicy) Do(req *policy.Request) (*http.Response, error
...
+			// the request failed, don't try any further
+			return resp, nil
+		}
 		err = k.findScopeAndTenant(resp)
 		if err != nil {
 			return nil, err
```
seankane-msft, line 67:
```
if err != nil || resp.StatusCode > 299 {
    return nil, err
}
``` 
Something similar to this

sdk/keyvault/internal/challenge_policy.go
jhendrixMSFT: The problem here is that there's no `WWW-Authenticate` header in a failure response.  I noticed then when running a test against a key vault that had been deleted.  We kept retrying due to the missing header value.

seankane-msft: Can we extend this conditional to say `if resp.StatusCode > 299 && _, ok := resp.Header.Get("WWW-Authenticate"); !ok {...}` ? We don't want to return the error all the time, only when we have no info from the service and the status code is > 299

jhendrixMSFT: Not sure I follow.  Also, I think the condition should probably be > 399

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -57,6 +58,10 @@ func (k *KeyVaultChallengePolicy) Do(req *policy.Request) (*http.Response, error
...
 		}
 
+		if resp.StatusCode > 399 {
+			// the request failed, don't try any further
+			return resp, nil
+		}
```
seankane-msft, line 64: This `resp` is nearly always going to fail, because it's a challenge auth (will return a 401). This should drop down until after the `err = k.findScopeAndTenant(resp)`.

jhendrixMSFT: Fixed by skipping the 401.

seankane-msft (approved):

PR #16933, opened 2022-01-27
[KeyVault] Removing embedded models in azkeys

seankane-msft:
- [ ] Remove embedded models in keyvault libraries to be inline with current autorest code.
- [ ] Name changes to align with other Tier 1 libraries
- [ ] Remove unnecessary generated code.


sdk/keyvault/azkeys/models.go
```diff
@@ -77,12 +59,12 @@ func (k *KeyAttributes) toGenerated() *generated.KeyAttributes {
 }
 
 // converts *generated.KeyAttributes to *KeyAttributes
-func keyAttributesFromGenerated(i *generated.KeyAttributes) *KeyAttributes {
+func keyAttributesFromGenerated(i *generated.KeyAttributes) *KeyProperties {
```
chlowell, line 62:
```suggestion
func keyPropertiesFromGenerated(i *generated.KeyAttributes) *KeyProperties {
```

sdk/keyvault/azkeys/models.go
```diff
@@ -94,10 +76,10 @@ func keyAttributesFromGenerated(i *generated.KeyAttributes) *KeyAttributes {
 	}
 }
 
-// KeyBundle - A KeyBundle consisting of a WebKey plus its attributes.
-type KeyBundle struct {
-	// The key management attributes.
-	Attributes *KeyAttributes `json:"attributes,omitempty"`
+// KeyVaultKey - A KeyVaultKey consists of a WebKey plus its attributes.
+type KeyVaultKey struct {
```
chlowell, line 80: The motivation for this name in other SDKs was to prevent confusion with `Key` types from other namespaces. In a Go application the type will be qualified, so I think there's a case for `azkeys.Key` here.

seankane-msft: @heaths thoughts?

sdk/keyvault/azsecrets/models.go
```diff
@@ -14,13 +14,35 @@ import (
 
 // DeletedSecretBundle - A Deleted Secret consisting of its previous id, attributes and its tags, as well as information on when it will be purged.
 type DeletedSecretBundle struct {
```
mccoyp, line 16: Is there a reason for `DeletedKeyBundle` being renamed to `DeletedKey` but not dropping "Bundle" here?

seankane-msft: Haven't gotten to secrets yet. I'm going to undo the changes to secrets and do it in another PR.

sdk/keyvault/azkeys/models.go
heaths: @chlowell is right, but I don't see a strong reason to deviate from naming without due cause (e.g. some name happens to be a keyword in a language), but perhaps @JeffreyRichter has a strong opinion on this either way.

sdk/keyvault/azkeys/client.go
```diff
@@ -1375,7 +1294,7 @@ func getKeyRotationPolicyResponseFromGenerated(i generated.KeyVaultClientGetKeyR
 		attribs = &KeyRotationPolicyAttributes{
 			ExpiryTime: i.Attributes.ExpiryTime,
 			Created:    i.Attributes.Created,
```
heaths, line 1296: Should be `CreatedOn`.

sdk/keyvault/azkeys/client.go
```diff
@@ -1508,7 +1427,7 @@ func updateKeyRotationPolicyResponseFromGenerated(i generated.KeyVaultClientUpda
...
```
heaths, line 1429: Same here (I won't mention it again for brevity, so please check for all occurrences).

sdk/keyvault/azkeys/client_test.go
```diff
@@ -581,9 +579,8 @@ func TestImportKey(t *testing.T) {
 			client, err := createClient(t, testType)
 			require.NoError(t, err)
 
-			r := RSA
 			jwk := JSONWebKey{
-				KeyType: &r,
+				KeyType: KeyTypeRSA.ToPtr(),
```
heaths, line 583: Interesting that this is a pointer. It is a required field of a JWK, or are there cases - either in the swagger or in the SDK - where it could be `nil`?

sdk/keyvault/azkeys/models.go
```diff
@@ -259,35 +245,62 @@ func keyItemFromGenerated(i *generated.KeyItem) *KeyItem {
...
 	RecoveryID *string `json:"recoveryId,omitempty"`
 
+	// The policy rules under which the key can be exported.
+	ReleasePolicy *KeyReleasePolicy `json:"release_policy,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]string `json:"tags,omitempty"`
+
 	// READ-ONLY; The time when the key was deleted, in UTC
 	DeletedDate *time.Time `json:"deletedDate,omitempty" azure:"ro"`
```
heaths, line 273: Should be `DeletedOn`.

sdk/keyvault/azkeys/models.go
```diff
@@ -259,35 +245,62 @@ func keyItemFromGenerated(i *generated.KeyItem) *KeyItem {
...
+
 	// READ-ONLY; The time when the key is scheduled to be purged, in UTC
 	ScheduledPurgeDate *time.Time `json:"scheduledPurgeDate,omitempty" azure:"ro"`
 }
 
 // DeletedKeyItem - The deleted key item containing the deleted key metadata and information about deletion.
 type DeletedKeyItem struct {
-	KeyItem
+	// The key management attributes.
+	Attributes *KeyProperties `json:"attributes,omitempty"`
```
heaths, line 286: 1. Should be `Properties`, but...
2. Why is `DeletedKeyItem` public? We don't expose this in other language SDKs, and not sure how it differs from a `DeletedKey` anyway.

sdk/keyvault/azkeys/models.go
```diff
@@ -259,35 +245,62 @@ func keyItemFromGenerated(i *generated.KeyItem) *KeyItem {
...
+
+	// Key identifier.
+	KID *string `json:"kid,omitempty"`
```
heaths, line 289: If this type is to be exposed, this should be `KeyID`.

sdk/keyvault/azkeys/models.go
```diff
@@ -259,35 +245,62 @@ func keyItemFromGenerated(i *generated.KeyItem) *KeyItem {
...
+
 	// The url of the recovery object, used to identify and recover the deleted key.
 	RecoveryID *string `json:"recoveryId,omitempty"`
 
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]string `json:"tags,omitempty"`
+
 	// READ-ONLY; The time when the key was deleted, in UTC
 	DeletedDate *time.Time `json:"deletedDate,omitempty" azure:"ro"`
```
heaths, line 298: If this type is to be exposed, this should be `DeletedOn`.

heaths (changes requested):

sdk/keyvault/azkeys/models.go
JeffreyRichter: I don't love this name. Could it be VaultKey in the azkeys package?

seankane-msft: I think `Key` is better than `VaultKey`

sdk/keyvault/azkeys/client_test.go
seankane-msft: This is in response types which are empty when an error occurs. We want to have it as nil for this scenario where there's a difference between "not there because the operation failed" and "zero value"

sdk/keyvault/azkeys/client.go
```diff
@@ -72,7 +72,7 @@ type CreateKeyOptions struct {
 	Curve *KeyCurveName `json:"crv,omitempty"`
 
 	// The attributes of a key managed by the key vault service.
-	KeyAttributes *KeyAttributes  `json:"attributes,omitempty"`
+	KeyProperties *KeyProperties  `json:"attributes,omitempty"`
```
JeffreyRichter, line 75: Do we need to repeat Key everywhere? Can't the field names be just Properties and Operations?
Where else can we apply this shortening?

JeffreyRichter (approved):

sdk/keyvault/azkeys/client.go
heaths: These were consistent with every other language, not to mention fairly consistent with the underlying data types common for a JWT like "kid" for "KeyID", "key_ops" for "KeyOperations", etc. I understand that language idiomaticity is primary, but is there any great reason to differ from other languages?

JeffreyRichter: Go tries very hard not to repeat names which they call stuttering. So, repeating "Key" falls under this category. You can get a sense of how much there is written about by doing this [search](https://www.google.com/search?q=golang+stuttering). So, for Go, we want to reduce this as much as possible as it is what the community expects.

jhendrixMSFT: To rant just a little, the verbosity of .NET bleeds into everything.  This is a prime example.  As a Go developer, it's somewhat frustrating that we get saddled with patterns from tier 1 languages that clash with our idioms.

sdk/keyvault/azkeys/client.go
```diff
@@ -47,7 +47,7 @@ func (c *ClientOptions) toConnectionOptions() *policy.ClientOptions {
 }
 
 // NewClient returns a pointer to a Client object affinitized to a vaultUrl.
-func NewClient(vaultUrl string, credential azcore.TokenCredential, options *ClientOptions) (*Client, error) {
+func NewClient(vaultUrl string, credential azcore.TokenCredential, options *ClientOptions) (Client, error) {
```
heaths, line 50: IIRC, wasn't the guideline to always return a client pointer?

heaths: https://azure.github.io/azure-sdk/golang_introduction.html#golang-client-constructors

heaths (approved):

heaths (approved):
Does azsecrets need to be updated as well?

sdk/keyvault/azkeys/client.go
```diff
@@ -197,10 +197,10 @@ func createECKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyRespons
 // EC keys can be created in Standard or Premium SKU vaults, ECHSM can be created in Premium SKU vaults or Managed HSMs.
 // It requires the keys/create permission. Pass nil to use the default options.
 func (c *Client) CreateECKey(ctx context.Context, name string, options *CreateECKeyOptions) (CreateECKeyResponse, error) {
-	keyType := EC
+	keyType := KeyTypeEC
 
-	if options != nil && options.HardwareProtected {
-		keyType = ECHSM
+	if options != nil && options.HardwareProtected != nil && *options.HardwareProtected {
```
chlowell, line 202: You could simplify this a little and avoid double checking `options == nil`:
```go
if options == nil {
	options = &CreateECKeyOptions{}
}
if options.HardwareProtected != nil && *options.HardwareProtected {
	keyType = KeyTypeECHSM
}
```

sdk/keyvault/azkeys/client.go
```diff
@@ -290,40 +290,40 @@ type CreateRSAKeyOptions struct {
 	Tags map[string]string `json:"tags,omitempty"`
 
 	// Elliptic curve name. For valid values, see KeyCurveName.
-	Curve *KeyCurveName `json:"crv,omitempty"`
+	Curve *CurveName `json:"crv,omitempty"`
```
chlowell, line 293:
```suggestion
```

sdk/keyvault/azkeys/client.go
```diff
@@ -290,7 +290,6 @@ type CreateRSAKeyOptions struct {
 	Tags map[string]string `json:"tags,omitempty"`
 
 	// Elliptic curve name. For valid values, see KeyCurveName.
-	Curve *CurveName `json:"crv,omitempty"`
```
heaths: Probably should delete the comment and resulting blank lines as well.

sdk/keyvault/azkeys/constants.go
```diff
@@ -133,16 +169,77 @@ func (c ActionType) ToPtr() *ActionType {
...
+}
+
+// EncryptionAlgorithm - The encryption algorithm to use to protected the exported key material
+type EncryptionAlgorithm string
 
 const (
-	KeyEncryptionAlgorithmCKMRSAAESKEYWRAP KeyEncryptionAlgorithm = "CKM_RSA_AES_KEY_WRAP"
-	KeyEncryptionAlgorithmRSAAESKEYWRAP256 KeyEncryptionAlgorithm = "RSA_AES_KEY_WRAP_256"
-	KeyEncryptionAlgorithmRSAAESKEYWRAP384 KeyEncryptionAlgorithm = "RSA_AES_KEY_WRAP_384"
+	EncryptionAlgorithmCKMRSAAESKEYWRAP EncryptionAlgorithm = "CKM_RSA_AES_KEY_WRAP"
```
chlowell, line 184: Yikes, the acronyms make this unreadable but I suppose doing anything to break them up is unconventional 🙄 

Should this be `KeyWrapAlgorithmCKMRSAAES`?

sdk/keyvault/azkeys/constants.go
```diff
@@ -78,46 +78,82 @@ func recoveryLevelToGenerated(d *DeletionRecoveryLevel) *generated.DeletionRecov
 	}
 }
 
-// KeyCurveName - Elliptic curve name. For valid values, see KeyCurveName.
-type KeyCurveName string
+// PossibleDeletionRecoveryLevelValues provides a slice of all possible DeletionRecoveryLevels
+func PossibleDeletionRecoveryLevelValues() []DeletionRecoveryLevel {
```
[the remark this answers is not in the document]
jhendrixMSFT: We added this in track 1 at the request of TF.  I believe they used in it cases where a const value could come from external (customer?) input, and they needed a way to validate that the provided value is valid.

sdk/keyvault/azkeys/constants.go
seankane-msft: Agree, they are ugly but I can't think of a better way to handle them.

sdk/keyvault/azkeys/constants.go
```diff
@@ -78,46 +78,82 @@ func recoveryLevelToGenerated(d *DeletionRecoveryLevel) *generated.DeletionRecov
...
+		DeletionRecoveryLevelCustomizedRecoverableProtectedSubscription,
+		DeletionRecoveryLevelCustomizedRecoverablePurgeable,
+		DeletionRecoveryLevelPurgeable,
+		DeletionRecoveryLevelRecoverable,
+		DeletionRecoveryLevelRecoverableProtectedSubscription,
+		DeletionRecoveryLevelRecoverablePurgeable,
+	}
+}
+
+// CurveName - Elliptic curve name. For valid values, see CurveName.
```
chlowell, line 94:
```suggestion
// CurveName - Elliptic curve name. For valid values, see PossibleCurveNameValues.
```

sdk/keyvault/azkeys/constants.go
chlowell: About the prefix then--will Key Vault allow these algorithms in `encrypt`?

seankane-msft: Should I remove the `KEYWRAP` portion from all the models in this scenario?

chlowell: Although "key wrap" is part of the algorithm name I think a `KeyWrap` prefix implies it strongly enough so, `KeyWrapAlgorithmRSAAES256` > `KeyWrapAlgorithmRSAAESKEYWRAP256`. @heaths, what do you think?

heaths: I don't think "key wrap" should be removed from the algorithm name. They are long, but many, many constants in crypto are. Also, I don't think we should preface the constants with "KeyWrap" when used for the `Encrypt` function. The algorithms that `Encrypt` and `WrapKey` support are different.

chlowell: > The algorithms that `Encrypt` and `WrapKey` support are different.

That's what I was getting at originally. These constants are for `WrapKey`, so we should have something like
```suggestion
	KeyWrapAlgorithmCKMRSAAESKEYWRAP KeyWrapAlgorithm = "CKM_RSA_AES_KEY_WRAP"
```
But taking a closer look, I see we have another bunch of EncryptionAlgorithms and WrapAlgorithms in `crypto/constants.go`. What are these for?

seankane-msft: * `WrapAlgorithm` is used in the `crypto.Client.WrapKey/UnwrapKey`
* `EncryptionAlgorithm` is used in `crypto.Client.Encrypt/Decrypt`

chlowell: Ah, these constants I'm commenting on are for `ReleaseKey` only? .NET calls the type [KeyExportEncryptionAlgorithm](https://github.com/Azure/azure-sdk-for-net/blob/3b295806bf2bf3af6ec2380173363db131f931ca/sdk/keyvault/Azure.Security.KeyVault.Keys/src/ReleaseKeyOptions.cs#L33)

heaths: In general, I would keep separate sets of constants for different methods (or sets of related methods) / models based on the swagger. That they are in different models means they can - and most likely will - diverge and that would make matters even more confusing for developers.

seankane-msft: Thanks for the catch @chlowell , I've updated the name

heaths (approved):

chlowell (approved):

PR #16937, opened 2022-01-28
Fix deadlock when erroring writes are slow

MasslessParticle:
`c.getErr` can return before slow writes complete. If there is more than one write and those writes result in errors, the second write blocks on a full `errCh`. The result is a deadlocked `copyFromReader` call because the wait groups never clear.

The test included tests demonstrates the deadlock. 

This PR causes `close` to simultaneously drain the `errCh` while it waits for writes to complete. It combines any errors that are returned.

I looked at the CHANGELOG, but not sure this change fits. Let me know if I need to do something else.

### DIFF
--- a/sdk/storage/azblob/blockblob/chunkwriting.go
+++ b/sdk/storage/azblob/blockblob/chunkwriting.go
@@ -13,12 +13,12 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared"
 	"io"
 	"sync"
 	"sync/atomic"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/uuid"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared"
 )
 
 // blockWriter provides methods to upload blocks that represent a file to a server and commit them.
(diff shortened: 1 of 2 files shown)


sdk/storage/azblob/zt_chunkwriting_test.go
```diff
@@ -195,6 +201,46 @@ func (s *azblobUnrecordedTestSuite) TestGetErr() {
...
+	}
+	defer from.Close()
+
+	br := newFakeBlockWriter()
+	defer br.cleanup()
+
+	br.stageDelay = 200 * time.Millisecond
+	br.errOnBlock = 0
+
+	errs := make(chan error, 1)
```
MasslessParticle, line 226: The goroutine here is a bit weird, but I wanted to listen for the success condition in a `select` with a context so any potential regressions cause a useful error rather than just hanging.

sdk/storage/azblob/zt_chunkwriting_test.go
```diff
@@ -302,6 +348,7 @@ func (s *azblobUnrecordedTestSuite) TestCopyFromReader() {
 		if err != nil {
 			panic(err)
 		}
+		defer from.Close()
```
MasslessParticle, line 351: This is unrelated to the change, but I noticed this wasn't being cleaned up.

mohsha-msft:
Hey @MasslessParticle ,

Thanks a lot for your contribution to our repo.
I'll review this PR and include this in release 0.4.0 in March. But in the meantime, I'll merge the changes in an internal `azblob` dev branch so if you need it, you can pull in that specific branch to your code. Does that sound good to you?

MasslessParticle:
Thanks for your reply! 

No need to merge to a dev branch. I'm excited for the `0.4.0` release

MasslessParticle:
@mohsha-msft is there still interest in accepting this PR?

MasslessParticle:
@jhendrixMSFT @mohsha-msft Is there still interest in this PR?

jhendrixMSFT (approved):

PR #16942, opened 2022-01-29
Add cloud configuration API to azcore

### LINKED ISSUE #16677: Cannot set custom scopes when creating a pipeline
In the case of ARM, we calculate the scopes based on the provided endpoint.  While this works in most cases, when a service is in private preview it might not.  Here's a snippet from the customer.

> We are stuck on trying to authenticate with ARM. The issue seems to be - our scope needs to be "management.azure.com", while the ARMEndpoint needs to be centraluseuap.management.azure.com. The reason is our bits are deployed only centralesuap and this is still private preview.

For data-plane, scopes look to be hard-coded.

chlowell:
## usage
```go
import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"

// the cloud package has predefined configuration for public clouds
conf := cloud.AzureChina

// users can create configuration for private clouds
conf = cloud.Configuration{LoginEndpoint: "..."}

// Clients default to Azure Public Cloud. An azcore.ClientOptions field overrides the default.
clientOpts := azcore.ClientOptions{Cloud: conf}

// client option types already embed azcore.ClientOptions, creating a consistent API for cloud configuration:
cred, err := azidentity.NewDefaultAzureCredential(
    &azidentity.DefaultAzureCredentialOptions{ClientOptions: clientOpts}
)
handle(err)

client, err := armsubscription.NewClient(
    cred, &arm.ClientOptions{ClientOptions: clientOpts}
)
handle(err)
```

Closes #16677


JeffreyRichter (approved):

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -7,36 +7,52 @@
...
-	ep := options.Endpoint
-	if len(ep) == 0 {
-		ep = arm.AzurePublicCloud
+	conf, err := getConfiguration(options)
+	if err != nil {
+		return pipeline.Pipeline{}, err
 	}
 	authPolicy := NewBearerTokenPolicy(cred, &armpolicy.BearerTokenOptions{
-		Scopes:           []string{shared.EndpointToScope(string(ep))},
+		Scopes:           []string{shared.EndpointToScope(conf.Endpoint)},
```
jhendrixMSFT, line 32: With the introduction of `ServiceConfiguration.Audiences` can we get rid of the `EndpointToScope()`?

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -7,36 +7,52 @@
...
-	return azruntime.NewPipeline(module, version, plOpts, &options.ClientOptions)
+	return azruntime.NewPipeline(module, version, plOpts, &options.ClientOptions), nil
+}
+
+func getConfiguration(o *arm.ClientOptions) (cloud.ServiceConfiguration, error) {
+	c := cloud.WellKnownClouds[cloud.AzurePublicCloud]
+	if !reflect.ValueOf(o.Cloud).IsZero() {
+		c = o.Cloud
+	}
+	if conf, ok := c.Services[cloud.ResourceManager]; ok && conf.Endpoint != "" && len(conf.Audiences) > 0 {
```
jhendrixMSFT, line 53: Do we need any similar validation when creating a non-ARM pipeline?

sdk/azcore/cloud/cloud.go
```diff
@@ -0,0 +1,180 @@
...
+			WellKnownClouds[AzurePublicCloud] = cloud
+		case "AzureUSGovernment":
+			WellKnownClouds[AzureGovernment] = cloud
+		}
+	}
+}
+
+type ServiceName string
+
+const ResourceManager ServiceName = "resourceManager"
```
jhendrixMSFT, line 38: This, and the below types and their fields, should have doc comments.

jhendrixMSFT (changes requested):
Missing doc comments.

sdk/azcore/arm/runtime/pipeline.go
chlowell: Not in `NewPipeline()`, but each client/credential reading settings from this API should perform similar validation.

sdk/azcore/arm/runtime/pipeline.go
chlowell: Yes! Commit forthcoming.

sdk/azcore/arm/runtime/pipeline.go
chlowell: Correct. No service will use all fields, e.g. none will use both Endpoint and Suffix, so I think the best we could do in general is ensure configuration isn't zero-valued, which doesn't mean it's complete. (Speaking of Suffix, I'll remove it from this PR; we don't have a client that would use it and can add it when we do.)

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -7,36 +7,55 @@
...
+}
+
+func getConfiguration(o *azpolicy.ClientOptions) (cloud.ServiceConfiguration, error) {
+	c := cloud.WellKnownClouds[cloud.AzurePublicCloud]
+	if !reflect.ValueOf(o.Cloud).IsZero() {
+		c = o.Cloud
+	}
+	if conf, ok := c.Services[cloud.ResourceManager]; ok && conf.Endpoint != "" && len(conf.Audiences) > 0 {
+		return conf, nil
+	} else {
```
jhendrixMSFT, line 58: Nit, `else` can be removed.

jhendrixMSFT (approved):

sdk/azcore/arm/runtime/pipeline.go
chlowell: The `else` makes `conf` available after the `if` block. I liked this better than declaring `var conf cloud.ServiceConfiguration` before the conditional; is that better style?

sdk/azcore/cloud/cloud.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+type cloudConfiguration struct {
+	Authentication  authentication    `json:"authentication,omitempty"`
+	Name            string            `json:"name,omitempty"`
+	ResourceManager string            `json:"resourceManager,omitempty"`
+	Suffixes        map[string]string `json:"suffixes,omitempty"`
+}
+
+// https://management.azure.com/metadata/endpoints?api-version=2019-05-01 (minus Azure Germany)
```
richardpark-msft, line 103: Why is Azure Germany left out?

chlowell: It closed in October: https://www.microsoft.com/en-us/cloud-platform/germany-cloud-regions

sdk/azcore/arm/runtime/pipeline.go
jhendrixMSFT: I missed that part.  In this case it's fine.

sdk/azcore/cloud/cloud.go
```diff
@@ -0,0 +1,192 @@
...
+var WellKnownClouds = map[Name]Configuration{}
+
+// Name identifies a cloud.
+type Name string
+
+const (
+	// AzureChina is a global constant identifying Azure China.
+	AzureChina Name = "AzureChina"
+	// AzureGovernment is a global constant identifying Azure Government.
+	AzureGovernment Name = "AzureGovernment"
```
benbp, line 21: I think this should be `AzureUSGovernment` consistent with the naming in other tooling, and also `AzureGovernment` isn't really specific if you're not US-based anyway.

chlowell: `AzureUSGovernment` is a good name and consistent with ARM metadata at least. I'm not sure it's the right name though. Our GA Identity libraries have an `AzureGovernment` constant because the cloud is called "Azure Government" in docs. I also doubt we want to prioritize consistency with ARM metadata. That would mean having `AzureCloud` instead of `AzurePublicCloud`. The latter isn't a great name but it's clearer than the former. I chose to start with inter-SDK consistency and intend to raise this with architects.

sdk/azcore/arm/runtime/pipeline.go
```diff
@@ -7,36 +7,62 @@
...
 	}
-	ep := options.Endpoint
-	if len(ep) == 0 {
-		ep = arm.AzurePublicCloud
+	conf, err := getConfiguration(&options.ClientOptions)
+	if err != nil {
+		// TODO: return the error
+		panic(err)
+	}
+	scope := conf.Audiences[0]
```
jhendrixMSFT, line 35: Do we need to doc that the first audience is the scope?  In what scenarios will more than one audience be specified?

jhendrixMSFT (approved):
One question on the scopes.

sdk/azcore/arm/runtime/pipeline.go
chlowell: Thanks for commenting on this line, there's a bug here (scope = audience + "/.default"). 

AAD requires exactly one audience per token. Arguably we should have `Audience string` instead (item for future discussion). However, ARM metadata contains two audiences. I believe this is for historical reasons. One audience is for Service Management, the other for Resource Manager. The latter accepts tokens for the former, so I don't know any functional reason to prefer one over the other and don't think we need to document our arbitrary choice.

jhendrixMSFT:
@chlowell do we need to take this back to the arch board?

sdk/azcore/cloud/cloud.go
```diff
@@ -0,0 +1,44 @@
+//go:build go1.16
+// +build go1.16
```
jhendrixMSFT, line 2: The build constraints will need to be updated to go1.18 after rebasing on `main`.

jhendrixMSFT (approved):

chlowell:
I'll ping the Teams channel to see whether any architect wants another board meeting for this. I think offline discussion is more likely.

PR #16945, opened 2022-01-31
[Storage] improving documentation

seankane-msft:
Improving documentation for `pkg.go.dev` viewers


sdk/storage/azblob/zt_examples_test.go
```diff
@@ -133,20 +114,21 @@ func Example() {
...
-	_, err = container.Delete(ctx, nil)
+	// Delete the container.
+	_, err = container.Delete(context.TODO(), nil)
 	if err != nil {
 		log.Fatal(err)
 	}
 }
 
 // Awaiting Mohit's test PR, StorageError is barren for now.
+// TODO: is this still valid?
```
seankane-msft, line 131: @mohsha-msft is this true? should this example be shown?

sdk/storage/azblob/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.3.0 (Unreleased)
 
 ### Features Added
+* Exported fields `Offset` and `Count` on `HttpRange`
```
seankane-msft, line 6: This is not exported, but is accessed as if it is exported in one of the examples.

mohsha-msft: I've fixed this change in my other PR. So we should be able to export it from next release onwards.

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: That must be some old comment. Will remove it.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -52,76 +37,72 @@ func Example() {
...
-	cred, err := NewSharedKeyCredential(accountName, accountKey)
+	cred, err := azblob.NewSharedKeyCredential(accountName, accountKey)
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	// Open up a service client.
-	// You'll need to specify a service URL, which for blob endpoints usually makes up the syntax http(s)://<account>.blob.core.windows.net/
-	service, err := NewServiceClientWithSharedKey("https://"+accountName+".blob.core.windows.net/", cred, nil)
+	// The service URL for blobs endpoints is usually in the form: http(s)://<account>.blob.core.windows.net/
```
kimprice, line 45:
```suggestion
	// The service URL for blob endpoints is usually in the form: http(s)://<account>.blob.core.windows.net/
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -52,76 +37,72 @@ func Example() {
...
-	// All operations in the Azure Storage Blob SDK for Go operate on a context.Context, allowing you to control cancellation/timeout.
-	ctx := context.Background() // This example has no expiry.
-
-	// This example showcases several common operations to help you get started, such as:
-
 	// ===== 1. Creating a container =====
 
-	// First, branch off of the service client and create a container client.
+	// First, create a container client, and use the Create method to create a new container in your account
 	container := service.NewContainerClient("myContainer")
```
kimprice, line 54: I thought we fixed this before, but that container name will cause an error because we can't use capital letters. More details on naming here: https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-and-Referencing-Containers--Blobs--and-Metadata
```suggestion
	container := service.NewContainerClient("my-container")
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -52,76 +37,72 @@ func Example() {
...
-	// We'll specify our data up-front, rather than reading a file for simplicity's sake.
 	data := "Hello world!"
 
-	// Branch off of the container into a block blob client
+	// Create a new BlockBlobClient from the ContainerClient
 	blockBlob := container.NewBlockBlobClient("HelloWorld.txt")
 
 	// Upload data to the block blob
-	_, err = blockBlob.Upload(ctx, NopCloser(strings.NewReader(data)), nil)
+	_, err = blockBlob.Upload(context.TODO(), streaming.NopCloser(strings.NewReader(data)), nil)
```
kimprice, line 70: Just curious why we started using `context.TODO()` instead of `context.Background()`?

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -52,76 +37,72 @@ func Example() {
...
 		return
 	}
 	if data != downloadedData.String() {
-		log.Fatal("downloaded data doesn't match uploaded data")
+		log.Fatal("downloaded data does not match uploaded data")
+	} else {
+		fmt.Printf("Downloaded data: %s\n", downloadedData.String())
 	}
 
 	// ===== 3. list blobs =====
```
kimprice, line 99: Nitpick for consistency with the other steps.
```suggestion
	// ===== 3. Listing blobs =====
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -52,76 +37,72 @@ func Example() {
...
+	get, err := blockBlob.Download(context.TODO(), nil)
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	// Open a buffer, reader, and then download!
+	// Use the bytes.Buffer object to read the downloaded data.
 	downloadedData := &bytes.Buffer{}
-	reader := get.Body(RetryReaderOptions{}) // RetryReaderOptions has a lot of in-depth tuning abilities, but for the sake of simplicity, we'll omit those here.
+	reader := get.Body(azblob.RetryReaderOptions{}) // RetryReaderOptions options for tuning downloads.
```
kimprice, line 83: Should we mention something in the comment about the options being empty or default? 
This is one of the bugs found during the study - that we should allow 'nil' to be passed in here for default options in the future. Do we have an issue filed for this?

sdk/storage/azblob/zt_examples_test.go
kimprice: Another thing to file an issue for if we haven't already: modifying the Upload function to take a ReadSeeker instead of a ReadSeekCloser, removing the need to use NopCloser.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
-	// Note that: SAS tokens may be limited to a specific container or blob.
-	// You should be careful about modifying SAS tokens, as you might take them outside of their original scope accidentally.
-	parts.SAS = SASQueryParameters{}
+	parts := azblob.NewBlobURLParts(u)
+
+	// The BlobURLParts allows access to individual portions of a Blob URL
+	fmt.Printf("Host: %s\nContainerName: %s\nBlobName: %s\nSnapshot: %s\n", parts.Host, parts.ContainerName, parts.BlobName, parts.Snapshot)
+	fmt.Printf("Version: %s\nResource: %s\nStartTime: %s\nExpiryTime: %s\nPermissions: %s\n", parts.SAS.Version(), parts.SAS.Resource(), parts.SAS.StartTime(), parts.SAS.ExpiryTime(), parts.SAS.Permissions())
+
+	// You can alter fields toconstruct a new URL:
```
kimprice, line 199:
```suggestion
	// You can alter fields to construct a new URL:
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
 	parts.Snapshot = ""
 	parts.ContainerName = "othercontainer"
 
 	// construct a new URL from the parts
 	fmt.Print(parts.URL())
 }
 
-// This example demonstrates how to use the SAS token convenience generators.
-// Though this example focuses on account SAS, these generators exist across all clients (Service, Container, Blob, and specialized Blob clients)
+// This example demonstrates how to use the SAS token generators, this example focuses on account SAS,
```
kimprice, line 209:
```suggestion
// This example demonstrates how to use the SAS token generators.
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
+// these generators exist across all clients (Service, Container, Blob, and specialized Blob clients)
```
kimprice, line 210:
```suggestion
// This example specifically generates an account SAS token. 
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
 func ExampleServiceClient_GetSASToken() {
-	// Initialize a service client
```
kimprice, line 233: Why remove this comment?

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
 	// You can also break a blob URL up into it's constituent parts
-	blobURLParts := NewBlobURLParts(serviceClient.URL())
+	blobURLParts := azblob.NewBlobURLParts(serviceClient.URL())
 	fmt.Printf("SAS expiry time = %s\n", blobURLParts.SAS.ExpiryTime())
 }
 
 // This example demonstrates how to create and use a Blob service Shared Access Signature (SAS)
 func ExampleBlobSASSignatureValues() {
-	// Gather your account key and name from the Azure portal
-	// Supplying them via environment variables is recommended.
```
kimprice, line 306: Why we remove the comment? Are we assuming people would be reading through all examples?

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
 	// You can also break a blob URL up into it's constituent parts
-	blobURLParts := NewBlobURLParts(serviceClient.URL())
+	blobURLParts := azblob.NewBlobURLParts(serviceClientUsingSAS.URL())
 	fmt.Printf("SAS expiry time = %s\n", blobURLParts.SAS.ExpiryTime())
 }
 
 // This example shows how to create and use an Azure Storage account Shared Access Signature (SAS).
 func ExampleAccountSASSignatureValues_Sign() {
-	accountName, accountKey := accountInfo()
+	accountName, accountKey := os.Getenv("AZURE_STORAGE_ACCOUNT_NAME"), os.Getenv("AZURE_STORAGE_ACCOUNT_KEY")
```
kimprice, line 260: Should we include a similar comment here on creating the account?

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -397,287 +380,292 @@ func ExampleContainerClient_SetAccessPolicy() {
...
 		log.Fatal(err)
 	}
 
-	// NOTE: The SetMetadata & SetProperties methods update the container's ETag & LastModified properties
+	// NOTE: SetMetadata & SetProperties methods update the container's ETag & LastModified properties
 }
 
-// This examples shows how to create a blob with metadata and then how to read & update
-// the blob's read-only properties and metadata.
+// This examples shows how to create a blob with metadata, read blob metadata, and update a blob's read-only properties and metadata.
```
kimprice, line 545:
```suggestion
// This example shows how to create a blob with metadata, read blob metadata, and update a blob's read-only properties and metadata.
```

kimprice:
In general, throughout the examples some comments were removed without a replacement. A lot of them were cases like the comments for setting the account information, which may seem redundant/unnecessary after a while, but if people are only looking at one example at a time, it's probably best to just keep them. Are there reasons we would want to remove some comments completely or is it more likely an oversight?

sdk/storage/azblob/zt_examples_test.go
seankane-msft: I use `context.TODO()` in documentation for two reasons: 
1. It's easily findable with a search of "TODO" in a code editor and it grabs the eye of something to change. TODO is also equal to Background in 
2. Also per the [documentation](https://pkg.go.dev/context#TODO) it's recommended
> TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

sdk/storage/azblob/zt_examples_test.go
seankane-msft: Not sure if there is an issue for this, however it looks like a small change I might be able to do this quickly.

seankane-msft:
> In general, throughout the examples some comments were removed without a replacement. A lot of them were cases like the comments for setting the account information, which may seem redundant/unnecessary after a while, but if people are only looking at one example at a time, it's probably best to just keep them. Are there reasons we would want to remove some comments completely or is it more likely an oversight?

@kimprice my goal was to remove excess comments from code on portions that I thought were obvious. I think that's a conversation we should take offline to discuss what knowledge we expect doc readers to already have and which is same to assume they have.

sdk/storage/azblob/zt_examples_test.go
kimprice: Should we then change the other examples? I saw some later in this file that still use Background.

sdk/storage/azblob/zt_examples_test.go
seankane-msft: For clarification, remove the comment or remove the example?

sdk/storage/azblob/zt_examples_test.go
seankane-msft: Yes, thanks for catching that. Updated.

sdk/storage/azblob/zt_examples_test.go
seankane-msft: The service client wasn't being initialized in the line below and I don't think that we need to clarify that portion.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -201,182 +183,175 @@ func ExampleStorageError() {
...
+	accountName, accountKey := os.Getenv("AZURE_STORAGE_ACCOUNT_NAME"), os.Getenv("AZURE_STORAGE_ACCOUNT_KEY")
 
-	// Use your storage account's name and key to form a credential object
-	credential, err := NewSharedKeyCredential(accountName, accountKey)
+	credential, err := azblob.NewSharedKeyCredential(accountName, accountKey)
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	containerName := "myContainer"
```
kimprice: Just noticed this one as well.
```suggestion
	containerName := "mycontainer"
```

sdk/storage/azblob/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-The Microsoft Azure Storage SDK for Go allows you to build applications that takes advantage of Azure's scalable cloud storage. 
+The Microsoft Azure Storage SDK for Go allows you to build applications that takes advantage of Azure's scalable cloud storage.
 This SDK replaces the previously previewed [azblob package](https://github.com/azure/azure-storage-blob-go).
```
kimprice, line 6: We should use beta/stable to refer to our libraries and preview/GA to refer to our services.
```suggestion
This is the new beta client library for Azure Blob Storage, which follows our [Azure SDK Design Guidelines for Go](https://azure.github.io/azure-sdk/golang_introduction.html) and replaces the previous beta [azblob package](https://github.com/azure/azure-storage-blob-go).
```

sdk/storage/azblob/README.md
```diff
@@ -20,10 +20,10 @@ The Azure Blob SDK can access an Azure Storage account.
 
 ### Install the package
 * Install the Azure blob storage for Go with `go get`:
```
kimprice, line 44:
```suggestion
* Install the Azure Blob Storage package for Go with `go get`:
```

sdk/storage/azblob/README.md
```diff
@@ -2,7 +2,7 @@
```
kimprice, line 2: I think this title should actually be **Azure Blob Storage client module for Go**. I see that other SDKs call it "Azure Storage Blobs" but the service is called "Azure Blob Storage." Thoughts?

seankane-msft: I think it's better to be consistent across SDKs. I think we've also had this discussion about Tables in the past.

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: Remove the comment. Not example.

sdk/storage/azblob/README.md
```diff
@@ -12,18 +12,40 @@ The Azure Blob SDK can access an Azure Storage account.
...
+	az storage account create --resource-group my-resource-group --name my-storage-account
+	```
+
+	The storage account name can be queried with:
+	```
+	az storage account show -n mystorageaccount -g MyResourceGroup --query "primaryEndpoints.blob"
+	```
+	You can set this as an environment variable with:
+	```pwsh
+	$ENV:AZURE_STORAGE_ACCOUNT_NAME="mystorageaccount"
```
mohsha-msft, line 31: This is quite specific to Windows. Either we provide way for both Unix and Window or we can just ask them to setup on their own by specifying what env vars need to be set.

sdk/storage/azblob/CHANGELOG.md
```diff
@@ -2,15 +2,13 @@
...
-
 ### Breaking Changes
 * Updated to latest `azcore`.  Public surface area is unchanged.
 * [#16978](https://github.com/Azure/azure-sdk-for-go/pull/16978): The `DownloadResponse.Body` parameter is now `*RetryReaderOptions`.
 
 ### Bugs Fixed
-* Fixed Issue #16193 : azblob.GetSASToken wrong singed resource. 
-* Fixed Issue #16223 : HttpRange does not expose its fields. 
+* Fixed Issue #16193 : azblob.GetSASToken wrong singed resource.
+* Fixed Issue #16223 : HttpRange does not expose its fields.
```
mohsha-msft, line 11: Can you add https://github.com/Azure/azure-sdk-for-go/issues/16988 fix as well?

sdk/storage/azblob/README.md
```diff
@@ -12,18 +12,40 @@ The Azure Blob SDK can access an Azure Storage account.
 ### Prerequisites
 
 * Go versions 1.16 or higher
-* You must have an [Azure storage account][azure_storage_account]
+* You must have an [Azure storage account][azure_storage_account]. If you need to create one, you can use the [Azure Cloud Shell](https://shell.azure.com/bash) to create one with these commands (replace `"my-resource-group"` and `"myStorageAccount"` with your own unique names):
```
mohsha-msft, line 15: Lowercase letters only for storage account `myStorageAccount`. Can you also validate nomenclature for resource group as well?

sdk/storage/azblob/README.md
```diff
@@ -12,18 +12,40 @@ The Azure Blob SDK can access an Azure Storage account.
...
+	(Optional) if you want a new resource group to hold the Storage Account:
+	```
+	az group create --name my-resource-group --location westus2
+	```
+	Create the storage account:
+	```
+	az storage account create --resource-group my-resource-group --name my-storage-account
```
mohsha-msft, line 22: same as above. No hyphen in storage account.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-package azblob
+package azblob_test
```
mohsha-msft, line 4: Nice: This was missing, Thanks 👍

sdk/storage/azblob/README.md
seankane-msft: Added portion for bash

sdk/storage/azblob/README.md
kimprice: I talked to Kyle and he linked me to our style guide which says to use "Azure Blob Storage" so we should use that. And we'll work on getting the other SDKs updated to follow the same convention since we definitely want to be consistent.

Issue #16768, opened 2022-01-06
[KeyVault] Add Client Methods for Certificates


seankane-msft:
- [x] `BackupCertificate`
- [x] `BeginCreateCertificate`
- [x] `BeginDeleteCertificate`
- [x] `BeginRecoverDeletedCertificate`
- [x] `CreateContacts`
- [x] `CreateIssuer`
- [X] `DeleteCertificateOperation`
- [x] `DeleteContacts`
- [x] `DeleteIssuer`
- [x] `GetCertificate`
- [x] `GetCertificateOperation`
- [x] `GetCertificatePolicy`
- [x] ~~`GetCertificateVersion`~~ (GetCertificate has an optional parameter for `Version`)
- [x] `GetContacts`
- [x] `GetDeletedCertificate`
- [x] `GetIssuer`
- [x] `ListCertificates`
- [x] `ListCertificateVersions`
- [x] `ListIssuers`
- [x] `ImportCertificate`
- [x] `MergeCertificate`
- [x] `PurgeDeletedCertificate`
- [x] `RestoreDeletedCertificate`
- [x] `UpdateCertificatePolicy`
- [x] `UpdateCertificateProperties`
- [x] `UpdateIssuer`

PR #16947, opened 2022-01-31
[KeyVault] Create azcertificates library

seankane-msft:
#16769 #16768 #16767

APIView: https://apiview.dev/Assemblies/Review/4ffcffe7d06547c8b4f95635d71b196e
Champion Scenarios: https://gist.github.com/seankane-msft/882d9468bc20434efdc771c62d94cebf


sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+	return &generated.KeyVaultClientPurgeDeletedCertificateOptions{}
+}
+
+// PurgeDeletedCertificateResponse contains the response from method Client.PurgeDeletedCertificate.
+type PurgeDeletedCertificateResponse struct {
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+}
+
+// PurgeDeletedCertificate operation performs an irreversible deletion of the specified certificate, without possibility for  recovery. The operation
```
heaths, line 447: Nit:
```suggestion
// PurgeDeletedCertificate operation performs an irreversible deletion of the specified certificate, without possibility for recovery. The operation
```

sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+// NextPage fetches the next available page of results from the service. If the fetched page
+// contains results, the return value is true, else false. Results fetched from the service
+// can be evaluated by calling PageResponse on this Pager.
+func (l *ListCertificatesPager) NextPage(ctx context.Context) bool {
+	return l.genPager.NextPage(ctx)
+}
+
+// ListCertificatesOptions contains the optional parameters for the Client.ListCertificates method
+type ListCertificatesOptions struct {
+	MaxResults *int32
```
heaths, line 623: We don't expose this in other languages and it should probably be removed.

sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+// contains results, the return value is true, else false. Results fetched from the service
+// can be evaluated by calling PageResponse on this Pager.
+func (l *ListCertificateVersionsPager) NextPage(ctx context.Context) bool {
+	return l.genPager.NextPage(ctx)
+}
+
+// ListCertificateVersionsOptions contains the options for the ListCertificateVersions operations
+type ListCertificateVersionsOptions struct {
+	// Maximum number of results to return in a page. If not specified the service will return up to 25 results.
+	MaxResults *int32
```
heaths, line 698: Same here and anywhere `MaxResults` is exposed. Shouldn't even be necessary to declare since we don't pass it to the server (leaving the default in place).

sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+
+// CreateIssuerResponse is the response struct for the Client.CreateIssuer function
+type CreateIssuerResponse struct {
+	IssuerBundle
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+}
+
+// CreateIssuer adds or updates the specified certificate issuer. This operation requires the certificates/setissuers permission.
+func (c *Client) CreateIssuer(ctx context.Context, issuerName string, provider string, options *CreateIssuerOptions) (CreateIssuerResponse, error) {
```
heaths, line 776: This doesn't appear to be close to https://github.com/Azure/azure-sdk-for-net/blob/a7743d7564c9d60dff0ec17822af0f18323f5f15/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateClient.cs#L1131 or its related types like [CertificateIssuer](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateIssuer.cs).

sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+
+	return ListIssuersPage{
+		RawResponse: i.RawResponse,
+		Issuers:     vals,
+	}
+}
+
+// ListIssuers returns a pager that can be used to get the set of certificate issuer resources in the specified key vault. This operation
+// requires the certificates/manageissuers/getissuers permission.
+func (c *Client) ListIssuers(options *ListIssuersOptions) ListIssuersPager {
```
heaths, line 905: We all call this `ListPropertiesOfIssuers` since it returns their properties and not an actual `CertificateIssue`: https://github.com/Azure/azure-sdk-for-net/blob/a7743d7564c9d60dff0ec17822af0f18323f5f15/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateClient.cs#L1375

sdk/keyvault/azcertificates/client.go
```diff
@@ -0,0 +1,1568 @@
...
+// MergeCertificateResponse contains the response from method Client.MergeCertificate.
+type MergeCertificateResponse struct {
+	CertificateBundle
+
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+}
+
+// The MergeCertificate operation performs the merging of a certificate or certificate chain with a key pair currently available in the service. This operation requires the certificates/create permission.
+func (c *Client) MergeCertificate(ctx context.Context, certName string, certificates [][]byte, options *MergeCertificateOptions) (MergeCertificateResponse, error) {
```
heaths, line 1248: We pack every parameter into a `MergeCertificateOptions`: https://github.com/Azure/azure-sdk-for-net/blob/a7743d7564c9d60dff0ec17822af0f18323f5f15/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateClient.cs#L1606

No other parameters are exposed (sans pipeline options, etc.).

sdk/keyvault/azcertificates/constants.go
```diff
@@ -0,0 +1,113 @@
...
+
+package azcertificates
+
+// ActionType - The type of the action.
+type ActionType string
+
+const (
+	ActionTypeEmailContacts ActionType = "EmailContacts"
+	ActionTypeAutoRenew     ActionType = "AutoRenew"
+)
```
heaths, line 15: `CertificatePolicyAction`:

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificatePolicyAction.cs

sdk/keyvault/azcertificates/constants.go
```diff
@@ -0,0 +1,113 @@
...
+	DeletionRecoveryLevelRecoverable DeletionRecoveryLevel = "Recoverable"
+	// DeletionRecoveryLevelRecoverableProtectedSubscription - Denotes a vault and subscription state in which deletion is recoverable within retention interval
+	// (90 days), immediate and permanent deletion (i.e. purge) is not permitted, and in which the subscription itself cannot be permanently canceled. System
+	// wil permanently delete it after 90 days, if not recovered
+	DeletionRecoveryLevelRecoverableProtectedSubscription DeletionRecoveryLevel = "Recoverable+ProtectedSubscription"
+	// DeletionRecoveryLevelRecoverablePurgeable - Denotes a vault state in which deletion is recoverable, and which also permits immediate and permanent deletion
+	// (i.e. purge). This level guarantees the recoverability of the deleted entity during the retention interval (90 days), unless a Purge operation is requested,
+	// or the subscription is cancelled. System wil permanently delete it after 90 days, if not recovered
+	DeletionRecoveryLevelRecoverablePurgeable DeletionRecoveryLevel = "Recoverable+Purgeable"
+)
```
heaths, line 57: You don't need to declare an enum for these. Not only can - and have - changed, we just use strings because of that fact. It's not something users can pass back and only informational.

sdk/keyvault/azcertificates/constants.go
```diff
@@ -0,0 +1,113 @@
...
+
+// JSONWebKeyCurveName - Elliptic curve name. For valid values, see JsonWebKeyCurveName.
+type JSONWebKeyCurveName string
+
+const (
+	JSONWebKeyCurveNameP256  JSONWebKeyCurveName = "P-256"
+	JSONWebKeyCurveNameP256K JSONWebKeyCurveName = "P-256K"
+	JSONWebKeyCurveNameP384  JSONWebKeyCurveName = "P-384"
+	JSONWebKeyCurveNameP521  JSONWebKeyCurveName = "P-521"
+)
```
heaths, line 72: `CertificateKeyCurveName`: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateKeyCurveName.cs

Basically the same as Keys' `KeyCurveName` but with `Certificate` prepended to avoid conflicts in most other languages that don't force a module reference like Go does.

sdk/keyvault/azcertificates/constants.go
```diff
@@ -0,0 +1,113 @@
...
+type JSONWebKeyType string
+
+const (
+	JSONWebKeyTypeEC     JSONWebKeyType = "EC"
+	JSONWebKeyTypeECHSM  JSONWebKeyType = "EC-HSM"
+	JSONWebKeyTypeOct    JSONWebKeyType = "oct"
+	JSONWebKeyTypeOctHSM JSONWebKeyType = "oct-HSM"
+	JSONWebKeyTypeRSA    JSONWebKeyType = "RSA"
+	JSONWebKeyTypeRSAHSM JSONWebKeyType = "RSA-HSM"
+)
```
heaths, line 89: `CertificateKeyType`: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateKeyType.cs

sdk/keyvault/azcertificates/constants.go
```diff
@@ -0,0 +1,113 @@
...
+	KeyUsageTypeCRLSign          KeyUsageType = "cRLSign"
+	KeyUsageTypeDataEncipherment KeyUsageType = "dataEncipherment"
+	KeyUsageTypeDecipherOnly     KeyUsageType = "decipherOnly"
+	KeyUsageTypeDigitalSignature KeyUsageType = "digitalSignature"
+	KeyUsageTypeEncipherOnly     KeyUsageType = "encipherOnly"
+	KeyUsageTypeKeyAgreement     KeyUsageType = "keyAgreement"
+	KeyUsageTypeKeyCertSign      KeyUsageType = "keyCertSign"
+	KeyUsageTypeKeyEncipherment  KeyUsageType = "keyEncipherment"
+	KeyUsageTypeNonRepudiation   KeyUsageType = "nonRepudiation"
+)
```
heaths, line 108: `CertificateKeyUsage`: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificateKeyUsage.cs

heaths (changes requested):

sdk/keyvault/azcertificates/test-resources.json
```diff
@@ -0,0 +1,331 @@
...
+            "type": "string",
+            "value": "[variables('blobContainerName')]"
+        },
+        "AZURE_KEYVAULT_ATTESTATION_URL": {
+            "type": "string",
+            "condition": "[parameters('enableHsm')]",
+            "value": "[format('https://{0}/', reference(variables('attestationSite')).defaultHostName)]"
+        }
+    }
+}
```
benbp, line 331: Can you convert this to bicep? We should be using bicep going forwards and I'd love for the go repo to be a good example of that. I can help with conversion/generation if you need it. `az bicep decompile -f test-resources.json`

heaths: This is a copy of what all other languages use. Normally I create/update this for KV and everyone just copies it. So it's not really a new definition. I'd like to do this for all other services, too. It's not an efficient use of time for every language to maintain their own test-resources.json (and related files).

I plan to convert this to bicep at some point, but it's not been a priority. The ARM templates work just fine and so far we've only needed simple incremental changes on top them.

sdk/keyvault/azcertificates/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+```go
+import (
+    "github.com/Azure/azure-sdk-for-go/sdk/keyvault/azcertificates"
+    "github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+)
+
+func main() {
+    credential, err := azidentity.NewDefaultAzureCredential(nil)
+    handle(err)
```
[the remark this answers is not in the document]
seankane-msft: Fixed to check if nil and panic if it is

sdk/keyvault/azcertificates/README.md
```diff
@@ -119,10 +119,14 @@ import (
 
 func main() {
     credential, err := azidentity.NewDefaultAzureCredential(nil)
-    handle(err)
+	if err != nil {
+		panic(err)
+	}
```
heaths, line 124: Nit: spacing seems off.

sdk/keyvault/azcertificates/client.go
seankane-msft: I believe the `IssuerBundle` and `CertificateIssuer` are essentially the same type. I'm changing the name of IssuerBundle, to CertificateIssuer

sdk/keyvault/azcertificates/client.go
seankane-msft: The Certificates are required parameters correct, can you pass a nil parameter and have a successful response from the service? If you can then you're correct we can pack it into the options. `certName` is also a required parameter so we cannot  pack it into the options. The thought being, a user should always be able to pass in `nil` for an Options struct for the default options.

sdk/keyvault/azcertificates/client.go
heaths: None of the *Bundle model names made it into the track 2 SDKs.

sdk/keyvault/azcertificates/client.go
heaths: "options" doesn't mean optional, per architects on past reviews. For a brief point in time, they wanted to use *Parameters for any options bag that had required parameters (hence our `CryptographyClient` method signautres), but changed to just use *Options for everything - required or not. So in this case, `MergeCertificateOptions` is required and has required fields, but also optional fields. That's what we do in other languages as well.

sdk/keyvault/azcertificates/models.go
```diff
@@ -590,7 +590,7 @@ func issuerParametersFromGenerated(g *generated.IssuerParameters) *IssuerParamet
 // KeyProperties - Properties of the key pair backing a certificate.
 type KeyProperties struct {
 	// Elliptic curve name. For valid values, see JsonWebKeyCurveName.
-	Curve *JSONWebKeyCurveName `json:"crv,omitempty"`
+	Curve *CertificateKeyCurveName `json:"crv,omitempty"`
```
heaths, line 593: We don't expose anything like `KeyProperties` for certificates. All this (and with slightly different names, like `KeyCurveName` here) hangs off of `CertificatePolicy`, which is more or less the input to create a certificate: https://github.com/Azure/azure-sdk-for-net/blob/75b74e3d5bdd65a7d34c5087faded65595b6a23a/sdk/keyvault/Azure.Security.KeyVault.Certificates/src/CertificatePolicy.cs#L163-L183

Or, if it's easier, see our APIView. This was reviewed across languages as well: https://apiview.dev/Assemblies/Review/da210106190946c0874b30d5f84f41b1

heaths (changes requested):
Looks better, but there's still a few methods and structures here that don't match what we do for other languages (even taking idioms into account). Might be better to look at APIView, which we already reviewed across languages.

sdk/keyvault/azcertificates/client.go
jhendrixMSFT: As Go doesn't have default/optional parameters, we have a `<ClientName><Method>Options` type that contains all non-mandatory parameters.  All methods take this regardless of if they contain optional parameters.  All mandatory parameters are pass individually.  This is the pattern all of our SDKs follow, please see the [guidelines](https://azure.github.io/azure-sdk/golang_introduction.html) for more info.

sdk/keyvault/azcertificates/client.go
seankane-msft: Fixed

sdk/keyvault/azcertificates/models.go
seankane-msft: I believe I am more inline with this API while staying idiomatic to Go. Can you take another look at this PR now?

sdk/keyvault/azcertificates/test-resources.json
benbp: Ok let's just treat the keyvault ones as a batch then. I'll approach the conversion on a service to service basis instead of a language to language basis.

heaths: I think it's also completely valid to encourage conversion. That said, I'm typically very much of the mind: if it ain't broke, don't fix it. But for new services we should perhaps encourage it through PRs.

heaths (commented):
Almost. If we had https://github.com/Azure/azure-sdk-tools/issues/676 or https://github.com/Azure/azure-sdk-tools/issues/1703 it would be much easier to compare.

sdk/keyvault/azcertificates/models.go
```diff
@@ -295,14 +251,26 @@ func certificateOperationFromGenerated(g generated.CertificateOperation) Certifi
...
+	// The certificate properties.
+	Properties *CertificateProperties `json:"attributes,omitempty"`
 
 	// Parameters for the issuer of the X509 component of a certificate.
 	IssuerParameters *IssuerParameters `json:"issuer,omitempty"`
 
-	// Properties of the key backing a certificate.
-	KeyProperties *KeyProperties `json:"key_props,omitempty"`
+	// Elliptic curve name. For valid values, see JsonWebKeyCurveName.
+	Curve *CertificateKeyCurveName `json:"crv,omitempty"`
```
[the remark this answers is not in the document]
seankane-msft: Fixed

sdk/keyvault/azcertificates/client.go
```diff
@@ -582,15 +582,15 @@ func (c *Client) ImportCertificate(ctx context.Context, certName string, base64E
...
 		RawResponse: resp.RawResponse,
-		CertificateBundle: CertificateBundle{
-			Attributes:     certificateAttributesFromGenerated(resp.Attributes),
+		Certificate: Certificate{
+			Properties:     certificateAttributesFromGenerated(resp.Attributes),
 			Cer:            resp.Cer,
 			ContentType:    resp.ContentType,
 			Tags:           convertGeneratedMap(resp.Tags),
 			ID:             resp.ID,
 			Kid:            resp.Kid,
```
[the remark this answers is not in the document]
seankane-msft: Fixed

sdk/keyvault/azcertificates/client.go
```diff
@@ -216,7 +216,7 @@ type GetCertificateOptions struct {
 
 // GetCertificateResponse contains the result from method Client.GetCertificate.
 type GetCertificateResponse struct {
-	CertificateBundle
+	Certificate
```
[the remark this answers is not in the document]
seankane-msft: Added this model and reflected the response models with it to reflect .NET.

sdk/keyvault/azcertificates/models.go
```diff
@@ -101,8 +101,35 @@ func certificateAttributesFromGenerated(g *generated.CertificateAttributes) *Cer
...
+	Cer []byte `json:"cer,omitempty"`
+
+	// The content type of the secret. eg. 'application/x-pem-file' or 'application/x-pkcs12',
+	ContentType *string `json:"contentType,omitempty"`
+
+	// Application specific metadata in the form of key-value pairs
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// READ-ONLY; The certificate id.
+	ID *string `json:"id,omitempty" azure:"ro"`
```
heaths, line 119: `azure:"ro"`

Whoa, that's neat! Where can I learn more about how that's processed?

sdk/keyvault/azcertificates/models.go
```diff
@@ -101,8 +101,35 @@ func certificateAttributesFromGenerated(g *generated.CertificateAttributes) *Cer
...
+
+	// READ-ONLY; The secret ID.
+	SecretID *string `json:"sid,omitempty" azure:"ro"`
+
+	// READ-ONLY; Thumbprint of the certificate.
+	X509Thumbprint []byte `json:"x5t,omitempty" azure:"ro"`
+}
+
+// KeyVaultCertificateWithPolicy - A certificate bundle consists of a certificate (X509) with a policy, and its properties.
+type KeyVaultCertificateWithPolicy struct {
```
heaths, line 132: Nit: wouldn't it make sense to just declare `KeyVaultCertificate` as a nested struct e.g.

```go
type KeyVaultCertificate struct {
  // ...
}

type KeyVAultCertificteWithPolicy struct {
  KeyVaultCertificate

  // READ-ONLY; The management policy.
  Policy *CertificatePolicy `json:"policy,omitempty" azure:"ro"`
}
```

It's derived from KeyVaultCertificate in .NET and Java, and practically in other languages.

heaths (approved):

sdk/keyvault/azcertificates/models.go
jhendrixMSFT: See https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/request.go#L153

sdk/keyvault/azcertificates/models.go
jhendrixMSFT: Composition makes it painful when creating struct literals of these things due to the extra nesting.  It can also introduce some unwanted side-effects by promoting methods on `KeyVaultCertificate` to the parent.

heaths: Understood, but in this case please at least consider:

1. These models are intended as a derivation, so members appearing in the `KeyVaultCertificateWithPolicy` model from `KeyVaultCertificate` are indeed intended.
2. This model shouldn't be creatable by the client anyway, so construction woes with struct literals (thanks, I forgot the name) shouldn't be much of a concern.

That said, I signed off. Merely an idea/nit, and certain up to your idiomatic designs to decide. 🙂

PR #16955, opened 2022-02-01
[azservicebus/stress] Make sure you pull the image always

richardpark-msft:
Make sure we always pull the stress test image.

Without this you'll end up in a situation where, if you're repeatedly testing, you end up running the last cached image, which is probably not what you want.


richardpark-msft:
@benbp - I'm wondering if this is already happening somewhere and I just missed it. And if it's not it should probably be the default (I think we can eat the bandwidth/disk wastage, should it come up).

benbp (approved):

PR #16956, opened 2022-02-01
[Core] updating changelog and version

seankane-msft:
Updates for release on 2/3

### DIFF
--- a/sdk/azcore/internal/shared/constants.go
+++ b/sdk/azcore/internal/shared/constants.go
@@ -35,5 +35,5 @@ const (
 	Module = "azcore"
 
 	// Version is the semantic version (see http://semver.org) of this module.
-	Version = "v0.22.0"
+	Version = "v0.21.0"
 )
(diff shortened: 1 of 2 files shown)


sdk/azcore/internal/shared/constants.go
```diff
@@ -35,5 +35,5 @@ const (
 	Module = "azcore"
 
 	// Version is the semantic version (see http://semver.org) of this module.
-	Version = "v0.21.1"
+	Version = "v0.22.0"
```
jhendrixMSFT, line 38: We should be able to keep this at v0.21.1 as there were no breaking changes introduced correct?

chlowell:
I intend to merge #16942, which takes care of this, unless someone objects.

seankane-msft:
> I intend to merge #16942, which takes care of this, unless someone objects.

That is fine by me. Could you include the issue numbers in the changelog entries (I've updated the existing two here). I will close this issue now.

PR #16967, opened 2022-02-03
[KeyVault] Release prep for keys and secrets

seankane-msft:


sdk/keyvault/azkeys/utils_test.go
```diff
@@ -31,7 +31,20 @@ var enableHSM = true
 
 func TestMain(m *testing.M) {
 	// Initialize
-	if recording.GetRecordMode() == "record" {
+	switch recording.GetRecordMode() {
+	case recording.PlaybackMode:
```
richardpark-msft, line 35: Comment here on why we don't do playback mode?

seankane-msft: All of these changes are only for the response bodies. There's no affect on request bodies.

jhendrixMSFT (approved):

PR #16973, opened 2022-02-03
Convert to package, JSON null funcs to generics

jhendrixMSFT:
NullValue now takes a generic type parameter instead of an interface arg
to determine the type of null sentinel value to create.
IsNullValue infers its generic parameter to determine the type of null
sentinel value to look for.
At present, there is no way to express a 'nillable' generic type
constraint so the funcs simply take a [T any] which should be fine as
they typically take/return pointer-to-types.
The 'to' package has been reduced to two funcs.

### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -23,14 +23,16 @@ type TokenCredential = shared.TokenCredential
 // holds sentinel values used to send nulls
 var nullables map[reflect.Type]interface{} = map[reflect.Type]interface{}{}
 
+func typeOfT[T any]() reflect.Type {
+	// you can't, at present, obtain the type of
+	// a type parameter, so this is the trick
+	return reflect.TypeOf((*T)(nil)).Elem()
+}
+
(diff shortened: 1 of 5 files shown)


sdk/azcore/core.go
```diff
@@ -23,39 +23,41 @@ type TokenCredential = shared.TokenCredential
...
-	}
+func NullValue[T any]() T {
+	t := typeOfT[T]()
 	v, found := nullables[t]
 	if !found {
 		var o reflect.Value
-		if k := t.Kind(); k == reflect.Slice || k == reflect.Map {
-			o = reflect.New(t) // *[]type / *map[]
-			o = o.Elem()       // []type / map[]
+		if k := t.Kind(); k == reflect.Map {
```
jhendrixMSFT, line 39: This is the same change that's pending in main.  I hope to get that merged then rebase and update this PR first.

seankane-msft (approved):
So much reduced code 😄

PR #16975, opened 2022-02-03
Disable linter for now

jhendrixMSFT:
Blows up on generic type parameters so disabling until it's working.
Restored installation to its original form.


seankane-msft (approved):
Fine by me, curious if we could add a pre-commit hook that runs `go fmt` on the files. I have to add a new commit often just to pass the formatting check

jhendrixMSFT:
We could, but I'm not a big fan of having CI make changes to source code and commit them.  Of course, formatting is a bit innocuous so maybe that's not so bad.  Curious though, is your VSCode not auto-formatting on save?

seankane-msft:
> We could, but I'm not a big fan of having CI make changes to source code and commit them. Of course, formatting is a bit innocuous so maybe that's not so bad. Curious though, is your VSCode not auto-formatting on save?

It does auto-format on save, but my machine is a bit slow (especially with the Go language server). Occassionally that lag doesn't format before I close VS Code.

PR #16985, opened 2022-02-04
[azservicebus] Adding in MessageState property for message

### LINKED ISSUE #16961: [azservicebus] set ReceivedMessage.MessageState from the system properties
MessageState is a useful property for consumers when they get a message from an API like peek,. 

MessageState can be:
- Active
- Deferred
- Scheduled

The value comes from SystemProperties with the key `x-opt-message-state`.

(PR adding this in M.A.S .NET lib): https://github.com/Azure/azure-sdk-for-net/pull/25158

richardpark-msft:
Adding in State property for message.

The State gives you information about whether a message is Active, Deferred or Scheduled.

Fixes #16961

### DIFF
--- a/sdk/messaging/azservicebus/message.go
+++ b/sdk/messaging/azservicebus/message.go
@@ -382,7 +382,7 @@ func asInt64(v interface{}, defVal int64) int64 {
 		return int64(v2)
 	case int:
 		return int64(v2)
+	default:
+		return defVal
 	}
-
-	return defVal
 }
--- a/sdk/messaging/azservicebus/message.go
+++ b/sdk/messaging/azservicebus/message.go
@@ -60,11 +60,15 @@ type ReceivedMessage struct {
 	deferred bool
 }
 
+// MessageState represents the current state of a message (Active, Scheduled, Deferred).
 type MessageState int32
 
 const (
-	MessageStateActive    MessageState = 0
-	MessageStateDeferred  MessageState = 1
+	// MessageStateActive indicates the message is active.
+	MessageStateActive MessageState = 0
+	// MessageStateDeferred indicates the message is deferred.
+	MessageStateDeferred MessageState = 1
+	// MessageStateScheduled indicates the message is scheduled.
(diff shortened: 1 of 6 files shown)


sdk/messaging/azservicebus/message.go
```diff
@@ -347,3 +373,16 @@ func uuidFromLockTokenBytes(bytes []byte) (*amqp.UUID, error) {
 
 	return &amqpUUID, nil
 }
+
+func asInt64(v interface{}, defVal int64) int64 {
```
jhendrixMSFT, line 377: When would `defVal` be returned and is this an expected scenario or just defensive programming?  Also, if it would always be zero, consider removing the arg and returning zero in the `default` clause.

richardpark-msft: It's defensive at the moment: if the value wasn't there (ie, was 'nil' or incorrectly typed). 

For today, zero would be right but it's also nice to have that documented locally (ie, at the call site) since it's pretty specific to the usage (ie, 'nil' might just mean the property was optional).

sdk/messaging/azservicebus/message.go
```diff
@@ -270,6 +287,15 @@ func newReceivedMessage(amqpMsg *amqp.Message) *ReceivedMessage {
 			msg.TransactionPartitionKey = to.StringPtr(viaPartitionKey.(string))
 		}
 
+		switch asInt64(amqpMsg.Annotations[messageStateAnnotation], 0) {
+		case 1:
+			msg.State = MessageStateDeferred
+		case 2:
+			msg.State = MessageStateScheduled
+		default:
```
jhendrixMSFT, line 295: Is it possible for more states to be added?  If so, they'd fall into this clause until we updated it.

richardpark-msft: This is the "official" list, so I think the fall through will be unexercised (this matches the .net logic, btw).

jhendrixMSFT (approved):

PR #16986, opened 2022-02-05
Revise azidentity error messages

### LINKED ISSUE #15923: Improve DefaultAzureCredential diagnosability
DefaultAzureCredential only adds to its chain the credentials it can instantiate, and surfaces instantiation failures only when it can't instantiate any credential. That never happens because AzureCLICredential is at the end of the chain. This means that DefaultAzureCredential never surfaces problems with environment configuration or managed identity. Its constructor can always instantiate AzureCLICredential, so it never returns an error, and errors from `DefaultAzureCredential.GetToken()` never mention problems with environment configuration or managed identity because the corresponding credentials aren't part of the chain when such problems exist.

chlowell:
This makes error messages in general more consistent and improves them in some key scenarios. For example, when no inner credential can provide a token, DefaultAzureCredential.GetToken() now includes messages from those credentials' constructors (fixes #15923). There are a couple functional changes I doubt will have any impact:
- ChainedTokenCredential returns credentialUnavailableError if and only if all its sources return that type of error
- AzureCLICredential returns credentialUnavailableError on Windows when it can't identify a safe working directory

### DIFF
--- a/sdk/azidentity/chained_token_credential.go
+++ b/sdk/azidentity/chained_token_credential.go
@@ -56,45 +56,39 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
 // GetToken calls GetToken on the chained credentials in turn, stopping when one returns a token. This method is called automatically by Azure SDK clients.
 // ctx: Context controlling the request lifetime.
 // opts: Options for the token request, in particular the desired scope of the access token.
-func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (token *azcore.AccessToken, err error) {
+func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
 	if c.successfulCredential != nil && !c.retrySources {
 		return c.successfulCredential.GetToken(ctx, opts)
 	}
-	var errList []credentialUnavailableError
+
+	var errs []error
 	for _, cred := range c.sources {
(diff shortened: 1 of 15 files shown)


sdk/azidentity/chained_token_credential.go
```diff
@@ -56,45 +56,39 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
-			return nil, err
-		} else {
+		token, err := cred.GetToken(ctx, opts)
+		if err == nil {
 			log.Writef(EventAuthentication, "Azure Identity => %s authenticated with %s", c.name, extractCredentialName(cred))
 			c.successfulCredential = cred
 			return token, nil
 		}
+		errs = append(errs, err)
+		if _, ok := err.(credentialUnavailableError); !ok {
```
jhendrixMSFT, line 73: Shouldn't this be checking for `authenticationFailedError`?

chlowell: The only expected, non-fatal error type is `credentialUnavailableError`. If we have any other type of error, I want to return `AuthenticationFailedError`.

jhendrixMSFT: Is there actually an HTTP response in this case though?  I thought that we get this error if there's not enough info to even attempt to authenticate.

jhendrixMSFT: Ooops I just realized the condition is `!ok`, makes sense now.

jhendrixMSFT (approved):

PR #16354, opened 2021-12-01
Adreed/fix pagers

adreed-msft:
Fixes #16295 by overriding the advancers, since the generated code was incorrect

PR #16992, opened 2022-02-07
azblob: fix ListBlob to preserve prefix while paging

serprex:
#16295 was not fixed by #16354 because _all_ options must be repeated in following requests, not just marker. ie prefix would be dropped when paging

This PR opts to modify the passed in options. It's probably preferable to clone the options to a variable within scope to avoid mutating caller's options


serprex:
@adreed-msft @mohsha-msft

mohsha-msft:
Thanks for your contribution @serprex . I've merged it with the next release candidate.

PR #17027, opened 2022-02-09
Add generic pager implementation

jhendrixMSFT:

### DIFF
--- a/sdk/azcore/runtime/pager.go
+++ b/sdk/azcore/runtime/pager.go
@@ -38,21 +38,22 @@ type PageProcessor[T any] struct {
 // Pager provides operations for iterating over paged responses.
 type Pager[T any] struct {
 	current   T
-	handler   PageProcessor[T]
-	lroSecond bool
+	processor PageProcessor[T]
+	firstPage bool
 }
 
 // NewPager creates an instance of Pager using the specified PageProcessor.
-func NewPager[T any](pageHandler PageProcessor[T]) *Pager[T] {
+func NewPager[T any](processor PageProcessor[T]) *Pager[T] {
 	return &Pager[T]{
-		handler: pageHandler,
+		processor: processor,
+		firstPage: true,
 	}
 }
 
 // More returns true if there are more pages to retrieve.
 func (p *Pager[T]) More() bool {
 	if !reflect.ValueOf(p.current).IsZero() {
-		return p.handler.More(p.current)
+		return p.processor.More(p.current)
 	}
 	return true
 }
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
I'll add tests once we close on the naming.

sdk/azcore/runtime/pager.go
```diff
@@ -0,0 +1,95 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package runtime
```
jhendrixMSFT, line 7: I placed this in the `runtime` package as customers never need to create these.  That said, I know that pagers (and pollers) are core concepts, so I could add a type alias in the `azcore` package if we think there's value exposing the type there (the constructor and supporting types would remain in `runtime`).

jhendrixMSFT: The current Go 1.18 beta doesn't support aliasing a generic type.  Unclear if this will land in 1.18 or not.

jhendrixMSFT: It looks like we won't have type-aliasing until Go 1.19.  I personally don't feel this is a blocking issue, and we can add the type-alias later which is non-breaking.

JeffreyRichter: I just tried this in some code I'm working on with Go 1.18 beta 2:
type Q = collections.Queue[int]   // Type alias to a generic type

var q Q

and this compiles just fine. What do you mean by type-aliasing of a generic isn't in 1.18 beta 2?

jhendrixMSFT: In `azcore`, I'd like to declare `type Pager = runtime.Pager` which is not yet supported.

Your example is an alias to a generic instantiation which is slightly different.

JeffreyRichter:
Got it. Thanks.

sdk/azcore/runtime/pager.go
```diff
@@ -0,0 +1,95 @@
...
+type PageProcessor[T any] struct {
+	// Do sends the request to fetch the next page.
+	Do func(*policy.Request) (*http.Response, error)
+
+	// More returns a boolean indicating if there are more pages to fetch.
+	// It uses the provided page to make the determination.
+	More func(T) bool
+
+	// Requester creates the request to fetch the first page.
+	Requester func(context.Context) (*policy.Request, error)
```
chlowell, line 29: Not that I want to redesign the API here, but do we need two methods for fetching pages? Naively, it seems `Requester` would be a straightforward special case for `Advancer`.

sdk/azcore/runtime/pager.go
```diff
@@ -0,0 +1,95 @@
...
+
+// NextPage advances the pager to the next page.
+func (p *Pager[T]) NextPage(ctx context.Context) (T, error) {
+	var req *policy.Request
+	var err error
+	if !reflect.ValueOf(p.current).IsZero() && p.firstPage {
+		// we get here if it's an LRO-pager, we already have the first page
+		p.firstPage = false
+		return p.current, nil
+	} else if !reflect.ValueOf(p.current).IsZero() {
```
chlowell, line 69: Nitpicking: already evaluated this condition

chlowell (approved):

sdk/azcore/runtime/pager.go
jhendrixMSFT: That's a good point.  Maybe we can fold this into a func named `Fetcher`, and when requesting the first page, the `T` is its zero-value?  I don't think there could ever be a case where that's ambiguous (or we just make it pointer-to-T).

sdk/azcore/runtime/pager.go
jhendrixMSFT: Fixed

sdk/azcore/runtime/pager.go
jhendrixMSFT: I went with the pointer-to-T as it simplifies the condition and removed the need to use reflection.

sdk/azcore/runtime/pager.go
```diff
@@ -0,0 +1,93 @@
...
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// PageProcessor contains the required data for constructing a Pager.
+type PageProcessor[T any] struct {
+	// Do sends the request to fetch the next page.
+	Do func(*policy.Request) (*http.Response, error)
```
jhendrixMSFT, line 21: We might also be able to do some further consolidation here, get rid of `Do` and merge `Fetcher` and `Responder`.  Let me try it out.

jhendrixMSFT: This does work.  It makes creating fetchers a bit more complex but is likely worth it in the long run.  Plus, it removes the hard-coded HTTP status code check which bugged me.

jhendrixMSFT:
Closing as I don't think we'll be using a generic pager (or poller).

PR #17036, opened 2022-02-10
[azservicebus] Fixing bug with panic caused by an incorrect error return in go-amqp

### LINKED ISSUE #17029: [azservicebus] nil pointer dereference in GetRecoveryKind()
- pkg: github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
- version: v0.3.4
- go version: go 1.17

**Description**

In testing I'm seeing a nil pointer dereference error generated by [this](https://github.com/Azure/azure-sdk-for-go/blob/2bb27d4b7ddec29b5a58291c1f2206865846d377/sdk/messaging/azservicebus/internal/errors.go#L157) line within `errors.go`.

Even though other functions calling `GetRecoveryKind()` have checked to see if `err == nil`, they have not accounted for the fact that an `error` interface containing a nil pointer to an `amqp.Error` type is non-nil. So there is still a possibility of a nil pointer error here.

The following code may make the problem clearer:

```
        var detachErr *amqp.Error
	var err error
	err = detachErr

	// err is not nil, even though it contains a nil pointer
	if err == nil {
		return
	}

	var amqpErr *amqp.Error
	if errors.As(err, &amqpErr) {
		// trying to access amqpErr.Condition will cause a panic
		t.Log(amqpErr.Condition)
	}
```
It would be great if you could address this problem.

richardpark-msft:
go-amqp was returning a nil *amqp.Error (accidentally) which was causing a nil panic for us in getRecoveryKind().

The root cause was actually that it should have been returning an error indicating that it was detached, so when we went to drain it we could see that the link was no longer valid. 

This has been fixed in the latest go-amqp (v.0.17.4)

Fixes #17029


sdk/messaging/azservicebus/go.mod
```diff
@@ -7,13 +7,12 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.13.0
 	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.3
 	github.com/Azure/azure-sdk-for-go/sdk/messaging/internal v0.0.0-20211208010914-2b10e91d237e
-	github.com/Azure/go-amqp v0.17.0
+	github.com/Azure/go-amqp v0.17.4
```
jhendrixMSFT, line 10: mod tidy?

sdk/messaging/azservicebus/receiver_test.go
```diff
@@ -85,6 +85,43 @@ func TestReceiverForceTimeoutWithTooFewMessages(t *testing.T) {
...
+
+	err = sender.SendMessage(context.Background(), &Message{
+		Body: []byte("hello"),
+	})
+	require.NoError(t, err)
+
+	receiver, err := serviceBusClient.NewReceiverForQueue(queueName, nil)
+	require.NoError(t, err)
+
+	// close the link _just_ before the drain credit to simulate connection loss.
```
jhendrixMSFT, line 104: The link recovery logic prevents us from seeing this connection hiccup right?

richardpark-msft: In this it's not the link recovery - it's the fact that we received some messages (this falls into the we just represent "partial failure as success" bucket).

I have another changeset to try to do connection recovery but haven't finished that yet.

sdk/messaging/azservicebus/go.mod
richardpark-msft: Will do once the next go-amqp is out.

richardpark-msft: Done.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/receiver_test.go
richardpark-msft: Okay, fixed this a slightly different way.

Basically, if the drain fails the link is detached (or dead) so there's nothing left to drain here. So instead, I'll just return. The next call will properly recover the link/connection if need be.

richardpark-msft:
Having a bit of a weird time getting all the validation systems to work properly on this one so I'm going to override checkenforcer. However:
- live test pipeline completed (had to stop the 'release' stages, but that's correct in this case) [link](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1360661&view=results)
- CI pipeline also passed [link](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1360579&view=results)

PR #17039, opened 2022-02-11
Revise azidentity logging

chlowell:
A couple small changes I hope will make authentication logs less noisy and more useful:
- Stop logging errors returned from `GetToken()` because the application gets the error and is compelled to handle it
- Remove "Azure Identity => " prefix because listeners don't need it for filtering or formatting

### DIFF
--- a/sdk/azidentity/default_azure_credential_test.go
+++ b/sdk/azidentity/default_azure_credential_test.go
@@ -5,7 +5,6 @@ package azidentity
 
 import (
 	"context"
-	"os"
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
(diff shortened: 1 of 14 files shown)


sdk/azidentity/default_azure_credential_test.go
```diff
@@ -52,37 +42,22 @@ func TestDefaultAzureCredential_defaultAzureCredentialConstructorErrorHandlerNoS
...
-		"<credential-name>: <error-message>",
-	}
 	err = defaultAzureCredentialConstructorErrorHandler(1, errorMessages)
 	if err != nil {
-		t.Fatal("Unexpected error", err.Error())
+		t.Fatal(err)
 	}
 
-	expectedLogs := `Azure Identity => Failed to initialize some credentials on the Default Azure Credential:
+	expectedLogs := `NewDefaultAzureCredential failed to initialize some credentials:
```
richardpark-msft, line 55: This is thorough in a good way, I should test my log messages too.

richardpark-msft (approved):
Looks good, and it's a good point. I was thinking about this the other day so this is pretty timely :)

jhendrixMSFT (approved):

PR #17082, opened 2022-02-16
[Blob] migrating to new test proxy

seankane-msft:
Migrating test recordings to the new test proxy.


sdk/storage/azblob/zt_access_policy_test.go
```diff
@@ -4,39 +4,43 @@
...
-		s.Fail("Unable to fetch service client because " + err.Error())
-	}
+func TestSetEmptyAccessPolicy(t *testing.T) {
+	stop := start(t)
+	defer stop()
+
+	_assert := assert.New(t)
+	testName := t.Name()
+	svcClient, err := createServiceClient(t, testAccountDefault)
+	require.NoError(t, err)
```
mohsha-msft: Any particular reason to use `require` instead of  `_assert` we were using before?

sdk/storage/azblob/zt_access_policy_test.go
```diff
@@ -4,39 +4,43 @@
...
 
 	containerName := generateContainerName(testName)
-	containerClient := createNewContainer(_assert, containerName, svcClient)
+	containerClient := createNewContainer(t, containerName, svcClient)
 	defer deleteContainer(_assert, containerClient)
```
mohsha-msft: In case we use `require`, we need to get rid of `_assert` completely.

sdk/storage/azblob/zt_access_policy_test.go
```diff
@@ -417,28 +322,29 @@ func (s *azblobTestSuite) TestContainerSetPermissionsACLMoreThanFive() {
 		},
 	}
 	_, err = containerClient.SetAccessPolicy(ctx, &setAccessPolicyOptions)
-	_assert.NotNil(err)
+	require.Error(t, err)
 
 	validateStorageError(_assert, err, StorageErrorCodeInvalidXMLDocument)
 }
 
-func (s *azblobTestSuite) TestContainerSetPermissionsDeleteAndModifyACL() {
```
mohsha-msft: We have two test suites - `azblobUnrecordedTestSuite`  and `azblobTestSuite`.  We run the former suite live and later suite as recorded tests. How do these changes distinguish between them?

mohsha-msft: I'll need your help setting up Live tests in CI too.

sdk/storage/azblob/zt_access_policy_test.go
seankane-msft: The `require` library will fail whereas `_assert` will continue testing and logs that the statement was not what was expected. This causes nil pointer exceptions, panics, and early exits whenever the first test fails. I think it's a huge positive to run all the tests in a suite rather than exit the suite on the first failure.

sdk/storage/azblob/zt_access_policy_test.go
seankane-msft: Agree, but I didn't want to touch every single test to do this. I'll make the complete removal in the second PR.

sdk/storage/azblob/zt_access_policy_test.go
seankane-msft: Live tests will already be set up, but I can help verify they are working as expected. I included an ARM template which the CI picks up to deploy resources. In the second PR we can verify that everything is working as expected.

seankane-msft: > We have two test suites - `azblobUnrecordedTestSuite` and `azblobTestSuite`. We run the former suite live and later suite as recorded tests. How do these changes distinguish between them?

What is the semantic difference between the two? I didn't find failures running the unrecorded tests in playback. I can mark those tests as live only but curious what the use-case is for that. I think it's best to run all playback tests whenever possible.

mohsha-msft (approved):
I don't see any major functional issues with the PR. We can go ahead and merge this. But can we please get rid of `_assert` in this PR itself? The only usage of _assert I see, after it has been replaced with require, is that it is getting used in deleting containers and maybe some occurrences here and there.

seankane-msft:
> I don't see any major functional issues with the PR. We can go ahead and merge this. But can we please get rid of `_assert` in this PR itself? The only usage of _assert I see, after it has been replaced with require, is that it is getting used in deleting containers and maybe some occurrences here and there.

There is a new `_assert` in every single test, so this would get rid of the need for two PRs. I just looked it up, there are 420 more tests that would have to be changed. At that point why have two separete PRs if I'm going to touch every testcase in this PR? This is really just a mediating step, they'll be removed in the next PR.

PR #17084, opened 2022-02-16
Add includeResponse policy to pipeline

jhendrixMSFT:
The policy enables the retrieval of the raw HTTP response from an API
call via the context.


sdk/azcore/policy/policy_test.go
```diff
@@ -52,3 +53,13 @@ func TestWithRetryOptions(t *testing.T) {
...
 }
+
+func TestIncludeResponse(t *testing.T) {
+	ctx := IncludeResponse(context.Background())
+	require.NotNil(t, ctx)
+	raw := ctx.Value(shared.CtxIncludeResponseKey{})
+	_, ok := raw.(**http.Response)
+	if !ok {
+		t.Fatalf("unexpected type %T", raw)
+	}
```
seankane-msft, line 64: I believe you can use require here too:
```suggestion
        require.Truef(t, ok, "unexpected type %T", raw)
```

seankane-msft (approved):

sdk/azcore/policy/policy_test.go
jhendrixMSFT: Good idea, fixed.

jhendrixMSFT: I was hoping that there was a "required type" check in testify but I couldn't find it.

sdk/azcore/policy/policy.go
```diff
@@ -126,3 +126,10 @@ func WithHTTPHeader(parent context.Context, header http.Header) context.Context
 func WithRetryOptions(parent context.Context, options RetryOptions) context.Context {
 	return context.WithValue(parent, shared.CtxWithRetryOptionsKey{}, options)
 }
+
+// IncludeResponse applies the HTTP response retrieval annotation to the parent context.
+// Call runtime.ResponseFromContext() to retrieve the HTTP response from the context.
+func IncludeResponse(parent context.Context) context.Context {
```
jhendrixMSFT, line 132: Note that I changed the name from `WithRawResponse` to `IncludeResponse`.  I felt that using the `With` prefix was misleading in cases where we're enabling a behavior (I also removed `Raw` to shorten the name).  Maybe this should be `EnableResponse`?

sdk/azcore/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Features Added
 * Added header `WWW-Authenticate` to the default allow-list of headers for logging.
+* Added a pipeline policy that enables the retrieval of HTTP responses from API calls.
+  * Added `policy.IncludeResponse` to enable the policy at the API level (off by default).
+  * Added `runtime.ResponseFromContext` to retreive the HTTP response from the context when enabled.
```
jhendrixMSFT, line 9: Are we ok with having the functionality split across two packages?  I feel like `ResponseFromContext` makes sense in the `runtime` package.

sdk/azcore/policy/policy_test.go
jhendrixMSFT: OK there is `require.IsType` but that's a little awkward for this case.

sdk/azcore/CHANGELOG.md
chlowell: I think `IncludeResponse` could go in `runtime` as well because it configures runtime behavior. The fact that an unexported policy implements the behavior is an implementation detail.

jhendrixMSFT: I think you're right.  This also makes me thing that the other WithXxx funcs should go in `runtime` too.  What do you think?

sdk/azcore/policy/policy.go
chlowell: I actually prefer `Include`, `Add` or even `With` because this func causes the policy to augment the context with something it wouldn't otherwise have.

jhendrixMSFT: Will keep it as `Include`.

sdk/azcore/runtime/policy_include_response_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRawResponsePolicy(t *testing.T) {
```
chlowell, line 19:
```suggestion
func TestIncludeResponsePolicy(t *testing.T) {
```

sdk/azcore/runtime/policy_include_response_test.go
```diff
@@ -0,0 +1,41 @@
...
+	srv, close := mock.NewServer()
+	defer close()
+	// add a generic HTTP 200 response
+	srv.SetResponse()
```
chlowell, line 23: It would be good to also verify the policy provides the final response after polling or retrying

chlowell (approved):

sdk/azcore/runtime/policy_include_response_test.go
jhendrixMSFT: Fixed.

sdk/azcore/runtime/policy_include_response_test.go
jhendrixMSFT: Good idea, fixed.

PR #17086, opened 2022-02-16
Upgrade App Service managed identity version

chlowell:
Upgrading from 2017-09-01 to 2019-08-01 because the service team told us the newer version is available everywhere now. I tested this live in addition to updating the offline tests.

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -31,7 +31,6 @@ const (
 	headerMetadata           = "Metadata"
 	imdsEndpoint             = "http://169.254.169.254/metadata/identity/oauth2/token"
 	msiEndpoint              = "MSI_ENDPOINT"
-	msiSecret                = "MSI_SECRET"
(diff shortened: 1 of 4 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -216,17 +212,7 @@ func (c *managedIdentityClient) createAccessToken(res *http.Response) (*azcore.A
 		if expiresOn, err := strconv.Atoi(v); err == nil {
 			return &azcore.AccessToken{Token: value.Token, ExpiresOn: time.Unix(int64(expiresOn), 0).UTC()}, nil
 		}
-		// this is the case when expires_on is a time string
```
jhendrixMSFT, line 219: The format is now unified I take it?

chlowell: Yes, the former App Service version was the quirky one. All other platforms return Unix time as a string.

jhendrixMSFT (approved):

RickWinter (approved):

PR #17093, opened 2022-02-17
Azure App Configuration SDK

antkmsft:
* Client type, and initialization (NewClient())
* SyncTokenPolicy, UpdateSyncToken()
* Types and constants for ConfigurationSetting and SettingSelector
* AddConfigurationSetting()
* GetConfigurationSetting()
* SetConfigurationSetting()
* DeleteConfigurationSetting()
* SetReadOnly()
* GetRevisions()
* Documenting comments for public entities
* NewClientFromConnectionString() (will also be useful for live tests, as constructing this way does not require Identity dependency; Plus .NET and Java do have this constructor overload as well)
* Tests


sdk/appconfiguration/azappconfiguration/client_api_add.go
```diff
@@ -0,0 +1,28 @@
...
+
+func (c *Client) AddConfigurationSetting(ctx context.Context, setting ConfigurationSetting, options *AddConfigurationSettingOptions) (AddConfigurationSettingResponse, error) {
+	etagAny := azcore.ETagAny
+	resp, err := c.appConfigClient.PutKeyValue(ctx, *setting.Key, setting.toGeneratedPutOptions(nil, &etagAny))
+	if err != nil {
+		return AddConfigurationSettingResponse{}, err
+	}
+
+	return (AddConfigurationSettingResponse)(responseFromGeneratedPut(resp)), nil
+}
```
seankane-msft, line 28: Bit of a nit: we try to keep all client methods in the `client.go` file. The response and struct options also are included in this file. It's simpler to have related functions in the same file.

sdk/appconfiguration/azappconfiguration/client_api_get.go
```diff
@@ -0,0 +1,48 @@
...
+		Label:          cs.Label,
+	}
+}
+
+func (c *Client) GetConfigurationSetting(ctx context.Context, setting ConfigurationSetting, options *GetConfigurationSettingOptions) (GetConfigurationSettingResponse, error) {
+	var ifNoneMatch *azcore.ETag
+	var acceptDateTime *time.Time
+	if options != nil {
+		if options.OnlyIfChanged {
+			ifNoneMatch = setting.etag
```
seankane-msft, line 36: is the `setting.etag` expected to be provided by the customer? If so, the etag property needs to be exported.

sdk/appconfiguration/azappconfiguration/client_responses.go
```diff
@@ -0,0 +1,102 @@
...
+	"sdk/appconfiguration/sdk/appconfiguration/azappconfiguration/internal/generated"
+)
+
+type ConfigurationSettingResult struct {
+	ConfigurationSetting
+	ETag      *azcore.ETag
+	SyncToken *string
+}
+
+type configurationSettingResponse struct {
```
seankane-msft, line 24: We should be exporting the response type

sdk/appconfiguration/azappconfiguration/client_responses.go
```diff
@@ -0,0 +1,102 @@
...
+
+type GetConfigurationSettingResult struct {
+	ConfigurationSettingResult
+	LastModified *time.Time
+}
+
+type GetConfigurationSettingResponse struct {
+	Result      GetConfigurationSettingResult
+	RawResponse *http.Response
+}
```
seankane-msft, line 81: I think we can merge these two types. It will reduce the number of exported models and make it more friendly for customers. Thoughts @jhendrixMSFT ?

sdk/appconfiguration/azappconfiguration/configuration_setting.go
```diff
@@ -0,0 +1,69 @@
...
+type ConfigurationSetting struct {
+	Key         *string
+	Value       *string
+	Label       *string
+	ContentType *string
+
+	etag         *azcore.ETag
+	tags         map[string]*string
+	lastModified *time.Time
+	isReadOnly   *bool
```
seankane-msft, line 26: These should be exported as well. A user might want to inspect etag, lastModified, and isReadOnly. Tags can be edited by the user directly and changed on a submit.

sdk/appconfiguration/azappconfiguration/internal/sync_token.go
```diff
@@ -0,0 +1,126 @@
...
+
+type SyncTokenPolicy struct {
+	syncTokens map[string]syncToken
+}
+
+func NewSyncTokenPolicy() *SyncTokenPolicy {
+	return &SyncTokenPolicy{}
+}
+
+func parseToken(tok string) (syncToken, error) {
```
seankane-msft, line 32: For reference, here is the Python implementation of this: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_sync_token.py#L44-L54

seankane-msft: I think you can simplify this by first using `strings.Split(s, ";")` and `strings.Split(s, "=")` on the semicolon separated portions.

seankane-msft (changes requested):
You'll also need a `ci.yml` file, which you can copy from any existing data-plane library. You can copy a `test-resources.json` from one of the tier-1 languages.

sdk/appconfiguration/azappconfiguration/client_api_add.go
```diff
@@ -0,0 +1,28 @@
...
+
+package azappconfiguration
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+type AddConfigurationSettingResponse configurationSettingResponse
```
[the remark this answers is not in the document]
jhendrixMSFT: We really need a CI check that flags all exported symbols that are missing doc comments.

sdk/appconfiguration/azappconfiguration/client_responses.go
antkmsft: It is private, but is made public by
```
type AddConfigurationSettingResponse configurationSettingResponse
type DeleteConfigurationSettingResponse configurationSettingResponse
type SetConfigurationSettingResponse configurationSettingResponse
type SetReadOnlySettingResponse configurationSettingResponse // soon
```

sdk/appconfiguration/azappconfiguration/configuration_setting.go
antkmsft: There are getters below - `ETag()`, `Tags()`, `LastModified()`, and `IsReadOnly()`. But I see that some  SDK do allow them to be modified, so I guess there would be no harm in turning this to a struct with all public members. C# only has 4 out of 8 modifiable, while in Java SDK you can set 7 out of 8 (`setReadOnly` is private for customers).

sdk/appconfiguration/azappconfiguration/internal/sync_token.go
antkmsft: Thank you! I thought something like that must be possible, I'll update it.
(What I currently have here is word for word what C# SDK has https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/SyncTokenUtils.cs, but I don't see why `Split()` can't be used to a large extent - thanks!)

sdk/appconfiguration/azappconfiguration/configuration_setting.go
seankane-msft: Is the configuration setting a model is input and output? (ie returned by the SDK and a parameter to the SDK)

sdk/appconfiguration/azappconfiguration/client_responses.go
antkmsft: My thought process is that we have specialized types for a response for each API (so that in the future if only one of them adds an extra property, we modify just one, and don't make a breaking change. So, users see we have "separate" types, but internally we don't copy&paste the same struct four times.

sdk/appconfiguration/azappconfiguration/configuration_setting.go
antkmsft: Yes, it it input and output. I guess one other advantage to use the same type, is that you can invoke Get(), and then Set() with the same variable. Other SDKs are also reusing the same abstraction for inputs and outputs. Plus, they have overloads, which allows them to have overload methods like `SetConfigurationSetting(key, value, label = default, options = default)` (in addition to ones that take ConfigurationSetting). We don't have overloads, so I think it is important to make interactions with `ConfigurationSetting` type syntactiaclly light, and struct would give us that - `SetConfigurationSetting(context, ConfigurationSetting{ Key: "name", Value: "value", Label: "label" }, null)`

sdk/appconfiguration/azappconfiguration/client_responses.go
seankane-msft: That's a fine reasoning, I don't know if that's the most idiomatic way of doing it. @jhendrixMSFT what are your thoughts?

seankane-msft: Also, how does this show up on intellisense?

seankane-msft (commented):
Samples/tests would help me understand how this is used as well.

antkmsft:
@seankane-msft , samples and public method documentation is in progress. They will be published in this PR when or before it switches from being a Draft.

sdk/appconfiguration/azappconfiguration/client_responses.go
antkmsft: A view from another package:
image

image

image

image

sdk/appconfiguration/azappconfiguration/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+stages:
+- template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
+  parameters:
+    ServiceDirectory: 'appconfiguration/azappconfiguration'
+    RunLiveTests: true
+    AdditionalMatrixConfigs:
+      - Name: keyvault_test_matrix_addons
+        Path: sdk/appconfiguration/azappconfiguration/platform-matrix.json
+        Selection: sparse
+        GenerateVMJobs: true
```
seankane-msft, line 33: Are these necessary?

sdk/appconfiguration/azappconfiguration/client_api_readonly.go
```diff
@@ -0,0 +1,65 @@
...
+		IfMatch: (*string)(ifMatch),
+		Label:   cs.label,
+	}
+}
+
+// SetReadOnly sets an existing configuration setting to read only or read write state in the configuration store.
+func (c *Client) SetReadOnly(ctx context.Context, setting ConfigurationSetting, isReadOnly bool, options *SetReadOnlyOptions) (SetReadOnlyResponse, error) {
+	var ifMatch *azcore.ETag
+	if options != nil && options.OnlyIfUnchanged {
+		ifMatch = setting.etag
```
seankane-msft, line 46: Should etag be exported? We would want a user to do an unconditional update and remove the etag from a `ConfigurationSetting`

sdk/appconfiguration/azappconfiguration/client_api_revisions.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/appconfiguration/azappconfiguration/internal/generated"
+)
+
+// GetRevisionsPager is a Pager for revision list operations.
+//
+// NextPage should be called first. It fetches the next available page of results from the service.
+// If the fetched page contains results, the return value is true, else false.
+// Results fetched from the service can be evaluated by calling PageResponse on this Pager.
+// If the result is false, the value of Err() will indicate if an error occurred.
+type GetRevisionsPager interface {
```
seankane-msft, line 21: Should be ListRevisionsPager, also the new guidance is to return a struct not an interface (it might not be updated in the guidelines).

sdk/appconfiguration/azappconfiguration/client_api_revisions.go
```diff
@@ -0,0 +1,56 @@
...
+func (p *getRevisionsPager) NextPage(ctx context.Context) bool {
+	return p.genPager.NextPage(ctx)
+}
+
+// GetRevisionsOptions contains the optional parameters for the GetRevisions method.
+type GetRevisionsOptions struct {
+}
+
+// GetRevisions retrieves the revisions of one or more configuration setting entities that match the specified setting selector.
+func (c *Client) GetRevisions(selector SettingSelector, options *GetRevisionsOptions) GetRevisionsPager {
```
seankane-msft, line 53: `ListRevisions`

seankane-msft: are selectors required? If not, can we combine them into the options?

sdk/appconfiguration/azappconfiguration/client_responses.go
```diff
@@ -0,0 +1,152 @@
...
+			LastModified: t,
+		},
+		RawResponse: g.RawResponse,
+	}
+}
+
+// ConfigurationSettingResult contains the configuration setting returned from Azure App Configuration client methods.
+type GetRevisionsResult struct {
+	// Contains the configuration settings returned that match the setting selector provided.
+	Items []ConfigurationSetting
```
seankane-msft, line 122:
```suggestion
        ConfigurationSettings []ConfigurationSetting
```

sdk/appconfiguration/azappconfiguration/policy_sync_token_parse_test.go
```diff
@@ -0,0 +1,223 @@
...
+	{
+		tok1, err1 := parseToken("n=v;sn=1")
+		require.Empty(t, err1)
+		require.NotEmpty(t, tok1)
+		require.Equal(t, tok1.id, "n")
+		require.Equal(t, tok1.value, "v")
+		require.Equal(t, tok1.seqNo, 1)
+	}
+
+	{
```
seankane-msft, line 50: The blocks separated by `{` and `}` aren't necessary here, feels like this might be more a c/c++ thing.

sdk/appconfiguration/azappconfiguration/client_api_add.go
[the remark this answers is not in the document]
antkmsft: There should be comments for all public entities now

sdk/appconfiguration/azappconfiguration/client_api_add.go
antkmsft: I will keep them separated if you don't mind, I find it more convenient to at least  develop in this way - everything related to one method is in one file, and nothing else. Since there aren't that many other entities, it does not clutter the files list.

sdk/appconfiguration/azappconfiguration/client_api_get.go
antkmsft: I made it all public per Jeffrey's suggestion from API review session we've held.

sdk/appconfiguration/azappconfiguration/client_responses.go
antkmsft: I think the screenshots look good.

sdk/appconfiguration/azappconfiguration/internal/sync_token.go
antkmsft: I am going to leave it like this for beta, if you don't mind. It did not take me as much of a mental resource to translate this code from the other language SDK, but in order to make sure it would always produce the same result using potentially more effective methods, I will need to think more deeply for a bit more. But I was considering your suggestion, just decided to not do it right now.

sdk/appconfiguration/azappconfiguration/client_api_revisions.go
antkmsft: Could go either way, but other language SDKs have them. Potentially, it can be used in two methods.

sdk/appconfiguration/azappconfiguration/policy_sync_token_parse_test.go
antkmsft: I'd say it is not very common in C/C++ as well, it's just me :) But I removed them.

sdk/appconfiguration/azappconfiguration/client_responses.go
antkmsft: Jeffrey has told me that we don't need to have the `RawResponse` anymore (there should be ways to get the response some other way), so I merged Response into Type, but I'll check with @jhendrixMSFT

sdk/appconfiguration/azappconfiguration/client_api_readonly.go
antkmsft: That's the logic from other SDKs: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L965

sdk/appconfiguration/azappconfiguration/internal/sync_token.go
antkmsft: Now that there was unit test coverage, I rewrote it to use string split.

sdk/appconfiguration/azappconfiguration/client_api_add.go
antkmsft: I have moved all methods into a single file now.

antkmsft:
@seankane-msft, ping! Do you still see any blockers?
I have reviewed API with Jeffrey and took his feedback. He said it looks good.
Tests are passing, all the service methods do work.

sdk/appconfiguration/azappconfig/client.go
```diff
@@ -0,0 +1,443 @@
...
+
+	pl := runtime.NewPipeline(generated.ModuleName, generated.ModuleVersion, runtime.PipelineOptions{}, genOptions)
+	return Client{
+		appConfigClient: generated.NewAzureAppConfigurationClient(endpointUrl, nil, pl),
+		syncTokenPolicy: syncTokenPolicy,
+	}, nil
+}
+
+// NewClientFromConnectionString parses the connection string and returns a pointer to a Client object.
+func NewClientFromConnectionString(connectionString string, options *ClientOptions) (Client, error) {
```
seankane-msft, line 86: Clients should be returned by reference: https://azure.github.io/azure-sdk/golang_introduction.html#service-client-constructors

sdk/appconfiguration/azappconfig/client.go
```diff
@@ -0,0 +1,443 @@
...
+
+func fromGeneratedAdd(g generated.AzureAppConfigurationClientPutKeyValueResponse) AddSettingResult {
+	return AddSettingResult{
+		Setting:   settingFromGenerated(g.KeyValue),
+		SyncToken: g.SyncToken,
+	}
+}
+
+// AddSettingOptions contains the optional parameters for the AddSetting method.
+type AddSettingOptions struct {
```
seankane-msft, line 142: Should have a comment like [this](https://github.com/Azure/autorest.go/blob/track2/test/autorest/complexgroup/zz_generated_response_types.go#L28)

sdk/appconfiguration/azappconfig/client.go
```diff
@@ -0,0 +1,443 @@
...
+
+func (cs Setting) toGeneratedDeleteOptions(ifMatch *azcore.ETag) *generated.AzureAppConfigurationClientDeleteKeyValueOptions {
+	return &generated.AzureAppConfigurationClientDeleteKeyValueOptions{
+		IfMatch: (*string)(ifMatch),
+		Label:   cs.Label,
+	}
+}
+
+// DeleteSetting deletes a configuration setting from the configuration store.
+func (c *Client) DeleteSetting(ctx context.Context, setting Setting, options *DeleteSettingOptions) (DeleteSettingResult, error) {
```
seankane-msft, line 187: Why are we returning `***Result`? I thought the guideline was `***Response`, I'm assuming Jeff okayed this but curious why the difference.

sdk/appconfiguration/azappconfig/client.go
```diff
@@ -0,0 +1,443 @@
...
+	return p.genPager.NextPage(ctx)
+}
+
+// ListRevisionsOptions contains the optional parameters for the ListRevisions method.
+type ListRevisionsOptions struct {
+}
+
+// NewListRevisionsPager retrieves the revisions of one or more configuration setting entities that match the specified setting selector.
+func (c *Client) NewListRevisionsPager(selector SettingSelector, options *ListRevisionsOptions) ListRevisionsPager {
+	_ = options
```
seankane-msft, line 441: Do you need this for the linter? I think you can ignore it and you should be okay, just a nit.

sdk/appconfiguration/azappconfig/setting.go
```diff
@@ -0,0 +1,73 @@
...
+	// The content type of the configuration setting's value.
+	// Providing a proper content-type can enable transformations of values when they are retrieved by applications.
+	ContentType *string
+
+	// An ETag indicating the state of a configuration setting within a configuration store.
+	ETag *azcore.ETag
+
+	// A dictionary of tags used to assign additional properties to a configuration setting.
+	// These can be used to indicate how a configuration setting may be applied.
+	Tags map[string]*string
```
seankane-msft, line 39: In keyvault and tables we change this to `map[string]string` on the Client side. I'm not sure there is a use case of a nil value.

seankane-msft: `map[string]string` is also easier for customers to use

seankane-msft (approved):
Approving, I don't think any of these are blocking but a few comments.

sdk/appconfiguration/azappconfig/client.go
jhendrixMSFT: Good point.  The pattern is to return `FooResponse`.

sdk/appconfiguration/azappconfig/client.go
jhendrixMSFT: I believe we reverted back to naming the pageable operations just `ListFoo`.

sdk/appconfiguration/azappconfig/client.go
antkmsft: I can change once again before checking in, but in the API view you wrote 3 days ago that new guideline is to return Client by value not reference.
https://apiview.dev/Assemblies/Review/2194037c3c534536b31edad97ed193ec

sdk/appconfiguration/azappconfig/client.go
antkmsft: Jeff told me that there will be a new way of getting raw responses, it is no longer needed in methods.

sdk/appconfiguration/azappconfig/policy_hmac_auth_parse_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	require.Len(t, sc, 3)
+	require.Equal(t, byte('f'), sc[0])
+	require.Equal(t, byte('f'), sc[1])
+	require.Equal(t, byte('f'), sc[2])
+}
+
+func TestHmacAuthParseConnectionStringMissingEndoint(t *testing.T) {
+	_, _, _, err := parseConnectionString("Id=yY;Secret=ZmZm")
+	require.NotEmpty(t, err)
```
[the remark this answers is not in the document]
antkmsft: In this case, as long as `err` is not empty, implementation may return anything in other return values, tests will be fine with either.

sdk/appconfiguration/azappconfig/client.go
seankane-msft: Yes , I was informed that is not our standard. I had to make this change myself

sdk/appconfiguration/azappconfig/client.go
seankane-msft: The change we are referring to is `DeleteSettingResult` should be `DeleteSettingResponse`

sdk/appconfiguration/azappconfig/policy_hmac_auth_parse_test.go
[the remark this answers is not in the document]
seankane-msft: I mistyped, I meant to point `require.NoError` (and `require.Error` for when you are expecting an error).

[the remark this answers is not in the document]
antkmsft: Thanks!

antkmsft:
Thank you all, especially Sean, so much for the feedback, I have applied it all, including the last portion of comments.

PR #17097, opened 2022-02-17
[KeyVault] adding a comment about nil options

seankane-msft:
Improve doc comments

### DIFF
--- a/sdk/keyvault/azkeys/client.go
+++ b/sdk/keyvault/azkeys/client.go
@@ -194,7 +194,7 @@ func createECKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyRespons
 
 // CreateECKey can be used to create a new elliptic key curve in Azure Key Vault. If the
 // named key already exists, Azure Key Vault creates a new version of the key.
-// EC key can be created on standard vaults, ECHSM can be created on Premium SKU Vaults or Managed HSMs.
+// EC keys can be created in Standard or Premium SKU vaults, ECHSM can be created in Premium SKU vaults or Managed HSMs.
 // It requires the keys/create permission. Pass nil to use the default options.
 func (c *Client) CreateECKey(ctx context.Context, name string, options *CreateECKeyOptions) (CreateECKeyResponse, error) {
 	keyType := EC
--- a/sdk/keyvault/azkeys/client.go
+++ b/sdk/keyvault/azkeys/client.go
@@ -193,8 +193,9 @@ func createECKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyRespons
 }
 
 // CreateECKey can be used to create a new elliptic key curve in Azure Key Vault. If the
-// named key already exists, Azure Key Vault creates a new version of the key. It requires
-// the keys/create permission. Pass nil to use the default options.
+// named key already exists, Azure Key Vault creates a new version of the key.
+// EC key can be created on standard vaults, ECHSM can be created on Premium SKU Vaults or Managed HSMs.
+// It requires the keys/create permission. Pass nil to use the default options.
 func (c *Client) CreateECKey(ctx context.Context, name string, options *CreateECKeyOptions) (CreateECKeyResponse, error) {
 	keyType := EC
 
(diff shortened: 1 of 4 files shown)


sdk/keyvault/azkeys/client.go
```diff
@@ -136,9 +136,9 @@ func createKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyResponse)
 	}
 }
 
-// CreateKey - The create key operation can be used to create any key type in Azure Key Vault.
-// If the named key already exists, Azure Key Vault creates
-// a new version of the key. It requires the keys/create  permission.
+// CreateKey can be used to create any key type in Azure Key Vault.  If the named key already exists,
+// Azure Key Vault creates a new version of the key. It requires the keys/create permission. Pass in
```
JeffreyRichter, line 140: How about just: "Default options are used if nil is passed."

sdk/keyvault/azkeys/client.go
```diff
@@ -254,9 +254,9 @@ func createOCTKeyResponseFromGenerated(i generated.KeyVaultClientCreateKeyRespon
 	}
 }
 
-// CreateKey - The create key operation can be used to create a new octet sequence (symmetric) key in Azure Key Vault.
-// If the named key already exists, Azure Key Vault creates
-// a new version of the key. It requires the keys/create permission.
+// CreateOCTKey can be used to create a new octet sequence (symmetric) key in Azure Key Vault.
```
kimprice, line 257: Just wondering if this function should be CreateOctKey instead to align with the official naming?  And if there is only an oct-HSM (no oct), should the default options set HardwareProtected to true? https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys

Also, should the doc comment include info that the oct-HSM key requires a managed HSM rather than a vault?
Similarly, should we include doc comments for the other key types to indicate vault compatibility or would the service provide a useful error message in those cases where the vault or managed HSM does not support the key type?

sdk/keyvault/azkeys/client.go
kimprice: Or "Pass nil to use the default options"

sdk/keyvault/azkeys/client.go
```diff
@@ -1065,6 +1072,7 @@ func (l *ListDeletedKeysOptions) toGenerated() *generated.KeyVaultClientGetDelet
 
 // ListDeletedKeys lists all versions of the specified key. The full key identifier and attributes are provided
```
kimprice, line 1073: "specified key" sounds like there should be a key name/id parameter. If this function lists all the deleted keys in the vault, then I suggest something like the following:

```suggestion
// ListDeletedKeys lists all versions of the deleted keys. The full key identifier and attributes are provided
```

kimprice: Actually, this looks like it may have been a copy/paste from ListKeyVersions so perhaps this comment needs to be updated completely.

sdk/keyvault/azkeys/client.go
seankane-msft: Good catch, Changed everything from OCT -> Oct

kimprice: To follow up on the vault compatibility - is that something we should specify in the doc comments or do we rely on the service to provide an error? For some of the other functions we do seem to specify compatibility. For example GetDeletedKey mentions that it only works for soft-delete enabled vaults.

sdk/keyvault/azkeys/client.go
```diff
@@ -340,9 +341,10 @@ func createRSAKeyResponseFromGenerated(i generated.KeyVaultClientCreateKeyRespon
 	}
 }
 
-// CreateKey - The create key operation can be used to create a new RSA key in Azure Key Vault.
-// If the named key already exists, Azure Key Vault creates
-// a new version of the key. It requires the keys/create  permission.
+// CreateRSAKey can be used to create a new RSA key in Azure Key Vault. If the named key already
+// exists, Azure Key Vault creates a new version of the key. // RSA key can be created on standard
```
kimprice:
```suggestion
// exists, Azure Key Vault creates a new version of the key. // RSA keys can be created on Standard or
```

sdk/keyvault/azkeys/client.go
```diff
@@ -340,9 +341,10 @@ func createRSAKeyResponseFromGenerated(i generated.KeyVaultClientCreateKeyRespon
...
+// vaults, RSAHSM can be created on Premium SKU Vaults or Managed HSMs.
```
kimprice, line 346:
```suggestion
// Premium SKU vaults, RSAHSM can be created in Premium SKU vaults or Managed HSMs.
```

sdk/keyvault/azkeys/client.go
seankane-msft: oct and oct-HSM are treated the same by the service. If the user is targeting a MHSM and they send a request to create a `oct` key, the service will create an `oct-HSM` key. That is the reason we have both types here.

PR #17099, opened 2022-02-18
azidentity: environmnet credential add supports for the cert password

magodo:
The PR adds a new env var `AZURE_CLIENT_CERTIFICATE_PASSWORD` to allow users to set the certificate password for the `EnvironmentCredential`.


chlowell:
Thanks for doing this! I think we can add this but probably not in the next version. I need to discuss it with the rest of the team (we would add it to the other Azure SDKs as well) and our architecture review board first. In the meantime, could you add a test?

magodo:
@chlowell Seems only the integration test need to be added, while it needs access to the CI setup, i.e. a new env var something like the `PFX_CONTENTS_ENCRYPT`?

chlowell:
To update on the timeline, I'm aiming for this in the next beta release, which will go out in early June.

chlowell (approved):
LGTM for the next beta, thanks again for doing this

PR #17146, opened 2022-02-22
[Tables] GA Preparation

seankane-msft:
Preparing Tables for GA

### DIFF
--- a/sdk/data/aztables/CHANGELOG.md
+++ b/sdk/data/aztables/CHANGELOG.md
@@ -5,6 +5,12 @@
 ### Features Added
 
 ### Breaking Changes
+* Prefixed all `TransactionType` constants with `TransactionType`.
+* Prefixed all `EntityUpdateMode` constants with `EntityUpdateMode`.
+* Changed the `SharedKeyCredential.ComputeHMACSHA256` method to a private method.
+* Changed the `ListTablesPager` and `ListEntitiesPager` to structs.
+* Renamed the `ResponseProperties` type to `TableProperties`.
+* Returns `Client` and `ServiceClient` by value instead of by reference.
 
 ### Bugs Fixed
 
--- a/sdk/data/aztables/CHANGELOG.md
+++ b/sdk/data/aztables/CHANGELOG.md
@@ -11,6 +11,9 @@
 * Changed the `ListTablesPager` and `ListEntitiesPager` to structs.
 * Renamed the `ResponseProperties` type to `TableProperties`.
 * Removing `ContentType` from the `TransactionResponse` struct.
+* Update `ListEntitiesPager` and `ListTablesPager`.
+    * The `More` method checks whether there are more pages to retrieve.
+    * The `NextPage(context.Context)` method gets the next page and returns a response and an `error`.
 
 ### Bugs Fixed
 
--- a/sdk/data/aztables/CHANGELOG.md
+++ b/sdk/data/aztables/CHANGELOG.md
@@ -10,7 +10,6 @@
 * Changed the `SharedKeyCredential.ComputeHMACSHA256` method to a private method.
 * Changed the `ListTablesPager` and `ListEntitiesPager` to structs.
 * Renamed the `ResponseProperties` type to `TableProperties`.
-* Returns `Client` and `ServiceClient` by value instead of by reference.
 * Removing `ContentType` from the `TransactionResponse` struct.
 
 ### Bugs Fixed
--- a/sdk/data/aztables/CHANGELOG.md
+++ b/sdk/data/aztables/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Changed the `ListTablesPager` and `ListEntitiesPager` to structs.
 * Renamed the `ResponseProperties` type to `TableProperties`.
 * Returns `Client` and `ServiceClient` by value instead of by reference.
+* Removing `ContentType` from the `TransactionResponse` struct.
 
 ### Bugs Fixed
 
--- a/sdk/data/aztables/client.go
+++ b/sdk/data/aztables/client.go
@@ -302,6 +302,7 @@ func (t *Client) List(listOptions *ListEntitiesOptions) ListEntitiesPager {
 
 // Options for Client.GetEntity method
 type GetEntityOptions struct {
+	// placeholder for future optional parameters
 }
 
 func (g *GetEntityOptions) toGenerated() (*generated.TableClientQueryEntityWithPartitionAndRowKeyOptions, *generated.QueryOptions) {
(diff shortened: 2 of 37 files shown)


sdk/data/aztables/CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
 ### Features Added
 
 ### Breaking Changes
+* Prefixed all `TransactionType` constants with `TransactionType`.
+* Prefixed all `EntityUpdateMode` constants with `EntityUpdateMode`.
+* Changed the `SharedKeyCredential.ComputeHMACSHA256` method to a private method.
+* Changed the `ListTablesPager` and `ListEntitiesPager` to structs.
+* Renamed the `ResponseProperties` type to `TableProperties`.
+* Returns `Client` and `ServiceClient` by value instead of by reference.
```
chlowell, line 13: Why? Client constructors should return pointers.

seankane-msft: Fixed

seankane-msft: Mis-remembered the guidelines

sdk/data/aztables/transactional_batch.go
```diff
@@ -219,14 +225,14 @@ func buildTransactionResponse(req *policy.Request, resp *http.Response, itemCoun
 			if err != nil {
 				return &TransactionResponse{}, err
 			} else {
-				innerResponses = []http.Response{*r}
+				// innerResponses = []*http.Response{r}
```
jhendrixMSFT, line 228: Does this need to go away?

sdk/data/aztables/service_client.go
```diff
@@ -106,15 +107,15 @@ func (t *ServiceClient) CreateTable(ctx context.Context, name string, options *C
 }
 
 // Options for Client.Delete and ServiceClient.DeleteTable methods
-type DeleteTableOptions struct {
-}
+type DeleteTableOptions struct{}
```
jhendrixMSFT, line 110: In generated code, we add a comment indicating this is for future expansion.  See [example](https://github.com/Azure/autorest.go/blob/track2/test/autorest/additionalpropsgroup/zz_generated_models.go#L96-L98)

sdk/data/aztables/service_client.go
```diff
@@ -189,87 +173,109 @@ type ListTablesPage struct {
...
+	nextTableName     *string
 }
 
 // NextPage fetches the next available page of results from the service.
 // If the fetched page contains results, the return value is true, else false.
 // Results fetched from the service can be evaulated by calling PageResponse on this Pager.
-func (p *tableQueryResponsePager) NextPage(ctx context.Context) bool {
-	if p.err != nil || (p.current != nil && p.current.XMSContinuationNextTableName == nil) {
-		return false
+func (p *ListTablesPager) NextPage(ctx context.Context) (ListTablesPage, error) {
```
jhendrixMSFT, line 242: I think the return type should be named `ListTablesPageResponse` like our other response types.

sdk/data/aztables/service_client.go
```diff
@@ -189,87 +173,109 @@ type ListTablesPage struct {
...
+		}
+	}
+	return true
 }
 
-// Err returns an error value if the most recent call to NextPage was not successful, else nil.
-func (p *tableQueryResponsePager) Err() error {
-	return p.err
+// NextPageTableName returns the continuation token for the ListTablesPager
+func (p *ListTablesPager) NextPageTableName() *string {
```
jhendrixMSFT, line 277: Is this necessary?  Generally speaking, if we need to expose the continuation token from a pager, we should have a unified name for the API, maybe `ContinuationToken() string` (I assume it doesn't need to be a pointer-to-string as an empty string would mean no more pages).

sdk/data/aztables/client.go
```diff
@@ -213,76 +220,57 @@ func newListEntitiesPage(resp *generated.TableClientQueryEntitiesResponse) (List
...
 
-type tableEntityQueryResponsePager struct {
-	tableClient       *Client
-	current           *ListEntitiesPage
-	tableQueryOptions *generated.TableClientQueryEntitiesOptions
-	listOptions       *ListEntitiesOptions
-	err               error
+// NextPagePartitionKey returns one of the continuation tokens for the ListEntitiesPager.
+// Use in conjunction with the NewPageRowKey.
+func (p *ListEntitiesPager) NextPagePartitionKey() *string {
```
jhendrixMSFT, line 240: Does this need to be pointer-to-string?  Same question for `NextPageRowKey()`

sdk/data/aztables/service_client.go
seankane-msft: Another oddity for Tables, paging entities has two continuation tokens, we can return a `(string, string)` but Jeff and I decided to go with two different methods to retrieve them. I then kept the pattern the same for the Tables pager, even though it only has a single CT.

sdk/data/aztables/transactional_batch.go
seankane-msft: Leaving for now

sdk/data/aztables/service_client.go
seankane-msft: Added

sdk/data/aztables/client.go
seankane-msft: When the pager is returned from `Client.List`, there are no continuation tokens. How should we return in this edge case? I tested with an empty string (`""`), but this will not page correctly. I think a string is the correct return type here, but when they are nil should I return an empty string?

sdk/data/aztables/service_client.go
seankane-msft: Fixed

sdk/data/aztables/transactional_batch.go
```diff
@@ -219,7 +225,7 @@ func buildTransactionResponse(req *policy.Request, resp *http.Response, itemCoun
 			if err != nil {
 				return &TransactionResponse{}, err
 			} else {
-				innerResponses = []http.Response{*r}
+				// innerResponses = []http.Response{*r}
```
jhendrixMSFT, line 228: Nit, this can be removed.

sdk/data/aztables/service_client.go
jhendrixMSFT: I see.  Are the CTs exposed in the response though?  If not, then I assume that's why they're methods on the pager.

sdk/data/aztables/client.go
jhendrixMSFT: Looks like this has been resolved?

jhendrixMSFT (approved):

sdk/data/aztables/service_client.go
seankane-msft: They are exposed in the response, we could remove these methods and just expose them from the responses.

seankane-msft: I'm going to remove the methods exposing continuation tokens, they are available in the response structs which is more intuitive.

PR #17156, opened 2022-02-24
Refine HTTP response captuing APIs

jhendrixMSFT:
Renamed IncludeResponse to WithCaptureResponse and changed its return
values to return a func used to retrieve the response.  This clarifies
the contract of the API and how to use it.
Removed ResponseFromContext as it's no longer required.

### DIFF
--- a/sdk/azcore/runtime/policy_include_response.go
+++ b/sdk/azcore/runtime/policy_include_response.go
@@ -27,19 +27,13 @@ func includeResponsePolicy(req *policy.Request) (*http.Response, error) {
 	return resp, err
 }
 
-// IncludeResponse applies the HTTP response retrieval annotation to the parent context.
-// Call runtime.ResponseFromContext() to retrieve the HTTP response from the context.
-func IncludeResponse(parent context.Context) context.Context {
-	var rawResp *http.Response
-	return context.WithValue(parent, shared.CtxIncludeResponseKey{}, &rawResp)
-}
-
-// ResponseFromContext retrieves the raw HTTP response from the specified context.
-// Disabled by default.  Use policy.IncludeResponse() to enable on the calling context.
-func ResponseFromContext(ctx context.Context) *http.Response {
-	if httpOutRaw := ctx.Value(shared.CtxIncludeResponseKey{}); httpOutRaw != nil {
-		httpOut := httpOutRaw.(**http.Response)
-		return *httpOut
+// WithCaptureResponse applies the HTTP response retrieval annotation to the parent context.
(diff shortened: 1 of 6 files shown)


sdk/azcore/runtime/policy_include_response.go
```diff
@@ -27,19 +27,13 @@ func includeResponsePolicy(req *policy.Request) (*http.Response, error) {
...
-		httpOut := httpOutRaw.(**http.Response)
-		return *httpOut
+// WithCaptureResponse applies the HTTP response retrieval annotation to the parent context.
+// Call the returned function to retrieve the HTTP response after the request has completed.
+// The created context is NOT safe to use across multiple goroutines.
+func WithCaptureResponse(parent context.Context) (context.Context, func() *http.Response) {
+	var resp *http.Response
+	return context.WithValue(parent, shared.CtxIncludeResponseKey{}, &resp), func() *http.Response {
+		cap := &resp
+		return *cap
```
JeffreyRichter, line 37: Can't you just return resp?
And, you don't actually need to pass &resp to WithValue.

JeffreyRichter (approved):

sdk/azcore/runtime/policy_include_response.go
jhendrixMSFT: We don't have the response at this point, all we're doing here is allocating the storage for it.  The value is populated before the policy exits.  This also means that the implementation isn't safe to share across goroutines (I added a note about this in the doc comment).  I did some thinking on how to resolve this but couldn't come up with a solution.

seankane-msft (approved):
Do we have sample documentation/examples on how to use this?

jhendrixMSFT:
@richardpark-msft brought up a good point that using an out-parameter like before makes it very clear on the shared nature of the response.  I worry that the current design makes this less clear, unless you read the doc comments.  Thoughts?

jhendrixMSFT:
Yes, it would look like this.
```go
var resp *http.Response
captureCtx := runtime.WithCaptureResponse(parentCtx, &resp)
```
You are correct about the last one winning.  However, this is true to all context values.

JeffreyRichter:
Yes, it's true but they can't tell this because Value() returns the last one. What we're doing that is problematic is that these Context values are supposed to be read-only but we're treating this one as read/write. This is what's causing some consternation (race conditions) here. But we're saying that this is an advanced (not mainstream) scenario and so we're offering an escape hatch mechanism for anyone who might need it and therefore we're willing to make some compromise. Now, it's a quest as to how far can we go to make things work as expected as possible but we will not hit perfection since the *http.Response is read/write.

PR #17170, opened 2022-02-25
Prevent ChainedTokenCredential data race

### LINKED ISSUE #17144: DefaultAzureCredential causes race data condition in goroutines
### Bug Report

Package: github.com/Azure/azure-sdk-for-go/sdk/azidentity
SDK version: latest
Package version: v0.13.1
go version go1.17.6 darwin/arm64

Using DefaultAzureCredential in more than one goroutine at the same time can cause a race data condition to be raised. The problem seems to be the NewChainedTokenCredential overwriting the stored credential from another routine vs. finding the cached one from a previous call on the credential. Since this is deferred until later in execution when the token is needed an actual API call needs to be made before the goroutines can use the credential or the race condition will happen. It would be ideal if the routines would prevent this race condition, use the cached credential or at least provide a way to obtain the credential and cache it on the azidentity API and not have to execute an API action to cause the credential to be obtained before calling other APIs in goroutines just to prevent this race condition.

chlowell:
Fixes #17144 by adding synchronization around `ChainedTokenCredential.successfulCredential`. The field is assigned at most once, so I used a read lock to minimize performance impact on subsequent calls.

### DIFF
--- a/sdk/azidentity/chained_token_credential.go
+++ b/sdk/azidentity/chained_token_credential.go
@@ -75,6 +75,8 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.Token
 			}
 			if !c.iterating {
 				c.iterating = true
+				// allow other goroutines to wait while this one iterates
+				c.cond.L.Unlock()
 				break
 			}
 			c.cond.Wait()
@@ -99,6 +101,7 @@ func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.Token
 		}
 	}
 	if c.iterating {
+		c.cond.L.Lock()
 		c.iterating = false
 		c.cond.L.Unlock()
 		c.cond.Broadcast()
--- a/sdk/azidentity/chained_token_credential.go
+++ b/sdk/azidentity/chained_token_credential.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
@@ -31,6 +32,9 @@ type ChainedTokenCredential struct {
 	successfulCredential azcore.TokenCredential
 	retrySources         bool
 	name                 string
+	cond                 *sync.Cond
+	iterating            bool
+	mut                  *sync.RWMutex
 }
 
 // NewChainedTokenCredential creates a ChainedTokenCredential.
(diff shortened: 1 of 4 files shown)


sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +54,72 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+	}, nil
 }
 
 // GetToken calls GetToken on the chained credentials in turn, stopping when one returns a token. This method is called automatically by Azure SDK clients.
 // ctx: Context controlling the request lifetime.
 // opts: Options for the token request, in particular the desired scope of the access token.
 func (c *ChainedTokenCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
-	if c.successfulCredential != nil && !c.retrySources {
-		return c.successfulCredential.GetToken(ctx, opts)
+	if !c.retrySources {
```
jhendrixMSFT, line 71: Before I dig into this, is it possible to reuse the [`ExpiringResource`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/shared/expiring_resource.go) for this?  Ideally, we'd move this machinery into the `internal` module so it can be used in `azcore` and here.

chlowell: There's probably a way to make it work but it would be awkward. The value here never expires and exclusive locking around it is an unnecessary performance hit.

jhendrixMSFT: If it's going to make it more complicated, then it's not worth it.  The pattern looked similar, so I was just curious if this possibility had been looked at.

sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +54,72 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+		// ensure only one goroutine at a time iterates the sources and perhaps sets c.successfulCredential
+		for {
+			c.mut.RLock()
+			mustIterate := c.successfulCredential == nil
+			c.mut.RUnlock()
+			if !mustIterate {
+				return c.successfulCredential.GetToken(ctx, opts)
+			}
+			c.cond.L.Lock()
+			defer c.cond.L.Unlock()
```
jhendrixMSFT, line 81: I don't think this is correct (it's safe but doesn't take advantage of the condition).  The exclusive lock should be released once the for loop exits so that other goroutines will hit `c.cond.Wait()`.

sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +54,72 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
 	var errs []error
+	var token *azcore.AccessToken
 	for _, cred := range c.sources {
-		token, err := cred.GetToken(ctx, opts)
+		token, err = cred.GetToken(ctx, opts)
 		if err == nil {
 			log.Writef(EventAuthentication, "%s authenticated with %s", c.name, extractCredentialName(cred))
-			c.successfulCredential = cred
-			return token, nil
+			if c.iterating {
```
jhendrixMSFT, line 97: I think you can move `if c.iterating` to wrap around the `for` loop, then you also don't need to check it twice (similar to how we do it in expiring resource).

sdk/azidentity/chained_token_credential.go
chlowell: You're right, the condition isn't effective here. I'll have to give it more thought. At the moment I'm thinking my read locking approach prevents straightforward use of `Cond` because there are really two conditions here, only one of which is checked under an exclusive lock as `Cond` expects. If that's the case, I would rather give up on `Cond` than read locking.

sdk/azidentity/chained_token_credential.go
chlowell: I ended up with this because I don't always want to synchronize the loop. The credential may be configured to iterate its sources every time, in which case there's no reason to access `c.successfulCredential` and `c.iterating` is false.

sdk/azidentity/chained_token_credential.go
```diff
@@ -70,15 +67,12 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
-			c.mut.RUnlock()
-			if !mustIterate {
+			if c.successfulCredential != nil {
+				c.cond.L.Unlock()
 				return c.successfulCredential.GetToken(ctx, opts)
 			}
-			c.cond.L.Lock()
-			defer c.cond.L.Unlock()
 			if !c.iterating {
 				c.iterating = true
```
jhendrixMSFT, line 77: I think you need to `c.cond.L.Unlock()` after this assignment so the remaining goroutines unblock from line 70 and block on `c.cond.Wait()`.

sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +53,68 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+			break
 		}
 		errs = append(errs, err)
 		if _, ok := err.(credentialUnavailableError); !ok {
+			break
+		}
+	}
+	if c.iterating {
+		c.iterating = false
+		c.cond.L.Unlock()
```
jhendrixMSFT, line 103: Unlocking on line 78 would then preclude the need to unlock here.

sdk/azidentity/chained_token_credential.go
chlowell: That would more efficiently get goroutines to the `Wait()` but it reintroduces the race because the iterating goroutine would assign the variable while not holding the lock.

jhendrixMSFT: But all the other goroutines would be waiting yes?

jhendrixMSFT: Hmm I suppose not, if one arrives late.

chlowell: I'm with you in believing it ought to be safe to unlock here because only the goroutine which does the assignment would iterate, while all others wait for the signal. However, unlocking here triggers the race detector, so either it or me isn't smart enough to figure this out 😆

jhendrixMSFT: Ah yes, I get it now.  The lock must be acquired before writing to shared state (just like we do [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/shared/expiring_resource.go#L95)).  I think that after the call to `cred.GetToken()` you need to do this.
```go
if c.iterating {
    c.cond.L.Lock()
}
```
What do you think?

chlowell: Yes, that does the trick. A goroutine can iterate without holding the lock but must acquire the lock afterward.

sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +53,71 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
+	var token *azcore.AccessToken
 	for _, cred := range c.sources {
-		token, err := cred.GetToken(ctx, opts)
+		token, err = cred.GetToken(ctx, opts)
 		if err == nil {
 			log.Writef(EventAuthentication, "%s authenticated with %s", c.name, extractCredentialName(cred))
-			c.successfulCredential = cred
-			return token, nil
+			if c.iterating {
+				c.successfulCredential = cred
```
jhendrixMSFT, line 94: I think the lock needs to be held before the assignment here.

sdk/azidentity/chained_token_credential.go
```diff
@@ -50,35 +53,73 @@ func NewChainedTokenCredential(sources []azcore.TokenCredential, options *Chaine
...
-		token, err := cred.GetToken(ctx, opts)
+		token, err = cred.GetToken(ctx, opts)
 		if err == nil {
 			log.Writef(EventAuthentication, "%s authenticated with %s", c.name, extractCredentialName(cred))
-			c.successfulCredential = cred
-			return token, nil
+			if c.iterating {
+				c.cond.L.Lock()
+				c.successfulCredential = cred
+				c.cond.L.Unlock()
```
jhendrixMSFT, line 96: I think this unlock could be removed, and the lock on line 106.  What do you think?

chlowell: Works for the happy path but a goroutine that exits this loop due to an error doesn't execute this block. Such a goroutine would then go on to assign `c.iterating` and unlock while not holding the lock. This would be easier to reason about and a bit tidier without this block, let's try that.

jhendrixMSFT (approved):
Looks great, thanks for bearing with me.

chlowell:
Thanks for helping me fix this code without making it incomprehensible!

PR #17171, opened 2022-02-25
[azservicebus] Recategorize recovery for errors

richardpark-msft:
Recategorize recovery for errors:
- amqp.ErrorInternalError and amqp.ErrSessionClosed are now categorized as connection level recovery.
- ErrorTransferLimitExceeded is a link recovery event. From what I can see it signals a mismatch between wanting to send and not having credit available. A link reset should cause a reset on both sides.
- Default to connection recovery except for specific fatal errors.

### DIFF
--- a/sdk/messaging/azservicebus/internal/errors.go
+++ b/sdk/messaging/azservicebus/internal/errors.go
@@ -91,17 +91,18 @@ var amqpConditionsToRecoveryKind = map[amqp.ErrorCondition]recoveryKind{
 	amqp.ErrorCondition("com.microsoft:operation-cancelled"): RecoveryKindNone,
 
 	// Link recovery needed
-	amqp.ErrorDetachForced: RecoveryKindLink, // "amqp:link:detach-forced"
+	amqp.ErrorDetachForced:          RecoveryKindLink, // "amqp:link:detach-forced"
+	amqp.ErrorTransferLimitExceeded: RecoveryKindLink,
 
 	// Connection recovery needed
 	amqp.ErrorConnectionForced: RecoveryKindConn, // "amqp:connection:forced"
+	amqp.ErrorInternalError:    RecoveryKindConn, // "amqp:internal-error"
 
 	// No recovery possible - this operation is non retriable.
 	amqp.ErrorMessageSizeExceeded:                                 RecoveryKindFatal,
 	amqp.ErrorUnauthorizedAccess:                                  RecoveryKindFatal, // creds are bad
 	amqp.ErrorNotFound:                                            RecoveryKindFatal,
 	amqp.ErrorNotAllowed:                                          RecoveryKindFatal,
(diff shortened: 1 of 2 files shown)


sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -91,17 +91,18 @@ var amqpConditionsToRecoveryKind = map[amqp.ErrorCondition]recoveryKind{
 	amqp.ErrorCondition("com.microsoft:operation-cancelled"): RecoveryKindNone,
 
 	// Link recovery needed
-	amqp.ErrorDetachForced: RecoveryKindLink, // "amqp:link:detach-forced"
+	amqp.ErrorDetachForced:          RecoveryKindLink, // "amqp:link:detach-forced"
+	amqp.ErrorTransferLimitExceeded: RecoveryKindLink,
```
jhendrixMSFT, line 95: Curious if you've actually seen this error, or just adding it for thoroughness.

richardpark-msft: It was in the original list of AMQP codes handled in JS so I brought it over but I've not seen it in the wild.

I'm not entirely sure if this error can even surface for us, but, from the service, I think it would mean that we've exceeded the transfer credit for the link. At best, I think we can try to "recover" it by terminating the link, which would cause outgoing messages to stop and be retried against a new link (ie, with fresh credits).

jhendrixMSFT (approved):

sdk/messaging/azservicebus/internal/errors.go
RickWinter, line 95: Doc comment

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -91,17 +91,18 @@ var amqpConditionsToRecoveryKind = map[amqp.ErrorCondition]recoveryKind{
...
 
 	// Connection recovery needed
 	amqp.ErrorConnectionForced: RecoveryKindConn, // "amqp:connection:forced"
+	amqp.ErrorInternalError:    RecoveryKindConn, // "amqp:internal-error"
```
RickWinter, line 99: Internal errors are always recoverable at the connection level?

RickWinter (approved):

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: The main issue with InternalError is that the cause of it is not known on our end. So the safest thing to do (rather than getting stuck in an infinite loop) is to at least try to recreate the entire thing.

We do our best internally to not lose any state, so at worse it's hopefully just a performance disruption.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: Added. There were a few others, so I added comments for them as well.

PR #17178, opened 2022-02-25
[Performance] fixing ops/s prints, small cleanup

### LINKED ISSUE #17176: [Performance] ops/s should be printed as a float

seankane-msft:
Closes #17176 

* ops/s is printed as a float with 2 decimal points of precision
* Removed print statements from previous PR
* Put argsument parsing in an `init` function.
* moves the `runtime.GOMAXPROCS` function to after calling `flag.Parse`

### DIFF
--- a/sdk/internal/perf/perf.go
+++ b/sdk/internal/perf/perf.go
@@ -12,6 +12,25 @@ import (
 	"runtime"
 )
 
+func init() {
+	// Start with adding all of our arguments
+	flag.IntVar(&duration, "d", 10, "Duration of test in seconds.")
+	flag.IntVar(&duration, "duration", 10, "Duration of test in seconds.")
+
+	flag.StringVar(&testProxyURLs, "test-proxies", "", "test proxy URLs to target. This can be a semi-colon separated list for multiple proxies.")
+	flag.StringVar(&testProxyURLs, "x", "", "test proxy URLs to target. This can be a semi-colon separated list for multiple proxies.")
+
+	flag.IntVar(&warmUpDuration, "warmup", 5, "Duration of warmup in seconds.")
+	flag.IntVar(&warmUpDuration, "w", 5, "Duration of warmup in seconds.")
+
+	flag.IntVar(&parallelInstances, "parallel", 1, "Degree of parallelism to run with.")
+	flag.IntVar(&parallelInstances, "p", 1, "Degree of parallelism to run with.")
+
+	flag.IntVar(&numProcesses, "maxprocs", runtime.NumCPU(), "Number of CPUs to use.")
+
+	flag.BoolVar(&debug, "debug", false, "Print debugging information")
+}
+
 // GlobalPerfTest methods execute once per process
 type GlobalPerfTest interface {
 	// NewPerfTest creates an instance of a PerfTest for each goroutine.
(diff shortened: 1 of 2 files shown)


sdk/internal/perf/perf.go
```diff
@@ -12,6 +12,25 @@ import (
 	"runtime"
 )
 
+func init() {
+	// Start with adding all of our arguments
+	flag.IntVar(&duration, "d", 10, "Duration of test in seconds.")
+	flag.IntVar(&duration, "duration", 10, "Duration of test in seconds.")
+
+	flag.StringVar(&testProxyURLs, "test-proxies", "", "test proxy URLs to target. This can be a semi-colon separated list for multiple proxies.")
```
richardpark-msft, line 20: Tell me about your ambitions to have multiple text proxies? Is this failover or something interesting?

sdk/internal/perf/perf.go
```diff
@@ -12,6 +12,25 @@ import (
 	"runtime"
 )
 
+func init() {
+	// Start with adding all of our arguments
+	flag.IntVar(&duration, "d", 10, "Duration of test in seconds.")
```
richardpark-msft, line 17: If you threw away the short options that would actually be allright by me, mostly because now the options are self-documenting in reviews.

(no big deal either way)

sdk/internal/perf/perf.go
```diff
@@ -12,6 +12,25 @@ import (
 	"runtime"
 )
 
+func init() {
+	// Start with adding all of our arguments
```
richardpark-msft, line 16: Can probably remove this comment as it's obvious ;)

sdk/internal/perf/perf.go
richardpark-msft: (also 'comma separated' might be more intuitive, but not a big deal either way)

sdk/internal/perf/perf.go
```diff
@@ -12,6 +12,25 @@ import (
...
+
+	flag.IntVar(&warmUpDuration, "warmup", 5, "Duration of warmup in seconds.")
+	flag.IntVar(&warmUpDuration, "w", 5, "Duration of warmup in seconds.")
+
+	flag.IntVar(&parallelInstances, "parallel", 1, "Degree of parallelism to run with.")
+	flag.IntVar(&parallelInstances, "p", 1, "Degree of parallelism to run with.")
+
+	flag.IntVar(&numProcesses, "maxprocs", runtime.NumCPU(), "Number of CPUs to use.")
+
+	flag.BoolVar(&debug, "debug", false, "Print debugging information")
```
richardpark-msft, line 31: What about -d? Oh wait...

sdk/internal/perf/status.go
```diff
@@ -164,10 +164,10 @@ func printFinalResults(elapsedTimes [][]float64, perSecondCount [][]int, warmup
 	secondsPerOp := 1.0 / opsPerSecond
 	weightedAvgSec := float64(totalOperations) / opsPerSecond
 	fmt.Printf(
-		"Completed %s operations in a weighted-average of %.2fs (%s ops/s, %.3f s/op)\n",
+		"Completed %s operations in a weighted-average of %ss (%s ops/s, %s s/op)\n",
```
richardpark-msft, line 167: This is neat, I'd never heard of message.Printer before.

richardpark-msft (approved):
Nothing big, left a few comments.

sdk/internal/perf/perf.go
seankane-msft: semi-colon separated is the language agnostic way we've determined. This helps remove the test proxy as a bottleneck. If we have 8 parallel instances, we can also have 8 different proxies running and target one proxy per process.

sdk/internal/perf/status.go
seankane-msft: Does a very good job of making numbers more readable with comma separated values.

sdk/internal/perf/perf.go
seankane-msft: Keeping them for now because this is how other languages do them, we might need them for language agnostic automation, if not I can remove them later.

PR #17202, opened 2022-03-02
Bulk fix adding --scopes Parameter to Documentation

v-alje:
This is a bulk fix adding the --scopes parameter to `az ad sp create-for-rbac` commands in documentation.

### DIFF
--- a/documentation/previous-versions-quickstart.md
+++ b/documentation/previous-versions-quickstart.md
@@ -163,7 +163,7 @@ options offered by the SDK because it allows seamless use of both service
 principals and [Azure Managed Service Identity][]. Other options are listed
 below.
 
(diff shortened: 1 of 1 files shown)


v-alje:
@dbradish-microsoft

dbradish-microsoft:
@v-alje, please mark as ready for review.

dbradish-microsoft:
#sign-off with sufficient time given for author review.

The Azure subscription no longer defaults to the current active subscription in the `az ad sp create-for-rbac` reference command.  You must now specify a subscription ID.

dbradish-microsoft:
@mohsha-msft can you please merge this PR?  The subscription no longer defaults for the create-for-rbac reference. You must specify it. Thank you.

jhendrixMSFT (approved):

PR #17225, opened 2022-03-03
[Certificates] Remove `RawResponse`

seankane-msft:
Remove `RawResponse` and add documentation

### DIFF
--- a/sdk/keyvault/azcertificates/README.md
+++ b/sdk/keyvault/azcertificates/README.md
@@ -385,6 +385,20 @@ log.SetClassifications(log.Request, log.Response)
 > CAUTION: logs from credentials contain sensitive information.
 > These logs must be protected to avoid compromising account security.
 
+### Accessing `http.Response`
+You can access the raw `*http.Response` returned by the service using the `runtime.WithCaptureResponse` method and a context passed to any client method.
+
+```go
+var respFromCtx *http.Response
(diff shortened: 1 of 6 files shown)


sdk/keyvault/azcertificates/README.md
```diff
@@ -385,6 +385,20 @@ log.SetClassifications(log.Request, log.Response)
 > CAUTION: logs from credentials contain sensitive information.
 > These logs must be protected to avoid compromising account security.
 
+### Accessing `http.Response`
+You can access the raw `*http.Response` returned by the service using the `runtime.WithCaptureResponse` method and a context passed to any client method.
+
+```go
+var respFromCtx *http.Response
+ctx := runtime.WithCaptureResponse(context.Background(), &respFromCtx)
```
richardpark-msft, line 393: I know you didn't write this but looking at it now - is the intention that this be used similar to context.With<blah> functions? If so, should it return a cancel() func?

And if not (and the return value is just a convenience) is it safe to use `context.Background()`?

richardpark-msft: CC: @jhendrixMSFT

jhendrixMSFT: We were going to do that.  However, the implementation would make it less clear that shared state is being used.  See the discussion in https://github.com/Azure/azure-sdk-for-go/pull/17156 for more info.

jhendrixMSFT (approved):

PR #17231, opened 2022-03-04
[Perf] Move result processing to background threads

seankane-msft:


jhendrixMSFT (approved):

richardpark-msft (commented):
Just curious - have you run this with the `-race` flag? It'll help you ensure your async code is race free (or at least, it'll check quite a bit).

PR #17233, opened 2022-03-05
Remove unused function

antkmsft:
Remove this unused funtion. `Do` function does the same thing. This function was not removed by mistake.


seankane-msft (approved):

jhendrixMSFT:
Looks like this module is missing its `ci.yml` file.

Also, I think we should shorten the directory name to `appconfig` to align with the module name.

RickWinter (approved):

seankane-msft:
> Looks like this module is missing its `ci.yml` file.
> 
> Also, I think we should shorten the directory name to `appconfig` to align with the module name.

You can copy the tables ci here: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/ci.yml

antkmsft:
@jhendrixMSFT, @seankane-msft, thank you, I added ci.yml and renamed `appconfiguration` directory to `appconfig`.

seankane-msft:
> @jhendrixMSFT, @seankane-msft, thank you, I added ci.yml and renamed `appconfiguration` directory to `appconfig`.

Were there also changes to go.mod that need to be made?

jhendrixMSFT:
Oh yes the go.mod will need to be updated so it matches the directory structure.

sdk/appconfig/azappconfig/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/Azure/azure-sdk-for-go/sdk/appconfig/azappconfig
```
jhendrixMSFT, line 1: go.sum should be included

antkmsft: Thank you, I'll revert.

jhendrixMSFT:
@benbp do we need to run a script to get the pipeline set up?

benbp:
@jhendrixMSFT I just kicked it off ^^

benbp:
@antkmsft FYI the appconfig pipeline just created has some failures: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1414958&view=results

PR #17240, opened 2022-03-08
App Configuration: prepare release

antkmsft:
Release work item: https://dev.azure.com/azure-sdk/Release/_workitems/edit/10782/


sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,305 @@
...
+    connectionString := os.Getenv("APPCONFIGURATION_CONNECTION_STRING")
+    client, err = azappconfig.NewClientFromConnectionString(connectionString, nil)
+    if err != nil {
+        panic(err)
+    }
+
+    // Set configuration setting read only
+    key := "key"
+    label := "label"
+    resp, err := client.SetReadOnly(context.TODO(), Setting{Key: &key, Label: &label, true, nil)
```
seankane-msft, line 199: Also missing a bracket

sdk/appconfig/azappconfig/client_test.go
```diff
@@ -15,10 +15,16 @@ import (
 )
 
 func TestClient(t *testing.T) {
+	connectionString := os.Getenv("APPCONFIGURATION_CONNECTION_STRING")
+	if connectionString == "" {
+		// TODO: Need CI to deploy test-resources.json
```
seankane-msft, line 20: Note: you can use the python test-resources.json file [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/test-resources.json)

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,305 @@
...
+names):
+
+  (Optional) if you want a new resource group to hold the Azure App Configuration:
+  ```sh
+  az group create --name my-resource-group --location westus2
+  ```
+
+  Create the Key Vault:
+  ```Bash
+  az appconfig create --resource-group my-resource-group --name my-app-config --location westus2
```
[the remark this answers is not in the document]
antkmsft: Yes, I was able to create it.

sdk/appconfig/azappconfig/client_test.go
antkmsft: test-resources.json is already checked in, the problem is that CI is not deploying it. I run tests manually at this point.

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,305 @@
...
+    connectionString := os.Getenv("APPCONFIGURATION_CONNECTION_STRING")
+    client, err = azappconfig.NewClientFromConnectionString(connectionString, nil)
+    if err != nil {
+        panic(err)
+    }
+
+    any := "*"
+    revPgr := client.ListRevisions(SettingSelector{KeyFilter: &any, LabelFilter: &any, Fields: AllSettingFields()}, nil)
+    for revPgr.NextPage(context.TODO()) {
+        resp := revPgr.PageResponse()
```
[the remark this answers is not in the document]
antkmsft: Which package is using the new API so I can use it as example? I'd change.

sdk/appconfig/azappconfig/client_test.go
benbp: @antkmsft I see the template getting deployed here? https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1417572&view=logs&j=a46fb01e-fa54-52e9-355f-ff40079f227e&t=5d7559a5-142f-5cf9-0231-17146bcf4d47

antkmsft: Ah, I see, it does run live and pass in tests, it fails during regular unit tests. I'll try to see if I can mark it as live-only.

antkmsft: Currently it will run during the live test, when the template is deployed and the corresponding environment variable is set.

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,298 @@
...
+### Install packages
+Install [azappconfig][goget_azappconfig] with `go get`:
+```Bash
+go get github.com/Azure/azure-sdk-for-go/sdk/appconfig/azappconfig
+```
+
+### Prerequisites
+* An [Azure subscription][azure_sub]
+* Go version 1.16 or higher
+* A Key Vault. If you need to create one, you can use the [Azure Cloud Shell][azure_cloud_shell] to create one with these commands (replace `"my-resource-group"` and `"my-app-config"` with your own, unique
```
richardpark-msft, line 17:
```suggestion
* An App Configuration store. If you need to create one, you can use the [Azure Cloud Shell][azure_cloud_shell] to create one with these commands (replace `"my-resource-group"` and `"my-app-config"` with your own, unique
```

(Unless you're trying to show the keyvault integration, perhaps?)

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,298 @@
...
+  }
+  ```
+
+  > The `"endpoint"` property is the `endpointUrl` used by [Client][appconfig_client_src]
+
+### Authenticate the client
+This document demonstrates using the connection string. However, [Client][appconfig_client_src] accepts any [azidentity][azure_identity] credential. See the [azidentity][azure_identity] documentation for more information about other credentials.
+
+#### Create a client
+Constructing the client requires your vault's connection string, which you can get from the Azure Portal.
```
richardpark-msft, line 51:
```suggestion
Constructing the client requires your App Configuration connection string, which you can get from the Azure Portal.
```

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+func ExampleAddConfigurationSetting() {
+    connectionString := os.Getenv("APPCONFIGURATION_CONNECTION_STRING")
+    client, err = azappconfig.NewClientFromConnectionString(connectionString, nil)
+    if err != nil {
+        panic(err)
+    }
+
+    // Create configuration setting
+    resp, err := client.AddSetting(context.TODO(), azappconfig.Setting{Key: to.StringPtr("key"), Label: to.StringPtr("label"), Value: to.StringPtr("value")}, nil)
```
richardpark-msft, line 106: It would be nice to add some whitespace here, just to make it more clear you're setting three fields in a Setting:
(whatever gofmt ends up doing)

    resp, err := client.AddSetting(context.TODO(), 
      azappconfig.Setting{
        Key: to.StringPtr("key"), 
        Label: to.StringPtr("label"), 
        Value: to.StringPtr("value")
      }, nil)

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+func ExampleGetConfigurationSetting() {
+    connectionString := os.Getenv("APPCONFIGURATION_CONNECTION_STRING")
+    client, err = azappconfig.NewClientFromConnectionString(connectionString, nil)
+    if err != nil {
+        panic(err)
+    }
+
+    // Get configuration setting
+    resp, err := client.GetSetting(context.TODO(), azappconfig.Setting{Key: to.StringPtr("key"), Label: to.StringPtr("label")}, nil)
```
richardpark-msft, line 135: Does 'key' need to be defined as a string pointer? I wouldn't think it'd be optional at all (label is, however).

sdk/appconfig/azappconfig/README.md
richardpark-msft: (there are a few spots like this, in case you decide you want to change it)

sdk/appconfig/azappconfig/README.md
antkmsft: Yes - because `Setting` is a return type, and you may ask for the list of items via `ListRevisions`, and ask it to not include `Key` column. In that case, the service will return `Setting` with `nil` `Key`.

sdk/appconfig/azappconfig/README.md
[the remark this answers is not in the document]
antkmsft: I modernized the pager. But I think in the long term we should update codegen, so that it is easier to implement and so that it does not require to expose internal members.

sdk/appconfig/azappconfig/README.md
antkmsft: Thank you!

sdk/appconfig/azappconfig/README.md
antkmsft: Thank you, I updated the samples to include more spacing/better formatting using newlines.

sdk/appconfig/azappconfig/client_test.go
antkmsft: Live tests do run and pass now: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1419414&view=logs&j=a46fb01e-fa54-52e9-355f-ff40079f227e&t=db30c526-9df4-58f3-c8d1-1445458d3e15

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,356 @@
...
+
+```go
+import (
+    "github.com/Azure/azure-sdk-for-go/sdk/appconfig/azappconfig"
+    "github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+)
+
+credential, err := azidentity.NewDefaultAzureCredential(nil)
+
+client, err = azappconfig.NewClient("https://my-app-config.azconfig.io", credential, nil)
```
seankane-msft, line 75:
```suggestion
client, err := azappconfig.NewClient("https://my-app-config.azconfig.io", credential, nil)
```

seankane-msft: This occurs throughout, not going to point out all of them

sdk/appconfig/azappconfig/README.md
```diff
@@ -0,0 +1,356 @@
...
+    revPgr := client.ListRevisions(
+        azappconfig.SettingSelector{
+            KeyFilter: to.StringPtr("*"),
+            LabelFilter: to.StringPtr("*"),
+            Fields: azappconfig.AllSettingFields()
+        },
+        nil)
+
+    for revPgr.More() {
+        if revResp, revErr := revPgr.NextPage(context.TODO()); revErr != nil {
```
seankane-msft, line 285:
```suggestion
        if revResp, revErr := revPgr.NextPage(context.TODO()); revErr == nil {
```

seankane-msft (approved):
LGTM, thanks for your patience
:shipit:

sdk/appconfig/azappconfig/README.md
richardpark-msft: But this argument is for getting a setting - it has different semantics than ListSetting().

Is there a reason we're not just using 'key' as a positional argument?

antkmsft:
Thank you!

sdk/appconfig/azappconfig/README.md
antkmsft: Then the key will be duplicated in the `Setting`, and as a parameter, I think it will be confusing.
If you look at the Java or .NET SDK, they have either methods with this semantics, taking `Setting`, or an overload with string key and optional string label, but not a mix.

richardpark-msft: (this isn't a "delay this preview" thing. But let's discuss it for the next one)

richardpark-msft: (if you don't mind filing an issue)

antkmsft: As of ideas, initially, I had `Setting` being a struct with private fields, which you could only create as `NewSetting()` with string key parameter. It had getters to get fields as pointers, and "Setters", such as `WithLabel()`, `WithValue()` also taking strings and not pointers. Not all fields were writeable - the same as in other SDKs. And it was not possible then for user to create a setting without key. `With...()` were retutning a copy of the struct, with the corresponding field being set. But I was told such API is not Go-like, where everything is much simpler, uses bare structs, pointers, relies on comments etc.

The issue is: #17264

PR #17244, opened 2022-03-08
[Tables] Batching response struct is now empty

seankane-msft:
After discussion with Jeff, the `TransactionResponse` object is empty now and customers can parse the `azcore.ResponseError` to understand why the transaction failed.


richardpark-msft (approved):
Interesting...

PR #17257, opened 2022-03-09
feat: add testutil for resourcemanager

tadelesh:


sdk/resourcemanager/internal/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/internal
+
+go 1.16
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.22.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.13.1
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.9.1
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v0.3.1
```
ArcturusZhang, line 9: Would this be a cyclic reference between this module and resources module?

sdk/resourcemanager/internal/testutil/helper.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func GetEnv(key, fallback string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return fallback
+}
+
+func CreateResourceGroup(ctx context.Context, subscriptionId string, cred azcore.TokenCredential, options *arm.ClientOptions, location string) (*armresources.ResourceGroup, error) {
```
ArcturusZhang, line 30: Shall we let this function also return a delegate which cleans the resource group? So that we could do this in our test cases in RPs:
```go
rg, clean, err := testutil.CreateResourceGroup(...)
require.NoError(err)
defer clean() // this will delete the created resource group
```
In this way we do not need to explicitly call the `DeleteResourceGroup` function in our test cases

sdk/resourcemanager/internal/testutil/credential.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type FakeCredential struct {
+}
+
+func (c *FakeCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
+	return &azcore.AccessToken{Token: "FakeToken", ExpiresOn: time.Now().Add(time.Hour * 24).UTC()}, nil
+}
+
+func GetCredAndClientOptions(t *testing.T) (azcore.TokenCredential, *arm.ClientOptions) {
```
ArcturusZhang, line 28: I believe when we are using this, we should call `GetCredAndClientOptions` to get the fake credential and the corresponding client options, we will never instantiate a FakeCredential in our test cases.
If this is true, we should unexport the `FakeCredential` struct.

ArcturusZhang (commented):
Some minor comments.

sdk/resourcemanager/internal/go.mod
tadelesh: I notice the cyclic reference when add this dep. But I've tested the `testutil` in `armresources`, it seems no mod import problem.

ArcturusZhang: Importing these does not have problems. But we might have problems when we have to update these two. But this is a minor problem

sdk/resourcemanager/internal/testutil/helper.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}, nil
+}
+
+func DeleteResourceGroup(ctx context.Context, subscriptionId string, cred azcore.TokenCredential, options *arm.ClientOptions, resourceGroupName string) error {
+	rgClient := armresources.NewResourceGroupsClient(subscriptionId, cred, options)
+	poller, err := rgClient.BeginDelete(ctx, resourceGroupName, nil)
```
ArcturusZhang, line 52: Do we have to wait until the deletion is done in this function? For clean up, we do not have to wait for the clean up to be done to declare our test is over.
So that I believe we could just return the poller here and let the deletion begins during the clean up.

sdk/resourcemanager/internal/testutil/helper.go
```diff
@@ -0,0 +1,76 @@
...
+					return err
+				}
+				break
+			}
+		}
+	} else {
+		if _, err = poller.PollUntilDone(ctx, 10*time.Second); err != nil {
+			return err
+		}
+	}
```
ArcturusZhang, line 74: I kind of do not get the point here - do we have to do this for all pollers when it is in playback mode? If not, we do not have to treat this differently here

ArcturusZhang (approved):
LGTM besides some minor comments
BTW could we add some document to the public functions in this module?

sdk/resourcemanager/internal/go.mod
tadelesh: Got it. We may carefully handle it by specific release step.

sdk/resourcemanager/internal/testutil/helper.go
tadelesh: Yes, we need to do this for all `pollers`. It is because the interval time of `PollUntilDone` has a constraint of bigger than 1s. After discussed with Joel, the only way to accelerate the test speed when playback is to write own poll logic.

sdk/resourcemanager/internal/testutil/helper.go
tadelesh: Good point. As the randomly resource group name, async deletion is reasonable.

PR #17260, opened 2022-03-09
[Perf] adding a performance test for runtime.Pipeline

### LINKED ISSUE #17213: [Perf] Port pipeline client GET test
* Python example: https://github.com/Azure/azure-sdk-for-python/blob/bedff8a444413458c84bcc8d08caa7b7e84f2b80/tools/azure-devtools/src/azure_devtools/perfstress_tests/system_perfstress/pipeline_client_get_test.py

seankane-msft:
Currently blocked by #17231

Closes #17213

### DIFF
--- a/sdk/azcore/testdata/perf/client_get.go
+++ b/sdk/azcore/testdata/perf/client_get.go
@@ -1,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package main
+
+import (
+	"context"
+	"flag"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/pipeline"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/perf"
+)
+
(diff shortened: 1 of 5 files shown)


sdk/azcore/testdata/perf/client_get.go
```diff
@@ -0,0 +1,76 @@
...
+
+type clientGETTestOptions struct {
+	url string
+}
+
+var clientGetOpts clientGETTestOptions = clientGETTestOptions{url: ""}
+
+// sleepTestRegister is called once per process
+func clientTestRegister() {
+	flag.StringVar(&clientGetOpts.url, "url", "", "URL to send a GET request")
```
richardpark-msft, line 26: What do you typically use here now? (maybe the bing.com default was never actually triggered in practice/)

richardpark-msft (approved):
Looks okay and pretty minor.

sdk/azcore/testdata/perf/client_get.go
seankane-msft: When I'm testing for validation I have used bing.com. I'm not sure what @mikeharder uses in the real performance tests.

mikeharder: We currently do not run an HttpClient GET test in the perf automation.  If I were to run an HttpClient GET test manually, I would use a super-fast "hello world" web server (like https://github.com/mikeharder/MinKestrel/tree/master/MinKestrel) running on a VM in the same virtual network as the client VM.

PR #17261, opened 2022-03-09
[EngSys] Optionally extend test runtime

seankane-msft:
It happens sporadically, but `azcertificates` (full of LROs) live tests will fail because the tests do not complete in >600s (the default in Go). This change provides a way to specify in the `ci.yml` for larger test suites.

### DIFF
--- a/eng/scripts/run_tests.ps1
+++ b/eng/scripts/run_tests.ps1
@@ -6,9 +6,9 @@ Param(
 )
 
 Push-Location sdk/$serviceDirectory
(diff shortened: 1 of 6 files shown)


eng/scripts/run_tests.ps1
```diff
@@ -1,13 +1,14 @@
 #Requires -Version 7.0
 
 Param(
-    [string] $serviceDirectory
+    [string] $serviceDirectory,
+    [string] $runTime
```
benbp, line 5: Would `testTimeout`, `maxTestTime`, etc. be a more descriptive name?

benbp (approved):

PR #17266, opened 2022-03-09
[Blob] returning clients by pointer

seankane-msft:


antkmsft (approved):

sdk/storage/azblob/zc_append_blob_client.go
```diff
@@ -17,55 +17,55 @@ type AppendBlobClient struct {
...
+	return &AppendBlobClient{
 		client:     &appendBlobClient{con: con},
 		BlobClient: BlobClient{client: &blobClient{con: con}},
 	}
 }
 
 // WithVersionID creates a new AppendBlobURL object identical to the source but with the specified version id.
 // Pass "" to remove the versionID returning a URL to the base blob.
-func (ab AppendBlobClient) WithVersionID(versionID string) AppendBlobClient {
+func (ab AppendBlobClient) WithVersionID(versionID string) *AppendBlobClient {
```
mohsha-msft, line 63: Can you also change the receivers `(ab AppendBlobClient)` to pointers `(ab *AppendBlobClient)`.
If we're returning clients as pointers, it makes sense to have all receivers as pointers as well

mohsha-msft (commented):
Changes look straightforward.
Please fix the change suggested.
Also, please keep base branch as `mohsha-msft/bug-shared-cred` instead of  `main` please. I've done changes in the clients and I'd love to have your review + commit in [this PR](https://github.com/Azure/azure-sdk-for-go/pull/17304/files?file-filters%5B%5D=.go&show-viewed-files=true). Some clients were not returning error so I fixed that and some didn't have sharedkey initialized so SAS Generation wasn't happening properly.

PR #17274, opened 2022-03-10
Fix blob versioning bug

mohsha-msft:


sdk/storage/azblob/testdata/perf/download_blob.go
```diff
@@ -1,140 +0,0 @@
-// // Copyright (c) Microsoft Corporation. All rights reserved.
```
seankane-msft, line 1: Can we leave the perf tests?

sdk/storage/azblob/zc_page_blob_client.go
```diff
@@ -70,7 +70,7 @@ func (pb PageBlobClient) WithSnapshot(snapshot string) PageBlobClient {
 // WithVersionID creates a new PageBlobURL object identical to the source but with the specified snapshot timestamp.
 // Pass "" to remove the version returning a URL to the base blob.
 func (pb PageBlobClient) WithVersionID(versionID string) PageBlobClient {
-	p := NewBlobURLParts(pb.URL())
+	p, _ := NewBlobURLParts(pb.URL())
```
seankane-msft, line 73: We should check all of these errors

seankane-msft (approved):

sdk/storage/azblob/zc_page_blob_client.go
mohsha-msft: This for sure wouldn't be a problem. that's why I didn't check it. Problem occurs when customer provides us URL directly. In that case we do parsing error check inside the NewBlobURLParts() method.

mohsha-msft:
FYI @zezha-msft

Issue #17095, opened 2022-02-17
Add a CI check that flags all exported symbols that are missing doc comments.


RickWinter:
We really need a CI check that flags all exported symbols that are missing doc comments.

_Originally posted by @jhendrixMSFT in https://github.com/Azure/azure-sdk-for-go/pull/17093#r809211410_

benbp:
This does not appear to be a simple fix (see https://github.com/golangci/golangci-lint/issues/456), but I'm digging into config updates/workarounds for linting.

benbp:
Fixed by https://github.com/Azure/azure-sdk-for-go/issues/17095

PR #17278, opened 2022-03-10
[Eng] add tool for checking public docs, fixes to azcore docs

seankane-msft:
Adds a tool to validate all exported items will have a doc comment that's properly read by pkg.go.dev.


benbp (approved):

sdk/azcore/arm/resource_identifier.go
```diff
@@ -85,6 +83,7 @@ func ParseResourceID(id string) (*ResourceID, error) {
 	return appendNext(RootResourceID, parts, id)
 }
 
+// String returns the string of the ResourceId
```
jhendrixMSFT, line 86: ResourceID

eng/tools/doccheck/main.go
```diff
@@ -0,0 +1,111 @@
...
+	"path/filepath"
+	"strings"
+)
+
+func filter(f fs.FileInfo) bool {
+	return !strings.HasSuffix(f.Name(), "_test.go")
+}
+
+func findAllSubDirectories(root string) []string {
+	ret := make([]string, 0)
```
richardpark-msft, line 20: Small nit, but you can just do this:

```go
var ret []string
```

A 'nil' slice value is basically an empty slice - append (and len) handle it just fine.

sdk/azcore/arm/resource_type.go
```diff
@@ -11,19 +11,17 @@ import (
...
-var (
-	// SubscriptionResourceType is the ResourceType of a subscription
-	SubscriptionResourceType = NewResourceType(builtInResourceNamespace, "subscriptions")
+// SubscriptionResourceType is the ResourceType of a subscription
+var SubscriptionResourceType = NewResourceType(builtInResourceNamespace, "subscriptions")
 
-	// ResourceGroupResourceType is the ResourceType of a resource group
-	ResourceGroupResourceType = NewResourceType(builtInResourceNamespace, "resourceGroups")
+// ResourceGroupResourceType is the ResourceType of a resource group
+var ResourceGroupResourceType = NewResourceType(builtInResourceNamespace, "resourceGroups")
```
richardpark-msft, line 18: Conforming to a style rule or just something you wanted to change?

eng/tools/doccheck/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	filepath.Walk(root, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			panic(err)
+		}
+		if info.IsDir() && !strings.Contains(path, "internal") {
```
richardpark-msft, line 26: You can apparently stop recursing in specific folders by returning SkipDir (just found out about this)

> The error result returned by the function controls how Walk continues. If the function returns the special value SkipDir, Walk skips the current directory

This would also prevent the unlikely problem that someone has *internal* in their string and it's not actually an `internal` folder. Also, gets rid of some unneeded calls.

richardpark-msft (approved):
Minor stuff. Thank you for fixing azservicebus's doc comments :)

sdk/azcore/arm/resource_type.go
seankane-msft: Style rule, when `var`s or `const`s are grouped together, they are displayed as a code block.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17095

tadelesh:
@seankane-msft Lots of public types in mgmt. plane SDK which are auto generated have no doc comment. Can we make this check temporarily not mandatory? @jhendrixMSFT If some params have no description (one [example](https://github.com/Azure/azure-rest-api-specs/blob/96da969aa75e2beb27c3dbc2d69a76900856cc52/specification/storage/resource-manager/Microsoft.Storage/stable/2021-08-01/file.json#L982-L998)), can we put am empty comment to pass this check rule?

seankane-msft:
I'm making this change here https://github.com/Azure/azure-sdk-for-go/pull/17292

It will still run, but it will not fail on missing doc checks

PR #17279, opened 2022-03-11
[azservicebus] Fixing the issue with 404's failing fast-ish, but not fast.

### LINKED ISSUE #16088: [azservicebus] amqp-common's link.RetryableRPC() retries an entity does not exist

richardpark-msft:
The main issue is that the negotiateClaim() function was doing it's own
retries and it wasn't properly handling 404 as a "fatal" error. We'd eliminated
the outmost retry set (where we were basically doing 'n' retries around 'n' retries)
but there were still some incorrect retries happening in the negotiateClaim 
function itself.

Rather than continuing to patch the 'internal' version I'm just moving it
all into the azservicebus package and leveraging the normal retry and error 
recovery. I stopped short of doing a full change (to make it so we can hold 
onto an RPC link for claims negotiation).

Some supporting changes:
- mgmtError is now rpcError and is just generically the "request/response"
  error type. Also, the actual decision of creating an rpcError is built
  in to rpcLink.RPC(), rather than being done by callers.
- Small change to make NewRPCLink take an args structure and remove the
  options that weren't used.
- Removed 'tab' calls in code I'm modifying since we're going to be
  replacing them/moving them anyways.

Fixes #16088 (completely this time)


sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -171,10 +171,14 @@ func GetRecoveryKind(err error) recoveryKind {
...
 
-	var me mgmtError
+	var rpcErr rpcError
 
-	if errors.As(err, &me) {
-		code := me.RPCCode()
+	if errors.As(err, &rpcErr) {
+		code := rpcErr.RPCCode()
+
+		if code == 404 {
```
jhendrixMSFT, line 179: At some point would be good to switch to using the constants from the [`net/http`](https://pkg.go.dev/net/http#pkg-constants) package.

sdk/messaging/internal/cbs/cbs.go
```diff
@@ -1,90 +0,0 @@
-// Package cbs provides the functionality for negotiating claims-based security over AMQP for use in Azure Service Bus
```
jhendrixMSFT, line 1: What will this mean for Event Hubs?

richardpark-msft: When we get there I'll merge whatever we end up with back into internal.

richardpark-msft: (I have some more changes coming for this but I didn't want to do it all in the same PR)

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: They do seem to line up with HTTP status codes but they aren't coming back from HTTP in this case. 

I could make some aliases though since it's kind of silly to just have the values hanging out.

richardpark-msft: (going to leave this for now. I need to go actually repro some of these to see the descriptions that come back from the service)

PR #17280, opened 2022-03-11
[azservicebus] Stress test improvements, adding additional test for lazy/infrequent receiving

richardpark-msft:
- Changing the infiniteSendAndReceive test to be cleaner. Rather than having several links we just have the one (since multiple wasn't really giving us additional information).
- Adding in a new test - mostly_idle_receiver.go. Some customers have reported issues where their receiver seems to not receive after x amount of time. This test creates a connection, let's it sit idle and then sends and receive a message. This should ensure that even the laziest/slowest pattern of receiving works. Goes all the way up to 24h.
- Updated the bicep and deploy-job.yaml so we can just uncomment a few lines and deploy to southeastasia. There's an issue submitted for engsys to make this not require editing in the future.


sdk/messaging/azservicebus/internal/stress/shared/stress_context.go
```diff
@@ -89,7 +89,7 @@ func MustCreateStressContext(testName string) *StressContext {
 		Nano:             testRunID, // the same for now
 		ConnectionString: cs,
 		TelemetryClient:  telemetryClient,
-		statsPrinter:     newStatsPrinter(ctx, testName, 5*time.Second, telemetryClient),
+		statsPrinter:     newStatsPrinter(ctx, testName, time.Minute, telemetryClient),
```
RickWinter, line 92: Nice to comment on why we want a minute.

sdk/messaging/azservicebus/internal/stress/stress-test-resources.bicep
```diff
@@ -5,6 +5,7 @@ param baseName string = resourceGroup().name
 param testApplicationOid string
 
 var apiVersion = '2017-04-01'
+//var location = 'southeastasia'
```
RickWinter, line 8: Remove rather than comment

RickWinter (approved):

sdk/messaging/azservicebus/internal/stress/stress-test-resources.bicep
richardpark-msft: It's commented so I don't have to keep looking it up. We're in this weird in-between state where we have a manually created set of nodes for southeastasia, so you have to manually target them. (I'll delete it when we finalize all of that).

sdk/messaging/azservicebus/internal/stress/shared/stress_context.go
richardpark-msft: Oh, yeah, it's nothing special. Just long enough so I don't feel like I'm waiting forever, but not so often that @benbp  complains I'm eating up all of our diagnostics log space.

richardpark-msft: Updated!

sdk/messaging/azservicebus/internal/stress/stress-test-resources.bicep
richardpark-msft: Another comment added.

PR #17282, opened 2022-03-11
doc fixes for KV and Tables

seankane-msft:


sdk/data/aztables/entity.go
```diff
@@ -133,10 +135,12 @@ func (e *EDMEntity) UnmarshalJSON(data []byte) (err error) {
 // EDMBinary will also receive the correct odata annotation for round-trip accuracy.
 type EDMBinary []byte
 
+// MarshalText implements the json.Marshal method
```
jhendrixMSFT, line 138: implements the encoding.TextMarshaler interface

sdk/data/aztables/sas_account.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 )
 
+// SASVersion is the default SAS Service Version
```
jhendrixMSFT, line 14: SAS Version

seankane-msft: Comments are supposed to start with the name of the type they are documenting, so in this case leaving it all as one word.

chlowell: Why is this a variable, and exported?

jhendrixMSFT: Sorry I was referring to the end of the comment.  "SASVersion is the default SAS version".  The extra "Service" seems redundant.

sdk/keyvault/azkeys/crypto/constants.go
```diff
@@ -38,23 +47,50 @@ func PossibleWrapAlgorithmValues() []WrapAlgorithm {
...
-	EncryptionAlgorithmA192KW     EncryptionAlgorithm = "A192KW"
-	EncryptionAlgorithmA256CBC    EncryptionAlgorithm = "A256CBC"
-	EncryptionAlgorithmA256CBCPAD EncryptionAlgorithm = "A256CBCPAD"
-	EncryptionAlgorithmA256GCM    EncryptionAlgorithm = "A256GCM"
-	EncryptionAlgorithmA256KW     EncryptionAlgorithm = "A256KW"
-	EncryptionAlgorithmRSA15      EncryptionAlgorithm = "RSA1_5"
-	EncryptionAlgorithmRSAOAEP    EncryptionAlgorithm = "RSA-OAEP"
-	EncryptionAlgorithmRSAOAEP256 EncryptionAlgorithm = "RSA-OAEP-256"
-)
+// EncryptionAlgorithmA128CBC is the 'A128CBC' encryption algorithm
```
chlowell, line 50: Is there a reason to get rid of all these `const` blocks, other than just not liking `const` blocks? 😸

chlowell (approved):

sdk/keyvault/azkeys/crypto/constants.go
jhendrixMSFT: Is this because the tool is choking on the block?  If so, we should fix the tool instead (same for `var` blocks).

sdk/data/aztables/sas_account.go
seankane-msft: @chlowell This is the default value used for `AccountSASSignatureValues`

PR #17295, opened 2022-03-14
[azservicebus] Adding tests for multi-link and multi-entity

richardpark-msft:
AMQP lets you create multiple links on a single connection. As a simple sanity check I'm actually using that - sending and receiving to separate queues using the same connection and also just creating multiple sender/receiver links and sending/receiving to the same queue.

### DIFF
--- a/sdk/messaging/azservicebus/liveTestHelpers_test.go
+++ b/sdk/messaging/azservicebus/liveTestHelpers_test.go
@@ -27,15 +27,12 @@ func setupLiveTest(t *testing.T, props *admin.QueueProperties) (*Client, func(),
 	testCleanup := func() {
 		require.NoError(t, serviceBusClient.Close(context.Background()))
 		cleanupQueue()
-		require.NoError(t, serviceBusClient.Close(context.Background()))
 	}
 
 	return serviceBusClient, testCleanup, queueName
 }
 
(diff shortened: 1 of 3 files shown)


sdk/messaging/azservicebus/liveTestHelpers_test.go
```diff
@@ -27,15 +27,12 @@ func setupLiveTest(t *testing.T, props *admin.QueueProperties) (*Client, func(),
 	testCleanup := func() {
 		require.NoError(t, serviceBusClient.Close(context.Background()))
 		cleanupQueue()
-		require.NoError(t, serviceBusClient.Close(context.Background()))
```
seankane-msft, line 30: Does closing twice return an error now?

seankane-msft (approved):

sdk/messaging/azservicebus/liveTestHelpers_test.go
richardpark-msft: It just seemed unnecessary but it shouldn't cause an error (well, it never has, even now). But maybe I should keep it, why not. I'll add it back.

richardpark-msft: Added it back and popped a comment on there.

Issue #17177, opened 2022-02-25
[Performance] Cancel context at end of tests


seankane-msft:
Perf framework should cancel context at end of warmup/duration, and tests should best-effort respect the cancellation request

seankane-msft:
Closed with #17296

Issue #17175, opened 2022-02-25
[Performance] Handle results in a background thread


seankane-msft:

seankane-msft:
azblob tests are broken at the moment.

seankane-msft:
Add `go build .` to the CI

seankane-msft:
Closed with https://github.com/Azure/azure-sdk-for-go/pull/17296

PR #17296, opened 2022-03-14
[Perf] Fixing race conditions in performance

seankane-msft:
Hopefully this is the last _large_ PR for re-writing portions of the Perf framework

What's changed
* created a `perfRunner` struct for handling running a single test
* Using `sync.Mutex`s and `atomic.Load/Add` for eliminating data races
* improved simplicity of status updates


sdk/internal/perf/perf.go
```diff
@@ -64,6 +66,35 @@ type PerfTestOptions struct {
...
+func newPerfTestOptions(name string) PerfTestOptions {
+	return PerfTestOptions{
+		Name:          name,
+		runElapsed:    newAtomicFloat64(0),
+		warmupElapsed: newAtomicFloat64(0),
+	}
+}
+
+// increment does an atomic increment of the warmup or non-warmup performance test
+func (p *PerfTestOptions) incrememt(warmup bool) {
```
richardpark-msft, line 92:
```suggestion
func (p *PerfTestOptions) increment(warmup bool) {
```

sdk/internal/perf/atomic_float.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package perf
+
+import (
+	"sync"
+)
+
+type atomicFloat64 struct {
```
richardpark-msft, line 10: Something to consider in a future PR - would it be simpler to store off the time.Duration values, which fit into an int64 and just use the plain `atomic` functions instead? 

(I only briefly looked but it looks like this is meant to be "fractional seconds", but you really can only measure to the granularity of the time keeping in Go anyways)

sdk/internal/perf/implementation.go
```diff
@@ -4,295 +4,36 @@
...
+	numProcesses int
+)
 
-// parse the TestProxy input into a slice of strings
+// parseProxyURLs splits the --test-proxy input with the delimiter ';'
 func parseProxyURLS() []string {
-	var ret []string
 	if testProxyURLs == "" {
-		return ret
+		return []string{}
```
richardpark-msft, line 33: Generally, `nil` should be the same as empty slice. Would that work here? I think it's more idiomatic.

sdk/internal/perf/testdata/perf/sleep_perf.go
```diff
@@ -58,8 +58,13 @@ func (g *globalSleepPerfTest) NewPerfTest(ctx context.Context, options *perf.Per
...
 
 func (s *sleepPerfTest) Run(ctx context.Context) error {
-	time.Sleep(s.sleepInterval)
-	s.sleepInterval = time.Duration(float64(s.sleepInterval.Nanoseconds())*sleepTestOpts.iterationGrowthFactor) * time.Nanosecond
+	select {
+	case <-ctx.Done():
+		return nil
+	default:
+		time.Sleep(s.sleepInterval)
+		s.sleepInterval = time.Duration(float64(s.sleepInterval.Nanoseconds())*sleepTestOpts.iterationGrowthFactor) * time.Nanosecond
```
richardpark-msft, line 66: This case is very gnarly. I mentioned it above, but if there was a way to just keep it as a time.Duration value it'd probably simplify some things. (perhaps iterationGrowthFactor expects floats though so that'd make it difficult)?

sdk/internal/perf/perf_runner.go
```diff
@@ -0,0 +1,432 @@
...
+	if debug {
+		fmt.Printf("finished %d warmups\n", val)
+	}
+
+	opts.runStart = time.Now()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(duration))
+	defer cancel()
+
+	lastSavedTime := time.Now()
+	for time.Since(opts.runStart).Seconds() < float64(duration) {
```
richardpark-msft, line 404: Just briefly eyeballing this but it looks like warmup and this block are basically identical, except you'd probably parameterize which 'opts.<elapsed>' you'd update. Probably worth consolidating a bit. 

Might also give you a simple testable function afterwards, as well.

richardpark-msft (approved):
Given that this is an internal tool and you're obsessively running it I'm not scrutinizing it too hard at the moment. Just found a few things here and there that are probably worth looking at - if you decide to do it later just file an issue and it's all good.

I'd be interested in if you have some other areas you're going to look into here or if this is pretty much the last drop for awhile. If this is going to go into a less active state we could possibly look at adding more testing.

sdk/internal/perf/atomic_float.go
seankane-msft: This worked out fine, there's a couple extra type conversions that are a bit annoying but it's cleaner than rolling a new atomic type.

sdk/internal/perf/testdata/perf/sleep_perf.go
seankane-msft: iterationGrowthFactor is a `float64` which makes this difficult. I'm going to leave it as is for now.

seankane-msft:
> I'd be interested in if you have some other areas you're going to look into here or if this is pretty much the last drop for awhile. If this is going to go into a less active state we could possibly look at adding more testing.

@richardpark-msft I think this PR is going to put us in a more stable place going forward. I'd love it if you would add servicebus tests. I'm not sure how much overlap there is between your stress tests and these performance tests.

This closes #17177 and #17175

PR #17298, opened 2022-03-14
[azservicebus] Moving nhooyr.io into a file that's not auto-downloaded with the package

richardpark-msft:
Moving nhooyr.io into `testdata` so I don't have to reference it directly in our go.mod file for azservicebus. 

The actual library doesn't use it but it is referenced in our tests and examples so users know how to connect a websocket package up with the client.


sdk/messaging/azservicebus/testdata/examples/example_client_websocket_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"net"
+	"os"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/admin"
+	"github.com/joho/godotenv"
+	"nhooyr.io/websocket"
+)
+
+func Example_usingWebsockets() {
```
jhendrixMSFT, line 18: How important is it that we have a compiling example?  If we really felt compelled to provide an example for this specific implementation, would an entry in the README be sufficient?

richardpark-msft: We could certainly leave it in place but I'd want something a bit more concrete that guarantees we didn't accidentally break it.

As it is it's now just this one example, and the example actually does create the client and test it. So if we're okay with it living in here for now then we should be okay.

seankane-msft: Alternatively, could we include it in the sdk/samples directory? We can add a check to CI to build any test files there

richardpark-msft: @jhendrixMSFT , @seankane-msft :

Feels like we need to have a larger conversation about the whole SDK. 

Are we okay with this solution temporarily and then we can have the bigger  conversation since that affects all packages (and might have a discovery problem)?

CC: @Duffney , since we were literally talking about examples in docs.

seankane-msft: Do you want this example in your documentation? If so, it needs to be in the main package or in the README.md. I'm fine putting it here but if we want the sample readily available to customers it probably belongs in the README.

jhendrixMSFT: I feel like we're having to jump through hoops due to a bad actor that we don't really even care about other than a small example.  Given that the support for web sockets is implementation agnostic, I feel like we should just kick this dependency out for now.  We can always bring it back if we think there's value, once they fix their dependencies.

richardpark-msft: >As long as the interface doesn't change, we shouldn't ever break it though correct?

Yeah, this is less about "me breaking" and more about something changing in the websocket implementation that means it doesn't work at all anymore. It is starting to feel like the external dep is dragging us a bit but this seemed like a better solution than ending up directly depending on it. 

> I feel like we're having to jump through hoops due to a bad actor that we don't really even care about other than a small example. Given that the support for web sockets is implementation agnostic, I feel like we should just kick this dependency out for now. We can always bring it back if we think there's value, once they fix their dependencies.

It's a good point, and TBH while the bad actor thing is a driver it's not the only package that we'd probably want to do this with. 

Azure Identity is another example I have that drags in a non-required dep into go.mod. I reference the package currently so I can show a working example with DefaultAzureCredential(). We can just put these in readmes but it was nice to also take advantage of the natural examples* as well since those gets shown in pkg.go.dev prominently as well. 

In JS, for instance, these would be in our equivalent through samples, where we don't have the same requirements for external deps.

I can delete this sample and migrate it to readme.md for now, but I think we'll need to discuss again at some point.

jhendrixMSFT: In track 1, we had a similar situation with an external dependency for parsing JWTs.  It was clear the project had been abandoned so we ended up using a fork that fixed the CVE.  Fortunately, the license allowed for this.  So I would say, if you can't avoid the dependency, then ensure that its license permits you to fork it in the event it becomes abandoned.

Taking a step back though, I see we have a `require` clause to update the problematic dependency.  What's flagging this as problematic?

richardpark-msft: > In track 1, we had a similar situation with an external dependency for parsing JWTs. It was clear the project had been abandoned so we ended up using a fork that fixed the CVE.

It's MIT licensed, so forks are possible if it comes to that. It's been almost a year since it's last release and I'm not sure we want to own a fork of this. I did just try to use one of the other recommended packages (gorilla/websocket) but it's not as simple as the nhoory.io version, which means we'd need to put in some more effort to swap.

> Taking a step back though, I see we have a require clause to update the problematic dependency. What's flagging this as problematic?

It's not flagging our repo anymore since we have the `require` directive but it does cause the bad dep to get downloaded if someone uses our package (ie, the require only protects us).

I just tried this out and you do end up with gin-gonic 1.6.3, transitively, in my go.sum when referencing v0.3.6:

go mod graph line:
```
nhooyr.io/websocket@v1.8.6 github.com/gin-gonic/gin@v1.6.3
```

jhendrixMSFT: What does `go mod list -m` tell you?  The `go.sum` file will contain all versions evaluated during version selection, even the unused ones, so inspecting the `go.sum` file isn't enough.

richardpark-msft: Talked with @jhendrixMSFT about this and there was a misconception on my part.

TLDR: we can close this PR.

Longer:

`go.sum` here is referencing every visited version of a package. `go list -m all` (slight typo from the comment above) lists the _actual_ referenced package that will get built. When I go get v0.3.6 (the latest azservicebus package) it respects the version of gin-gonic that we've required:

```
github.com/gin-gonic/gin v1.7.7
```

This means we should be okay to just close this PR as it's a cosmetic issue but not an actual problem.

richardpark-msft:
Talked with @jhendrixMSFT about this and there was a misconception on my part.

TLDR: we can close this PR.

Longer:

`go.sum` here is referencing every visited version of a package. `go list -m all` (slight typo from the comment above) lists the _actual_ referenced package that will get built. When I go get v0.3.6 (the latest azservicebus package) it respects the version of gin-gonic that we've required:

```
github.com/gin-gonic/gin v1.7.7
```

This means we should be okay to just close this PR as it's a cosmetic issue but not an actual problem.

PR #17312, opened 2022-03-15
Do not add release stage for scheduled pipelines

benbp:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -192,7 +192,7 @@ stages:
                   Cloud: ${{ cloud.key }}
 
   # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.
-  - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
+  - ${{if and(in(variables['Build.Reason'], 'Manual', ''), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
```
weshaggard, line 195: What is empty Build.Reason?

benbp: According to past you (https://github.com/Azure/azure-sdk-for-net/pull/24132): 

> When you are trying to queue a manual run DevOps sets Build.Reason to empty which blocks the full generation of stages in the stage selection UI. To work around this we check for Build.Reason empty or Manual. This allows for teams to uncheck a release stage for any packages that they don't plan to release.

This is similar to how it is defined in the other repos (e.g. https://github.com/Azure/azure-sdk-for-net/blob/main/eng/pipelines/templates/stages/archetype-net-release.yml#L39).

RickWinter (approved):
Looks reasonable to me ;)

seankane-msft (approved):

tadelesh:
@benbp Mgmt plane use release job to auto check whether a new version need to be released. If it is set to manual, we need to trigger almost 200 rps' pipeline manually if we need to do a round of refresh. Is there any workaround of this or I need to separate the logic?

benbp:
> @benbp Ben Broderick Phillips FTE Mgmt plane use release job to auto check whether a new version need to be released. If it is set to manual, we need to trigger almost 200 rps' pipeline manually if we need to do a round of refresh. Is there any workaround of this or I need to separate the logic?

@tadelesh let's discuss using something like an opt-in parameter? Apologies I missed the context here, I should have added you as a reviewer. I'll reach out on chat.

PR #17318, opened 2022-03-16
[azservicebus] Fixing batch algorithm to calculate accurately

### LINKED ISSUE #15562: [azservicebus] Make our batch size calculation consistent with other implementations
Just a small check - our batch size calculation should match the other implementations.

### LINKED ISSUE #16090: [azservicebus] Batches can exceed link size, causing errors when sending
As part of a test I'm creating messages, adding them to batches and sending them.

It looks like we've got a miscalculation in how the message size gets encoded, which causes a failure in `Sender.SendMessageBatch`:

```go
payload := [1000]byte{0}

for i := 0; i < 1000; i++ {
	added, err := messageBatch.Add(&azservicebus.Message{
		//Body: []byte("hello world"),
		Body: payload[:],
	})

	if added {
		continue
	} else if err == nil {
		log.Printf("Sending batch")

	/*
	panic: encoded message size exceeds max of 262144

	goroutine 1 [running]:
	main.panicky(...)
				/home/codespace/src/_/stress/sdk/messaging/azservicebus/sandbox/sandbox_send_receive.go:138
	main.main()
				/home/codespace/src/_/stress/sdk/messaging/azservicebus/sandbox/sandbox_send_receive.go:65 +0xc53
	exit status 2
	*/
		err := sender.SendMessageBatch(context.Background(), messageBatch)
		panicky(err)

		messageBatch, err = sender.NewMessageBatch(context.Background(), nil)
		panicky(err)
	}

	panicky(err)
}
```

Depending on whether we have an encoding bug, this might be fixed by reconciling our batch calculation in https://github.com/Azure/azure-sdk-for-go/issues/15562

richardpark-msft:
Message batches were being calculated incorrectly, using an assumed fixed size overhead which wasn't accurate. We're now doing this dynamically. 

This also made it so I could fix a stress test that's been broken because of the batch size.

Fixes #16090
Fixes #15562

### DIFF
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 ### Features Added
 
-- Support for using a SharedAccessSignature in a connection string. Ex: `Endpoint=sb://<sb>.servicebus.windows.net;SharedAccessSignature=SharedAccessSignature sr=<sb>.servicebus.windows.net&sig=<base64-sig>&se=<expiry>&skn=<keyname>` (#TBD)
+- Support for using a SharedAccessSignature in a connection string. Ex: `Endpoint=sb://<sb>.servicebus.windows.net;SharedAccessSignature=SharedAccessSignature sr=<sb>.servicebus.windows.net&sig=<base64-sig>&se=<expiry>&skn=<keyname>` (#17314)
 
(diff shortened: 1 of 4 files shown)


seankane-msft (approved):
LGTM :shipit:

jhendrixMSFT (approved):

PR #17321, opened 2022-03-16
[Blob] Return all clients by ptr

seankane-msft:


seankane-msft:
@mohsha-msft you should transition the `mohsha-msft/bug-shared-cred` branch into a feature branch or merge it into main. There are no CI checks that run when you use this branch.

PR #17326, opened 2022-03-17
[azservicebus] Fixing a few errors related to either cleanup or closing

### LINKED ISSUE #17183: [azservicebus] Client.Close() needs to close the namespace.Client()
The client close is closing all the links, leaving an idle connection but it must _also_ close the client as well or else it'll leave the connection open.

There is some potential lock interaction here with the amqpLinks recovery, so we need to be careful to avoid deadlock if:

1. amqpLinks is recovering the connection (thus closing namespace.Client)
2. user calls client.Close()
3. client.Close() tries to close namespace.Client() as well as all the child links

richardpark-msft:
Fixing a few errors related to either cleanup or closing:
- negotiateClaim() would cycle infinitely if there was a fatal error, rather than exiting
- the func() returned from negotiateClaim would return before the claim loop had completely exited. Not fatal, but the intention was that there was no further claim activity after the cancel func was called. Fixed.
 - Connection was not being by client.Close(). Also, added in a "permanent close" error that gets returned from all the normal spots.
- Fixing some race conditions that were in the _test_ code

Fixes #17183

### DIFF
--- a/sdk/messaging/azservicebus/internal/namespace.go
+++ b/sdk/messaging/azservicebus/internal/namespace.go
@@ -50,6 +50,9 @@ type (
 		clientMu         sync.RWMutex
 		client           *amqp.Client
 		negotiateClaimMu sync.Mutex
+		// indicates that the client was closed permanently, and not just
+		// for recovery.
+		closedPermanently bool
 	}
 
 	// NamespaceOption provides structure for configuring a new Service Bus namespace
@@ -59,16 +62,17 @@ type (
 // NamespaceWithNewAMQPLinks is the Namespace surface for consumers of AMQPLinks.
 type NamespaceWithNewAMQPLinks interface {
 	NewAMQPLinks(entityPath string, createLinkFunc CreateLinkFunc) AMQPLinks
+	Check() error
 }
 
 // NamespaceForAMQPLinks is the Namespace surface needed for the internals of AMQPLinks.
 type NamespaceForAMQPLinks interface {
-	NegotiateClaim(ctx context.Context, entityPath string) (func() <-chan struct{}, error)
+	NegotiateClaim(ctx context.Context, entityPath string) (context.CancelFunc, <-chan struct{}, error)
 	NewAMQPSession(ctx context.Context) (AMQPSessionCloser, uint64, error)
 	NewRPCLink(ctx context.Context, managementPath string) (RPCLink, error)
 	GetEntityAudience(entityPath string) string
 	Recover(ctx context.Context, clientRevision uint64) (bool, error)
-	Close(ctx context.Context) error
+	Close(ctx context.Context, permanently bool) error
 }
 
 // NamespaceWithConnectionString configures a namespace with the information provided in a Service Bus connection string
@@ -232,9 +236,13 @@ func (ns *Namespace) NewAMQPLinks(entityPath string, createLinkFunc CreateLinkFu
 }
 
 // Close closes the current cached client.
-func (ns *Namespace) Close(ctx context.Context) error {
+func (ns *Namespace) Close(ctx context.Context, permanently bool) error {
 	ns.clientMu.Lock()
 	defer ns.clientMu.Unlock()
+
+	if permanently {
+		ns.closedPermanently = true
+	}
 
 	if ns.client != nil {
 		return ns.client.Close()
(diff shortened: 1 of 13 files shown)


richardpark-msft:
Ran gofmt and also fixed some race conditions in the test code itself (around when/how to activate logging)

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -50,6 +50,9 @@ type (
 		clientMu         sync.RWMutex
 		client           *amqp.Client
 		negotiateClaimMu sync.Mutex
+		// indicates that the client was closed permanently, and not just
+		// for recovery.
+		closedPermanently bool
```
jhendrixMSFT, line 55: I have mixed feelings on this.  Shouldn't the owner of ns know if it was closed for recovery vs really closed?

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -287,19 +317,23 @@ func (ns *Namespace) Recover(ctx context.Context, theirConnID uint64) (bool, err
 
 // negotiateClaim performs initial authentication and starts periodic refresh of credentials.
 // the returned func is to cancel() the refresh goroutine.
-func (ns *Namespace) NegotiateClaim(ctx context.Context, entityPath string) (func() <-chan struct{}, error) {
+func (ns *Namespace) NegotiateClaim(ctx context.Context, entityPath string) (context.CancelFunc, <-chan struct{}, error) {
```
jhendrixMSFT, line 320: I don't think this needs to be split into three return args.  Looks like the bug was introduced in the track 2 migration.  See the [track 1](https://github.com/Azure/azure-service-bus-go/blob/master/namespace.go#L293-L296) impl where we closed the channel when the refresh goroutine exits.

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: The namespace and the client are really 1:1. I just let the namespace hold the flag since it's basically the authority. So it's the same concept, unless I'm missing something.

sdk/messaging/azservicebus/internal/namespace.go
richardpark-msft: This change was less about a bug and more about just wanting to make it simpler to do the common operations. Basically, when I cancel I never had a case where I intentionally was trying to do it and walk way - I always wanted to block and wait until the refresh loop exits.

I also had a use case, in tests, where I wanted to make sure that the refresh loop exits without me explicitly cancelling it (ie, fatal recovery error).

So there's some internal expedience, and also just making it simpler to use it in the way that I actually used it.

jhendrixMSFT: With the old impl, you would write `<-cancelFun()` to wait until the exit channel was closed (so it wasn't fire-and-forget).  But you have a good point about wanting to detect the refresh goroutine exiting due to a terminal failure, so I can see the sense in breaking this apart.

sdk/messaging/azservicebus/internal/namespace.go
jhendrixMSFT: Whenever I see a "really the thing" flag it gives me pause for concern.  Seems like the API is performing double-duty, masquerading as two slightly different APIs.  How does a person know when to write Close(true) or Close(false)?  I know this is all internal stuff so maybe it's a bit academic.

richardpark-msft: It's _very_ internal (and it's a good question):

Ignoring the bool part, the non-permanent-close is when we're doing recovery and expect that a later call will reconstitute/reopen the link. So if anyone is in the middle of a retry loop the next call will open it back up.

Close permanently allows us (and other callers) to know that retries should end - we're in a locally non-retriable state. This will exit any retry loops.

Right now the way the code is written is that each caller is basically doing a liveness check when they get the client or use the client. This is just an extension of that logic.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/internal/namespace.go
jhendrixMSFT: Makes sense given that we do lazy recreation.

PR #17352, opened 2022-03-22
[KeyVault] regenerate azcertificates from 7.3 swagger

seankane-msft:
regenerating from latest GA swagger


mccoyp (approved):
It looks like recordings were manually updated instead of being generated by new test runs -- is that for a particular reason?

seankane-msft:
> It looks like recordings were manually updated instead of being generated by new test runs -- is that for a particular reason?

Not a great one, running a live test to verify the tests pass.

PR #17353, opened 2022-03-22
[KeyVault] regenerating azsecrets for 7.3 swagger

seankane-msft:
Regenerate azsecrets for latest GA swagger.


sdk/keyvault/azsecrets/autorest.md
```diff
@@ -6,13 +6,13 @@ These settings apply only when `--go` is specified on the command line.
...
+- https://github.com/Azure/azure-rest-api-specs/blob/e2ef44b87405b412403ccb005bfb3975411adf60/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.3/secrets.json
 license-header: MICROSOFT_MIT_NO_VERSION
 clear-output-folder: true
 output-folder: internal
 tag: package-7.2
 credential-scope: none
-use: "@autorest/go@4.0.0-preview.36"
-module-version: 0.6.0
+use: "@autorest/go@4.0.0-preview.37"
+module-version: 0.6.1
```
jhendrixMSFT, line 16: Looks like there are some breaking changes?  If so, this should be `0.7.0`.

sdk/keyvault/azsecrets/client.go
```diff
@@ -218,32 +216,36 @@ type DeleteSecretPoller struct {
...
 }
 
 // Poll fetches the latest state of the LRO. It returns an HTTP response or error.(
 // If the LRO has completed successfully, the poller's state is updated and the HTTP response is returned.
 // If the LRO has completed with failure or was cancelled, the poller's state is updated and the error is returned.)
 func (s *DeleteSecretPoller) Poll(ctx context.Context) (*http.Response, error) {
+	var rawResp *http.Response
+	ctx = runtime.WithCaptureResponse(ctx, &rawResp)
 	resp, err := s.client.GetDeletedSecret(ctx, s.vaultUrl, s.secretName, nil)
+	s.rawResponse = rawResp
```
jhendrixMSFT, line 237: I'm not sure what `GetDeletedSecret` does under the hood, but could there be a case where `err != nil` and `rawResp` is `nil`?  So you'd overwrite the last response with a `nil` one.  Maybe this is ok, just pointing it out.

jhendrixMSFT (changes requested):
The removal of raw response requires a minor version bump.

sdk/keyvault/azsecrets/autorest.md
seankane-msft: Fixed

Issue #17351, opened 2022-03-22
Restore auto-release stage behavior for mgmt plane pipelines


benbp:
See https://github.com/Azure/azure-sdk-for-go/pull/17312#issuecomment-1074998790

PR #17356, opened 2022-03-23
ci: add auto release flag for mgmt. plane release

tadelesh:

Resolve https://github.com/Azure/azure-sdk-for-go/issues/17351

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -68,6 +68,9 @@ parameters:
   - name: TestRunTime
     type: string
     default: '600s'
+  - name: AutoRelease
+    type: boolean
+    default: false
 
 
 stages:
@@ -192,8 +195,11 @@ stages:
                   Location: ${{ coalesce(parameters.Location, cloud.value.Location) }}
                   Cloud: ${{ cloud.key }}
 
-  # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.
-  - ${{if and(in(variables['Build.Reason'], 'Manual', ''), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
+  # The Prerelease and Release stages are conditioned on:
+  # 1. Internal trigger
+  # 2. Not weekly build
+  # 3. Manual trigger (include manual defined Build.Reason param with empty) / Non-PR trigger with 'AutoRelease' flag
+  - ${{if and(or(in(variables['Build.Reason'], 'Manual', ''), and(ne(variables['Build.Reason'], 'PullRequest'), eq(parameters.AutoRelease, 'true'))), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
     - template: archetype-go-release.yml
       parameters:
         DependsOn:
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -68,7 +68,7 @@ parameters:
   - name: TestRunTime
     type: string
     default: '600s'
-  - name: AutoRelease
+  - name: IncludeRelease
     type: boolean
     default: false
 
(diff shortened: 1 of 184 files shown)


tadelesh:
Test for release: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1453746&view=results

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -192,8 +195,11 @@ stages:
                   Location: ${{ coalesce(parameters.Location, cloud.value.Location) }}
                   Cloud: ${{ cloud.key }}
 
-  # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.
-  - ${{if and(in(variables['Build.Reason'], 'Manual', ''), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
+  # The Prerelease and Release stages are conditioned on:
+  # 1. Internal trigger
+  # 2. Not weekly build
+  # 3. Manual trigger (include manual defined Build.Reason param with empty) / Non-PR trigger with 'AutoRelease' flag
+  - ${{if and(or(in(variables['Build.Reason'], 'Manual', ''), and(ne(variables['Build.Reason'], 'PullRequest'), eq(parameters.AutoRelease, 'true'))), not(contains(variables['Build.DefinitionName'], 'weekly')), eq(variables['System.TeamProject'], 'internal'))}}:
```
weshaggard, line 202: This condition is definitely starting to become impossible to understand. @benbp can you give more context on why you made the change in https://github.com/Azure/azure-sdk-for-go/pull/17312? 

Should we consider doing auto release check behavior for all go libraries both mgmt and data-plane?

benbp: @weshaggard I made this change based on an ask from @RickWinter and @seankane-msft to hide the release stage for builds that should not release, with the idea that it would make release tracking/pipeline history easier to read, reduce confusion, and be consistent with the pipeline UI for other languages.

I think adding `AutoRelease` does enable it for all go libraries, where the data plane ones just default to false and so don't need to specify it. We would still want to do the other checks here like build reason, etc. if we wanted to preserve the release stage hiding behavior.

benbp:
Looks good to me but holding approval until we reach a consensus on https://github.com/Azure/azure-sdk-for-go/pull/17356#discussion_r833463975

eng/pipelines/templates/jobs/archetype-sdk-client.yml
weshaggard: I do think it might make it a little easier to understand and read if we can put some of the condition as the default value for. Perhaps we just keep the internal and AutoRelease condition here and move the others to the default value.

Also I think AutoRelease is a little misleading, we should probably name this to IncludeRelease.

tadelesh:
@seankane-msft could you have a review as I [exclude the doc validation](https://github.com/Azure/azure-sdk-for-go/pull/17356/commits/0a893133acd56b5ad5b773ef27ea377bd86bc5d9) for the eng/tools folder?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
tadelesh: I don't quite get the idea of the default value. Is that mean we just use another variable to judge other condition except internal and AutoRelease?

RickWinter:
@benbp @weshaggard What does AutoRelease do?  Does that mean a human never clicks the final approval?  Do we have sufficient permissions and checks in place to prevent unintentional releases?

RickWinter (changes requested):
We need to ensure proper safeguards are in place to prevent unintentional triggers and resulting releases.

tadelesh:
> @benbp @weshaggard What does AutoRelease do? Does that mean a human never clicks the final approval? Do we have sufficient permissions and checks in place to prevent unintentional releases?

The word might be a little bit confusing. Actually `Include Release` might be more accurate. When it is set to true, the release pipeline will be trigger. Then a pre-release job will check if there is a new version need to be released. If true, the release job will do the changelog check, etc., then WAITING FOR APPROVAL for the truly release. This is used for auto trigger, not for approval the truly release as we want to just do one final approval instead of too many manual work (find pipeline, manual trigger, waiting result, approve the release) for mgmt. plane release. The pipeline has working fine for a long time (Here is the original [PR](https://github.com/Azure/azure-sdk-for-go/pull/16207) I added it) until the latest change of the condition. Ben and I are trying to fix it with introducing a new flag.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
benbp: I agree re `IncludeRelease` naming because having the release included doesn't guarantee that it will be auto released due to the "verify release" stage.

We were chatting offline and @weshaggard had the idea that maybe we could improve the readability of the conditionals by using nested ifs (since I don't think devops supports multi-line expressions). The downside is that the template below has a higher indentation level, but I think that's fine.

```
  - ${{ if and(ne(variables['Build.Reason'], 'PullRequest'], eq(variables['System.TeamProject'], 'internal')) }}:
    - ${{ if not(contains(variables['Build.DefinitionName'], 'weekly')) }}:
      - ${{ if or(eq(variables['Build.Reason'], 'Manual'), eq(${{ parameters.IncludeRelease }}, true)) }}:
```

By "default value" we were thinking about trying to declare some of the conditional expressions in a parameter declaration as the default value, but I'm not sure devops would evaluate that properly at template compile time.

tadelesh: Looks great. I've changed the PR according to this. BTW, the final if for `Manual` part, I use `in(variables['Build.Reason'], 'Manual', '')` to keep the same as your previous logic.

tadelesh: One single test: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1460469&view=results

weshaggard (approved):
Looks reasonable to me. However I defer to @RickWinter on whether or not there are other considerations.

RickWinter (approved):

PR #17306, opened 2022-03-15
[EngSys] adding ci checks for performance

seankane-msft:
* Runs `go build` and `go vet` on the testdata/perf directory if it exists

PR #17365, opened 2022-03-24
Fix performance test build task

benbp:
Fixes invalid yaml syntax introduced in #17306 (related to #17364 as well).

The template pipeline triggers against `eng/` changes to catch potential issues. If there are syntactic errors in the yaml then the pipeline "fails to run" which isn't a valid criteria/event to get posted back to the check runs.

For example: [failing pipeline](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1456127&view=results)

Currently, the only way to verify eng pipeline yaml changes are valid would be to manually invoke the pipeline like `/azp run go - aztemplate` and make sure it was successfully run. I'm not sure if it's going to be possible to get the check runs to detect pipeline initialization failures.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -62,7 +62,7 @@ steps:
       ${{ insert }}: ${{ parameters.EnvVars }}
 
   - task: PowerShell@2
-    display: 'Build Performance Tests'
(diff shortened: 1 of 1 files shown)


benbp:
FYI @weshaggard if you know of/can think of any workarounds to the problem detailed in this PR description.

seankane-msft (approved):

PR #17374, opened 2022-03-24
Fix pkg.go.dev link for certificates README

kimprice:
The link right now is going to the Python library. Also fixed a couple errors in the sample code.

### DIFF
--- a/sdk/keyvault/azcertificates/README.md
+++ b/sdk/keyvault/azcertificates/README.md
@@ -123,7 +123,7 @@ func main() {
 		panic(err)
 	}
 
-	client, err = azcertificates.NewClient("https://my-key-vault.vault.azure.net/", credential, nil)
+	client, err := azcertificates.NewClient("https://my-key-vault.vault.azure.net/", credential, nil)
 	if err != nil {
 		panic(err)
 	}
@@ -162,7 +162,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	client, err = azcertificates.NewClient("https://my-key-vault.vault.azure.net/", credential, nil)
+	client, err := azcertificates.NewClient("https://my-key-vault.vault.azure.net/", credential, nil)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azcertificates/README.md
```diff
@@ -450,12 +450,12 @@ This project has adopted the [Microsoft Open Source Code of Conduct][code_of_con
 [azure_sub]: https://azure.microsoft.com/free/
 [code_of_conduct]: https://opensource.microsoft.com/codeofconduct/
 [keyvault_docs]: https://docs.microsoft.com/azure/key-vault/
-[pkggodev_azcerts]: https://pypi.org/project/azure-keyvault-certificates/
+[pkggodev_azcerts]: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azcertificates
 [certificate_client_docs]: https://aka.ms/azsdk/go/azcertificates
 [rbac_guide]: https://docs.microsoft.com/azure/key-vault/general/rbac-guide
 [reference_docs]: https://aka.ms/azsdk/go/azcertificates
```
kimprice, line 456: I'm also wondering about the reference docs link. In some of our other READMEs the API Reference is a link to the pkg.go.dev, so it seems redundant here.

kimprice:
I wonder if we should also have a general README at the root for KeyVault with just basic information on KeyVault like what's at the beginning of this README. Then in the READMEs for the individual package, they would just focus on that package and the service docs linked could be for that specific Key Vault management.

seankane-msft (approved):

seankane-msft:
> I wonder if we should also have a general README at the root for KeyVault with just basic information on KeyVault like what's at the beginning of this README. Then in the READMEs for the individual package, they would just focus on that package and the service docs linked could be for that specific Key Vault management.

I checked the other language repos, they don't do this but I think it could be worth leading the charge here.

kimprice:
> > I wonder if we should also have a general README at the root for KeyVault with just basic information on KeyVault like what's at the beginning of this README. Then in the READMEs for the individual package, they would just focus on that package and the service docs linked could be for that specific Key Vault management.
> 
> I checked the other language repos, they don't do this but I think it could be worth leading the charge here.

I guess the only downside is that it wouldn't technically be part of a package. So maybe what we need to do is have a consistent intro section for all the Key Vault libraries that starts by introducing Key Vault and describes the three separate packages (exactly like what's in this README).

Issue #17017, opened 2022-02-09
[azservicebus] Timeout in ReceiveMessages


praminda:
### Bug Report

- pkg: `github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus`
- SDK version: `v0.3.2`
- go version: `go 1.16.2`

- What happened?
We recently moved from `azure-service-bus-go` and implemented an asb client using `azure-sdk-for-go`. This implementation was working fine for few days and then started to give below timeout error frequently. A manual service restart seem to fix the issue temporarily, but after some time error appears again in the logs. Our implementation is supposed to retry retrieving messages indefinitely but it doesn't seem to recover from the error and getting the messages from the topic.
Pls note we are subscribing to multiple topics using multiple clients here.
You can find our client implementation [here](https://github.com/wso2/product-microgateway/blob/348cad1cd2dbedf32672234c4cdb1990462dfd3e/adapter/pkg/messaging/azure_listener.go#L70).
**Error:**

    ```
    Failed to receive messages from ASB. read tcp 1*.*.*.113:xxxxxx->**.*.*.88:xxxx: i/o timeout
    ```

- What did you expect or want to happen?
If something unexpected causes this timeout error we expect it to recover when retrying.

- Anything we should know about your environment.
Our service is running on AKS

praminda:
Thanks for the fix @richardpark-msft. 
When we will be able to get the fixed pkg released?

pubudu538:
Hi team, Can we get a release with the latest fixes?

paul2u:
Thank you @richardpark-msft for checking this issue. Would you update us the latest status? Thank you so much for you help.

richardpark-msft:
(sorry for the late response, was on vacation)

For this fix you can do a `go get` on the commit hash.

```bash
go get github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus@ad8d573a3cea4dd0f094550a838d56fc3ecbcddf
```

pubudu538:
Hi @richardpark-msft , 

We did the version upgrade based on your above comment. But still we see the connectivity issues. This time we don't see any issues in the application side. The app stopped receiving messages from ASB. When we check the subscription, we can see the queued messages. Could you please help with this?

Thank you!

pubudu538:
---- Update ----

We thought of doing upgrading the client sdk to [1] as it has fixed some of the issues. 

[...]

PR #17382, opened 2022-03-25
[azservicebus] Handle 410 properly for session and non-session based links.

### LINKED ISSUE #17325: [azservicebus] Sessions detach can be inconsistent
After a link detach it seems like there's a delay between the session lock expiring and me being able to reconnect to it.

This makes it so recovery is difficult - our current error code just assumes that attempting to open a session either works because we obtain the lock, or fails because the session is already locked. It doesn't attempt to account for a potential delay until the session becomes available again.

I need to investigate this a bit more - I have a simple test that just does a detach on an entity and then attempts to recover the link.

This is related to me wanting to fix the issue with message completion retrying unnecessarily when the lock has expired.

richardpark-msft:
410 (lock lost) handling needed to be separated for sessions and non-session links.

For non-session links losing a link is essentially unrecoverable - the lock cannot be reobtained with just the lock token so further retries will always fail.

For _session_ based links it's different since you actually obtain a session lock by opening a link for a particular session ID. You can retry re-opening the link and there are cases where, on detach, that you might have some overlap until the session is actually available again (this is shown in the `TestSessionReceiver_Detach` test).

So the split here is simple - all retrying is really handled in amqpLinks, so now it takes a parameter giving it the proper function to determine if an error is fatal or retryable. As part of this I also made the parameters to amqpLinks an arg, which had some small rippling effects in callers.

Fixes #17325
(probable fix for the latest issue mentioned in #17017 since this affects settlement time)

### DIFF
--- a/sdk/messaging/azservicebus/internal/errors.go
+++ b/sdk/messaging/azservicebus/internal/errors.go
@@ -26,16 +26,18 @@ func NewErrNonRetriable(message string) error {
 
 func (e errNonRetriable) Error() string { return e.Message }
 
-type recoveryKind string
+// RecoveryKind dictates what kind of recovery is possible. Used with
+// GetRecoveryKind().
+type RecoveryKind string
 
-const RecoveryKindNone recoveryKind = ""
-const RecoveryKindFatal recoveryKind = "fatal"
-const RecoveryKindLink recoveryKind = "link"
-const RecoveryKindConn recoveryKind = "connection"
+const RecoveryKindNone RecoveryKind = ""
+const RecoveryKindFatal RecoveryKind = "fatal"
+const RecoveryKindLink RecoveryKind = "link"
+const RecoveryKindConn RecoveryKind = "connection"
 
 type SBErrInfo struct {
 	inner        error
-	RecoveryKind recoveryKind
+	RecoveryKind RecoveryKind
 }
 
 func (sbe *SBErrInfo) String() string {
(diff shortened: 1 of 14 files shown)


seankane-msft (approved):
LGTM

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -176,15 +190,13 @@ func GetRecoveryKind(err error) recoveryKind {
...
+		if code == RPCResponseCodeNotFound || code == RPCResponseCodeLockLost {
 			return RecoveryKindFatal
 		}
 
 		// this can happen when we're recovering the link - the client gets closed and the old link is still being
 		// used by this instance of the client. It needs to recover and attempt it again.
-		if code == 401 ||
-			// we lost the session lock, attempt link recovery
-			code == 410 {
+		if code == 401 {
```
jhendrixMSFT, line 199: Add a named `const` for 401?

jhendrixMSFT (approved):
Couple nits.

sdk/messaging/azservicebus/internal/errors.go
richardpark-msft: Sounds good!

richardpark-msft: I did some looking - these actually are supposed to map to HTTP status codes! I swapped out all of the constants for the http status code except for `RPCResponseCodeLockLost` since that one maps to StatusCodeGone and isn't necessarily intuitive like the others.

PR #17397, opened 2022-03-28
[Keyvault] `azsecrets` GA readiness

seankane-msft:


sdk/keyvault/azsecrets/models.go
```diff
@@ -52,49 +35,76 @@ type Secret struct {
...
 
-	// READ-ONLY; Creation time in UTC.
-	CreatedOn *time.Time `json:"created,omitempty" azure:"ro"`
+	// READ-ONLY; If this is a secret backing a KV certificate, then this field specifies the corresponding key backing the KV
+	// certificate.
+	KeyID *string `json:"kid,omitempty" azure:"ro"`
 
-	// READ-ONLY; Last updated time in UTC.
-	UpdatedOn *time.Time `json:"updated,omitempty" azure:"ro"`
+	// Not before date in UTC.
```
chlowell, line 83: Doc comments should begin with the field name
```suggestion
	// NotBefore is the secret's not before date in UTC.
```

sdk/keyvault/azsecrets/README.md
```diff
@@ -171,15 +171,15 @@ func main() {
 		panic(err)
 	}
 
-	secretName := "mySecretName"
-	secretValue := "mySecretValue"
+	secret := "mySecret"
+	value := "mySecretValue"
```
heaths, line 175: I thought these were going to change to `name` and `secret` (or `value` is also fine) per archboard review. As written, this was confusing.

heaths (commented):
APIs LGTM sans the few changes we talked about in APIView offline. I'll wait to see those to sign off.

sdk/keyvault/azsecrets/client.go
```diff
@@ -605,9 +676,15 @@ func (b *RecoverDeletedSecretPoller) pollUntilDone(ctx context.Context, t time.D
 	return recoverDeletedSecretResponseFromGenerated(b.recoverResponse), nil
 }
 
+// ResumeToken returns a token for resuming polling at a later time
+func (s *RecoverDeletedSecretPoller) ResumeToken() (string, error) {
+	return string(s.resumeToken), nil
```
chlowell, line 681: `resumeToken` is a string?
```suggestion
	return s.resumeToken, nil
```

sdk/keyvault/azsecrets/internal/keyvault_client.go
```diff
@@ -33,7 +34,7 @@ func NewKeyVaultClient(options *azcore.ClientOptions) *KeyVaultClient {
 		options = &azcore.ClientOptions{}
 	}
 	client := &KeyVaultClient{
-		pl: runtime.NewPipeline(moduleName, moduleVersion, runtime.PipelineOptions{}, options),
+		Pl: runtime.NewPipeline(ModuleName, ModuleVersion, runtime.PipelineOptions{}, options),
```
chlowell, line 37: This gives requests the wrong User-Agent: `azsdk-go-internal/v0.7.0 ...`

sdk/keyvault/azsecrets/models.go
```diff
@@ -10,39 +10,26 @@ import (
...
 	// The secret management attributes.
 	Properties *Properties `json:"attributes,omitempty"`
 
-	// The content type of the secret.
-	ContentType *string `json:"contentType,omitempty"`
-
 	// The secret id.
 	ID *string `json:"id,omitempty"`
 
+	// The name of the string
```
chlowell, line 24:
```suggestion
	// Name of the secret
```

sdk/keyvault/azsecrets/client.go
```diff
@@ -330,93 +379,107 @@ type GetDeletedSecretResponse struct {
...
 	}
 
 	resp, err := c.kvClient.UpdateSecret(
 		ctx,
 		c.vaultUrl,
-		secretName,
-		options.Version,
-		options.toGeneratedProperties(),
-		options.toGenerated(),
+		name,
```
chlowell, line 479: If a caller constructed the `Secret` instead of getting it from another method, `name` could be empty, in which case the request can't succeed. I guess the vault would respond 404, maybe with a baffling HTML payload; the caller could ultimately get a serialization error. Should we forego the request and return an error when `name == ""`?

sdk/keyvault/azsecrets/internal/keyvault_client.go
seankane-msft: fixed to `azsdk-go-azsecrets/v0.7.0` is that correct?

sdk/keyvault/azsecrets/client.go
seankane-msft: There is a check in the generated code that validates name is not empty. It returns `errors.New("parameter secretName cannot be empty")`

sdk/keyvault/azsecrets/models.go
```diff
@@ -52,49 +39,79 @@ type Secret struct {
...
-	// The content type of the secret.
-	ContentType *string `json:"contentType,omitempty"`
-
 	// The secret id.
 	ID *string `json:"id,omitempty"`
 
-	// Application specific metadata in the form of key-value pairs.
-	Tags map[string]string `json:"tags,omitempty"`
+	// The name of the secret
+	Name *string
```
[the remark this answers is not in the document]
seankane-msft: If there was an error with the request this will be nil, like all other properties. It technically doesn't have to be a pointer but I made it a pointer for consistency reasons

sdk/keyvault/azsecrets/internal/keyvault_client.go
chlowell: I believe so. @kimprice @mikekistler is `azsdk-go-azsecrets/v0.7.0 azsdk-go-azcore/v0.23.0 (go1.18; Windows_NT)` the right format for our telemetry?

sdk/keyvault/azsecrets/client.go
chlowell: That's good, if awkward because there's no public `secretName`. I wouldn't block the release on the error message in this unlikely scenario though

sdk/keyvault/azsecrets/testdata/recordings/TestBeginDeleteSecretRehydrated.json
```diff
@@ -0,0 +1,732 @@
+{
+  "Entries": [
+    {
+      "RequestUri": "https://fakekvurl.vault.azure.net/secrets/begindeleterehydrat3511093089?api-version=7.3",
+      "RequestMethod": "PUT",
+      "RequestHeaders": {
+        "Accept": "application/json",
+        "Accept-Encoding": "gzip",
+        "Content-Length": "0",
+        "User-Agent": "azsdk-go-internal/v0.7.0 azsdk-go-azcore/v0.23.0 (go1.18; Windows_NT)"
```
chlowell, line 10: Shouldn't the test fail now that the client's User-Agent doesn't match the recording?

chlowell (approved):

sdk/keyvault/azsecrets/testdata/recordings/TestBeginDeleteSecretRehydrated.json
seankane-msft: By default we don't check the User-Agent header because then we'd have to regen for every version update to core and each service.

sdk/keyvault/azsecrets/internal/keyvault_client.go
kimprice: That looks right according to this: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

PR #17398, opened 2022-03-28
[Keyvault] `azkeys` GA preparation

seankane-msft:


sdk/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,18 +1,12 @@
 # Release History
 
-## 0.4.1 (Unreleased)
-
-### Features Added
+## 0.5.0 (2022-04-05)
 
 ### Breaking Changes
 * `ListDeletedKeysPager` has `More() bool` and `NextPage(context.Context) (ListDeletedKeysPage, error)` for paging over deleted keys.
 * `ListPropertiesOfKeyVersionssPager` has `More() bool` and `NextPage(context.Context) (ListPropertiesOfKeyVersionsPage, error)` for paging over deleted keys.
```
chlowell, line 7:
```suggestion
* `ListPropertiesOfKeyVersionsPager` has `More() bool` and `NextPage(context.Context) (ListPropertiesOfKeyVersionsPage, error)` for paging over deleted keys.
```

sdk/keyvault/azkeys/crypto/crypto_client.go
```diff
@@ -120,22 +120,22 @@ func NewClient(keyURL string, credential azcore.TokenCredential, options *Client
...
 	// Additional data to authenticate but not encrypt/decrypt when using authenticated crypto algorithms.
-	AAD []byte `json:"aad,omitempty"`
+	AuthData []byte `json:"aad,omitempty"`
 
 	// Initialization vector for symmetric algorithms.
 	IV []byte `json:"iv,omitempty"`
 
 	// The tag to authenticate when performing decryption with an authenticated algorithm.
-	Tag []byte `json:"tag,omitempty"`
+	AuthenticationTag []byte `json:"tag,omitempty"`
```
chlowell, line 129:
```suggestion
	AuthTag []byte `json:"tag,omitempty"`
```

sdk/keyvault/azkeys/models.go
```diff
@@ -23,22 +27,34 @@ type Properties struct {
...
-	RecoveryLevel *DeletionRecoveryLevel `json:"recoveryLevel,omitempty" azure:"ro"`
+	RecoveryLevel *string `json:"recoveryLevel,omitempty" azure:"ro"`
+
+	// Application specific metadata in the form of key-value pairs.
+	Tags map[string]string `json:"tags,omitempty"`
 
 	// READ-ONLY; Last updated time in UTC.
 	UpdatedOn *time.Time `json:"updated,omitempty" azure:"ro"`
+
+	VaultURL *string
```
chlowell, line 55: Missing doc comments

sdk/keyvault/azkeys/models.go
```diff
@@ -23,22 +27,34 @@ type Properties struct {
...
 	// READ-ONLY; softDelete data retention days. Value should be >=7 and <=90 when softDelete enabled, otherwise 0.
 	RecoverableDays *int32 `json:"recoverableDays,omitempty" azure:"ro"`
 
 	// READ-ONLY; Reflects the deletion recovery level currently in effect for keys in the current vault. If it contains 'Purgeable'
 	// the key can be permanently deleted by a privileged user; otherwise, only the system
 	// can purge the key, at the end of the retention interval.
-	RecoveryLevel *DeletionRecoveryLevel `json:"recoveryLevel,omitempty" azure:"ro"`
+	RecoveryLevel *string `json:"recoveryLevel,omitempty" azure:"ro"`
+
+	// Application specific metadata in the form of key-value pairs.
```
chlowell, line 49:
```suggestion
	// Tags contains application specific metadata in the form of key-value pairs.
```

sdk/keyvault/azkeys/models.go
```diff
@@ -87,11 +109,30 @@ type Key struct {
 	// The policy rules under which the key can be exported.
 	ReleasePolicy *ReleasePolicy `json:"release_policy,omitempty"`
 
-	// Application specific metadata in the form of key-value pairs.
-	Tags map[string]string `json:"tags,omitempty"`
+	// Key identifier.
```
chlowell, line 112: Something like
```suggestion
	// ID identifies the key
```

sdk/keyvault/azkeys/models.go
```diff
@@ -290,7 +316,7 @@ func deletedKeyItemFromGenerated(i *generated.DeletedKeyItem) *DeletedKeyItem {
 			CreatedOn:       i.Attributes.Created,
 			UpdatedOn:       i.Attributes.Updated,
 			RecoverableDays: i.Attributes.RecoverableDays,
-			RecoveryLevel:   (*DeletionRecoveryLevel)(i.Attributes.RecoveryLevel),
+			RecoveryLevel:   (*string)(i.Attributes.RecoveryLevel),
```
chlowell, line 319: Would this work?
```suggestion
			RecoveryLevel:   to.Ptr(i.Attributes.RecoveryLevel),
```

sdk/keyvault/azkeys/models.go
```diff
@@ -351,11 +377,32 @@ type RotationPolicy struct {
...
+		Attributes:      attribs,
+	}
+}
+
 // RotationPolicyAttributes - The key rotation policy attributes.
 type RotationPolicyAttributes struct {
-	// The expiryTime will be applied on the new key version. It should be at least 28 days. It will be in ISO 8601 Format. Examples: 90 days: P90D, 3 months:
-	// P3M, 48 hours: PT48H, 1 year and 10 days: P1Y10D
-	ExpiryTime *string `json:"expiryTime,omitempty"`
+	// The expiryTime will be applied on the new key version. It should be at least 28 days.
```
chlowell, line 402:
```suggestion
	// ExpiresIn will be applied on the new key version. It should be at least 28 days.
```

sdk/keyvault/azkeys/utils_test.go
```diff
@@ -203,3 +203,55 @@ func toBytes(s string, t *testing.T) []byte {
...
+
+func validateProperties(t *testing.T, props *Properties) {
+	require.NotNil(t, props)
+	if props.CreatedOn == nil {
+		t.Fatalf("expected CreatedOn to be not nil")
+	}
+	if props.Enabled == nil {
+		t.Fatalf("expected Enabled to be not nil")
+	}
+	// if props.ExpiresOn == nil {
```
chlowell, line 224: Can you delete the commented code?

sdk/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,18 +1,12 @@
 # Release History
 
-## 0.4.1 (Unreleased)
-
-### Features Added
+## 0.5.0 (2022-04-05)
```
heaths, line 3: Shouldn't requiring 1.18 be a breaking change?

heaths (commented):
APIs LGTM sans the few changes we talked about in APIView offline. I'll wait to see those to sign off.

sdk/keyvault/azkeys/models.go
seankane-msft: No, it has to be cast to a string first

sdk/keyvault/azkeys/CHANGELOG.md
chlowell: Our convention for v0.x.y is to increment x for breaking changes, y for everything else.

sdk/keyvault/azkeys/client.go
```diff
@@ -615,60 +667,67 @@ func (s *DeleteKeyPoller) pollUntilDone(ctx context.Context, t time.Duration) (D
...
-// DeleteKeyPollerResponse contains the response from the Client.BeginDeleteKey method
-type DeleteKeyPollerResponse struct {
-	// PollUntilDone will poll the service endpoint until a terminal state is reached or an error occurs
-	PollUntilDone func(context.Context, time.Duration) (DeleteKeyResponse, error)
-
-	// Poller contains an initialized WidgetPoller
-	Poller *DeleteKeyPoller
+// ResumeToken returns a token for resuming polling at a later time
+func (s *DeleteKeyPoller) ResumeToken() (string, error) {
+	return string(s.resumeToken), nil
```
chlowell, line 685:
```suggestion
	return s.resumeToken, nil
```

sdk/keyvault/azkeys/client.go
```diff
@@ -615,60 +667,67 @@ func (s *DeleteKeyPoller) pollUntilDone(ctx context.Context, t time.Duration) (D
...
 // PollUntilDone function can be used to poll until completion. Pass nil to use the default options.
-func (c *Client) BeginDeleteKey(ctx context.Context, keyName string, options *BeginDeleteKeyOptions) (DeleteKeyPollerResponse, error) {
+func (c *Client) BeginDeleteKey(ctx context.Context, name string, options *BeginDeleteKeyOptions) (*DeleteKeyPoller, error) {
 	if options == nil {
 		options = &BeginDeleteKeyOptions{}
 	}
-	resp, err := c.kvClient.DeleteKey(ctx, c.vaultUrl, keyName, options.toGenerated())
-	if err != nil {
-		return DeleteKeyPollerResponse{}, err
+	var resumeToken string
```
chlowell, line 695: Same feedback as on secrets, seems like the poller ought to carry the token in both cases

sdk/keyvault/azkeys/client.go
```diff
@@ -727,38 +786,48 @@ type RecoverDeletedKeyPoller struct {
...
 }
 
 // FinalResponse returns the final response after the operations has finished
 func (p *RecoverDeletedKeyPoller) FinalResponse(ctx context.Context) (RecoverDeletedKeyResponse, error) {
 	return recoverDeletedKeyResponseFromGenerated(p.recoverResponse), nil
 }
 
-// pollUntilDone is the method for the Response.PollUntilDone struct
-func (p *RecoverDeletedKeyPoller) pollUntilDone(ctx context.Context, t time.Duration) (RecoverDeletedKeyResponse, error) {
+// PollUntilDone is the method for the Response.PollUntilDone struct
```
chlowell, line 829: Suggest writing something to indicate why a user might want to call this method

sdk/keyvault/azkeys/client.go
```diff
@@ -788,107 +858,75 @@ type RecoverDeletedKeyResponse struct {
...
-// RecoverDeletedKeyPollerResponse contains the response of the Client.BeginRecoverDeletedKey operations
-type RecoverDeletedKeyPollerResponse struct {
-	// PollUntilDone will poll the service endpoint until a terminal state is reached or an error occurs
-	PollUntilDone func(context.Context, time.Duration) (RecoverDeletedKeyResponse, error)
-
-	// Poller contains an initialized RecoverDeletedKeyPoller
-	Poller *RecoverDeletedKeyPoller
+// ResumeToken returns a token for resuming polling at a later time
+func (r *RecoverDeletedKeyPoller) ResumeToken() (string, error) {
+	return string(r.resumeToken), nil
```
chlowell, line 875:
```suggestion
	return r.resumeToken, nil
```

sdk/keyvault/azkeys/client.go
```diff
@@ -1109,12 +1154,12 @@ func (c *Client) RestoreKeyBackup(ctx context.Context, keyBackup []byte, options
 // ImportKeyOptions contains the optional parameters for the Client.ImportKeyOptions parameter
 type ImportKeyOptions struct {
 	// Whether to import as a hardware key (HSM) or software key.
-	Hsm *bool `json:"Hsm,omitempty"`
+	HardwareProtected *bool `json:"Hsm,omitempty"`
```
chlowell, line 1157: Should this be
```suggestion
	HardwareProtected *bool `json:"hsm,omitempty"`
```

sdk/keyvault/azkeys/client.go
```diff
@@ -1109,12 +1154,12 @@ func (c *Client) RestoreKeyBackup(ctx context.Context, keyBackup []byte, options
...
 
 	// The key management attributes.
 	Properties *Properties `json:"attributes,omitempty"`
 
-	// Application specific metadata in the form of key-value pairs.
+	// Tag contains application specific metadata in the form of key-value pairs.
```
chlowell, line 1162:
```suggestion
	// Tags contains application specific metadata in the form of key-value pairs.
```

sdk/keyvault/azkeys/models.go
chlowell: New question: does this go over the wire as part of a request or response body?

sdk/keyvault/azkeys/internal/generated/constants.go
```diff
@@ -10,7 +10,7 @@ package generated
 
 const (
 	ModuleName    = "generated"
-	ModuleVersion = "v0.4.1"
+	ModuleVersion = "v0.5.0"
```
chlowell, line 13: "User-Agent": "azsdk-go-generated/v0.5.0 ..."

sdk/keyvault/azkeys/client.go
seankane-msft: No, in the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.3/keys.json#L1727) it is capitalized

seankane-msft:
@chlowell the VaultURL does not go over the wire, nor is it returned by the service, it's parsed from the Secret ID

sdk/keyvault/azkeys/models.go
chlowell: > @chlowell the VaultURL does not go over the wire, nor is it returned by the service, it's parsed from the Secret ID

What's it for? Why is it a pointer?

seankane-msft: It can be used for debugging. It is a consistency item with other languages as well. I made it a pointer to be consistent with all the other strings on this model. If the service call returns an error this will be nil and the user couldn't use the VaultURL to do anything like create a new Client (if they were to skip checking the error).

chlowell (approved):

sdk/keyvault/azkeys/crypto/crypto_client.go
```diff
@@ -120,39 +121,49 @@ func NewClient(keyURL string, credential azcore.TokenCredential, options *Client
...
-			IV:                          i.Iv,
-			ID:                          i.Kid,
-			Result:                      i.Result,
-		},
+		AuthData:   i.AdditionalAuthenticatedData,
+		AuthTag:    i.AuthenticationTag,
+		IV:         i.Iv,
+		KeyID:      i.Kid,
+		Ciphertext: i.Result,
+		Algorithm:  to.Ptr(alg),
```
richardpark-msft, line 166: Should this also just be 'Alg'? I like it the way it is, but wasn't sure if it was supposed to match it.

sdk/keyvault/azkeys/crypto/crypto_client.go
```diff
@@ -173,52 +184,55 @@ func (c *Client) Encrypt(ctx context.Context, alg EncryptionAlgorithm, value []b
...
 	// Additional data to authenticate but not encrypt/decrypt when using authenticated crypto algorithms.
-	AAD []byte `json:"aad,omitempty"`
+	AuthData []byte `json:"aad,omitempty"`
 
 	// Initialization vector for symmetric algorithms.
 	IV []byte `json:"iv,omitempty"`
 
 	// The tag to authenticate when performing decryption with an authenticated algorithm.
-	Tag []byte `json:"tag,omitempty"`
+	AuthTag []byte `json:"tag,omitempty"`
```
richardpark-msft, line 206: If AuthTag and AuthData are related you could place them together in the struct.. I don't know if we alpha-sort but that would have done it too.

sdk/keyvault/azkeys/client.go
```diff
@@ -66,22 +67,30 @@ func NewClient(vaultUrl string, credential azcore.TokenCredential, options *Clie
...
+	return c.vaultUrl
+}
+
 // CreateKeyOptions contains the optional parameters for the KeyVaultClient.CreateKey method.
 type CreateKeyOptions struct {
 	// Elliptic curve name. For valid values, see PossibleCurveNameValues.
 	Curve *CurveName `json:"crv,omitempty"`
 
-	// The properties of a key managed by the key vault service.
+	// The attributes of a key managed by the key vault service.
```
richardpark-msft, line 80: Silly question - why isn't it "properties"? The field name is Properties?

sdk/keyvault/azkeys/client.go
```diff
@@ -66,22 +67,30 @@ func NewClient(vaultUrl string, credential azcore.TokenCredential, options *Clie
 	}, nil
 }
 
+// VaultURL returns a string of the vault URL
+func (c *Client) VaultURL() string {
+	return c.vaultUrl
```
richardpark-msft, line 72: NIT: casing, Field name should probably be c.vaultURL as well

sdk/keyvault/azkeys/client.go
```diff
@@ -153,19 +164,36 @@ type CreateECKeyOptions struct {
...
+
+	// The policy rules under which the key can be exported.
+	ReleasePolicy *ReleasePolicy `json:"release_policy,omitempty"`
 }
 
 // convert CreateECKeyOptions to generated.KeyCreateParameters
 func (c *CreateECKeyOptions) toKeyCreateParameters(keyType KeyType) generated.KeyCreateParameters {
+	var keyOps []*generated.JSONWebKeyOperation
+	if c.Operations != nil {
+		keyOps = make([]*generated.JSONWebKeyOperation, len(c.Operations))
```
richardpark-msft, line 185: For future you can just loop and append() - you don't even need to initialize 'keyOps' since append works with a nil slice ptr just fine.

richardpark-msft: (you might have just been trying to be more efficient, however)

sdk/keyvault/azkeys/example_test.go
```diff
@@ -102,18 +102,19 @@ func ExampleClient_UpdateKeyProperties() {
...
-			RecoveryLevel: azkeys.DeletionRecoveryLevelCustomizedRecoverablePurgeable.ToPtr(),
-		},
-	})
+	resp, err := client.GetKey(context.TODO(), "key-to-update", nil)
+	if err != nil {
+		panic(err)
+	}
+
+	resp.Key.Properties.Tags = map[string]string{"Tag1": "val1"}
+	resp.Key.Properties.Enabled = to.Ptr(true)
```
richardpark-msft, line 111: Is the default Disabled? Or is this likely already true?

sdk/keyvault/azkeys/client.go
```diff
@@ -123,12 +133,13 @@ type CreateKeyResponse struct {
 
 // creates CreateKeyResponse from generated.KeyVaultClient.CreateKeyResponse
 func createKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyResponse) CreateKeyResponse {
+	vaultURL, name, version := shared.ParseID(g.Key.Kid)
 	return CreateKeyResponse{
 		Key: Key{
-			Properties: keyPropertiesFromGenerated(g.Attributes),
+			Properties: keyPropertiesFromGenerated(g.Attributes, g.Key.Kid, name, version, g.Managed, vaultURL, g.Tags),
```
richardpark-msft, line 139: Someday you'll probably want to pass in a struct or something with named fields, to avoid accidental tranposition with params.

sdk/keyvault/azkeys/models.go
```diff
@@ -351,11 +391,32 @@ type RotationPolicy struct {
 	ID *string `json:"id,omitempty" azure:"ro"`
 }
 
+func (u RotationPolicy) toGenerated() generated.KeyRotationPolicy {
+	var attribs *generated.KeyRotationPolicyAttributes
+	if u.Attributes != nil {
+		attribs = u.Attributes.toGenerated()
+	}
+	var la []*generated.LifetimeActions
+	if la != nil {
```
richardpark-msft, line 400: Won't this always be false?

richardpark-msft (commented):
Some comments. There was a spot that seems like it might be a logic error.

sdk/keyvault/azkeys/client.go
seankane-msft: This was an efficiency thing. I didn't want to create a slice that would be nil, and if it's left as nil the value won't be serialized when we write the request body

sdk/keyvault/azkeys/models.go
seankane-msft: Yes :)

sdk/keyvault/azkeys/crypto/crypto_client.go
seankane-msft: I'm going to leave it as Algorithm, it's consistent with other languages.

sdk/keyvault/azkeys/client.go
seankane-msft: Agreed. I could move some of the parsing into the method at a later date

sdk/keyvault/azkeys/example_test.go
seankane-msft: It's likely already true, but a sample that we provide in other languages

sdk/keyvault/azkeys/client.go
```diff
@@ -23,7 +23,7 @@ import (
 // Client is the struct for interacting with a KeyVault Keys instance
 type Client struct {
 	kvClient *generated.KeyVaultClient
-	vaultUrl string
+	vaultURL string
```
heaths, line 26: Weren't you also going to expose the `VaultURL` for debugging? This would likely be better as a `func` to avoid mutability, which would be disastrous once the credential token is initially fetched.

Issue #6846, opened 2020-01-13
go get on master branch fails.


rebecca-makar:
### Bug Report

- go get on entire directory. Following README.md
- SDK version - master

- What happened?

```
+ go get -v -u github.com/Azure/azure-sdk-for-go/...
github.com/Azure/azure-sdk-for-go (download)
github.com/Azure/go-autorest (download)
github.com/dgrijalva/jwt-go (download)
github.com/satori/go.uuid (download)
github.com/shopspring/decimal (download)
Fetching https://golang.org/x/crypto/pkcs12?go-get=1
Parsing meta tags from https://golang.org/x/crypto/pkcs12?go-get=1 (status code 200)
get "golang.org/x/crypto/pkcs12": found meta tag get.metaImport{Prefix:"golang.org/x/crypto", VCS:"git", RepoRoot:"https://go.googlesource.com/crypto"} at https://golang.org/x/crypto/pkcs12?go-get=1
get "golang.org/x/crypto/pkcs12": verifying non-authoritative meta tag
Fetching https://golang.org/x/crypto?go-get=1
Parsing meta tags from https://golang.org/x/crypto?go-get=1 (status code 200)
golang.org/x/crypto (download)
Fetching https://golang.org/x/crypto/pkcs12/internal/rc2?go-get=1
Parsing meta tags from https://golang.org/x/crypto/pkcs12/internal/rc2?go-get=1 (status code 200)
get "golang.org/x/crypto/pkcs12/internal/rc2": found meta tag get.metaImport{Prefix:"golang.org/x/crypto", VCS:"git", RepoRoot:"https://go.googlesource.com/crypto"} at https://golang.org/x/crypto/pkcs12/internal/rc2?go-get=1
get "golang.org/x/crypto/pkcs12/internal/rc2": verifying non-authoritative meta tag
github.com/globalsign/mgo (download)
github.com/dimchansky/utfbom (download)
github.com/mitchellh/go-homedir (download)
github.com/spf13/cobra (download)
github.com/spf13/pflag (download)
github.com/Masterminds/semver (download)
github.com/pkg/errors (download)
Fetching https://golang.org/x/tools/imports?go-get=1
Parsing meta tags from https://golang.org/x/tools/imports?go-get=1 (status code 200)
[...]

PR #17402, opened 2022-03-29
[KeyVault] Updating internal to go1.18

seankane-msft:
This will need a release after the azcore release

### DIFF
--- a/sdk/keyvault/internal/challenge_policy.go
+++ b/sdk/keyvault/internal/challenge_policy.go
@@ -1,5 +1,5 @@
-//go:build go1.16
-// +build go1.16
+//go:build go1.18
+// +build go1.18
 
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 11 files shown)


chlowell (approved):

RickWinter (approved):

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -1,5 +1,5 @@
-//go:build go1.16
-// +build go1.16
+//go:build go1.18
+// +build go1.18
```
chlowell, line 2: The `+build` tags are only needed for Go < 1.17, so they're redundant now.

jhendrixMSFT: I'll update the code generator with this change shortly.

jhendrixMSFT: I will point out though that removing them will break building with older versions of Go.

seankane-msft: This uses the `to.Ptr` method, it won't build with older versions of Go anyways.

jhendrixMSFT: To clarify, I'm talking about track 1.  We had this problem in the past, see https://github.com/Azure/azure-sdk-for-go/issues/6846.  We might need to keep these until track 1 is out of `main` (needs investigation).

seankane-msft: I'll keep these for now

PR #17411, opened 2022-03-30
Remove pollerID parameter from poller constructors

jhendrixMSFT:
The generic type parameter has removed the need for an explicit poller
ID parameter during construction.
Create the poller ID name from the generic type parameter.

### DIFF
--- a/sdk/azcore/arm/runtime/poller.go
+++ b/sdk/azcore/arm/runtime/poller.go
@@ -23,8 +23,7 @@ import (
 )
 
 // NewPoller creates a Poller based on the provided initial response.
-// pollerID - a unique identifier for an LRO.  it's usually the client.Method string.
-func NewPoller[T any](pollerID string, finalState string, resp *http.Response, pl pipeline.Pipeline, rt *T) (*Poller[T], error) {
+func NewPoller[T any](finalState string, resp *http.Response, pl pipeline.Pipeline, rt *T) (*Poller[T], error) {
 	defer resp.Body.Close()
 	// this is a back-stop in case the swagger is incorrect (i.e. missing one or more status codes for success).
 	// ideally the codegen should return an error if the initial response failed and not even create a poller.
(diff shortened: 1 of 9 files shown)


sdk/azcore/internal/pollers/util.go
```diff
@@ -53,6 +53,17 @@ func IsValidURL(s string) bool {
...
 
 const idSeparator = ";"
+const emptyTypeName = "-empty-"
+
+// PollerTypeName returns the type name to use when constructing the poller ID.
+// It handles the empty type struct{} with a special name instead of the empty string.
+func PollerTypeName[T any]() string {
+	n := shared.TypeOfT[T]().Name()
+	if n == "" {
+		n = emptyTypeName
```
jhendrixMSFT, line 63: We could also just disallow specifying an empty type (e.g. `struct{}`) instead, although this was useful for tests where the type is irrelevant.  If people think it's safer to just disallow instead of handling empty types specially I can make that change.

jhendrixMSFT: I went ahead and made this change as it's representative of real-world scenarios.  We can always relax it in a non-breaking way in the future if needed.

chlowell:
I'm puzzled by `FinalState`. Would I use it to configure a poller to consider its operation done upon receiving some status other than "Succeeded"? Do we have any tests covering it and `Response`?

jhendrixMSFT:
It's something that was added to the [`x-ms-long-running-operation`](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-long-running-operation) extension to handle some ARM services that don't quite follow the spec.

sdk/azcore/arm/runtime/poller.go
```diff
@@ -22,26 +22,41 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
+// NewPollerOptions contains the optional parameters for NewPoller.
+type NewPollerOptions[T any] struct {
+	// FinalState contains the final-state-via value for the LRO.
+	FinalState string
```
jhendrixMSFT, line 28: @chlowell rename to `FinalStateVia`?

chlowell: Sure, that lines up better with the schema. I take it this is too obscure to warrant an enum?

jhendrixMSFT: Hmm we could add an enum.  Per the docs, there is only a predefined set of values.

jhendrixMSFT: Also, it looks like this might not be just for ARM.  I'm following up on that.

jhendrixMSFT: Confirmed as I see it in some CS specs.  Will update.

jhendrixMSFT: Actually I'll do that separately as the data-plane pollers need to be updated to support it.

jhendrixMSFT: Opened https://github.com/Azure/azure-sdk-for-go/issues/17415 to track it.

jhendrixMSFT:
To answer your other question about `Response`, it's currently only used by paged pollers.  We do have a test covering this in the `autorest.go` test suite.

jhendrixMSFT:
To answer your other question about `Response`, it's currently only used by paged pollers.  We do have a test covering this in the `autorest.go` test suite.

chlowell (approved):

PR #17422, opened 2022-04-01
Fixing an issue where we wouldn't return cancellation errors even whe…

### LINKED ISSUE #17412: [azservicebus] ReceiveMessages() does not always return the proper error when a context is cancelled
ReceiveMessages() does a bit work to ensure that we don't return both an error and messages since Go users are accustomed to only checking one or the other (but not both - typically `err` being non-nil will take precedence).

In the code it looks like there are two spots where a cancellation error can be "erased" (harmlessly) and one where a cancellation error can be returned unnecessarily.

The erasure can happen when the following happens:
1. User calls ReceiveMessages()
2. context is cancelled after .Receive() is called
3. drain occurs and does NOT fail or cancel
4. _no_ messages were received either by Receive() or by Drain()
5. We return `nil` error, "erasing" the cancellation error.

The unnecessary cancellation return happens if:
1. User calls ReceiveMessages()
2. context is cancelled
3. drain occurs and times out (deadline exceeded on our internal context)
4. We return the cancellation error if the there were no messages

richardpark-msft:
Fixing an issue where we wouldn't return cancellation errors even when it made sense.

This would manifest as a user getting an empty array on a ReceiveMessages() call, even if the actual problem is that the cancellation timeout is being set too low.

Small cleanups on some comments as well.

Fixes #17412

### DIFF
--- a/sdk/messaging/azservicebus/receiver_test.go
+++ b/sdk/messaging/azservicebus/receiver_test.go
@@ -19,6 +19,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestReceiverCancel(t *testing.T) {
+	serviceBusClient, cleanup, queueName := setupLiveTest(t, nil)
+	defer cleanup()
+
+	receiver, err := serviceBusClient.NewReceiverForQueue(queueName, nil)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	messages, err := receiver.ReceiveMessages(ctx, 5, nil)
+	require.NoError(t, err)
+	require.Empty(t, messages)
+}
+
 func TestReceiverSendFiveReceiveFive(t *testing.T) {
 	serviceBusClient, cleanup, queueName := setupLiveTest(t, nil)
 	defer cleanup()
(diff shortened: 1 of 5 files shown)


sdk/messaging/azservicebus/receiver_test.go
```diff
@@ -19,6 +19,21 @@ import (
...
+	defer cleanup()
+
+	receiver, err := serviceBusClient.NewReceiverForQueue(queueName, nil)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	messages, err := receiver.ReceiveMessages(ctx, 5, nil)
+	require.NoError(t, err)
```
jhendrixMSFT, line 33: Maybe I'm missing something here but why is `err` expected to be `nil`?  Hasn't the context timed out?

richardpark-msft: Oops, sorry, need to fix that one. I was focusing on the unit tests, but yes, the cancel error should come back here.

richardpark-msft: (this was the previous undesirable behavior!)

richardpark-msft: (fixed, btw)

jhendrixMSFT (approved):

Issue #17415, opened 2022-03-31
Add support for final-state-via for data-plane pollers


jhendrixMSFT:
This isn't just for ARM.

As part of the work, change the `FinalStateVia` field to a `const` type with applicable values.

jhendrixMSFT:
Once the public surface area changes are in, we need to update autorest with data-plane scenarios to ensure we get the algorithm correct.

jhendrixMSFT:
Fixed in upcoming `azcore@v0.24.0`

PR #17426, opened 2022-04-01
Add FinalStateVia to data-plane poller options

jhendrixMSFT:
Converted FinalStateVia to a const as its values are constrained.

NOTE: this makes the necessary changes to public surface area.  However, we don't have coverage in autorest for data-plane LROs at present, so the implementation is incomplete.  This doesn't impact the current behavior though.

Partially fixes https://github.com/Azure/azure-sdk-for-go/issues/17415

### DIFF
--- a/sdk/azcore/internal/pollers/loc/loc.go
+++ b/sdk/azcore/internal/pollers/loc/loc.go
@@ -28,11 +28,12 @@ func Applicable(resp *http.Response) bool {
 type Poller struct {
 	Type     string `json:"type"`
 	PollURL  string `json:"pollURL"`
+	FinalGET string `json:"finalGET"`
 	CurState int    `json:"state"`
 }
 
 // New creates a new Poller from the provided initial response.
-func New(resp *http.Response, pollerID string) (*Poller, error) {
+func New(resp *http.Response, finalState pollers.FinalStateVia, pollerID string) (*Poller, error) {
 	log.Write(log.EventLRO, "Using Location poller.")
 	locURL := resp.Header.Get(shared.HeaderLocation)
 	if locURL == "" {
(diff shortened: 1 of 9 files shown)


sdk/azcore/internal/pollers/op/op.go
```diff
@@ -61,6 +61,7 @@ func New(resp *http.Response, pollerID string) (*Poller, error) {
 	// calculate the tentative final GET URL.
 	// can change if we receive a resourceLocation.
 	// it's ok for it to be empty in some cases.
+	// TODO: finalState
```
seankane-msft, line 64: Is there still a TODO item here?

sdk/azcore/internal/pollers/loc/loc.go
```diff
@@ -41,9 +42,12 @@ func New(resp *http.Response, pollerID string) (*Poller, error) {
 	if !pollers.IsValidURL(locURL) {
 		return nil, fmt.Errorf("invalid polling URL %s", locURL)
 	}
+	// TODO: calculate the final GET URL.  can be empty
```
seankane-msft, line 45: Is there still a TODO item here?

sdk/azcore/internal/pollers/op/op.go
jhendrixMSFT: Yeah.  We need to add test coverage for data-plane pollers in autorest to ensure we get the algorithm right (comment applies to both TODOs).

jhendrixMSFT: Tracked in https://github.com/Azure/azure-sdk-for-go/issues/17415

chlowell (approved):

PR #17438, opened 2022-04-04
Update azidentity to use new cloud config API

chlowell:
Removes `azidentity.AuthorityHost`; credentials now get host values from `azcore.ClientOptions.Cloud`.


sdk/azidentity/go.mod
```diff
@@ -1,16 +1,25 @@
 module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
-go 1.16
+go 1.18
 
 require (
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.21.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.22.1-0.20220401193339-cf00edc55f3c
```
chlowell, line 6: I'll update this after the azcore release (before merging this PR)

sdk/azidentity/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.14.0 (2022-04-05)
 
 ### Breaking Changes
+* This module now requires Go 1.18
```
richardpark-msft, line 6: Ooh, good reminder. I need to do the same.

sdk/azidentity/azidentity.go
```diff
@@ -30,23 +32,17 @@ const (
...
-	host := string(authorityHost)
+// setAuthorityHost initializes the authority host for credentials. Precedence is:
+// 1. cloud.Configuration.LoginEndpoint value set by user
+// 2. value of AZURE_AUTHORITY_HOST
+// 3. default: Azure Public Cloud
+func setAuthorityHost(cc cloud.Configuration) (string, error) {
+	host := cc.LoginEndpoint
 	if host == "" {
-		host = string(AzurePublicCloud)
+		if !reflect.ValueOf(cc).IsZero() {
```
richardpark-msft, line 42: The only other field to check here is if `Services` is nil, right? Do we need reflection for that?

sdk/azidentity/doc.go
```diff
@@ -8,13 +8,28 @@
...
 
 	cred, err := azidentity.NewDefaultAzureCredential(nil)
 	...
-	client := armresources.NewResourceGroupsClient("subscription ID", cred, nil)
+	client := armsubscription.NewClient(cred, nil)
+)
 
-Different credential types implement different authentication flows. Each credential's
-documentation describes how it authenticates.
+Different credential types implement different authentication flows and so have different
```
richardpark-msft, line 18: This is nicely written :)

sdk/azidentity/azidentity.go
chlowell: Good point! Checking whether `Services` is empty would suffice.

richardpark-msft (approved):
One minor comment about the reflect() call, otherwise seems fine.

PR #17464, opened 2022-04-06
Wait for test proxy startup

### LINKED ISSUE #17461: Pipeline flakiness due to test proxy initialization race
Test runs occasionally fail because the test proxy isn't available before the module under test sends its first request ([for example](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1483554&view=logs&j=7c2e6c09-e0bf-5c90-9850-b294a30ca1c6&t=e00182e0-c89a-5584-1fee-0622017fb833&s=6884a131-87da-5381-61f3-d7acc3b91d76)). I guess this is a race because the pipeline continues to the next step immediately after starting the proxy, which isn't immediately ready for requests.

benbp:
Fixes #17461

After adding this and testing it, it seems like it's really just a couple seconds that's needed to ensure the proxy is ready to receive requests. We could consider adding `Start-Sleep 2` as an alternative to all this extra checking.

CC @chlowell


benbp:
Didn't realize this was in eng/common. Moving PR to tools.

PR #17474, opened 2022-04-07
dont block release on perf go.mod files

### LINKED ISSUE #17441: [Eng] Checks should not run on `testdata` directories
@chlowell was blocked on releasing azcore because the testdata/perf directory referenced the local version of internal/perf. The testdata directory is ignored by the go tooling, so we can ignore it as well.

seankane-msft:
@benbp based on my searches this was the most elegant way to tackle this situation. I don't want to block a release if a perf test's go.mod file uses a replace directive, this checks if `testdata` is in the path for a go.mod file and skips it if this is correct.

Closes #17441

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -37,7 +37,7 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(\w*/)?([^/]+/)?(?<modName>[^/]+$)))")
   {
     $modPath = $matches["modPath"]
     $modName = $matches["modName"] # We may need to start readong this from the go.mod file if the path and mod config start to differ
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -37,7 +37,7 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(\w*/)?([^/]+/)?(?<modName>[^/]+$)))")
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
   {
     $modPath = $matches["modPath"]
     $modName = $matches["modName"] # We may need to start readong this from the go.mod file if the path and mod config start to differ
--- a/eng/scripts/validate_go_mod.ps1
+++ b/eng/scripts/validate_go_mod.ps1
@@ -15,10 +15,8 @@ if ($goModFiles.Length -eq 0) {
 
 $hasError = $false
 foreach ($goMod in $goModFiles) {
-    Write-Host $goMod.Directory
-    $packageProperties = Get-GoModuleProperties $goMod.Directory
-    Write-Host "Props $packageProperties"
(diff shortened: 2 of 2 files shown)


eng/scripts/validate_go_mod.ps1
```diff
@@ -7,17 +7,19 @@ Push-Location sdk/$serviceDir
...
 }
 
 $hasError = $false
 foreach ($goMod in $goModFiles) {
-    $name = $goMod.FullName
-    $patternMatches = Get-Content $name | Select-String -Pattern "replace "
-    if ($patternMatches.Length -ne 0) {
-        Write-Host "Found a replace directive in go.mod file at $name"
-        $hasError = $true
+    if (-Not ($goMod.FullName -Like "*testdata*")) {
```
weshaggard, line 16: I'd suggest using the common https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L37 they already return null for these non-standard go.mod files.

eng/scripts/Language-Settings.ps1
```diff
@@ -37,7 +37,7 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(\w*/)?([^/]+/)?(?<modName>[^/]+$)))")
```
weshaggard, line 40: Why this change? We intentionally only had resourcemanager because they were the only approved projects that were deeper in paths.

eng/scripts/validate_go_mod.ps1
```diff
@@ -2,22 +2,29 @@ Param(
     [string] $serviceDir
 )
 
+. (Join-Path $PSScriptRoot . Language-Settings.ps1)
```
weshaggard, line 5:
```suggestion
. (Join-Path $PSScriptRoot .. common scripts common.ps1)
```

eng/scripts/validate_go_mod.ps1
```diff
@@ -2,22 +2,29 @@ Param(
...
 
 $hasError = $false
 foreach ($goMod in $goModFiles) {
-    $name = $goMod.FullName
-    $patternMatches = Get-Content $name | Select-String -Pattern "replace "
-    if ($patternMatches.Length -ne 0) {
-        Write-Host "Found a replace directive in go.mod file at $name"
-        $hasError = $true
+    $mod = Get-GoModuleProperties ($goMod -replace ".*\/azure-sdk-for-go\/")
+    if ($mod -ne $null) {
```
weshaggard, line 19:
```suggestion
    if ($mod) {
```

eng/scripts/Language-Settings.ps1
seankane-msft: @benbp helped out here. This regex does not match for the multi directory (ie sdk/data/aztables). It will match for sdk/azcore and sdk/azidentity, but not further. Is there a need to keep the resourcemanager portion?

weshaggard: This should be matching sdk/data/aztables just fine, if it weren't then that pipeline would be completely broken. Is there some evidence that it isn't working for them?

seankane-msft: @weshaggard https://regex101.com/r/yE8yp3/1

benbp: @weshaggard I agree that it's worth investigating why the pipeline is working because this regex does not appear to work based on our testing in that regex101 link.

weshaggard: Generally the path passed here is a directory not the full path to go.mod, see https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L108 as an example. If we want to support the full path as well we could add `(/go.mod)?` at the end of the regex.

seankane-msft: I passed in the directory and this worked as expected. Thanks for clarifying @weshaggard

weshaggard (approved):

PR #17492, opened 2022-04-08
[KeyVault] Adding performance tests

### LINKED ISSUE #17362: [Perf] Add performance tests for KeyVault
- [ ] Certificates: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Certificates/perf/Scenarios
  - [ ] DownloadCertificate
  - [ ] GetCertificate
- [ ] Keys: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Keys/perf/Scenarios
  - [ ] GetKey
  - [ ] crypto tests
- [ ] Secrets: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Secrets/perf/Scenarios
  - [ ] GetSecret
  - [ ] ListSecrets

seankane-msft:
closes #17362


sdk/keyvault/azkeys/testdata/perf/decrypt.go
```diff
@@ -0,0 +1,131 @@
...
+	client, err := azkeys.NewClient(vaultURL, cred, &azkeys.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
+			Transport: options.Transporter,
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := client.CreateRSAKey(ctx, d.keyName, nil)
```
heaths, line 60: To align with .NET, at least, I'd specify a key size of 2048, unless that's already your default. I presume, though, you let the service default which can change and change the perf measurements. I'd hardcode it to 2048.

sdk/keyvault/azkeys/testdata/perf/decrypt.go
```diff
@@ -0,0 +1,131 @@
...
+	_, err = rand.Read(b)
+	if err != nil {
+		return nil, err
+	}
+
+	result, err := d.cryptoClient.Encrypt(ctx, d.encrypAlg, b, nil)
+	if err != nil {
+		return nil, err
+	}
+	d.cipherText = result.Ciphertext
```
heaths, line 87: If possible, I'd do this once per run - probably in `NewPerfTest` - to avoid potential caching of results upstream: https://github.com/Azure/azure-sdk-for-net/blob/1b56a4b5c85e427d912dcc2ffa5f5a627bf3e5b7/sdk/keyvault/Azure.Security.KeyVault.Keys/perf/Scenarios/Decrypt.cs#L25-L36

sdk/keyvault/azkeys/testdata/perf/get_key.go
```diff
@@ -0,0 +1,95 @@
...
+		panic(err)
+	}
+
+	client, err := azkeys.NewClient(vaultURL, cred, &azkeys.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
+			Transport: options.Transporter,
+		},
+	})
+
+	_, err = client.CreateRSAKey(ctx, d.keyName, nil)
```
heaths, line 51: I'd hardcode to 2048 bit key size. Same across other scenario tests.

sdk/keyvault/azkeys/testdata/perf/unwrap.go
```diff
@@ -0,0 +1,128 @@
...
+	b := make([]byte, 32)
+	_, err = rand.Read(b)
+	if err != nil {
+		return nil, err
+	}
+
+	result, err := cryptoClient.WrapKey(ctx, d.alg, b, nil)
+	if err != nil {
+		return nil, err
+	}
```
heaths, line 83: As with decrypt, I'd move this to per-run to avoid caching of results: https://github.com/Azure/azure-sdk-for-net/blob/1b56a4b5c85e427d912dcc2ffa5f5a627bf3e5b7/sdk/keyvault/Azure.Security.KeyVault.Keys/perf/Scenarios/Unwrap.cs#L26-L36

heaths (approved):
A few suggestions, but otherwise LGTM.

sdk/keyvault/azkeys/testdata/perf/decrypt.go
seankane-msft: This method is only run once the currently.

sdk/keyvault/azkeys/testdata/perf/unwrap.go
seankane-msft: This runs once per-run right now

sdk/keyvault/azkeys/testdata/perf/decrypt.go
heaths: Reading the method names and looking at how they are used, it seems `newDecryptTest` is called once per run but `NewPerfTest` is called once per perf function run, correct? That corresponds to how .NET has both per-run and per-scenario setups and tear-downs. If so, the idea here is to create the client once per run, but to use different plaintext, keys to wrap, etc. per test invoccation. If you do those once, you run the risk of caching either by the service, or by the SDK (if we ever started caching) and will skew results.

seankane-msft: `newDecryptTest` is called once per process, this is related to the `GlobalSetup` in other languages. `decryptTest.NewPerfTest` is called once per parallel thread, similar to `Setup` in other languages.

heaths: Yes, that's what I understood as well. Creating keys to wrap or encrypting plaintext should be done in `Setup` so that the actual perf test scenario gets a different value each time. Only create the client and create the encryption key in `GlobalSetup`.

mikeharder: `GlobalSetup` is called once per test run, while `Setup` is called once per parallel instance.  So if you are running a test with `parallel = 4`, `GlobalSetup` will be called exactly once, and `Setup` will be called exactly 4 times.

There is no `Setup` method called once per operation (per call to `Run`).  If you want to do something on each operation, it needs to be inside `Run`, and its impact will be included in the throughput results.

heaths: By "operation" I meant "run". Can you confirm this is correct:

* GlobalSetup
  * 1..n
    * Setup
    * Run
    * Cleanup
* GlobalCleanup

mikeharder: I don't think that is correct.  The core of the perf test algorithm is here: https://github.com/Azure/azure-sdk-for-net/blob/2a35ed235fafe7b7ad7c8920e34118c3d06f0492/common/Perf/Azure.Test.Perf/PerfProgram.cs#L102-L200.  In pseudo-code:

* Create `Parallel` instances of the test class
* Call `GlobalSetup()` on instance 0
* Call `Setup()` on all instances (in parallel)
* Call `Run()` in a loop on all instances (in parallel)
* Call `Cleanup()` on all instances (in parallel)
* Call `GlobalCleanup()` on instance 0

heaths: So then:

* GlobalSetup
  * 1..n
    * Setup
      * 1..m
        * Run
    * Cleanup
* GlobalCleanup

Which is still fine. The goal was to not use the same decryption key, plaintext, etc. in every iteration to avoid skewed measurement because of potential caching by the framework, SDK, or even service. We also don't want to measure that time in the `Run` method (or equivalent) because it's significant and not what we're intending to measure.

mikeharder: That flowchart looks correct.  However I still don't follow your argument about doing something in GlobalSetup vs Setup.  The main reason to do something in Setup instead of GlobalSetup is if each parallel instance needs to do something different.

For example, when downloading a blob there is no contention if parallel instances download the same blob, so the test uploads a single blob in GlobalSetup that all parallel instances can download.

However, if multiple parallel instances try to upload the same blob, they will be throttled by contention.  So the upload test generates a unique blob name per parallel instance in Setup instead of GlobalSetup.

Looking at https://github.com/Azure/azure-sdk-for-net/blob/1b56a4b5c85e427d912dcc2ffa5f5a627bf3e5b7/sdk/keyvault/Azure.Security.KeyVault.Keys/perf/Scenarios/Decrypt.cs#L25-L36, I would expect to see the same results whether this code lives in Setup or GlobalSetup.  There might be a slight difference between each parallel instance using the same or different ciphertext (e.g. more efficient use of CPU instruction cache), but either way each call to Run within a parallel instance is using the same ciphertext, so if there is any benefit to caching, we are already seeing it.

heaths: If we used the same symmetric key there is a non-zero change something - like the results - could be cached. The idea was to at least use different keys across runs or at least parallel instances (now that you've cleared that up) to avoid skewing the results. Imagine any particular service where only the first run actually hits the service and gets back a non-204 response, but then gets subsequent 204 messages with no body and no work to do in the client. Measurements would be greatly skewed toward the 204 response timings.

heaths: ...though, perhaps better we take this offline or start a new discussion (perhaps in Azure/azure-sdk-tools?) rather than keep using a resolved thread on a merged PR. 😉

PR #17493, opened 2022-04-08
AppConfig: Fix HMAC auth for Germany West Central

### LINKED ISSUE #17424: [AppConfig]: Error while reading Settings from Azure App Configuration
### Bug Report

**Environment**

```
go 1.17

require (
	github.com/Azure/azure-sdk-for-go/sdk/appconfig/azappconfig v0.1.0
	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.22.0
	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v0.3.0
)
```

**What happened?**

Reading settings from Azure App Configuration (AppConfig) does not work. When using a connection string for authentication,  requests fail with an HTTP `401` and `HMAC-SHA256 error="invalid_token", error_description="Invalid Credential"` provided as in the response header.

When authenticating with `azidentity` same underlying request fail with an HTTP `403`. I had `azidentity` configured to re-use existing authentication token from Azure CLI (`az`). 

Reading the same setting works perfectly fine via `az appconfig kv show`

**What did you expect or want to happen?**

I expect to retrieve the setting from AppConfig 😄 

**How can we reproduce it?**

Here the fragment of the code used for authenticating with a connection string

```go
func test(ctx context.Context, key string, label string, connectionString string) {
	client, err := azappconfig.NewClientFromConnectionString(connectionString, nil)
	// removed err handling because it does not matter for this conversation
	res, err := client.GetSetting(ctx, azappconfig.Setting{
		Key:   to.StringPtr(key),
		Label: to.StringPtr(label),
	}, nil)
	if err != nil {
		fmt.Printf("Error while reading setting: %s", err)
	}
}
```

**Anything we should know about your environment.**

local: I tested it on latest macOS (Intel)
cloud: see the following `az` script to spin up the required services:

```bash
appCfgName=foobar
rgName=rg-appcfg-sample
location=germanywestcentral

az group create -n $rgName -l $location
az appconfig create -n $appCfgName -g $rgName -l $location --sku Standard
```

antkmsft:
Closes #17424

### DIFF
--- a/sdk/appconfig/azappconfig/policy_hmac_auth.go
+++ b/sdk/appconfig/azappconfig/policy_hmac_auth.go
@@ -69,8 +69,8 @@ func (policy *hmacAuthenticationPolicy) Do(request *policy.Request) (*http.Respo
 	}
 
 	req.Header.Set("x-ms-content-sha256", contentHash)
-	req.Header.Set("x-ms-date", timestamp)
-	req.Header.Set("Authorization", "HMAC-SHA256 Credential="+id+", SignedHeaders=x-ms-date;host;x-ms-content-sha256, Signature="+signature)
+	req.Header.Set("Date", timestamp)
+	req.Header.Set("Authorization", "HMAC-SHA256 Credential="+id+", SignedHeaders=date;host;x-ms-content-sha256, Signature="+signature)
 
 	return request.Next()
 }
--- a/sdk/appconfig/azappconfig/CHANGELOG.md
+++ b/sdk/appconfig/azappconfig/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed authentication in Germany West Central using connection string (#17424).
 
 ### Other Changes
 


sdk/appconfig/azappconfig/policy_hmac_auth.go
```diff
@@ -69,8 +69,8 @@ func (policy *hmacAuthenticationPolicy) Do(request *policy.Request) (*http.Respo
 	}
 
 	req.Header.Set("x-ms-content-sha256", contentHash)
-	req.Header.Set("x-ms-date", timestamp)
-	req.Header.Set("Authorization", "HMAC-SHA256 Credential="+id+", SignedHeaders=x-ms-date;host;x-ms-content-sha256, Signature="+signature)
+	req.Header.Set("Date", timestamp)
+	req.Header.Set("Authorization", "HMAC-SHA256 Credential="+id+", SignedHeaders=date;host;x-ms-content-sha256, Signature="+signature)
```
seankane-msft, line 73: Does `Date` need to be capitalized in the header? Otherwise, LGTM

seankane-msft (approved):

sdk/appconfig/azappconfig/policy_hmac_auth.go
antkmsft: Good point, but I'd leave it capitalized for 3 reasons:
1. This is what .NET SDK does
2. This is what is confirmed to work
3. In RFCs, such as RFC 2616, "Date" and "Authorization" are capitalized.

seankane-msft: What about in the Authorization header ie `SignedHeaders=date;...` vs `SignedHeaders=Date`

antkmsft: I would not change it, this is how .NET and also Java put it, and it works.

antkmsft: BTW, do you know if Go SDK does lowercase headers? I know we do it in C++ SDK.

PR #17524, opened 2022-04-11
[Tables] Adding error codes

### LINKED ISSUE #17449: [Tables] Add table error codes
https://apiview.dev/Assemblies/Review/14024004d53648b6bc325087299aead9#Azure.Data.Tables.Models.TableErrorCode

seankane-msft:
On service errors, parse the response body and change the ErrorCode to the returned error code when applicable

closes #17449


sdk/data/aztables/error_codes.go
```diff
@@ -0,0 +1,107 @@
...
+func parseErrorCode(svcErr error) error {
+	var httpErr *azcore.ResponseError
+	if errors.As(svcErr, &httpErr) {
+		body, err := ioutil.ReadAll(httpErr.RawResponse.Body)
+		if err != nil {
+			return svcErr
+		}
+
+		var oerr oDataError
+		err = json.Unmarshal(body, &oerr)
```
jhendrixMSFT, line 94: I'm surprised this is JSON.  When I was testing the changes for `SetAccessPolicy` the error response was XML.
```xml
<?xml version="1.0" encoding="utf-8"?>
<m:error xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata">
  <m:code>InvalidXmlDocument</m:code>
  <m:message xml:lang="en-US">
    XML specified is not syntactically valid.
    RequestId:84770ad1-9002-00ae-1d5d-4fc246000000
    Time:2022-04-13T17:39:33.4024706Z</m:message>
</m:error>
```
Do you know when the error response is JSON vs XML?

JeffreyRichter: I assume Sean tested this, right?

JeffreyRichter (approved):

sdk/data/aztables/error_codes.go
```diff
@@ -0,0 +1,107 @@
...
+
+type oDataMessage struct {
+	Code TableErrorCode `json:"code"`
+}
+
+// parseErrorCode creates a azcore.ResponseError with the Code as one of the TableErrorCodes
+func parseErrorCode(svcErr error) error {
+	var httpErr *azcore.ResponseError
+	if errors.As(svcErr, &httpErr) {
+		body, err := ioutil.ReadAll(httpErr.RawResponse.Body)
```
jhendrixMSFT, line 88: It's a bit unfortunate that we have to parse this again.  We have some handling for error code extraction in `azcore` [already](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/shared/response_error.go#L47).  Might be better to fix that up to handle the `odata.error` case.

sdk/data/aztables/error_codes.go
jhendrixMSFT: Also, I submitted a fix to handle the XML case when namespaces are involved.  Once that fix is in, the XML case should just work.

sdk/data/aztables/error_codes.go
seankane-msft: I added this PR to add the case: https://github.com/Azure/azure-sdk-for-go/pull/17579

If that PR looks good, this will be much more simplified

sdk/data/aztables/error_codes.go
```diff
@@ -0,0 +1,93 @@
...
+		UpdateConditionNotSatisfied,
+		XMethodIncorrectCount,
+		XMethodIncorrectValue,
+		XMethodNotUsingPost,
+	}
+}
+
+// validateResponseError creates a response error for transactional batches and validates
+// that the ErrorCode has been set
+func validateResponseError(e error) error {
```
jhendrixMSFT, line 79: Curious why this is needed?

sdk/data/aztables/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/Azure/azure-sdk-for-go/sdk/data/aztables
 
 go 1.18
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../../azcore
```
jhendrixMSFT, line 5: Can we release the updated `azcore` today?

sdk/data/aztables/error_codes.go
seankane-msft: Something I thought I was going to need but did not. Removed.

sdk/data/aztables/go.mod
seankane-msft: I don't have an issue with that. Let me do the release and prepare portions.

jhendrixMSFT (approved):

Issue #14932, opened 2021-06-29
Add support for challenge based multi-tenant authentication scenario


catalinaperalta:
Similar to: https://github.com/Azure/azure-sdk-for-net/pull/21378

chlowell:
Waiting on https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/296. Public API for this was prematurely implemented in `azcore` and removed by #17640.

chlowell:
Closing this because it's part of #17159 we don't need to track separately.

PR #17525, opened 2022-04-11
[Tables] Multi tenant authentication

seankane-msft:
Adds a challenge policy for clients created with an azidentity credential

#17332


sdk/data/aztables/internal/auth/challenge_policy.go
```diff
@@ -0,0 +1,236 @@
...
+
+type StorageChallengePolicy struct {
+	// mainResource is the resource to be retrieved using the tenant specified in the credential
+	mainResource *ExpiringResource
+	cred         azcore.TokenCredential
+	scope        *string
+	tenantID     *string
+}
+
+func NewStorageChallengePolicy(cred azcore.TokenCredential) *StorageChallengePolicy {
```
christothes, line 36: Should we name this tables instead of storage?

sdk/data/aztables/zc_client_options.go
```diff
@@ -5,15 +5,20 @@ package aztables
...
 )
 
 // ClientOptions are the optional parameters for the NewClient method
 type ClientOptions struct {
 	azcore.ClientOptions
 }
 
-func (c *ClientOptions) toPolicyOptions() *azcore.ClientOptions {
-	return &azcore.ClientOptions{
+func (c *ClientOptions) toPolicyOptions() *policy.ClientOptions {
```
christothes, line 16: should we make it possible to disable the tenant discovery behavior through clientOptions?

seankane-msft: Did you add that for .NET in storage? I don't have it in the KV clients, but if it's included in other languages I can add it in Go

sdk/data/aztables/internal/auth/challenge_policy.go
```diff
@@ -0,0 +1,236 @@
...
+
+// sets the scp.scope and scp.tenantID from the WWW-Authenticate header
+func (scp *TablesChallengePolicy) findScopeAndTenant(resp *http.Response) error {
+	authHeader := resp.Header.Get("WWW-Authenticate")
+	if authHeader == "" {
+		return &challengePolicyError{err: errors.New("response has no WWW-Authenticate header for challenge authentication")}
+	}
+
+	// Strip down to auth and resource
+	// Format is "Bearer authorization=\"<site>\" resource=\"<site>\"" OR
```
chlowell, line 162: I recall Storage challenges as having unquoted values, are these from another RP?

sdk/data/aztables/internal/auth/expiring_resource.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var err error
+	if acquire {
+		// This thread/goroutine has been selected to acquire/update the resource
+		var expiration time.Time
+		resource, expiration, err = er.acquireResource(state)
+
+		// Atomically, update the shared resource's new value & expiration.
+		er.cond.L.Lock()
+		if err == nil {
```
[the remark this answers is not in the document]
jhendrixMSFT: If this code is going to be used in multiple places, can it be moved to `internal` instead?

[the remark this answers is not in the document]
chlowell: I like that better than vendoring our own code but I think we want to extend `azcore/runtime.BearerTokenPolicy` (#17554) and share that instead of `ExpiringResource`.

sdk/data/aztables/client_test.go
```diff
@@ -544,3 +547,39 @@ func TestAzurite(t *testing.T) {
 	}
 	require.Equal(t, 1, count)
 }
+
+func TestMultiTenantAuth(t *testing.T) {
+	recording.LiveOnly(t)
```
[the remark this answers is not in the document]
seankane-msft: The live tests used a SharedKeyCredential, this could be recorded and I can make that change

sdk/data/aztables/internal/auth/challenge_policy.go
seankane-msft: I looked at the response, you are correct. I have updated this

sdk/data/aztables/internal/auth/expiring_resource.go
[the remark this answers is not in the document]
seankane-msft: I agree with Charles, are we ok with adding this for now and migrating later or do we want to make the migration into core now?

sdk/data/aztables/internal/auth/challenge_policy.go
```diff
@@ -0,0 +1,236 @@
...
+
+// acquire acquires or updates the resource; only one
+// thread/goroutine at a time ever calls this function
+func acquire(state interface{}) (newResource interface{}, newExpiration time.Time, err error) {
+	s := state.(acquiringResourceState)
+	tk, err := s.p.cred.GetToken(
+		s.req.Raw().Context(),
+		policy.TokenRequestOptions{
+			Scopes:   []string{*s.p.scope},
+			TenantID: *s.p.scope,
```
chlowell, line 229: This won't work with `azidentity` credentials until we address #14932. I thought we intended to remove this field to avoid suggesting it might have an effect; @jhendrixMSFT do you recall discussing that?

seankane-msft: If that's the case, wouldn't this fail for keyvault which uses the same logic? 
Reference: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/keyvault/internal/challenge_policy.go#L223-L236

chlowell: Only when the resource isn't in the credential's configured tenant. You would need to ensure it isn't in order to test multitenancy (the live test you're adding here doesn't).

jhendrixMSFT: @chlowell we use it today in the implementation for ARM's multitenant auth [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/runtime/policy_bearer_token.go#L29).

Issue #10813, opened 2020-06-24
How to tell if a table has successfully been deleted


pabowers:
Deleting a table is asynchronous so even though the REST endpoint returns as success the table might not actually be deleted at that point.

jhendrixMSFT:
According to the [docs](https://docs.microsoft.com/en-us/rest/api/storageservices/delete-table), if the table is in process of being deleted, any future operations on that table will return a 409 (status conflict) with additional info that the table is being deleted.

jhendrixMSFT:
Apparently there is some corner-case where GET will return a 404 (not a 409 as doc'ed) however when you subsequently try to recreate the table it fails with a 409 as the delete is still in progress.

jhendrixMSFT:
@seankane-msft do you know if this has been fixed in `aztables`?

seankane-msft:
I'm not sure what the solution to this issue is. There is a slight delay by the service to delete a table, and you cannot create a table with the same name immediately afterwards. However, you can loop on a `CreateTable` call when the response is a 409, but there is not built in method in the SDK for retrying this operation.

jhendrixMSFT:
Would it make sense to model Delete as an LRO?  What do the other languages do?

seankane-msft:
Other languages do not model delete as an LRO, we have the same implementation as other languages. I don't believe other languages have a built in way to do any polling for this method.

jhendrixMSFT:
I've updated the docs for `CreateTable` and `Delete` to explain the nuance here.

PR #17570, opened 2022-04-13
Doc updates for aztables

jhendrixMSFT:
Detail behavior when deleting a table.

### DIFF
--- a/sdk/data/aztables/client.go
+++ b/sdk/data/aztables/client.go
@@ -118,6 +118,8 @@ func createTableResponseFromGen(g *generated.TableClientCreateResponse) CreateTa
 
 // CreateTable creates the table with the tableName specified when NewClient was called. If the service returns a non-successful
 // HTTP status code, the function returns an *azcore.ResponseError type. Specify nil for options if you want to use the default options.
+// NOTE: creating a table with the same name as a table that's in the process of being deleted will return an *azcore.ResponseError
+// with error code TableBeingDeleted and status code http.StatusConflict.
 func (t *Client) CreateTable(ctx context.Context, options *CreateTableOptions) (CreateTableResponse, error) {
 	if options == nil {
 		options = &CreateTableOptions{}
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/10813

sdk/data/aztables/client.go
```diff
@@ -653,7 +658,7 @@ func (s *SetAccessPolicyOptions) toGenerated() *generated.TableClientSetAccessPo
 // Specify nil for options if you want to use the default options.
 func (t *Client) SetAccessPolicy(ctx context.Context, options *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
```
jhendrixMSFT, line 659: While fixing the panic, I wonder if `SetAccessPolicyOptions` should actually be optional?  Would one ever call this API with a `nil` options?

jhendrixMSFT: To clarify, should `TableACL` be optional?

sdk/data/aztables/client.go
```diff
@@ -128,6 +128,8 @@ func (t *Client) CreateTable(ctx context.Context, options *CreateTableOptions) (
 
 // Delete deletes the table with the tableName specified when NewClient was called. If the service returns a non-successful HTTP status
 // code, the function returns an *azcore.ResponseError type. Specify nil for options if you want to use the default options.
+// NOTE: deleting a table can take up to 40 seconds or more to complete.  If a table with the same name is created while the delete is still
+// in progress, an *azcore.ResponseError is returned with error code TableBeingDeleted and status code http.StatusConflict.
```
jhendrixMSFT, line 132: Unfortunately, the only operation that returns a 409 is creating the table.  Other operations return a 404 so you can't spin on them to deterministically tell when the table has been deleted.  It means that callers of `CreateTable` would need to be resilient to this 409 case.

jhendrixMSFT: Added doc comment to `CreateTable` explaining this.

sdk/data/aztables/client.go
seankane-msft: That's a good point. TableACL probably should not be optional. I'm surprised this didn't come up in archboard. I wonder if by specifying nil you can remove access policies. Let me investigate.

jhendrixMSFT: I've removed the other changes so this PR is just doc updates now.

seankane-msft (approved):

PR #17571, opened 2022-04-13
[Tables] improving testing for access policies

seankane-msft:
Brought on by Joel's comment [here](https://github.com/Azure/azure-sdk-for-go/pull/17570/files#r849644507), there is a reason to pass in `nil` options for `SetAccessPolicy`, it will remove any previously set access policies. AccessPolicies is `nil`-able and doesn't fit the same mold as what I have generally used in named parameters (that they can not be nil by the service).

### DIFF
--- a/sdk/data/aztables/client.go
+++ b/sdk/data/aztables/client.go
@@ -652,6 +652,7 @@ func (s *SetAccessPolicyOptions) toGenerated() *generated.TableClientSetAccessPo
 // If the service returns a non-successful HTTP status code, the function returns an *azcore.ResponseError type.
 // Specify nil for options if you want to use the default options.
 func (t *Client) SetAccessPolicy(ctx context.Context, options *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
+	if options == nil {options = &SetAccessPolicyOptions{}}
 	response, err := t.client.SetAccessPolicy(ctx, t.name, generated.Enum4ACL, options.toGenerated())
 	if len(options.TableACL) > 5 {
 		err = errTooManyAccessPoliciesError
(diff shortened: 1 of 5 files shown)


sdk/data/aztables/client.go
```diff
@@ -652,6 +652,7 @@ func (s *SetAccessPolicyOptions) toGenerated() *generated.TableClientSetAccessPo
 // If the service returns a non-successful HTTP status code, the function returns an *azcore.ResponseError type.
 // Specify nil for options if you want to use the default options.
 func (t *Client) SetAccessPolicy(ctx context.Context, options *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
+	if options == nil {options = &SetAccessPolicyOptions{}}
 	response, err := t.client.SetAccessPolicy(ctx, t.name, generated.Enum4ACL, options.toGenerated())
 	if len(options.TableACL) > 5 {
```
jhendrixMSFT, line 657: Should this check be moved before the call to the internal client?

seankane-msft: That depends on if we want to do client side validation. We don't do this before sending to the service in other languages so I kept the pattern in Go. It also prevents us from having to issue a fix if the service were to accept more than 5 policies in a request.

sdk/data/aztables/client.go
```diff
@@ -652,8 +652,11 @@ func (s *SetAccessPolicyOptions) toGenerated() *generated.TableClientSetAccessPo
 // If the service returns a non-successful HTTP status code, the function returns an *azcore.ResponseError type.
 // Specify nil for options if you want to use the default options.
 func (t *Client) SetAccessPolicy(ctx context.Context, options *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
+	if options == nil {
+		options = &SetAccessPolicyOptions{}
+	}
 	response, err := t.client.SetAccessPolicy(ctx, t.name, generated.Enum4ACL, options.toGenerated())
-	if len(options.TableACL) > 5 {
+	if err != nil && options != nil && len(options.TableACL) > 5 {
```
jhendrixMSFT, line 659: The nil check can be removed now.

jhendrixMSFT (approved):

PR #17574, opened 2022-04-13
Remove AuthorizationCodeCredential

chlowell:
This credential is awkward to use because it offers no help in acquiring the auth code it requires. It's also redundant next to InteractiveBrowserCredential, which implements the same authentication flow with a better experience for most users. Still, AuthorizationCodeCredential would be useful to apps running in the browser that want more control over the flow, so the JS team is working on an improved design. Since I doubt we have any Go users who need today's AuthorizationCodeCredential, I want to remove it to clear the way for something better (which may simply be InteractiveBrowserCredential, if we can't identify a Go scenario needing more flexibility).


jhendrixMSFT (approved):

sdk/azidentity/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features Added
 
 ### Breaking Changes
+* Removed `AuthorizationCodeCredential`
```
richardpark-msft, line 8: Maybe add a small note of what they should use instead - sounds like it should be InteractiveBrowserCredential?

seankane-msft (approved):

PR #17575, opened 2022-04-13
[azservicebus] Export our log.Event constants 

### LINKED ISSUE #17553: Constants for azservicebus function azlog.SetEvents are not publicly accessible
I was running through README for `azservicebus` and found that the sample code favored passing string literals in the `SetEvents` function.
`azlog.SetEvents("azsb.Conn",...)`
[azservicebus/README.md#L249](https://github.com/Azure/azure-sdk-for-go/blob/6b8b2d958181884c8aa5eedbab4170c8c7a92907/sdk/messaging/azservicebus/README.md?plain=1#L249)
As it is difficult for the user to remember these strings it would be better to replace them with constants which are searchable in IntelliSense. 
@JeffreyRichter and @jhendrixMSFT informed me that constants are available for these string literals, during my hunt for these constants I found out that `azsb` log event constants are in an internal package. 
[azservicebus/internal/log.go#L8](https://github.com/Azure/azure-sdk-for-go/blob/6b8b2d958181884c8aa5eedbab4170c8c7a92907/sdk/messaging/azservicebus/internal/log.go#L8)
I propose we should expose these constants to public so that users would have IntelliSense support and update the README accordingly.
Proposal Example:
```go
import (
    "fmt"
    "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus"
    azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
)

// print log output to stdout
azlog.SetListener(func(event azlog.Event, s string) {
    fmt.Printf("[%s] %s\n", event, s)
})

// pick the set of events to log
azlog.SetEvents(
    // connection/reconnect related events)
    azservicebus.EventConn,
    // authentication events
    azservicebus.EventAuth,
    // receiver specific events
    azservicebus.EventReceiver,
    // management link related events
    azservicebus.EventMgmtLink,
    // retry related events
    azservicebus.EventRetry,
)
```

richardpark-msft:
Made the log constants publicly available so people can use them, rather than having to hardcode/know the strings that are there.

This also helps with intellisense to discover the set of available values.

Also:
- Added example for log constants and one for retry options
- Added a new log.Event for `Sender' (log.EventSender)
- Removed log.EventMgmtLink and just made them all log.EventReceiver or log.EventSender, which made more sense.
- Removed log.EventRetry in favor of inheriting/using whatever the actual client's EventType is.

Fixes #17553

### DIFF
--- a/sdk/messaging/azservicebus/log.go
+++ b/sdk/messaging/azservicebus/log.go
@@ -1,14 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-package internal
+package azservicebus
 
-import (
-	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
-	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/atom"
-)
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
 
-// NOTE: these are intended to mirror the constant values in azservicebus/log.go
 const (
 	// EventConn is used whenever we create a connection or any links (ie: receivers, senders).
 	EventConn log.Event = "azsb.Conn"
@@ -23,5 +19,5 @@ const (
 	EventSender log.Event = "azsb.Sender"
 
 	// EventAdmin is used for operations in the azservicebus/admin.Client
-	EventAdmin log.Event = atom.EventAdmin
+	EventAdmin log.Event = "azsb.Admin"
 )
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Features Added
 
+- Exported log.Event constants for azservicebus. This will maek them easier to
+  discover and they are also documented. The text of log messages themselves 
+  are not guaranteed to be stable. (#TBD)
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/messaging/azservicebus/README.md
+++ b/sdk/messaging/azservicebus/README.md
@@ -247,16 +247,16 @@ azlog.SetListener(func(event azlog.Event, s string) {
 
 // pick the set of events to log
 azlog.SetEvents(
-  // connection/reconnect related events)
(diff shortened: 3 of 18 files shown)


sandeep-sen (approved):
Looks good

sdk/messaging/azservicebus/log.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azservicebus
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+
+const (
+	// EventConn is used whenever we create a connection or any links (ie: receivers, senders).
+	EventConn log.Event = "azsb.Conn"
```
jhendrixMSFT, line 10: This brings up a good question about how SDKs should name their events.

CC @JeffreyRichter

JeffreyRichter: Since only the Azure SDK team can define these strings and we mostly need them for HTTP and some for AMQP, I think the event strings should be geared towards wire protocol more than towards service. Could service bus over AMQP use the same log event strings, for example (I'd hope so).

richardpark-msft: These constants are actually just SB/AMQP (except for the last one, `EventAdmin`, which covers the HTTP/ATOM client).

I'm guessing I can delete that one _soon_ once I transition over to using the normal azcore HTTP pipeline.

richardpark-msft: @JeffreyRichter , @jhendrixMSFT : Were we okay with the prefixing of the sb specific events?

jhendrixMSFT: We need to distill a general pattern.  None of our other events have a prefix, so if we want to do this, then we should be consistent.
I'm on the fence about it.  I see the value in preventing collisions in event names, but it reads weird to me.

JeffreyRichter:
I also prefer no prefix. Can we just have descriptive-enough strings that are likely to be unique (butt not guaranteed to be unique). Since we own these, if we end up with a conflict, we can easily resolve it ourselves

PR #17576, opened 2022-04-13
Rename LoginEndpoint to ActiveDirectoryAuthorityHost

chlowell:
Suggested by @schaabs, seems reasonable to me. "LoginEndpoint" suggests the value could be the URL for token requests when actually we want the base URL. "ActiveDirectoryAuthorityHost" clarifies that we want only the host of an Active Directory (not necessarily _Azure_ Active Directory) instance; "Authority" jives with the environment variable we also read this value from, `AZURE_AUTHORITY_HOST`.


sdk/azidentity/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
 go 1.18
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../azcore
```
jhendrixMSFT, line 5: What's the POR to release `azcore` so we can remove this?

chlowell: Nothing out of the ordinary, just release `azcore` first. `azidentity` is the only other module that references this field.

jhendrixMSFT: Yeah I saw that, no codegen changes (hooray).

jhendrixMSFT (approved):

JeffreyRichter (approved):

PR #17579, opened 2022-04-13
[Core] extending parsing for storage

seankane-msft:
Extending parsing on ResponseError to handle `odata.error` case from storage/tables

### DIFF
--- a/sdk/azcore/internal/shared/response_error.go
+++ b/sdk/azcore/internal/shared/response_error.go
@@ -59,10 +59,8 @@ func extractErrorCodeJSON(body []byte) string {
 			return ""
 		}
 		rawObj = unwrapped
-	}
-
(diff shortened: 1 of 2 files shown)


sdk/azcore/internal/shared/response_error.go
```diff
@@ -61,6 +61,15 @@ func extractErrorCodeJSON(body []byte) string {
 		rawObj = unwrapped
 	}
 
+	// check if this a wrapped odata error, i.e. { "odata.error": { ... } }
+	if wrapped, ok := rawObj["odata.error"]; ok {
```
jhendrixMSFT, line 65: Nit: this should be an `else if` clause to the above `if`.

jhendrixMSFT (approved):

PR #17567, opened 2022-04-13
Added setuping field to NewBlockBlobClientWithSharedKey

godvlpr:
Added setup of one of the field (sharedKey) to `azblob.NewBlockBlobClientWithSharedKey` constructor. Problem was about the using client generated by this constructor.
Steps to reproduce:

Create client using `azblob.NewBlockBlobClientWithSharedKey`
Call `GetSASToken(permissions BlobSASPermissions, start time.Time, expiry time.Time) (SASQueryParameters, error)`
Expecting result is returning structure `azblob.SASQueryParameters`.
Actual result is error: `credential is not a SharedKeyCredential. SAS can only be signed with a SharedKeyCredential.`
It happens be cause `azblob.SASQueryParameters` has rows for checking of existing the `sharedKey`:
```
if b.sharedKey == nil {
		return SASQueryParameters{}, errors.New("credential is not a SharedKeyCredential. SAS can only be signed with a 
        SharedKeyCredential")
	}
```
Problem can be fixed if in the `azblob.NewBlockBlobClientWithSharedKey` field sharedKey will be set.

PR #17588, opened 2022-04-14
Fix/function azblob.new block blob client with shared key

godvlpr:
Added setup of one of the field (sharedKey) to `azblob.NewBlockBlobClientWithSharedKey` constructor. Problem was about the using client generated by this constructor.
Steps to reproduce:

- Create client using `azblob.NewBlockBlobClientWithSharedKey`
- Call `GetSASToken(permissions BlobSASPermissions, start time.Time, expiry time.Time) (SASQueryParameters, error)`
Expecting result is returning structure `azblob.SASQueryParameters`.
Actual result is error: `credential is not a SharedKeyCredential. SAS can only be signed with a SharedKeyCredential`.
It happens be cause `azblob.SASQueryParameters` has rows for checking of existing the sharedKey:
```
if b.sharedKey == nil {
		return SASQueryParameters{}, errors.New("credential is not a SharedKeyCredential. SAS can only be signed with a 
        SharedKeyCredential")
	}
```
Problem can be fixed if in the `azblob.NewBlockBlobClientWithSharedKey` field sharedKey will be set.


godvlpr:
Related PR: https://github.com/Azure/azure-sdk-for-go/pull/17567

Issue #17522, opened 2022-04-11
sdk/azcore: TokenCredential should not be internal


ItalyPaleAle:
### Bug Report

In `sdk/azcore`, the `TokenCredential` interface, as well as related structs such as `TokenRequestOptions` are in an internal package and cannot be used by third-party code.

We are working with APIs that are not supported by the SDKs yet, including some that have no roadmap to be supported. In order to interact with them, we need to retrieve access tokens. Because the `TokenRequestOptions` is internal, we cannot invoke `GetToken` on an `azcore.TokenCredential` object and retrieve an access token. Thus, we cannot make API calls without having to _also_ re-implement the entire auth flow.

- import path of package in question, e.g. `sdk/azcore`
- SDK version e.g. `0.23.0`
- output of `go version`: `go version go1.18 linux/amd64`

chlowell:
`TokenCredential` and `TokenRequestOptions` are exported, though from different packages, so it's possible to implement custom credential types and acquire a token from any credential type in `azidentity`:
```go
package main

import (
    "context"

    "github.com/Azure/azure-sdk-for-go/sdk/azcore"
    "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
    "github.com/Azure/azure-sdk-for-go/sdk/azidentity"
)

// implementing TokenCredential
type CustomCredential struct {}

func (c *CustomCredential) GetToken(ctx context.Context, options policy.TokenRequestOptions) (*azcore.AccessToken, error) {
	return nil, nil
}

var _ azcore.TokenCredential = (*CustomCredential)(nil)

// calling GetToken()
func main() {
	cred, err := azidentity.NewDefaultAzureCredential(nil)
	if err != nil {
		panic(err)
	}
	tk, err := cred.GetToken(
		context.TODO(), policy.TokenRequestOptions{Scopes: []string{"https://management.azure.com/.default"}},
	)
	if err != nil {
		panic(err)
	}
}

var _ azcore.TokenCredential = (*CustomCredential)(nil)
```

ItalyPaleAle:
Thanks for the explanation. I'll take a look at that and will let you know if I run into any issue.

jhendrixMSFT:
For some additional context, there are two reasons why we have internal types that are exported via type aliases.
- To break cyclic dependencies.
- To offer a type in one package (e.g. `azcore`) but its usage in another (e.g. `runtime`).

[...]

PR #17595, opened 2022-04-14
Refactor exported, internal public surface area

jhendrixMSFT:
Internal packages are used to break circular dependencies.  Some of this
internal content is actually publicly exported via type aliases.  To
make it clear which internal parts are public, they've been moved into
the internal/exported package along with comments indicating from where
they're exposed.
Moved some shared stuff that was just for pollers into internal/pollers.
Added shared.BytesSetter interface to decouple the payload caching from
the underlying implementation.
Moved some non-shared constants around.
Removed unused x-ms-date constant.

No functional or public surface area changes.

### DIFF
--- a/sdk/azcore/internal/exported/exported.go
+++ b/sdk/azcore/internal/exported/exported.go
@@ -7,39 +7,13 @@
 package exported
 
 import (
-	"context"
 	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
 )
-
-// TokenRequestOptions contain specific parameter that may be used by credentials types when attempting to get a token.
-// Exported as policy.TokenRequestOptions.
-type TokenRequestOptions struct {
-	// Scopes contains the list of permission scopes required for the token.
-	Scopes []string
-	// TenantID contains the tenant ID to use in a multi-tenant authentication scenario, if TenantID is set
-	// it will override the tenant ID that was added at credential creation time.
-	TenantID string
-}
-
-// TokenCredential represents a credential capable of providing an OAuth token.
-// Exported as azcore.TokenCredential.
-type TokenCredential interface {
-	// GetToken requests an access token for the specified set of scopes.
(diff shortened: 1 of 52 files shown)


sdk/azcore/internal/exported/exported.go
```diff
@@ -0,0 +1,143 @@
...
+	// Scopes contains the list of permission scopes required for the token.
+	Scopes []string
+	// TenantID contains the tenant ID to use in a multi-tenant authentication scenario, if TenantID is set
+	// it will override the tenant ID that was added at credential creation time.
+	TenantID string
+}
+
+// TokenCredential represents a credential capable of providing an OAuth token.
+// Exported as azcore.TokenCredential.
+type TokenCredential interface {
```
chlowell, line 32: Should we just define this in azcore? That doesn't appear to create a cycle

jhendrixMSFT: We could.  In fact, looks like there might be a few others too (I'll play around with it).  My only concern is if this might cause future hardships.  I've been trying to keep the dependencies flowing one way, i.e. parents can reach into children and cross-sibling (there are a few exceptions).  Maybe it will be ok.

jhendrixMSFT: The other cases introduce cycles, so this looks to be the only case.

I'm tempted to leave this as-is to avoid introducing more cycles.

chlowell: I lean toward defining things where they're exported whenever possible, to get better documentation and code completion. Is moving a definition behind an alias breaking in some way?

chlowell (approved):

sdk/azcore/internal/exported/exported.go
jhendrixMSFT: No, it would be compatible.  We'd just have to undo the change that directly exports it (touches 11 files).  Not terrible but might not be obvious at first glance.

jhendrixMSFT: Fixed

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17522

PR #17596, opened 2022-04-14
[azservicebus] Move re-exported constants into a more recognizable subpackage.

richardpark-msft:
Move any constants/types that get re-exported from `internal` to the public package to the  'azservicebus/internal/exports' sub package to make them easier to keep track of.


richardpark-msft:
@RickWinter - this should help a bit with making it visually more prominent. I'm sure we could write a lint rule as well. I was using a 'grep' to find most of it.

And since it's a regex I'm sure it could have been even more simple ;)
```bash
grep -n -P "^\s*type\s+[^ ]+.+\s+=\s+(exports|internal|atom|utils|amqpwrap|sbauth|test|tracing)\..*$" *.go
```

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/atom"
-	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/utils"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/exports"
```
jhendrixMSFT, line 15: I used package named `exported` for similar changes in `azcore` as I felt it read better.  Minor issue but would be good to be consistent.  And if the majority thinks `exports` is better I can update my PR accordingly.

JeffreyRichter: Yes, I like exports. A "utils" package is explicitly frowned upon by the Go community.

JeffreyRichter (approved):

sdk/messaging/azservicebus/admin/admin_client.go
richardpark-msft: @jhendrixMSFT , went with 'exported'

PR #17598, opened 2022-04-15
[azservicebus] Allow link creation to be cancelled

### LINKED ISSUE #17565: Unable to cancel client.AcceptNextSessionForQueue(..) function
pkg: github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus

SDK version: v0.4.0

go version: go 1.18

What happened?

function client.AcceptNextSessionForQueue(ctx, queueName, options) is not cancellable by context, passed as first argument. For example, if I want to interrupt this function by signal or by timeout, I need to wait about 1 minute before function exits with error "context canceled".

richardpark-msft:
AcceptNextSessionFor(Queue|Subscription) can block for a long time (server dependent) if there are no available sessions. HOWEVER, it was intended to be cancellable, which wasn't working.

This is a simpler workaround until we get context support plumbed through go-amqp itself.

Fixes #17565


sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap.go
```diff
@@ -78,22 +78,22 @@ func (w *AMQPClientWrapper) NewSession(opts ...amqp.SessionOption) (AMQPSession,
...
 
 	return &AMQPSessionWrapper{
-		inner: sess,
+		Inner: sess,
 	}, nil
 }
 
 type AMQPSessionWrapper struct {
-	inner *amqp.Session
+	Inner *amqp.Session
```
jhendrixMSFT, line 86: Why is this exported now?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -507,15 +507,47 @@ func (e *entity) SetSubQueue(subQueue SubQueue) error {
...
+func createReceiverLink(ctx context.Context, session amqpwrap.AMQPSession, linkOptions []amqp.LinkOption) (internal.AMQPReceiverCloser, error) {
+	// If you're doing an AcceptNextSession it's possible for this call to take a long time before timing out
+	// on its own (it's by design - it's waiting for any empty session to become available).
+	type ret = struct {
+		Receiver internal.AMQPReceiverCloser
+		Err      error
 	}
 
-	return amqpReceiver, nil
+	done := make(chan ret, 1)
```
jhendrixMSFT, line 518: I don't think this needs to be a buffered channel does it?

sdk/messaging/azservicebus/internal/amqp_test_utils.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/amqpwrap"
```
jhendrixMSFT, line 11: Unrelated to this PR.

Is the contents of this file used only for tests?  If so, shouldn't the name be `amqp_utils_test.go` to limit it to tests?

richardpark-msft: I wanted that too - it's re-used by tests in the azservicebus package and sub-package _test.go files aren't available in that case.

sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap.go
richardpark-msft: I'm wrapping a session in the `azservicebus` package so I need to set it when I create the AMQPSessionWrapper.

I could create a constructor func but it seemed like it'd be more work for me in the long term since we have just the single field.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: You were correct! In fact, it was better unbuffered.

jhendrixMSFT (approved):

PR #17616, opened 2022-04-16
[azservicebus] Adding authorization and subscription rule support

### LINKED ISSUE #16045: [azservicebus] AdminClient - Rule management functions
#15090 covered the primary entities (Queues,Topics,Rules). There are two small bits of work remaining:
- [x] GetNamespaceProperties (added)
- [ ] Topic Rule management (#17616)
- [ ] Authorization rule management (#17616)

richardpark-msft:
Adding in support for:
* Getting/updating AuthorizationRules for topics and queues
* CRUD for working with Subscription rules.

Fixes #16045


sdk/messaging/azservicebus/CHANGELOG.md
```diff
@@ -5,8 +5,10 @@
 ### Features Added
 
 - Exported log.Event constants for azservicebus. This will maek them easier to
```
jhendrixMSFT, line 7: /maek/make/s

sdk/messaging/azservicebus/admin/admin_client_models.go
```diff
@@ -16,3 +22,87 @@ const (
...
+		PrimaryKey *string
+
+		// SecondaryKey for this rule.
+		SecondaryKey *string
+	}
+)
+
+const (
+	// AccessRightManage allows changes to an entity.
+	AccessRightManage = "Manage"
```
jhendrixMSFT, line 54:
```go
AccessRightManage AccessRight = "Manage`
```

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -214,3 +215,22 @@ func (ep *entityPager[TFeed, T, TOutput]) Fetcher(ctx context.Context) ([]TOutpu
 
 	return finalItems, nil
 }
+
+// mapATOMError checks if the error is a legitimate 404 or a "fake" 404 (where the service succeeded but gave us back an
+// empty feed instead). This "fake" behavior comes about because the API here is not truly a CRUD API (it's extremely close)
+// so we have to do some small workarounds.
+// NOTE: we had a debate about whether to return a nil instance or try to fabricate an HTTP 404 response instead (even if
+// one didn't come back) and went with 'nil' to avoid having a fake HTTP response, which would have been confusing.
+func mapATOMError[T any](err error) (*T, error) {
```
jhendrixMSFT, line 224: Can this type parameter be removed?  The `T` isn't actually used, it's always `nil`.

sdk/messaging/azservicebus/admin/admin_client_models.go
```diff
@@ -16,3 +22,87 @@ const (
...
+	// AccessRightSend allows you to send messages to this entity.
+	AccessRightSend = "Send"
+	// AccessRightListen allows you to receive messages from this entity.
+	AccessRightListen = "Listen"
+)
+
+func internalAccessRightsToPublic(internalRules []atom.AuthorizationRule) []AuthorizationRule {
```
jhendrixMSFT, line 61: Are we using a generated ATOM layer now, which is why we need to perform these translations?

jhendrixMSFT:
Does this include the fix for https://github.com/Azure/azure-service-bus-go/issues/204?

sdk/messaging/azservicebus/admin/admin_client.go
richardpark-msft: I want to leave it since it makes this code more compact:

https://github.com/Azure/azure-sdk-for-go/pull/17616/files/7642f0d8b6bdb6cbf41e726014e5533c9c9606bc#diff-c0a58bbe8d6af602f609af18f19b0873cb94974f9ae461a8a12ab800cb8111f3R183

```go
func (ac *Client) GetQueue(ctx context.Context, queueName string, options *GetQueueOptions) (*GetQueueResponse, error) {
  // other code
	if err != nil {
		return mapATOMError[GetQueueResponse](err)
  }
}
```

Otherwise I'd need to do another check to return the right type of nil, etc...

richardpark-msft:
> Does this include the fix for [Azure/azure-service-bus-go#204](https://github.com/Azure/azure-service-bus-go/issues/204)? I think the answer is yes but looking for an ack.

It does, we deserialize the "map[string]interface{}" by hand, since it can't be done automatically. (ie: KeyValueOfstringanyType).

sdk/messaging/azservicebus/admin/admin_client_models.go
richardpark-msft: Done

sdk/messaging/azservicebus/admin/admin_client_rules.go
```diff
@@ -0,0 +1,558 @@
...
+	"strconv"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/atom"
+)
+
+// Filters allows you to specify a single filter for a particular rule.
+// Only one of these properties can be set.
```
jhendrixMSFT, line 19: Is this how other languages model the filters?  It seems a little clumsy to have a type with all of these mutually exclusive fields.  Can be refined in a separate PR as necessary.

sdk/messaging/azservicebus/admin/admin_client_rules.go
```diff
@@ -0,0 +1,558 @@
...
+}
+
+// ListRulesOptions contains the optional parameters for Client.ListRules
+type ListRulesOptions struct {
+	// MaxPageSize is the maximum size of each page of results.
+	MaxPageSize int32
+}
+
+// ListRules creates a pager that can list rules for a subscription.
+func (ac *Client) ListRules(topicName string, subscriptionName string, options *ListRulesOptions) *runtime.Pager[ListRulesResponse] {
```
jhendrixMSFT, line 202: This should be named `NewListRulesPager(...)`

sdk/messaging/azservicebus/admin/admin_client_rules.go
richardpark-msft: The other languages basically have the equivalent of an 'any' field for the filter, which I believe we could also do. So the user would have to do a casting but that would work. 

What do you think of that?

sdk/messaging/azservicebus/admin/admin_client_rules.go
richardpark-msft: Will rename (all others too)

sdk/messaging/azservicebus/admin/admin_client_rules.go
richardpark-msft: (also, for better or for worse, I modeled it after this: https://github.com/Azure/azure-sdk-for-go/blob/8d5dc19fa7e427d90690ecc116feb8159c9e9c45/services/servicebus/mgmt/2017-04-01/servicebus/models.go#L2176)

jhendrixMSFT: Let's schedule some time with @JeffreyRichter to come up with a design.  I assume that there's no urgency in releasing this and we can clean it up in a separate PR later.

richardpark-msft: I changed it to any 'any' field for now anyways. We can work on it more but this is at least inline with what other languages do.

sdk/messaging/azservicebus/admin/admin_client_rules.go
richardpark-msft: Renamed

sdk/messaging/azservicebus/admin/admin_client_rules.go
```diff
@@ -0,0 +1,633 @@
...
+// which properly represents this type.
+type UnknownRuleAction struct {
+	// Type is the Service Bus type for this action.
+	Type string
+
+	// RawXML is the raw XML for this action that could not be parsed.
+	RawXML string
+
+	// RawAttrs are attributes for the raw XML element that wasn't covered by our known properties.
+	RawAttrs []xml.Attr
```
jhendrixMSFT, line 65: Why are the attributes separate, I would have thought them to be in `RawXML`.

richardpark-msft: Agree, I'll unify.

richardpark-msft: Unified. 

I also just made all the "RawXML" fields just be a byte slice, just to remove one unneeded conversion when calling xml.Unmarshal (should a user choose to do so).

sdk/messaging/azservicebus/admin/admin_client.go
```diff
@@ -20,6 +21,11 @@ import (
 // NOTE: For sending and receiving messages you'll need to use the `azservicebus.Client` type instead.
 type Client struct {
 	em atom.EntityManager
+
+	// For testing: treatFiltersAndActionsAsUnknown causes the deserialization of rule filters and rule actions
+	//  to behave as if they were unhandled types, forcing a return of an "unknown" filter. We use this to test that
+	// rule round-tripping works, even for types added after this library was published.
+	treatFiltersAndActionsAsUnknown bool
```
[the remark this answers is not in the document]
richardpark-msft: Now that I've seen it work in reality a bunch of times I can't see a reason to. Let me see how bad that'll be.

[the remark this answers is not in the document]
richardpark-msft: Fixed. I just do a replace of the entityManager when I'm running the test so the code runs the same in tests or without.

PR #17671, opened 2022-04-20
Renamed pager APIs per new guidelines

jhendrixMSFT:
Regenerated code with latest code generator (no public changes).
Added directive to autorest.md to fix up parts of codegen.

### DIFF
--- a/sdk/data/aztables/autorest.md
+++ b/sdk/data/aztables/autorest.md
@@ -11,12 +11,27 @@ clear-output-folder: false
 output-folder: internal
 tag: package-2019-02
 credential-scope: none
-use: "@autorest/go@4.0.0-preview.37"
-module-version: 0.7.0
+use: "@autorest/go@4.0.0-preview.39"
+module-version: 0.8.0
 security: "AADToken"
 security-scopes: "https://storage.azure.com/.default"
 modelerfour:
   group-parameters: false
+
+directive:
+  - from: source-file-go
+    where: $
+    transform: >-
+      return $.
+        replace(/moduleName\s+=\s+"internal"/, `moduleName = "aztables"`).
+        replace(/\(client \*TableClient\) deleteEntityCreateRequest\(/, `(client *TableClient) DeleteEntityCreateRequest(`).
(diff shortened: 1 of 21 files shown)


sdk/data/aztables/autorest.md
```diff
@@ -11,17 +11,32 @@ clear-output-folder: false
...
+      return $.
+        replace(/moduleName\s+=\s+"internal"/, `moduleName = "aztables"`).
+        replace(/\(client \*TableClient\) deleteEntityCreateRequest\(/, `(client *TableClient) DeleteEntityCreateRequest(`).
+        replace(/\(client \*TableClient\) insertEntityCreateRequest\(/, `(client *TableClient) InsertEntityCreateRequest(`).
+        replace(/\(client \*TableClient\) mergeEntityCreateRequest\(/, `(client *TableClient) MergeEntityCreateRequest(`).
+        replace(/\(client \*TableClient\) updateEntityCreateRequest\(/, `(client *TableClient) UpdateEntityCreateRequest(`).
+        replace(/= client\.deleteEntityCreateRequest\(/, `= client.DeleteEntityCreateRequest(`).
+        replace(/= client\.insertEntityCreateRequest\(/, `= client.InsertEntityCreateRequest(`).
+        replace(/= client\.mergeEntityCreateRequest\(/, `= client.MergeEntityCreateRequest(`).
+        replace(/= client\.updateEntityCreateRequest\(/, `= client.UpdateEntityCreateRequest(`);
```
chlowell, line 34: Why doesn't the generator export these methods?

chlowell (approved):

sdk/data/aztables/autorest.md
jhendrixMSFT: We export them in track 1, however it adds a lot of noise to the API-set to cater to an uncommon case.  In addition, the original thinking for data-plane was that the generated content would be used from the same package as the convenience layer (e.g. `azblob`), so there would be no need to call these directly.  If we think that more data-plane packages will be using this pattern, we could switch to exporting these by default.

Issue #17698, opened 2022-04-21
Fold arm/runtime/poller.go logic into runtime/poller.go


jhendrixMSFT:
After the LRO refactoring, we might no longer need separate poller.go files.

To do:
Testing end-2-end
Flushing out the bugs :)

jhendrixMSFT:
Keeping this separate.

jhendrixMSFT:
Fixed in upcoming release of `azcore`.

PR #17708, opened 2022-04-22
Merge arm/runtime/poller.go into runtime/poller.go

jhendrixMSFT:
LRO refactoring has alleviated the need for separation of ARM and
data-plane pollers so the content has been merged into the runtime
package.  The duplicated content under arm/runtime will be removed in a
future release.
The arm/internal/pollers/loc package was renamed to armloc to avoid
colliding with the existing data-plane package.  These will need to be
consolidated in a separate change.
No functional changes to the relocated, internal, pollers.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17698

jhendrixMSFT:
I've tested the changes with the autorest test suite in `autorest.go` and everything is working as expected.

sdk/azcore/arm/runtime/poller.go
```diff
@@ -4,6 +4,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+// Deprecated: use azcore/runtime instead.
```
chlowell, line 7: Does this put a deprecation notice on the entire package?

jhendrixMSFT: No, in fact it has no discernible effect, other than on the reader if they happen to see this.  I felt like I should add _something_, but perhaps the entry in the changelog is enough.

I don't feel too strong about this, so I can remove it if there are any concerns.

chlowell: I was curious enough to try this out. godoc apparently uses the first line of comment it finds on a package declaration as the synopsis of that package:
image
But this probably has no impact because there's no semantic effect and the audience is probably only SDK devs.

chlowell (approved):

sdk/azcore/arm/runtime/poller.go
jhendrixMSFT: Thanks for checking the docs.  I've updated the deprecation notice.

PR #17716, opened 2022-04-23
Get ContentMD5 for ListBlobsFlat response.

mdales:
The server returns this information, but it is currently not propagated back to the caller.

Interestingly the API returns a base64 encoded string, but the other APIs (and the old API) use []byte to store ContentMD5, so I've implemented the conversion at the unmarshalling time for consistency.

### DIFF
--- a/sdk/storage/azblob/CHANGELOG.md
+++ b/sdk/storage/azblob/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+* ContentMD5 is returned for ListBlobsFlat where appropriate.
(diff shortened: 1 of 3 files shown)


mdales:
Guidance would be appreciated on the test failure. The test I have added fails with:

```        	
Error:      	Required environment variable not set: AZURE_STORAGE_ACCOUNT_NAME
```

Which I hit locally, and just assumed I needed a storage account for testing. However, my test is based on other tests in the suite, so I'm not sure why my test is failing this way but say TestRehydrateStatus isn't.

PR #17719, opened 2022-04-24
[Test] sdk/resourcemanager/resources/armresources/0.5.1

Alancere:


sdk/resourcemanager/resources/armresources/ze_generated_example_providers_client_test.go
```diff
@@ -54,7 +54,6 @@ func ExampleProvidersClient_NewListPager() {
 		nextResult, err := pager.NextPage(ctx)
 		if err != nil {
 			log.Fatalf("failed to advance page: %v", err)
-			return
 		}
 		for _, v := range nextResult.Value {
 			// TODO: use page item
```
RickWinter, line 59: Will you be adding validation of the action and return types?

tadelesh: For example, no. We do validation in mock test. I explained it with the mail reply.

Issue #17450, opened 2022-04-05
Reconsider exporting symbols from internal modules


chlowell:
[`azcore/log`](https://github.com/Azure/azure-sdk-for-go/blob/a7dfae471f382c8e7bf80492cc054d2566ae4c9f/sdk/azcore/log/log.go#L15-L31) exports a type and constants from `sdk/internal/log`. This instance is low risk but the pattern is troubling in general because it makes `sdk/internal` not really internal. We should discuss whether to allow this pattern and, if we decide to allow it, what guard rails to place around `internal` modules (we'd want them around e.g. `keyvault/internal` as well).

JeffreyRichter:
Sometimes we do this because of dependency management. Internal defines symbols we need low in the system (like by Core) and we also expose them publicly. I think that is the case here. Joel would know for sure.

-- Jeffrey Richter
________________________________

richardpark-msft:
I have a similar situation in `azservicebus` - I have a set of constants that represent the receive mode and I use them in both `azservicebus/internal` and `azservicebus` code.

chlowell:
I think it's okay for a module to re-export things from its own `internal` subpackages when necessary. I'm concerned about internal modules like `sdk/internal`. These are nominally private, accessible only to SDK authors who may not know parts of them are functionally public due to being re-exported by some public module.

chkohner:
Also, if you're going to export internals, please make sure they are fully exported. Atm, you export `azcore.TokenCredential = shared.TokenCredential` but not the corresponding TokenRequestOptions used as a parameter (thus making the interface unusable in practice). Also, it's unclear from a consumer perspective what the distinction between azcore and azpolicy is. (Which, btw, `azpolicy.TokenRequestOptions` IS exporting the shared internal atm, where azcore is not...)

chkohner:
As an additional note, if you're considering NOT exporting key internals at all, that'd be really unfortunate. If they are exported (correctly), I can at least leverage key components (e.g. `azidentity` for managed identities) to make generic REST calls to Azure endpoints that don't have great Go SDK support yet.

jhendrixMSFT:
I agree that `azcore` shouldn't be exporting content from `internal`.  I think some of this design predates when we refactored `azcore`.  The only wrinkle here is how to handle `Should()`, `Write()`, and `Writef()`.  The current implementation permits only SDK authors to call these funcs as they live in `internal` which at some point we thought was desirable.  How much do we really care about that?

[...]

PR #17729, opened 2022-04-25
Removed log constants

jhendrixMSFT:
They should be defined in azcore/log, not here.
Added comments indicating which parts of log are public surface area.

### DIFF
--- a/sdk/internal/log/log.go
+++ b/sdk/internal/log/log.go
@@ -12,31 +12,12 @@ import (
 	"time"
 )
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// NOTE: The following are exported as public surface area from azcore.  DO NOT MODIFY
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
 // Event is used to group entries.  Each group can be toggled on or off.
 type Event string
-
-const (
-	// EventRequest entries contain information about HTTP requests.
-	// This includes information like the URL, query parameters, and headers.
-	EventRequest Event = "Request"
-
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17450

sdk/internal/log/log.go
```diff
@@ -12,32 +12,13 @@ import (
 	"time"
 )
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// NOTE: The following are exported as public surface area from azcore.  DO NOT MODIFY
```
chlowell, line 16: Why doesn't this stuff live in `azcore`?

jhendrixMSFT: Logging is a very oddball, and hopefully rare, case.  It's a mix of public and internal surface area.  We only want customers to interact with a small subset of the APIs (the ones in this comment block).  The rest are intended for SDK authors so they can't go in `azcore`.

jhendrixMSFT: We could define `Event` in `azcore` and use `string` in `internal`, we'd just have to do a little casting (not terrible, just a little ugly).  However, we'll still need a comment block indicating that some of these APIs are public surface area.

jhendrixMSFT: Thinking a bit more about that, I'm not in favor of it as SDK authors would also have to perform casting.

richardpark-msft (approved):
LGTM. Approved so long as @chlowell is good with the answer on the azcore/internal split.

chlowell (approved):

PR #17746, opened 2022-04-27
Revise azidentity doc comments

chlowell:
I revised some content, but my main goal is to make doc comments look better when rendered by godoc. For example, compare EnvironmentCredential [before](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#EnvironmentCredential) and after: image


sdk/azidentity/chained_token_credential.go
```diff
@@ -17,16 +17,14 @@ import (
...
+	// authenticate through each source in turn, stopping when one succeeds. When false, the credential authenticates
+	// only through this first successful source--it never again tries the sources which failed.
 	RetrySources bool
 }
 
-// ChainedTokenCredential is a chain of credentials that enables fallback behavior when a credential can't authenticate.
-// By default, this credential will assume that the first successful credential should be the only credential used on future requests.
-// If the `RetrySources` option is set to true, it will always try to get a token using all of the originally provided credentials.
+// ChainedTokenCredential links together multiple credentials and tries them sequentially when authenticating. By default,
+// it tries all the credentials until one authenticates, after which it always uses that credential.
```
richardpark-msft, line 27: Should the second sentence of this be removed since it contradicts what the ChainedTokenCredentialOption is saying for 'RetrySources'? Or am I misreading it?

sdk/azidentity/environment_credential.go
```diff
@@ -25,29 +25,36 @@ type EnvironmentCredentialOptions struct {
...
-// - AZURE_TENANT_ID: (optional) tenant to authenticate in. If not set, defaults to the "organizations" tenant, which
-//   can authenticate only Azure Active Directory work or school accounts.
+// AZURE_CLIENT_ID: the service principal's client ID
+//
+// AZURE_CLIENT_SECRET: one of the service principal's client secrets
+//
+// Service principal with certificate
+//
+// AZURE_TENANT_ID: ID of the service principal's tenant. Also called its "directory" ID.
+//
```
richardpark-msft, line 39: It seems slightly harder to see that some of these options are grouped since everything has an extra newline after it (IMO).

richardpark-msft (approved):
Looks good. 

If I'm getting the idea we're just trying to remove the `param:` style comments and adding in some docs to indicate that you can pass nil to options (which sometimes people have trouble figuring out). Is that right?

sdk/azidentity/environment_credential.go
richardpark-msft: Ah, I just noticed your PR description. Ignore this one.

chlowell:
That's the gist. The prior comment formatting looked good in source code but terrible when rendered because godoc doesn't do anything special with parameter names.

PR #17760, opened 2022-04-27
AppConfig: inline method parameters instead of using Setting struct

### LINKED ISSUE #17264: App Configuration: Setting semantics
But this argument is for getting a setting - it has different semantics than ListSetting().

Is there a reason we're not just using 'key' as a positional argument?

_Originally posted by @richardpark-msft in https://github.com/Azure/azure-sdk-for-go/pull/17240#discussion_r822956769_

antkmsft:
We were using `Setting` struct as input parameter for client methods.

But it was not clear, which ones are mandatory (`key`, sometimes nullable `value`), which are optional (`label`), and which are ignored (the rest).

It was also confusing with `SetReadOnly()`: setting does have `IsReadOnly` field, and also `SetReadOnly()` has mandatory boolean argument. Now, `SetReadOnly()` takes `key`, and boolean `isReadOnly`.

Closes #17264.

### DIFF
--- a/sdk/appconfig/azappconfig/client.go
+++ b/sdk/appconfig/azappconfig/client.go
@@ -140,11 +140,19 @@ func fromGeneratedAdd(g generated.AzureAppConfigurationClientPutKeyValueResponse
 
 // AddSettingOptions contains the optional parameters for the AddSetting method.
 type AddSettingOptions struct {
-	// placeholder for future options
+	// Configuration setting label.
+	Label *string
 }
 
 // AddSetting creates a configuration setting only if the setting does not already exist in the configuration store.
-func (c *Client) AddSetting(ctx context.Context, setting Setting, options *AddSettingOptions) (AddSettingResponse, error) {
+func (c *Client) AddSetting(ctx context.Context, key string, value *string, options *AddSettingOptions) (AddSettingResponse, error) {
+	var label *string
+	if options != nil {
+		label = options.Label
+	}
+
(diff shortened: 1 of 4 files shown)


sdk/appconfig/azappconfig/client.go
```diff
@@ -140,11 +140,19 @@ func fromGeneratedAdd(g generated.AzureAppConfigurationClientPutKeyValueResponse
...
 // AddSettingOptions contains the optional parameters for the AddSetting method.
 type AddSettingOptions struct {
-	// placeholder for future options
+	// Configuration setting label.
+	Label *string
 }
 
 // AddSetting creates a configuration setting only if the setting does not already exist in the configuration store.
-func (c *Client) AddSetting(ctx context.Context, setting Setting, options *AddSettingOptions) (AddSettingResponse, error) {
+func (c *Client) AddSetting(ctx context.Context, key string, value *string, options *AddSettingOptions) (AddSettingResponse, error) {
```
jhendrixMSFT, line 148: Is specifying a `nil` value for `value` the way to "erase" a setting?  If so it would be good to document that, else it gives the appearance that `value` is optional.

antkmsft: Value can be `nil`, or can be `""`. AppConfig does distinguish between them, i.e. if you set `nil`, you read `nil` back, the same for `""`. In the Azure Portal, you can't tell the difference.

PR #17773, opened 2022-04-28
Use case-insensitive comparison for success case

jhendrixMSFT:

### DIFF
--- a/sdk/azcore/internal/pollers/util.go
+++ b/sdk/azcore/internal/pollers/util.go
@@ -78,6 +78,11 @@ func Failed(s string) bool {
 	return strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled)
 }
 
+// Succeeded returns true if the LRO's state is terminal success.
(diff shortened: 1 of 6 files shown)


sdk/azcore/internal/pollers/util.go
```diff
@@ -78,6 +78,11 @@ func Failed(s string) bool {
 	return strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled)
 }
 
+// Succeeded returns true if the LRO's state is terminal success.
+func Succeeded(s string) bool {
+	return strings.EqualFold(s, StatusSucceeded)
```
richardpark-msft, line 83: Do we not control the constant value here? Why would it change case?

richardpark-msft: Talked with @jhendrixMSFT - the value of the string is part of the contract but the casing of the string isn't fixed from other services.

richardpark-msft (approved):
Best EqualFold() call I've seen this year.

PR #17786, opened 2022-04-30
[azservicebus] Adding `azservicebus.Error` to represent errors that a customer can use programatically.

### LINKED ISSUE #17283: [azservicebus] Provide a Service Bus error type that customers can use for actionable errors.
(This has come up in a few different customer issues (ex: [#16088](https://github.com/Azure/azure-sdk-for-go/issues/16088#issuecomment-1062454958)), and it's worth a discussion.)

From looking at a few customers programs, I've seen a pattern that basically looks like this:

```go
for {
  client, receiver := GetReceiver()

ReceiveLoop:
  for {
    messages, err := receiver.ReceiveMessages(5)

    if err != nil {
      // customer might add in their own backoff.
      receiver.Close()
      client.Close()

      client, receiver = GetReceiver()
      continue
    }

    for _, msg := range messages {
      err = receiver.CompleteMessage(msg)

      if err != nil {
        // customer might add in their own backoff.
        client, receiver = GetReceiver()
        continue ReceiveLoop
      }
    }
  }
}
```

Where basically any error is treated as a "restart the client". This is probably typical and is basically a variant on "error from stateless call, probably retry it all". 

I think it's worthwhile to consider if there are some areas where that kind of retry logic would be wasteful or incorrect and it'd be worth adding in a programmatically useful type.

Something similar to this:

```go
package azservicebus

type Error struct {
  Code Code
  Inner error
}

func (e Error) Error() string { 
  return e.Inner.Error()
}

// NOTE: we should also make this work with errors.As()

type Code string

const (
  // (this is my first pass, so there might be more, just documenting one that immediately occurs to me).
  CodeNotFound = "notfound"
  CodeLockLost = "locklost"
  CodeUnauthorized = "unauthorized"
)
```

### Message lock lost (ie: 'locklost')

Messages (in peek lock mode) have a lock, which expires after a set period of time. Restarting a connection/link here would be unnecessary and expensive, so the "standard" error handling in the top level example would not be a good idea.

There are two internal paths that can happen when settling, both of which return separate error types:

```
// go-amqp amqp.Error, if you settled via normal AMQP constructs

{
  Condition:"com.microsoft:message-lock-lost"
  Description:"The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. For more information please see https://aka.ms/ServiceBusExceptions . Reference:a5507261-5c0a-4aee-9df8-8af1326dea2a, TrackingId:8dd5ab0300007e6b008d6d84622bde3d_G38_B17, SystemTracker:<host>:Queue:queue-16db7812269186b4, Timestamp:2022-03-11T23:43:33"
  Info:map[string]interface {} nil
}
```

We also have a mode called 'backup' message settlement, where we use a separate link. When we do this the error comes back as an RPCError{}, since it happens on our RPC $management link, not the original receiver link.

```
// azservicebus rpcError (it's not exposed right now)
{
  RPCResponse: {
  	Code: 410
    Description: "The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. For more information please see https://aka.ms/ServiceBusExceptions . Reference:a0ecf2d4-3f4f-4b72-9604-ce936e2c78e2, TrackingId:ed803096-27bb-401b-a037-ea12471d9259_B17, SystemTracker:<host>:Queue:queue-16db7aaedfe1797b, Timestamp:2022-03-12T00:31:03"
    Message: <the actual AMQP message received back from the RPC link>
  },
  RPCCode: 410,     // (replicated from the actual RPCResponse)
}
```

Customers would need to check both to check for the single logical error.

### Invalid permissions/Entity not found ('unauthorized'/'notfound')

These are "fail and fail hard" type of errors. They're probably deployment issues or similar, and will require human intervention. The "retry and recreate" logic in the sample at the top would also be bad here.

You'd normally get something like a 401 or possibly a 404 in HTTP parlance, and we could do something similar here. This gives customers that want to a chance to do something indicating the fatal nature of these problems.

### Possibly common questions

- Let the customer trigger off the actual error type 

  This ends up being somewhat messy. In some cases the error can come from different code paths, and won't be the same type. See the message lock lost section above for an example.

  In addition, this would place a dependency on the customer's code on lower level libraries like go-amqp, basically making it's public surface _our_ public surface. We don't do that today.

- Can we re-use some existing error type?
   
  `azcore.ResponseError` is the closest analogue to this, but won't work since it assumes you have a net.Response and is basically HTTP centric.

- Do we need string error codes?

  No, but it seemed nice to avoid the temptation of trying to shoe-horn all the error codes into what might appear to be http status codes.

richardpark-msft:
Created `azservicebus.Error` to represent errors that a customer can use programatically.

azservicebus.Error contains a code that customers can use, which is considered stable and part of the supported API. 

Today we only have two codes - CodeLockLost, which comes back if the lock has expired and CodeConnectionLost, which comes back if we failed on an error that is retryable.

Fixes #17283

### DIFF
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -217,6 +217,7 @@ func (client *Client) NewSender(queueOrTopic string, options *NewSenderOptions) 
 
 // AcceptSessionForQueue accepts a session from a queue with a specific session ID.
 // NOTE: this receiver is initialized immediately, not lazily.
+// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptSessionForQueue(ctx context.Context, queueName string, sessionID string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -243,6 +244,7 @@ func (client *Client) AcceptSessionForQueue(ctx context.Context, queueName strin
 
 // AcceptSessionForSubscription accepts a session from a subscription with a specific session ID.
 // NOTE: this receiver is initialized immediately, not lazily.
+// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptSessionForSubscription(ctx context.Context, topicName string, subscriptionName string, sessionID string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -270,6 +272,7 @@ func (client *Client) AcceptSessionForSubscription(ctx context.Context, topicNam
 
 // AcceptNextSessionForQueue accepts the next available session from a queue.
 // NOTE: this receiver is initialized immediately, not lazily.
+// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptNextSessionForQueue(ctx context.Context, queueName string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -296,6 +299,7 @@ func (client *Client) AcceptNextSessionForQueue(ctx context.Context, queueName s
 
 // AcceptNextSessionForSubscription accepts the next available session from a subscription.
 // NOTE: this receiver is initialized immediately, not lazily.
+// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptNextSessionForSubscription(ctx context.Context, topicName string, subscriptionName string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -217,7 +217,7 @@ func (client *Client) NewSender(queueOrTopic string, options *NewSenderOptions) 
 
 // AcceptSessionForQueue accepts a session from a queue with a specific session ID.
 // NOTE: this receiver is initialized immediately, not lazily.
-// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
+// If the operation fails it can return an *azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptSessionForQueue(ctx context.Context, queueName string, sessionID string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -244,7 +244,7 @@ func (client *Client) AcceptSessionForQueue(ctx context.Context, queueName strin
 
 // AcceptSessionForSubscription accepts a session from a subscription with a specific session ID.
 // NOTE: this receiver is initialized immediately, not lazily.
-// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
+// If the operation fails it can return an *azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptSessionForSubscription(ctx context.Context, topicName string, subscriptionName string, sessionID string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -272,7 +272,7 @@ func (client *Client) AcceptSessionForSubscription(ctx context.Context, topicNam
 
 // AcceptNextSessionForQueue accepts the next available session from a queue.
 // NOTE: this receiver is initialized immediately, not lazily.
-// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
+// If the operation fails it can return an *azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptNextSessionForQueue(ctx context.Context, queueName string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
@@ -299,7 +299,7 @@ func (client *Client) AcceptNextSessionForQueue(ctx context.Context, queueName s
 
 // AcceptNextSessionForSubscription accepts the next available session from a subscription.
 // NOTE: this receiver is initialized immediately, not lazily.
-// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
+// If the operation fails it can return an *azservicebus.Error type if the failure is actionable.
 func (client *Client) AcceptNextSessionForSubscription(ctx context.Context, topicName string, subscriptionName string, options *SessionReceiverOptions) (*SessionReceiver, error) {
 	id, cleanupOnClose := client.getCleanupForCloseable()
 	sessionReceiver, err := newSessionReceiver(
--- a/sdk/messaging/azservicebus/error.go
+++ b/sdk/messaging/azservicebus/error.go
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azservicebus
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/exported"
+)
+
+// Code is an error code, usable by consuming code to work with
+// programatically.
+type Code = exported.Code
+
+const (
+	// CodeConnectionLost means our connection was lost and all retry attempts failed.
+	// This typicall reflects an extended outage or connection disruption and may
+	// require manual intervention.
+	CodeConnectionLost = exported.CodeConnectionLost
+
+	// CodeLockLost means that the lock token you have for a message has expired.
+	// This message will be available again after the lock period expires, or, potentially
+	// go to the dead letter queue if delivery attempts have been exceeded.
+	CodeLockLost = exported.CodeLockLost
+)
+
+// Error represents a Service Bus specific error.
+type Error = exported.Error
--- a/sdk/messaging/azservicebus/error.go
+++ b/sdk/messaging/azservicebus/error.go
@@ -24,4 +24,7 @@ const (
 )
 
 // Error represents a Service Bus specific error.
+// NOTE: the Code is considered part of the published API but the message that
+// comes back from Error(), as well as the underlying wrapped error, are NOT and
+// are subject to change.
 type Error = exported.Error
--- a/sdk/messaging/azservicebus/error.go
+++ b/sdk/messaging/azservicebus/error.go
@@ -13,7 +13,7 @@ type Code = exported.Code
 
 const (
 	// CodeConnectionLost means our connection was lost and all retry attempts failed.
-	// This typicall reflects an extended outage or connection disruption and may
+	// This typically reflects an extended outage or connection disruption and may
 	// require manual intervention.
 	CodeConnectionLost = exported.CodeConnectionLost
 
--- a/sdk/messaging/azservicebus/example_receiver_test.go
+++ b/sdk/messaging/azservicebus/example_receiver_test.go
@@ -93,7 +93,7 @@ func ExampleReceiver_ReceiveMessages() {
 		err = receiver.CompleteMessage(context.TODO(), message, nil)
 
 		if err != nil {
-			var sbErr azservicebus.Error
+			var sbErr *azservicebus.Error
 
 			if errors.As(err, &sbErr) && sbErr.Code == azservicebus.CodeLockLost {
 				// The message lock has expired. This isn't fatal for the client, but it does mean
--- a/sdk/messaging/azservicebus/example_receiver_test.go
+++ b/sdk/messaging/azservicebus/example_receiver_test.go
@@ -5,6 +5,7 @@ package azservicebus_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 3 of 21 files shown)


sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Unwrap is implemented so this error can be used with errors.As()
+func (e Error) Unwrap() error {
+	return e.innerErr
+}
+
+// NewError creates a new `Error` instance.
+// NOTE: this function is only exported so it can be used by the `internal`
+// package. It is not available for customers.
+func NewError(code Code, innerErr error) error {
```
chlowell, line 45: Do you need to return the interface?
```suggestion
func NewError(code Code, innerErr error) Error {
```

sdk/messaging/azservicebus/receiver_test.go
```diff
@@ -752,6 +755,38 @@ func TestReceiverMultiTopic(t *testing.T) {
...
+	require.NoError(t, err)
+
+	receiver, err := client.NewReceiverForQueue(queueName, nil)
+	require.NoError(t, err)
+
+	messages, err := receiver.ReceiveMessages(context.Background(), 1, nil)
+	require.NoError(t, err)
+
+	// sleep so our message locks expire
+	time.Sleep(6 * time.Second)
```
chlowell, line 777: Can you skip this in playback?

sdk/messaging/azservicebus/receiver_unit_test.go
```diff
@@ -394,24 +395,73 @@ func TestReceiverOptions(t *testing.T) {
...
-	}
+		entity: entity{
+			Queue: "queue",
+		},
+		cleanupOnClose:      func() {},
+		getRecoveryKindFunc: internal.GetRecoveryKind,
+		newLinkFn: func(ctx context.Context, session amqpwrap.AMQPSession) (internal.AMQPSenderCloser, internal.AMQPReceiverCloser, error) {
+			return nil, nil, nil
+		},
+		retryOptions: exported.RetryOptions{
```
chlowell, line 413:
```suggestion
		retryOptions: RetryOptions{
```

sdk/messaging/azservicebus/sender_unit_test.go
```diff
@@ -0,0 +1,59 @@
...
+func TestSender_UserFacingError(t *testing.T) {
+	fakeAMQPLinks := &internal.FakeAMQPLinks{}
+
+	sender, err := newSender(newSenderArgs{
+		ns: &internal.FakeNS{
+			AMQPLinks: fakeAMQPLinks,
+		},
+		queueOrTopic:   "queue",
+		cleanupOnClose: func() {},
+		retryOptions: exported.RetryOptions{
```
chlowell, line 26:
```suggestion
		retryOptions: RetryOptions{
```

sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,50 @@
...
+	innerErr error
+}
+
+// Error is an error message containing the code and a user friendly message, if any.
+func (e Error) Error() string {
+	return fmt.Sprintf("(%s): %s", e.Code, e.innerErr.Error())
+}
+
+// Unwrap is implemented so this error can be used with errors.As()
+func (e Error) Unwrap() error {
```
chlowell, line 38: When a type doesn't define `Unwrap`, `errors.As` just doesn't call it and the error chain ends with that type. So, you need this method only if you want to enable applications to access `e.innerErr` via `errors.As`. I guess you don't, because `e.innerErr` will be an unexported type anyway?

sdk/messaging/azservicebus/error.go
```diff
@@ -0,0 +1,27 @@
...
+
+package azservicebus
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/exported"
+)
+
+// Code is an error code, usable by consuming code to work with
+// programatically.
+type Code = exported.Code
```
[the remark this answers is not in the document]
richardpark-msft: This is the "export internal stuff" pattern, so this one does come from `exported`. This specifically is the spot where we officially export the code type from the azservicebus package.

sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -37,38 +38,45 @@ const (
...
 	if err == nil {
 		return nil
 	}
 
-	sbe := &SBErrInfo{
-		inner:        err,
-		RecoveryKind: GetRecoveryKind(err),
+	_, ok := err.(exported.Error)
+
+	if ok {
```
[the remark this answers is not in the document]
richardpark-msft: These bytes need to be saved!

sdk/messaging/azservicebus/internal/exported/error.go
richardpark-msft: I could, although in this particular case it really should conform to the `error` interface. So I think this does add a good type check and I don't think I ever try to use it as an `Error` after that.

sdk/messaging/azservicebus/receiver_test.go
richardpark-msft: I wish! This is actually AMQP underneath, so no recordings.

sdk/messaging/azservicebus/internal/exported/error.go
richardpark-msft: This is intentional as we're only just providing a basic wrapper. The real details are always in the underlying error, but the only part we're on the hook for, programmatically, is providing the code.

sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,50 @@
...
+type Error struct {
+	// Code is a stable error code which can be used as part of programatic error handling.
+	// The codes can expand in the future, but the values (and their meaning) will remain the same.
+	Code     Code
+	innerErr error
+}
+
+// Error is an error message containing the code and a user friendly message, if any.
+func (e Error) Error() string {
+	return fmt.Sprintf("(%s): %s", e.Code, e.innerErr.Error())
```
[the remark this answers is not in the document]
richardpark-msft: The usage of this is restricted to only a few spots, all of which only create this error because they're intentionally wrapping an error they classified. So in this case it's on the caller to do it correctly but they'd have to go out of their way to not do so.

Maybe there's a better way. I didn't want to end up with `NewError(err innerErr) (Error, error)` for instance, which would at least allow me to do the validation in the constructor but that's the only way I can see preventing this potential problem.

sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,50 @@
...
+	CodeConnectionLost Code = "connlost"
+
+	// CodeLockLost means that the lock token you have for a message has expired.
+	// This message will be available again after the lock period expires, or, potentially
+	// go to the dead letter queue if delivery attempts have been exceeded.
+	CodeLockLost Code = "locklost"
+)
+
+// Error represents a Service Bus specific error.
+type Error struct {
```
jhendrixMSFT, line 25: I think we need to decide if this is the path we want to take.  Per the [guidelines](https://azure.github.io/azure-sdk/golang_introduction.html#error-handling) we currently forbid the creation of custom error types; plus, we went to some lengths to update other SDKs to return an `*azcore.ResponseError`.  Having a unified error type across the SDKs simplifies the experience.

CC @JeffreyRichter

richardpark-msft: We had this discussion, actually:
https://github.com/Azure/azure-sdk-for-go/issues/17283#issuecomment-1071013052

sdk/messaging/azservicebus/internal/exported/error.go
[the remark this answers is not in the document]
richardpark-msft: For now I'll just do a check with nil and print out unknown error (just so it's easy to search and find should we get bug reports)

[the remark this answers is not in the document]
richardpark-msft: Fixed! Also added a couple of tests, let me know if it's what you were thinking of.

sdk/messaging/azservicebus/internal/exported/error.go
jhendrixMSFT: I didn't see explicitly where it was said that `azcore.ResponseError` won't work, but if you don't think it will then that's fine.

Is there any commonality with event hubs such that they can share an error type, or is it different enough that we'll end up with `azeventhubs.Error`?

richardpark-msft: After discussing we're going to leave this as-is but we want to add in doc comments to indicate where this error type comes back.

richardpark-msft: (doc comments added)

sdk/messaging/azservicebus/client.go
```diff
@@ -217,6 +217,7 @@ func (client *Client) NewSender(queueOrTopic string, options *NewSenderOptions)
 
 // AcceptSessionForQueue accepts a session from a queue with a specific session ID.
 // NOTE: this receiver is initialized immediately, not lazily.
+// If the operation fails it can return an azservicebus.Error type if the failure is actionable.
```
jhendrixMSFT, line 220: It's an `*azservicebus.Error` right?

sdk/messaging/azservicebus/example_receiver_test.go
```diff
@@ -80,6 +93,18 @@ func ExampleReceiver_ReceiveMessages() {
 		err = receiver.CompleteMessage(context.TODO(), message, nil)
 
 		if err != nil {
+			var sbErr azservicebus.Error
```
jhendrixMSFT, line 96: Looks like the error isn't being returned by reference.

sdk/messaging/azservicebus/internal/exported/error.go
jhendrixMSFT: Implementing `Unwrap()` means that the wrapped error implicitly becomes part of the error contract, as `errors.As()` and friends will have access to it.  Is this really necessary?  We've not been implementing `Unwrap()` in other error types.

sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,57 @@
...
+// Unwrap is implemented so this error can be used with errors.As()
+func (e Error) Unwrap() error {
+	return e.innerErr
+}
+
+// NewError creates a new `Error` instance.
+// NOTE: this function is only exported so it can be used by the `internal`
+// package. It is not available for customers.
+func NewError(code Code, innerErr error) error {
+	return Error{
```
jhendrixMSFT, line 53: Should return by reference.

jhendrixMSFT (changes requested):

sdk/messaging/azservicebus/receiver_test.go
```diff
@@ -554,8 +555,10 @@ func TestReceiver_RenewMessageLock(t *testing.T) {
 	endCaptureFn := test.CaptureLogsForTest()
 	defer endCaptureFn()
 	expectedLockBadError := receiver.RenewMessageLock(context.Background(), messages[0], nil)
-	// String matching can go away once we fix #15644
-	// For now it at least provides the user with good context that something is incorrect about their lock token.
+
+	var asSBError Error
+	require.ErrorAs(t, expectedLockBadError, &asSBError)
```
jhendrixMSFT, line 560: Presumably this succeeds at present due to the `Unwrap()` implementation on `Error`?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -196,7 +197,8 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, maxMessages int, options
 		return nil, errors.New("receiver is already receiving messages. ReceiveMessages() cannot be called concurrently")
 	}
 
-	return r.receiveMessagesImpl(ctx, maxMessages, options)
+	messages, err := r.receiveMessagesImpl(ctx, maxMessages, options)
+	return messages, internal.TransformError(err)
```
jhendrixMSFT, line 201: Is it possible to move this error transformation further down the stack so that implementers of new APIs don't need to remember to do this?

jhendrixMSFT: If you want to save that refactoring for post-GA that's fine, just open an issue to track it.

sdk/messaging/azservicebus/client.go
richardpark-msft: It will be...

sdk/messaging/azservicebus/receiver.go
richardpark-msft: This approach was a bit annoying but it has the benefit of letting me shuffle the internals around more without having to worry about which piece of code was responsible for transforming it.

It's not a huge deal, but this was a bit easier to manually review.

sdk/messaging/azservicebus/internal/exported/error.go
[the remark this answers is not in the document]
richardpark-msft: Unwrap() has now been removed.

sdk/messaging/azservicebus/internal/exported/error.go
richardpark-msft: Unwrap() has now been removed.

sdk/messaging/azservicebus/client.go
richardpark-msft: It's not returned by pointer.

sdk/messaging/azservicebus/example_receiver_test.go
richardpark-msft: It is now and examples have been updated.

sdk/messaging/azservicebus/internal/exported/error.go
richardpark-msft: Done.

sdk/messaging/azservicebus/internal/exported/error.go
```diff
@@ -0,0 +1,52 @@
...
+// are subject to change.
+type Error struct {
+	// Code is a stable error code which can be used as part of programatic error handling.
+	// The codes can expand in the future, but the values (and their meaning) will remain the same.
+	Code     Code
+	innerErr error
+}
+
+// Error is an error message containing the code and a user friendly message, if any.
+func (e Error) Error() string {
```
jhendrixMSFT, line 36: `(e *Error)` I'm surprised the linter didn't flag this.

richardpark-msft: I am too, actually, bit it's still valid I suppose. I switched it over and added in a couple of related tests.

jhendrixMSFT (approved):

Issue #17743, opened 2022-04-26
Provide abstraction for SDK authors to create custom polling trackers


jhendrixMSFT:
The current `runtime.Poller[T]` uses internal "polling trackers" to determine the behavior (Azure-AsyncOperation, Location, etc).  Some services might require custom polling trackers, e.g. the custom model trainer API for Form Recognizer uses a custom implementation (see python example [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_polling.py)).  While there's no need for this at present, it will become a requirement if/when we produce such SDKs.

I believe this can be done in a non-breaking way by adding a `PollingTracker` field (we can refine the name) to the `NewPollerOptions` type that, when specified, will be used in lieu of our polling tracker detection logic.

jhendrixMSFT:
This prevents us from using the generic poller in `azcore` for things like key vault.

RickWinter:
@jhendrixMSFT to post the PR

jhendrixMSFT:
Fixed in upcoming `azcore@v0.24.0`

PR #17793, opened 2022-05-03
Add support for custom poller implementations

jhendrixMSFT:


jhendrixMSFT:
Fixes:
- https://github.com/Azure/azure-sdk-for-go/issues/17415
- https://github.com/Azure/azure-sdk-for-go/issues/17743

sdk/azcore/runtime/poller.go
```diff
@@ -104,67 +103,82 @@ type NewPollerFromResumeTokenOptions[T any] struct {
...
 	return &Poller[T]{
-		pt: pollers.NewPoller(lro, nil, pl),
-		rt: options.Response,
+		op:     opr,
+		result: options.Response,
 	}, nil
 }
 
+// Operation[T] abstracts the differences among long-running operation implementations.
+type Operation[T any] interface {
```
jhendrixMSFT, line 152: This is the new piece of public surface area.  Open to other naming suggestions.

sdk/azcore/runtime/poller.go
```diff
@@ -174,12 +188,49 @@ func (p *Poller[T]) PollUntilDone(ctx context.Context, options *PollUntilDoneOpt
 	if cp.Frequency == 0 {
 		cp.Frequency = 30 * time.Second
 	}
-	var resp T
-	if p.rt != nil {
-		resp = *p.rt
+
+	if cp.Frequency < time.Second {
```
jhendrixMSFT, line 192: This was copied from the internal implementation.

jhendrixMSFT:
Testing includes autorest test server, `armcompute`, `armredis`, and updating `azcertificates` with a functional, custom, poller.

sdk/azcore/runtime/poller.go
RickWinter: Operation is used in C++ and .NET.  I like aligning on that name here as well.

jhendrixMSFT: That was mostly why I chose the name.  The only thing here is that we use the term `Poller` instead of `Operation` for our LROs.  It would be nice if the term reflected that, but I didn't want to use something like `PollerImpl`.

chlowell: I'm fine with `Operation` but if you're looking for another name, `PollingMethod` seems reasonable too

sdk/azcore/CHANGELOG.md
```diff
@@ -3,20 +3,30 @@
...
+* The `Operation-Location` poller now properly handles `final-state-via` values.
 
 ### Other Changes
-* The functionality in `arm/runtime/poller.go` has been merged into `runtime/poller.go` so it should be used instead.
-  * `arm/runtime/poller.go` will be removed in a future release.
+* The internal poller implementation has been refactored.
+  * The implementation in `internal/pollers/poller.go` has been merged into `runtime/poller.go` with some slight modification.
+  * Constructor `arm/runtime.NewPoller[T]()` is now implemented in terms of `runtime.NewPoller[T]()`, leveraging the custom poller feature.
+  * The internal poller types had their methods updated to conform to the `runtime.Operation` interface.
+  * The creation of resume tokens has been refactored so that implementers of `runtime.Operation` don't need to know about it.
```
chlowell, line 28: 🥳

chlowell (approved):

sdk/azcore/runtime/poller.go
jhendrixMSFT: `PollingPolicy`
`PollerMethod`

chlowell: Eh, that suggests it's a `policy.Policy`

jhendrixMSFT: Yeah I had the same thought.

How about `PollerMethod` then.  It's like `Poller` but with a suffix.

Issue #17771, opened 2022-04-28
Cosmos DB - TransactionalBatch support


ealsur:
Add support for https://docs.microsoft.com/en-us/azure/cosmos-db/sql/transactional-batch

## Proposed API

The exposed new types are:

* TransactionalBatch - Represents the batch itself
* TransactionalBatchOptions - Options for execution of the batch
* TransactionalBatchItemOptions - Options at the individual operation level
* TransactionalBatchResponse - The batch response
* TransactionalBatchResponseOperationResult - The individual operation result within the batch response

### Creating a Transactional Batch

```
func (c *ContainerClient) NewTransactionalBatch(partitionKey PartitionKey) TransactionalBatch
```

On the Container, we can create the TransactionalBatch like so:

```
partitionKey := NewPartitionKeyString("myPartitionKeyValue")
batch := container.NewTransactionalBatch(partitionKey)
```

The batch is scoped to a Partition Key Value.

### Populating operations

A batch contains a list of operations to be executed as a single unit:

```
// CreateItem adds a create operation to the batch.
func (b *TransactionalBatch) CreateItem(item []byte, o *TransactionalBatchItemOptions)

// DeleteItem adds a delete operation to the batch.
func (b *TransactionalBatch) DeleteItem(itemId string, o *TransactionalBatchItemOptions)

// ReplaceItem adds a replace operation to the batch.
func (b *TransactionalBatch) ReplaceItem(itemId string, item []byte, o *TransactionalBatchItemOptions)

// UpsertItem adds an upsert operation to the batch.
func (b *TransactionalBatch) UpsertItem(item []byte, o *TransactionalBatchItemOptions)

// ReadItem adds a read operation to the batch.
func (b *TransactionalBatch) ReadItem(itemId string, o *TransactionalBatchItemOptions) 
```

Each operation that can be added contains some potential options:

```
type TransactionalBatchItemOptions struct {
	// IfMatchEtag is used to ensure optimistic concurrency control.
	// https://docs.microsoft.com/azure/cosmos-db/sql/database-transactions-optimistic-concurrency#optimistic-concurrency-control
	IfMatchEtag *azcore.ETag
}
```

### Executing a batch

After a batch is created, operations need to be added, and then the ExecuteTransactionalBatch method is used:

```
func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b TransactionalBatch, o *TransactionalBatchOptions) (TransactionalBatchResponse, error)
```

[...]

PR #17795, opened 2022-05-03
Cosmos DB - Adds TransactionalBatch support

ealsur:
## APIView

https://apiview.dev/Assemblies/Review/7e272b28ee30452aaa7e3e77ad8cd4b9/23b1fc488ec84e9e855cce2bb2ab94d8?diffRevisionId=6b6ad7e3a9b74575915fb585bfc99929&doc=False&diffOnly=True

## Description

The exposed new types are:

* TransactionalBatch - Represents the batch itself
* TransactionalBatchOptions - Options for execution of the batch
* TransactionalBatchItemOptions - Options at the individual operation level
* TransactionalBatchResponse - The batch response
* TransactionalBatchResult - The individual operation result within the batch response

### Creating a Transactional Batch

```
func (c *ContainerClient) NewTransactionalBatch(partitionKey PartitionKey) TransactionalBatch
```

On the Container, we can create the TransactionalBatch like so:

```
partitionKey := NewPartitionKeyString("myPartitionKeyValue")
batch := container.NewTransactionalBatch(partitionKey)
```

The batch is scoped to a Partition Key Value.

### Populating operations

A batch contains a list of operations to be executed as a single unit:

```
// CreateItem adds a create operation to the batch.
func (b *TransactionalBatch) CreateItem(item []byte, o *TransactionalBatchItemOptions)

// DeleteItem adds a delete operation to the batch.
func (b *TransactionalBatch) DeleteItem(itemID string, o *TransactionalBatchItemOptions)

// ReplaceItem adds a replace operation to the batch.
func (b *TransactionalBatch) ReplaceItem(itemID string, item []byte, o *TransactionalBatchItemOptions)

// UpsertItem adds an upsert operation to the batch.
func (b *TransactionalBatch) UpsertItem(item []byte, o *TransactionalBatchItemOptions)

// ReadItem adds a read operation to the batch.
func (b *TransactionalBatch) ReadItem(itemID string, o *TransactionalBatchItemOptions) 
```

Each operation that can be added contains some potential options:

```
type TransactionalBatchItemOptions struct {
	// IfMatchETag is used to ensure optimistic concurrency control.
	// https://docs.microsoft.com/azure/cosmos-db/sql/database-transactions-optimistic-concurrency#optimistic-concurrency-control
	IfMatchETag *azcore.ETag
}
```

### Executing a batch

After a batch is created, operations need to be added, and then the ExecuteTransactionalBatch method is used:

```
func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b TransactionalBatch, o *TransactionalBatchOptions) (TransactionalBatchResponse, error)
```

Example:

```
partitionKey := NewPartitionKeyString("myPartitionKeyValue")
batch := container.NewTransactionalBatch(partitionKey)
batch.CreateItem(bodyOfAnItemToCreate, nil)
batch.ReadItem("idOfAnItemToRead", nil)
//... other operations

batchResponse, err := container.ExecuteTransactionalBatch(context, batch, nil)
```

The ExecuteTransactionalBatch method contains options for the execution:

```
// TransactionalBatchOptions includes options for transactional batch operations.
type TransactionalBatchOptions struct {
	SessionToken string
	ConsistencyLevel *ConsistencyLevel
}
```

### Handling batch response

The batch response contains the raw HTTP response plus Request Charge, SessionToken, ActivityId and 2 important properties:

```
// QueryItemsResponse contains response from the query operation.
type TransactionalBatchResponse struct {
	Response
	// SessionToken contains the value from the session token header to be used on session consistency.
	SessionToken string
	// OperationResults contains the individual batch operation results.
	// The order of the results is the same as the order of the operations in the batch.
	OperationResults []TransactionalBatchResponseOperationResult
	// Success indicates if the transaction was successfully committed.
	// If false, one of the operations in the batch failed.
	// Inspect the OperationResults, any operation with status code http.StatusFailedDependency is a dependency failure.
	// The cause of the batch failure is the first operation with status code different from http.StatusFailedDependency.
	Success bool
}
```

`OperationResults` contain all the individual operation results, each operation in a batch has its own response.

Each operation result exposes:

```
// TransactionalBatchResponseOperationResult represents the result of a single operation in a batch.
type TransactionalBatchResult struct {
	// StatusCode contains the status code of the operation.
	StatusCode int32
	// RequestCharge contains the request charge for the operation.
	RequestCharge float32
	// ResourceBody contains the body response of the operation.
	// This property is available depending on the EnableContentResponseOnWrite option.
	ResourceBody []byte
	// ETag contains the ETag of the operation.
	ETag azcore.ETag
}
```

`Success` is used to determine if the batch committed all operations or if it had to be rolled back due to one operation failing.

After receiving the response, we need to inspect if the batch committed, and if not, which operation was the source of the failure:

```
if batchResponse.Success {
		// Transaction succeeded
		// We can inspect the individual operation results 
		for index, operation := range batchResponse.OperationResults {
			fmt.Printf("Operation %v completed with status code %v consumed %v RU", index, operation.StatusCode, operation.RequestCharge)
			if index == 1 {
				// Read operation would have body available
				var itemResponseBody map[string]string
				err = json.Unmarshal(operation.ResourceBody, &itemResponseBody)
				if err != nil {
					panic(err)
				}
			}
		}
	} else {
		// Transaction failed, look for the offending operation
		for index, operation := range batchResponse.OperationResults {
			if operation.StatusCode != http.StatusFailedDependency {
				fmt.Printf("Transaction failed due to operation %v which failed with status code %v", index, operation.StatusCode)
			}
		}
	}
```

Closes https://github.com/Azure/azure-sdk-for-go/issues/17771


sdk/data/azcosmos/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # Release History
 
-## 0.3.1 (Unreleased)
+## 0.4.0 (Unreleased)
```
jhendrixMSFT, line 3: The convention we've been using for prerelease is to bump minor for breaking changes, and bump patch for features/bug fixes.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -238,6 +238,26 @@ func (c *Client) sendDeleteRequest(
 	return c.executeAndEnsureSuccessResponse(req)
 }
 
+func (c *Client) sendBatchRequest(
+	path string,
+	ctx context.Context,
```
jhendrixMSFT, line 243: For funcs/methods that take a `context.Context`, by convention, it should always be the first parameter.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -456,6 +457,66 @@ func (c *ContainerClient) NewQueryItemsPager(query string, partitionKey Partitio
...
+// NewTransactionalBatch creates a batch of operations to be committed as a single unit.
+// See https://docs.microsoft.com/azure/cosmos-db/sql/transactional-batch
+func (c *ContainerClient) NewTransactionalBatch(partitionKey PartitionKey) TransactionalBatch {
+	return TransactionalBatch{partitionKey: partitionKey}
+}
+
+// ExecuteTransactionalBatch executes a transactional batch.
+// Once executed, verify the Success property of the response to determine if the batch was committed
+func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b TransactionalBatch, o *TransactionalBatchOptions) (TransactionalBatchResponse, error) {
+	if len(b.operations) == 0 {
```
jhendrixMSFT, line 469: The guidelines state that we should omit client-side validation of parameters and let the service reject the request.  But before we change it, I have some questions.

- Does the service return an intelligible error in this case?
- Would the caller be charged $$ for such a call?

sdk/data/azcosmos/cosmos_transactional_batch_response.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func newTransactionalBatchResponse(resp *http.Response) (TransactionalBatchResponse, error) {
+	response := TransactionalBatchResponse{
+		Response: newResponse(resp),
+	}
+
+	response.SessionToken = resp.Header.Get(cosmosHeaderSessionToken)
+
+	response.Success = resp.StatusCode != http.StatusMultiStatus
```
jhendrixMSFT, line 36: Are 4xx and 5xx status codes handled elsewhere?

sdk/data/azcosmos/cosmos_transactional_batch_response.go
```diff
@@ -0,0 +1,89 @@
...
+		return TransactionalBatchResponse{}, err
+	}
+
+	return response, nil
+}
+
+// TransactionalBatchResult represents the result of a single operation in a batch.
+type TransactionalBatchResult struct {
+	// StatusCode contains the status code of the operation.
+	StatusCode int32
```
jhendrixMSFT, line 48: You can add tags to each field to handle JSON marshalling/unmarshalling which should allow you to remove your custom unmarshaller.  Here's an example.
```go
StatusCode int32 `json:"statusCode"`
```
See the [docs](https://pkg.go.dev/encoding/json#Marshal) for more info.

sdk/data/azcosmos/cosmos_transactional_batch_response.go
ealsur: Yes, those would be returned as errors, not a response. MultiStatus is a 207 (that's the status code the service picked) so it's not a failure, but a valid response, which the user needs to read and understand.

sdk/data/azcosmos/cosmos_transactional_batch_response.go
ealsur: I tried the tags. The reason I need the Unmarshall is that the body of a particular item we are returning it as a `[]byte` to be inline with all the other item responses (ItemResponse). And that needs to be obtained from the server response for that particular item. The server returns a JSON array, each element contains attributes (like the statusCode) and a JSON object as the body, I need to convert the JSON body into a []byte. The `json` decorator does not work on that scenario.

sdk/data/azcosmos/cosmos_container.go
ealsur: The service returns a BadRequest but not exactly the reason for the BadRequest, so I preferred to be more specific for this case.

sdk/data/azcosmos/cosmos_transactional_batch_response.go
ealsur: The contract that was defined service side is, for scenarios where one of the operations generates a rollback, it returns a 207 (instead of 200) and the individual operation result should be inspected (same as it needs to be done on other language SDKs). If the service returns a 4xx or 5xx, then that is a failure and the call returns the `err` to the caller.

sdk/data/azcosmos/cosmos_container.go
jhendrixMSFT: I see, we can leave it for now then.

Long-term, can this get fixed on the service side?  Then the error is uniform across all SDKs as well as customers that directly use the HTTP endpoints.

ealsur: That would be ideal, I agree. And as per previous PRs, I try to avoid any client side validation, but this case I feel customers would be confused.

jhendrixMSFT (approved):

Issue #17752, opened 2022-04-27
Export ExpiringResource from /sdk/internal so SDKs can implement their own authentication policy


jhendrixMSFT:
### Feature Request

The bearer token policy uses a hard-coded format string for populating the `Authentication` header value.  This means that the policy can't be used for other services that use an OAuth token for authentication with a slightly different value (e.g. Cosmos).

We can add an option to the `BearerTokenPolicyOptions` type (which is current empty) that allows specifying an optional format string.

```go
// BearerTokenOptions configures the bearer token policy's behavior.
type BearerTokenOptions struct {
	// Format is the format string used when constructing the value for the Authentication header.
	// The empty string defaults to format "Bearer %s", where %s is replaced with the OAuth token.
	Format string
}
```

ealsur:
For reference, in Cosmos we have a different Auth header format for AAD: https://github.com/Azure/azure-sdk-for-go/blob/4b68333d02a427eb47985a8c61b828fdc16b9dd4/sdk/data/azcosmos/cosmos_policy_bearer_token.go#L26

In the current state, the only alternative is to have a policy read the current Bearer Auth header and replace it with our format.

chlowell:
Given that this wouldn't be generally useful, other languages haven't added a similar option, and that we have only one use case for it here, I think we should wait to see more need before adding it. Today's workarounds don't seem too bad to me. Here's another one:
```go
type cosmosCredentialWrapper struct {
	cred azcore.TokenCredential
}

func (c *cosmosCredentialWrapper) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (*azcore.AccessToken, error) {
	tk, err := c.cred.GetToken(ctx, opts)
	if err != nil {
		return nil, err
	}
	// this is okay because our auth policies handle tokens as opaque strings
	tk.Token = "type=aad&ver=1.0&sig=" + tk.Token
	return tk, err
}
```

jhendrixMSFT:
This will preserve the `Bearer ` prefix though which is problematic for Cosmos.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_bearer_token.go#L61

jhendrixMSFT:
Good point about the RFC.  I think then the best option is to expose the necessary primitives like `ExpiringResource` so that SDK authors can implement their own policy.

jhendrixMSFT:
Fixed in upcoming `internal@v0.10.0` release.

PR #17798, opened 2022-05-03
Added temporal package for handling of temporal resources

jhendrixMSFT:

### DIFF
--- a/sdk/internal/CHANGELOG.md
+++ b/sdk/internal/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.10.0 (Unreleased)
 
 ### Features Added
+* Added `temporal` package for handling of temporal resources.
 
 ### Breaking Changes
 * Removed logging constants from the `log` package.
@@ -10,6 +11,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+* Updated build constraints and Go versioin to 1.18.
 
 ## 0.9.2 (2022-04-04)
 
--- a/sdk/internal/CHANGELOG.md
+++ b/sdk/internal/CHANGELOG.md
@@ -11,7 +11,7 @@
 ### Bugs Fixed
 
 ### Other Changes
-* Updated build constraints and Go versioin to 1.18.
+* Updated build constraints and Go version to 1.18.
 
 ## 0.9.2 (2022-04-04)
 
--- a/sdk/internal/temporal/resource.go
+++ b/sdk/internal/temporal/resource.go
@@ -1,0 +1,120 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package temporal
+
+import (
+	"sync"
+	"time"
+)
+
+// AcquireResource abstracts a method for refreshing a temporal resource.
+type AcquireResource[TResource, TState any] func(state TState) (newResource TResource, newExpiration time.Time, err error)
+
+// Resource is a temporal resource (usually a credential) that requires periodic refreshing.
+type Resource[TResource, TState any] struct {
+	// cond is used to synchronize access to the shared resource embodied by the remaining fields
+	cond *sync.Cond
+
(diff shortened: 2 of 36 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17752

sdk/internal/temporal/resource.go
```diff
@@ -0,0 +1,120 @@
...
+
+package temporal
+
+import (
+	"sync"
+	"time"
+)
+
+// AcquireResource abstracts a method for refreshing a temporal resource.
+type AcquireResource[TResource, TState any] func(state TState) (newResource TResource, newExpiration time.Time, err error)
```
jhendrixMSFT, line 15: The contents were copied from `azcore` with a few of the type names slightly changed.

sdk/internal/temporal/resource.go
```diff
@@ -0,0 +1,120 @@
...
+
+		// Wake up any waiting threads/goroutines since there is a resource they can ALL use
+		er.cond.L.Unlock()
+		er.cond.Broadcast()
+	}
+	return resource, err // Return the resource this thread/goroutine can use
+}
+
+// Expire marks the resource as expired, ensuring it's refreshed on the next call to GetResource().
+func (er *Resource[TResource, TState]) Expire() {
```
jhendrixMSFT, line 114: This is a new method.  It's used by the key vault challenge policy.

sdk/internal/CHANGELOG.md
```diff
@@ -3,13 +3,15 @@
...
 ### Features Added
+* Added `temporal` package for handling of temporal resources.
 
 ### Breaking Changes
 * Removed logging constants from the `log` package.
 
 ### Bugs Fixed
 
 ### Other Changes
+* Updated build constraints and Go versioin to 1.18.
```
chlowell, line 14:
```suggestion
* Updated build constraints and Go version to 1.18.
```

RickWinter (approved):

sdk/internal/temporal/resource.go
chlowell: Why does Key Vault's policy need to force refresh a token?

sdk/internal/CHANGELOG.md
jhendrixMSFT: Fixed

sdk/internal/temporal/resource.go
jhendrixMSFT: Good question.  Honestly, I'm not sure as I wasn't involved in this implementation.

Per the [comments](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/keyvault/internal/challenge_policy.go#L88-L91).

```go
// If it fails and has a 401, try it with a new token
```

@schaabs do you have any context on why we might want to explicitly expire a token credential in this scenario?

chlowell: The policy discards its cached token when it sees a 401 because the error means the token is invalid despite not being expired. What I'm getting at is whether the policy needs `Expire()` to do this. I would think it can just create a new `ExpiringResource`, which makes sense conceptually because the policy may need to acquire a token in a new tenant, or with new claims.

jhendrixMSFT: I see your point.  Presumably using `Expire()` would be a bit more GC-friendly in this case compared to creating a new `ExpiringResource`.  However, it's probably infrequent enough that it doesn't really matter (well, I'm guessing here but I suspect this is why it was implemented in this way).

chlowell: It should be a rare event and even if not, I expect the perf impact due to collection is trivial compared to the impact of reauthenticating and resending the service request.

jhendrixMSFT: At this point I'm inclined to leave it given that key vault has tested with this implementation.  It's also not available to the public.  Do you feel particularly strong about removing it?

chlowell: I think it's okay for now, being internal, but Key Vault's policy needs another look after #14932

chlowell (approved):

PR #17804, opened 2022-05-04
Warn on unsupported managed identity configuration

### LINKED ISSUE #17754: Align user-assigned identity behavior with other SDKs
When the application specifies a user-assigned identity in an environment known not to support user-assigned identities, send the identity's ID as a query parameter anyway and log a warning.

chlowell:
Closes #17754

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -293,7 +293,7 @@ func (c *managedIdentityClient) createServiceFabricAuthRequest(ctx context.Conte
 	q.Add("api-version", serviceFabricAPIVersion)
 	q.Add("resource", strings.Join(scopes, " "))
 	if id != nil {
-		log.Write(EventAuthentication, "Service Fabric doesn't support selecting a user-assigned identity at runtime")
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -293,7 +293,12 @@ func (c *managedIdentityClient) createServiceFabricAuthRequest(ctx context.Conte
 	q.Add("api-version", serviceFabricAPIVersion)
 	q.Add("resource", strings.Join(scopes, " "))
 	if id != nil {
-		q.Add(qpClientID, id.String())
+		log.Write(EventAuthentication, "Service Fabric doesn't support selecting a user-assigned identity at runtime")
```
jhendrixMSFT, line 296: Nit: This log message is a bit different than the others, is that intentional?

jhendrixMSFT: Also, should these messages have some kind of prefix like `WARNING: ` so they stand out a bit more?

chlowell: Service Fabric does support user-assigned identities, but you configure one in the application manifest and can't select a different one at runtime as you can on App Service, for example. Arc and Cloud Shell don't support user-assigned identity at all.

jhendrixMSFT (approved):

PR #17830, opened 2022-05-06
Do not include Authorization header if access token is empty

ItalyPaleAle:
If the token provider returns an empty token, then do not include the Authorization header.

Why -> This is helpful when working against dev environments that do not use headers for authorization. This way, we can simply provide a "no-op" token provider like:

```go
type NoOpTokenProvider struct{}

func (NoOpTokenProvider) GetToken(ctx context.Context, options policy.TokenRequestOptions) (*azcore.AccessToken, error) {
	return &azcore.AccessToken{
		Token:     "",
		ExpiresOn: time.Now().Add(10000 * time.Hour),
	}, nil
}
```

- [ X The PR does not update generated files.

### DIFF
--- a/sdk/azcore/arm/runtime/policy_bearer_token.go
+++ b/sdk/azcore/arm/runtime/policy_bearer_token.go
@@ -68,6 +68,9 @@ func (b *BearerTokenPolicy) Do(req *azpolicy.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	if tk.Token == "" {
+		return req.Next()
+	}
 	req.Raw().Header.Set(shared.HeaderAuthorization, shared.BearerTokenPrefix+tk.Token)
 	auxTokens := []string{}
 	for tenant, er := range b.auxResources {


sdk/azcore/arm/runtime/policy_bearer_token.go
```diff
@@ -68,6 +68,9 @@ func (b *BearerTokenPolicy) Do(req *azpolicy.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	if tk.Token == "" {
```
jhendrixMSFT, line 71: Could you please add a comment explaining when an empty token might happen?

chlowell:
Can you give some more details about the scenario here? If the token provider can't acquire a token, it should return an error and the policy then returns that error instead of sending the request. Have you observed one of our credentials in `azidentity` returning an empty token and nil error?

jhendrixMSFT:
@chlowell comment got me thinking.  While we don't explicitly state this (maybe we should), the contract for `TokenCredential` is to return a token or an error.  IMO an empty token doesn't fulfil that contract.

chlowell:
I agree with that. I think it's reasonable and idiomatic for `BearerTokenPolicy` to interpret a non-nil error from `GetToken()` as meaning the token authenticates the principal. `BearerTokenPolicy` has no opinion about token content because it's responsible only for authorizing requests. The service validates the tokens. If a service requires requests _not_ to have an Authorization header, its clients shouldn't have `BearerTokenPolicy` in their pipelines.

ItalyPaleAle:
Thanks for the comments :) I agree this is a very niche situation and I achieve the same result by implementing my own pipeline, but thought I'd share this since it's a very small fix and it could help. I'd rather avoid creating my own pipeline so I can rely on the existing SDK modules as much as possible.

I am working on something that needs to communicate with a dev cluster of Azure Container Apps. In this case, the Resource Manager API calls need to go a different endpoint (which ends in `.azurewebsites.net`) and authentication is performed through a mTLS certificate (which I pass by using a custom `http.Client` as `Transporter`). Including an Authorization header in this case is not necessary and adds the additional complexity on my side as I need to retrieve an access token.

jhendrixMSFT:
@ItalyPaleAle if your `NoOpTokenProvider` returns a nonsensical value for the token (i.e. any string value), is the request rejected?  It's not super-clean, but is maybe a viable workaround.

Another idea would be to inject your own policy that removes the `Authorization` header from the request before sending it.

ItalyPaleAle:
@jhendrixMSFT I am having other issues right now and not sure if an invalid token would work or not. I will tell you later in the day.

As per using a policy to remove the header, I tried it, but it seems that it got fired before the policy that injected the header, so it didn't work.

I did get this to work (removing the header) by creating a custom Transporter however. But I'd rather get out of that business :)

jhendrixMSFT:
It appears that the caller-provided policies are added before the SDK policies which is why that didn't work.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/pipeline.go#L56-L57

I went through the history, and I don't think this was a conscious decision, it just happened to land that way.

@chlowell I'm thinking that we should reverse the order here to enable a caller policies to "interact" with the in-box ones.  Thoughts?

chlowell:
Probably it doesn't much matter, this being a rarely used feature, but after hearing the architecture board's long discussion on the perils of policy ordering, I'm wary of making that part of our contract more fine-grained. Today we promise that custom policies run either once per call or once per retry. Further promising that custom policies run after our default policies in each category seems to support applications in depending on what I consider implementation details of the default pipeline i.e., its constituents and their ordering. Those are my perhaps overly anxious thoughts 😄. Seems like something to discuss with the architect(s) before implementing though.

ItalyPaleAle:
…or you could perhaps just merge this PR and then we wouldn't have to deal with that 😊 If you don't like this approach, however I can continue using a custom Transporter and I'll close this PR.

jhendrixMSFT:
https://github.com/Azure/azure-sdk-for-go/pull/17857 changes the ordering of user-provided policies to execute after the in-box ones.  So you should be able to write your own policy to remove the `Authorization` header now.

chlowell:
Regarding the proposed change, I don't think we should add this quirk to the policy. The policy implements bearer token authorization for clients that need it. The Authorization header isn't optional for these clients, so the policy should never send a request without it.

PR #17839, opened 2022-05-06
Update CODEOWNERS

RickWinter:


.github/CODEOWNERS
```diff
@@ -6,7 +6,7 @@
 ################
 
 # Git Hub integration and bot rules
-/.github/                @AlexGhiondea @jsquire @jhendrixMSFT @rickwinter
+/.github/                    @jsquire @jhendrixMSFT @rickwinter
```
jhendrixMSFT, line 9: Is this errant whitespace or does it now line up with other things?

RickWinter: Lines up with the rest of the file

PR #17841, opened 2022-05-06
Update skipped azkeys tests

chlowell:
Fixing some test setup issues and unskipping and re-recording rotation policy and secure key release tests.

### DIFF
--- a/sdk/keyvault/azkeys/utils_test.go
+++ b/sdk/keyvault/azkeys/utils_test.go
@@ -107,19 +107,16 @@ func startTest(t *testing.T) func() {
 	}
 }
 
+// skipHSM skips live MHSM tests when AZURE_MANAGEDHSM_URL has no value
 func skipHSM(t *testing.T, testType string) {
-	if testType == HSMTEST && !enableHSM {
-		if recording.GetRecordMode() != recording.PlaybackMode {
-			t.Log("Skipping HSM Test")
-			t.Skip()
-		}
+	if recording.GetRecordMode() != recording.PlaybackMode && testType == HSMTEST && !enableHSM {
+		t.Skip("set AZURE_MANAGEDHSM_URL to run this test")
 	}
 }
 
 func alwaysSkipHSM(t *testing.T, testType string) {
 	if testType == HSMTEST {
-		t.Log("Skipping HSM Test")
-		t.Skip()
+		t.Skip("this feature isn't supported by MHSM")
 	}
 }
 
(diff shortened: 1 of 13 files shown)


sdk/keyvault/azkeys/utils_test.go
```diff
@@ -22,15 +23,35 @@ import (
...
+)
 
 func TestMain(m *testing.M) {
-	// Initialize
+	liveVaultURL = strings.TrimSuffix(os.Getenv("AZURE_KEYVAULT_URL"), "/")
+	liveMHSMURL = strings.TrimSuffix(os.Getenv("AZURE_MANAGEDHSM_URL"), "/")
+	enableHSM = liveMHSMURL != ""
+
+	if recording.GetRecordMode() != recording.LiveMode {
+		err := recording.ResetProxy(nil)
```
jhendrixMSFT, line 44: Why do we need to do this before/after the tests?  Is this just one of those "ensure the proxy is in a known state" type of thing?

sdk/keyvault/azkeys/utils_test.go
```diff
@@ -40,48 +61,37 @@ func TestMain(m *testing.M) {
...
-
-		err = recording.AddBodyKeySanitizer("$.recoveryId", fakeKvURL, vaultUrl, nil)
-		if err != nil {
-			panic(err)
+		if liveVaultURL == "" {
+			panic("no value for AZURE_KEYVAULT_URL")
 		}
-
-		tenantID := os.Getenv("AZKEYS_TENANT_ID")
-		err = recording.AddHeaderRegexSanitizer("WWW-Authenticate", "00000000-0000-0000-0000-000000000000", tenantID, nil)
```
jhendrixMSFT, line 60: Why was this removed?

sdk/keyvault/azkeys/utils_test.go
chlowell: The latter. Just trying to prevent hard to debug problems when working on multiple modules.

sdk/keyvault/azkeys/utils_test.go
chlowell: There's no need to replace tenant IDs because they aren't secret or considered in request matching. However, when the variable isn't set and `tenantID == ""`, this sanitizer will malform the header and break playback.

jhendrixMSFT: For the first part, that makes sense.

For the second part, sounds like this is something we should fix in `recording`.

heaths (approved):

jhendrixMSFT (approved):
Get a tracking issue opened for the problem with `recording`.

sdk/keyvault/azkeys/utils_test.go
chlowell: #17842

PR #17848, opened 2022-05-07
Service Bus: export NewReceivedMessage

ItalyPaleAle:
In sdk/messaging/azureservicebus, the method newReceivedMessage was not exported. This made it really hard for us (Dapr) to create our own unit tests, as without that method we cannot create a ReceivedMessage object with a body.

### DIFF
--- a/sdk/messaging/azservicebus/message.go
+++ b/sdk/messaging/azservicebus/message.go
@@ -205,10 +205,10 @@ func (m *Message) toAMQPMessage() *amqp.Message {
 	return amqpMsg
 }
 
-// newReceivedMessage creates a received message from an AMQP message.
+// NewReceivedMessage creates a received message from an AMQP message.
 // NOTE: this converter assumes that the Body of this message will be the first
 // serialized byte array in the Data section of the messsage.
-func newReceivedMessage(amqpMsg *amqp.Message) *ReceivedMessage {
+func NewReceivedMessage(amqpMsg *amqp.Message) *ReceivedMessage {
 	msg := &ReceivedMessage{
 		rawAMQPMessage: amqpMsg,
 		State:          MessageStateActive,
--- a/sdk/messaging/azservicebus/message_test.go
+++ b/sdk/messaging/azservicebus/message_test.go
@@ -52,7 +52,7 @@ func TestMessageUnitTest(t *testing.T) {
 func TestAMQPMessageToReceivedMessage(t *testing.T) {
 	t.Run("empty_message", func(t *testing.T) {
 		// nothing should blow up.
-		rm := newReceivedMessage(&amqp.Message{})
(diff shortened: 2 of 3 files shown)


richardpark-msft:
Let me see if we can brainstorm some solutions for this. 

We can't export the current NewReceivedMessage() function because it references a type from another package, which we do not normally do.

jhendrixMSFT:
We should change `Body()` to a `Body []byte` field on the `ReceivedMessage` type instead.

ItalyPaleAle:
@jhendrixMSFT Although that would fix the issue and make it possible for us to write a unit test, it does leave us with a lot more responsibility to code the equivalent of `NewReceivedMessage`.

Our preference would be for the method to be exported, so we don't have to write a lot of code, and especially maintain it in sync with the SDK.

PS: If `Message.toAMQPMessage` could be exported too, that would be a nice-to-have as well, for more unit tests.

richardpark-msft:
@ItalyPaleAle : 

> @jhendrixMSFT Although that would fix the issue and make it possible for us to write a unit test, it does leave us with a lot more responsibility to code the equivalent of `NewReceivedMessage`.
> 
> Our preference would be for the method to be exported, so we don't have to write a lot of code, and especially maintain it in sync with the SDK.
> 
> PS: If `Message.toAMQPMessage` could be exported too, that would be a nice-to-have as well, for more unit tests.

I might be misunderstanding what you're trying to do here because, at least for a test, you'd need to fill in the ReceivedMessage for your test to work, correct? Even NewReceivedMessage will require an entire amqp.Message to use (or equivalent) so you'd have to fill out _something_.

Do you happen to have a link to a test you're trying to write?

ItalyPaleAle:
I am working on updating Dapr to use the "track2" SDK for Service Bus. The PR is here: https://github.com/dapr/components-contrib/pull/1702

In the previous version, there was a test called `TestNewPubsubMessageFromAzServiceBusMessage` which creates one of our PubSub messages from a message returned by the SDK. With the current SDK we cannot create a test like that because we have no way to create a `Message` object with a body, so I had to [delete the test entirely](https://github.com/dapr/components-contrib/pull/1702/files#r867404192). Although the test isn't _strictly_ required, removing it does reduce our test coverage further.

We need a way to create `Message` objects with some data that we can then read. What you propose would allow us to create an object with a body, but we'd have to manually add _something_ to the body. If we could use `NewReceivedMessage`, we could pre-populate the object with a message from amqp and leverage all the work the SDK does for parsing, validation, etc.

richardpark-msft:
Closing this PR in favor of #17888.

After talking about it we're just going to convert the .Body() function back to a .Body field. There will be some minor edge cases where it can be nil, but this will be apparent to customers since they would have had to arrange it that way from the sending side.

PR #17860, opened 2022-05-09
Remove azcore version info from default User-Agent

chlowell:
Making this change was only a little more work than opening an issue to discuss the change, so here's a PR for discussion 😄 (I don't think this is a release blocker)

The other SDKs don't include their core library version in User-Agent strings, I suppose because the template in the [general design guidelines](https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy) doesn't include it. Does anyone recall the reasoning behind ourformat? Is `azcore` adoption interesting independent of client module adoption?

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -32,6 +32,7 @@
   * The internal poller types had their methods updated to conform to the `runtime.PollingHandler` interface.
   * The creation of resume tokens has been refactored so that implementers of `runtime.PollingHandler` don't need to know about it.
 * `NewPipeline()` places policies from `ClientOptions` after policies from `PipelineOptions`
+* Default User-Agent headers no longer include `azcore` version information
 
 ## 0.23.1 (2022-04-14)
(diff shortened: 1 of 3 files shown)


mikekistler (approved):
This looks good to me! I'm all for consistency across languages except where there's a compelling need to differ.

jhendrixMSFT (changes requested):
I believe, and could be wrong, that Go is unique here to the other languages in that you can use an arbitrary version of `azcore` with modules as long as its version is >= the minimum version as determined by semantic import versioning.

Do other languages have similar versioning rules?

jhendrixMSFT:
You bring up good points.  I was thinking about this from a diagnostic perspective, after upgrading `azcore` the failure rate for operation X increases.  However, User-Agent probably isn't the correct tool for this (we probably don't even capture failure rates).

We could also add it later if we do find a need for it.  Presumably, this wouldn't be a breaking change.

jhendrixMSFT (approved):

PR #17872, opened 2022-05-10
[azservicebus] Update readme, fix doc comments, etc...

richardpark-msft:
Some more doc comment and doc updates.

### DIFF
--- a/sdk/messaging/azservicebus/README.md
+++ b/sdk/messaging/azservicebus/README.md
@@ -87,7 +87,7 @@ func main() {
 Once you've created a [Client][godoc_client], you can interact with resources within a Service Bus Namespace:
 
 - [Queues][queue_concept]: Allows for sending and receiving messages. Often used for point-to-point communication.
-- [Topics][topic_concept]: As opposed to Queues, Topics are better suited to publish/subscribe scenarios. A topic can be sent to, but requires a subscription, of which there can be multiple in parallel, to consume from.
+- [Topics][topic_concept]: A topic is similar to a queue but splits the receiving and sending into separate entities. Messages are sent to a topic and are broadcast to Subscriptions, where they can be consumed independently, and in parallel by Receivers.
 - [Subscriptions][subscription_concept]: The mechanism to consume from a Topic. Each subscription is independent, and receives a copy of each message sent to the topic. Rules and Filters can be used to tailor which messages are received by a specific subscription.
 
 For more information about these resources, see [What is Azure Service Bus?][service_bus_overview].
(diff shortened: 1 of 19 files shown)


sdk/messaging/azservicebus/README.md
```diff
@@ -113,48 +113,79 @@ Once you've created a [Client][godoc_client] you can create a [Sender][godoc_sen
...
-
-if err != nil {
-  panic(err)
-}
+Sending a single message:
 
-// send a single message
-err = sender.SendMessage(context.TODO(), &azservicebus.Message{
+```go
+// Send a single message
```
chlowell, line 119:
```suggestion
```
(redundant)

sdk/messaging/azservicebus/README.md
```diff
@@ -113,48 +113,79 @@ Once you've created a [Client][godoc_client] you can create a [Sender][godoc_sen
...
 }, nil)
 ```
 
+#### Sending multiple messages using a batch
+
 You can also send messages in batches, which can be more efficient than sending them individually
 
 ```go
+// Sending multiple messages at a time using a MessageBatch
+
```
chlowell, line 131:
```suggestion
```

sdk/messaging/azservicebus/migrationguide.md
```diff
@@ -43,6 +34,10 @@ New (using `azservicebus`):
 // new code
 
 client, err := azservicebus.NewClientFromConnectionString(connectionString, nil)
+
+//  or, if you would like to use an `azidentity` credential:
```
chlowell, line 38:
```suggestion
//  or, if you would like to use an azidentity credential:
```
But also consider splitting this into its own block so you can show more context, like importing `azidentity` and calling `NewDefaultAzureCredential`. That might help people getting started with track 2.

sdk/messaging/azservicebus/README.md
```diff
@@ -113,48 +113,79 @@ Once you've created a [Client][godoc_client] you can create a [Sender][godoc_sen
...
+    // was full so it didn't go out with the previous SendMessageBatch call).
+    i--
+  } else if err != nil {
+    panic(err)
+  }
+}
+
+// check if any messages are remaining to be sent.
+if currentMessageBatch.NumMessages() > 0 {
+  err := sender.SendMessageBatch(context.TODO(), currentMessageBatch, nil)
```
chlowell, line 184: This is pretty big for a readme snippet. Could it be an example test instead?

chlowell (approved):

sdk/messaging/azservicebus/README.md
richardpark-msft: It is a bit big, but my worry is that if someone copy/pastes they'll miss the pattern they really need to follow. I think we get a lot of copy/paste programming so I just wanted to be safe.

sdk/messaging/azservicebus/migrationguide.md
richardpark-msft: Updated. I had a section later on in the doc specifically dedicated to the use of Azure Identity credentials, so I just linked to that.

PR #17888, opened 2022-05-11
[azservicebus] Changing ReceivedMessage.Body() to ReceivedMessage.Body (a field)

### LINKED ISSUE #17877: ReceivedMessage exports fields and a method which breaks mocking
Return types that mix exported fields and methods are not mockable and require a constructor.  We should avoid this.

richardpark-msft:
Changed ReceivedMessage.Body() to be a field of type `[]byte`. 

The entire reason it was a function was to return an error in the rare case that the underlying AMQP message was using the Value or Sequence section, rather than Data (or that Data itself was multiple byte arrays, not just a single one). Being a function makes it harder for users to create tests where they pass around `ReceivedMessage` since they can't create it themselves.

In the end it was agreed that just having a nil Body was enough signal for users to know that something was wrong, and that the error that was returned wouldn't have really been actionable (much like Body being nil, it was basically a true/false with exactly one error message).

Thanks to @ItalyPaleAle for pointing this out!

Fixes #17877

### DIFF
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 - `admin.Client` can now be configured using `azcore.Options`. (#17796)
 - `ReceivedMessage.TransactionPartitionKey` has been removed as this library doesn't support transactions.
+- `ReceivedMessage.Body()` is now a field. `Body` will be nil in the cases where it would have returned an error (where the underlying AMQP message had a payload in .Value, .Sequence or had multiple byte slices in .Data). (#TBD)
 
 ### Bugs Fixed
 
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -21,7 +21,7 @@
 
 - `admin.Client` can now be configured using `azcore.Options`. (#17796)
 - `ReceivedMessage.TransactionPartitionKey` has been removed as this library doesn't support transactions.
-- `ReceivedMessage.Body()` is now a field. `Body` will be nil in the cases where it would have returned an error (where the underlying AMQP message had a payload in .Value, .Sequence or had multiple byte slices in .Data). (#TBD)
+- `ReceivedMessage.Body()` is now a field. `Body` will be nil in the cases where it would have returned an error (where the underlying AMQP message had a payload in .Value, .Sequence or had multiple byte slices in .Data). (#17888)
 
 ### Bugs Fixed
 
--- a/sdk/messaging/azservicebus/message.go
+++ b/sdk/messaging/azservicebus/message.go
@@ -314,7 +314,7 @@ func newReceivedMessage(amqpMsg *amqp.Message) *ReceivedMessage {
 		State:          MessageStateActive,
 	}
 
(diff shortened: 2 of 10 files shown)


sdk/messaging/azservicebus/message.go
```diff
@@ -18,6 +17,11 @@ type ReceivedMessage struct {
 	// ApplicationProperties can be used to store custom metadata for a message.
 	ApplicationProperties map[string]interface{}
 
+	// Body corresponds to the first byte array in the Data section of an AMQP message.
```
jhendrixMSFT, line 20: Will this extra info be useful for the customer?  i.e. how many will know (or need to know) about these details?  And could this ever change?  As long as we're ok with formalizing this contract.

Also, the change description implies that a `nil` body _might_ be an error.  Am I reading too much into that?  I know that AMQP allows sending messages with no body, does SB support the same?  I want to make sure we don't have some ambiguous situation when the body is nil.

sdk/messaging/azservicebus/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 - `admin.Client` can now be configured using `azcore.Options`. (#17796)
 - `ReceivedMessage.TransactionPartitionKey` has been removed as this library doesn't support transactions.
+- `ReceivedMessage.Body()` is now a field. `Body` will be nil in the cases where it would have returned an error (where the underlying AMQP message had a payload in .Value, .Sequence or had multiple byte slices in .Data). (#17888)
```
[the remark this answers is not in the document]
richardpark-msft: No, this was the only one, which made it even more notable.

sdk/messaging/azservicebus/message.go
richardpark-msft: I think the key here is that only the user knows what a nil Body means. We can tell them we couldn't convert it, but the fact that they expected some payload but got nil tells them something is wrong (which is why it felt okay to just convert this back to a field).

If Service Bus didn't support raw AMQP messages then it'd be unneeded, but since we do it makes sense to document where this simplified envelope doesn't work (IMO).

When we add in raw AMQP message support they'll also be able to do something about it but for now it's just informational.

sdk/messaging/azservicebus/message.go
```diff
@@ -323,6 +316,10 @@ func newReceivedMessage(amqpMsg *amqp.Message) *ReceivedMessage {
 		State:          MessageStateActive,
 	}
 
+	if msg.rawAMQPMessage.Data != nil && len(msg.rawAMQPMessage.Data) == 1 {
```
jhendrixMSFT, line 319: Nit the `nil` check can be removed as a `nil` slice has a length of zero.

sdk/messaging/azservicebus/message.go
richardpark-msft: Removed.

richardpark-msft: Whoops, one second.

richardpark-msft: Now it's removed.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/message.go
richardpark-msft: Indeed! Fixed.

PR #17893, opened 2022-05-12
[azservicebus] Quality of life fixes for logging, fix for prefetch.

richardpark-msft:
Found a bug while adding in some more debug logging.

- Fixing a bug where the prefetch cache might not be drained if the user had cancelled the receive call. The prefetch call was specifically made so it no longer incurs any network I/O, so cancellation is something we should NOT do as it could cause us to lose messages we've already received.
- Adding (and fixing) logging to make things easier on diagnostics later.
- Cleaning up the close of the RPCLink so you can distinguish errors from normal shutdown. This eliminates a source of noise where it appears the negotiateClaim()'s response router was dying, when it was actually just the client itself being closed.

### DIFF
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -373,19 +373,22 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
 	}
 
 	flushAndReturn := func(lastErr error) ([]*ReceivedMessage, error) {
-		flushPrefetchedMessages(ctx, linksWithID.Receiver, &all)
+		flushPrefetchedMessages(linksWithID.Receiver, &all)
 
 		if len(all) > 0 {
 			// users don't typically check the other values in the return if the 'err' is non-nil
 			// so if we have any messages we'll just return them and nil out the error.
 			// If it's persistent they'll see it on their next ReceiveMessages() call.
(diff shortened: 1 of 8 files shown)


sdk/messaging/azservicebus/receiver.go
```diff
@@ -462,16 +474,24 @@ func fetchMessages(ctx context.Context, receiver internal.AMQPReceiver, maxMessa
...
+	count := 0
+	defer func() {
+		if count > 0 {
+			log.Writef(EventReceiver, "Flushed %d messages from receiver's internal cache", count)
+		}
+	}()
+
 	for {
-		am, err := receiver.Prefetched(ctx)
+		am, err := receiver.Prefetched(context.Background())
```
jhendrixMSFT, line 486: I think we should change the `Prefetched()` API to not take a context.  It doesn't perform any IO and is always non-blocking.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/receiver.go
richardpark-msft: I'll file an issue in go-amqp for that one.

richardpark-msft: https://github.com/Azure/go-amqp/issues/147

PR #17942, opened 2022-05-13
Add NewRecordingWithOptions constructor

jhendrixMSFT:
The NewRecording constructor defaults to a poor location for the
recording files and provides no facilities to override it.
Added a new constructor with options for specifying a custom recordings
directory.  The default is to now place the recordings under the
testdata directory.  The old constructor maintains its poor behavior.

### DIFF
--- a/sdk/internal/CHANGELOG.md
+++ b/sdk/internal/CHANGELOG.md
@@ -1,14 +1,16 @@
 # Release History
 
-## 1.0.1 (Unreleased)
+## 1.1.0 (Unreleased)
 
 ### Features Added
+* Added `recording.NewRecordingWithOptions` and corresponding `NewRecordingOptions` type to allow specifying custom recordings directory.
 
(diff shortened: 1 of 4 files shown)


jhendrixMSFT:
Bumped into this problem when attempting to move the recordings for `azblob` under `testdata`.  :(

jhendrixMSFT:
I believe we'll need this to fix https://github.com/Azure/azure-sdk-for-go/issues/17943

chlowell (approved):
We seem to have a couple generations of recording API. I haven't examined the differences yet but wonder whether we should deprecate this one and recommend using the newer `Start()` instead.

jhendrixMSFT:
Oh interesting.  Sounds like we need to do an audit of the current public surface area.  Maybe we should do that before we start changing things that might be deprecated in full.

richardpark-msft (approved):
Ouch.

PR #17946, opened 2022-05-13
[azservicebus] Vendoring go-amqp

richardpark-msft:
There are enough interface changes for go-amqp that we want to insulate this package somewhat. Simplest solution that we could come up with was to temporarily vendor go-amqp here. 

So this PR is a literal copy of go-amqp's code after I did a `go mod vendor` with:
- Fixing up of references to use our internally vendored copy of go-amqp:
- Adding in a NOTICE.txt to make sure we credit Kale and respect the sub-license.


sdk/messaging/azservicebus/NOTICE.txt
```diff
@@ -0,0 +1,55 @@
...
+           azgosdkhelp@microsoft.com
+
+The attached notices are provided for information only.
+
+License notice for go-amqp
+------------------------------------------------------------------------------
+
+    MIT License
+
+    Copyright (C) 2017 Kale Blankenship
```
richardpark-msft, line 36: @RickWinter, this is an exact copy of what's in the LICENSE.txt for go-amqp.

sdk/messaging/azservicebus/internal/go-amqp/.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Binary files (no line-ending conversions), diff using hexdump
```
jhendrixMSFT, line 1: Do we need this file?  Looks like a few others could be pruned too.

richardpark-msft: Yeah, let me go look. I tried to remove a few obvious ones.

richardpark-msft: Got rid of some more files.

sdk/messaging/azservicebus/internal/go-amqp/client.go
```diff
@@ -0,0 +1,203 @@
+// Copyright (C) 2017 Kale Blankenship
```
richardpark-msft, line 1: @RickWinter  - an example of our "combined" copyright (obviously subject to whatever we need to change).

sdk/messaging/azservicebus/internal/go-amqp/doc.go
```diff
@@ -0,0 +1,13 @@
...
+AMQP 1.0 is not compatible with AMQP 0-9-1 or 0-10, which are
+the most common AMQP protocols in use today.
+
+The example below shows how to use this package to connect
+to a Microsoft Azure Service Bus queue.
+*/
+
+// Copyright (C) 2017 Kale Blankenship
+// Portions Copyright (c) Microsoft Corporation
+package amqp // import "github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/go-amqp"
```
jhendrixMSFT, line 13: Trailing import comment?

RickWinter (approved):

sdk/messaging/azservicebus/internal/go-amqp/doc.go
richardpark-msft: Removed.

Issue #17759, opened 2022-04-27
Some key vault pollers have broken logic


jhendrixMSFT:
As a result, they might report that an LRO has completed when it's actually still in progress.

We've seen this with `azcertificates`, there might be other cases.

jhendrixMSFT:
All pollers should be fixed now.

PR #17984, opened 2022-05-16
Use generic runtime.Poller[T] for LROs

jhendrixMSFT:
LROs now return a *runtime.Poller[T] with their respective response
types and the in-box poller types have been removed.
Fixed poller terminal state detection so that calls to PollUntilDone
exit after the LRO has reached a terminal state.


jhendrixMSFT:
@heaths @schaabs please look at the changes in `client.go` and `pollers.go` as this is where the LRO changes reside.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17759

sdk/keyvault/azcertificates/client.go
```diff
@@ -84,82 +82,11 @@ func (b BeginCreateCertificateOptions) toGenerated() *generated.KeyVaultClientCr
...
-			return CreateCertificateResponse{}, err
-		}
-		b.getRawResponse = resp
-		if b.Done() {
-			break
-		}
-		time.Sleep(t)
-	}
-	return b.createResponse, nil
+	CertificateWithPolicy
```
jhendrixMSFT, line 85: Note that we now return the created certificate with the LRO completes instead of this intermediate `Operation` type.

jhendrixMSFT:
There appears to be a bug with URI sanitizing somewhere.  I patched the recorded response for now and opened https://github.com/Azure/azure-sdk-for-go/issues/17985 to track the bug.

sdk/keyvault/azcertificates/pollers.go
```diff
@@ -0,0 +1,136 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+type beginDeleteCertificateOperation struct {
+	resp *http.Response
+	poll func(context.Context) (*http.Response, error)
+}
+
+func (b *beginDeleteCertificateOperation) Done() bool {
+	return b.resp != nil && b.resp.StatusCode == http.StatusOK
```
chlowell, line 73: The client may be allowed to delete certs but forbidden to get them, in which case every poll will get a 403
```suggestion
	return runtime.HasStatusCode(b.resp, http.StatusOK, http.StatusForbidden)
```

sdk/keyvault/azcertificates/pollers.go
```diff
@@ -0,0 +1,136 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+type beginRecoverDeletedCertificate struct {
+	resp *http.Response
+	poll func(context.Context) (*http.Response, error)
+}
+
+func (b *beginRecoverDeletedCertificate) Done() bool {
+	return b.resp != nil && b.resp.StatusCode == http.StatusOK
```
chlowell, line 110: Similarly, the client may be allowed to recover certs but forbidden to get them
```suggestion
	return runtime.HasStatusCode(b.resp, http.StatusOK, http.StatusForbidden)
```

sdk/keyvault/azcertificates/pollers.go
```diff
@@ -0,0 +1,136 @@
...
+func (b *beginDeleteCertificateOperation) Done() bool {
+	return b.resp != nil && b.resp.StatusCode == http.StatusOK
+}
+
+func (b *beginDeleteCertificateOperation) Poll(ctx context.Context) (*http.Response, error) {
+	resp, err := b.poll(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusNotFound, http.StatusForbidden) {
```
heaths, line 81: Ah, here it is.

heaths (approved):

sdk/keyvault/azcertificates/client_test.go
```diff
@@ -652,6 +643,20 @@ func TestMergeCertificate(t *testing.T) {
 		},
 	}
 
+	poller, err := client.BeginCreateCertificate(ctx, certName, certPolicy, nil)
+	require.NoError(t, err)
+	// can't PollUntilDone for this scenario
+	resp, err := poller.Poll(ctx)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	if retryAfter := resp.Header.Get("Retry-After"); retryAfter != "" {
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

sdk/keyvault/azcertificates/pollers.go
jhendrixMSFT: This makes me think that instead we should remove 403 as successful status code in the `Poll()` methods so that an error is returned.  Thoughts?

jhendrixMSFT: The 403s have been removed from the list of successful status codes.

chlowell (approved):
Thanks for doing this!

PR #17987, opened 2022-05-17
Revise azsecrets readme and doc comments

chlowell:
Updating snippets, adding aka.ms links, making docs more consistent, removing references to access policy permissions because Key Vault now also supports RBAC, and other miscellany.


jhendrixMSFT (approved):
A few minor things to tweak.

PR #18075, opened 2022-05-17
Revise azkeys readme and doc comments

chlowell:
Revising azkeys docs similarly to #17987. I also found an inconsistent field name which I changed from CurveName -> Curve.


sdk/keyvault/azkeys/README.md
```diff
@@ -361,7 +361,7 @@ func main() {
 		panic(err)
 	}
 
-	pager := client.NewListPropertiesOfKeysPager(nil)
+	pager := client.ListPropertiesOfKeys(nil)
```
jhendrixMSFT, line 364: I guess this API still needs renaming?

sdk/keyvault/azkeys/client.go
```diff
@@ -237,26 +240,28 @@ func (c *Client) CreateECKey(ctx context.Context, name string, options *CreateEC
 	return createECKeyResponseFromGenerated(resp), nil
 }
 
-// CreateOctKeyOptions contains the optional parameters for the Client.CreateOCTKey method
+// CreateOctKeyOptions contains optional parameters for CreateOctKey
```
jhendrixMSFT, line 243: We can handle separately, but I think that `Oct` should be capitalized.

sdk/keyvault/azkeys/client.go
```diff
@@ -237,26 +240,28 @@ func (c *Client) CreateECKey(ctx context.Context, name string, options *CreateEC
...
 type CreateOctKeyOptions struct {
-	// Hardware Protected OCT Key
+	// HardwareProtected determines whether the key is is created in a hardware security module (HSM).
 	HardwareProtected *bool
 
-	// The key size in bits. For example: 2048, 3072, or 4096 for RSA.
+	// Size is the key size in bits.
```
jhendrixMSFT, line 248: The previous comments suggest that the value _should_ be a power of two and that there _might_ be a limit on acceptable values.  Are either of these true?

sdk/keyvault/azkeys/client.go
```diff
@@ -565,7 +569,7 @@ func purgeDeletedKeyResponseFromGenerated(i generated.KeyVaultClientPurgeDeleted
 
 // PurgeDeletedKey deletes the specified key. The purge deleted key operation removes the key permanently, without the possibility of recovery.
 // This operation can only be enabled on a soft-delete enabled vault. This operation requires the key/purge permission.
-// Pass nil to use the default options.
+// Pass nil for options to accept default values.
```
jhendrixMSFT, line 572: Given that this doesn't really complete after the API returns, seems like a comment should be added to indicate this quirk.

sdk/keyvault/azkeys/client.go
```diff
@@ -1222,14 +1213,13 @@ func (g GetRandomBytesOptions) toGenerated() *generated.KeyVaultClientGetRandomB
 	return &generated.KeyVaultClientGetRandomBytesOptions{}
 }
 
-// GetRandomBytesResponse is the response struct for the Client.GetRandomBytes function.
+// GetRandomBytesResponse is returned by GetRandomBytes.
 type GetRandomBytesResponse struct {
-	// The bytes encoded as a base64url string.
+	// Value is the bytes, URL-safe base64 encoded. TODO: is that so?
```
jhendrixMSFT, line 1218: Is there still an open question here?

sdk/keyvault/azkeys/client.go
```diff
@@ -1351,26 +1338,25 @@ func (c *Client) GetKeyRotationPolicy(ctx context.Context, keyName string, optio
...
-// ReleaseKeyOptions contains optional parameters for Client.ReleaseKey
+// ReleaseKeyOptions contains optional parameters for Client.ReleaseKey.
 type ReleaseKeyOptions struct {
-	// The version of the key to release
+	// Version is the version of the key to release
 	Version string
 
-	// The encryption algorithm to use to protected the exported key material
+	// Enc is the encryption algorithm used to protected exported key material.
 	Enc *ExportEncryptionAlg `json:"enc,omitempty"`
```
jhendrixMSFT, line 1347: I don't know if this was covered in the API review, but `Enc` is a bit too terse and non-descript IMO.

sdk/keyvault/azkeys/README.md
chlowell: No, this is evidence of me bungling a merge 😊

sdk/keyvault/azkeys/client.go
chlowell: Yes. I'll list the legal values.

sdk/keyvault/azkeys/client.go
chlowell: Sure, let's discuss it elsewhere. Counterargument is that "oct" is short for "octet sequence"; it isn't an acronym like "EC" and "RSA".

jhendrixMSFT: Good point, no need for further discussion.

jhendrixMSFT (approved):

sdk/keyvault/azkeys/client.go
chlowell: Agreed. Pretty sure we want `Algorithm` here, which is what other languages chose. I'll open a PR.

Issue #17905, opened 2022-05-12
Move azappconfig under data directory


jhendrixMSFT:
Per discussion we'll move `azappconfig`

Old location
`
github.com/Azure/azure-sdk-for-go/sdk/appconfig/azappconfig
`

New location
`
https://github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig
`

PR #18077, opened 2022-05-17
Copy azappconfig to its new location

jhendrixMSFT:
Updated references and examples to use the new path.


jhendrixMSFT:
@benbp what's the command to create the pipeline again?

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17905

richardpark-msft (commented):
Is this an actual copy, and not just a 'git mv'? This PR looks like it loses history.

jhendrixMSFT:
We have a chicken-and-egg problem here.  We can't move the old module until it's been deprecated, and we can't deprecate it until the new module is in place.

Well, I suppose we could cheat and deprecate it now, even though the replacement doesn't exist yet.  That would then allow us to `git mv` so history is preserved.

richardpark-msft:
> We have a chicken-and-egg problem here. We can't move the old module until it's been deprecated, and we can't deprecate it until the new module is in place.
> 
> Well, I suppose we could cheat and deprecate it now, even though the replacement doesn't exist yet. That would then allow us to `git mv` so history is preserved.

Let's not do that - we've put people through enough on that front. :)

richardpark-msft (approved):
Looks like the pipelines aren't quite there yet but I don't see any problems with the code itself that we'd address here.

jhendrixMSFT:
I renamed the current pipelines for `azappconfig` but I guess that doesn't fix the pipeline creation problem.

benbp:
> I renamed the current pipelines for `azappconfig` but I guess that doesn't fix the pipeline creation problem.

@jhendrixMSFT Ah, I was hoping with the service directory change the names wouldn't clash, but I guess we just look at the child path for simplicity. I would change the name of the deprecated azappconfig ci.yml file to like `ci.azappconfigold.yml` or something. Alternatively, you could just delete the old ci.yml file, and then `prepare` should work.

Another option is to edit the current pipelines to point to the new ci.yml file, however the pipeline generation for go all up will still be broken, so moving/renaming the old file is probably best.

jhendrixMSFT:
@benbp I'll still need the old pipeline to be functional in order to deprecate and release a final revision of the old module.

Can I rename the other `ci.yml` file in this PR?

Issue #14284, opened 2021-02-05
Add claims field to TokenRequestOptions


catalinaperalta:
Add claims field as part of CAE addition to azidentity.

RickWinter:
@chlowell please review this ask and update the cost for the work.

baywet:
Hi from the Microsoft Graph SDKs team 👋
We're working on the implementation of CA/CAE in the latest generation of our SDKs, so far the experience with azure identity has been great.
However, we need this dependency to fully close the loop on our end.
Do you have any timeline to ship CAE support in azure identity? Could we at least add the field in the next minor/patch so we can pass the information from our client and things will kick in when the implementation work is done? (happy to submit a PR to add the field).
Thanks for the insights!

CC @RickWinter @catalinaperalta @chlowell

PR #18166, opened 2022-05-18
adds a claims field placeholder

baywet:
partial #14284 adds a claims field placeholder to unlock Microsoft Graph SDKs work.

- ~[ ] Tests are included and/or updated for code changes.~ N/A

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Other Changes
 
+* Added a placeholder `Claims` field to `TokenRequestOptions` to allow consumers to start passing the value.
+
 ## 1.0.0 (2022-05-12)
 
 ### Features Added
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -111,6 +111,8 @@ type TelemetryOptions struct {
 type TokenRequestOptions struct {
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
+	// Claims contains the additional claims to be included in the token. See https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter for more information on format and content. Note this parameter is a placeholder and hasn't been implemented yet.
+	Claims string
 }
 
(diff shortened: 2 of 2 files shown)


chlowell:
I'm reluctant to add this field apart from a full CAE implementation because by itself it has no function and invites confusion. Can you elaborate on how this unblocks the Graph SDK? Even if `azcore.TokenCredential` had the API it would need to support CAE, actually doing so would require custom credential types (or perhaps a hacky adapter?) because no `azidentity` credential will claim to support CAE or request tokens with particular claims.

chlowell:
I'm reluctant to add this field apart from a full CAE implementation because by itself it has no function and invites confusion. Can you elaborate on how this unblocks the Graph SDK? Even if `azcore.TokenCredential` had the API it would need to support CAE, actually doing so would require custom credential types (or perhaps a hacky adapter?) because no `azidentity` credential will claim to support CAE or request tokens with particular claims.

baywet:
@chlowell thanks for the reply.
mainly because I had to return an error when I see the claim coming back from the http adapter. Having the field would allow all the infrastructure to be in place in the Microsoft Graph SDK, even though "it wouldn't do anything". And then when you ship the full implementation of CAE in the credentials, simply updating the reference would light it up in the Microsoft Graph SDKs.

See [this issue](https://github.com/microsoft/kiota-authentication-azure-go/issues/13) for more details

chlowell:
I see how this would be convenient by enabling you to do some of your CAE work in advance. However, I still think it's best to wait on this until we actually support CAE. Setting aside for the moment that I don't want to ship nonfunctional API, I see practical problems. I also don't want to commit to (release as stable) part of a feature in isolation from the rest of it. Hypothetically then--I don't think we should really do this--I suppose we would release this field in an `azcore` beta and then maintain parallel beta and stable versions until we add CAE support. That's a lot of work for a field that doesn't do anything 😅

baywet:
alright, thanks for the rationale. We'll keep returning errors that point to the issues in this repo then until the implementation work is done. Hopefully this will help driving the demand from the customer side. If you could share timelines on the issues it'd be helpful as well. Closing.

PR #18169, opened 2022-05-18
azsecrets uses runtime.Poller[T] for LROs

chlowell:
Replacing the bespoke poller types with `runtime.Poller[T]`. Along the way I moved the module version out of generated code, so we don't have to regenerate or edit generated code to release patch versions, and fixed some test problems.


sdk/keyvault/azsecrets/version.go
```diff
@@ -0,0 +1,12 @@
...
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azsecrets
+
+const (
+	moduleName = "azsecrets"
+	version    = "v0.8.0"
```
jhendrixMSFT, line 11: Will the CI scripts get confused if there are two locations for `version`?

jhendrixMSFT (approved):

sdk/keyvault/azsecrets/version.go
chlowell: Let's find out 😉 

If they do, I propose aligning on this pattern.

jhendrixMSFT:
> record with newer test proxy

How do we ensure we're recording with the latest test proxy that's compatible with what's in CI?

sdk/keyvault/azsecrets/version.go
jhendrixMSFT: We could, but I worry this will get us into trouble at some point as we now have two sources of truth that either must be kept in sync, or we know which is the "real" one.

Since we don't create a pipeline for data-plane packages, perhaps we just omit these constants entirely and leave managing them to the SDK author?  They aren't used by the generated code.

chlowell:
Destroying your container and running `eng/common/testproxy/docker-start-proxy.ps1 start` will get the image used in CI, which is recorded in `eng/common/testproxy/target_version.txt`. Maybe there's a better story when the proxy isn't in a container.

sdk/keyvault/azsecrets/version.go
chlowell: Sounds good to me: https://github.com/Azure/autorest.go/issues/825

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17759

PR #18170, opened 2022-05-18
Rename azkeys.ReleaseKeyOptions.Enc to .Algorithm

chlowell:
`Algorithm` is clearer and the name for this option in other languages.


sdk/keyvault/azkeys/client.go
```diff
@@ -1138,8 +1138,8 @@ type ReleaseKeyOptions struct {
 	// Version is the version of the key to release
 	Version string
 
-	// Enc is the encryption algorithm used to protected exported key material.
-	Enc *ExportEncryptionAlg `json:"enc,omitempty"`
+	// Algorithm is the encryption algorithm used to protected exported key material.
+	Algorithm *ExportEncryptionAlg `json:"algorithm,omitempty"`
```
jhendrixMSFT, line 1142: Is the tag change needed?  How is the value sent over the wire?

chlowell: Convenience models don't go over the wire so, I don't know why we have the tags. I assumed the purpose was to help applications de/serialize.

jhendrixMSFT: Good point.  We can leave it as-is, but you bring up an interesting point that these tags might be superfluous.

jhendrixMSFT (approved):

chlowell:
I found a few more breaking changes we should make in azkeys. Converting this to a draft while I add them here.

heaths (approved):

chlowell:
The further breaking changes are:
- Make model field types consistently pointers to things e.g. `Tags map[string]string` -> `map[string]*string`
- Delete redundant `Tags` and `Managed` fields on `ImportKeyOptions`, `DeletedKeyItem` (e.g. `ImportKeyOptions.Properties` already has these fields)
- Use the existing `Operation` enum type instead of `string`

sdk/keyvault/azkeys/client.go
```diff
@@ -90,7 +90,7 @@ type CreateKeyOptions struct {
 	PublicExponent *int32 `json:"public_exponent,omitempty"`
 
 	// Tags is application specific metadata in the form of key-value pairs.
-	Tags map[string]string `json:"tags,omitempty"`
+	Tags map[string]*string `json:"tags,omitempty"`
```
heaths, line 93: Is this consistent with other SDKs? Tag dictionaries on Azure can't even have an empty/null value so this should never be nil nor could it ever be empty which I know is the zero value if using only a single return from `map[]` instead of the 2 value return. Just seems the indirection here is unnecessary.

chlowell: Yes. This is how autorest.go models it, so all our management plane libraries use this type for tags. I agree the pattern is awkward, but it enables us to accommodate future null elements, however unlikely they may be, without a breaking change.

heaths: Okay. I wouldn't expect it in this PR, but just wanted to make sure you were aware. Sorry about that. ;)

sdk/keyvault/azkeys/models.go
```diff
@@ -194,14 +194,38 @@ func jsonWebKeyFromGenerated(i *generated.JSONWebKey) *JSONWebKey {
...
 	}
 
+	ops := make([]*Operation, len(i.KeyOps))
+	for j, op := range i.KeyOps {
+		if op == nil {
+			ops[j] = nil
+		} else {
+			switch *op {
+			case string(OperationDecrypt):
+				ops[j] = to.Ptr(OperationDecrypt)
```
jhendrixMSFT, line 204: Seels like `op` is already a pointer-to-type, can't we just use its value?

chlowell: Yes, there's no need to confine the values to the set of defined constants.

sdk/keyvault/azkeys/models.go
```diff
@@ -194,14 +194,23 @@ func jsonWebKeyFromGenerated(i *generated.JSONWebKey) *JSONWebKey {
 		return &JSONWebKey{}
 	}
 
+	ops := make([]*Operation, len(i.KeyOps))
+	for j, op := range i.KeyOps {
+		if op == nil {
+			ops[j] = nil
+		} else {
+			ops[j] = to.Ptr(Operation(*op))
```
jhendrixMSFT, line 202: Can't this be simplified to just `ops[j] = op`?  Which then makes me wonder if this `for` loop can just be removed completely.

chlowell: The loop translates `[]*string` to `[]*Operation` because we need the latter type and the two aren't interchangeable. `Operation` does have underlying type `string`; do you mean there's a terser way to do this?

jhendrixMSFT: Oh I thought that `i.KeyOps` was of type `[]*Operation` was also changed in this PR?

chlowell: `i.KeyOps` here is a generated field of type `[]*string`. `Operation` is a convenience type. This code follows from changing the type of the convenience layer's `JSONWebKey.KeyOps` from `[]*string` to `[]*Operation` (also in this PR).

jhendrixMSFT: Ah yes I totally missed that, I get it now.

You can simplify this a bit though.

```go
ops[j] = (*Operation)(op)
```

jhendrixMSFT (approved):

sdk/keyvault/azkeys/models.go
chlowell: Thanks for pointing that out, it's a lot tidier.

PR #18192, opened 2022-05-19
expose getkey in crypto for simpler use

tg123:

### DIFF
--- a/sdk/keyvault/azkeys/crypto/crypto_client.go
+++ b/sdk/keyvault/azkeys/crypto/crypto_client.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys"
 	generated "github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys/internal/generated"
 	shared "github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal"
 )
@@ -25,10 +26,11 @@ import (
 // is able to get that material from Key Vault. When the required key material is unavailable,
 // cryptographic operations are performed by the Key Vault service.
 type Client struct {
-	kvClient   *generated.KeyVaultClient
-	vaultURL   string
-	keyID      string
-	keyVersion string
+	kvClient    *generated.KeyVaultClient
+	vaultURL    string
+	keyID       string
+	keyVersion  string
+	vaultClient *azkeys.Client
 }
 
 // ClientOptions are the configurable options on a Client.
@@ -109,11 +111,17 @@ func NewClient(keyURL string, credential azcore.TokenCredential, options *Client
 		return nil, err
 	}
 
+	vaultClient, err := azkeys.NewClient(vaultURL, credential, &azkeys.ClientOptions{options.ClientOptions})
+	if err != nil {
+		return nil, err
+	}
+
 	return &Client{
-		kvClient:   generated.NewKeyVaultClient(pl),
-		vaultURL:   vaultURL,
-		keyID:      keyID,
-		keyVersion: keyVersion,
+		kvClient:    generated.NewKeyVaultClient(pl),
+		vaultURL:    vaultURL,
+		keyID:       keyID,
+		keyVersion:  keyVersion,
+		vaultClient: vaultClient,
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


jhendrixMSFT (changes requested):
What's the scenario here?

tg123:
> What's the scenario here?

it has to create 2 clients, Vault client to get key info and crypto client to sign with keys
Here to add a shortcut to getkey to make life easier

tg123:
the real motivation is I am updating https://github.com/tg123/azkeyvault
and wish the ctor just take `crypto.Client`

and if az sdk accept go Signer binding, I am happy to contribute it to az sdk

jhendrixMSFT:
The separate clients are intentional; however, I wasn't involved in the design so I can't comment on why.  @heaths do you know why this design was selected?

heaths:
A `KeyClient` acts on a Key Vault endpoint, while a `CryptographyClient` acts on a key URI (versioned or not). They have very different purposes and lifetimes. That said, all the languages did add a convenience method to get a `CryptographyClient` from a `KeyClient`: https://docs.microsoft.com/dotnet/api/azure.security.keyvault.keys.keyclient.getcryptographyclient?view=azure-dotnet 

This should take a required key name and optional key version. The intent - for those client libraries where it matters - was to share a pipeline as well.

heaths:
I might've misunderstood now that I look at the code. This is added a `Get` to the crypto client. We didn't do that in other languages because we figured you already have/know the key if you created the crypto client, in which case just use the key client. Also, since the crypto client wraps *using* the key + some value add in some languages with local crypto, we didn't see value in getting the key to use it in different ways. For a low-level library, IMO, it's better to have fewer, less ambiguous ways to do the same thing.

jhendrixMSFT:
I see, thanks for the explanation.

@tg123 can you please update the PR per Heath's comment?

tg123:
I was trying to add this

[`CryptographyClient (Azure.Security.KeyVault.Keys.JsonWebKey key)`](https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.keys.cryptography.cryptographyclient.-ctor?view=azure-dotnet#azure-security-keyvault-keys-cryptography-cryptographyclient-ctor(azure-security-keyvault-keys-jsonwebkey))

at first.

the idea is to create a crypto client from a [key object](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys#Key) but I have no way to get `credential` to create it

thus, my workaround is as this PR: to do the reverse way, create the key object from crypto client.
then I can get a `crypto client` as param and get the key info without asking `credential` from users.

jhendrixMSFT:
I see your point and ran into this same thing while working on some refactoring prototype for `azblob`.  I have a pattern to fix this.  Can you please open an issue so we can track this feature request?

heaths:
Note that construction patterns in other libraries that take a JWT expect that the caller has a full JWT complete with whatever keys they need e.g., private keys for decryption, public keys for encryption, etc. It is not meant to call back to Key Vault for any reason, thus doesn't need a credential. Supporting that scenario for the Go SDK might make sense only if their crypto library is doing any local crypto work; otherwise, it should not take a JWT because it does no good. Import keys from a JWT into Key Vault and then you could use the crypto client, if that's more desirable.

heaths:
If Key Vault has to be accessed, you need a credential. If you already have a key you can encrypt/verify/wrap with a public key already. I don't understand why you want to create a crypto client from a key if you already have the key. We don't do this in other language. We do support creating a crypto client for a specific key in key vault. When possible, we initially attempt to download the public key material (if possible - may not have `get` permissions) and do with that what we can locally; if nothing else, we call into Key Vault to do, for example, private key operations. This will always require a credential.

With the other languages, when we support creating a crypto client from a key client, it's because the underlying pipeline already has the credential and the user doesn't need to specify it again. If they need a different credential, they can create a crypto client explicitly as before.

tg123:
I am creating a library which ideally just requires `1`  client, and without cred, as input

heaths:
> I am creating a library which ideally just requires `1` client, and without cred, as input

How do you expect to get a key from Key Vault without credentials?

heaths:
Can you describe your exact scenario you want enabled and how our existing APIs don't already provide that?

jhendrixMSFT:
@tg123 per the C# example linked earlier, the established pattern is to have a factory method on `azkeys.Client` that creates a `crypto.Client`.  Which means you'd only need to keep an `azkeys.Client`.  Presumably this would address your scenario?

tg123:
> @tg123 per the C# example linked earlier, the established pattern is to have a factory method on `azkeys.Client` that creates a `crypto.Client`. Which means you'd only need to keep an `azkeys.Client`. Presumably this would address your scenario?

Yes that is good

jhendrixMSFT:
Closing in favor of https://github.com/Azure/azure-sdk-for-go/pull/18260

PR #18193, opened 2022-05-19
Remove locale in link

jhendrixMSFT:
Per guidelines in https://aka.ms/azsdk/guideline/links

### DIFF
--- a/documentation/new-version-quickstart.md
+++ b/documentation/new-version-quickstart.md
@@ -19,7 +19,7 @@ You will need to authenticate to Azure either by using Azure CLI to sign in or s
 
 ### Using Azure CLI to Sign In
 
-You could easily use `az login` in command line to sign in to Azure via your default browser. Detail instructions can be found in [Sign in with Azure CLI](https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli).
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@benbp is it possible to set up a pipeline for docs that just performs link verification and nothing else?  This break was noticed after-the-fact due to lack of CI.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=1588275&view=results

benbp:
@jhendrixMSFT let me look into it. I'm vaguely recalling that we had to disable CI link verification due to mgmt plane onboarding chicken or egg problems? Otherwise I would have expected this same pipeline that's failing to now to have failed as a PR gate.

tadelesh:
> Looks like it was overridden, probably due to CI hanging like it was for me this morning.
> 
> #17763

I override the check because of the chicken or egg problem for the new created doc. Sorry for miss this link problem. It is weird in my region; the URL works fine with `en-us`.

tadelesh (approved):

tadelesh:
> > > Looks like it was overridden, probably due to CI hanging like it was for me this morning.
> > > #17763
> > 
> > 
> > I override the check because of the chicken or egg problem for the new created doc. Sorry for miss this link problem. It is weird in my region; the URL works fine with `en-us`.
> 
> @tadelesh we explicitly disallow region/language links in favor of the non-region specific URLs which is probably why it failed.

Got it. Thanks.

PR #18202, opened 2022-05-20
Update aztemplate with latest azcore and azidentity

jhendrixMSFT:
Added version.go per pattern and demonstrated how to create a pipeline.
Tweaked APIs to be indicative of real-world APIs.


sdk/template/aztemplate/client.go
```diff
@@ -1,37 +1,53 @@
...
+	}, &options.ClientOptions)
 
-	return &Client{client: tc}, nil
+	return &Client{client: internal.NewTemplateClient(pl)}, nil
+
+}
 
+// SomeServiceActionOptions contains the optional values for the Client.SomeServiceAction method.
+type SomeServiceActionOptions struct {
+	// placeholder for future options
```
chlowell, line 47: Should this demonstrate when fields should be pointers to things (stuff that goes on the wire) and when they should be the things themselves (stuff that does not like resume tokens)?

chlowell (approved):

sdk/template/aztemplate/client.go
jhendrixMSFT: Good point.  In fact, the exiting API needs further refinement to return a response type as well as an error.  And if we were to really flesh this out, examples for pagers and pollers should be added.

PR #18204, opened 2022-05-20
Removing floor for polling frequency when running tests

jhendrixMSFT:
The floor adds unnecessary delays when executing unit tests and can
decrease end-to-end test execution time by 3x.

### DIFF
--- a/sdk/azcore/runtime/poller_test.go
+++ b/sdk/azcore/runtime/poller_test.go
@@ -102,7 +102,7 @@ func TestLocPollerSimple(t *testing.T) {
 	}
 	var respFromCtx *http.Response
 	ctxWithResp := WithCaptureResponse(context.Background(), &respFromCtx)
-	_, err = lro.PollUntilDone(ctxWithResp, &PollUntilDoneOptions{Frequency: time.Second})
+	_, err = lro.PollUntilDone(ctxWithResp, &PollUntilDoneOptions{Frequency: time.Millisecond})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Test time for `azcertificates`.
- before: 157.119s
- after: 21.203s

sdk/azcore/runtime/poller_test.go
```diff
@@ -102,7 +102,7 @@ func TestLocPollerSimple(t *testing.T) {
 	}
 	var respFromCtx *http.Response
 	ctxWithResp := WithCaptureResponse(context.Background(), &respFromCtx)
-	_, err = lro.PollUntilDone(ctxWithResp, &PollUntilDoneOptions{Frequency: time.Second})
+	_, err = lro.PollUntilDone(ctxWithResp, &PollUntilDoneOptions{Frequency: time.Millisecond})
```
chlowell, line 105: Do smaller frequencies break something?

chlowell (approved):
I love it! So much simpler than my first idea 🤣

sdk/azcore/runtime/poller_test.go
chlowell: Really I'm wondering whether something will go wrong in recording playback, where I'll be trying `time.Nanosecond`. I suppose the worst thing that could happen would be a somewhat longer sleep on systems with coarser clock resolution.

jhendrixMSFT: Ah I see, I misunderstood the question.  This is what I get for answering before my first cup of coffee.

Issue #18179, opened 2022-05-19
Fix ci `Install GoLintCLI and dependencies.` failure


tadelesh:
`Install GoLintCLI and dependencies.` may accidentally fail because of network issue. Add retry to avoid false alert for both daily and weekly ci.

PR #18239, opened 2022-05-25
ci: add retry to the golangci-lint install step to prevent failure of network issue

tadelesh:
There are many daily ci failure (e.g. [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1598529&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=45ee1ca9-5b97-59ac-94ed-a114ab499701)) because of the failure of golangci-lint install step. Add retry to resolve it.
Resolve https://github.com/Azure/azure-sdk-for-go/issues/18179

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -23,6 +23,7 @@ steps:
       golangci-lint --version
     displayName: 'Install GoLintCLI and dependencies.'
     workingDirectory: $(System.DefaultWorkingDirectory)
+    retryCountOnTaskFailure: 3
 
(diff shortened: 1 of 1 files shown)


richardpark-msft (approved):
(not the final approver)

Looks good. It'd be nice if we could cache the tools on something we have greater control on but I'm not sure if we do that with tools in general.

weshaggard:
>It'd be nice if we could cache the tools on something we have greater control on but I'm not sure if we do that with tools in general.

We do cache tools and packages in other places and if this continues to be an issue, we can add something like that here as well.

PR #18258, opened 2022-05-26
Revise azsecrets model fields

chlowell:
Making model field types consistently pointers to things instead of values, removing redundant fields, and renaming `Properties.IsManaged` to `.Managed` to align with `azkeys` and the other Azure SDKs.


jhendrixMSFT (approved):

sdk/keyvault/azsecrets/client.go
```diff
@@ -650,12 +539,12 @@ type ListPropertiesOfSecretVersionsResponse struct {
...
 
 	// Secrets is the page's content.
-	Secrets []SecretItem `json:"value,omitempty" azure:"ro"`
+	Secrets []*SecretItem `json:"value,omitempty" azure:"ro"`
 }
 
 // create ListSecretsPage from generated pager
 func listSecretVersionsPageFromGenerated(i generated.KeyVaultClientGetSecretVersionsResponse) ListPropertiesOfSecretVersionsResponse {
-	var secrets []SecretItem
+	var secrets []*SecretItem
```
richardpark-msft, line 547: Why do these need to be pointers? I feel like we go back and forth about when things are or are not - in this case it seems like they wouldn't need to be unless you're saying it can sometimes be nil?

jhendrixMSFT: The short answer is "for consistency with our generated modules".

For the long answer, this was done for versioning tolerance and potential support for `x-nullable` in the future as we didn't want that to be a breaking change (it's not for other languages as they use reference types).

chlowell: It's future proofing. These elements won't be null today but if Key Vault makes them nullable, we'd have to make this breaking change at that time.

PR #18259, opened 2022-05-26
Fix key release always returning an error

chlowell:
Secure key release doesn't work when the application doesn't specify a version, because the version is a path parameter and generated code doesn't allow empty path segments. Key Vault allows the version segment to be empty, so this PR adds autorest transforms to remove the validation, fixes the test that was intended to cover this and another with a similar bug, and adds coverage of the version-specified scenario.


sdk/keyvault/azkeys/autorest.md
```diff
@@ -8,13 +8,24 @@ version: "^3.0.0"
...
 license-header: MICROSOFT_MIT_NO_VERSION
-clear-output-folder: true
+clear-output-folder: false
 output-folder: internal/generated
-module: azkeys
 openapi-type: "data-plane"
 security: "AADToken"
 security-scopes:  "https://vault.azure.net/.default"
-use: "@autorest/go@4.0.0-preview.38"
+use: "@autorest/go@4.0.0-preview.41"
```
jhendrixMSFT, line 16: FYI I released `preview.42` this morning that contains the change to omit the `moduleName` and `moduleVersion` constants.

sdk/keyvault/azkeys/client_test.go
```diff
@@ -504,7 +484,7 @@ func TestUpdateKeyPropertiesImmutable(t *testing.T) {
 					},
 					ReleasePolicy: &ReleasePolicy{
 						Immutable:     to.Ptr(true),
-						EncodedPolicy: marshalledPolicy,
+						EncodedPolicy: getMarshalledReleasePolicy(testAttestationURL),
```
jhendrixMSFT, line 487: When we write a test, how do we know we need to use this?

chlowell: Tribal knowledge and observing the prior art, I suppose. Setting a release policy is only necessary when testing key release, which we already do, so I don't expect this helper will be more commonly used.

jhendrixMSFT (approved):

PR #18266, opened 2022-05-27
CLI Credential: Avoid calling Azure CLI via cmd.exe/sh

magodo:

It is not necessary to invoke Azure CLI via either `cmd.exe` or `sh`, as itself is
executable and has proper interpreter settings. This avoids some
unneeded assumptions to the running environment, e.g. in Windows it is
then not needed to always ensure `cmd.exe` is available and correctly
set in the %PATH% (see: https://github.com/Azure/aztfy/issues/119).

Also, the working directory for the spawned process is also not
needed to set. By default, it uses the current working directory, which
is just fine.

### DIFF
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -110,8 +110,14 @@ func defaultTokenProvider() func(ctx context.Context, resource string, tenantID 
 		var cliCmd *exec.Cmd
 		if runtime.GOOS == "windows" {
 			cliCmd = exec.CommandContext(ctx, "az", args...)
+			dir := os.Getenv("SYSTEMROOT")
+			if dir == "" {
+				return nil, newCredentialUnavailableError(credNameAzureCLI, "environment variable 'SYSTEMROOT' has no value")
+			}
+			cliCmd.Dir = dir
 		} else {
 			cliCmd = exec.CommandContext(ctx, "az", args...)
+			cliCmd.Dir = "/bin"
 		}
 		cliCmd.Env = os.Environ()
 		var stderr bytes.Buffer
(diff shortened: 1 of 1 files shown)


chlowell (changes requested):
> Also, the working directory for the spawned process is also not
needed to set. By default, it uses the current working directory, which
is just fine.

Setting the working directory is actually a requirement that followed from a security review. Invoking `az` from a protected directory is an attempt to avoid executing a malicious executable placed in the application's working directory. This does mean that CLI authentication doesn't work when `az` isn't on the path. Can you elaborate more on the scenario motivating this? Is it not feasible to set the path?

magodo:
@chlowell I've added the `dir` part back. The user scenario can be found here: https://github.com/Azure/aztfy/issues/119

jhendrixMSFT:
@magodo do we know why `cmd.exe` isn't on the path?  I didn't see an explanation for that.

jhendrixMSFT:
@schaabs did you do the original threat modeling for CLI credential?

magodo:
There is a new report of this issue in https://github.com/Azure/aztfexport/issues/484

RickWinter:
FYI, @chlowell

chlowell:
I don't see a solid technical reason not to exec az as you propose and would probably take that approach if I were writing this code today. However, I don't see practical benefit in making this change now. We have a stable, well understood implementation that has apparently worked in virtually all cases for years. Execing az *might* prevent an error for users without cmd.exe on their path, but it's not certain to do so, that's a very unusual case, and this change could break applications whose users rely on shell behavior implicit in the current implementation. So, I see more risk than benefit in this change and would like to understand the scenario better before trying to address it.

PR #18272, opened 2022-05-30
Deprecated sdk/resourcemanager/machinelearningservices/armmachinelearningservices

Alancere:


tadelesh (changes requested):
You need to upgrade a fix version for the deprecation.

tadelesh (approved):

PR #18282, opened 2022-05-31
Revise azcertificates models

chlowell:
Mostly renames in alignment with the other Key Vault modules. Deleting all the fields of `UpdateCertificatePropertiesOptions` looks drastic, but they were unused and/or redundant.

### DIFF
--- a/sdk/keyvault/azcertificates/client.go
+++ b/sdk/keyvault/azcertificates/client.go
@@ -434,8 +434,8 @@ func (c *Client) ImportCertificate(ctx context.Context, certificateName string, 
 	}, nil
 }
 
-// ListCertificatesOptions contains optional parameters for Client.ListCertificates
-type ListCertificatesOptions struct {
+// ListPropertiesOfCertificatesOptions contains optional parameters for Client.ListCertificates
+type ListPropertiesOfCertificatesOptions struct {
 	// placeholder for future optional parameters.
 }
 
@@ -469,7 +469,7 @@ func listCertsPageFromGenerated(i generated.KeyVaultClientGetCertificatesRespons
 // public part of a stored certificate. The LIST operation is applicable to all certificate types, however only the
 // base certificate identifier, attributes, and tags are provided in the response. Individual versions of a
 // certificate are not listed in the response. This operation requires the certificates/list permission.
-func (c *Client) NewListPropertiesOfCertificatesPager(options *ListCertificatesOptions) *runtime.Pager[ListPropertiesOfCertificatesResponse] {
+func (c *Client) NewListPropertiesOfCertificatesPager(options *ListPropertiesOfCertificatesOptions) *runtime.Pager[ListPropertiesOfCertificatesResponse] {
 	pager := c.genClient.NewGetCertificatesPager(c.vaultURL, nil)
 	return runtime.NewPager(runtime.PagingHandler[ListPropertiesOfCertificatesResponse]{
 		More: func(page ListPropertiesOfCertificatesResponse) bool {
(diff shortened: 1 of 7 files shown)


jhendrixMSFT (approved):

sdk/keyvault/azcertificates/client.go
```diff
@@ -70,7 +70,7 @@ type BeginCreateCertificateOptions struct {
 	Enabled *bool `json:"enabled,omitempty"`
 
 	// Application specific metadata in the form of key-value pairs
-	Tags map[string]string `json:"tags,omitempty"`
+	Tags map[string]*string `json:"tags,omitempty"`
```
richardpark-msft, line 73: Is setting a value to 'nil' here equivalent to deleting it?

sdk/keyvault/azcertificates/client.go
```diff
@@ -1081,18 +1066,18 @@ func (c *Client) UpdateCertificateProperties(ctx context.Context, certificateNam
...
+	version := ""
+	if properties.Version != nil {
+		version = *properties.Version
 	}
 	resp, err := c.genClient.UpdateCertificate(
 		ctx,
 		c.vaultURL,
 		certificateName,
-		options.Version,
+		version,
```
richardpark-msft, line 1077: Empty version string is okay?

sdk/keyvault/azcertificates/testdata/recordings/TestCRUDOperations.json
```diff
@@ -4,28 +4,27 @@
...
         "Cache-Control": "no-cache",
         "Content-Length": "97",
         "Content-Type": "application/json; charset=utf-8",
-        "Date": "Mon, 16 May 2022 18:13:51 GMT",
+        "Date": "Wed, 01 Jun 2022 16:09:58 GMT",
         "Expires": "-1",
         "Pragma": "no-cache",
         "Strict-Transport-Security": "max-age=31536000;includeSubDomains",
-        "WWW-Authenticate": "Bearer authorization=\u0022https://login.windows.net/00000000-0000-0000-0000-000000000000\u0022, resource=\u0022https://vault.azure.net\u0022",
+        "WWW-Authenticate": "Bearer authorization=\u0022https://login.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47\u0022, resource=\u0022https://vault.azure.net\u0022",
```
richardpark-msft, line 22: Probably scrub this.

sdk/keyvault/azcertificates/utils_test.go
```diff
@@ -58,12 +58,6 @@ func TestMain(m *testing.M) {
 		if err != nil {
 			panic(err)
 		}
-
```
richardpark-msft, line 61: Don't you still need this? I guess we'll always use a particular tenant, but if there were an external contributor or even one of us testing against a personal account we'd want this to be replaced, right?

richardpark-msft (changes requested):

sdk/keyvault/azcertificates/client.go
chlowell: Yes.

sdk/keyvault/azcertificates/testdata/recordings/TestCRUDOperations.json
chlowell: It's okay--tenant IDs aren't secret.

sdk/keyvault/azcertificates/utils_test.go
chlowell: No. We need a tenant ID in the recorded `WWW-Authenticate` header, to prevent a parsing error in the challenge authentication policy, but the value doesn't matter because it isn't used in matching and we don't authenticate in playback. Since the live tenant ID isn't secret, we can simply record that.

sdk/keyvault/azcertificates/client.go
richardpark-msft: Talked with @chlowell offline - we're kind of stuck just due to the fact that string can't be nil and (even if it's not valid today) we need to do it this particular way to handle any potential future change in the API.

richardpark-msft (approved):

PR #18286, opened 2022-06-01
Remove trailing slashes from parsed vault URLs

### LINKED ISSUE #18190: [Key Vault] Ensure port numbers are preserved in vault URLs parsed from IDs
See https://github.com/Azure/azure-sdk-for-python/issues/24446 for issue context, and https://github.com/Azure/azure-sdk-for-python/pull/24466 for Python's fix.

chlowell:
While adding test cases verifying we preserve ports when parsing Key Vault identifiers (closes #18190), I noticed we add trailing slashes to parsed vault URLs e.g., we return `https://foo.vault.com/` as the vault URL of `https://foo.vault.com/keys/mykey`. I think we should stop doing this because although I doubt the trailing slash breaks anything, it's unnecessary and semantically incorrect in implying a directory or an empty path.

### DIFF
--- a/sdk/keyvault/internal/CHANGELOG.md
+++ b/sdk/keyvault/internal/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Release History
 
-## 0.5.1 (Unreleased)
+## 0.6.0 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
+* `ParseID()` no longer appends a trailing slash to vault URLs
 
 ### Bugs Fixed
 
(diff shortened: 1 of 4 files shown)


jhendrixMSFT (approved):
Presumably the trailing `/` was included to simplify joining the paths together?  At any rate, we do have `runtime.JoinPaths()` to help with this.

richardpark-msft (changes requested):
Was this path ever directly returned from any of our APIs? (if not, no big deal). If so, let's add it to the "breaking changes" list in the changelog.

richardpark-msft (approved):

PR #18308, opened 2022-06-02
Client secret and certificate credentials support ESTS-R

### LINKED ISSUE #15605: Support ESTS-R
Would be easiest with MSAL handling the details (MSAL adoption is tracked by #14920).

chlowell:
Closes #15605

### DIFF
--- a/sdk/azidentity/client_certificate_credential.go
+++ b/sdk/azidentity/client_certificate_credential.go
@@ -12,10 +12,10 @@ import (
 	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/x509"
-
 	"encoding/base64"
 	"encoding/pem"
 	"errors"
+	"os"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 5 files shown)


sdk/azidentity/client_certificate_credential.go
```diff
@@ -71,6 +71,7 @@ func NewClientCertificateCredential(tenantID string, clientID string, certs []*x
 	o := []confidential.Option{
 		confidential.WithAuthority(runtime.JoinPaths(authorityHost, tenantID)),
 		confidential.WithHTTPClient(newPipelineAdapter(&options.ClientOptions)),
+		confidential.WithAzureRegion(os.Getenv(azureRegionalAuthorityName)),
```
jhendrixMSFT, line 74: I guess `WithAzureRegion()` handles the empty string as a no-op?

chlowell: Yes. MSAL sets a string field with this value. Passing an empty string is equivalent to accepting that field's zero value, which selects the default, non-regional authority.

jhendrixMSFT (approved):

PR #18309, opened 2022-06-03
Consistent Update*Properties methods for Key Vault clients

chlowell:
Methods for updating Key Vault resource properties are inconsistent across our 3 clients. The cert client is the odd duck, taking a resource name and a `Properties` struct instead of a certificate model:
```go
UpdateCertificateProperties(ctx context.Context, certificateName string, properties Properties, options *UpdateCertificatePropertiesOptions)
UpdateKeyProperties(ctx context.Context, key Key, options *UpdateKeyPropertiesOptions)
UpdateSecretProperties(ctx context.Context, secret Secret, options *UpdateSecretPropertiesOptions)
```

First thought then is, why not have `UpdateCertificateProperties` follow along by taking a `Certificate` instead of the string and `Properties`? Well, there are two models representing certificates, reflecting a quirk of the REST API; either of these would work for `UpdateCertificateProperties`. Both certificate model types have a `Properties` field, as do all resource models, so I chose instead to have all update methods take one of those. That gives us these signatures:

```go
UpdateCertificateProperties(ctx context.Context, properties Properties, options *UpdateCertificatePropertiesOptions)
UpdateKeyProperties(ctx context.Context, properties Properties, options *UpdateKeyPropertiesOptions)
UpdateSecretProperties(ctx context.Context, properties Properties, options *UpdateSecretPropertiesOptions)
```

This design has a practical problem: `UpdateKeyProperties` can update a key's release policy, but the policy is part of `Key` and not `Properties`. So, I moved that field to `Properties`, as have the other SDKs (I suppose for this reason).

### DIFF
--- a/sdk/keyvault/azkeys/client.go
+++ b/sdk/keyvault/azkeys/client.go
@@ -148,7 +148,7 @@ func createKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyResponse)
 	vaultURL, name, version := shared.ParseID(g.Key.Kid)
 	return CreateKeyResponse{
 		Key: Key{
-			Properties: keyPropertiesFromGenerated(g.Attributes, g.Key.Kid, name, version, g.Managed, vaultURL, g.Tags),
+			Properties: keyPropertiesFromGenerated(g.Attributes, g.Key.Kid, name, version, g.Managed, vaultURL, g.Tags, g.ReleasePolicy),
 			JSONWebKey: jsonWebKeyFromGenerated(g.Key),
 			ID:         g.Key.Kid,
 			Name:       name,
@@ -222,7 +222,7 @@ func createECKeyResponseFromGenerated(g generated.KeyVaultClientCreateKeyRespons
 	vaultURL, name, version := shared.ParseID(g.Key.Kid)
 	return CreateECKeyResponse{
 		Key: Key{
-			Properties: keyPropertiesFromGenerated(g.Attributes, g.Key.Kid, name, version, g.Managed, vaultURL, g.Tags),
+			Properties: keyPropertiesFromGenerated(g.Attributes, g.Key.Kid, name, version, g.Managed, vaultURL, g.Tags, g.ReleasePolicy),
 			JSONWebKey: jsonWebKeyFromGenerated(g.Key),
 			ID:         g.Key.Kid,
 			Name:       name,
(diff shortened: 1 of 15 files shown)


jhendrixMSFT (approved):

sdk/keyvault/azkeys/client.go
```diff
@@ -760,21 +760,22 @@ func updateKeyPropertiesFromGenerated(g generated.KeyVaultClientUpdateKeyRespons
...
 // UpdateKeyProperties updates the management properties of a key, but not its cryptographic material.
 // Pass nil for options to accept default values.
-func (c *Client) UpdateKeyProperties(ctx context.Context, key Key, options *UpdateKeyPropertiesOptions) (UpdateKeyPropertiesResponse, error) {
-	name, version := "", ""
-	if key.Properties != nil && key.Properties.Name != nil {
-		name = *key.Properties.Name
-	}
-	if key.Properties != nil && key.Properties.Version != nil {
-		version = *key.Properties.Version
+func (c *Client) UpdateKeyProperties(ctx context.Context, properties Properties, options *UpdateKeyPropertiesOptions) (UpdateKeyPropertiesResponse, error) {
```
heaths, line 763: How/where do they update keyOps: https://docs.microsoft.com/en-us/dotnet/api/azure.security.keyvault.keys.keyclient.updatekeyproperties?view=azure-dotnet#azure-security-keyvault-keys-keyclient-updatekeyproperties(azure-security-keyvault-keys-keyproperties-system-collections-generic-ienumerable((azure-security-keyvault-keys-keyoperation))-system-threading-cancellationtoken) ?

heaths (approved):
Not blocking, but am wondering how keyOps are updated. Presumably through the existing UpdateKeyProperties struct?

sdk/keyvault/azkeys/client.go
chlowell: Thanks for pointing this out! Looks like it was never possible, although it could be in the current API because the `Key` type carries the operations. Why did we choose to add another parameter for the operations instead of adding them to the properties object?

heaths: The permissible operations can't be updated in all the same places properties can.

chlowell: Are you thinking of import, which also takes `Properties` but gets the operations from the JWK?

heaths: Partly, but in C# - and some other languages I think as well - our _properties_ object is the same one that hangs off a `KeyVaultKey`. In hindsight, probably over-sharing the type.

chlowell: That's the design here as well. I see we have some additional over-sharing in Go: several options bags have `Properties` fields and therefore have redundant fields and read-only options. I'll flatten these in another PR (tracked by #18324).

PR #18311, opened 2022-06-03
Fail tool installation when writing to stderr

jhendrixMSFT:
Fail the "install coverage and junit dependencies" step if any of the
tools fail to install.
Add a few retries if the step fails.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -41,11 +41,14 @@ steps:
 
   - pwsh: |
       go install github.com/jstemmer/go-junit-report@v0.9.1
+      if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
       go install github.com/axw/gocov/gocov@v1.1.0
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
In response to https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1618275&view=logs&j=d10f4d72-f086-52d5-0c2f-aa1bdc250c55&t=5146fa69-6c9d-5ce6-f380-4f7e6ef3dd0e

weshaggard:
@jhendrixMSFT do you know if that is writing to stderr? We should also consider just checking exit codes and failing on those but it isn't clear how go install treats errors.

jhendrixMSFT:
Most of the tools do, I will look into how this one behaves.

Is there a standard pattern we use for PS to check the exit code and fail?

weshaggard:
```
if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
```

That is the standard way to check exit code in PS.

jhendrixMSFT:
I dug into the impl for `go install` and it doesn't write to `stderr` but does return a non-zero exit code.

weshaggard (approved):

PR #18313, opened 2022-06-03
Prepare azcore v1.1.0 for release

chlowell:


sdk/azcore/internal/shared/constants.go
```diff
@@ -30,5 +30,5 @@ const (
 	Module = "azcore"
 
 	// Version is the semantic version (see http://semver.org) of this module.
-	Version = "v1.0.1"
+	Version = "v1.1.0"
```
jhendrixMSFT, line 33: If this release only updates the MSAL dependency (I didn't see that mentioned in the changelog) then I think it should be a patch update.

chlowell: This is for `azcore`, where the change is removing the floor on polling frequency during tests. Seems like a minor version bump to me because it's new behavior, though it is a very minor change.

jhendrixMSFT: Reading comprehension fail for me today.  Agreed with this.

jhendrixMSFT (approved):

Issue #18276, opened 2022-05-30
Problem for `Dump Package properties`


tadelesh:
This new added step cannot handle some internal tools under `eng/tools` folder. Log is [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1607851&view=logs&j=730d91e9-ccca-5f92-9343-3b1124b19442&t=57014ab0-c321-5206-21d1-de9826f53695).

benbp:
@tadelesh not sure if this issue has been noticed or not? I'm going to add it to our engsys project so it comes past our planning meeting.

PR #18349, opened 2022-06-09
Update release-v2 tooling

tadelesh:
1. Update tooling to support mgmt. plane SDK code filename without 'zz_generated_' prefix.
2. Fix some tool issue (https://github.com/Azure/azure-sdk-for-go/issues/18407)


eng/tools/generator/template/rpName/packageName/build.go.tpl
```diff
@@ -2,6 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 // This file enables 'go generate' to regenerate this specific SDK
-//go:generate pwsh.exe ../../../../eng/scripts/build.ps1 -skipBuild -cleanGenerated -format -tidy -generate resourcemanager/{{rpName}}/{{packageName}}
+//go:generate pwsh.exe ../../../../eng/scripts/build.ps1 -skipBuild -cleanGenerated -format -tidy -generate -filePrefix "" resourcemanager/{{rpName}}/{{packageName}}
```
lirenhe, line 5: Can we have default value for -filePrefix as empty instead of zz_generated_,  then we don't need to change the below files.

tadelesh: For previous build.ps1, the default is `zz_generated_`. I don't quite sure it is safe to change it default to empty. So, I chose to change our build template.

lirenhe: But as 'empty' is the default behaviour, so shall we update our script to make it default?
What do you think of the risk of this change?

tadelesh: I've searched the code base and it seems only mgmt. plane uses build.ps1 to do the code generation. So, maybe it only has impact for our mgmt. SDK, I've changed the default behavior and will track with any possible issue related to it after merge.

lirenhe (approved):

eng/tools/generator/cmd/template/templateCmd.go
```diff
@@ -85,6 +85,7 @@ type Flags struct {
 	ReleaseDate   string
 	PackageConfig string
 	GoVersion     string
+	Version       string
```
ArcturusZhang, line 88: What is this `Version` property used for? Could we have some comment on these properties?

ArcturusZhang: And this new property is not added to the `BindFlags` above, is this an issue?

tadelesh: Changed to PackageVersion.

eng/scripts/build.ps1
```diff
@@ -29,7 +30,7 @@ function Process-Sdk ()
     if ($cleanGenerated)
     {
         Write-Host "##[command]Cleaning auto-generated files in" $currentDirectory
-        Remove-Item "zz_generated_*"
+        Get-ChildItem "*.go" | Select-String -Pattern "Code generated by Microsoft (R) AutoRest Code Generator." | Select-Object -ExcludeProperty path | Remove-Item -Force -Confirm
```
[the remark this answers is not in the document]
tadelesh: Yes, this will delete auto-generated test file. At first, I think it's OK to delete them all. But now I change my mind. It could be clearer to only delete what this script generates, and we have test generation script to deal with the test.

ArcturusZhang (approved):

tadelesh:
Override check-enforcer as this [issue](https://github.com/Azure/azure-sdk-for-go/issues/18276) still not resolved.

PR #18399, opened 2022-06-13
Release v66.0.0

Alancere:


lirenhe:
> > Please hold the merge & release of this PR. @tadelesh, could you check whether we need to add deprecation message for data plane SDKs before releasing it? Thanks
> 
> @lirenhe , merging this PR does not mean release v66.0.0 and also data plane deprecation should be in another PR. The release (tag from `legacy` branch) will be held until data plane confirm.

I see. I removed the label of 'DoNotMerge'.
It would be good to provide this information in the descritpion and the title 'Release v66.0.0.' give me a feeling that we will release this one.    @Alancere .

PR #18400, opened 2022-06-13
[Release] sdk/resourcemanager/eventgrid/armeventgrid/2.0.0

Alancere:
https://github.com/Azure/sdk-release-request/issues/2911


tadelesh:
@jhendrixMSFT this breaking is from adding the `x-ms-enum` ([previous](https://github.com/Azure/azure-rest-api-specs/blob/c767823fdfd9d5e96bad245e3ea4d14d94a716bb/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2021-12-01/EventGrid.json), [now](https://github.com/Azure/azure-rest-api-specs/blob/f88928d723133dc392e3297e6d61b7f6d10501fd/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2022-06-15/EventGrid.json#L10763)). It seems such kind of naming change (including definition naming change or enum naming change) will always cause breaking changes.

jhendrixMSFT:
In this case, the service team explicitly changed the name of the enum with the `x-ms-enum:name`.  If the breaking change was approved then there's nothing we can do.

RickWinter (dismissed):
cc: @JeffreyRichter can you weigh in on this.

JeffreyRichter:
I'm not sure what I'm being asked to weigh in on. This is a breaking change. I don't recall this one so I don't think the Breaking Change Review Board approved this.

ArcturusZhang:
The corresponding swagger PR might be reviewed by me - they are adding an `x-ms-enum` to an enum that previously does not have this extension, and this causes this breaking change.
This kind of breaking changes might not be caught by the current "cross-version breaking change pipeline".
I will have a double confirm on the information above

tadelesh:
I confirmed with swagger team that this `x-ms-enum` change will not be a breaking change as its implementation depends on SDK.

JeffreyRichter:
Yes, that's true but a team can't change swagger if that will break sdks.

tadelesh:
> @tadelesh I'm not sure that's true. If the value of `x-ms-enum.name` doesn't match the property name then it will be a breaking change for sure. Even if it does match, I believe it will still be a breaking change as M4 synthesizes the enum's type name based on the property name (I don't think it uses it verbatim but would be good to confirm).

@jhendrixMSFT We all think it is a breaking. But unfortunately, swagger breaking detection does not report it. We are trying to convince swagger team to add some breaking detection rule for the `x-ms-enum` change.

tadelesh:
Got breaking approval from [here](https://github.com/Azure/azure-rest-api-specs/pull/19386).

tadelesh (approved):

PR #18413, opened 2022-06-15
[azservicebus] Adding in raw AMQP support for sending and receiving messages.

### LINKED ISSUE #15260: [azservicebus] Support "rawAnnotatedMessage" (AMQP message)
The recent Service Bus SDKs have made it easier to both access and send messages that are a 1:1 translation of an AMQP message. This includes being able to encode the body to separate sections (data, value, or sequence).

Using an AMQPAnnotatedMessage (type name TBD in Go) is supported in these methods in the Sender:

- Sender.SendMessage()
- Sender.ServiceBusMessageBatch.TryAdd()
- Sender.ScheduleMessage

And each message received by the receiver contains an AMQPAnnotatedMessage in a field in the ServiceBusReceivedMessage itself.

In languages that have the ability updating fields in the annotated message will be reflected in the outer message - this is easy to do in languages like C# and Java, where properties are backed by code. 

In Go, this is unlikely to be the case but should be discussed.

TODO:
- [ ] Validate the list of fields we export against what we've exported in other languages.
- [ ] Do we need settlement methods specifically for AMQP messages? Or can we require that the ReceivedMessage instance is kept around?
- [ ] Do we need AMQPBodyType? It might be considered redundant.

richardpark-msft:
For sending AMQP messages you can use either Sender.SendAMQPMessages() or schedule them using Sender.ScheduleAMQPMessages().
    
For receiving, each ReceivedMessage now has a RawAMQPMessage property.

I've also added in a way to export the TLS pre-master key so you can easily wireshark and trace the actual AMQP interaction, which can be helpful in validating that things are encoding properly.

Fixes #15260

### DIFF
--- a/sdk/messaging/azservicebus/amqp_message.go
+++ b/sdk/messaging/azservicebus/amqp_message.go
@@ -29,7 +29,7 @@ type AMQPMessage struct {
 	ApplicationProperties map[string]any
 
 	// Body represents the body of an AMQP message.
-	Body AMQPBody
+	Body AMQPMessageBody
 
 	// DeliveryAnnotations corresponds to the "delivery-annotations" section in an AMQP message.
 	//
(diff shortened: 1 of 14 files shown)


sdk/messaging/azservicebus/amqp_message.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// MessageAnnotations corresponds to the message-annotations section of an AMQP message.
+	//
+	// The values of the map are restricted to AMQP simple types, as listed in the comment for AMQPMessage.
+	MessageAnnotations map[any]any
+
+	// Properties corresponds to the properties section of an AMQP message.
+	Properties *AMQPMessageProperties
+
+	linkName string
```
jhendrixMSFT, line 59: While not unique to this change, having unexported fields on a POD type means customers can't create a fake `AMQPMessage` for testing purposes.

jhendrixMSFT: I suspect there's a fair amount of work that needs to be done in order to support mocks.  Also, I don't know what our actual testing strategy and recommendation is for `azservicebus`; this was just a general observation.  Feel free to open a tracking issue to address it later.

richardpark-msft: @jhendrixMSFT: Just to check I'm getting what you mean - customers can create azservicebus.AMQPMessage{}, they just won't be able to to initialize those unexported fields. That's actually okay. They're only used when you actually call an operation (like Receiver.CompleteMessage) where they would have to be filled in.

For just mocking (ie, I'm passing an azservicebus.AMQPMessage around) it should be fine.

jhendrixMSFT: Good point, I think I overgeneralized a bit (whenever I see exported POD types that mix un/exported fields my inner alarm goes off).  Agreed that passing around a mock version of this to/from mock APIs would be fine.  And it looks like there aren't any public APIs on this type that depend on the unexported fields so it should be fine.

sdk/messaging/azservicebus/amqp_message.go
```diff
@@ -0,0 +1,279 @@
...
+			Priority:      goAMQPMessage.Header.Priority,
+			TTL:           goAMQPMessage.Header.TTL,
+		}
+	}
+
+	var properties *AMQPMessageProperties
+
+	if goAMQPMessage.Properties != nil {
+		properties = &AMQPMessageProperties{
+			AbsoluteExpiryTime: goAMQPMessage.Properties.AbsoluteExpiryTime,
```
jhendrixMSFT, line 241: If `amqp.MessageProperties` and `AMQPMessageProperties` have the exact same layout you can cast one to the other and let the compiler do the copying for you.

sdk/messaging/azservicebus/sender.go
```diff
@@ -96,14 +100,42 @@ type ScheduleMessagesOptions struct {
...
+// ScheduleAMQPMessagesOptions contains optional parameters for the ScheduleAMQPMessages function.
+type ScheduleAMQPMessagesOptions struct {
+	// For future expansion
+}
+
+// ScheduleAMQPMessages schedules a slice of Messages to appear on Service Bus Queue/Subscription at a later time.
+// Returns the sequence numbers of the messages that were scheduled.  Messages that haven't been
+// delivered can be cancelled using `Receiver.CancelScheduleMessage(s)`
+// If the operation fails it can return an *azservicebus.Error type if the failure is actionable.
+func (s *Sender) ScheduleAMQPMessages(ctx context.Context, messages []*AMQPMessage, scheduledEnqueueTime time.Time, options *ScheduleAMQPMessagesOptions) ([]int64, error) {
```
jhendrixMSFT, line 115: Presumably other languages use overloads to avoid introducing additional methods.  What made you chose this over having `ScheduleMessage()` take an interface instead?

jhendrixMSFT: Technically that would be a breaking change, so we're stuck.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/amqp_message.go
richardpark-msft: We'll need to do that separately. They have the same content, but their field layout is in a different order which appears to be enough to tank the auto-copying.

I don't think we did that on purpose in go-amqp, so we can patch it there and then vendor that change back into here.

PR #18425, opened 2022-06-16
Fix 18424: aztables incorrectly disallows empty rowkey

### LINKED ISSUE #18424: aztables incorrectly disallows empty rowkey resulting in unreadable data
### Bug Report

- SDK version: `github.com/Azure/azure-sdk-for-go/sdk/data/aztables@v1.0.0`
- output of `go version`
`go version go1.18.3 darwin/arm64`

The `aztables` SDK and previous Azure Table Storage SDKs allow insertion of entities with empty row key - only a partition key is required. However, the `aztables` SDK (including v1.0.0) explicitly throws an error when performing a Get or Update requests on entities with empty row key. This is a breaking change from the previous SDK and renders data written with the previous SDK not readable. Furthermore, this is inconsistent with this SDK itself as it is still possible to create entities with empty row key values.

Here is what the old SDK did in the Get path:
https://github.com/Azure/azure-sdk-for-go/blob/1cdb007c833eea301e1866a02e5e6d41240f3ec4/storage/entity.go#L226

And here is what the new "Track 2" SDK is doing which is incorrect:
https://github.com/Azure/azure-sdk-for-go/blob/1cdb007c833eea301e1866a02e5e6d41240f3ec4/sdk/data/aztables/internal/table_client.go#L727

berndverst:
Signed-off-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

Fixes #18424 

### DIFF
--- a/sdk/data/aztables/internal/table_client.go
+++ b/sdk/data/aztables/internal/table_client.go
@@ -12,14 +12,15 @@ import (
 	"context"
 	"encoding/xml"
 	"errors"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
I made some updates to the PR so the hand-edits to the generated code aren't lost.

I think [this](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/client.go#L471-L472) also needs to be fixed yes?  If `rowKey` is missing the type-assertion will panic.

PR #18446, opened 2022-06-21
Remove azsecrets convenience API

chlowell:
Replacing the handwritten API with a generated one, slightly modified (see autorest.md for the transformations). [APIView](https://apiviewuat.azurewebsites.net/Assemblies/Review/93269ff4353a4e63b880bc1a13796988) gives a better view of the new surface than the diff here.


sdk/keyvault/azsecrets/autorest.md
```diff
@@ -2,29 +2,79 @@
...
+      - response_types.go
+    where: $
+    transform: return $.replace(/Client(\w+)((?:Options|Response))/g, "$1$2");
+
+  # make secret IDs a convenience type so we can add parsing methods
+  - from: models.go
+    where: $
+    transform: return $.replace(/(\sID \*)string(\s+.*)/g, "$1ID$2")
+
+  # Maxresults -> MaxResults
```
chlowell, line 69: This one is a good candidate to become part of autorest.go's default behavior

sdk/keyvault/azsecrets/autorest.md
```diff
@@ -2,29 +2,79 @@
...
+      from: SecretRestoreParameters
+      to: RestoreSecretParameters
+  - rename-model:
+      from: SecretSetParameters
+      to: SetSecretParameters
+  - rename-model:
+      from: SecretUpdateParameters
+      to: UpdateSecretParameters
+
+  # delete the Attributes model defined in common.json (it's used only with allOf)
```
chlowell, line 38: That is to say, the spec defines this model only so other models can inherit its fields. No operation uses `Attributes`. I imagine we could make the generator recognize that and not output the model. @jhendrixMSFT would that be a task for autorest.go? I imagine we wouldn't want that kind of filtering in autorest core because languages with inheritance may want to use that to model the allOf relationship.

JeffreyRichter (approved):

sdk/keyvault/azsecrets/custom_client.go
```diff
@@ -0,0 +1,39 @@
...
+	authPolicy := internal.NewKeyVaultChallengePolicy(credential)
+	pl := runtime.NewPipeline(moduleName, version, runtime.PipelineOptions{PerRetry: []policy.Policy{authPolicy}}, options)
+	return &Client{endpoint: vaultURL, pl: pl}
+}
+
+// ID is a secret's unique ID, containing its name and version.
+type ID string
+
+// Name of the secret.
+func (i ID) Name() string {
```
jhendrixMSFT, line 30: Since `ID` is returned by pointer, this will dereference `i` and could panic if it's ever `nil`.  It might be worth changing the receivers to byref and just return the empty string if `i` is `nil`.

jhendrixMSFT: I don't feel too strong about this though, and you could make the argument that `ID` is never `nil` so if you do get a panic then you likely have a bug in your code (I'm somewhat guessing on this latter part, let me know if you disagree).

JeffreyRichter: I'm OK with panics for developer bugs. But since it is returned by pointer, these probably should be pointers too.

chlowell: Thanks for pointing this out, I'll make them pointer receivers. I want to keep the value return because every secret has an ID and Key Vault always includes it in responses so, without a bug we would want to know about or some other extraordinary circumstance, `ID` won't be `nil`.

jhendrixMSFT (approved):

jhendrixMSFT (approved):

PR #18469, opened 2022-06-23
Remove create-go-workspace step from track 2 pipelines

benbp:
I think these are unnecessary and add unnecessary time to the pipeline.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client-samples.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client-samples.yml
@@ -36,8 +36,6 @@ stages:
           version: '$(go.version)'
         displayName: "Select Go Version"
 
-      - template: ../steps/create-go-workspace.yml
-
       - template: ../steps/build-test.yml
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client-samples.yml
```diff
@@ -36,8 +36,6 @@ stages:
           version: '$(go.version)'
         displayName: "Select Go Version"
 
-      - template: ../steps/create-go-workspace.yml
```
weshaggard, line 39: Should we delete this template as well?

weshaggard (commented):
I don't have enough context to approve this or not you should try to get someone from the go team to sign-off on it.

eng/pipelines/templates/jobs/archetype-sdk-client-samples.yml
benbp: It's still used by a cosmos pipeline that I think needs it...

PR #18474, opened 2022-06-24
[Release] sdk/resourcemanager/compute/armcompute/3.0.0

Alancere:
https://github.com/Azure/sdk-release-request/issues/2945


tadelesh:
Breaking change from known swagger change and codegen change.

tadelesh (approved):

lirenhe:
> Breaking change from known swagger change and codegen change.

Do you mean part of breaking is the result from Codegen change?  Could you share more details and is this the area that we should avoid?

tadelesh:
> > Breaking change from known swagger change and codegen change.
> 
> Do you mean part of breaking is the result from Codegen change? Could you share more details and is this the area that we should avoid?

We discussed this change last week and the final decision is not allow optional body param in PATCH/PUT. For now, we will always set the body param of PATCH/PUT operation as required in the codegen after this [change](https://github.com/Azure/autorest.go/pull/839). There are many existed issues in swagger which will cause breaking.

PR #18477, opened 2022-06-24
[azservicebus] Allow link creation to be cancelled (#17598)

franklinkim:
Go routine is never closed if the ctx is being cancelled since there is no receiver for the done channel.


richardpark-msft:
Hi @franklinkim, you've definitely found a bug! 

Looking at it more deeply there's also a race condition. I'll need to do a bit more changes in a new PR but thank you for finding this bug, even if we're not taking this particular fix.

franklinkim:
Hi @richardpark-msft,

no problem, happy to help out.
please keep me (this pr) up to date cause it triggered some unexpected failures in our infrastructure.

Thanks

PR #18495, opened 2022-06-29
Skip dumping package properties for eng directories

### LINKED ISSUE #18276: Problem for `Dump Package properties`

benbp:
Fixes #18276

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-eng-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-eng-client.yml
@@ -1,0 +1,85 @@
+parameters:
+  - name: ServiceDirectory
+    type: string
+    default: ''
+  - name: NonShipping
+    type: boolean
+    default: false
+  - name: LicenseCheck
+    type: boolean
+    default: true
+  - name: EnvVars
+    type: object
+    default: {}
+  - name: TestRunTime
+    type: string
+    default: '600s'
+
+
+stages:
+  - stage: Build
+    jobs:
+    - job: Test
+      displayName: Build and Test on
+      variables:
+        - template: /eng/pipelines/templates/variables/globals.yml
+      strategy:
+        matrix:
+          Linux_Go118:
+            pool.name: azsdk-pool-mms-ubuntu-2004-general
+            image.name: MMSUbuntu20.04
+            go.version: '1.18.2'
+          Windows_Go118:
+            pool.name: azsdk-pool-mms-win-2019-general
+            image.name: MMS2019
+            go.version: '1.18.2'
+            generate.bom: true
+      pool:
+        name: $(pool.name)
+        vmImage: $(image.name)
+      steps:
+      - template: /eng/common/pipelines/templates/steps/verify-agent-os.yml
+        parameters:
+          AgentImage: $(image.name)
+
+      - task: GoTool@0
+        inputs:
+          version: '$(go.version)'
+        displayName: "Select Go Version"
+
+      - template: ../steps/create-go-workspace.yml
+
+      - template: ../steps/build-test.yml
+        parameters:
+          ServiceDirectory: ${{ parameters.ServiceDirectory }}
+          Image: $(vm.image)
+          GoVersion: $(go.version)
+          TestProxy: true
+          TestRunTime: ${{ parameters.TestRunTime }}
+          EnvVars:
+            AZURE_RECORD_MODE: 'playback'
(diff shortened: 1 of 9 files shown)


benbp:
FYI @tadelesh

benbp:
@tadelesh if it looks good, can you approve when you get the chance? Thanks!

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -16,7 +16,7 @@ steps:
       pwsh: true
       workingDirectory: $(Pipeline.Workspace)
     displayName: Dump Package properties
-    condition: succeeded()
+    condition: and(succeeded(), not(contains('${{ parameters.ServiceDirectory }}', '/eng/')))
```
weshaggard, line 19: I don't know how this fails but perhaps we just make sure our script can handle service directories that don't have any packages.

benbp: It fails at https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/Save-Package-Properties.ps1#L124 when it returns empty package properties. So we could move the exclusion for the `eng` directory into the script itself, but it seems like it's intended behavior that it would fail in other situations vs. passing silently. I assume in most/all places where this script is used there are downstream steps that rely on the package properties not being empty.

It would be cleaner in the script from the go pipeline configs, but at the same time it is putting an exception for a single language into a core component, so ¯\_(ツ)_/¯ Thoughts?

weshaggard: That is fair. I wouldn't add an "eng" special casing inside of the script I was just wondering if there are assumptions built in that would break if we just let an empty list come back. 

Do you like adding a "/eng/" explicit condition here or just a "Skip" variable that we could set in those eng pipelines?

benbp: I think Skip is probably cleaner, but we also don't have a config driven way to set that up, so it could make new eng tool pipelines break in the future. I'm not the biggest fan of this current `/eng/` skip condition, so maybe it's worth the trade-off?

weshaggard: I was thinking of some skip variable would make this more opt-in to the exception cases which I like better then adding this special cases. If things under /eng/ is special enough maybe we create a template with this variable setup for those pipelines.

benbp: @weshaggard I added a separate pipeline. I think that will be the most flexible for handling similar situations that come up in the future.

tadelesh (approved):
LGTM

tadelesh:
@benbp Sorry for that I did not recognize the `auto-merge` label and it seems this PR auto merged.

benbp:
@tadelesh that's my bad. I had it on auto-merge when the change was simpler, and forgot to remove it. @weshaggard let me know if you want me to revert anything :)

eng/pipelines/templates/jobs/archetype-sdk-eng-client.yml
```diff
@@ -0,0 +1,85 @@
+parameters:
```
weshaggard, line 1: I suspect we can just name this eng.yml or something similar as the client in the name isn't correct for these now.

While I agree we probably do want a separate template I worry a little about the duplication.

eng/pipelines/templates/jobs/archetype-sdk-eng-client.yml
```diff
@@ -0,0 +1,85 @@
...
+  - name: ServiceDirectory
+    type: string
+    default: ''
+  - name: NonShipping
```
weshaggard, line 5: Isn't this now implied?

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -4,6 +4,7 @@ parameters:
   LintVersion: ''
   NonShipping: false
   LicenseCheck: true
+  IsSdkLibrary: true
```
weshaggard, line 7: Should IsSdkLibrary and NonShipping be combined? They seem like they are representing something similar.

PR #18519, opened 2022-07-02
[azblob] Enable errors.Is checking on wrapped InternalError error

wkalt:
This changes the implementation of the Is method of azblob.InternalError
to check the target error against the wrapped error value instead of the
InternalError type itself, if the target type is not itself
InternalError. Prior to this commit, type information about the wrapped
error was not accessible. The motivation for this change is to allow the
caller to check a returned InternalError to see if it is wrapping a
context cancelation.

For instance, given a BlockBlobClient, one may like to call

    _, err := client.Download(ctx, nil)
    if err != nil {
      if errors.Is(err, context.Canceled) {
        return // all good
      }
      // handle error
    }

and be able to recognize that the failure was due to context
cancelation.

Existing logic around checking if an InternalError type assertion
succeeds is left in place to account for compatibility issues. New logic
is in place only if that check fails.

### DIFF
--- a/sdk/storage/azblob/zc_storage_error_test.go
+++ b/sdk/storage/azblob/zc_storage_error_test.go
@@ -7,7 +7,9 @@
 package azblob
 
 import (
+	"context"
 	"encoding/xml"
+	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


sdk/storage/azblob/zc_storage_error_test.go
```diff
@@ -38,3 +40,38 @@ func (s *azblobUnrecordedTestSuite) TestErrorResponseUnmarshal() {
...
+	t := s.T()
+
+	cases := []struct {
+		name   string
+		err    *InternalError
+		target error
+		result bool
+	}{
+		{
+			"errors.Is reports true if checked against an internal error",
```
wkalt, line 54: This was the existing behavior, just captured in a test now.

I'm not sure if this behavior is actually desirable but I don't know what the compatibility concerns may be.

jhendrixMSFT:
We are refactoring the `azblob` module, and as a part of that work, the `InternalError` type is going away.

PR #18530, opened 2022-07-05
[azservicebus] Updating to handle disposition hang fix

richardpark-msft:
Applying patch from https://github.com/Azure/go-amqp/pull/171

The disposition functions, which we use for message settlement, can hang if the link is closed. The issue here is message disposition's initial send is done through the session and connection, which means a dead link won't be detected. 

When this happens we end up blocking on a channel that won't be closed since the disposition response will never come back.


jhendrixMSFT (approved):

mgevantmakher:
Hello Richard, would you tag a release, this bug you cloud is critical for us, we hit this several times already

PR #18532, opened 2022-07-06
[azservicebus] Updating to go-amqp:ca689527ff4c98693075d2fbc0ef98ef415fd08b

richardpark-msft:
Absorbing go-amqp@ca689527ff4c98693075d2fbc0ef98ef415fd08b

This contains a bug fix for settlement hanging if the receiver had been closed. This showed up in our constant-detach stress tests - we'd receive messages but then hang for an indeterminate time while trying to complete messages. The problem (from what I could trace from the log) was that the go-amqp Receiver didn't check the local link state before attempting to disposition a message. This would result in us waiting infinitely for a disposition response that would never happen.

The other changes are primarily just go-amqp transitioning from variadic config to plain old struct config.


richardpark-msft:
(stress test launched as ripark-preview-disp-fix)

PR #18582, opened 2022-07-13
Added SignerForKey to azkeys.Client

ItalyPaleAle:
This change adds a new method `SignerForKey` to `azkeys.Client`, handwritten.

This method allows creating an object that implements the `crypto.Signer` interface. It allows using keys stored in Azure Key Vault to create digital signatures in Go applications anywhere where `crypto.Signer` is supported. It makes the SDK essentially a "plug-in" for a lot more systems.

Example:

```go
if err != nil {
	panic(err)
}

client := azkeys.NewClient("https://<vault name>.vault.azure.net", cred, nil)
signer := client.SignerForKey(context.Background(), "<key name>", "")

msg := sha256.Sum256([]byte("hello world"))
res, err := signer.Sign(nil, msg[:], crypto.SHA256)
fmt.Println(res, err)
```

### DIFF
--- a/sdk/keyvault/azkeys/constants.go
+++ b/sdk/keyvault/azkeys/constants.go
@@ -251,6 +251,26 @@ func PossibleJSONWebKeyTypeValues() []JSONWebKeyType {
 	}
 }
 
+// IsRSAKey returns true if the key is an RSA key (RSA or RSA-HSM).
+func (kt JSONWebKeyType) IsRSAKey() bool {
+	return kt == JSONWebKeyTypeRSA || kt == JSONWebKeyTypeRSAHSM
+}
+
+// IsECKey returns true if the key is an EC key (EC or EC-HSM).
+func (kt JSONWebKeyType) IsECKey() bool {
+	return kt == JSONWebKeyTypeEC || kt == JSONWebKeyTypeECHSM
+}
+
+// IsSymmetricKey returns true if the key is a symmetric key (Oct or OctHSM).
+func (kt JSONWebKeyType) IsSymmetricKey() bool {
+	return kt == JSONWebKeyTypeOct || kt == JSONWebKeyTypeOctHSM
+}
+
+// IsAsymmetric returns true if the key type is asymmetric.
+func (kt JSONWebKeyType) IsAsymmetric() bool {
+	return kt.IsECKey() || kt.IsRSAKey()
+}
+
 // KeyEncryptionAlgorithm - The encryption algorithm to use to protected the exported key material
 type KeyEncryptionAlgorithm string
 
--- a/sdk/keyvault/azkeys/custom_client.go
+++ b/sdk/keyvault/azkeys/custom_client.go
@@ -109,12 +109,9 @@ func (signer *signer) Sign(_ io.Reader, value []byte, opts crypto.SignerOpts) ([
 		// If we have an opts object and that is of type *rsa.PSSOptions, then use RSA-PSS
 		// (however, we are ignoring the SaltLength property of the object as it's not supported by Key Vault).
 		// Otherwise, use RSA in PKCS #1 v1.5 mode.
-		if opts != nil {
-			if _, ok := opts.(*rsa.PSSOptions); ok {
-				signAlg = "PS"
-			}
-		}
-		if signAlg == "" {
+		if _, ok := opts.(*rsa.PSSOptions); ok {
+			signAlg = "PS"
+		} else {
 			signAlg = "RS"
 		}
 	case *ecdsa.PublicKey:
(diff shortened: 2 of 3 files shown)


chlowell:
I like this idea, thanks for implementing it! My main concern about the API is that `Sign` and `Public` send HTTP requests but can't have a `Context` parameter as required by the [SDK design guidelines](https://azure.github.io/azure-sdk/golang_introduction.html#cancellation). The intent of the `Context` guideline is to establish a consistent, idiomatic cancellation pattern across the SDK: methods which perform I/O are cancelled via `Context`; methods which don't take a `Context` don't perform I/O. This new API reverses the pattern, and applies cancellation to multiple operations, including future ones.

What do you think of turning this into a sample?

ItalyPaleAle:
Thanks for the review @chlowell!

Sadly, the `crypto.Signer` interface does not support passing a context and the goal was to be compliant with that. This way you can pass an `azkeys.Signer` object just like you'd pass a `crypto.PrivateKey` around methods in the standard library that accept a `crypto.Signer`. Scenarios include things such as signing TLS certificates.

I thought of passing the context to the `SignerForKey` method as a workaround, although I agree that it's not ideal. If you do have a better idea, I'd be more than happy to change the implementation.

As for making this a code sample... It does seem like it's a lot of code to make it a sample IMHO, to the point where perhaps it could be its own library.

At the very least, would you be open to merging the changes that are in the [`custom_models.go` file](https://github.com/Azure/azure-sdk-for-go/pull/18582/files#diff-72ffce2bbd684efb8a69abab9f26614350d63ceb140edc875bbfb321da9551f6)? Those are utility methods that can be used for a variety of purposes.

chlowell:
> I thought of passing the context to the `SignerForKey` method as a workaround, although I agree that it's not ideal. If you do have a better idea, I'd be more than happy to change the implementation.
> 

I haven't come up with one. I suppose we need a `crypto.SignerWithContext` interface to make this compatible with our `Context` requirements, but I can imagine why the standard library doesn't have such a thing yet--remote crypto isn't the norm.

> As for making this a code sample... It does seem like it's a lot of code to make it a sample IMHO, to the point where perhaps it could be its own library.
> 

I agree it's perhaps too big for a doc example. I was imagining it as part of the [Azure Samples collection](https://docs.microsoft.com/en-us/samples/browse/?products=azure&languages=go). I suppose it could work as a separate library as well, though that might be less discoverable.

> At the very least, would you be open to merging the changes that are in the [`custom_models.go` file](https://github.com/Azure/azure-sdk-for-go/pull/18582/files#diff-72ffce2bbd684efb8a69abab9f26614350d63ceb140edc875bbfb321da9551f6)? Those are utility methods that can be used for a variety of purposes.

We can consider the `Public()` helper (`Private()` would always return nil for Key Vault keys). We want to keep the current API stable until the 1.0 release, so it would have to wait on that. No problem adding it to the docs any time though.

ItalyPaleAle:
> I suppose we need a `crypto.SignerWithContext` interface to make this compatible with our `Context` requirements, but I can imagine why the standard library doesn't have such a thing yet--remote crypto isn't the norm.

Yeah. Looks like there's an open issue in the Golang repo, that's been open for years and still undecided: https://github.com/golang/go/issues/28427 Possibly scheduled for Go2 - whenever that will be.

> I was imagining it as part of the [Azure Samples collection](https://docs.microsoft.com/en-us/samples/browse/?products=azure&languages=go).

That could be an option actually, didn't think of that.

> We can consider the Public() helper (Private() would always return nil for Key Vault keys). We want to keep the current API stable until the 1.0 release, so it would have to wait on that.

Yes, you're right that `Private()` would always return nil because the keys are in the vault. At least having the `Public()` method could be of use, for a lot of scenarios IMHO. Do you want me to open a separate PR for that part?

chlowell:
> At least having the `Public()` method could be of use, for a lot of scenarios IMHO. Do you want me to open a separate PR for that part?

Sure. If you're thinking of the docs, I believe a `func ExampleJSONWebKey_public()` in example_test.go would land in the right place, and we can merge that any time. I still want to hold public API changes until after the first stable release though.

ItalyPaleAle:
I will wait for the stable release then :)

chlowell:
This appeared on our list of stale PRs today, and we discussed your `Public()` method. We want to learn more about how, and how often, customers would use it before we decide whether to add it to the public API. Also, we may add a local crypto implementation to `azkeys` in the future and don't want to add similar API prematurely.

However, `Public()` would fit well in the documentation as an example. I added an issue on the backlog to track that (#19140). Feel free to open another PR to address that issue. I'm closing this one to shorten the stale PR list. Thanks again for opening it--this is an interesting integration we hadn't considered.

PR #18583, opened 2022-07-13
Fix broken pkg.go.dev link in azkeys README

hiddeco:
This adds back the `goget_azkeys` link reference which seems to have
been accidentally deleted in https://github.com/Azure/azure-sdk-for-go/commit/32be770a92b5199ac7fee8340344d1b7520b8a46,
making the rendered Markdown link clickable again.


chlowell (approved):
Thanks for fixing this!

PR #18657, opened 2022-07-20
[azservicebus] ReceiveMessages() algorithm change to use amqpReceiver.Release() instead of flow+drain

richardpark-msft:
Changing Receiver.ReceiveMessages() from the existing implementation, which uses AMQP's flow+drain to doing amqp.Release() on messages that arrive when there is not an active listener.
    
It seems particularly hard for our AMQP stack to properly recover if we drain and then have to cancel, which can happen in some cases where the drain ACK takes a longer time to arrive. Switching over to simply releasing messages when there's not an active ReceiveMessages() call allows us to avoid waiting for the ACK (faster exit) as well as allowing us to keep the link alive for longer, which also avoids some scenarios where the link would get closed.

This change also modifies some stress tests that help demonstrate this problem and show that it now works properly.

### DIFF
--- a/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
+++ b/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
@@ -1,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package amqpwrap
+
+
--- a/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
+++ b/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
@@ -1,4 +1,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-package amqpwrap
--- a/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
+++ b/sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
@@ -2,5 +2,3 @@
 // Licensed under the MIT License.
 
 package amqpwrap
-
-
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
@@ -59,6 +60,8 @@ type Receiver struct {
 
 	defaultDrainTimeout      time.Duration
 	defaultTimeAfterFirstMsg time.Duration
+
+	cancelReleaser *atomic.Value
 }
 
 // ReceiverOptions contains options for the `Client.NewReceiverForQueue` or `Client.NewReceiverForSubscription`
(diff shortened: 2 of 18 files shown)


richardpark-msft:
@gracewilcox , adding you in case you're interested in seeing some of what's happening in azservicebus.

sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package amqpwrap
```
jhendrixMSFT, line 4: Empty file?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -357,141 +372,69 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
+	creditsToIssue := int64(maxMessages) - currentReceiverCredits
+	log.Writef(EventReceiver, "Asking for %d credits", maxMessages)
 
-			// there's no material difference here, so let them fail on the next call to ReceiveMessages() instead.
-			log.Writef(EventReceiver, "Non-fatal error occurred and no messages were received. Will recover on next call: %v", lastErr)
-			return nil, nil
-		}
-	}
+	if creditsToIssue > 0 {
+		log.Writef(EventReceiver, "Only need to issue %d additional credits", maxMessages)
```
jhendrixMSFT, line 383: Was this meant to log `creditsToIssue`?

sdk/messaging/azservicebus/internal/amqpwrap/amqpwrap_test.go
richardpark-msft: Will remove (ended up testing this in a live scenario instead)

richardpark-msft: Deleted.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Fixed, good spot!

sdk/messaging/azservicebus/receiver.go
```diff
@@ -59,6 +60,8 @@ type Receiver struct {
 
 	defaultDrainTimeout      time.Duration
 	defaultTimeAfterFirstMsg time.Duration
+
+	cancelReleaser *atomic.Value
```
jhendrixMSFT, line 64: Why is this an `atomic.Value`?  Not saying it shouldn't be, but I'd like to understand the reasoning.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -357,141 +372,69 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
-				// return the cancellation error.
-				err = nil
-			}
-
-			return flushAndReturn(err)
-		}
+	result := r.fetchMessages(ctx, linksWithID.Receiver, maxMessages, r.defaultTimeAfterFirstMsg)
 
-		return flushAndReturn(ctx.Err())
+	if result.Messages != nil {
```
jhendrixMSFT, line 394: You don't want it to log if the message count is zero?

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Close() and Receive() don't have to happen in the same thread.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: That's a good point, will fix.

sdk/messaging/azservicebus/receiver.go
jhendrixMSFT: I came to the same conclusion

jhendrixMSFT (approved):

Issue #18620, opened 2022-07-19
Subscriptions List should include all the fields available in the API


hermanschaaf:
### Feature Request

The subscriptions [Model struct](https://github.com/Azure/azure-sdk-for-go/blob/v61.6.0/services/subscription/mgmt/2020-09-01/subscription/models.go#L293-L308) includes a number of fields, but not all the fields available in the REST API according to the documentation. Specifically, the following fields are missing in the Go struct:

- `tenantId` - The subscription tenant ID.
- `tags` - The tags attached to the subscription.
- `managedByTenants` - An array containing the tenants managing the subscription.

We are particularly interested in the tenant ID in order to automatically determine the tenant for a given subscription.

hermanschaaf:
After digging a bit more, I can see that this is because the fields are not part of the OpenAPI definition here: https://github.com/Azure/azure-rest-api-specs/blob/0c21498807897f2d523f3b63166a75e278ca0c64/specification/subscription/resource-manager/Microsoft.Subscription/stable/2016-06-01/subscriptions.json#L210

However, these fields are returned by the API and are documented here: https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list?tabs=HTTP#subscription

Would the fix be to add the fields to the REST API definition? Would a PR to that effect be accepted?

RickWinter:
The package you are using is part of the previous versions.  The most recent version is available here:
https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/subscription/armsubscription

hermanschaaf:
@RickWinter Thanks, I wasn't aware that I should be using the more recent version there. 

Having checked that implementation, it seems like the new version is also missing subscription fields from the API like `tenantId` and `tags`: https://github.com/Azure/azure-sdk-for-go/blob/77190ec7d546094c51ff1b61073dd9580394a579/sdk/resourcemanager/subscription/armsubscription/zz_generated_models.go#L445-L465 Is there a way to map subscriptions to their tenants in either the new or the old SDK versions?

navba-MSFT:
@hermanschaaf Thanks for reaching out to us and reporting this issue. We are looking into this issue and we will provide an update. I have created the above PR and it is currently under review.

hermanschaaf:
Hi @navba-MSFT, I see this issue has the `needs-author-feedback` tag attached to it--is there something I can help with at this stage?

navba-MSFT:
[...]

PR #18666, opened 2022-07-21
(ARMSubscription) fixes Azure/azure-sdk-for-go#18620

navba-MSFT:
fixes Azure/azure-sdk-for-go#18620

The subscriptions Model struct includes a number of fields, but not all the fields available in the REST API according to the documentation. Specifically, the following fields are missing in the Go struct:

tenantId - The subscription tenant ID.
tags - The tags attached to the subscription.

Note that these fields are returned by the API and are documented here: https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/list?tabs=HTTP#subscription

### DIFF
--- a/sdk/resourcemanager/subscription/armsubscription/zz_generated_models.go
+++ b/sdk/resourcemanager/subscription/armsubscription/zz_generated_models.go
@@ -462,6 +462,12 @@ type Subscription struct {
 
 	// READ-ONLY; The subscription ID.
 	SubscriptionID *string `json:"subscriptionId,omitempty" azure:"ro"`
+	
+	// READ-ONLY; The Tenant ID.
+        SubscriptionTenantID *string `json:"subscriptionTenantId,omitempty" azure:"ro"`
+	
+	// READ-ONLY; The Tags.
+        Tags map[string]*string `json:"tags,omitempty" azure:"ro"`
 }
 
 // SubscriptionsClientGetOptions contains the optional parameters for the SubscriptionsClient.Get method.


hermanschaaf:
@navba-MSFT Thank you for looking into this issue :bow: 

I'm not super familiar with this repo, but as far as I can tell, the fix will need to be in the [underlying OpenAPI JSON spec](https://github.com/Azure/azure-rest-api-specs/blob/0c21498807897f2d523f3b63166a75e278ca0c64/specification/subscription/resource-manager/Microsoft.Subscription/stable/2016-06-01/subscriptions.json#L210) first, and then the code here should be regerenated. As it stands, it looks like this PR's change will be lost the next time the code is regenerated.

navba-MSFT:
> @navba-MSFT Thank you for looking into this issue 🙇
> 
> I'm not super familiar with this repo, but as far as I can tell, the fix will need to be in the [underlying OpenAPI JSON spec](https://github.com/Azure/azure-rest-api-specs/blob/0c21498807897f2d523f3b63166a75e278ca0c64/specification/subscription/resource-manager/Microsoft.Subscription/stable/2016-06-01/subscriptions.json#L210) first, and then the code here should be regerenated. As it stands, it looks like this PR's change will be lost the next time the code is regenerated.

@hermanschaaf You are right. I have created another PR in the RestAPi specs repo:
https://github.com/Azure/azure-rest-api-specs/pull/19870

tadelesh:
@navba-MSFT We do not change the SDK manually. This PR will be closed and after the swagger updated, we will release the new SDK according to service team's release request.

Issue #18619, opened 2022-07-19
Tenants List should return all the fields available in the API


hermanschaaf:
### Feature Request

The [TenantIDDescription](https://github.com/Azure/azure-sdk-for-go/blob/v61.6.0/services/subscription/mgmt/2020-09-01/subscription/models.go#L459-L465) struct returned by calls to `TenantsClient.ListComplete` contains only two fields: `ID` and `TenantID`:

```
// TenantIDDescription tenant Id information.
type TenantIDDescription struct {
	// ID - READ-ONLY; The fully qualified ID of the tenant. For example, /tenants/00000000-0000-0000-0000-000000000000.
	ID *string `json:"id,omitempty"`
	// TenantID - READ-ONLY; The tenant ID. For example, 00000000-0000-0000-0000-000000000000.
	TenantID *string `json:"tenantId,omitempty"`
}
```

However, the [underlying API](https://docs.microsoft.com/en-us/rest/api/resources/tenants/list?tabs=HTTP) returns many more useful fields, including `countryCode`, `displayName`, `domains`, `tenantCategory` `defaultDomain` and `tenantType`. It would be great if these fields could also be made accessible through the Go SDK.

navba-MSFT:
@hermanschaaf Apologies for the late reply. Thanks for reaching out to us and reporting this issue. We are looking into this issue and we will provide an update.

navba-MSFT:
The above PR has been filed and it is currently under review.

navba-MSFT:
@hermanschaaf Did you try to list tenant using the below package and check if that helps ?

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions#TenantsClient

hermanschaaf:
Hi @navba-MSFT, ah thanks, yes this SDK looks like it would work--it definitely contains the fields, at least! From the comment on the linked PR it sounds like the one I was using is actually deprecated? I will use the new one, thanks so much for your help.

PR #18670, opened 2022-07-22
{AzureSubscription} Fixes Azure/azure-sdk-for-go#18619

navba-MSFT:
Fixes Azure/azure-sdk-for-go#18619

The TenantIDDescription struct returned by calls to TenantsClient.ListComplete contains only two fields: ID and TenantID:

However, the underlying API returns many more useful fields, including countryCode, displayName, domains, tenantCategory defaultDomain and tenantType. It would be great if these fields could also be made accessible through the Go SDK.

### DIFF
--- a/services/subscription/mgmt/2020-09-01/subscription/models.go
+++ b/services/subscription/mgmt/2020-09-01/subscription/models.go
@@ -462,6 +462,20 @@ type TenantIDDescription struct {
 	ID *string `json:"id,omitempty"`
 	// TenantID - READ-ONLY; The tenant ID. For example, 00000000-0000-0000-0000-000000000000.
 	TenantID *string `json:"tenantId,omitempty"`
+        // TenantCategory - READ-ONLY; Category of the tenant. Possible values include: 'TenantCategoryHome', 'TenantCategoryProjectedBy', 'TenantCategoryManagedBy'
+	TenantCategory TenantCategory `json:"tenantCategory,omitempty"`
+	// Country - READ-ONLY; Country/region name of the address for the tenant.
+	Country *string `json:"country,omitempty"`
+	// CountryCode - READ-ONLY; Country/region abbreviation for the tenant.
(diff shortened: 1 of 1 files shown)


hermanschaaf:
Hi @navba-MSFT, thanks again for championing this change. I just wanted to note that I believe this is similar to the subscription case, in that the file here is generated from an underlying JSON source, so it will need to be updated at the source first.

chlowell (changes requested):
This file is generated. If it needs an update, we should regenerate the package rather than edit its files manually. The code here is correct though. It was generated for a REST API whose TenantIdDescription model has only two fields, as you can see in its specification [here](https://github.com/Azure/azure-rest-api-specs/blob/7a2ac91de424f271cf91cc8009f3fe9ee8249086/specification/subscription/resource-manager/Microsoft.Subscription/stable/2016-06-01/subscriptions.json#L299) (sorry I don't know where this API is found in the REST docs). The SDK package corresponding to the REST API mentioned in the linked issue is [services/resources/mgmt/2021-01-01/subscriptions](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2021-01-01/subscriptions#TenantIDDescription). However, that package is deprecated, so I recommend using the [new armsubscriptions module](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions#TenantsClient) instead.

This is a confusing couple of packages/APIs. Can another reviewer comment on why we have [armsubscription and armsubscriptions](https://pkg.go.dev/search?q=armsubscriptions) packages?

tadelesh:
@navba-MSFT As explanation in this [PR](https://github.com/Azure/azure-sdk-for-go/pull/18666), I'll close this PR. Thanks @chlowell to help to explain the deprecation of the package. I'll check these two subscription packages.

PR #18694, opened 2022-07-27
[azeventhubs] First checkin

richardpark-msft:
Can do:

- Can send batches of events.
- Can receive events, per partition.
- Can get the properties of partitions and the event hub itself.

What's missing:
- No processor/checkpoint store support (yet)
- AMQP message support not yet exposed
- No buffered producer
- Automated stress tests


jhendrixMSFT:
Is there an APIView for this or are we not there yet?

richardpark-msft:
@jhendrixMSFT

APIView here:
https://apiview.dev/Assemblies/Review/8245669789dd4453bea087a9afaf2705

sdk/messaging/azeventhubs/internal/auth/token.go
```diff
@@ -0,0 +1,58 @@
+// Package auth provides an abstraction over claims-based security for Azure Event Hub and Service Bus.
+package auth
+
+//	MIT License
```
jhendrixMSFT, line 4: Why does this file have a different license block format?

sdk/messaging/azeventhubs/internal/eh/eh_internal.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+package eh
+
+// ConvertToInt64 converts any int-like value to be an int64.
+func ConvertToInt64(intValue any) (int64, bool) {
```
jhendrixMSFT, line 6: Is the input to this always (or at least sometimes) `any`?  If not, you could use a generic implementation.
```go
func ToInt64[T int|int8|int16|int32|int64](v T) int64 {
	return int64(v)
}
```

sdk/messaging/azeventhubs/internal/errors.go
```diff
@@ -3,12 +3,334 @@
...
+type RecoveryKind string
+
+const (
+	RecoveryKindNone  RecoveryKind = ""
+	RecoveryKindFatal RecoveryKind = "fatal"
+	RecoveryKindLink  RecoveryKind = "link"
+	RecoveryKindConn  RecoveryKind = "connection"
+)
+
+func IsFatalSBError(err error) bool {
```
jhendrixMSFT, line 41: `IsFatalEHError()`

sdk/messaging/azeventhubs/log.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal/exported"
+)
+
+const (
+	// EventConn is used whenever we create a connection or any links (ie: producers, consumers).
+	EventConn log.Event = exported.EventConn
+
+	// EventAuth is used when we're doing authentication/claims negotiation.
+	EventAuth log.Event = "azeh.Auth"
```
jhendrixMSFT, line 16: This and other events should be type-aliased to their exported counterparts.

sdk/messaging/azeventhubs/internal/namespace.go
```diff
@@ -5,60 +5,147 @@ package internal
...
 )
 
 type (
-	namespace struct {
-		name          string
-		tokenProvider auth.TokenProvider
-		host          string
-		useWebSocket  bool
+	// Namespace is an abstraction over an amqp.Client, allowing us to hold onto a single
+	// instance of a connection per ServiceBusClient.
```
jhendrixMSFT, line 32: `EventHubsClient`

sdk/messaging/azeventhubs/mgmt.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	return EventHubProperties{
+		Name:         name,
+		CreatedOn:    createdOn,
+		PartitionIDs: partitionIDs,
+	}, nil
+}
+
+func newPartitionProperties(amqpValue any) (PartitionProperties, error) {
```
jhendrixMSFT, line 137: Why is the type `any` if the only acceptable type is `map[string]any`?

sdk/messaging/azeventhubs/internal/eh/eh_internal.go
richardpark-msft: That would be nicer, actually. The value that gets passed in here comes from the AMQP message properties, so it ends up being typed 'any' since it comes across the wire.

jhendrixMSFT: Ah bummer, makes sense though.

sdk/messaging/azeventhubs/internal/auth/token.go
richardpark-msft: Will remove. It's also in SB so I can trim it there as well.

sdk/messaging/azeventhubs/internal/eh/eh_internal.go
richardpark-msft: Yeah, I felt the same way. I did use some generics in other spots though :)

sdk/messaging/azeventhubs/mgmt.go
richardpark-msft: This is actually the amqpMessage.Value field, so it's typed 'any' already (more of the 'dynamic' stuff)

sdk/messaging/azeventhubs/internal/auth/token.go
richardpark-msft: (I'll create a separate PR to merge up whatever we change here)

richardpark-msft: (fixed here now)

sdk/messaging/azeventhubs/internal/errors.go
richardpark-msft: Fixed.

sdk/messaging/azeventhubs/internal/namespace.go
richardpark-msft: Fixed.

sdk/messaging/azeventhubs/log.go
richardpark-msft: Done.

jhendrixMSFT (approved):

sdk/messaging/azeventhubs/consumer_client.go
```diff
@@ -106,32 +106,33 @@ func NewConsumerClient(consumerGroup string, fullyQualifiedNamespace string, eve
...
+//
+// connectionString can be one of the following formats:
+//
+// Connection string, no EntityPath. In this case eventHub cannot be empty.
+// ex: Endpoint=sb://<your-namespace>.servicebus.windows.net/;SharedAccessKeyName=<key-name>;SharedAccessKey=<key>
+//
+// Connection string, has EntityPath. In this case eventHub must be empty.
 // ex: Endpoint=sb://<your-namespace>.servicebus.windows.net/;SharedAccessKeyName=<key-name>;SharedAccessKey=<key>;EntityPath=<entity path>
-func NewConsumerClientFromConnectionString(consumerGroup string, connectionString string, partitionID string, options *ConsumerClientOptions) (*ConsumerClient, error) {
+func NewConsumerClientFromConnectionString(connectionString string, eventHub string, partitionID string, consumerGroup string, options *ConsumerClientOptions) (*ConsumerClient, error) {
```
jhendrixMSFT, line 119: Shouldn't this be.
```go
func NewConsumerClientFromConnectionString(eventHub string, partitionID string, consumerGroup string, connectionString string, options *ConsumerClientOptions) (*ConsumerClient, error)
```

jhendrixMSFT: `(<RP specific params>, <cred param>, <options bag>)`

richardpark-msft: In this particular case the connection string is the endpoint. I can move it.

jhendrixMSFT: Leave as first.

richardpark-msft: Talked with @jhendrixMSFT , we'll leave it for now. Looks like we did this in the blob client, and led with the connection string.

In this particular case the connection string contains both an endpoint and a key, so it can make sense in either position.

PR #18797, opened 2022-08-05
Update Key Vault documentation

### LINKED ISSUE #18505: Update Key Vault documentation
We need to update the Key Vault documentation in a few places:

- [ ] **Authentication in Samples:** Our samples should show the use of Managed Identity instead of Service Principals with Client Secrets as the default authentication method. We can still show how to use the former for scenarios where the use of Managed Identity is not possible.
- [ ] **Cryptographic operations using CBC and CBCPAD:** We should make it clear that there are some considerations to have when using these algorithms for local cryptographic operations using our clients:
  - We should add a notice similar to the following:
    > Microsoft recommends you not use CBC without first ensuring the integrity of the ciphertext using an HMAC, for example. See https://docs.microsoft.com/dotnet/standard/security/vulnerabilities-cbc-mode for more information.
  - When it comes to a user provided IV, we should make it clear that if they pass their own IV, they need to make sure they use a cryptographically random, non-repeating IV. Here's an example:
    > Optional initialization vector (IV). If you pass your own IV, make sure you use a cryptographically random, non-repeating IV. If null, a cryptographically random IV will be choosing using {RandomNumberGenerator | whatever cryptorng your language provides}.

chlowell:
Closes #18505 by regenerating `azkeys` and replacing authentication and vault creation instructions with links to more comprehensive docs.

### DIFF
--- a/sdk/keyvault/azcertificates/README.md
+++ b/sdk/keyvault/azcertificates/README.md
@@ -1,7 +1,7 @@
 # Azure Key Vault Certificates client module for Go
 
 * Certificate management (this module) - create, manage, and deploy public and private SSL/TLS certificates
-* Cryptographic key management (([azkeys](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys))) - create, store, and control access to the keys used to encrypt your data
+* Cryptographic key management ([azkeys](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azkeys)) - create, store, and control access to the keys used to encrypt your data
 * Secrets management ([azsecrets](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets)) - securely store and control access to tokens, passwords, certificates, API keys, and other secrets
 
 [Source code][certificates_client_src] | [Package (pkg.go.dev)][reference_docs] |  [Product documentation][keyvault_docs] | [Samples][certificates_samples]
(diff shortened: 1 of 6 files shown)


jhendrixMSFT (approved):

sdk/keyvault/azcertificates/README.md
```diff
@@ -20,92 +20,17 @@ go get github.com/Azure/azure-sdk-for-go/sdk/azidentity
...
-
-#### Create a service principal (optional)
-
-This [Azure Cloud Shell][azure_cloud_shell] snippet shows how to create a new service principal. Before using it, replace "my-application" with an appropriate name for your application.
-
-Create a service principal:
-```Bash
-az ad sp create-for-rbac --name http://my-application --skip-assignment
-```
+* A supported Go version (the Azure SDK supports the two most recent Go releases)
```
vcolin7, line 23: Are these version numbers fixed or can they change in the future? If so, when a new version of Go is released would we drop support for the older one and add support for the newer one right away?

sdk/keyvault/azcertificates/README.md
```diff
@@ -20,92 +20,17 @@ go get github.com/Azure/azure-sdk-for-go/sdk/azidentity
...
+* A Key Vault. If you need to create one, see the Key Vault documentation for instructions on doing so in the [Azure Portal][azure_keyvault_portal] or with the [Azure CLI][azure_keyvault_cli].
```
vcolin7, line 24: nit; In official documentation, references to the Key Vault service start with uppercase letters but mentions to key vault instances do not.
```suggestion
* A key vault. If you need to create one, see the Key Vault documentation for instructions on doing so in the [Azure Portal][azure_keyvault_portal] or with the [Azure CLI][azure_keyvault_cli].
```

sdk/keyvault/azcertificates/README.md
chlowell: I removed the specific version because the requirement will change over time as Google releases new versions (roughly every 6 months). We won't ordinarily ship a module only to upgrade its required Go version.

Issue #15615, opened 2021-09-22
Support workload identity federation


chlowell:
~Blocked on #17417~

Workload Identity isn't yet supported out of the box but my comment below (https://github.com/Azure/azure-sdk-for-go/issues/15615#issuecomment-1211012677) has an implementation you can copy in the meantime for use with azidentity v1.2.0-beta.2 or later.

s-bauer:
Hey any eta on when this will be added?

chlowell:
May at the earliest, in the current plan. #16902 has an example you could follow in the meantime if you want to implement it yourself.

tkent:
I hope this is added soon. This is pretty important feature since [azure-workload-identity](https://github.com/Azure/azure-workload-identity) is replacing the prior method of k8s identity assignment (aad-pod-identity](https://github.com/Azure/aad-pod-identity), which does work with the existing SDK) 

Getting `azure-workload-identity` setup is straightforward enough, but once you do you quickly realize it's unusable if your applications use identities resolved using the `azure-go-sdk`.

tkent:
For other folks stuck dealing with this, the code below is what we ended up using for our application. It's been working for the past few days in our AWS EKS clusters using Azure Workload Identity. 

```golang

import (
	"context"
	"errors"
	"fmt"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
	confidential "github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
	"os"
)

// !! Warning - A regrettable hack !!
//
// MSFT has put their golang sdk consumers in a tough spot. Azure Workload
// Identity is replacing AAD Pod Identity as the solution to provide
// dynamically assigned credentials to workloads in a k8s cluster.
//
// Unfortunately, the azure-go-sdk's identity module simply doesn't
// support this type of authentication and the request to do so has
// been open since 09/2021.
//
//  https://github.com/Azure/azure-sdk-for-go/issues/15615
//
// The contents of this file provides a functional azcore.TokenCredential
// implementation that will work with the assertion provided by Azure
// Workload Identity.
//
// It is very difficult to test this and no care around caching was taken.
// This has worked well in the short time we've used it and the hope is that
// MSFT will add the functionality into their golang sdk in the next months.
//
// A sample of using this alongside the `DefaultAzureCredential` to mimic
// the behavior in other SDKs:
//
[...]

PR #18807, opened 2022-08-08
Add azidentity.ClientAssertionCredential

### LINKED ISSUE #17417: Support for client assertions
### Feature Request

In the [Azure Key Vault Provider for Secrets Store CSI Driver](https://github.com/Azure/secrets-store-csi-driver-provider-azure) we're using:

1. [adal](https://github.com/Azure/go-autorest/blob/master/autorest/adal/cmd/adal.go) for service principal (client secret), managed identity (user-assigned and system-assigned)
2. [msal](https://github.com/AzureAD/microsoft-authentication-library-for-go) for the new workload identity federation support.

We eventually want to move to using Azure Identity go SDK for supported auth methods instead of relying on adal/msal individually. Using azure identity sdk will provide us with the benefit of token cache and relying on the sdk to make backward compatible changes when there are changes to msal in the future. 

Opening this issue as the azure identity go sdk doesn't support client assertions.

chlowell:
Closes #17417 by adding a credential that grants applications full control over the assertions they use to authenticate. This credential enables advanced scenarios like authenticating with a remote certificate and will be the basis of our future Workload Identity Federation support (#15615).

~~This is a draft because it requires the next, as yet unreleased, version of MSAL for Go.~~


sdk/azidentity/client_assertion_credential.go
```diff
@@ -0,0 +1,84 @@
...
+type ClientAssertionCredential struct {
+	client confidentialClient
+}
+
+// ClientAssertionCredentialOptions contains optional parameters for ClientAssertionCredential.
+type ClientAssertionCredentialOptions struct {
+	azcore.ClientOptions
+}
+
+// NewClientAssertionCredential constructs a ClientAssertionCredential. The getAssertion function must be thread safe. Pass nil for options to accept defaults.
```
jhendrixMSFT, line 34: `...must be safe for concurrent use.`

Does the return value need to be documented?

chlowell: I'll add a link to the Azure AD documentation of the format.

jhendrixMSFT (approved):

PR #18808, opened 2022-08-08
Fix formatting and linting errors following 1.19 upgrade

chlowell:
The new `gofmt` adds rules for doc comments. These will create diffs, and thus failures, in CI after upgrading pipelines to 1.19, so for this PR I ran the updated `gofmt` on `sdk/**/*`. We also have new linter warnings because the upgraded golangci-lint warns on io/ioutil usage.


richardpark-msft (approved):
The best comments I've ever seen.

sdk/resourcemanager/resources/armpolicy/zz_generated_assignments_client.go
```diff
@@ -449,7 +449,7 @@ func (client *AssignmentsClient) listHandleResponse(resp *http.Response) (Assign
 
 // NewListForManagementGroupPager - This operation retrieves the list of all policy assignments applicable to the management
 // group that match the given $filter. Valid values for $filter are: 'atScope()', 'atExactScope()' or
-// 'policyDefinitionId eq '{value}''. If $filter=atScope() is provided, the returned list includes all policy assignments
+// 'policyDefinitionId eq '{value}”. If $filter=atScope() is provided, the returned list includes all policy assignments
```
tadelesh, line 452: This format is not correct. There is an [issue](https://github.com/golang/go/issues/54312) for this. But it seems no good solution for now.

tadelesh (commented):
Replacing two single quotes with double quotes in the new `gofmt` will lead to some incorrect comments. Maybe we need to refine the swagger description to avoid use of two single quotes.

sdk/resourcemanager/resources/armpolicy/zz_generated_assignments_client.go
chlowell: This has no impact on published documentation because godoc was already replacing `''` with `”` before 1.19. For example, see the [published documentation for this package](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armpolicy#AssignmentsClient.NewListForResourcePager). The new `gofmt` only changes comment text to match godoc's rendering. But it is annoying we can't represent the empty string as `''` in docs.

richardpark-msft:
Hey @chlowell, I had to gofmt azservicebus so there'll be some conflicts. Probably easiest to just merge with main.

chlowell:
cherry-picked #18822 because some modules in this PR need that fix as well 🙄

jhendrixMSFT (approved):

sdk/resourcemanager/resources/armpolicy/zz_generated_assignments_client.go
tadelesh: Got it. Only `armpolicy` has such problem that we could change swagger at least.

PR #18813, opened 2022-08-08
[azservicebus] Ignore the leading XMLNS for properties in subscription rules

### LINKED ISSUE #18785: [azservicebus] NewListRulesPager not working with Custom Properties in Correlation Filter
### Bug Report

NewListRulesPager not working with Custom Properties in Correlation Filter

- SDK version - github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v1.0.2
- output of `go version` - go version go1.18.1 linux/amd64

- What happened?
listRulesPager.NextPage(ctx) gives below error when the rule filter is of type Correlation filter and contains Custom Properties.
Error: type d5p1:string of parameter check is not a handled type for SQL parameters

NewListRulesPager  should not return error.
- How can we reproduce it?
Create a Rule on Subscription with filter of type Correlation filter and add Custom field in it. 
Try to fetch the Rule using NewListRulesPager
image

richardpark-msft:
The individual properties for a rule in Service Bus come with an XML ns prefix. We don't actually care about this as we're just trying to figure out what the assigned type is (string, int, etc..) so this just strips out the prefix.

Fixes #18785

### DIFF
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -19,6 +19,7 @@
     not being delivered, but still incrementing their delivery count and requiring the message lock 
     timeout to expire.
   - Link detach could result in messages being ignored, requiring the message lock timeout to expire.
+- Subscription rules weren't deserializing properly when created from the portal (PR#18813)
 
 ## 1.0.2-beta.0 (2022-07-07)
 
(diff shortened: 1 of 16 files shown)


jhendrixMSFT (approved):

richardpark-msft:
(had to go 1.19 gofmt)

richardpark-msft:
The CI pipeline seems to be having issues unrelated to tests or code (it's failing to upload an artifact, but it's CI).

Bypassing - live tests and other CI pipelines have passed

PR #18821, opened 2022-08-09
temperarily add honor-body-placement tag to prevent codegen breaking

tadelesh:


ArcturusZhang (commented):
can we add a swtich for build.ps1 script so that this could be set individually for every RP, and its default value is turned on?
In this way we could make sure things work both fine for those that turned this on and off

ArcturusZhang (approved):

PR #18834, opened 2022-08-10
[azeventhubs] Bug fixes: send to 'any' partition, also offset position strings were incorrect.

richardpark-msft:
Fixing two bugs and adding in some tests:
- When you send a batch of events you can choose to send it to any available partition. I missed a nil check, added in and added a test.
- The offset position filter strings were being incorrectly formatted. It appears that the sequence number one was accepted _anyways_, but I've now fixed it so they're aligned with what JS does.

### DIFF
--- a/sdk/messaging/azeventhubs/consumer_client.go
+++ b/sdk/messaging/azeventhubs/consumer_client.go
@@ -5,6 +5,7 @@ package azeventhubs
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"time"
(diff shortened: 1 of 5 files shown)


sdk/messaging/azeventhubs/consumer_client.go
```diff
@@ -220,34 +217,66 @@ func (cc *ConsumerClient) Close(ctx context.Context) error {
...
 
-func getOffsetExpression(startPosition StartPosition) string {
+func getOffsetExpression(startPosition StartPosition) (string, error) {
 	lt := ">"
 
 	if startPosition.Inclusive {
 		lt = ">="
 	}
 
+	var errMultipleFieldsSet = errors.New("Only a single start point can be set: Earliest, EnqueuedTime, Latest, Offset, or SequenceNumber")
```
jhendrixMSFT, line 227: Don't capitalize the first letter in error messages (the linter will bark at this).

jhendrixMSFT (approved):
One little nit to clean up.

sdk/messaging/azeventhubs/consumer_client.go
richardpark-msft: Gah, and I _know_ this and still did it.

Issue #16830, opened 2022-01-14
Add support for profiles


jhendrixMSFT:
Investigate adding profile support to track 2.  Ideally this would be based on resource type, not resource provider like we did in track 1.

cc: @kimprice @mikekistler

jhendrixMSFT:
Sketching out an implementation (not comprehensive but conveys the general pattern).
```
/sdk/profiles
+------------/20200901hybrid
             +--------------/doc.go
             +--------------/go.mod
             +--------------/keyvault
                            +--------/azkeys
                            +--------/azsecrets
             +--------------/resourcemanager
                            +---------------/compute
                                            +-------/armcompute
                            +---------------/resources
                                            +---------/armlocks
                                            +---------/armpolicy
```

jhendrixMSFT:
One problem I see is for the data-plane packages (key vault and blob storage).  Our data-plane packages typically have a hand-written convenience layer on top of the generated code.  Scaling this to all profiles isn't sustainable.  For key vault, we could probably get away with a 100% codegen solution, although it would mean that the profile versions of key vault would have a slightly different API than their non-profile counterparts which will impact migrating away from a profile or sharing some common code with profile/non-profile services (I don't know how common this is).  Blob storage is the most problematic as a 100% codegen solution for it just isn't feasible.

bganapa:
Current AzureStack SDKs have resource types that are not supported in the AzureStack cloud.  Main reason being autorest generates based on the input files given. Core resource providers like compute has not split the swagger files in to per resource type  files. However wherever possible, we include only the resource types supported while generating the SDK. e.g Network has per resource type files and we have profile specific tags including only the needed files

Also we need new track 2 SDK  only for the profile 2020-09-01-hybrid profile. We do not need to ship track2 SDKs for the 2019,2018,2017 profiles

For Storage/Keyvault dataplane SDKs,  Java/C# are going towards having single package supporting multiple api versions. This means that there is no need to ship separate packages for azure stack.  For Go, we could copy the storage plane sdks from the existing tag?

Alos it sounds like KV and storage have manual written part which would differ from the latest. Ideally profile specific dataplane sdks should follow the same programming paradigm as Azure. The most common scenario is porting apps written for Azure in to Azure Stack.

tadelesh:
[...]

PR #18841, opened 2022-08-10
Preview version of mgmt. profile package for Azure Stack

tadelesh:
Partially resolve https://github.com/Azure/azure-sdk-for-go/issues/16830.
- Add some packages to generate mgmt. plane config and SDK for profile.
- First version of profile config `2020-09-01-hybrid` package which only contains mgmt. plane modules.


sdk/profiles/hybrid20200901/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/azure-sdk-for-go/sdk/profiles/hybrid20200901
+
+go 1.19
```
jhendrixMSFT, line 3: This should be `go 1.18`

jhendrixMSFT:
Few questions.

- Does this include full generated code, i.e. no removal of resource types?
- When will data-plane parts (i.e. key vault) be added?

tadelesh:
> Few questions.
> 
> * Does this include full generated code, i.e. no removal of resource types?
> * When will data-plane parts (i.e. key vault) be added?

* Does this include full generated code, i.e. no removal of resource types?
Yes. I found it's hard to pair the config with operation/operation group. For example, which operation belong to `storageAccounts/blobServices` in the [spec](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/storage/resource-manager/Microsoft.Storage/stable/2019-06-01/storage.json) according to this [config](https://github.com/Azure/azure-rest-api-specs/blob/80065490402157d0df0dd37ab347c651b22eb576/profile/2020-09-01-hybrid.json#L462-L463). If I could figure out how to pair them, I think I could find some way to remove useless resource types.

* When will data-plane parts (i.e. key vault) be added?
Since we will announce the deprecation for track1 publicly later this September, I think we need to finish the work before that. According to the [issue](https://github.com/Azure/azure-sdk-for-go/issues/16830), it seems we don't have a conclusion for how we add the data plane modules. Is there any update for this?

sdk/profiles/hybrid20200901/resourcemanager/authorization/armauthorization/constants.go
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+// DO NOT EDIT.
+
+package armauthorization
+
+const (
+	moduleName    = "hybrid20200901/armauthorization"
+	moduleVersion = "v1.0.0"
```
jhendrixMSFT, line 14: I don't think this is correct. The module name and version should be declared at the module level, not per package. Will need codegen support.

tadelesh: For profile, we only have one module for all the RPs. If we only declare the model info in the root level, then telemetry data will not include any sub-package info. It that OK or we have other guideline for sub-package telemetry data format?

tadelesh: @jhendrixMSFT I have removed all the `moduleName` and `moduleVersion` from the sub-packages and introduce a root replacement. The side effect is that we need to export these two variables. Do you think it is OK?

sdk/profiles/hybrid20200901/constants.go
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+// DO NOT EDIT.
+
+package hybrid20200901
+
+const (
+	ModuleName    = "hybrid20200901"
```
jhendrixMSFT, line 13: This can go under an internal path to avoid exporting it.
```
sdk/profiles/hybrid20200901/internal
```

tadelesh:
I've moved tooling change out of this PR and currently this PR should be a clear beta version for profile `hybrid20200901`, but only support mgmt. plane.

sdk/profiles/hybrid20200901/internal/constants.go
```diff
@@ -0,0 +1,15 @@
...
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+// DO NOT EDIT.
+
+package hybrid20200901
+
+const (
+	ModuleName    = "hybrid20200901"
+	ModuleVersion = "v0.1.0"
+)
```
jhendrixMSFT, line 15: Missing new-line is causing the analysis failure.

sdk/profiles/hybrid20200901/internal/constants.go
```diff
@@ -0,0 +1,15 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+// DO NOT EDIT.
+
+package hybrid20200901
```
jhendrixMSFT, line 10: Package name doesn't match directory name.

sdk/profiles/hybrid20200901/definition.json
```diff
@@ -0,0 +1,82 @@
+{
```
jhendrixMSFT, line 1: This info looks redundant with what's in the autorest.md files.  Can it be removed?

sdk/profiles/hybrid20200901/build.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+// This file enables 'go generate' to regenerate this specific SDK
+//go:generate pwsh ../../../eng/scripts/build.ps1 -skipBuild -format -tidy profiles/hybrid20200901
```
jhendrixMSFT, line 5: This will regenerate the entire profile right?

sdk/profiles/hybrid20200901/resourcemanager/authorization/armauthorization/autorest.md
```diff
@@ -0,0 +1,17 @@
...
+
+> see https://aka.ms/autorest
+
+``` yaml
+azure-arm: true
+require:
+- https://github.com/Azure/azure-rest-api-specs/blob/53b1affe357b3bfbb53721d0a2002382a046d3b0/specification/authorization/resource-manager/readme.md
+- https://github.com/Azure/azure-rest-api-specs/blob/53b1affe357b3bfbb53721d0a2002382a046d3b0/specification/authorization/resource-manager/readme.go.md
+license-header: MICROSOFT_MIT_NO_VERSION
+module-version: 0.1.0
```
jhendrixMSFT, line 11: This, and `module-name`, should be removed as the sub packages aren't modules.

sdk/profiles/hybrid20200901/definition.json
tadelesh: This config is used to generate autorest.md files for all the profile packages.

sdk/profiles/hybrid20200901/resourcemanager/authorization/armauthorization/autorest.md
tadelesh: For current codegen, without module-version configured, arm package could not be generated successfully.

jhendrixMSFT:
I see that the root `build.go` file was removed.  While you don't have to fix it now, we really should ensure that all modules can be built via `go generate`.

sdk/profiles/hybrid20200901/resourcemanager/authorization/armauthorization/autorest.md
jhendrixMSFT: What was the error?  Is this somethin we need to fix in the code generator?

tadelesh: Current codegen will always generate `moduleVersion` and `moduleName` constants ([code](https://github.com/Azure/autorest.go/blob/2fe13f887fd5a7def9a77590d294545239ec8433/src/generator/constants.ts#L34). And these two constants will be used when build pipeline ([code](https://github.com/Azure/autorest.go/blob/2fe13f887fd5a7def9a77590d294545239ec8433/src/generator/operations.ts#L214)). If we need to change codegen, we have to decide when it is a sub-package, what we should do for the pipeline building.

tadelesh:
> I see that the root `build.go` file was removed. While you don't have to fix it now, we really should ensure that all modules can be built via `go generate`.

OK. I think if we decide to use profile, I'll do the fix.

sdk/profiles/hybrid20200901/resourcemanager/authorization/armauthorization/autorest.md
jhendrixMSFT: Got it, thanks for clarifying.

tadelesh: @jhendrixMSFT Regarding the solution you and Jeff propose, do we need to change code generator to support generate the sub package, or just like this PR, do some manual work or maybe could use directive to change the code?

jhendrixMSFT: I suspect the solution will be a combination of small changes to the code generator and some autorest transforms.

tadelesh:
@jhendrixMSFT @JeffreyRichter @RickWinter I've refined the PR with the new module path: `github.com/Azure/azure-sdk-for-go/profile/v20200928`. One thing I'm not sure is the `v20200928` is the pattern for Go module semantic version, will it be indexed by `pkg.go.dev` as profile's version `v20200928`?

tadelesh:
@benbp It seems some scripts could not deal with our new module path, could you help?

JeffreyRichter:
What is there to review really? If it appears in the repo correctly, then it's right.

benbp:
@tadelesh sorry I'm missing context on why this isn't in the `sdk` folder? The `sdk/*` convention is a pretty hard requirement for almost all of our engineering system tools. If an exception is needed, we'll need to have a more in-depth conversation about what's possible and how much work it will involve.

profile/v20200901/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+      - release/*
+  paths:
+    include:
+    - profile/v20200901/
+
+stages:
+- template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
+  parameters:
+    IncludeRelease: true
+    ServiceDirectory: '../profile/v20200901'
```
benbp, line 28: How frequently will new profiles be added, that is to say, do we need to think about a large amount of pipelines getting created that differ by profile date/version?

tadelesh: I think there won't be too much new profile, maybe one per year or even longer interval, it is only for Azure Stack. Thus, I could bear with little manual work to let pipeline work.

tadelesh:
> @tadelesh sorry I'm missing context on why this isn't in the `sdk` folder? The `sdk/*` convention is a pretty hard requirement for almost all of our engineering system tools. If an exception is needed, we'll need to have a more in-depth conversation about what's possible and how much work it will involve.

We have a discussion about how to deal with profile for our Go track2 SDK recently. As we do not want to miss-lead customer to wrongly use profile module, we are thinking about put them in `profile` parallelly to `sdk`.

tadelesh:
@jhendrixMSFT I've updated the PR, could you help to approve?

benbp:
Ok that yaml is good with me @tadelesh

tadelesh:
Then it seems everything is good. @jhendrixMSFT do you have any further concern? After your approval, I will merge and release the first beta.

profile/v20200901/resourcemanager/appservice/armappservice/client_factory.go
```diff
@@ -0,0 +1,70 @@
...
+	options        *arm.ClientOptions
+}
+
+// NewClientFactory creates a new instance of ClientFactory with the specified values.
+// The parameter values will be propagated to any client created from this factory.
+//   - subscriptionID - Your Azure subscription ID. This is a GUID-formatted string (e.g. 00000000-0000-0000-0000-000000000000).
+//   - credential - used to authorize requests. Usually a credential from azidentity.
+//   - options - pass nil to accept the default values.
+func NewClientFactory(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
+	_, err := arm.NewClient(internal.ModuleName+".ClientFactory", internal.ModuleVersion, credential, options)
```
jhendrixMSFT, line 32: This is going to be ambiguous for all client factories.

profile/v20200901/resourcemanager/appservice/armappservice/certificates_client.go
```diff
@@ -0,0 +1,391 @@
...
+	internal       *arm.Client
+	subscriptionID string
+}
+
+// NewCertificatesClient creates a new instance of CertificatesClient with the specified values.
+//   - subscriptionID - Your Azure subscription ID. This is a GUID-formatted string (e.g. 00000000-0000-0000-0000-000000000000).
+//   - credential - used to authorize requests. Usually a credential from azidentity.
+//   - options - pass nil to accept the default values.
+func NewCertificatesClient(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*CertificatesClient, error) {
+	cl, err := arm.NewClient(internal.ModuleName+".CertificatesClient", internal.ModuleVersion, credential, options)
```
jhendrixMSFT, line 37: This might also be ambiguous for some clients.

profile/v20200901/resourcemanager/appservice/armappservice/client_factory.go
tadelesh: Oh, yes. It will cause distributed tracing ambiguous. I added a package prefix to it (`internal.ModuleName+"/armappservice.ClientFactory"`). Is that OK?

jhendrixMSFT: This will show up in the telemetry data as `v20200901/armappservice.ClientFactory`.  How much do we care about being able to distinguish the profile version of this client with the non-profile version?

CC @JeffreyRichter

tadelesh: The complete telemetry will be `azsdk-go-v20200901/armappservice/v0.1.0`. A little bit ugly, but I think it is useful when analyzing user data.

JeffreyRichter: I'm OK with this too.

jhendrixMSFT (approved):

Issue #17943, opened 2022-05-13
azblob: Test recordings are being shipped with the product


jhendrixMSFT:
All non-production assets must go under the `testdata` directory.  At present, `azblob` is shipping ~2MB of recording data in every release.

Looks like this is not currently configurable in `internal/recording`, this needs to be fixed too.

jhendrixMSFT:
Fixed in upcoming `azblob@v0.5.0`

PR #18850, opened 2022-08-11
Move test recordings under the testdata directory

jhendrixMSFT:
This will omit megabytes of test assets from the module.
Updates test setup and teardown to use the latest APIs which allow
specifying the recordings directory.
Sanitized parts of the recordings to make the diffs less noisy. More
work is needed though.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/17943


mohsha-msft:
There are changes required in the test framework 
1. To move the test files in their respected packages and then 
2. Moving recordings to testdata/
3. Upgrade testing infra to take recording path insead of default path "recording/Test/testname"

But I realised it would need more time and seperate efforts.

jhendrixMSFT:
Unfortunately, there are two sets of APIs in the test framework.  The original ones, which `azblob` is currently using, and updated ones which are introduced in this PR.  The updated ones let you specify the location and most of the tests are working.  I need to tweak the sanitization a bit more to fix the remaining tests.  I'll have this resolved today.

jhendrixMSFT:
The problem was freezing the time stamps as this broke tests that use conditionals (e.g. `if-modified-since`).  I reverted that part of the change for now.

mohsha-msft (approved):
:shipit:

PR #18864, opened 2022-08-11
Move lease clients to lease package

jhendrixMSFT:
Keep lease clients separate from other client types as per the pattern.
There are two lease clients, one for containers and the other for all
blob client types.
Removed NewLeaseClient() methods from other client types.


sdk/storage/azblob/lease/blob_client.go
```diff
@@ -0,0 +1,141 @@
...
+
+// LeaseID returns leaseID of the client.
+func (c *BlobClient) LeaseID() *string {
+	return c.leaseID
+}
+
+// AcquireLease acquires a lease on the blob for write and delete operations.
+// The lease Duration must be between 15 and 60 seconds, or infinite (-1).
+// For more information, see https://docs.microsoft.com/rest/api/storageservices/lease-blob.
+func (c *BlobClient) AcquireLease(ctx context.Context, o *BlobAcquireOptions) (BlobAcquireResponse, error) {
```
jhendrixMSFT, line 84: I'm thinking we should remove the `Lease` suffix from these APIs as they're already on a type within the `lease` package.  Thoughts?

sdk/storage/azblob/lease/blob_client.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// BlobClientOptions contains the optional values when creating a BlobClient.
+type BlobClientOptions struct {
+	// LeaseID contains a caller-provided lease ID.
+	LeaseID *string
+}
+
+// NewBlobClient creates a blob lease client for the provided blob client.
+func NewBlobClient[T appendblob.Client | blob.Client | blockblob.Client | pageblob.Client](client *T, options *BlobClientOptions) (*BlobClient, error) {
```
jhendrixMSFT, line 36: The approach used here is the same as C#.  The lease client is built from an existing client.

sdk/storage/azblob/lease/blob_client.go
mohsha-msft: I think we should keep it as it makes API names more readable. The function is built on client that may not have lease in the name. `client.Acquire()` is less readable than `client.AcquireLease()`.

mohsha-msft (approved):

PR #18872, opened 2022-08-12
Separate client options per client type

jhendrixMSFT:
This allows client options to independently evolve.


sdk/storage/azblob/blockblob/client.go
```diff
@@ -27,39 +27,44 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared"
 )
 
+// ClientOptions contains the optional parameters when creating a Client.
+type ClientOptions struct {
+	azcore.ClientOptions
```
richardpark-msft, line 32: So for _input_ types would we prefer to inline all the options rather than embed? Or perhaps give it a name of 'Core' or something so it's a bit easier to initialize?

richardpark-msft (commented):
Seems sensible. One comment about the "input types with embedded data" thing, only because we've talked about this kind of pattern before.

chlowell (approved):

sdk/storage/azblob/blockblob/client.go
jhendrixMSFT: Client options is a bit of a special case.  We want to be able to add a new policy and its options type in `azcore` only without requiring an update of SDKs.

richardpark-msft: Seems sensible to me.

richardpark-msft (approved):

PR #18873, opened 2022-08-13
Generate API review file and publish as artifact

praveenkuttappan:
Add new step in build pipeline to create API review code file and publish it as artifact. This will be used by APIView to create review.


jhendrixMSFT:
@praveenkuttappan is this ready to go? Where can I find examples of the generated APIViews? Should it be adding a link in the PR?

PR #18887, opened 2022-08-17
APIView Review Comment Fixes

mohsha-msft:


sdk/storage/azblob/blob/client.go
```diff
@@ -301,10 +268,10 @@ func (b *Client) GetSASToken(permissions SASPermissions, start time.Time, expiry
 
 // Concurrent Download Functions -----------------------------------------------------------------------------------------
 
-// DownloadToWriterAt downloads an Azure blob to a WriterAt in parallel.
-func (b *Client) DownloadToWriterAt(ctx context.Context, writer io.WriterAt, o *DownloadToWriterAtOptions) error {
+// downloadWriterAt downloads an Azure blob to a WriterAt in parallel.
+func (b *Client) downloadWriterAt(ctx context.Context, writer io.WriterAt, o *DownloadWriterAtOptions) error {
```
jhendrixMSFT, line 272: Do we need to export `DownloadWriterAtOptions`?

sdk/storage/azblob/blob/client.go
```diff
@@ -370,14 +337,47 @@ func (b *Client) DownloadToWriterAt(ctx context.Context, writer io.WriterAt, o *
...
+	return DownloadStreamResponse{
+		b:                          b,
+		BlobClientDownloadResponse: dr,
+		getInfo:                    HTTPGetterInfo{Offset: offset, Count: count, ETag: eTag},
+		ObjectReplicationRules:     deserializeORSPolicies(dr.ObjectReplicationRules),
+	}, err
+}
+
+// DownloadBuffer downloads an Azure blob to a buffer with parallel.
+func (b *Client) DownloadBuffer(ctx context.Context, buffer []byte, o *DownloadBufferOptions) error {
```
jhendrixMSFT, line 374: This needs to return the number of bytes downloaded.

sdk/storage/azblob/blob/client.go
```diff
@@ -370,14 +337,47 @@ func (b *Client) DownloadToWriterAt(ctx context.Context, writer io.WriterAt, o *
...
+	return b.downloadWriterAt(ctx, shared.NewBytesWriter(buffer), o)
 }
 
-// DownloadToFile downloads an Azure blob to a local file.
+// DownloadFile downloads an Azure blob to a local file.
 // The file would be truncated if the size doesn't match.
-func (b *Client) DownloadToFile(ctx context.Context, file *os.File, o *DownloadToFileOptions) error {
+func (b *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFileOptions) (int64, error) {
```
jhendrixMSFT, line 380: Do we need to return the number of bytes here too?

sdk/storage/azblob/blob/models.go
```diff
@@ -146,11 +155,11 @@ func (o *DownloadToWriterAtOptions) getDownloadBlobOptions(offSet, count int64,
 	}
 }
 
-// DownloadToBufferOptions identifies options used by the DownloadToBuffer and DownloadToFile functions.
-type DownloadToBufferOptions = DownloadToWriterAtOptions
+// DownloadBufferOptions identifies options used by the DownloadBuffer and DownloadFile functions.
+type DownloadBufferOptions = DownloadWriterAtOptions
```
jhendrixMSFT, line 159: This type, and `DownloadFileOptions`, should be discrete types so that they can independently evolve over time.  They can anonymously embed the `DownloadWriterAtOptions`.

sdk/storage/azblob/blob/client.go
mohsha-msft: As per the API review discussion, we need to return the size of the file downloaded. And truncate the local file size if required.

sdk/storage/azblob/blob/models.go
```diff
@@ -100,8 +109,8 @@ func (o *DownloadToStreamOptions) format() (*generated.BlobClientDownloadOptions
 
 // ---------------------------------------------------------------------------------------------------------------------
 
-// DownloadToWriterAtOptions identifies options used by the DownloadToBuffer and DownloadToFile functions.
-type DownloadToWriterAtOptions struct {
+// DownloadWriterAtOptions identifies options used by the DownloadBuffer and DownloadFile functions.
+type DownloadWriterAtOptions struct {
```
jhendrixMSFT, line 113: Since we're no longer exported the matching method for this, we should rename it to something simpler like `DownloadOptions`.

sdk/storage/azblob/blob/client.go
mohsha-msft: yes of course. Let me complete that.

sdk/storage/azblob/blockblob/models.go
```diff
@@ -240,9 +248,9 @@ type TransferManager = shared.TransferManager
 
 // UploadStreamOptions provides set of configurations for UploadStream operation
 type UploadStreamOptions struct {
-	// TransferManager provides a TransferManager that controls buffer allocation/reuse and
+	// transferManager provides a transferManager that controls buffer allocation/reuse and
 	// concurrency. This overrides BufferSize and MaxBuffers if set.
-	TransferManager      TransferManager
+	transferManager      TransferManager
```
jhendrixMSFT, line 253: We should also remove the exported `TranseferManager` type on line 247.

sdk/storage/azblob/blob/models.go
mohsha-msft: Hmm, I think it makes sense. Alright.

sdk/storage/azblob/internal/exported/blob_url.go
```diff
@@ -100,9 +100,9 @@ func ParseBlobURL(u string) (BlobURLParts, error) {
 	return up, nil
 }
 
-// URL returns a URL object whose fields are initialized from the BlobURLParts fields. The URL's RawQuery
+// URL returns a URL object whose fields are initialized from the URLParts fields. The URL's RawQuery
 // field contains the SAS, snapshot, and unparsed query parameters.
-func (up BlobURLParts) URL() string {
+func (up URLParts) URL() string {
```
jhendrixMSFT, line 105: This should be `String()` (the `fmt.Stringer` interface).

sdk/storage/azblob/internal/generated/zz_models.go
```diff
@@ -1720,7 +1720,7 @@ type UserDelegationKey struct {
 	SignedExpiry *time.Time `xml:"SignedExpiry"`
 
 	// REQUIRED; The Azure Active Directory object ID in GUID format.
-	SignedOid *string `xml:"SignedOid"`
+	SignedOid *string `xml:"SignedOID"`
```
jhendrixMSFT, line 1723: Was this, and the change on line 1732, intentional or due to find/replace.

sdk/storage/azblob/blob/models.go
mohsha-msft: It'll collide with the actual Download() function present in the package. So I think we can keep it like this.

jhendrixMSFT: I'm not following, can you please clarify?

mohsha-msft: There's already a DownloadOptions for Download() function. So renaming it DownloadOption will create a conflict. May be some other name perhaps?

sdk/storage/azblob/blockblob/responses.go
```diff
@@ -98,3 +98,9 @@ func toUploadReaderAtResponseFromCommitBlockListResponse(resp CommitBlockListRes
 		XMSContentCRC64:     resp.XMSContentCRC64,
 	}
 }
+
+// UploadFileResponse contains the response from method Client.UploadBuffer/Client.UploadFile.
+type UploadFileResponse = UploadReaderAtResponse
```
jhendrixMSFT, line 103: We should rename `UploadReaderAtResponse` to simply `UploadResponse` as we don't have any `UploadReaderAt` API.

Also, can you please take a look at the various response types for the `Upload*` APIs.  E.g. I see a type `UploadReaderAtResponse` which looks to be a copy of a type from the generated code instead of a type alias.

sdk/storage/azblob/blob/models.go
mohsha-msft: Ohh never mind. Apologies for the confusion.

sdk/storage/azblob/blockblob/responses.go
mohsha-msft: There is already a function `func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *UploadOptions)` for Upload() API.  `UploadReaderAtResponse` is response received from `uploadReaderAtToBlockBlob` so we can not export it but we it'll be required.

sdk/storage/azblob/blob/client.go
mohsha-msft: Done. Not required to export to made it lowercase.

sdk/storage/azblob/internal/generated/zz_models.go
mohsha-msft: Ohh Goland must have caused this. Reverted it back

sdk/storage/azblob/blob/models.go
mohsha-msft: Renamed both the function and option bag - `download` `downloadOptions`

sdk/storage/azblob/blob/models.go
```diff
@@ -146,11 +155,11 @@ func (o *DownloadToWriterAtOptions) getDownloadBlobOptions(offSet, count int64,
...
 
-// DownloadToBufferOptions identifies options used by the DownloadToBuffer and DownloadToFile functions.
-type DownloadToBufferOptions = DownloadToWriterAtOptions
+// DownloadBufferOptions identifies options used by the DownloadBuffer and DownloadFile functions.
+type DownloadBufferOptions = downloadOptions
 
-// DownloadToFileOptions identifies options used by the DownloadToBuffer and DownloadToFile functions.
-type DownloadToFileOptions = DownloadToWriterAtOptions
+// DownloadFileOptions identifies options used by the DownloadBuffer and DownloadFile functions.
+type DownloadFileOptions = downloadOptions
```
jhendrixMSFT, line 162: This and `DownloadBufferOptions` should be their own types.

sdk/storage/azblob/blockblob/responses.go
jhendrixMSFT: Do we actually need to export this though?  Don't we want customers to use the Upload* variants?

sdk/storage/azblob/blob/models.go
mohsha-msft: I am not clear on this one Joel. If I anonymously embed the unexported type, the fields of `downloadOptions` won't be visible, correct?

JeffreyRichter (approved):

sdk/storage/azblob/blob/models.go
jhendrixMSFT: I will say though that if we anonymously embed a common `DownloadOptions` type it does make for a clunky user experience.

jhendrixMSFT:
@mohsha-msft please in the future don't resolve comments that haven't been addressed to the author's satisfaction.  I will open a separate PR to clean up the remaining things shortly.

sdk/storage/azblob/blob/models.go
JeffreyRichter: Perhaps the cleanest/most sustainable (versioning) solution is to define the public Download<Operation>Options struct with public fields with what you want and then when this is passed to the method, create an instance of the common/private struct and manually copy the fields over and then pass the common type along to the next method.

You could use that method I wrote that uses reflection to copy the same fields/types if you wanted to but this may be overkill for just these scenarios.

PR #18896, opened 2022-08-18
Split buffer and file download options into their own types

jhendrixMSFT:
This allows the options for each API to independently evolve over time.
Fixed potential panic in some helper methods.

### DIFF
--- a/sdk/storage/azblob/blob/client.go
+++ b/sdk/storage/azblob/blob/client.go
@@ -269,10 +269,7 @@ func (b *Client) GetSASToken(permissions SASPermissions, start time.Time, expiry
 // Concurrent Download Functions -----------------------------------------------------------------------------------------
 
 // download downloads an Azure blob to a WriterAt in parallel.
-func (b *Client) download(ctx context.Context, writer io.WriterAt, o *downloadOptions) (int64, error) {
-	if o == nil {
-		o = &downloadOptions{}
-	}
+func (b *Client) download(ctx context.Context, writer io.WriterAt, o downloadOptions) (int64, error) {
 	if o.BlockSize == 0 {
 		o.BlockSize = DefaultDownloadBlockSize
 	}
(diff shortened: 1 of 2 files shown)


sdk/storage/azblob/blob/models.go
```diff
@@ -109,7 +109,7 @@ func (o *DownloadStreamOptions) format() (*generated.BlobClientDownloadOptions,
 
 // ---------------------------------------------------------------------------------------------------------------------
 
-// downloadOptions identifies options used by the DownloadBuffer and DownloadFile functions.
+// downloadOptions contain common options used by the DownloadBuffer and DownloadFile functions.
```
JeffreyRichter, line 112:
```suggestion
// downloadOptions contains common options used by the DownloadBuffer and DownloadFile functions.
```

JeffreyRichter (approved):

sdk/storage/azblob/blob/models.go
jhendrixMSFT: Fixed

sdk/storage/azblob/blob/client.go
```diff
@@ -372,24 +369,32 @@ func (b *Client) DownloadStream(ctx context.Context, o *DownloadStreamOptions) (
...
 	// 1. Calculate the size of the destination file
 	var size int64
 
-	count := o.Count
+	count := do.Count
 	if count == CountToEnd {
 		// Try to get Azure blob's size
-		getBlobPropertiesOptions := o.getBlobPropertiesOptions()
+		getBlobPropertiesOptions := do.getBlobPropertiesOptions()
 		props, err := b.GetProperties(ctx, getBlobPropertiesOptions)
```
jhendrixMSFT, line 393: @JeffreyRichter I noticed that we make two API calls in `DownloadFile` which we try to avoid.  In this case it's to get the length of the blob.  Is there another way to achieve this or can we leave it as-is?

jhendrixMSFT: Will follow up separately on this.

JeffreyRichter: To do memory-mapped file writes to a file, the file has to be sized correctly first and then memory-mapped. So, we have to get the size of the blob first, create the file & set its length, memory-map it, and then download into the MMF (in parallel). There are ways to be more clever here like do a single GET block (which includes the size and the data), use the size to create/size file & MMF it and then write the first block while now starting more concurrent GETs but this is more complicated/brittle logic for a perf improvement.

Also, for these high-level APIs that do "magic" I think of them as customers just accepting the "magic". The lower-level APIs exist that do not magic and customers can control the behavior exactly as they want.

PR #18900, opened 2022-08-19
fix reverse tags

Alancere:


eng/tools/generator/cmd/v2/common/changlogProcessor.go
```diff
@@ -48,8 +47,11 @@ func GetAllVersionTags(rpName, namespaceName string) ([]string, error) {
 			tags = append(tags, tag["ref"].(string))
 		}
 	}
-	sort.Sort(sort.Reverse(sort.StringSlice(tags)))
-
+	// reverse tags
```
tadelesh, line 50: We cannot rely on the sequence of the GitHub return. You need to add a customer sort function.

tadelesh (approved):

PR #18904, opened 2022-08-19
Generate test report on test failures

jhendrixMSFT:

### DIFF
--- a/eng/scripts/run_tests.ps1
+++ b/eng/scripts/run_tests.ps1
@@ -10,6 +10,7 @@ Write-Host "##[command] Executing 'go test -timeout $testTimeout -v -coverprofil
 
 go test -timeout $testTimeout -v -coverprofile coverage.txt ./... | Tee-Object -FilePath outfile.txt
 # go test will return a non-zero exit code on test failures so don't skip generating the report in this case
+$GOTESTEXITCODE = $LASTEXITCODE
 
 Get-Content outfile.txt | go-junit-report > report.xml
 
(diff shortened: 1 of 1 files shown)


eng/scripts/run_tests.ps1
```diff
@@ -9,9 +9,7 @@ Push-Location sdk/$serviceDirectory
 Write-Host "##[command] Executing 'go test -timeout $testTimeout -v -coverprofile coverage.txt ./...' in sdk/$serviceDirectory"
 
 go test -timeout $testTimeout -v -coverprofile coverage.txt ./... | Tee-Object -FilePath outfile.txt
-if ($LASTEXITCODE) {
-    exit $LASTEXITCODE
-}
+# go test will return a non-zero exit code on test failures so don't skip generating the report in this case
```
benbp, line 12: Would you still want to capture the non-zero exit code and defer the failure to the end of the script, that way the test task still fails and can be tracked.

jhendrixMSFT: Any failed tests will still cause CI to fail.  E.g. we don't check the exit code for `go test` in other (legacy) repos and things work as expected.

RickWinter: I agree that capturing the non-zero and returning that after we generate the report would be safer.

jhendrixMSFT: It shouldn't be necessary though.  E.g. we don't do this for [`go-autorest`](https://github.com/Azure/go-autorest/blob/main/azure-pipelines.yml#L68) and CI works as expected.

benbp: I agree it's not necessary for failing CI. Capturing failures at the specific step in which they failed is helpful for error tracking, e.g. when we need to formulate historical queries to track when/how a certain step failed. And I find it's also useful to discern quickly whether the tests themselves failed vs. some engsys component of the coverage checking failed. Basically when we lose granularity it makes differentiating engsys vs. language failures a little bit harder in some cases.

RickWinter (approved):

PR #18926, opened 2022-08-23
Azblob Tinyblob Examples

mohsha-msft:

### DIFF
--- a/sdk/storage/azblob/constants.go
+++ b/sdk/storage/azblob/constants.go
@@ -10,10 +10,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 )
 
-const (
-	CountToEnd = 0
-)
-
 // PublicAccessType defines values for AccessType - private (default) or blob or container
 type PublicAccessType = generated.PublicAccessType
 
--- a/sdk/storage/azblob/constants.go
+++ b/sdk/storage/azblob/constants.go
@@ -10,6 +10,10 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 )
 
+const (
+	CountToEnd = 0
+)
+
 // PublicAccessType defines values for AccessType - private (default) or blob or container
 type PublicAccessType = generated.PublicAccessType
 
@@ -34,4 +38,33 @@ const (
 // PossibleDeleteSnapshotsOptionTypeValues returns the possible values for the DeleteSnapshotsOptionType const type.
 func PossibleDeleteSnapshotsOptionTypeValues() []DeleteSnapshotsOptionType {
 	return generated.PossibleDeleteSnapshotsOptionTypeValues()
+}
+
+const (
+	ListBlobsIncludeItemCopy                ListBlobsIncludeItem = "copy"
+	ListBlobsIncludeItemDeleted             ListBlobsIncludeItem = "deleted"
+	ListBlobsIncludeItemMetadata            ListBlobsIncludeItem = "metadata"
+	ListBlobsIncludeItemSnapshots           ListBlobsIncludeItem = "snapshots"
+	ListBlobsIncludeItemUncommittedblobs    ListBlobsIncludeItem = "uncommittedblobs"
+	ListBlobsIncludeItemVersions            ListBlobsIncludeItem = "versions"
+	ListBlobsIncludeItemTags                ListBlobsIncludeItem = "tags"
+	ListBlobsIncludeItemImmutabilitypolicy  ListBlobsIncludeItem = "immutabilitypolicy"
+	ListBlobsIncludeItemLegalhold           ListBlobsIncludeItem = "legalhold"
+	ListBlobsIncludeItemDeletedwithversions ListBlobsIncludeItem = "deletedwithversions"
+)
+
+// PossibleListBlobsIncludeItemValues returns the possible values for the ListBlobsIncludeItem const type.
+func PossibleListBlobsIncludeItemValues() []ListBlobsIncludeItem {
+	return []ListBlobsIncludeItem{
+		ListBlobsIncludeItemCopy,
+		ListBlobsIncludeItemDeleted,
+		ListBlobsIncludeItemMetadata,
+		ListBlobsIncludeItemSnapshots,
+		ListBlobsIncludeItemUncommittedblobs,
+		ListBlobsIncludeItemVersions,
+		ListBlobsIncludeItemTags,
+		ListBlobsIncludeItemImmutabilitypolicy,
+		ListBlobsIncludeItemLegalhold,
+		ListBlobsIncludeItemDeletedwithversions,
+	}
 }
--- a/sdk/storage/azblob/zt_examples_test.go
+++ b/sdk/storage/azblob/zt_examples_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/lease"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/pageblob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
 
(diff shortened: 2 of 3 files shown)


sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
 	handleError(err)
 
 	// Delete the container.
-	_, err = containerClient.Delete(context.TODO(), nil)
+	_, err = client.DeleteContainer(context.TODO(), containerName, nil)
 	handleError(err)
 }
 
-// ---------------------------------------------------------------------------------------------------------------------
+// tiny blob client ----------------------------------------------------------------------------------------------------
```
RickWinter, line 129:
```suggestion
// blob client ----------------------------------------------------------------------------------------------------
```

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -60,46 +61,39 @@ func Example() {
...
-		Metadata: map[string]string{"Foo": "Bar"},
-		Tags:     map[string]string{"Year": "2022"},
-	}
-	_, err = blockBlobClient.Upload(context.TODO(), streaming.NopCloser(strings.NewReader(uploadData)), &blockBlobUploadOptions)
+	blobData := "Hello world!"
+	blobName := "HelloWorld.txt"
+	uploadResp, err := client.UploadStream(context.TODO(),
+		containerName,
+		blobName,
+		streaming.NopCloser(strings.NewReader(blobData)),
```
JeffreyRichter, line 79: We shouldn't need to wrap a NopCloser around this.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	handleError(err)
+
+	serviceClient, err := azblob.NewClient(serviceURL, cred, nil)
+	handleError(err)
+
+	// Perform download
+	containerName := "testcontainer"
+	blobName := "virtual/dir/path/" + destFileName
+	blockSize := 1024
```
JeffreyRichter, line 307: Why don't we kill these variables and just put the values directly on lines 313-316. Do the same for the upload file example above. Keep the examples short. Also, don't explicitly initialize any variables/fields to 0 like doenloadOffset & downloadCount.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
+				fmt.Println(bytesTransferred)
+			},
+		})
+
+	// Assert download was successful
+	handleError(err)
+
+	fileSize := 8 * 1024 * 1024 // assuming blob size is 8MB.
+
+	// Assert downloaded data is consistent
```
JeffreyRichter, line 328: We shouldn't have validation code in the samples. This is not test code; it is sample code. The reader assumes the samples works correctly. If we want test, then let's have _test files.

Again, the samples, must be short and sweet - no extra junk that may confuse a reader.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
+
+	serviceClient, err := azblob.NewClient(serviceURL, cred, nil)
+	handleError(err)
+
+	// Set up test blob
+	containerName := "testcontainer"
+	bufferSize := 8 * 1024 * 1024
+	blobName := "test_upload_stream.bin"
+	blobData := make([]byte, bufferSize)
+	blobContentReader := streaming.NopCloser(bytes.NewReader(blobData))
```
JeffreyRichter, line 403: Shouldn't need the NopCloser

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
+	}
+	serviceURL := fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	handleError(err)
+
+	serviceClient, err := azblob.NewClient(serviceURL, cred, nil)
+	handleError(err)
+
+	// Download the blob to verify
```
JeffreyRichter, line 426: No verification code

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,325 @@ func Example() {
...
+
+	pager := serviceClient.NewListBlobsPager("testcontainer", &azblob.ListBlobsOptions{
+		Include: []azblob.ListBlobsIncludeItem{azblob.ListBlobsIncludeItemVersions, azblob.ListBlobsIncludeItemDeleted},
+	})
+
+	for pager.More() {
+		resp, err := pager.NextPage(ctx)
+		handleError(err) // if err is not nil, break the loop.
+		for _, _blob := range resp.Segment.BlobItems {
+			fmt.Printf("%v", _blob)
```
JeffreyRichter, line 362: Please modify this code to take each blob's URL and pass it to the function that parses the URL into parts and then print just the blob name.  This will demo how to do that and is why we kept that function in the azblob package.

JeffreyRichter (approved):

jhendrixMSFT (changes requested):
Fixes per Jeff's review.

sdk/storage/azblob/constants.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 )
 
+const (
+	CountToEnd = 0
```
jhendrixMSFT, line 14: I don't think this is required.

JeffreyRichter: We can get rid of this and just document that 0 means "to end"

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: `blob client` is different from `tiny blob client`. It basically represents the collection of APIs exposed to general user (not super user).

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: I just wanted to increase the readibility of the examples by writing exactly what each parameter means. But yeah we can remove the variables to reduce the clutter and present small examples.

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: Agreed.

sdk/storage/azblob/constants.go
mohsha-msft: Correct. I've removed `CountToEnd`.

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: Fixed. Does this look good?
```golang
for _, _blob := range resp.Segment.BlobItems {
	u := fmt.Sprintf("https://%s.blob.core.windows.net/%s/%s", accountName, "testcontainer", *_blob.Name)
	blobURLParts, err := azblob.ParseURL(u)
	handleError(err) // if err is not nil, break the loop.
	fmt.Printf("%s", blobURLParts.String())
}
```

sdk/storage/azblob/zt_examples_test.go
mohsha-msft: Yeah. bytes.NewReader anyway is returning `*Reader` back to us.
Removed.

sdk/storage/azblob/zt_examples_test.go
JeffreyRichter: There is no such thing as "tiny blob" to a customer; this is just a term we've been using internally.  So the term shouldn't be leaked exteranlly.

sdk/storage/azblob/zt_examples_test.go
JeffreyRichter: This isn't what I was thinking. I thought that List Blobs returned blob URLs but I jsut checked the docs and I was wrong. So, keep the code the way it was.
But, now I'm not sure we want ParseURL in azblob - leave it here for now and I'll think about it more.

sdk/storage/azblob/zt_examples_test.go
jhendrixMSFT: Please not that this will eventually go one step further.  i.e. the _only_ examples we should have in the `azblob` package are for its contents.  Everything else will be moved to the corresponding sub packages.

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -112,7 +106,7 @@ func Example() {
 	// List methods returns a pager object which can be used to iterate over the results of a paging operation.
 	// To iterate over a page use the NextPage(context.Context) to fetch the next page of results.
 	// PageResponse() can be used to iterate over the results of the specific page.
-	pager := containerClient.NewListBlobsFlatPager(nil)
+	pager := client.NewListBlobsPager(containerName, nil)
 	for pager.More() {
 		resp, err := pager.NextPage(context.TODO())
 		if err != nil {
```
jhendrixMSFT, line 112: Shouldn't line 113 be `handleError(err)` to be consistent with other error handling?

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,301 @@ func Example() {
...
+	serviceClient, err := azblob.NewClient(serviceURL, cred, nil)
+	handleError(err)
+
+	// Perform download
+	// 0 represents the end of the blob.
+	downloadOffset, downloadCount := 0, 0
+
+	_, err = serviceClient.DownloadFile(context.TODO(), "testcontainer", "virtual/dir/path/"+destFileName, destFile,
+		&blob.DownloadFileOptions{
+			Count:       int64(downloadCount),
```
jhendrixMSFT, line 306: `Count` and `Offset` default to zero, so we shouldn't need to explicitly set them for this example (only if we were downloading a subset of the blob).

sdk/storage/azblob/zt_examples_test.go
```diff
@@ -124,15 +118,301 @@ func Example() {
...
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	handleError(err)
+
+	serviceClient, err := azblob.NewClient(serviceURL, cred, nil)
+	handleError(err)
+
+	// Set up test blob
+	containerName := "testcontainer"
+	bufferSize := 8 * 1024 * 1024
```
jhendrixMSFT, line 379: How was this value determined?  What's the guidance?  A comment would help.

jhendrixMSFT: The default is 1MB.  Can we just leave it at that?

jhendrixMSFT (approved):

PR #18999, opened 2022-09-01
[azeventhubs] Adding in the Processor type

richardpark-msft:
Adding in the Processor type for Event Hubs. Callers can use this automatically load balance with other consumers, coordinating using Azure Blob storage.


sdk/messaging/azeventhubs/internal/blob/ci.yml
```diff
@@ -0,0 +1,28 @@
+trigger:
```
jhendrixMSFT, line 1: Presumably this file doesn't need to be here.

sdk/messaging/azeventhubs/internal/blob/doc.go
```diff
@@ -0,0 +1,214 @@
+//go:build go1.18
```
jhendrixMSFT, line 1: Another one for removal.

sdk/messaging/azeventhubs/internal/blob/README.md
```diff
@@ -0,0 +1,397 @@
+# Azure Blob Storage SDK for Go
```
jhendrixMSFT, line 1: This can be removed.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,398 @@
...
+
+	return &blobCheckpointStore{
+		cc: cc,
+	}, nil
+}
+
+// NewBlobCheckpointStore creates a checkpoint store that stores ownership and checkpoints in
+// Azure Blob storage. The container client will be used to store blobs.
+// NOTE: the container must exist before the checkpoint store can be used.
+func NewWithSharedKey(containerURL string, cred *SharedKeyCredential, options *NewOptions) (azeventhubs.CheckpointStore, error) {
```
jhendrixMSFT, line 60: Do we really need this variant for the beta?  We're still trying to work out if we want a centralized `NamedKeyCredential`.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,398 @@
...
+// NewOptions contains optional parameters for the New, NewFromConnectionString and NewWithSharedKey
+// functions
+type NewOptions struct {
+	azcore.ClientOptions
+}
+
+// New creates a checkpoint store that stores ownership and checkpoints in
+// Azure Blob storage. The container client will be used to store blobs.
+// NOTE: the container must exist before the checkpoint store can be used.
+func New(containerURL string, cred azcore.TokenCredential, options *NewOptions) (azeventhubs.CheckpointStore, error) {
```
jhendrixMSFT, line 30: Returning an interface isn't versioning tolerant (plus the idiom is to return by type, accept by interface).  Instead, we should be returning a `*CheckpointStore` type.  It can abstract the underlying checkpoint store implementation (we use this same pattern for `runtime.Poller[T]`.

jhendrixMSFT (changes requested):

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,398 @@
...
+				OwnerID:          po.OwnershipData.OwnerID,
+				LastModifiedTime: *lastModified,
+			},
+		})
+	}
+
+	return ownerships, nil
+}
+
+func (b *blobCheckpointStore) ListCheckpoints(ctx context.Context, fullyQualifiedNamespace string, eventHubName string, consumerGroup string, options *azeventhubs.ListCheckpointsOptions) ([]azeventhubs.Checkpoint, error) {
```
jhendrixMSFT, line 121: Should this return a *runtime.Pager[T] instead? How many ownerships are typical?

sdk/messaging/azeventhubs/consumer_client.go
```diff
@@ -9,8 +9,27 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal/amqpwrap"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal/exported"
+)
+
+type Error = exported.Error
+
+// Code is an error code, usable by consuming code to work with
+// programatically.
+type Code exported.Code
```
jhendrixMSFT, line 22: Was this meant to be a type alias?

sdk/messaging/azeventhubs/internal/blob/CHANGELOG.md
```diff
@@ -0,0 +1,54 @@
+# Release History
```
jhendrixMSFT, line 1: Can be removed.  Also be sure to open a tracking issue that we should move to the refactored `azblob` at some point which should make it easy to remove the unnecessary content (it's ~1MB for a handful of APIs).

sdk/messaging/azeventhubs/partition_client.go
```diff
@@ -56,7 +56,8 @@ type PartitionClient struct {
 
 	offsetExpression string
 
-	links *internal.Links[amqpwrap.AMQPReceiverCloser]
+	//links *internal.Links[amqpwrap.AMQPReceiverCloser]
```
jhendrixMSFT, line 59: Superfluous comment?

sdk/messaging/azeventhubs/processor_partition_client.go
```diff
@@ -0,0 +1,55 @@
...
+// NOTE: This type is instantiated using the Processor type, which handles dynamic load balancing.
+type ProcessorPartitionClient struct {
+	partitionID           string
+	innerClient           *PartitionClient // *azeventhubs.PartitionClient
+	checkpointStore       CheckpointStore
+	cleanupFn             func()
+	consumerClientDetails consumerClientDetails
+}
+
+// ReceiveEvents receives events until the context has expired or been cancelled.
```
jhendrixMSFT, line 18: Does the `count` param mean this `receives the number of events...`?

sdk/messaging/azeventhubs/processor.go
```diff
@@ -0,0 +1,349 @@
...
+	ProcessorStrategyBalanced ProcessorStrategy = "balanced"
+
+	// ProcessorStrategyGreedy will attempt to claim as many partitions at a time as it can, ignoring
+	// balance.
+	ProcessorStrategyGreedy ProcessorStrategy = "greedy"
+)
+
+// NewProcessorOptions are the options for the NewProcessor
+// function.
+type NewProcessorOptions struct {
```
jhendrixMSFT, line 30: Can you add descriptions for the default values for these fields?

sdk/messaging/azeventhubs/internal/blob/ci.yml
richardpark-msft: Done. Also, removed all the .md files.

sdk/messaging/azeventhubs/internal/blob/README.md
richardpark-msft: Also gone!

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Changed, now returning BlobCheckpointStore instead.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: 32, max. So probably not worth a pager.

richardpark-msft: With that said, one change @jsquire  and I have been talking about is just getting the checkpoint by name, rather than listing them all, as that's the more common case.

sdk/messaging/azeventhubs/consumer_client.go
richardpark-msft: Gah, yes. Fixed.

sdk/messaging/azeventhubs/partition_client.go
richardpark-msft: Gone

sdk/messaging/azeventhubs/processor_partition_client.go
richardpark-msft: Fixed. Made it more clear, and also fixed PartitionClient, which has the same function.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Removed, for now

sdk/messaging/azeventhubs/processor.go
richardpark-msft: Added.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,384 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &BlobCheckpointStore{
+		cc: cc,
+	}, nil
+}
+
+func (b *BlobCheckpointStore) ClaimOwnership(ctx context.Context, partitionOwnership []azeventhubs.Ownership, options *azeventhubs.ClaimOwnershipOptions) ([]azeventhubs.Ownership, error) {
```
jhendrixMSFT, line 62: Doc comment.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,384 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs"
+	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal/blob"
+)
+
+// BlobCheckpointStore is a CheckpointStore implementation that uses Azure Blob storage.
+type BlobCheckpointStore struct {
```
jhendrixMSFT, line 22: I feel like the name is a little long and redundant.

`checkpointstore.BlobCheckpointStore`

`checkpointstore.Blob` would be more concise but I get that just `Blob` is a little weird.  Is there some other name we can use so that we can remove the redundant suffix?

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
```diff
@@ -0,0 +1,384 @@
...
+// NewOptions contains optional parameters for the New, NewFromConnectionString and NewWithSharedKey
+// functions
+type NewOptions struct {
+	azcore.ClientOptions
+}
+
+// New creates a checkpoint store that stores ownership and checkpoints in
+// Azure Blob storage. The container client will be used to store blobs.
+// NOTE: the container must exist before the checkpoint store can be used.
+func New(containerURL string, cred azcore.TokenCredential, options *NewOptions) (*BlobCheckpointStore, error) {
```
jhendrixMSFT, line 35: This should be `NewBlobCheckpointStore()` or whatever name we end up landing on.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Blob feels good to me since it's under the checkpointstore package already. Let's go with that one.

richardpark-msft: Hm...maybe BlobStore?

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Yeah, I'll get these and also add it to the interface.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: NewBlobStore seems okay.

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Okay, renamed to:

checkpoints.BlobStore

sdk/messaging/azeventhubs/checkpointstore/blob_checkpoint_store.go
richardpark-msft: Added.

sdk/messaging/azeventhubs/checkpoints/blob_store.go
```diff
@@ -0,0 +1,390 @@
...
+	offsetStr, ok := metadata["offset"]
+
+	if !ok || offsetStr == nil {
+		return azeventhubs.CheckpointData{}, errors.New("offset is missing from metadata")
+	}
+
+	offset, err := strconv.ParseInt(*offsetStr, 10, 64)
+
+	if err != nil {
+		return azeventhubs.CheckpointData{}, fmt.Errorf("offset could not be parsed as an int64: %w", err)
```
jhendrixMSFT, line 337: I don't think you want `%w` here as that exposes the underlying error type to callers.

jhendrixMSFT: Generally speaking, I think the default position is that `%w` is the wrong thing to do, unless you really intend to expose the underlying error type.

sdk/messaging/azeventhubs/processor_load_balancer.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// extraPartitionPossible is true if the partitions cannot split up evenly
+	// amongst all the known consumers.
+	extraPartitionPossible bool
+
+	raw []Ownership
+}
+
+// loadBalance calls through to the user's configured load balancing algorithm.
+func (lb *processorLoadBalancer) LoadBalance(ctx context.Context, partitionIDs []string) ([]Ownership, error) {
```
jhendrixMSFT, line 55: Is there some prior art here, i.e. are the algorithms adapted from some other implementation or is this all new stuff?

richardpark-msft: Prior art - these are the two algorithms we have in the other track 2 libraries.

sdk/messaging/azeventhubs/checkpoints/blob_store.go
richardpark-msft: Ah, good point. Just habit when wrapping an error :)

richardpark-msft: Fixed.

sdk/messaging/azeventhubs/processor_load_balancer.go
```diff
@@ -0,0 +1,247 @@
...
+// consumers so each one has an even amount (or even + 1, if the # of consumers and #
+// of partitions doesn't divide evenly).
+//
+// NOTE: the checkpoint store itself does not have a concept of 'presence' that doesn't
+// ALSO involve owning a partition. It's possible for a consumer to get boxed out for a
+// bit until it manages to steal at least one partition since the other consumers don't
+// know it exists until then.
+func (lb *processorLoadBalancer) balancedLoadBalancer(ctx context.Context, lbinfo loadBalancerInfo) *Ownership {
+	if len(lbinfo.unownedOrExpired) > 0 {
+		idx := rand.Intn(len(lbinfo.unownedOrExpired))
```
jhendrixMSFT, line 211: Is the seed being set somewhere?  I didn't see it in this file.

jhendrixMSFT (approved):

sdk/messaging/azeventhubs/processor_load_balancer.go
richardpark-msft: Created some rand.Rand's rather than seeding the global one. This does mean the local rand.Rand's aren't thread safe, but in the context they are both used it is fine.

Also, added some comments just in case, to warn people.

PR #19010, opened 2022-09-01
Move SAS-related functionality to the sas package

jhendrixMSFT:
SAS functionality has been moved out of internal/exported to the sas
package.  All type aliases to the former types have been removed.
Removed "SAS" prefix from funcs and types in the sas package.
URL parsing has been moved out of internal/exported to the blob package.


jhendrixMSFT:
@JeffreyRichter I couldn't find the free-standing func for creating a SAS URL in track 1.5.  Do you know where it is?

sdk/storage/azblob/blob/client.go
```diff
@@ -228,10 +229,10 @@ func (b *Client) CopyFromURL(ctx context.Context, copySource string, options *Co
 
 // GetSASToken is a convenience method for generating a SAS token for the currently pointed at blob.
 // It can only be used if the credential supplied during creation was a SharedKeyCredential.
-func (b *Client) GetSASToken(permissions SASPermissions, start time.Time, expiry time.Time) (string, error) {
-	urlParts, _ := exported.ParseURL(b.URL())
+func (b *Client) GetSASToken(permissions sas.BlobPermissions, start time.Time, expiry time.Time) (string, error) {
+	urlParts, _ := ParseURL(b.URL())
```
JeffreyRichter, line 233: I thought we were putting this function in the tiny blob (azblob) package?

JeffreyRichter (approved):

JeffreyRichter:
> @JeffreyRichter I couldn't find the free-standing func for creating a SAS URL in track 1.5. Do you know where it is?

It's a method on this struct: https://github.com/Azure/azure-storage-blob-go/blob/a95cc9323b2c1273a3e088be4a520d4a95a40dd8/azblob/sas_service.go#L12

Method: https://github.com/Azure/azure-storage-blob-go/blob/a95cc9323b2c1273a3e088be4a520d4a95a40dd8/azblob/sas_service.go#L45

sdk/storage/azblob/blob/client.go
jhendrixMSFT: Unfortunately, this will introduce a circular dependency as the SAS-parsing functionality in the `sas` package requires this type.  I think we'll need to move `ParseURL()` and its dependent types to `sas` and can type-alias them from other packages as needed.

JeffreyRichter: Is there another place we can put ParseURL to avoid the circular dep issue? It's not really a SAS thing at all.

JeffreyRichter: Yes, I think that's the better choice

jhendrixMSFT: I suppose we just place all of this in `internal/exported` and type-alias the relevant parts.  But before I do that, can you take a look at this part of the `URLParts` definitnion.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/exported/blob_url.go#L37-L38

This looks a little clunky to me in that we have one field for SAS-specific content (which is immutable no less, although the fields in the parent aren't) and an "unparsed" field for the other query parameters.

jhendrixMSFT: This has been fixed.

PR #19018, opened 2022-09-03
[azeventhubs] Some bug fixes, changelog update

richardpark-msft:
Fixed some bugs, and updated the changelog

- EventDataBatches weren't getting the partition key filled out for the individual messages
- Offset wasn't getting properly filled out (it comes back as a string, not an int64!)
- Embedded EventData in ReceivedEventData, so it's easier to resend an event if you want to.


sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,36 @@
...
+    batch, err := producerClient.NewEventDataBatch(context.TODO(), &azeventhubs.NewEventDataBatchOptions{
+		  PartitionKey: to.Ptr("partition key"),
+	  })
+    ```
+
 ### Bugs Fixed
 
-### Other Changes
+- ReceivedEventData.Offset was being incorrectly parsed, resulting in it always being 0.
+- Added some missing fields to ReceivedEventData and EventData (CorrelationID)
```
RickWinter, line 32:
```suggestion
- Added missing fields to ReceivedEventData and EventData (CorrelationID)
```

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,36 @@
...
+    ```go
+    batch, err := producerClient.NewEventDataBatch(context.TODO(), &azeventhubs.NewEventDataBatchOptions{
+		  PartitionKey: to.Ptr("partition key"),
+	  })
+    ```
+
 ### Bugs Fixed
 
-### Other Changes
+- ReceivedEventData.Offset was being incorrectly parsed, resulting in it always being 0.
```
RickWinter, line 31:
```suggestion
- ReceivedEventData.Offset was incorrectly parsed, resulting in it always being 0.
```

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,36 @@
...
+  in [example_processor_test.go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azeventhubs/example_processor_test.go).
+
 ### Breaking Changes
 
+- In the first beta, ConsumerClient took constructor parameter that required a partition ID, which meant you had to create
+  multiple ConsumerClients if you wanted to consume multiple partitions. ConsumerClient can now create multiple PartitionClient
+  instances (using ConsumerClient.NewPartitionClient), which allows you to share the same AMQP connection and receive from multiple
+  partitions simultaneously.
+- Changes to EventData/ReceivedEventData:
+  - ReceivedEventData now embeds EventData for fields common between the two, making it easier to change and resend, if wanted.
```
RickWinter, line 18:
```suggestion
  - ReceivedEventData now embeds EventData for fields common between the two, making it easier to change and resend.
```

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,36 @@
...
+  - ApplicationProperties have been renamed to Properties.
```
RickWinter, line 19:
```suggestion
  - `ApplicationProperties` renamed to `Properties`.
```

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,36 @@
...
+  - PartitionKey has been removed from EventData. To send events using a PartitionKey you need to set it in the options
```
RickWinter, line 20:
```suggestion
  - `PartitionKey` removed from `EventData`. To send events using a PartitionKey you must set it in the options
```

sdk/messaging/azeventhubs/event_data.go
```diff
@@ -21,42 +22,26 @@ type EventData struct {
 	// the format of Content-Type, specified by RFC2045 (ex: "application/json").
 	ContentType *string
 
+	// CorrelationID is a client-specific id that can be used to mark or identify messages
+	// between clients.
+	// The type of CorrelationID can be a uint64, UUID, []byte, or string
```
RickWinter, line 27:
```suggestion
	// CorrelationID can be a uint64, UUID, []byte, or string
```

sdk/messaging/azeventhubs/event_data.go
```diff
@@ -127,35 +109,60 @@ func newReceivedEventData(amqpMsg *amqp.Message) *ReceivedEventData {
...
-		}
+	updateFromAMQPAnnotations(amqpMsg, re)
+	return re
+}
 
-		if partitionKey, ok := amqpMsg.Annotations[partitionKeyAnnotation]; ok {
-			re.PartitionKey = to.Ptr(partitionKey.(string))
-		}
+// the "SystemProperties" in an EventData are any annotations that are
+// NOT available at the top level as normal fields. So excluing  sequence
```
RickWinter, line 127:
```suggestion
// NOT available at the top level as normal fields. So excluding sequence
```

sdk/messaging/azeventhubs/producer_client_test.go
```diff
@@ -75,35 +76,44 @@ func TestNewProducerClient_GetEventHubsProperties(t *testing.T) {
...
-	producer, err := azeventhubs.NewProducerClientFromConnectionString(testParams.ConnectionString, testParams.EventHubName, nil)
-	require.NoError(t, err)
-
-	batch, err := producer.NewEventDataBatch(context.Background(), nil)
-	require.NoError(t, err)
+	// there are two ways to "send to any" partition
+	// 1. Don't specify a partition ID or a partition key when creating the batch
+	// 2. Specify a partition key. This is useful if you want to send events and have them
+	//  be placed into the same partition but let the overall distribution of the partition keys
+	// happen through Event Hubs.
```
RickWinter, line 83:
```suggestion
	//    be placed into the same partition but let the overall distribution of the partition keys
	//    happen through Event Hubs.
```

RickWinter (approved):

sdk/messaging/azeventhubs/event_data_batch.go
```diff
@@ -77,15 +77,14 @@ func (mb *EventDataBatch) toAMQPMessage() *amqp.Message {
 	mb.batchEnvelope.Format = batchMessageFormat
 
 	if mb.partitionKey != nil {
-		mb.batchEnvelope.Annotations[partitionKeyAnnotation] = *mb.partitionKey
-	}
+		if mb.batchEnvelope.Annotations == nil {
+			mb.batchEnvelope.Annotations = make(amqp.Annotations)
+		}
 
-	if mb.partitionID != nil {
```
jhendrixMSFT, line 83: Why was this removed?

sdk/messaging/azeventhubs/event_data.go
```diff
@@ -127,35 +109,60 @@ func newReceivedEventData(amqpMsg *amqp.Message) *ReceivedEventData {
...
+			if asString, ok := v.(string); ok {
+				dest.PartitionKey = to.Ptr(asString)
+			}
+		case enqueuedTimeAnnotation:
+			if asTime, ok := v.(time.Time); ok {
+				dest.EnqueuedTime = &asTime
+			}
+		case offsetNumberAnnotation:
+			if offsetStr, ok := v.(string); ok {
+				if offset, err := strconv.ParseInt(offsetStr, 10, 64); err == nil {
```
jhendrixMSFT, line 156: What happens if `err != nil`?

sdk/messaging/azeventhubs/event_data.go
```diff
@@ -127,35 +109,60 @@ func newReceivedEventData(amqpMsg *amqp.Message) *ReceivedEventData {
...
+		if !keyIsString {
+			continue
 		}
 
-		if offset, ok := amqpMsg.ApplicationProperties[offsetNumberAnnotation]; ok {
-			if offsetInt64, ok := offset.(int64); ok {
-				re.Offset = to.Ptr(int64(offsetInt64))
+		switch keyStr {
+		case sequenceNumberAnnotation:
+			if asInt64, ok := v.(int64); ok {
```
jhendrixMSFT, line 143: Will this always be `int64`?  What happens if the type-assertion fails?

richardpark-msft: If it's not an int64 then we just leave SequenceNumber uninitialized (ie, 0)

This is relying on the go-amqp type promotion always being consistent. If we think that'll swap around (or is an implementation detail) then we could do an `int-like` type thing and just cycle through the various int types.

The reality is that this value does have a range all the way up to int64, however.

sdk/messaging/azeventhubs/event_data_batch.go
richardpark-msft: It turns out it's not needed. 

We already route the batch sending to the proper link if the user explicitly chooses a partition for the batch _or_ we send it to the general link for hub otherwise (and then there'd be no partition id).

sdk/messaging/azeventhubs/event_data.go
richardpark-msft: Same here. If they send us invalid data we just leave Offset uninitialized (ie, 0)

richardpark-msft: We return an error now. It'll abort the Receive operation but this isn't as catastrophic as it would be for service bus.

sdk/messaging/azeventhubs/event_data.go
richardpark-msft: We do an int check now, so we can promote up from the smallest int (int8) to the biggest (int64). If it doesn't convert (ie, not an int type) then we just return an error, like we do for Offset.

jhendrixMSFT (approved):

PR #19057, opened 2022-09-07
Add completed as teminal state in sdk/azcore

duncan485:
When working with the sdk/resourcemanager/costmanagement/armcostmanagement v1.0.0 package I encountered a bug when polling the result of NewGenerateDetailedCostReportClient.BeginCreateOperation.

When the report is succesfully generated by this function, it returns a status of 'Completed', however, 'Completed is not recognized as a terminal state in the underlying sdk/azcore package. This leads to an infinite loop in the PollUntilDone function. The bug can be replicated by taking the example of NewGenerateDetailedCostReportClient.BeginCreateOperation from [docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/costmanagement/armcostmanagement#GenerateDetailedCostReportClient.BeginCreateOperation)

This PR adds 'Completed' as a recognized terminal state.

```
package main

import (
	"context"
	"log"

	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/costmanagement/armcostmanagement"
)

func main() {
	if err != nil {
		log.Fatalf("failed to obtain a credential: %v", err)
	}
	ctx := context.Background()
	client, err := armcostmanagement.NewGenerateDetailedCostReportClient(cred, nil)
	if err != nil {
		log.Fatalf("failed to create client: %v", err)
	}
	poller, err := client.BeginCreateOperation(ctx,
		"providers/Microsoft.Billing/billingAccounts/12345",
		armcostmanagement.GenerateDetailedCostReportDefinition{
			BillingPeriod: to.Ptr("202008"),
			Metric:        to.Ptr(armcostmanagement.GenerateDetailedCostReportMetricTypeActualCost),
		},
		nil)
	if err != nil {
		log.Fatalf("failed to finish the request: %v", err)
	}
	res, err := poller.PollUntilDone(ctx, nil) // INFINITE LOOP HERE
	if err != nil {
		log.Fatalf("failed to pull the result: %v", err)
	}
	// TODO: use response item
	_ = res
}
```

- [ x] The purpose of this PR is explained in this or a referenced issue.
- [ x] The PR does not update generated files.
- [x ] Tests are included and/or updated for code changes.

### DIFF
--- a/sdk/azcore/internal/pollers/util.go
+++ b/sdk/azcore/internal/pollers/util.go
@@ -23,7 +23,7 @@ import (
 
 // the well-known set of LRO status/provisioning state values.
 const (
-	StatusCompleteed = "Completed"
+	StatusCompleted  = "Completed"
 	StatusSucceeded  = "Succeeded"
 	StatusCanceled   = "Canceled"
 	StatusFailed     = "Failed"
@@ -32,7 +32,7 @@ const (
 
 // IsTerminalState returns true if the LRO's state is terminal.
 func IsTerminalState(s string) bool {
-	return strings.EqualFold(s, StatusSucceeded) || strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled) || strings.EqualFold(s, StatusCompleteed)
+	return strings.EqualFold(s, StatusSucceeded) || strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled) || strings.EqualFold(s, StatusCompleted)
 }
 
 // Failed returns true if the LRO's state is terminal failure.
@@ -47,7 +47,7 @@ func Succeeded(s string) bool {
 
 // Completed return true if is terminal completed
 func Completed(s string) bool {
-	return strings.EqualFold(s, StatusCompleteed)
+	return strings.EqualFold(s, StatusCompleted)
 }
 
 // returns true if the LRO response contains a valid HTTP status code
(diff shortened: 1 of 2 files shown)


sdk/azcore/internal/pollers/util.go
```diff
@@ -23,6 +23,7 @@ import (
 
 // the well-known set of LRO status/provisioning state values.
 const (
+	StatusCompleted  = "Completed"
```
jhendrixMSFT, line 26: While I'm not opposed to adding this, I believe this is in violation of the [ARM RPC spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#provisioningstate-property).  Let me follow up on that.

duncan485: Thanks, I was unaware of the spec. Just encountered the bug and tried to solve it this way :) . Let me know if there is a more elegant solution.

duncan485: Hi @jhendrixMSFT , any update on this? 

If there is a better way we can fix the issue please let me know.

jhendrixMSFT: Sorry for the delay, I just pinged the owners again to get their input on this.

jhendrixMSFT: The service is in error here, it shouldn't be returned `Completed`.  I'm following up with the service team to get this fixed.

PR #19059, opened 2022-09-07
Simplify blob list options

jhendrixMSFT:
Use a struct of bools instead of a slice of options.

### DIFF
--- a/sdk/storage/azblob/container/models.go
+++ b/sdk/storage/azblob/container/models.go
@@ -7,6 +7,8 @@
 package container
 
 import (
+	"reflect"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/exported"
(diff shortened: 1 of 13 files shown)


sdk/storage/azblob/container/models.go
```diff
@@ -98,10 +100,56 @@ func (o *GetPropertiesOptions) format() (*generated.ContainerClientGetProperties
 
 // ---------------------------------------------------------------------------------------------------------------------
 
+// ListBlobsDetail indicates what additional information the service should return with each blob.
+type ListBlobsDetail struct {
```
jhendrixMSFT, line 104: This is the type-name from track 1.5, I'm not a huge fan.  What about `ListBlobsInclude`?  Something else?

JeffreyRichter (approved):

sdk/storage/azblob/container/models.go
JeffreyRichter: Sure, that's fine

PR #19061, opened 2022-09-07
Use azcore.ETag instead of string

jhendrixMSFT:

### DIFF
--- a/sdk/storage/azblob/blob/client.go
+++ b/sdk/storage/azblob/blob/client.go
@@ -362,14 +362,10 @@ func (b *Client) DownloadStream(ctx context.Context, o *DownloadStreamOptions) (
 		count = *o.Count
 	}
 
-	eTag := azcore.ETag("")
-	if dr.ETag != nil {
-		eTag = *dr.ETag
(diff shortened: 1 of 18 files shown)


sdk/storage/azblob/blob/client.go
```diff
@@ -362,14 +362,14 @@ func (b *Client) DownloadStream(ctx context.Context, o *DownloadStreamOptions) (
 		count = *o.Count
 	}
 
-	eTag := ""
+	eTag := azcore.ETag("")
```
jhendrixMSFT, line 365: This looks dubious.  Shouldn't we default to a `nil` ETag?

JeffreyRichter: Great catch. Yes make it nil. My guess is that it would work anyway but nil is technically correct.

JeffreyRichter (approved):

sdk/storage/azblob/blob/client.go
jhendrixMSFT: Fixed

Issue #18786, opened 2022-08-04
Stop providing snippets in markdown


RickWinter:
All snippets and samples should be compiled to ensure correctness.  The Go tooling provides this validation when the code is contained in documentation examples.  

For each package, we need to ensure

1. Readme samples are references to Examples in documentation on pkg.go.dev
2. Example code in markdown is removed.
3. All Champion scenarios have corresponding examples in doumentation.

jhendrixMSFT:
I agree with this.  However, I think there's value in having very targeted _examples_ (not end-to-end samples) inline with the docs.

That said, we should take this one step further.  Remove all samples/docs from README.md and keep things in `doc.go` or the like ([example](https://github.com/golang/go/blob/master/src/net/net.go)).

PR #19075, opened 2022-09-08
Clean up README

jhendrixMSFT:
Used the same template as other languages.
Simplified examples and provided links to end-to-end examples. Added initial doc blurb on specialized clients.

### DIFF
--- a/sdk/storage/azblob/README.md
+++ b/sdk/storage/azblob/README.md
@@ -1,443 +1,276 @@
 # Azure Blob Storage SDK for Go
 
-## Introduction
+> Server Version: 2020-10-02
 
-The Microsoft Azure Storage SDK for Go allows you to build applications that takes advantage of Azure's scalable cloud
-storage. This is the new beta client module for Azure Blob Storage, which follows
-our [Azure SDK Design Guidelines for Go](https://azure.github.io/azure-sdk/golang_introduction.html) and replaces the
-previous beta [azblob package](https://github.com/azure/azure-storage-blob-go).
+Azure Blob storage is Microsoft's object storage solution for the cloud. Blob
+storage is optimized for storing massive amounts of unstructured data.
+Unstructured data is data that does not adhere to a particular data model or
+definition, such as text or binary data.
 
-## Getting Started
+[Source code][source] | [API reference documentation][docs] | [REST API documentation][rest_docs] | [Product documentation][product_docs]
 
-The Azure Blob SDK can access an Azure Storage account.
-
-### Prerequisites
-
-* Go versions 1.18 or higher.
-* You must have an [Azure storage account][azure_storage_account].
-* You can also use the [Azure Cloud Shell](https://shell.azure.com/bash) to create one with following commands (
-  replace `my-resource-group`
-  and `mystorageaccount` with your own unique names):
-  (Optional) if you want a new resource group to hold the Storage Account:
-  ```pwsh
-  az group create --name my-resource-group --location westus2
-  ```
-  Create the storage account:
-  ```pwsh
-  az storage account create --resource-group my-resource-group --name mystorageaccount
-  ```
-
-  The storage account name can be queried with:
-  ```pwsh
-  az storage account show -n mystorageaccount -g my-resource-group --query "primaryEndpoints.blob"
-  ```
-  You can set this as an environment variable with:
-  ```pwsh
-  # PowerShell
-  $ENV:AZURE_STORAGE_ACCOUNT_NAME="mystorageaccount"
-  ```
-
-  ```bash
-  # bash
-  export AZURE_STORAGE_ACCOUNT_NAME="mystorageaccount"
-  ```
-
-  Query your storage account keys:
-  ```pwsh
-  az storage account keys list --resource-group my-resource-group -n mystorageaccount
-  ```
-
-  Output:
-  ```json
-  [
-      {
-          "creationTime": "2022-02-07T17:18:44.088870+00:00",
-          "keyName": "key1",
-          "permissions": "FULL",
-          "value": "..."
-      },
-      {
-          "creationTime": "2022-02-07T17:18:44.088870+00:00",
-          "keyName": "key2",
-          "permissions": "FULL",
-          "value": "..."
-      }
-  ]
-  ```
-
-  ```pwsh
-  # PowerShell
-  $ENV:AZURE_STORAGE_ACCOUNT_KEY="<mystorageaccountkey>"
-  ```
-  ```bash
-  # Bash
-  export AZURE_STORAGE_ACCOUNT_KEY="<mystorageaccountkey>"
-  ```
-  > You can obtain your account key from the Azure Portal under the "Access Keys" section on the left-hand pane of your
-  storage account.
-
-#### Create account
-
-* To create a new Storage account, you can use [Azure Portal][azure_portal_create_account]
-  , [Azure PowerShell][azure_powershell_create_account], or [Azure CLI][azure_cli_create_account].
+## Getting started
 
 ### Install the package
 
-* Install the Azure Blob Storage client module for Go with `go get`:
+Install the Azure Blob Storage SDK for Go with [go get][goget]:
 
-```bash
+```Powershell
 go get github.com/Azure/azure-sdk-for-go/sdk/storage/azblob
 ```
 
-> Optional: If you are going to use AAD authentication, install the `azidentity` package:
-
-```bash
+If you're going to authenticate with Azure Active Directory (recommended), install the [azidentity][azidentity] module.
+```Powershell
 go get github.com/Azure/azure-sdk-for-go/sdk/azidentity
 ```
 
-#### Create the client
+### Prerequisites
 
-`azblob` allows you to interact with three types of resources :-
+A supported [Go][godevdl] version (the Azure SDK supports the two most recent Go releases).
 
-* [Azure Storage Accounts][azure_storage_account].
-* [Containers](https://azure.microsoft.com/en-in/overview/what-is-a-container/#overview) within those storage accounts.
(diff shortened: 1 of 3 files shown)


sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
+- One ore more _blobs_ in a container
 
-```golang
-import "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+Instances of the `azblob.Client` type provide methods for manipulating containers and blobs within a storage account.
+The storage account is specified when the `azblob.Client` is constructed.
+Use the appropriate client constructor function for the authentication mechanism you wish to use.
 
-serviceURL := fmt.Sprintf("https://%s.blob.core.windows.net/", accountName)
+Learn more about options for authentication _(including Connection Strings, Shared Key, Shared Key Signatures, Active Directory, and anonymous public access)_ [in our examples.](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/examples_test.go)
```
richardpark-msft, line 79: Minor: I didn't find those terms in the examples file, so it's hard to find the right thing if I got there from this paragraph.

Might be worth adding some comments so they're ctrl-f'able.

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
 
-cred, err := service.NewSharedKeyCredential(accountName, accountKey)
-handleError(err)
+### Goroutine safety
```
richardpark-msft, line 81: This is a nice section. I should add this to the messaging packages.

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
-well. To do this, pass the connection string to the client's `NewServiceClientFromConnectionString` method. The
-connection string can be found in your storage account in the [Azure Portal][azure_portal_account_url] under the "Access
-Keys" section or with the following Azure CLI command:
+### Uploading a blob
 
-```bash
-az storage account show-connection-string -g my-resource-group -n mystorageaccount
-```
+```go
+const (
```
richardpark-msft, line 100: These'll migrate into their own example files, rather than being inlined at some point, right?

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
-```golang
-import "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+// authenticate with Azure Active Directory
+cred, err := azidentity.NewDefaultAzureCredential(nil)
+// TODO: handle error
 
-connectionString := "DefaultEndpointsProtocol=https;AccountName=<myAccountName>;AccountKey=<myAccountKey>;EndpointSuffix=core.windows.net"
+// create a client for the specified storage account
+client, err := azblob.NewClient(account, cred, nil)
+// TODO: handle error
```
richardpark-msft, line 113: For better or worse, I think we've agreed to if(err) {panic(err) } in samples/snippets.

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
-_, err = containerClient.Delete(context.TODO(), nil)
-handleError(err)
 ```
 
 ## Troubleshooting
 
-### Error Handling
+All Blob service operations will return an
+[*azcore.ResponseError][azcore_response_error] on failure with a
+populated `ErrorCode` field.  Many of these errors are recoverable.
```
richardpark-msft, line 178:
```suggestion
populated `ErrorCode` field. Many of these errors are recoverable.
```

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
-var responseErr *azcore.ResponseError
-errors.As(err, &responseErr)
+// try to delete the container, avoiding any potential race conditions with an in-progress or completed deletion
+_, err = client.DeleteContainer(context.TODO(), containerName, nil)
 
-if responseErr != nil {
-fmt.Println("ErrorCode: " + responseErr.ErrorCode + ", StatusCode: " + strconv.Itoa(responseErr.StatusCode))
-} else {
-fmt.Println(err.Error())
+if respErr := &(azcore.ResponseError{}); errors.As(err, &respErr) {
```
richardpark-msft, line 195: Ha, JeffR showed me this trick the other day. It looks odd, but it's nice for scoping.

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
+	if respErr.ErrorCode == string(bloberror.ContainerBeingDeleted) || respErr.ErrorCode == string(bloberror.ContainerNotFound) {
```
richardpark-msft, line 196: Future: should the bloberror's just be strings then? If the user is going to have to cast them each time then we're not getting much utility out of them being string-ish.

sdk/storage/azblob/README.md
```diff
@@ -1,443 +1,276 @@
...
-azlog.SetClassifications(azlog.Request, azlog.Response)
-```
+- [appendblob][append_blob] - [REST docs](https://docs.microsoft.com/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-append-blobs)
+- [blockblob][block_blob] - [REST docs](https://docs.microsoft.com/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-block-blobs)
+- [pageblob][page_blob] - [REST docs](https://docs.microsoft.com/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-page-blobs)
 
-> CAUTION: logs from credentials contain sensitive information.
-> These logs must be protected to avoid compromising account security.
->
+The [blob][blob] client contains APIs common to all blob types.  This includes APIs for deleting and undeleting a blob, setting metadata, and more.
```
richardpark-msft, line 218:
```suggestion
The [blob][blob] package contains APIs common to all blob types.  This includes APIs for deleting and undeleting a blob, setting metadata, and more.
```
(the link just takes me the package folder so I'm assuming you meant package, like the 'lease' package below?)

richardpark-msft: Or should these point to the pkg.go.dev docs instead?

richardpark-msft (approved):
Minor stuff. Fix what makes sense, file issues for remainder.

sdk/storage/azblob/README.md
jhendrixMSFT: I think having bare-bones, inline examples are still useful as you can demonstrate the API without any ceremony (I stole this idea from the C# version that I copy/pasta'ed).  Full end-to-end samples should go in source files.

sdk/storage/azblob/README.md
jhendrixMSFT: I think for examples we should omit it (less ceremony for brevity).  For end-to-end samples I agree that `panic()` is good as you get a stack trace.

sdk/storage/azblob/README.md
jhendrixMSFT: It is nice but I hate it :p.  I want to introduce a helper, `bloberror.Is(err error, codes ...Codes) bool` to simplify this.

sdk/storage/azblob/README.md
jhendrixMSFT: Yep it's ugly which is why I want to introduce a helper.  Or we could introduce an error type for `azblob` that has the `ErrorCode` field as the appropriate type.

sdk/storage/azblob/README.md
jhendrixMSFT: Agreed that this should point to `pkg.go.dev`.

For the wording, I was trying to distinguish between the fact that `blob` exposes a client and that `lease` exposes multiple things.  This might not age well though, so I'll update them all to be consistent.

sdk/storage/azblob/README.md
richardpark-msft: (+1 to the second idea. Similar to what we do for log, with the log.Event type?)

sdk/storage/azblob/README.md
jhendrixMSFT: Good idea!

sdk/storage/azblob/README.md
richardpark-msft: This is something we should standardize. I think we're doing things differently in this area and it's a little unclear.

richardpark-msft: (I'll file an issue for future discussion)

sdk/storage/azblob/README.md
jhendrixMSFT: Ugh ok I remember that I pointed to the directories because this module hasn't been published yet.  Will have to update this later.

jhendrixMSFT: I did for all except `lease` and `sas` as they don't have a single `client.go` file.

sdk/storage/azblob/README.md
jhendrixMSFT: Helper has been added and the example updated.

richardpark-msft: I don't see the update?

jhendrixMSFT: Oops just pushed it.

sdk/storage/azblob/README.md
jhendrixMSFT: I added the appropriate comments and one new example.  I don't have one for SAS at present though, to be added later.

sdk/storage/azblob/README.md
jhendrixMSFT: Added a comment about this in https://github.com/Azure/azure-sdk-for-go/issues/18786

jhendrixMSFT: In fact, I might try migrating this to a `doc.go` file in a separate PR.

PR #19082, opened 2022-09-09
[keyvault/internal] Verify challenge resource matches request domain

chlowell:


sdk/keyvault/internal/challenge_policy.go
```diff
@@ -170,17 +171,27 @@ func (k *KeyVaultChallengePolicy) findScopeAndTenant(resp *http.Response) error
...
+	parsed, err := url.Parse(scope)
+	if err != nil {
+		return &challengePolicyError{err: fmt.Errorf(`invalid challenge resource "%s": %v`, scope, err)}
+	}
+	if !strings.HasSuffix(req.URL.Host, "."+parsed.Host) {
+		return &challengePolicyError{err: fmt.Errorf(`the challenge resource "%s" doesn't match the requested domain`, scope)}
+	}
+	if !strings.HasSuffix(scope, "/.default") {
+		scope += "/.default"
+	}
```
heaths, line 193: Why not just always add "/.default" when checking for "resource" in condition block on line 177?

heaths (approved):
One nit for slightly better perf.

sdk/keyvault/internal/challenge_policy_test.go
```diff
@@ -136,15 +144,62 @@ func TestFindScopeAndTenant(t *testing.T) {
...
+		{authResource, "https://vault.azure.net", false},
+		{authScope, "https://vault.azure.net/.default", false},
+
+		// error cases: resource/scope doesn't match the requested vault's host (vault.azure.net)
+		{authResource, "https://vault.azure.cn", true},
+		{authResource, "https://myvault.azure.net", true},
+		{authScope, "https://vault.azure.cn/.default", true},
+		{authScope, "https://myvault.azure.net/.default", true},
+	} {
+		t.Run("", func(t *testing.T) {
```
richardpark-msft, line 170: t.Run(test.resource, func(t *testing.T) {} will probably render nicer when executed.

sdk/keyvault/internal/challenge_policy_test.go
```diff
@@ -136,15 +144,62 @@ func TestFindScopeAndTenant(t *testing.T) {
...
+				runtime.PipelineOptions{PerCall: []policy.Policy{p}},
+				&policy.ClientOptions{Transport: srv},
+			)
+			req, err := runtime.NewRequest(context.Background(), "GET", "https://42.vault.azure.net")
+			require.NoError(t, err)
+			_, err = pl.Do(req)
+			if test.err {
+				expected := fmt.Sprintf(`the challenge resource "%s" doesn't match the requested domain`, test.resource)
+				require.EqualError(t, err, expected)
+				if _, ok := err.(*challengePolicyError); !ok {
```
richardpark-msft, line 193: There's a require for this too if you want to use it:

`require.ErrorAs()`

richardpark-msft (approved):
Minor stuff.

sdk/keyvault/internal/challenge_policy_test.go
chlowell: I used the type assert because I want this test to fail if the returned error only wraps `challengePolicyError`.

sdk/keyvault/internal/challenge_policy.go
chlowell: I did this to simplify reporting the resource value in the error message. If the value were `https://foo`, I don't want the error to include `https://foo/.default`.

PR #19083, opened 2022-09-09
[keyvault clients] Verify challenge resource matches request domain

chlowell:
Depends on #19082


sdk/keyvault/azcertificates/utils_test.go
```diff
@@ -70,6 +70,10 @@ func TestMain(m *testing.M) {
 		if err != nil {
 			panic(err)
 		}
+		err = recording.AddHeaderRegexSanitizer("WWW-Authenticate", "https://local", `resource="(.*)"`, &recording.RecordingOptions{GroupForReplace: "1"})
```
heaths, line 73: Nit: the parens shouldn't be necessary since the asterisk only modifies the preceding character i.e., ".".

sdk/keyvault/azkeys/custom_client.go
```diff
@@ -20,7 +20,8 @@ type ClientOptions struct {
 	azcore.ClientOptions
 }
 
-// NewClient creates a client that accesses a Key Vault's keys.
+// NewClient creates a client that accesses a Key Vault's keys. You should validate that vaultURL
+// references a valid Key Vault or Managed HSM. See https://aka.ms/azsdk/blog/vault-uri for details.
```
heaths, line 24: Also do this for your cryptoclient. IIRC, you have one, but remote-only, right?

heaths (approved):

sdk/keyvault/azcertificates/utils_test.go
chlowell: I'm using the parens for grouping so I can replace only the resource URL.

sdk/keyvault/azkeys/custom_client.go
chlowell: No, we have only service clients (`azkeys.Client` has crypto methods).

richardpark-msft:
Updating this PR to take the newly released `internal` package.

richardpark-msft:
Closing this, moving over to #19092 so I can update it.

PR #19086, opened 2022-09-09
[azeventhubs] Fixing some issues with docs and samples

richardpark-msft:
The pkg.go.dev for this package looks pretty bad because of some simple conventions I accidentally violated:

1. Any comment that's directly above the `package` declaration will be included as the package comment, which resulted in a bunch of "Copyright Microsoft"'s being inserted into the Overview. Fixed that by just adding in a blank line.
2. Whole-file examples will not display the full file unless there's at least one other `package level` declaration in the file. I've introduced some that I think aren't too artificial.

I also found an inconsistency in the batch naming - it really should be NumEvents(), rather than NumMessages(), since we're going with the Events terminology rather than Messages (EventData, ReceivedEventData, SendEventDataBatch).

### DIFF
--- a/sdk/messaging/azeventhubs/event_data_batch.go
+++ b/sdk/messaging/azeventhubs/event_data_batch.go
@@ -59,8 +59,8 @@ func (mb *EventDataBatch) NumBytes() uint64 {
 	return mb.currentSize
 }
 
-// NumMessages returns the # of messages in the batch.
-func (mb *EventDataBatch) NumMessages() int32 {
+// NumEvents returns the # of events in the batch.
+func (mb *EventDataBatch) NumEvents() int32 {
 	mb.mu.RLock()
 	defer mb.mu.RUnlock()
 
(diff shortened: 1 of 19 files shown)


richardpark-msft:
@gracewilcox, making you the primary reviewer on this one.

richardpark-msft:
> @gracewilcox, making you the primary reviewer on this one.

Well, I rescind this now :)

sdk/messaging/azeventhubs/event_data_batch.go
```diff
@@ -59,8 +59,8 @@ func (mb *EventDataBatch) NumBytes() uint64 {
 	return mb.currentSize
 }
 
-// NumMessages returns the # of messages in the batch.
-func (mb *EventDataBatch) NumMessages() int32 {
+// NumEvents returns the # of events in the batch.
+func (mb *EventDataBatch) NumEvents() int32 {
```
jhendrixMSFT, line 63: Breaking change entry in the changelog?

richardpark-msft: Fixed.

PR #19108, opened 2022-09-13
Cosmos: Add Location Cache

### LINKED ISSUE #18982: Cosmos DB: Add LocationCache
LocationCache mantains a cache (https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/async_cache.go) of the Locations available to an account.

We need to:

* Implement ReadAccount as an internal operation
* Create a LocationCache that uses the async cache (reference https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs) to store and update the Read/Write regions of an account.

imanvt:

Adds the LocationCache, an internally used feature to manage location availability.

closes #18982

### DIFF
--- a/sdk/data/azcosmos/cosmos_location_cache.go
+++ b/sdk/data/azcosmos/cosmos_location_cache.go
@@ -185,6 +185,37 @@ func (lc *LocationCache) IsEndptUnavailable(endpoint url.URL, ops opType) bool {
 	}
 }
 
+func (lc *LocationCache) getPrefAvailableEndpts(endptsByLoc map[string]url.URL, locs []string, availOps opType, fallbackEndpt url.URL) []url.URL {
+	endpts := make([]url.URL, 0)
+	if lc.enableEndptDiscovery {
+		if lc.CanUseMultipleWriteLocs() || availOps&read != 0 {
+			unavailEndpts := make([]url.URL, 0)
+			unavailEndpts = append(unavailEndpts, fallbackEndpt)
+			for _, loc := range lc.locationInfo.prefLocations {
+				if endpt, ok := endptsByLoc[loc]; ok && endpt != fallbackEndpt {
+					if lc.IsEndptUnavailable(endpt, availOps) {
+						unavailEndpts = append(unavailEndpts, endpt)
+					} else {
+						endpts = append(endpts, endpt)
+					}
+				}
+
+			}
+			endpts = append(endpts, unavailEndpts...)
+		} else {
+			for _, loc := range locs {
+				if endpt, ok := endptsByLoc[loc]; ok && loc != "" {
+					endpts = append(endpts, endpt)
+				}
+			}
+		}
+	}
+	if len(endpts) == 0 {
+		endpts = append(endpts, fallbackEndpt)
+	}
+	return endpts
+}
+
 func NewdbdbAcctLocationsInfo(prefLocations []string, defaultEndpt url.URL) *dbAcctLocationsInfo {
 	availWriteLocs := make([]string, 0)
 	availReadLocs := make([]string, 0)
--- a/sdk/data/azcosmos/cosmos_location_cache.go
+++ b/sdk/data/azcosmos/cosmos_location_cache.go
@@ -64,7 +64,7 @@ type LocationCache struct {
 func NewLocationCache(prefLocations []string, defaultEndpt url.URL) *LocationCache {
 	return &LocationCache{
 		defaultEndpt:                      defaultEndpt,
-		locationInfo:                      *NewdbAcctLocationsInfo(prefLocations, defaultEndpt),
+		locationInfo:                      *NewDbAcctLocationsInfo(prefLocations, defaultEndpt),
 		locationUnavailabilityInfoMap:     make(map[url.URL]locationUnavailabilityInfo),
 		unavailableLocationExpirationTime: DefaultExpirationTime,
 	}
@@ -217,7 +217,7 @@ func (lc *LocationCache) IsEndptUnavailable(endpoint url.URL, ops opType) bool {
 	}
 	lc.rwMutex.RLock()
 	defer lc.rwMutex.RUnlock()
-	return time.Since(info.lastCheckTime) <= lc.unavailableLocationExpirationTime
+	return time.Since(info.lastCheckTime) < lc.unavailableLocationExpirationTime
 }
 
 func (lc *LocationCache) GetPrefAvailableEndpts(endptsByLoc map[string]url.URL, locs []string, availOps opType, fallbackEndpt url.URL) []url.URL {
@@ -270,7 +270,7 @@ func GetEndptsByLocation(locs []AcctRegion) (map[string]url.URL, []string, error
 	return endptsByLoc, parsedLocs, nil
 }
 
-func NewdbAcctLocationsInfo(prefLocations []string, defaultEndpt url.URL) *dbAcctLocationsInfo {
+func NewDbAcctLocationsInfo(prefLocations []string, defaultEndpt url.URL) *dbAcctLocationsInfo {
 	availWriteLocs := make([]string, 0)
 	availReadLocs := make([]string, 0)
 	availWriteEndptsByLocation := make(map[string]url.URL)
--- a/sdk/data/azcosmos/cosmos_location_cache.go
+++ b/sdk/data/azcosmos/cosmos_location_cache.go
@@ -15,13 +15,14 @@ const (
 	none opType = iota
 	read
 	write
+	all
 )
 
 type opType int
 
 type locationUnavailabilityInfo struct {
-	lastCheckTime time.Time
-	unavailableOp opType
+	lastCheckTime  time.Time
+	unavailableOps opType
 }
 
 type dbAcctLocationsInfo struct {
@@ -122,12 +123,12 @@ func (lc *LocationCache) MarkEndptUnavailable(endpoint url.URL, op opType) {
 	lc.mu.Lock()
 	if info, ok := lc.locationUnavailabilityInfoMap[endpoint]; ok {
 		info.lastCheckTime = currTime
-		info.unavailableOp |= op
+		info.unavailableOps |= op
 		lc.locationUnavailabilityInfoMap[endpoint] = info
 	} else {
 		info = locationUnavailabilityInfo{
-			lastCheckTime: currTime,
-			unavailableOp: op,
+			lastCheckTime:  currTime,
+			unavailableOps: op,
 		}
 		lc.locationUnavailabilityInfoMap[endpoint] = info
 	}
@@ -145,5 +146,19 @@ func (lc *LocationCache) RefreshStaleEndpts() {
 		if time.Since(info.lastCheckTime) > lc.unavailableLocationExpirationTime {
 			delete(lc.locationUnavailabilityInfoMap, endpoint)
 		}
+	}
+	lc.mu.Unlock()
+}
+
+func (lc *LocationCache) IsEndptUnavailable(endpoint url.URL, ops opType) bool {
+	lc.mu.Lock()
+	info, ok := lc.locationUnavailabilityInfoMap[endpoint]
+	if ops == none || !ok || ops&info.unavailableOps == 0 {
+		return false
+	} else {
+		if time.Since(info.lastCheckTime) > lc.unavailableLocationExpirationTime {
+			return false
+		}
+		return true
 	}
 }
--- a/sdk/data/azcosmos/cosmos_location_cache.go
+++ b/sdk/data/azcosmos/cosmos_location_cache.go
@@ -63,7 +63,7 @@ type LocationCache struct {
 func NewLocationCache(prefLocations []string, defaultEndpt url.URL) *LocationCache {
 	return &LocationCache{
 		defaultEndpt:                  defaultEndpt,
-		locationInfo:                  *NewdbdbAcctLocationsInfo(prefLocations, defaultEndpt),
+		locationInfo:                  *NewdbAcctLocationsInfo(prefLocations, defaultEndpt),
 		locationUnavailabilityInfoMap: make(map[url.URL]locationUnavailabilityInfo),
 	}
 }
@@ -232,7 +232,7 @@ func (lc *LocationCache) GetEndptsByLocation(locs []AcctRegion) (map[string]url.
 	return endptsByLoc, parsedLocs
 }
 
-func NewdbdbAcctLocationsInfo(prefLocations []string, defaultEndpt url.URL) *dbAcctLocationsInfo {
+func NewdbAcctLocationsInfo(prefLocations []string, defaultEndpt url.URL) *dbAcctLocationsInfo {
 	availWriteLocs := make([]string, 0)
 	availReadLocs := make([]string, 0)
 	availWriteEndptsByLocation := make(map[string]url.URL)
--- a/sdk/data/azcosmos/cosmos_location_cache.go
+++ b/sdk/data/azcosmos/cosmos_location_cache.go
@@ -9,7 +9,7 @@ import (
 	"time"
 )
 
-type opType int
+const DefaultExpirationTime time.Duration = time.Minute * 5
 
 const (
 	none opType = iota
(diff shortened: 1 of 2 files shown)


sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	availWriteLocations        []string
+	availReadLocations         []string
+	availWriteEndptsByLocation map[string]url.URL
+	availReadEndptsByLocation  map[string]url.URL
+	writeEndpts                []url.URL
+	readEndpts                 []url.URL
+}
+
+// AcctRegion represents a Azure Cosmos DB Database Account in a specific region.
+type AcctRegion struct {
```
ealsur, line 39:
```suggestion
type accountRegion struct {
```

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	name string
+	// The URL of the database account location in the Azure Cosmos DB service.
+	endpoint string
+	// props    map[string]interface{}
+}
+
+// AcctProperties represents a container for Azure Cosmos DB databases.
+type AcctProperties struct {
+	// The list of readable regions of this database account from the Azure Cosmos DB service.
+	readRegions []AcctRegion
```
ealsur, line 50: Are these meant to be deserialized using JSON? Then are we missing the json decorators?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	// Indicates whether multiple write locations are enabled for this database account.
+	enableMultipleWriteLocations bool
+}
+
+type locationCache struct {
+	locationInfo              dbAcctLocationsInfo
+	defaultEndpt              url.URL
+	enableEndptDiscovery      bool
+	useMultipleWriteLocations bool
+	// connLimit                         int
```
ealsur, line 62: Remove all comments, if this is not currently used, then we can add it later

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	lc.lastUpdateTime = time.Now()
+	lc.rwMutex.RUnlock()
+	lc.rwMutex.Lock()
+	lc.locationInfo = nextLoc
+	lc.rwMutex.Unlock()
+	return nil
+}
+
+func (lc *locationCache) readEndpts() ([]url.URL, error) {
+	lc.mapMutex.RLock()
```
ealsur, line 117: Why do we need the lock on the read operation? https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L92

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	nextLoc.readEndpts = lc.getPrefAvailableEndpts(nextLoc.availReadEndptsByLocation, nextLoc.availReadLocations, read, nextLoc.writeEndpts[0])
+	lc.lastUpdateTime = time.Now()
+	lc.rwMutex.RUnlock()
+	lc.rwMutex.Lock()
+	lc.locationInfo = nextLoc
+	lc.rwMutex.Unlock()
+	return nil
+}
+
+func (lc *locationCache) readEndpts() ([]url.URL, error) {
```
ealsur, line 116: Please don't abbreviate names

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	lc.mapMutex.RLock()
+	defer lc.mapMutex.RUnlock()
+	if time.Since(lc.lastUpdateTime) > lc.unavailableLocationExpirationTime && len(lc.locationUnavailabilityInfoMap) > 0 {
+		err := lc.update(nil, nil, nil, lc.enableMultipleWriteLocations)
```
ealsur, line 120: .NET alternative seems to not pass the last parameter https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L115

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: For the locks, these are the decisions I made:

- Utilize RWMutexes (Read-Write), which allow multiple read locks to be acquired or ONE Write lock concurrently.
- One RWmutex for the locationUnavailabilityInfoMap
- One RWmutex for all other values of the location cache. 
If we are not concerned with any race conditions for the things like the lastUpdateTime, we can remove some uses of the read locks.

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return lc.locationInfo.readEndpts, nil
+}
+
+func (lc *locationCache) writeEndpts() ([]url.URL, error) {
+	lc.mapMutex.RLock()
```
ealsur, line 129: Similar question as with the read endpoints, why do you need a lock?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return lc.locationInfo.writeEndpts, nil
+}
+
+func (lc *locationCache) getLocation(endpoint url.URL) string {
+	firstLoc := ""
+	lc.rwMutex.RLock()
```
ealsur, line 142: Why do we need the lock here?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	defer lc.rwMutex.RUnlock()
+	// TODO: Find workaround for firstLoc, maps are unordered
```
ealsur: Would be ideal to see if there is a shorter/cleaner way to do this, the code looks like it does the same as 
https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L131-L143 but it took me a bit to digest

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: I would try to align with the current .NET implementation, if there are no locks there, I would assume none are needed?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+		locationInfo:                      *newDbAcctLocationsInfo(prefLocations, defaultEndpt),
+		locationUnavailabilityInfoMap:     make(map[url.URL]locationUnavailabilityInfo),
+		unavailableLocationExpirationTime: defaultExpirationTime,
+	}
+}
+
+func (lc *locationCache) update(writeLocations []AcctRegion, readLocations []AcctRegion, prefList []string, enableMultipleWriteLocations bool) error {
+	lc.rwMutex.RLock()
+	nextLoc := copyDbAcctLocationsInfo(lc.locationInfo)
+	if prefList != nil {
```
ealsur, line 81: `nil` comparison will work if the parameter is a pointer, like `*[]string`

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+		nextLoc.prefLocations = prefList
+	}
+	lc.enableMultipleWriteLocations = enableMultipleWriteLocations
```
ealsur, line 84: Should be conditional? You can have optional parameters like `*bool`

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+	lc.refreshStaleEndpts()
+	if readLocations != nil {
```
ealsur, line 86: Requires input parameter to be a pointer to work correctly

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+	for endpoint, info := range lc.locationUnavailabilityInfoMap {
+		t := time.Since(info.lastCheckTime)
+		if t > lc.unavailableLocationExpirationTime {
+			delete(lc.locationUnavailabilityInfoMap, endpoint)
+		}
+	}
+	lc.mapMutex.Unlock()
+}
+
+func (lc *locationCache) isEndptUnavailable(endpoint url.URL, ops opType) bool {
```
ealsur, line 215: We already have OperationType enum available, add any you need there https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/operation_type.go

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+	lc.mapMutex.RLock()
+	info, ok := lc.locationUnavailabilityInfoMap[endpoint]
+	lc.mapMutex.RUnlock()
+	if ops == none || !ok || ops&info.unavailableOps != ops {
+		return false
+	}
+	lc.rwMutex.RLock()
```
ealsur, line 222: Why do we need the lock here?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+	enableMultipleWriteLocations bool
+}
+
+type locationCache struct {
+	locationInfo              dbAcctLocationsInfo
+	defaultEndpt              url.URL
+	enableEndptDiscovery      bool
+	useMultipleWriteLocations bool
+	rwMutex                           sync.RWMutex
+	locationUnavailabilityInfoMap     map[url.URL]locationUnavailabilityInfo
```
ealsur, line 62: Should this be a `sync.Map` ? https://pkg.go.dev/sync#Map

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,303 @@
...
+	parsedLocs := make([]string, 0)
+	for _, loc := range locs {
+		endpt, err := url.Parse(loc.endpoint)
+		if err != nil {
+			return nil, nil, err
+		}
+		if loc.name != "" {
+			endptsByLoc[loc.name] = *endpt
+			parsedLocs = append(parsedLocs, loc.name)
+		}
```
ealsur, line 270: Probably needs an else clause with a comment like a TODO: Log ? I know that we cannot do the logging now but at least leave a TODO in all the places where we might want to

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: If the type is not public and not exported, do we need to still provide the json tags? I am getting a compiler warning about this.

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: Just to clarify, `.Rlock()` is a read lock, which is non-blocking assuming there are no current write locks acquired. I added them to prevent any race conditions. For a `RWMutex`, you can acquire the write lock with just `.Lock()`

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: As far as I understand, the zero/default value for slice is nil, so it works fine as is in that regard. Passing parameters as pointers are typically only needed if they are going to be modified, correct? The only benefit I can think of is that, for large slices, it may be more efficient to pass the pointer vs. pass-by-value (i.e. copy the slice).

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: I used the mutex + normal map combination to deal with some things that a sync.Map doesn't support, such as getting the length/checking if non-empty. I can restructure this to use a sync.Map and figure out other workarounds, but the underlying functionality is the same.

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: I understand, let me know if you have any direct suggestions, but this was the way to work around the fact that there's no direct way to enumerate a map without a for-range that I know of, which is needed to imitate the FirstOrDefault in C#.

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: I think the type that I defined is still needed, perhaps a better name for it would be something along the lines of `requestedOperations`? Since these are treated like flags for the location cache to indicate one or more operations to check availability for.

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: The type needs to be internal, the property public: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/cosmos_offers.go#L18-L20

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: My question is more around, why don't we need this on .NET?

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: No suggestions, my lack of Go experience shows :)

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: If nil works (and we can verify with tests) sure. I was under the impression (wrongly) that `nil` only worked with pointers and optional parameters should use pointers. The pattern used for Go SDKs' public API is for all optional parameters to be pointers.

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: I just thought that using a language provided construct versus fabricating one ourselves would be better. Besides obtaining the length (which I think is not exactly that, you need to just know if there is at least 1 item, right?) which are the other gaps?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -0,0 +1,298 @@
...
+	"time"
+)
+
+const defaultExpirationTime time.Duration = time.Minute * 5
+
+const (
+	none opType = iota
+	read
+	write
+	all
```
ealsur, line 18: Is `all` used?

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: According to https://azure.github.io/azure-sdk/golang_introduction.html#optional-parameters, it seems that is best to create a struct type representing a set of optional parameters to provide. I believe this is only for the public-facing API, so in this case, I think we can leave as is. The tests already cover the functionality, including passing nil directly.

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: Yes. Currently, I used bitwise operations to treat this enum/iota as flags. So all = read&write

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: The other major reason that I preferred to utilize this approach is to have more control over it. I believe for sync.map, you cannot limit the key/value types, i.e. you can put anything into it. Do you think we should still change it?

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: At the moment, there is only one map, for which I simulated the concurrent dictionary using the variable `mapMutex`. So that's why it's needed here, in another comment thread, I explained why I took this approach rather than use the sync.Map

sdk/data/azcosmos/cosmos_location_cache.go
imanvt: See the remaining open comment thread, for the current implementation, this is required. Based on the decision in the other thread, I will either revise or leave as is.

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: This method is not using the map, so why do we need the lock on it?

ealsur: I understand methods working with `unavailableLocationExpirationTime` but what about those that don't?

ealsur: My concern is that this is the read path, yes, sure it can be concurrent, but so it is on other languages (C#, Java) and there are no locks there. If concurrency for these other parts (not related to the map) was a concern, then all the languages should have the locks.

imanvt: removed it

ealsur (approved):

PR #19113, opened 2022-09-14
[azservicebus] Create a programmatically useful code for when a session is not available

### LINKED ISSUE #19039: [azservicebus] timeout error when accepting next session is not exported type
pkg: github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus
version: v1.1.0
go version: go 1.18.5

In testing, if I call the `azservicebus` client's `AcceptNextSessionForSubscription()` method and there is no session to accept, I get the following timeout error:

```
*Error{Condition: com.microsoft:timeout, Description: The operation did not complete within the allotted timeout of 00:01:04.9800000. The time allotted to this operation may have been a portion of a longer timeout. For more information on exception types and proper exception handling, please refer to https://aka.ms/ServiceBusExceptions . TrackingId:0d5d0118-2aaa-4b60-92af-7a3882f93837_B0, SystemTracker:nalb-int-tests:Topic:14prizb4|0tzhfbtr, Timestamp:2022-09-06T19:25:22 TrackingId:03ff398b38f8438fba50b343d7b09534_G18, SystemTracker:gateway7, Timestamp:2022-09-06T19:25:22, Info: map[]}
```

This timeout error formerly contained a `*amqp.Error` which allowed for type checking. The timeout error is no longer exported, so I cannot check the type using `errors.Is()` or `errors.As()`, and specifically handle the case of a foreseeable timeout. The [documentation](https://github.com/Azure/azure-sdk-for-go/blob/sdk/messaging/azservicebus/v1.1.0/sdk/messaging/azservicebus/client.go#L306) for the `AcceptNextSessionForSubscription()` method claims that "if the operation fails it can return an *azservicebus.Error type if the failure is actionable." Would you please fix the current implementation so that the timeout error is exported?

richardpark-msft:
Adding a new error Code (`CodeTimeout`) for when we get back a `com.microsoft:timeout` error from Service Bus. The primary use is if we are doing an `AcceptNextSessionFor<Entity>()` and there aren't any available sessions.
    
This should allow customers to figure out if there's a genuine failure or if their session-enabled queue/subscription is simply "empty".

This also improves the tests and adds in some more customer exposed examples for reading through a a queue of sessions.
    
Fixes #19039

### DIFF
--- a/sdk/messaging/azservicebus/client.go
+++ b/sdk/messaging/azservicebus/client.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
@@ -37,6 +38,10 @@ type Client struct {
 		internal.NamespaceForAMQPLinks
 	}
 	retryOptions RetryOptions
+
+	// acceptNextTimeout controls how long the session accept can take before
+	// the server stops waiting.
+	acceptNextTimeout time.Duration
 }
 
 // ClientOptions contains options for the `NewClient` and `NewClientFromConnectionString`
@@ -262,11 +267,11 @@ func (client *Client) AcceptSessionForSubscription(ctx context.Context, topicNam
 		toReceiverOptions(options))
 
 	if err != nil {
-		return nil, err
+		return nil, internal.TransformError(err)
 	}
 
 	if err := sessionReceiver.init(ctx); err != nil {
-		return nil, err
+		return nil, internal.TransformError(err)
 	}
 
 	client.addCloseable(id, sessionReceiver)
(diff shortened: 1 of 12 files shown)


sdk/messaging/azservicebus/internal/errors.go
```diff
@@ -63,6 +63,13 @@ func TransformError(err error) error {
 		return exported.NewError(exported.CodeLockLost, err)
 	}
 
+	if isMicrosoftTimeoutError(err) {
+		// one scenario where this error pops up is if you're waiting for an available
+		// session and there are non available. It waits, up to one minute, and then
```
ryepup, line 68: nit: typo

```suggestion
		// session and there are none available. It waits, up to one minute, and then
```

sdk/messaging/azservicebus/client.go
```diff
@@ -37,6 +38,10 @@ type Client struct {
 		internal.NamespaceForAMQPLinks
 	}
 	retryOptions RetryOptions
+
+	// acceptNextTimeout controls how long the session accept can take before
+	// the server stops waiting.
+	acceptNextTimeout time.Duration
```
jhendrixMSFT, line 44: Is this meant to be publicly configurable?  I didn't see where this is exposed.

richardpark-msft: It's not publicly configurable for now. We have an open issue for it, but the functionality was useful for making some of these live tests practical (otherwise we have to wait for 1:05 for the timeout).

jhendrixMSFT (approved):

PR #19123, opened 2022-09-14
[keyvault/internal] Add KeyVaultChallengePolicyOptions

chlowell:


jhendrixMSFT (approved):
I can't comment on the correctness of the algorithm, but the change itself looks good.  Just need to be careful on release timing as this is a breaking change and would break older KV clients.

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -27,18 +27,29 @@ import (
 const headerAuthorization = "Authorization"
 const bearerHeader = "Bearer "
 
+type KeyVaultChallengePolicyOptions struct {
+	// DisableChallengeResourceVerification controls whether the policy requires the
+	// authentication challenge resource to match the Key Vault or Managed HSM domain
+	DisableChallengeResourceVerification bool
```
RickWinter, line 33: Is every language using this name?

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -27,18 +27,29 @@ import (
...
 	// mainResource is the resource to be retrieved using the tenant specified in the credential
-	mainResource *temporal.Resource[azcore.AccessToken, acquiringResourceState]
-	cred         azcore.TokenCredential
-	scope        *string
-	tenantID     *string
+	mainResource            *temporal.Resource[azcore.AccessToken, acquiringResourceState]
+	cred                    azcore.TokenCredential
+	scope                   *string
+	tenantID                *string
+	verifyChallengeResource bool
```
RickWinter, line 42: I suggest dropping the *Resource* postfix.  We have that on mainResource which is a Resource.  In this case this is a switch and not a resource.

```suggestion
	verifyChallenge bool
```

sdk/keyvault/internal/challenge_policy.go
heaths: Except for Python, yes, we plan to.

sdk/keyvault/internal/challenge_policy.go
heaths: We are specifically verifying the `resource` in the challenge, not the entire challenge itself. Dropping "resource" is misleading.

heaths (approved):

heaths (approved):

heaths (approved):

PR #19127, opened 2022-09-15
Fix sub-test case not getting recording variables

Alancere:
example：
  test name： ScheduledqueryrulesTestSuite
  sub-test name：ScheduledqueryrulesTestSuite/TestScheduledqueryrule

monitor live test: https://github.com/Azure/azure-sdk-for-go/pull/19117

### DIFF
--- a/sdk/internal/recording/recording.go
+++ b/sdk/internal/recording/recording.go
@@ -807,7 +807,8 @@ func IsLiveOnly(t *testing.T) bool {
 
 // GetVariables returns access to the variables stored by the test proxy for a specific test
 func GetVariables(t *testing.T) map[string]interface{} {
-	if s, ok := testSuite[t.Name()]; ok {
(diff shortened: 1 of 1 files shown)


sdk/internal/recording/recording.go
```diff
@@ -807,7 +807,8 @@ func IsLiveOnly(t *testing.T) bool {
 
 // GetVariables returns access to the variables stored by the test proxy for a specific test
```
RickWinter, line 808: Why do have the "/<extra stuff>" as the key name for the test if we are going to trim it ?

jhendrixMSFT:
Is this not a problem elsewhere (e.g. `azblob` tests) because those tests don't rely on recorded variables?

tadelesh:
@RickWinter @jhendrixMSFT The problem is that we start record in the main test and get variable in the sub test. I think it should be a wrong usage problem. Close the PR.

PR #19136, opened 2022-09-16
Container soft delete

souravgupta-msft:
Added code for restoring soft-deleted container

### DIFF
--- a/sdk/storage/azblob/container/client.go
+++ b/sdk/storage/azblob/container/client.go
@@ -160,6 +160,18 @@ func (c *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 	return resp, err
 }
 
+// Restore operation restore the contents and properties of a soft deleted container to a specified container.
+// For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/restore-container.
+func (c *Client) Restore(ctx context.Context, deletedContainerName string, deletedContainerVersion string) (RestoreResponse, error) {
+	opts := &generated.ContainerClientRestoreOptions{
+		DeletedContainerName:    &deletedContainerName,
+		DeletedContainerVersion: &deletedContainerVersion,
+	}
+	resp, err := c.generated().Restore(ctx, opts)
+
+	return resp, err
+}
+
 // GetProperties returns the container's properties.
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/get-container-metadata.
 func (c *Client) GetProperties(ctx context.Context, o *GetPropertiesOptions) (GetPropertiesResponse, error) {
--- a/sdk/storage/azblob/container/client.go
+++ b/sdk/storage/azblob/container/client.go
@@ -163,12 +163,14 @@ func (c *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 // Restore operation restore the contents and properties of a soft deleted container to a specified container.
 // For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/restore-container.
 func (c *Client) Restore(ctx context.Context, options *RestoreOptions) (RestoreResponse, error) {
-	var opts *generated.ContainerClientRestoreOptions
+	urlParts, err := blob.ParseURL(c.URL())
+	if err != nil {
+		return RestoreResponse{}, err
+	}
+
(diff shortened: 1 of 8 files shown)


sdk/storage/azblob/container/client.go
```diff
@@ -160,6 +160,18 @@ func (c *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 	return resp, err
 }
 
+// Restore operation restore the contents and properties of a soft deleted container to a specified container.
+// For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/restore-container.
+func (c *Client) Restore(ctx context.Context, deletedContainerName string, deletedContainerVersion string) (RestoreResponse, error) {
```
jhendrixMSFT, line 165: Our guidelines state that APIs should have an options type as the last argument.  This is to provide extensibility without a breaking change.

Also, are both of these args mandatory?  I see they're not in the generated code.  I know that sometimes the generated code is incorrect due to OpenAPI spec bugs.  Is that the case here?

sdk/storage/azblob/container/client.go
```diff
@@ -160,6 +160,18 @@ func (c *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
...
+	opts := &generated.ContainerClientRestoreOptions{
```
vibhansa-msft, line 166: We shall redefine this in exported code and that can be used as input options to this API.

sdk/storage/azblob/container/client.go
souravgupta-msft: Converted the args to options

sdk/storage/azblob/container/client.go
souravgupta-msft: Done

tasherif-msft (approved):

siminsavani-msft (approved):
LGTM!

sdk/storage/azblob/container/client.go
```diff
@@ -160,6 +160,23 @@ func (c *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 	return resp, err
 }
 
+// Restore operation restore the contents and properties of a soft deleted container to a specified container.
+// For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/restore-container.
+func (c *Client) Restore(ctx context.Context, options *RestoreOptions) (RestoreResponse, error) {
```
nakulkar-msft, line 165: Container Name and version are required parameters to restore a container. Both of these should be in the func signature. Options can be an empty struct for now. Else the using function with default value will always fail. 

Should we check if this API is needed at container level or just at service level?

sdk/storage/azblob/service/client.go
```diff
@@ -123,6 +123,14 @@ func (s *Client) DeleteContainer(ctx context.Context, containerName string, opti
 	return containerDeleteResp, err
 }
 
+// RestoreContainer restores soft-deleted container
+// Operation will only be successful if used within the specified number of days set in the delete retention policy
+func (s *Client) RestoreContainer(ctx context.Context, deletedContainerName string, options *RestoreContainerOptions) (RestoreContainerResponse, error) {
```
nakulkar-msft, line 128: Here as well. Required parameters should be in the signature.

nakulkar-msft (changes requested):

sdk/storage/azblob/container/client.go
jhendrixMSFT: This is the `container` client, we already have the container name (see the following lines where we obtain it).

souravgupta-msft: @jhendrixMSFT, should we add the deleted container version as a parameter of the function (this parameter is mandatory for restoring a container). We will also have the options as a parameter which will be empty for now. What do you think?
@nakulkar-msft, this is what you are referring to, right?

jhendrixMSFT: If version is required for the API to work, then yeah it should be a required parameter and the options type empty.

souravgupta-msft: Added deleted container version as a parameter and also keeping empty options struct

sdk/storage/azblob/service/client.go
souravgupta-msft: Added

vibhansa-msft (approved):

nakulkar-msft (approved):

Issue #18977, opened 2022-08-30
Support GetSASURL for User Delegation SAS within ContainerClient 


bmorck:
### Feature Request

Currently [ContainerClient.GetSASURL](https://github.com/Azure/azure-sdk-for-go/blob/sdk/storage/azblob/v0.4.1/sdk/storage/azblob/zc_container_client.go#L234) only supports Service SAS. This means that `ContainerClient.GetSASURL` will only supply a SAS if the client was authenticated using a Shared Key.

We would like to be able to call `GetSASURL` and receive a User Delegation SAS if the ContainerClient was authenticated using Azure Active Directory. 

Our use case does not use Shared Keys as they are long-lived credentials and provide less fine-grained permissions than Azure Active Directory. We will also be using SAS as it provides short-lived configurable access to resources. 

Additionally, I have posted another issue (https://github.com/Azure/azure-sdk-for-go/issues/18976) to support User Delegation Credentials.

zezha-msft:
Thank you for the feature request. It is being addressed in this PR: https://github.com/Azure/azure-sdk-for-go/pull/19141

jhendrixMSFT:
This has been added in `azblob@v0.5.0`

Issue #18976, opened 2022-08-30
Support `UserDelegationCredential` in BlobSASSignatureValues.NewSASQueryParameters and on ServiceClient


bmorck:
### Feature Request

Within the previous [azblob SDK](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob), there was support for a UserDelegationCredential. This credential could be supplied to [BlobSASSignatureValues.NewSASQueryParameters](https://github.com/Azure/azure-storage-blob-go/blob/v0.15.0/azblob/sas_service.go#L45) and is returned from [ServiceURL.GetUserDelegationCredential](https://github.com/Azure/azure-storage-blob-go/blob/v0.15.0/azblob/url_service.go#L33) or created using [NewUserDelegationCredential](https://github.com/Azure/azure-storage-blob-go/blob/v0.15.0/azblob/user_delegation_credential.go#L10) by supplying a UserDelegationKey. 

Currently, within the new azblob SDK the [UserDelegationKey is supported](https://github.com/Azure/azure-sdk-for-go/blob/sdk/storage/azblob/v0.4.1/sdk/storage/azblob/zz_generated_models.go#L1000) but is not used in any methods. Additionally there is a comment on  BlobSASSignatureValues.NewSASQueryParameters that indicated that `UserDelegationCredential` can be used, [but it only accepts a `SharedKeyCredential`](https://github.com/Azure/azure-sdk-for-go/blob/sdk/storage/azblob/v0.4.1/sdk/storage/azblob/zc_sas_service.go#L51)

> NewSASQueryParameters uses an account's StorageAccountCredential to sign this signature values to produce the proper SAS query parameters. See: StorageAccountCredential. Compatible with both UserDelegationCredential and SharedKeyCredential

Due to this, we would like the `UserDelegationCredential` to be supported on the ServiceClient (as it was supported on the ServiceURL in the previous azblob SDK, and in `BlobSASSignatureValues.NewSASQueryParameters`. 

Currently we are generating User Delegation Key using the [REST API](https://docs.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key) and manually generating the User Delegation SAS using the User Delegation Key returned from the API call. This is a verbose method of constructing the SAS and using the SDK would allow for a simpler and more readable method of working with User Delegation SAS.

zezha-msft:
Thanks for the feature request, it is being actively worked on.

siminsavani-msft:
[...]

Issue #16916, opened 2022-01-25
azblob - implement UserDelegationCredential


alexg-axis:
### Feature Request

This feature request is for the "new" `azblob` package.

#### Background

In order to create a SAS token, one uses the `azblob.BlobSASSignatureValues.NewSASQueryParameters` function:

https://github.com/Azure/azure-sdk-for-go/blob/8c965f7cbb93f9fdf1fec8df5416f193e0bd465e/sdk/storage/azblob/zc_sas_service.go#L33-L36

It specifies that it is compatible with `UserDelegationCredential`. This seems to be left from before the SDK was imported. There seems to be no `UserDelegationCredential` anywhere in this repository, in fact.

#### Request

It's quite the hassle to fetch a user delegation key and then manually create the final SAS token.

I request a feature to easily create a `UserDelegationCredential` for use with `NewSASQueryParameters` so that a delegated SAS token can be created using the SDK.

mohsha-msft:
Hey @alexg-axis,

Thanks for reaching out!
We're planning to add that feature in march release most likely!

alexg-axis:
@mohsha-msft Is it still planned for release in March?

mt35-rs:
I also need this functionality.  We are trying to obey the "principle of least privilege" here, but not being able to (easily) work with service principals makes this really difficult.  Having to use a very privileged access key to create signed URLs seems totally backward.

abdullah-lt:
Hi. Is there any update on this? 
Is there a way where I can generate SAS token without  SharedKeyCredential  as this requires the use of storage account key. But I want to use manage identity with `NewDefaultAzureCredential`

alexg-axis:
@mohsha-msft Hi! Is there any timeline on when this will be fixed?

zezha-msft:
Update: this feature is being actively worked on here -> https://github.com/Azure/azure-sdk-for-go/pull/19141

siminsavani-msft:
Hi @alexg-axis ! We have released the User Delegation feature (https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fstorage%2Fazblob%2Fv0.5.0) today. Please give it a try and let us know if you have any questions!

Examples can be found here:
https://github.com/Azure/azure-sdk-for-go/blob/7a3989f70fa6c2d4376d0c72c01b5afc32b527f7/sdk/storage/azblob/service/examples_test.go#L286

PR #19141, opened 2022-09-16
Adding UDK/UDC to service client

siminsavani-msft:
Adding the option to create a service client with UDK. Tests in progress!

https://github.com/Azure/azure-sdk-for-go/issues/18976, https://github.com/Azure/azure-sdk-for-go/issues/16916, https://github.com/Azure/azure-sdk-for-go/issues/18977


sdk/messaging/azeventhubs/internal/blob/zc_sas_account.go
```diff
@@ -27,7 +27,7 @@ type AccountSASSignatureValues struct {
 	ResourceTypes string      `param:"srt"` // Create by initializing AccountSASResourceTypes and then call String()
 }
 
-// Sign uses an account's shared key credential to sign this signature values to produce
+// Sign uses an account's SharedKeyCredential to sign this signature values to produce
```
richardpark-msft, line 30: This is just a copy of the library and will get auto-updated. Can you remove these changes from this PR?

siminsavani-msft: Thank you for bringing this to my attention! Fixed now :)

sdk/storage/azblob/constants.go
```diff
@@ -35,3 +36,5 @@ const (
 func PossibleDeleteSnapshotsOptionTypeValues() []DeleteSnapshotsOptionType {
 	return generated.PossibleDeleteSnapshotsOptionTypeValues()
 }
+
+const SASTimeFormat = exported.SASTimeFormat // "2017-07-27T00:00:00Z" // ISO 8601
```
jhendrixMSFT, line 40: This is already in the `sas` package.  Does it need to be exported here?

sdk/storage/azblob/container/client.go
```diff
@@ -308,3 +313,42 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, start time.Time
...
+	if c.userDelegationKey() == nil {
+		return "", errors.New("GetUDKSASURL can only be signed with a UserDelegationKey")
+	}
+
+	urlParts, err := blob.ParseURL(c.URL())
+	if err != nil {
+		return "", err
+	}
+
+	udc := azblob.UserDelegationCredential{
```
jhendrixMSFT, line 329: Please use the source from `exported`.  We don't want to add a dependency on the `azblob` package.

sdk/storage/azblob/sas/service.go
```diff
@@ -151,15 +131,119 @@ func (v BlobSignatureValues) Sign(sharedKeyCredential *SharedKeyCredential) (Que
...
-	//	p.signedStart = udk.SignedStart
-	//	p.signedExpiry = udk.SignedExpiry
-	//	p.signedService = udk.SignedService
-	//	p.signedVersion = udk.SignedVersion
-	//}
+	return p, nil
+}
+
+// SignUDK uses an account's UserDelegationCredential to sign this signature values to produce the proper SAS query parameters.
+func (v BlobSignatureValues) SignUDK(userDelegationCredential *UserDelegationCredential) (QueryParameters, error) {
```
jhendrixMSFT, line 138: I wonder if this should be named `SignWithUDK`, and then also rename `Sign` to `SignWithSharedKey`

sdk/storage/azblob/service/client.go
```diff
@@ -81,6 +81,32 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// NewClientWithUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func NewClientWithUserDelegationCredential(serviceURL string, ctx context.Context, info generated.KeyInfo, timeout *int32, requestID *string) (*Client, error) {
```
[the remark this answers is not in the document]
jhendrixMSFT: `NewClientWithUserDelegationCredential(serviceURL, cred *UserDelegationCredential, options *ClientOptions)`

Generally speaking, the `context.Context` is always the first parameter by convention.

sdk/storage/azblob/service/examples_test.go
```diff
@@ -246,3 +248,49 @@ func Example_service_SASSignatureValues_Sign() {
...
+
+	// Set current and past time and create key
+	currentTime := time.Now().UTC().Add(-10 * time.Second)
+	pastTime := currentTime.Add(48 * time.Hour)
+	info := generated.KeyInfo{
+		Start:  to.Ptr(currentTime.UTC().Format(sas.TimeFormat)),
+		Expiry: to.Ptr(pastTime.UTC().Format(sas.TimeFormat)),
+	}
+
+	_, err = service.GetUserDelegationCredential(svcClient.URL(), context.Background(), info, nil, nil)
```
jhendrixMSFT, line 276: Shouldn't the returned `UserDelegationCredential` be used?

sdk/storage/azblob/sas/service.go
```diff
@@ -151,15 +131,119 @@ func (v BlobSignatureValues) Sign(sharedKeyCredential *SharedKeyCredential) (Que
...
+		signedDirectoryDepth:       getDirectoryDepth(v.Directory),
+		preauthorizedAgentObjectID: v.PreauthorizedAgentObjectId,
+		agentObjectID:              v.AgentObjectId,
+		correlationID:              v.CorrelationId,
+		// Calculated SAS signature
+		signature: signature,
+	}
+
+	//User delegation SAS specific parameters
+	if udk != nil {
```
jhendrixMSFT, line 239: This should never be `nil` right?  It's being dereferenced earlier.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,49 @@
...
+	Key  UserDelegationKey
+}
+
+// AccountName returns the Storage account's Name
+func (f *UserDelegationCredential) AccountName() string {
+	return f.Name
+}
+
+// ComputeHMAC
+func (f *UserDelegationCredential) ComputeHMACSHA256(message string) (string, error) {
```
jhendrixMSFT, line 39: This method, and `GetUDKParams` below shouldn't be on the `UserDelegationCredential` as they become part of the public surface area.  Can you make them funcs instead?  This way, only the SDK can call them.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
+)
+
+// NewUserDelegationCredential creates a new UserDelegationCredential using a Storage account's Name and a user delegation Key from it
+func NewUserDelegationCredential(accountName string, udk generated.UserDelegationKey) *UserDelegationCredential {
```
jhendrixMSFT, line 17: The `udk` param should be the type-alias defined on line 25.

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,32 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
...
+	}
+	udk, err := sc.GetUserDelegationKey(ctx, info, &opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewUserDelegationCredential(url.Host, udk.UserDelegationKey), nil
+}
+
+func (s Client) GetUserDelegationKey(ctx context.Context, keyInfo generated.KeyInfo, options *GetUserDelegationKeyOptions) GetUserDelegationKeyResponse {
```
jhendrixMSFT, line 107: The `keyInfo` param must be a publicly accessible type.  You'll need to define a type-alias for it.

sdk/storage/azblob/blob/models.go
```diff
@@ -23,6 +23,9 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 	return exported.NewSharedKeyCredential(accountName, accountKey)
 }
 
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = generated.UserDelegationKey
```
jhendrixMSFT, line 27: I don't see this type being used in the `blob` package, can it be removed?

sdk/storage/azblob/container/models.go
```diff
@@ -22,6 +22,9 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 	return exported.NewSharedKeyCredential(accountName, accountKey)
 }
 
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = exported.UserDelegationKey
```
jhendrixMSFT, line 26: Unused type.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,59 @@
...
+		Key:  udk,
+	}
+}
+
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = generated.UserDelegationKey
+
+// UserDelegationCredential contains an account's name and its user delegation key.
+type UserDelegationCredential struct {
+	Name string
```
jhendrixMSFT, line 29: We typically don't export the fields in the credential (or clients) as we want them to be immutable once created.  Once a `UserDelegationCredential` has been created, would a customer ever need to update these fields?

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,59 @@
...
+type UserDelegationKey = generated.UserDelegationKey
+
+// UserDelegationCredential contains an account's name and its user delegation key.
+type UserDelegationCredential struct {
+	accountName       string
+	userDelegationKey UserDelegationKey
+}
+
+// AccountName returns the Storage account's Name
+func (f *UserDelegationCredential) AccountName() string {
```
jhendrixMSFT, line 34: Will customers need to call this, or can it be removed?  It can always be added later if needed.

sdk/storage/azblob/service/models.go
```diff
@@ -21,6 +21,22 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
...
+
+type KeyInfo = generated.KeyInfo
+
+// NewUserDelegationCredential creates a new UserDelegationCredential using a Storage account's name and a user delegation key from it
+func NewUserDelegationCredential(accountName string, udk UserDelegationKey) *UserDelegationCredential {
+	return exported.NewUserDelegationCredential(accountName, udk)
+}
+
+// GetUserDelegationKeyOptions contains request id and timeout
+type GetUserDelegationKeyOptions = generated.ServiceClientGetUserDelegationKeyOptions
```
jhendrixMSFT, line 38: I don't think we want to expose this underlying options type as all it contains is `RequestID` and `Timeout` fields which we hide from customers.  Probably best to define an empty options type for now (see `GetPropertiesOptions` in this file for the pattern).

sdk/storage/azblob/internal/exported/user_delegation_credential.go
siminsavani-msft: I do use it in my example test, but I could create a helper method for it (similar to GetUDKParams) or simply access the AccountName from the UserDelegationCredential itself.

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,32 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
...
+	}
+	udk, err := sc.GetUserDelegationKey(ctx, info, &opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewUserDelegationCredential(url.Host, udk.UserDelegationKey), nil
+}
+
+func (s Client) GetUserDelegationKey(ctx context.Context, keyInfo KeyInfo, options *GetUserDelegationKeyOptions) GetUserDelegationKeyResponse {
```
jhendrixMSFT, line 107: Will customers ever directly need to access a `GetUserDelegationKeyResponse`?  What I'm wondering is if this method and `GetUserDelegationCredential` should be combined into a single `GetUserDelegationCredential` API.

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,32 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// GetUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func GetUserDelegationCredential(serviceURL string, ctx context.Context, info generated.KeyInfo, timeout *int32, requestID *string) (*UserDelegationCredential, error) {
```
jhendrixMSFT, line 88: Why is this a free-standing func?  Seems like it should be a method on the `Client` type.

jhendrixMSFT (changes requested):
Some clean up on the API public surface area.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
siminsavani-msft: It's actually used in sas/service.go (line 193) and sas/account.go (112) to get canonical name:

`stringToSign := strings.Join([]string{
		userDelegationCredential.AccountName(),`

jhendrixMSFT: If this is only needed internally, I think you can use an internal-only helper to get the value (just like for `getUDKParams`).

sdk/storage/azblob/client.go
```diff
@@ -158,3 +158,7 @@ func (c *Client) DownloadStream(ctx context.Context, containerName string, blobN
 	o = shared.CopyOptions(o)
 	return c.svc.NewContainerClient(containerName).NewBlobClient(blobName).DownloadStream(ctx, o)
 }
+
+func (c *Client) GetUserDelegationCredential(background context.Context, info service.KeyInfo, opts *service.GetUserDelegationKeyOptions) (*service.UserDelegationCredential, error) {
```
JeffreyRichter, line 162: Change 'background' to 'ctx'.
Leave 'opts' for now but we need to standardize on 'options' or 'o'

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+
+// UserDelegationCredential contains an account's name and its user delegation key.
+type UserDelegationCredential struct {
+	accountName       string
+	userDelegationKey UserDelegationKey
+}
+
+// AccountName returns the Storage account's Name
+func (f *UserDelegationCredential) getAccountName() string {
+	return f.accountName
```
JeffreyRichter, line 35: Make this method public

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+func (f *UserDelegationCredential) computeHMACSHA256(message string) (string, error) {
+	bytes, _ := base64.StdEncoding.DecodeString(*f.userDelegationKey.Value)
+	h := hmac.New(sha256.New, bytes)
+	_, err := h.Write([]byte(message))
+	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
+}
+
+// ComputeUDCHMACSHA256 is a helper method for computing the signed string outside of this package.
+func ComputeUDCHMACSHA256(udc *UserDelegationCredential, message string) (string, error) {
+	return udc.computeHMACSHA256(message)
```
JeffreyRichter, line 53: I'm curious, is there some reason why you have the private methods and the public wrapper functions?

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// GetUDKParams returns UserDelegationKey
+func (f *UserDelegationCredential) getUDKParams() *UserDelegationKey {
+	return &f.userDelegationKey
+}
+
+// GetUDKParams is a helper method for accessing the user delegation key parameters outside of this package.
+func GetUDKParams(udc *UserDelegationCredential) *UserDelegationKey {
+	return udc.getUDKParams()
```
JeffreyRichter, line 63: Same here

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+	userDelegationKey UserDelegationKey
+}
+
+// AccountName returns the Storage account's Name
+func (f *UserDelegationCredential) getAccountName() string {
+	return f.accountName
+}
+
+// GetUDKParams is a helper method for accessing the user delegation key parameters outside of this package.
+func GetAccountName(udc *UserDelegationCredential) string {
```
JeffreyRichter, line 39: Delete this method

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+	return udc.getAccountName()
+}
+
+// computeHMACSHA256 generates a hash signature for an HTTP request or for a SAS.
+func (f *UserDelegationCredential) computeHMACSHA256(message string) (string, error) {
```
JeffreyRichter, line 44: Make this method public

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+	bytes, _ := base64.StdEncoding.DecodeString(*f.userDelegationKey.Value)
+	h := hmac.New(sha256.New, bytes)
+	_, err := h.Write([]byte(message))
+	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
+}
+
+// ComputeUDCHMACSHA256 is a helper method for computing the signed string outside of this package.
+func ComputeUDCHMACSHA256(udc *UserDelegationCredential, message string) (string, error) {
```
JeffreyRichter, line 52: Delete this method

JeffreyRichter (commented):
This PR looks great now! Nice work.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
jhendrixMSFT: `UserDelegationCredential` is a publicly exposed type, and we don't want this (and other) APIs as part of the public surface area as they're meant to be used internally.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
jhendrixMSFT: Yes, this is to avoid exposing `ComputeUDCHMACSHA256` to customers while still being able to call it internally, outside of this package.

sdk/storage/azblob/client.go
jhendrixMSFT, line 162: Is a user delegation credential commonly used, enough that we want it on the tiny client?  I don't know enough about its usage.

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,23 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// GetUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func (s *Client) GetUserDelegationCredential(ctx context.Context, info generated.KeyInfo, o *GetUserDelegationKeyOptions) (*UserDelegationCredential, error) {
```
jhendrixMSFT, line 88: `generated.KeyInfo` isn't a publicly available type.  This will need a type alias.

jhendrixMSFT: Looks like the type alias already exists, just need to change the param type here.

sdk/storage/azblob/internal/exported/user_delegation_credential.go
jhendrixMSFT: See the comment on line 61.  We need this and the other helpers to keep some of these APIs internal.

sdk/storage/azblob/sas/service.go
```diff
@@ -151,15 +131,117 @@ func (v BlobSignatureValues) Sign(sharedKeyCredential *SharedKeyCredential) (Que
...
-	//	p.signedStart = udk.SignedStart
-	//	p.signedExpiry = udk.SignedExpiry
-	//	p.signedService = udk.SignedService
-	//	p.signedVersion = udk.SignedVersion
-	//}
+	return p, nil
+}
+
+// SignWithUDK uses an account's UserDelegationCredential to sign this signature values to produce the proper SAS query parameters.
+func (v BlobSignatureValues) SignWithUDK(userDelegationCredential *UserDelegationCredential) (QueryParameters, error) {
```
jhendrixMSFT, line 138: Thinking about this name a bit more.  I know that UDK is "user delegation key".  However, the type we use isn't a key, but a credential.  But perhaps instead of abbreviating this to UDC, what if we named it `SignWithUserDelegation`?  This is also similar to the pattern for `SignWithSharedKey`, where we use the credential's type name minus the "Credential" suffix.  What do you think?

siminsavani-msft: I agree with this! It definitely will be a lot clearer and as you mentioned, we are using a credential type here.

sdk/storage/azblob/client.go
siminsavani-msft: There are quite a few requests for it, and I think it could be beneficial having it on the tiny client.

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,23 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// GetUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func (s *Client) GetUserDelegationCredential(ctx context.Context, info KeyInfo, o *GetUserDelegationKeyOptions) (*UserDelegationCredential, error) {
```
jhendrixMSFT, line 88: The options type name is slightly different, it should match the method name with the "Options" suffix.

sdk/storage/azblob/client.go
jhendrixMSFT: Could we leave it out for the first beta, until we get some feedback?  We can always add it later but taking it away can be more difficult, and we're really trying to keep the API set on the tiny client as simple as possible.

sdk/storage/azblob/service/client.go
jhendrixMSFT: Sorry not the parameter name (what you had there was fine).  It's the type name that needs to be renamed to `GetUserDelegationCredentialOptions`

sdk/storage/azblob/client.go
siminsavani-msft: My intention was to address this comment (https://github.com/Azure/azure-sdk-for-go/pull/19141#discussion_r977982704), but I may have misinterpreted this? 

And yes, I can keep it out from the tiny client!

sdk/storage/azblob/service/client.go
siminsavani-msft: All good! I updated it now, thank you for clarifying :)

sdk/storage/azblob/service/models.go
```diff
@@ -21,6 +21,28 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 	return exported.NewSharedKeyCredential(accountName, accountKey)
 }
 
+// UserDelegationCredential contains an account's name and its user delegation key.
+type UserDelegationCredential = exported.UserDelegationCredential
+
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = generated.UserDelegationKey
+
+type KeyInfo = generated.KeyInfo
```
jhendrixMSFT, line 30: Missing doc comment.

sdk/storage/azblob/client.go
jhendrixMSFT: Oh I see what you're saying.  Having it a method on the `service.Client` is correct.  Let's just keep it off the tiny client for now until we get some feedback.

siminsavani-msft: Sure! Just updated :)

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,25 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// GetUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func GetUserDelegationCredential(serviceURL string, ctx context.Context, info KeyInfo, o *GetUserDelegationCredentialOptions) (*UserDelegationCredential, error) {
```
jhendrixMSFT, line 88: You had this correct before when it was a method on `Client`.

sdk/storage/azblob/service/models.go
```diff
@@ -21,6 +21,29 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
...
+type UserDelegationCredential = exported.UserDelegationCredential
+
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = generated.UserDelegationKey
+
+// KeyInfo contains KeyInfo struct.
+type KeyInfo = generated.KeyInfo
+
+// NewUserDelegationCredential creates a new UserDelegationCredential using a Storage account's name and a user delegation key from it
+func NewUserDelegationCredential(accountName string, udk UserDelegationKey) *UserDelegationCredential {
```
jhendrixMSFT, line 34: Since `GetUserDelegationCredential` now returns a`*UserDelegationCredential` do we need to export this API?

jhendrixMSFT (approved):

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+// UserDelegationKey contains UserDelegationKey.
+type UserDelegationKey = generated.UserDelegationKey
+
+// UserDelegationCredential contains an account's name and its user delegation key.
+type UserDelegationCredential struct {
+	accountName       string
+	userDelegationKey UserDelegationKey
+}
+
+// AccountName returns the Storage account's Name
```
JeffreyRichter, line 33: This comment isn't right

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+func (f *UserDelegationCredential) getAccountName() string {
+	return f.accountName
+}
+
+// GetUDKParams is a helper method for accessing the user delegation key parameters outside of this package.
```
JeffreyRichter, line 38: This comment isn't right

sdk/storage/azblob/internal/exported/user_delegation_credential.go
```diff
@@ -0,0 +1,64 @@
...
+	_, err := h.Write([]byte(message))
+	return base64.StdEncoding.EncodeToString(h.Sum(nil)), err
+}
+
+// ComputeUDCHMACSHA256 is a helper method for computing the signed string outside of this package.
+func ComputeUDCHMACSHA256(udc *UserDelegationCredential, message string) (string, error) {
+	return udc.computeHMACSHA256(message)
+}
+
+// GetUDKParams returns UserDelegationKey
```
JeffreyRichter, line 56: This comment isn't right

sdk/storage/azblob/service/client.go
```diff
@@ -82,6 +83,23 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
 	return NewClientWithNoCredential(parsed.ServiceURL, options)
 }
 
+// GetUserDelegationCredential obtains a UserDelegationKey object using the base ServiceURL object.
+// OAuth is required for this call, as well as any role that can delegate access to the storage account.
+func (s *Client) GetUserDelegationCredential(ctx context.Context, info KeyInfo, o *GetUserDelegationCredentialOptions) (*UserDelegationCredential, error) {
```
JeffreyRichter, line 88: I don't think we need this method. We DO need a method that returns a UDK but the customer can easily create a credential from this and so I see no reason to provide this additional functionality; we don't have it for any other kind of credential.

PR #19142, opened 2022-09-17
[azeventhubs] Updating the documentation and API surface based on arch board feedback

### LINKED ISSUE #19152: [azeventhubs] ConsumerClientOptions has obsolete options for start position and owner level but those should be deleted.
Current breakdown for client responsibilities is:
- ConsumerClient: AMQP connection 
- PartitionClient: AMQP _link_ parameters

I moved OwnerLevel and StartPosition (both link parameters) into PartitionClient, but didn't delete them from ConsumerClient. They're getting ignored there but they should be deleted altogether as they are link level parameters.

richardpark-msft:
Some great discussion in the archboard. There are still some bigger issues that need to be discussed, but these were some of the discussion that are easily resolveable:

- Added a lot of documentation to areas, making it clear which types require Closing, and added a bunch of self-referential docs as well between the type descriptions to get to examples, from sub-types to show which types actually instantiate them (ie: PartitionClient, ProcessorPartitionClient both mention their parent types and which function you use to create them).
- All types that used to say 'New*Options' types are properly renamed to just be *Options instead, avoiding 'New' clutter at the package level.
- Flatten the CheckpointStore's Ownership and Checkpoint types. This removed 3 extra clutter types (OwnershipData, CheckpointData and CheckpointStoreAddress). 

There were some other things that were wrong as well that I've fixed now:

- If your whole-file example function has more than one underscore in it it doesn't get listed in pkg.go.dev. So I fixed some of those - your naming convention needs to match Example_camelCaseOnly)
- Removed StartPosition and OwnerLevel fields from the ConsumerClientOptions (they are only supposed to be in the PartitionClientOptions).

Fixes #19152

### DIFF
--- a/sdk/messaging/azeventhubs/checkpoint_store.go
+++ b/sdk/messaging/azeventhubs/checkpoint_store.go
@@ -26,39 +26,25 @@ type CheckpointStore interface {
 
 // Ownership tracks which consumer owns a particular partition.
 type Ownership struct {
-	CheckpointStoreAddress
-	OwnershipData
-}
-
-// OwnershipData is data specific to ownership, like the
-// current owner, by ID, and the last time the ownership was
-// updated, which is used to calculate if ownership has expired.
-type OwnershipData struct {
-	OwnerID          string
-	LastModifiedTime time.Time
-	ETag             string
-}
-
-// Checkpoint tracks the last succesfully processed event in a partition.
-type Checkpoint struct {
-	CheckpointStoreAddress
-	CheckpointData
-}
-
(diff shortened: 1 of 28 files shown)


richardpark-msft:
@LarryOsterman : This is the revamped example file, which I think does a better job of explaining. I can't show the rendered help until it's been published, but Processor.go also makes it more clear when things are blocking, etc..

https://github.com/Azure/azure-sdk-for-go/blob/c0768b86542ac5320e31efbe6161944bc20fa169/sdk/messaging/azeventhubs/example_processor_test.go

jhendrixMSFT (approved):

PR #19181, opened 2022-09-21
[BOT RULES] Recovery Services Backup

ronniegeraghty:


ronniegeraghty:
@RickWinter not sure why the tests above are failing, but this is just the addition of some mentions for when the Service Attention and Recovery Service Backup labels are added to an issue. Can these ADO tests be overridden?

RickWinter (approved):

PR #19210, opened 2022-09-27
add `Cancelled` as a terminal state for LRO

### LINKED ISSUE #19209: Can `Cancelled` be added as a terminal state?
This is related to https://github.com/Azure/azure-resource-manager-rpc/issues/144 where `Cancelled` is used in a large percentage of specifications instead of `Canceled`. You can see it is used a lot in this repository https://github.com/Azure/azure-sdk-for-go/search?q=cancelled . 

Rather than adding `Canceled` to the specifications that already use `Cancelled`, would it not be a better user experience to update the `IsTerminalState` to support the alternative spelling?

https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/internal/pollers/util.go#L24-L40

cataggar:
Fix #19209.

### DIFF
--- a/sdk/azcore/internal/pollers/util.go
+++ b/sdk/azcore/internal/pollers/util.go
@@ -25,18 +25,19 @@ import (
 const (
 	StatusSucceeded  = "Succeeded"
 	StatusCanceled   = "Canceled"
+	StatusCancelled  = "Cancelled"
 	StatusFailed     = "Failed"
 	StatusInProgress = "InProgress"
 )
 
 // IsTerminalState returns true if the LRO's state is terminal.
 func IsTerminalState(s string) bool {
-	return strings.EqualFold(s, StatusSucceeded) || strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled)
+	return strings.EqualFold(s, StatusSucceeded) || strings.EqualFold(s, StatusFailed) || strings.EqualFold(s, StatusCanceled) || strings.EqualFold(s, StatusCancelled)
 }
 
 // Failed returns true if the LRO's state is terminal failure.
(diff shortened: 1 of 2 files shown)


sdk/azcore/internal/pollers/util.go
```diff
@@ -25,18 +25,19 @@ import (
 const (
 	StatusSucceeded  = "Succeeded"
 	StatusCanceled   = "Canceled"
+	StatusCancelled  = "Cancelled"
```
jhendrixMSFT, line 28: This violates the [ARM RPC spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#provisioningstate-property) and is incorrectly spelled.  If services are returning this, then they need to be fixed.

jhendrixMSFT (changes requested):

sdk/azcore/internal/pollers/util.go
cataggar: I think it makes sense to change the ARM RPC spec since a large percentage of existing specifications use that valid spelling.

RickWinter: @JeffreyRichter @mikekistler, Should there be a sweep through all the REST specs to correct them to follow the RPC spec?  Adding support for something that violates the RPC spec is not something we want to do.

jhendrixMSFT: Ultimately, the ARM team will need to make this call.

cataggar: I did the analysis. There are only 8 control plane services affected. The one I manage, plus 7 others. All log bugs for each.

PR #19216, opened 2022-09-27
[azservicebus] Make it so retry sleeps can be cancelled.

richardpark-msft:
The sleep between retries was not cancellable but it should have been. This changes it to use a context.WithTimeout() instead.

### DIFF
--- a/sdk/messaging/azservicebus/internal/utils/retrier.go
+++ b/sdk/messaging/azservicebus/internal/utils/retrier.go
@@ -52,12 +52,10 @@ func Retry(ctx context.Context, eventName log.Event, operation string, fn func(c
 			sleep := calcDelay(ro, i)
 			log.Writef(eventName, "(%s) Retry attempt %d sleeping for %s", operation, i, sleep)
 
-			sleepCtx, cancelSleep := context.WithTimeout(ctx, sleep)
-			<-sleepCtx.Done()
-			cancelSleep()
-
-			if ctx.Err() != nil {
+			select {
+			case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


sdk/messaging/azservicebus/internal/utils/retrier.go
```diff
@@ -51,7 +51,14 @@ func Retry(ctx context.Context, eventName log.Event, operation string, fn func(c
 		if i > 0 {
 			sleep := calcDelay(ro, i)
 			log.Writef(eventName, "(%s) Retry attempt %d sleeping for %s", operation, i, sleep)
-			time.Sleep(sleep)
+
+			sleepCtx, cancelSleep := context.WithTimeout(ctx, sleep)
```
jhendrixMSFT, line 55: This is the typical pattern we use elsewhere.

```go
select {
case <-ctx.Done():
	return ctx.Err()
case <-time.After(sleep):
	// sleep elapsed
}
```

richardpark-msft: Yeah, I was aware of that one, but I did think this was pretty nice too.

I'll change it so we're consistent.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/internal/utils/retrier.go
richardpark-msft: Done! One nice advantage of the default approach is that we don't have to check ctx.Err() vs our local context. So I think it's clearer now.

PR #19240, opened 2022-09-30
Change ErrorInfo Format

gracewilcox:
This pull request changes the format of the ErrorInfo struct.

When you execute a single query (QueryWorkspace), the operation returns a Results struct and error (error is an azcore.ResponseError if there was a failure).

Additionally, the Results struct contains inside it an ErrorInfo struct. Sometimes, there is partial error information from Azure Monitor contained inside it. For example, an "EngineOverload" error.

The current structure of the ErrorInfo struct is below:
```
|---Error *ErrorInfo
	|---Code *string
	|---Message *string
	|---AdditionalProperties interface{}
	|---Details []*ErrorDetail
		|---Code *string
		|---Message *string
		|---AdditionalProperties interface{}
		|---Resources []*string
		|---Target *string
		|---Value *string
	|---Innererror *ErrorInfo
```

New structure is:
```
|---Error *ErrorInfo
	|---Code *string // required
	|---Message *string // required
	|---RawMessage json.RawMessage // contains entire error byte array
```

New structure prevents users from using in the logic of their code properties that can change over time. For example, the service can change the contents of ErrorDetail at any time.


sdk/monitor/azquery/custom_client.go
```diff
@@ -45,3 +48,46 @@ func NewMetricsClient(credential azcore.TokenCredential, options *MetricsClientO
...
+// ErrorInfo - The code and message for an error.
+type ErrorInfo struct {
+	// REQUIRED; A machine readable error code.
+	Code *string `json:"code,omitempty"`
+
+	// REQUIRED; A human readable error message.
+	Message *string `json:"message,omitempty"`
+
+	// A description in detail why the operation failed.
+	RawMessage json.RawMessage
```
JeffreyRichter, line 61: Joel, is this really what we want here? I can't remember our azcore.ResponseError but I thought it had a public Code field and a String() method that returns the raw JSON; anything else is private. I think we should mimic the azcore.ResponseError type.

gracewilcox: resolved in person with Jeff, new format detailed below

gracewilcox:
Talked it over with Jeff and the new format is as follows:
```
|---Error *ErrorInfo
	|---Code string // error code
	|---data []byte // full error message
```

Pull request also includes a new custom unmarshalor for `ErrorInfo`. `ErrorInfo` also now implements error interface. Users can use the Error() method to display the full error message.

JeffreyRichter (approved):

PR #19278, opened 2022-10-04
JoinPaths: don't add "/" before the query string

### LINKED ISSUE #19245: azblob v0.5.0 and blockBlobClient path joining behaviour
### Bug Report

- import path of package in question
```
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
```
- SDK version: `v0.5.0` of `sdk/storage/azblob`
- output of `go version`:   ` go1.18.2 linux/amd64`

- What happened?

Given code similar to:

```
package main

import (
        "context"
        "fmt"

        "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
        "github.com/sirupsen/logrus"
)

func main() {
        ctx := context.Background()
        token := "?some token goes here"
        containerURL := fmt.Sprintf("https://mystorage.blob.core.windows.net/mycontainer%s", token)
        containerClient, err := container.NewClientWithNoCredential(containerURL, nil)
        if err != nil {
                logrus.Fatal(err)
        }

        blobURL := "test.txt"
        blobClient := containerClient.NewBlockBlobClient(blobURL)
        props, err := blobClient.GetProperties(ctx, nil)
        if err != nil {
                logrus.Fatal(err)
        }
        fmt.Println(props.ContentMD5)
}
```

Results in the error:

```
FATA[0000] HEAD https://mystorage.blob.core.windows.net/mycontainer/test.txt/
RESPONSE 404: 404 The specified blob does not exist.
ERROR CODE: BlobNotFound
Response contained no body
```

Note the extra forward slash "/" at the end. This is because when instantiating the client the code calls runtime.JoinPaths [here](https://github.com/Azure/azure-sdk-for-go/blob/sdk/storage/azblob/v0.5.0/sdk/storage/azblob/container/client.go#L124). 

 Which is this code [here](https://github.com/Azure/azure-sdk-for-go/blob/sdk/azcore/v1.1.3/sdk/azcore/runtime/request.go#L48).

I was expecting I would be able to create a container client and a blockblock client, passing in the blob name and that would work but it results in this extra slash. Is this intended behaviour or am I not supposed to create the blockblob client in this way?
Should I be using the blockblob package and instantiating a blockblob.NewClientWithNoCredential instead?

It is worth noting that if I use `azidentity.NewDefaultAzureCredential` and `container.NewClient` passing in the credential - then I do not encounter this issue. Which makes me think `containerClient.NewBlockBlobClient` is only intended to be used when passing in an AAD credential. So perhaps this isn't a bug but intended behaviour. However, it would be good if there wasn't this difference between using SAS tokens and AAD tokens.

- How can we reproduce it?
Create a storage account, container and SAS token. Replace the values in the code and run it.

N/A

drakkan:
Fixes #19245

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bugs Fixed
 * Don't retry a request if the `Retry-After` delay is greater than the configured `RetryOptions.MaxRetryDelay`.
+* JoinPaths: do not unconditionally add a forward slash before the query string
 
 ### Other Changes
 * Removed logging URL from retry policy as it's redundant.
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 * Don't retry a request if the `Retry-After` delay is greater than the configured `RetryOptions.MaxRetryDelay`.
-* JoinPaths: do not unconditionally add a forward slash before the query string
+* `runtime.JoinPaths`: do not unconditionally add a forward slash before the query string
 
 ### Other Changes
 * Removed logging URL from retry policy as it's redundant.
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -57,26 +57,23 @@ func JoinPaths(root string, paths ...string) string {
 		root, qps = splitPath[0], splitPath[1]
 	}
 
-	var sb strings.Builder
-
-	sb.WriteString(strings.TrimRight(root, "/"))
-
-	joinedPaths := path.Join(paths...)
-	joinedPaths = strings.TrimLeft(joinedPaths, "/")
-
-	sb.WriteString("/")
-	sb.WriteString(joinedPaths)
+	p := path.Join(paths...)
 	// path.Join will remove any trailing slashes.
-	// Preserve a trailing slash if the last path element has it
-	if strings.HasSuffix(paths[len(paths)-1], "/") {
-		sb.WriteString("/")
+	// if one was provided, preserve it.
+	if strings.HasSuffix(paths[len(paths)-1], "/") && !strings.HasSuffix(p, "/") {
+		p += "/"
 	}
 
 	if qps != "" {
-		sb.WriteString("?")
-		sb.WriteString(qps)
+		p = p + "?" + qps
 	}
-	return sb.String()
+
+	if strings.HasSuffix(root, "/") && strings.HasPrefix(p, "/") {
+		root = root[:len(root)-1]
+	} else if !strings.HasSuffix(root, "/") && !strings.HasPrefix(p, "/") {
+		p = "/" + p
+	}
+	return root + p
 }
 
 // EncodeByteArray will base-64 encode the byte slice v.
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
+	"path"
 	"reflect"
 	"strings"
 	"time"
@@ -56,19 +57,26 @@ func JoinPaths(root string, paths ...string) string {
 		root, qps = splitPath[0], splitPath[1]
 	}
 
-	for i := 0; i < len(paths); i++ {
-		root = strings.TrimRight(root, "/")
-		paths[i] = strings.TrimLeft(paths[i], "/")
-		root += "/" + paths[i]
+	var sb strings.Builder
+
+	sb.WriteString(strings.TrimRight(root, "/"))
+
+	joinedPaths := path.Join(paths...)
+	joinedPaths = strings.TrimLeft(joinedPaths, "/")
+
+	sb.WriteString("/")
+	sb.WriteString(joinedPaths)
+	// path.Join will remove any trailing slashes.
+	// Preserve a trailing slash if the last path element has it
+	if strings.HasSuffix(paths[len(paths)-1], "/") {
+		sb.WriteString("/")
 	}
 
 	if qps != "" {
-		if !strings.HasSuffix(root, "/") {
-			root += "/"
-		}
-		return root + "?" + qps
+		sb.WriteString("?")
+		sb.WriteString(qps)
 	}
-	return root
+	return sb.String()
 }
 
 // EncodeByteArray will base-64 encode the byte slice v.
--- a/sdk/azcore/runtime/request_test.go
+++ b/sdk/azcore/runtime/request_test.go
@@ -581,6 +581,16 @@ func TestJoinPaths(t *testing.T) {
 			paths:    []string{"path/one", "path/two/"},
 			expected: "http://test.contoso.com/path/one/path/two/?qp1=abc&qp2=def",
 		},
+		{
+			root:     "http://test.contoso.com/?qp1=abc&qp2=def",
+			paths:    []string{"path/one", "path/two/"},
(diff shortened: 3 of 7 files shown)


jhendrixMSFT:
When you fix the broken test can you please also add an entry to the changelog?

drakkan:
> When you fix the broken test can you please also add an entry to the changelog?

done. I think it is ok to omit the slash before a query string. The url is valid and this change fix the linked issue

drakkan:
Maybe we should always remove the trailing slash, something like this:

```
func JoinPaths(root string, paths ...string) string {
	if len(paths) == 0 {
		return root
	}

	qps := ""
	if strings.Contains(root, "?") {
		splitPath := strings.Split(root, "?")
		root, qps = splitPath[0], splitPath[1]
	}

	joinedPaths := path.Join(paths...)
	root = strings.TrimRight(root, "/")
	joinedPaths = strings.TrimLeft(joinedPaths, "/")
	root += "/" + joinedPaths

	if qps != "" {
		return root + "?" + qps
	}
	return root
}
```

what do you think about? Thanks

jhendrixMSFT:
I was thinking about `TrimRight` earlier too. It seems like with your fix that `TrimRight` wouldn't need to do any work. However, having it in place couldn't hurt especially if it catches some edge-case I'm overlooking.

drakkan:
> I was thinking about `TrimRight` earlier too. It seems like with your fix that `TrimRight` wouldn't need to do any work. However, having it in place couldn't hurt especially if it catches some edge-case I'm overlooking.

`TrimRight` is required. If you have:

- root = `http://test.contoso.com/`
- paths = `[]string{"/path/one", "path/two", "/path/three/", "path/four/"}`

you get `http://test.contoso.com//path/one/path/two/path/three/path/four` if you remove `TrimRight`.

Refactored the to use `path.Join` and `strings.Builder`, added some more test cases

jhendrixMSFT (dismissed):

drakkan:
> I was discussing this with @chlowell and he brought up a good point that `http://foo/?bar=baz` is valid, although unlikely. So maybe it would be better to preserve a trailing `/` but don't add one by default. Thoughts?

`http://foo/?bar=baz`, `http://foo?bar=baz`, `http://foo/baz/?bar=baz`, `http://foo/baz?bar=baz` are all valid URIs as far as I know. The problem here is that an URI like `http://foo/baz/?bar=baz` (even if valid) creates issues for some azblob calls. I don't know azure URL requirements. If you prefer I have no problems to update the patch and preverse the trailing slash. Let me know, thanks

jhendrixMSFT:
That's what Charles and I were discussing. Probably best that `azcore` is unopinionated and preserves what's passed to it but doesn't add it.

jhendrixMSFT:
Hmm this might be problematic in some cases.  E.g. when obtaining a SAS URL for a storage service account via the portal, there will be a trailing `/`.

`https://jhendrixstorage1.blob.core.windows.net/?sv=2021-06-08&...`

Preserving the `/` in this case while joining container/blob names will be problematic.

So, while I still think that having `JoinPaths` be unopinionated is correct, we'll also need to fix this in `azblob`.

drakkan:
> Hmm this might be problematic in some cases. E.g. when obtaining a SAS URL for a storage service account via the portal, there will be a trailing `/`.
> 
> `https://jhendrixstorage1.blob.core.windows.net/?sv=2021-06-08&...`
> 
> Preserving the `/` in this case while joining container/blob names will be problematic.
> 
> So, while I still think that having `JoinPaths` be unopinionated is correct, we'll also need to fix this in `azblob`.

what do you think about something like this?

```
func JoinPaths(root string, paths ...string) string {
	if len(paths) == 0 {
		return root
	}

	qps := ""
	if strings.Contains(root, "?") {
		splitPath := strings.Split(root, "?")
		root, qps = splitPath[0], splitPath[1]
	}
	hasTrailingSlash := strings.HasSuffix(paths[len(paths)-1], "/")

	var sb strings.Builder

	sb.WriteString(strings.TrimRight(root, "/"))

	joinedPaths := path.Join(paths...)
	joinedPaths = strings.TrimLeft(joinedPaths, "/")

	sb.WriteString("/")
	sb.WriteString(joinedPaths)
	if hasTrailingSlash {
		sb.WriteString("/")
	}

	if qps != "" {
		sb.WriteString("?")
		sb.WriteString(qps)
	}
	return sb.String()
}
```

the trailing slash is preserved if provided in the last path element

jhendrixMSFT (approved):

drakkan:
@jhendrixMSFT thanks! I did a quick test and I confirm the merged patch fix the reported issue. I also wrote a small benchmark comparing the string builder approach with the used string concatenation

```
func BenchmarkJoinPaths(b *testing.B) {
	url := "http://192.168.1.12:10000/devstoreaccount1?sv=2021-04-10&ss=b&srt=sco&st=2022-10-04T17%3A30%3A22Z&se=2022-10-05T17%3A30%3A22Z&sp=rwdl&sig=3Zuos3PufKoEBM5OB9KbM8LJFZvT0JZvHNtoCf0kkSg%3D"
	blobName := "test"
	expected := "http://192.168.1.12:10000/devstoreaccount1/test?sv=2021-04-10&ss=b&srt=sco&st=2022-10-04T17%3A30%3A22Z&se=2022-10-05T17%3A30%3A22Z&sp=rwdl&sig=3Zuos3PufKoEBM5OB9KbM8LJFZvT0JZvHNtoCf0kkSg%3D"
	for i := 0; i < b.N; i++ {
		if p := JoinPaths(url, blobName); p != expected {
			panic(p)
		}
	}
}

func BenchmarkJoinPathsSB(b *testing.B) {
	url := "http://192.168.1.12:10000/devstoreaccount1?sv=2021-04-10&ss=b&srt=sco&st=2022-10-04T17%3A30%3A22Z&se=2022-10-05T17%3A30%3A22Z&sp=rwdl&sig=3Zuos3PufKoEBM5OB9KbM8LJFZvT0JZvHNtoCf0kkSg%3D"
	blobName := "test"
	expected := "http://192.168.1.12:10000/devstoreaccount1/test?sv=2021-04-10&ss=b&srt=sco&st=2022-10-04T17%3A30%3A22Z&se=2022-10-05T17%3A30%3A22Z&sp=rwdl&sig=3Zuos3PufKoEBM5OB9KbM8LJFZvT0JZvHNtoCf0kkSg%3D"
	for i := 0; i < b.N; i++ {
		if p := JoinPathsSB(url, blobName); p != expected {
			panic(p)
		}
	}
}
```

here the results:

```
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkJoinPaths-12    	 1623058	       616.4 ns/op	     552 B/op	       6 allocs/op

cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkJoinPathsSB-12    	 3126475	       432.1 ns/op	     280 B/op	       5 allocs/op
```

performance is OK even with the current approach, we can optimize the code in the future, thanks

jhendrixMSFT:
Thanks for the analysis. I more or less copied this from `url.Join()` in Go 1.19 thinking if it was good enough for the standard library, we can use it too.

Agreed there's future optimization work (and not just in this func).

drakkan:
> Thanks for the analysis. I more or less copied this from `url.Join()` in Go 1.19 thinking if it was good enough for the standard library, we can use it too.
> 
> Agreed there's future optimization work (and not just in this func).

I agree, v0.5.0 is a big step forward in usability over previous versions and `NewListBlobsHierarchyPager` also works much better (no longer making unnecessary recursive requests). Performance can be improved in the future. Thank's for your work.

PR #19281, opened 2022-10-04
[azeventhubs] Adding in prefetch and far more stress tests

richardpark-msft:
- Adding in prefetch, making it the default. Event Hubs has no message expiration or server side locks (like Service Bus) so it's a great fit.
- Adding in tests for more long running operations, adding variations for with and without prefetch, with and without the processor.
- Added in a new ConsumerClient.ID() field so a user can get the unique identifier for the ConsumerClient


sdk/messaging/azeventhubs/internal/namespace.go
```diff
@@ -259,9 +261,10 @@ func (ns *Namespace) Recover(ctx context.Context, theirConnID uint64) (bool, err
 	}
 
 	if ns.connID != theirConnID {
-		log.Writef(exported.EventConn, "Skipping connection recovery, already recovered: %d vs %d", ns.connID, theirConnID)
 		// we've already recovered since the client last tried.
-		return false, nil
+		// they should recover their links since their connection is old.
+		log.Writef(exported.EventConn, "Skipping connection recovery, already recovered: %d vs %d. Links will still be recovered.", ns.connID, theirConnID)
+		return true, nil
```
jhendrixMSFT, line 267: The fact that this used to return `false` was just a bug?

richardpark-msft: Yes. I'll be making this same change in azservicebus before we ship the next hotfix.

jhendrixMSFT: Probably worth noting in the changelog then.

richardpark-msft: Added.

richardpark-msft: Thinking about this more, I want to look at the code. I think it went through some evolutions, and I want to look at it fresh.

richardpark-msft: Talked with @jhendrixMSFT offline - I'm going to revert this because I think there are some more cases I want to cover and test. Will do that in the next PR so we can keep this one simple.

richardpark-msft:
@jhendrixMSFT  - recovery fix has been rolled back. Will address in next PR.

jhendrixMSFT (approved):

Issue #19249, opened 2022-10-03
Blob user delegation SAS wrong service version


yvespp:
### Bug Report

- import path of package in question:
```
"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
azcontainer "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
```
- SDK version e.g. `0.5.0`, lock file:
```
github.com/Azure/azure-sdk-for-go v66.0.0+incompatible h1:bmmC38SlE8/E81nNADlgmVGurPWMHDX2YNXVQMrBpEE=
github.com/Azure/azure-sdk-for-go v66.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
github.com/Azure/azure-sdk-for-go/sdk/azcore v1.1.3 h1:8LoU8N2lIUzkmstvwXvVfniMZlFbesfT2AmA1aqvRr8=
github.com/Azure/azure-sdk-for-go/sdk/azcore v1.1.3/go.mod h1:uGG2W01BaETf0Ozp+QxxKJdMBNRWPdstHG0Fmdwn1/U=
github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.1.0 h1:QkAcEIAKbNL4KoFr4SathZPhDhF4mVwpBMFlYjyAqy8=
github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.1.0/go.mod h1:bhXu1AjYL+wutSL/kpSq6s7733q2Rb0yuot9Zgfqa/0=
github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.1 h1:XUNQ4mw+zJmaA2KXzP9JlQiecy1SI+Eog7xVkPiqIbg=
github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.1/go.mod h1:eWRD7oawr1Mu1sLCawqVc0CUiF43ia3qQMxLscsKQ9w=
[...]

PR #19299, opened 2022-10-06
Updating SAS Version 

siminsavani-msft:
Updating the SAS Version won't affect any other string-to-string algorithms and is needed for UserDelegationSAS
(Fixes https://github.com/Azure/azure-sdk-for-go/issues/19249)

+ fixing comments to UDC file and test


sdk/storage/azblob/sas/query_params.go
```diff
@@ -23,7 +23,7 @@ const (
 
 var (
 	// Version is the default version encoded in the SAS token.
-	Version = "2019-12-12"
+	Version = "2020-02-10"
```
jhendrixMSFT, line 26: This value is also used in the `SignWithSharedKey()` APIs. You're sure this won't break them?

siminsavani-msft: I'm confident! The string-to-sign version 2020-12-06 is the one that will need some changes, but 2020-02-10 will fall under this format (https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#version-2018-11-09-and-later) which is our current string-to-sign behavior.

jhendrixMSFT: Thanks for confirming!

jhendrixMSFT (approved):

jhendrixMSFT:
Could you also please update `azcore` to `v1.1.4` please and add an entry to the changelog in the "other" section? This resolves an issue with incorrect path joining.

jhendrixMSFT (approved):

PR #19304, opened 2022-10-06
Retain contents of read-only fields when sending requests

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/843

### DIFF
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -95,7 +95,7 @@ func MarshalAsByteArray(req *policy.Request, v []byte, format Base64Encoding) er
 
 // MarshalAsJSON calls json.Marshal() to get the JSON encoding of v then calls SetBody.
 func MarshalAsJSON(req *policy.Request, v interface{}) error {
-	if omit := os.Getenv("AZURE_SDK_GO_OMIT_READONLY"); omit != "" {
+	if omit := os.Getenv("AZURE_SDK_GO_OMIT_READONLY"); omit == "true" {
 		v = cloneWithoutReadOnlyFields(v)
 	}
 	b, err := json.Marshal(v)
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
+	"os"
 	"path"
 	"reflect"
 	"strings"
@@ -94,7 +95,9 @@ func MarshalAsByteArray(req *policy.Request, v []byte, format Base64Encoding) er
 
 // MarshalAsJSON calls json.Marshal() to get the JSON encoding of v then calls SetBody.
 func MarshalAsJSON(req *policy.Request, v interface{}) error {
-	v = cloneWithoutReadOnlyFields(v)
+	if omit := os.Getenv("AZURE_SDK_GO_OMIT_READONLY"); omit != "" {
+		v = cloneWithoutReadOnlyFields(v)
+	}
 	b, err := json.Marshal(v)
 	if err != nil {
 		return fmt.Errorf("error marshalling type %T: %s", v, err)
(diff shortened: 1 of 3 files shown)


sdk/azcore/runtime/request_test.go
```diff
@@ -341,6 +342,10 @@ func TestCloneWithoutReadOnlyFieldsEndToEnd(t *testing.T) {
 		ID:   &id,
 		Name: &name,
 	}
+
+	os.Setenv("AZURE_SDK_GO_OMIT_READONLY", "true")
+	defer os.Unsetenv("AZURE_SDK_GO_OMIT_READONLY")
```
chlowell, line 347:
```suggestion
	t.Setenv("AZURE_SDK_GO_OMIT_READONLY", "true")
```

sdk/azcore/runtime/request.go
```diff
@@ -94,7 +95,9 @@ func MarshalAsByteArray(req *policy.Request, v []byte, format Base64Encoding) er
 
 // MarshalAsJSON calls json.Marshal() to get the JSON encoding of v then calls SetBody.
 func MarshalAsJSON(req *policy.Request, v interface{}) error {
-	v = cloneWithoutReadOnlyFields(v)
+	if omit := os.Getenv("AZURE_SDK_GO_OMIT_READONLY"); omit != "" {
```
richardpark-msft, line 98: I'm assuming this is related to a conversation that happeened in the arch board.

Are we sure this should be a global setting, and not one that's passed in so we can change it for each service, if necessary?

jhendrixMSFT: @JeffreyRichter and I debated this approach. For large code bases, it could be expensive for customers to toggle the setting at the client/API level. There was also some discussion with the ARM team about our approach and based on the provided [docs](https://armwiki.azurewebsites.net/api_contracts/guidelines/openapi.html) we think this is an acceptable path.

chlowell, line 98: `AZURE_SDK_GO_OMIT_READONLY=false` gets the same behavior as `AZURE_SDK_GO_OMIT_READONLY=true`. I'm okay with that provided we document that any value toggles the behavior, but then I think we should allow "":
```suggestion
	if _, ok := os.LookupEnv("AZURE_SDK_GO_OMIT_READONLY"); ok {
```

richardpark-msft: If we did ever decide to make it configurable, per-client, would we say that the newly added setting (in code, not the environment variable) would override the env variable's value?

In a similar vein, do we think people will set AZURE_SDK_GO_OMIT_READONLY='no' or 'false' and expect it to change the meaning?

jhendrixMSFT: The nuance here that I probably should have included in the PR description is that we don't want to publicly document this env var. The thinking is as follows.
- if a service is rejecting a request due to mutation of a RO field, the customer would be required to update their code
- if a service is rejecting a request due to inclusion of a RO field that hasn't been mutated, then the service must be fixed per the ARM guidelines. customers can use this escape hatch to get unblocked in the meantime

Ideally, support for this env var would be removed at some point, and the fact that it's undocumented _should_ in theory give us the latitude to do so (similar to undocumented compiler switches).

The undocumented aspect is why I was a bit loose with the allowed values, but maybe that wasn't such a great idea. I'll clean that up.

jhendrixMSFT: We really want to never make this configurable. It's just an escape hatch if a service has incorrect behavior.

richardpark-msft: Based on yoiur reply to @chlowell, I'm good. Adding some strictness around it would be good, and removing it at some point (which sounds like our plan) is even better.

sdk/azcore/runtime/request_test.go
jhendrixMSFT: That's awesome, I didn't know about this feature.

sdk/azcore/runtime/request.go
jhendrixMSFT: I've changed this to only accept `true` to enable this.

JeffreyRichter (approved):

chlowell (approved):

PR #19322, opened 2022-10-10
Add support for making recordings external to the repository

benbp:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4257

This adds support to the recording package to work with [test proxy asset sync](https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/documentation/asset-sync/README.md), which will allow the test recordings to be moved out of the go repository into https://github.com/Azure/azure-sdk-assets.

The primary behavior change is: if an `assets.json` file is found in the current test service directory or any parent, the recording framework will request external assets via the test proxy during test setup. If there is no `assets.json`, then pre-existing recordings will be used instead. This change should support running tests via `run_tests.ps1` or invoking `go test` directly.

I also added some minor improvements to the testing scripts.

- Improve error and pipeline handling in run_tests.ps1
- Add test proxy default assets directory to gitignore
- Support test proxy external asset mode for test recordings

### DIFF
--- a/eng/scripts/run_tests.ps1
+++ b/eng/scripts/run_tests.ps1
@@ -3,7 +3,7 @@
 Param(
     [Parameter(Mandatory = $true)]
     [string] $serviceDirectory,
-    [string] $testTimeout = "1s"
+    [string] $testTimeout = "10s"
 )
 
 $ErrorActionPreference = 'Stop'
--- a/eng/scripts/run_tests.ps1
+++ b/eng/scripts/run_tests.ps1
@@ -1,8 +1,9 @@
 #Requires -Version 7.0
 
 Param(
+    [Parameter(Mandatory = $true)]
     [string] $serviceDirectory,
-    [string] $testTimeout
+    [string] $testTimeout = "1s"
 )
 
 $ErrorActionPreference = 'Stop'
--- a/eng/scripts/run_tests.ps1
+++ b/eng/scripts/run_tests.ps1
@@ -5,6 +5,8 @@ Param(
     [string] $testTimeout
 )
 
+$ErrorActionPreference = 'Stop'
+
 Push-Location sdk/$serviceDirectory
 Write-Host "##[command] Executing 'go test -timeout $testTimeout -v -coverprofile coverage.txt ./...' in sdk/$serviceDirectory"
 
@@ -12,7 +14,7 @@ go test -timeout $testTimeout -v -coverprofile coverage.txt ./... | Tee-Object -
 # go test will return a non-zero exit code on test failures so don't skip generating the report in this case
 $GOTESTEXITCODE = $LASTEXITCODE
 
-Get-Content outfile.txt | go-junit-report > report.xml
+Get-Content -Raw outfile.txt | go-junit-report > report.xml
 
 # if no tests were actually run (e.g. examples) delete the coverage file so it's omitted from the coverage report
 if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simplematch -quiet) {
@@ -32,8 +34,8 @@ if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simpl
     Get-Content ./coverage.json | gocov-xml > ./coverage.xml
     Get-Content ./coverage.json | gocov-html > ./coverage.html
 
-    Move-Item ./coverage.xml $repoRoot
-    Move-Item ./coverage.html $repoRoot
+    Move-Item -Force ./coverage.xml $repoRoot
+    Move-Item -Force ./coverage.html $repoRoot
 
     # use internal tool to fail if coverage is too low
     Pop-Location
--- a/sdk/internal/CHANGELOG.md
+++ b/sdk/internal/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+Support test recording assets external to repository
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/internal/CHANGELOG.md
+++ b/sdk/internal/CHANGELOG.md
@@ -1,16 +1,10 @@
 # Release History
 
-## 1.0.2 (Unreleased)
+## 1.1.0 (2022-10-20)
 
 ### Features Added
 
-Support test recording assets external to repository
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+* Support test recording assets external to repository
 
 ## 1.0.1 (2022-08-22)
 
--- a/sdk/internal/recording/recording.go
+++ b/sdk/internal/recording/recording.go
@@ -20,7 +20,6 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -576,7 +575,7 @@ func (r RecordingOptions) baseURL() string {
 }
 
 func getTestId(pathToRecordings string, t *testing.T) string {
-	return path.Join(pathToRecordings, "recordings", t.Name()+".json")
+	return filepath.Join(pathToRecordings, "recordings", t.Name()+".json")
 }
 
 func getGitRoot(fromPath string) (string, error) {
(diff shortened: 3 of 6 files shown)


eng/scripts/run_tests.ps1
```diff
@@ -32,8 +35,8 @@ if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simpl
     Get-Content ./coverage.json | gocov-xml > ./coverage.xml
     Get-Content ./coverage.json | gocov-html > ./coverage.html
 
-    Move-Item ./coverage.xml $repoRoot
-    Move-Item ./coverage.html $repoRoot
+    Move-Item -Force ./coverage.xml $repoRoot
```
RickWinter, line 38: This will overwrite the file if it exists.  If the file exists can it be renamed first or prompted before overwriting?

Maybe its worth considering putting all output into directories that are specific to the run. i.e:
<rootdir>\builds\<timestamp>\

benbp: I guess it's not clear to me how this script is used more generally. I know it's used in the pipeline, but do developers also use it for testing or prefer straight `go test` commands? I ran into this because I was running the script back to back in dev, and had to remove the file each time otherwise it would fail at the end.

We could back up the files, but if someone is just using it to test locally they may end up with a large number of files.

benbp:
Looks like windows tests are failing due to some path handling differences with the OS. Investigating...

eng/scripts/run_tests.ps1
richardpark-msft: Since it's just code coverage output archiving is probably not worth it - it's totally ephemeral and just rerunning tests will recreate it. I'm guessing this just moves it to the root for some other tool to pick it up later.

richardpark-msft: @jhendrixMSFT , @gracewilcox , @chlowell  - do any of you use this script recreationally to run tests? I just use `go test` myself.

jhendrixMSFT: I don't. Overwriting it should be fine.

benbp:
@jhendrixMSFT @richardpark-msft do you anticipate a future requirement to define configs for recording assets in directories _below_ the service directory (e.g. deeper than `sdk/keyvault/azkeys`)? That is to say, will we need more granularity with recordings than the main package? I've skipped that functionality for now assuming it won't be needed.

jhendrixMSFT:
In [`azblob`](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob) there are tests in subdirectories. Does this mean the feature is required?

benbp:
> In [`azblob`](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob) there are tests in subdirectories. Does this mean the feature is required?

It's a weird scenario I'm thinking of where you would have tests in a sub-directory, and an asset configuration for those test recordings in the sub-directory, but you invoke the tests from a parent directory (either with `./...` or like `./container/...`). Let me dig back into it.

jhendrixMSFT:
We definitely need `go test ./...` from the `azblob` directory to work.

benbp:
Alright, the latest update is much more flexible with the working directory of go test and the asset config path relative to the recording path, as well as fixing some windows file and root detection issues.

scbedd (approved):
This looks great. Thank you so much @benbp

sdk/internal/recording/recording_test.go
```diff
@@ -550,6 +551,63 @@ func TestHostAndScheme(t *testing.T) {
...
 
+func TestGitRoot(t *testing.T) {
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	_, err = getGitRoot(cwd)
+	require.NoError(t, err)
+}
+
+func TestGitRootNotFound(t *testing.T) {
+	_, err := getGitRoot("../../../../")
```
RickWinter, line 562: I'm not a fan of relative pathing.  Could this be a variable we pass in or something set on the environment?

RickWinter (approved):

sdk/internal/recording/recording.go
```diff
@@ -577,6 +579,105 @@ func getTestId(pathToRecordings string, t *testing.T) string {
 	return path.Join(pathToRecordings, "recordings", t.Name()+".json")
 }
 
+func getGitRoot(fromPath string) (string, error) {
```
jhendrixMSFT, line 582: The better way to do this is to execute the following command.
```
git rev-parse --show-toplevel
```
This will return the repo root.
```
PS C:\git\Azure\azure-sdk-for-go\sdk\resourcemanager\compute\armcompute> git rev-parse --show-toplevel
C:/git/Azure/azure-sdk-for-go
```

benbp: Fair point, if we're ok shelling out to git in the unit tests (I don't see an issue with that).

sdk/internal/recording/recording_test.go
benbp: I'm going to switch to shelling out to git to get the root as Joel suggested below, so I'll be able to remove this test.

sdk/internal/recording/recording.go
jhendrixMSFT: I have no issues with it and have used this in CI in the past.  Looks like we also use this same technique in the [smoketest tool](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/tools/smoketests/cmd/root.go#L176-L180).

benbp: Updated, much simpler now. Thanks!

sdk/internal/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+Support test recording assets external to repository
```
jhendrixMSFT, line 7: We should bump the minor version here and in the [version.go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/version.go#L14) file.

benbp: @jhendrixMSFT should that be done in this PR or as part of a separate release process?

jhendrixMSFT: True we can wait for that.

sdk/internal/recording/recording.go
```diff
@@ -574,7 +575,100 @@ func (r RecordingOptions) baseURL() string {
...
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+	cmd.Dir = absPath
+
+	root, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("Unable to find git root for path '%s'", absPath)
+	}
+
+	// Wrap with Abs() to get os-specific path separators to support sub-path matching
+	return filepath.Abs(strings.TrimSpace(bytes.NewBuffer(root).String()))
```
jhendrixMSFT, line 595: You can cast `root` to a `string` to avoid the buffer.
```go
string(root)
```

sdk/internal/recording/recording.go
```diff
@@ -574,7 +575,100 @@ func (r RecordingOptions) baseURL() string {
...
+	}
+
+	if _, err := os.Stat(gitDirectoryPath); err == nil {
+		return "", nil
+	} else if !errors.Is(err, fs.ErrNotExist) {
+		return "", err
+	}
+
+	trimmedPath := strings.TrimRight(absPath, string(os.PathSeparator))
+	parentDir, _ := filepath.Split(trimmedPath)
```
jhendrixMSFT, line 621: To get the parent dir, I believe you can call `filepath.Dir(absPath)`.

sdk/internal/recording/recording.go
```diff
@@ -574,7 +575,100 @@ func (r RecordingOptions) baseURL() string {
...
+	assetConfigPath := filepath.Join(absPath, recordingAssetConfigName)
+	gitDirectoryPath := filepath.Join(absPath, ".git")
+
+	if _, err := os.Stat(assetConfigPath); err == nil {
+		return assetConfigPath, nil
+	} else if !errors.Is(err, fs.ErrNotExist) {
+		return "", err
+	}
+
+	if _, err := os.Stat(gitDirectoryPath); err == nil {
```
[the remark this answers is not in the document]
benbp: e.g. passing down a stop directory vs. the stop condition? Let me try that out.

sdk/internal/CHANGELOG.md
```diff
@@ -1,14 +1,10 @@
 # Release History
 
-## 1.0.2 (Unreleased)
+## 1.1.0 (2022-10-19)
```
jhendrixMSFT, line 3: This will need to change to today's date.

benbp: @jhendrixMSFT thanks, done

jhendrixMSFT (approved):

PR #19326, opened 2022-10-11
Making the recovery logic for connections simpler by having them only…

richardpark-msft:
Making the recovery logic for connections simpler by having them only be responsible for recovering themselves, not all other links.

Prior to this we'd do a blanket CloseLinks() for the first recovery that happened to reboot the connection. Since each of these recoveries has to happen for each link anyways it's a bit simpler to let them each recover individually. This also preps us for the receiver redirect feature, where we can't assume that all the links are actually part of the same connection.
    
And added a bunch of tests for both parts of this (Namespace.Recover and Links.Recover).

### DIFF
--- a/sdk/messaging/azeventhubs/internal/eh/stress/templates/deploy-job.yaml
+++ b/sdk/messaging/azeventhubs/internal/eh/stress/templates/deploy-job.yaml
@@ -3,6 +3,7 @@
 metadata:
   labels:
     testName: "go-eventhubs"
+    Skip.RemoveTestResources: "true"
 spec:
   # uncomment to deploy to the southeastasia region.
   # nodeSelector:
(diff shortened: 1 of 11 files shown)


sdk/messaging/azeventhubs/internal/rpc.go
```diff
@@ -98,6 +100,17 @@ type RPCLinkArgs struct {
 	LogEvent azlog.Event
 }
 
+func closeOrLog(name string, limit time.Duration, closeable interface {
+	Close(ctx context.Context) error
+}) {
+	ctx, cancel := context.WithTimeout(context.Background(), limit)
+	defer cancel()
+
+	if err := closeable.Close(ctx); err != nil {
```
jhendrixMSFT, line 109: What's the real-world impact here if `Close()` fails?  Does anything leak?

sdk/messaging/azeventhubs/internal/eh/stress/templates/deploy-job.yaml
```diff
@@ -21,30 +22,36 @@ spec:
       - batch
       - "-rounds"
       - "100"
+      - "-verbose"
```
RickWinter, line 25: Can -verbose be a flag that gets set at a higher level and flows everywhere?

sdk/messaging/azeventhubs/internal/rpc.go
richardpark-msft: It's _possible_ something leaks. I had an odd error where EH was complaining that I had the $cbs link opened twice and this was the only case I could think of where that might be.

sdk/messaging/azeventhubs/internal/rpc.go
```diff
@@ -108,6 +121,7 @@ func NewRPCLink(ctx context.Context, args RPCLinkArgs) (*rpcLink, error) {
 
 	linkID, err := uuid.New()
 	if err != nil {
+		closeOrLog("session", 5*time.Second, session)
```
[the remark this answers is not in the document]
richardpark-msft: It might be. I don't even know if it has been timing out, or how long it typically takes which is why I added in the logging.

sdk/messaging/azeventhubs/internal/eh/stress/templates/deploy-job.yaml
richardpark-msft: It could be, but with these args I like it to be as flat and DRY as possible, since it's annoying to have to trace inheritance/fallthroughts and such in templates.

jhendrixMSFT (approved):

PR #19357, opened 2022-10-13
Implement CRC64 hashing with custom storage polynomial

adreed-msft:


sdk/storage/azblob/hashing/hashing_reader.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (h *ReadWrapper) Seek(offset int64, whence int) (int64, error) {
+	return h.internal.Seek(offset, whence)
+}
+
+func (h *ReadWrapper) Close() error {
+	return nil
+}
+
+func NewHashingReadWrapper(r io.ReadSeekCloser, validationOption StorageTransferValidationOption) (*ReadWrapper, error) {
```
jhendrixMSFT, line 31: Since this is in the `hashing` package, you can omit the word "Hash" from the name.

sdk/storage/azblob/hashing/hashing_reader.go
```diff
@@ -0,0 +1,46 @@
+package hashing
```
jhendrixMSFT, line 1: Do we think there's enough content here (or future content) to warrant placing this in its own package?  Maybe it can just live in the `blob` package.  Not advocating for either at present but want to start the discussion.

jhendrixMSFT (changes requested):

sdk/storage/azblob/blockblob/client.go
```diff
@@ -169,6 +171,15 @@ func (bb *Client) StageBlock(ctx context.Context, base64BlockID string, body io.
 
 	opts, leaseAccessConditions, cpkInfo, cpkScopeInfo := options.format()
 
+	if options.TransactionalContentCRC64 == 0 && options.TransactionalValidationOption != hashing.StorageTransferValidationOptionNone {
+		body, err = hashing.NewHashingReadWrapper(body, options.TransactionalValidationOption)
+
+		if options.TransactionalValidationOption&hashing.StorageTransferValidationOptionCRC64 == hashing.StorageTransferValidationOptionCRC64 {
```
vibhansa-msft, line 177: This same check is done inside NewHashingReadWrapper() as well. In future if more options of validation come then this will be duplicate.

sdk/storage/azblob/hashing/hashing_reader.go
adreed-msft: I think it may well have room in the `blob` package. I can't see any particularly good reason why it would be bad to put it there, outside of a dependency loop, but all of the specialized blob packages already reference the core `blob` package.

adreed-msft: As for future substance, it is likely we'll utilize the ability to concatenate hashes together for use in the `chunkwriting.go` file in the `blockblob` package, both for upload and download.

sdk/storage/azblob/blockblob/client.go
adreed-msft: This check is done to prevent writing an empty header.

sdk/storage/azblob/hashing/transfer_validation_option.go
```diff
@@ -0,0 +1,14 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package hashing
+
+type StorageTransferValidationOption uint8
```
jhendrixMSFT, line 9: The pattern is to use the suffix `Type` for constants. Also, can we shorten the name? How about `TransferValidationType`?

jhendrixMSFT: Also, we typically just name the file `constants.go` for ones that contain all constants.

sdk/storage/azblob/blockblob/models.go
```diff
@@ -72,8 +74,12 @@ type StageBlockOptions struct {
 
 	LeaseAccessConditions *blob.LeaseAccessConditions
 
-	// Specify the transactional crc64 for the body, to be validated by the service.
-	TransactionalContentCRC64 []byte
+	// Let the SDK hash for you (providing the hash type(s) specified voids their relevance in this option).
+	// nil = None default
+	TransactionalValidationOption hashing.StorageTransferValidationOption
```
jhendrixMSFT, line 79: Does this change also need to be made in `uploadFromReaderOptions` and `UploadStreamOptions` types?

sdk/storage/azblob/appendblob/client.go
```diff
@@ -140,6 +142,15 @@ func (ab *Client) AppendBlock(ctx context.Context, body io.ReadSeekCloser, o *Ap
 
 	appendOptions, appendPositionAccessConditions, cpkInfo, cpkScope, modifiedAccessConditions, leaseAccessConditions := o.format()
 
+	if o.TransactionalContentCRC64 == 0 && o.TransactionalValidationOption != hashing.StorageTransferValidationOptionNone {
+		body, err = hashing.NewReadWrapper(body, o.TransactionalValidationOption)
```
jhendrixMSFT, line 146: Need to check `err` before using `body`.

sdk/storage/azblob/hashing/hashing_reader.go
jhendrixMSFT: I think it would be best to move this to the `blob` package. That way, we don't have to add more public surface area (e.g. `ReadWrapper`) that's not really required to be public.

sdk/storage/azblob/hashing/transfer_validation_option.go
jhendrixMSFT: Just move to `blob/constants.go` per other discussion.

sdk/storage/azblob/hashing/hashing_reader.go
adreed-msft: > I think it would be best to move this to the `blob` package. That way, we don't have to add more public surface area (e.g. `ReadWrapper`) that's not really required to be public.

ReadWrapper would have to be an exported type anyway as it's used by all 3 specialized blob types, which are separate packages; unless we wanted to incur code dupe.

adreed-msft: As per offline discussion, hashing type option has been moved to `internal\exported` and the rest to `internal\shared`.

sdk/storage/azblob/appendblob/models.go
```diff
@@ -73,8 +74,12 @@ func (o *CreateOptions) format() (*generated.AppendBlobClientCreateOptions, *gen
...
 // AppendBlockOptions contains the optional parameters for the Client.AppendBlock method.
 type AppendBlockOptions struct {
-	// Specify the transactional crc64 for the body, to be validated by the service.
-	TransactionalContentCRC64 []byte
+	// Let the SDK hash for you (providing the hash type(s) specified voids their relevance in this option).
+	// nil = None default
+	TransactionalValidationOption exported.TransferValidationType
+
+	// Specify the transactional crc64 for the body, to be validated by the service. Should be hashed using hashing.CRC64Table or hashing.CRC64Polynomial
+	TransactionalContentCRC64 uint64
```
jhendrixMSFT, line 82: If the SDK performs the hashing based on the value of the above field, does this field still need to exist?

sdk/storage/azblob/appendblob/models.go
```diff
@@ -73,8 +74,12 @@ func (o *CreateOptions) format() (*generated.AppendBlobClientCreateOptions, *gen
 
 // AppendBlockOptions contains the optional parameters for the Client.AppendBlock method.
 type AppendBlockOptions struct {
-	// Specify the transactional crc64 for the body, to be validated by the service.
-	TransactionalContentCRC64 []byte
+	// Let the SDK hash for you (providing the hash type(s) specified voids their relevance in this option).
+	// nil = None default
+	TransactionalValidationOption exported.TransferValidationType
```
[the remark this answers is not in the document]
jhendrixMSFT: Also, I think you can shorten the field name to just `TransactionalValidation`.  What do you think?

sdk/storage/azblob/appendblob/models.go
adreed-msft: These are for the user supplying their own hash. On a technicality, they could still be used to validate hashing functionality in testing, or a user could have their own pipeline where they've already hashed the incoming data and don't want to re-hash it.

jhendrixMSFT: Does this mean that the two options are mutually exclusive?

sdk/storage/azblob/appendblob/models.go
```diff
@@ -73,8 +74,12 @@ func (o *CreateOptions) format() (*generated.AppendBlobClientCreateOptions, *gen
 
 // AppendBlockOptions contains the optional parameters for the Client.AppendBlock method.
 type AppendBlockOptions struct {
-	// Specify the transactional crc64 for the body, to be validated by the service.
-	TransactionalContentCRC64 []byte
+	// Let the SDK hash for you (providing the hash type(s) specified voids their relevance in this option).
+	// nil = None default
+	TransactionalValidationOption blob.TransferValidationType
```
jhendrixMSFT, line 79: What do you think about shortening this to `TransactionalValidation`?

adreed-msft: Sounds like a good change to make

sdk/storage/azblob/appendblob/models.go
```diff
@@ -73,8 +74,12 @@ func (o *CreateOptions) format() (*generated.AppendBlobClientCreateOptions, *gen
 
 // AppendBlockOptions contains the optional parameters for the Client.AppendBlock method.
 type AppendBlockOptions struct {
-	// Specify the transactional crc64 for the body, to be validated by the service.
-	TransactionalContentCRC64 []byte
+	// Let the SDK hash for you (providing the hash type(s) specified voids their relevance in this option).
+	// nil = None default
+	TransactionalValidation blob.TransferValidationType
+
+	// Specify the transactional crc64 for the body, to be validated by the service. Should be hashed using hashing.CRC64Table or hashing.CRC64Polynomial
```
jhendrixMSFT, line 81: This comment needs to be updated after the package refactor (I didn't check any others).

sdk/storage/azblob/internal/exported/hashing_polynomial.go
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package exported
+
+import (
+	"hash/crc64"
+)
+
+const CRC64Polynomial uint64 = 0x9A6C9329AC4BC9B5
```
jhendrixMSFT, line 13: Looks like this, and `CRC64Table`, don't actually need to be exported from this package.

Also, it might be good to condense all of this content into one file as it's all related.

adreed-msft: I am going to condense it to one file, but I intend to leave them exported, as they are used in documentation for the package.

sdk/storage/azblob/appendblob/models.go
adreed-msft: They are.

sdk/storage/azblob/internal/exported/hashing_reader.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func NewReadWrapper(r io.ReadSeekCloser, validationOption TransferValidationType) (*ReadWrapper, error) {
+	buf, err := io.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	var crc uint64
+	if validationOption&TransferValidationTypeCRC64 == TransferValidationTypeCRC64 {
```
jhendrixMSFT, line 48: What happens if this is `false`?  Should it return an error?

adreed-msft: Absolutely nothing. The hash just doesn't get generated, and the read wrapper works as expected.

The goal of the wrapper was, per Jocelyn's request, to make it so that we didn't have to do a re-read for a hash, and then a re-read for submission.

Should the service adjust to accept HTTP footers, this read wrapper will adjust to generate the hash during the read and then have it be submitted afterwards.

jhendrixMSFT (approved):

siminsavani-msft (approved):

vibhansa-msft (approved):

jhendrixMSFT:
From the [docs](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties), it seems like the service returning the MD5 is expected.  But why did this start happening?

PR #19361, opened 2022-10-14
Version Level WORM Support

tasherif-msft:


vibhansa-msft (approved):

souravgupta-msft:
This should be added for page blob and append blob as well, right?

sdk/storage/azblob/blob/client_test.go
```diff
@@ -3012,6 +3012,86 @@ func (s *BlobRecordedTestsSuite) TestBlobSetMetadataIfNoneMatchFalse() {
...
+	policy := blob.ImmutabilityPolicySetting(blob.ImmutabilityPolicySettingLocked)
+	_require.Nil(err)
+
+	setImmutabilityPolicyOptions := blob.SetImmutabilityPolicyOptions{
+		ExpiryTime:               &currentTime,
+		Mode:                     &policy,
+		ModifiedAccessConditions: nil,
+	}
+	_, err = bbClient.SetImmutabilityPolicy(context.Background(), &setImmutabilityPolicyOptions)
+	_require.Nil(err)
```
siminsavani-msft, line 3040: It would be good to test if the ImmutabilityPolicy works in the test by attempting to delete the blob (which should fail).

sdk/storage/azblob/blob/client_test.go
```diff
@@ -3012,6 +3012,86 @@ func (s *BlobRecordedTestsSuite) TestBlobSetMetadataIfNoneMatchFalse() {
...
+	containerClient := testcommon.GetContainerClient(containerName, svcClient)
+
+	blockBlobName := testcommon.GenerateBlobName(testName)
+	bbClient := testcommon.CreateNewBlockBlob(context.Background(), _require, blockBlobName, containerClient)
+
+	_, err = bbClient.GetProperties(context.Background(), nil)
+	_require.Nil(err)
+
+	_, err = bbClient.SetLegalHold(context.Background(), true, nil)
+	_require.Nil(err)
```
siminsavani-msft, line 3092: Test if LegalHold works by attempting to delete blob.

sdk/storage/azblob/blob/client_test.go
```diff
@@ -3012,6 +3012,86 @@ func (s *BlobRecordedTestsSuite) TestBlobSetMetadataIfNoneMatchFalse() {
...
+
+	setImmutabilityPolicyOptions := blob.SetImmutabilityPolicyOptions{
+		ExpiryTime:               &currentTime,
+		Mode:                     &policy,
+		ModifiedAccessConditions: nil,
+	}
+	_, err = bbClient.SetImmutabilityPolicy(context.Background(), &setImmutabilityPolicyOptions)
+	_require.Nil(err)
+
```
siminsavani-msft, line 3070: Test if ImmutabilityPolicy is set by attempting to delete blob + test after deleting the ImmutabilityPolicy.

sdk/storage/azblob/appendblob/client.go
```diff
@@ -190,6 +190,24 @@ func (ab *Client) Undelete(ctx context.Context, o *blob.UndeleteOptions) (blob.U
 	return ab.BlobClient().Undelete(ctx, o)
 }
 
+// SetImmutabilityPolicy operation enables users to set the immutability policy on a blob.
+// https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-storage-overview
+func (ab *Client) SetImmutabilityPolicy(ctx context.Context, options *blob.SetImmutabilityPolicyOptions) (blob.SetImmutabilityPolicyResponse, error) {
```
souravgupta-msft, line 195: Add tests for append and page blobs.

sdk/storage/azblob/blob/client.go
```diff
@@ -231,6 +231,30 @@ func (b *Client) GetTags(ctx context.Context, options *GetTagsOptions) (GetTagsR
 
 }
 
+// SetImmutabilityPolicy operation enables users to set the immutability policy on a blob.
+// https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-storage-overview
+func (b *Client) SetImmutabilityPolicy(ctx context.Context, options *SetImmutabilityPolicyOptions) (SetImmutabilityPolicyResponse, error) {
```
jhendrixMSFT, line 236: Having all the params in `options` implies that they're all optional. Is this really the case? What happens if a customer calls `SetImmutabilityPolicy(ctx, nil)`? If this is legal, then the default behavior needs to be documented.

sdk/storage/azblob/blockblob/client.go
```diff
@@ -11,10 +11,6 @@ import (
 	"context"
 	"encoding/base64"
 	"errors"
-	"io"
```
jhendrixMSFT, line 14: Why did this ordering change?

jhendrixMSFT (changes requested):
Also needs entry in the changelog.

sdk/storage/azblob/blob/client.go
tasherif-msft: So, the mode defaults to unlocked (can be nil). However, we do need to provide an expiry time. So, I changed that one parameter in the struct from optional to required.

sdk/storage/azblob/blockblob/client.go
tasherif-msft: goland automatically enforces sorting

siminsavani-msft (approved):
LGTM! Just make sure the live tests pass in case there are any conflicts.

sdk/storage/azblob/blob/client.go
```diff
@@ -242,6 +242,30 @@ func (b *Client) GetTags(ctx context.Context, options *GetTagsOptions) (GetTagsR
 
 }
 
+// SetImmutabilityPolicy operation enables users to set the immutability policy on a blob. Mode defaults to "Unlocked".
+// https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-storage-overview
+func (b *Client) SetImmutabilityPolicy(ctx context.Context, options SetImmutabilityPolicyOptions) (SetImmutabilityPolicyResponse, error) {
```
jhendrixMSFT, line 247: `*SetImmutabilityPolicyOptions`

jhendrixMSFT: Comment applies to the other APIs.

jhendrixMSFT: Well wait, that means that everything is optional.  What would that mean to call this API with a `nil` options then?

tasherif-msft: I did not set it as optional since an ExpiryTime must be passed at minimum. Other fields can be nil in the SetImmutabilityPolicyOptions bag

jhendrixMSFT: I see.  The convention is that all mandatory parameters must be individually specified.  All optional params are in the `Options` type which is passed by pointer.

tasherif-msft: changed!

PR #19362, opened 2022-10-14
Add support for ResourceGUID property

rayterrill:
https://github.com/hashicorp/terraform-provider-azurerm/issues/17011

   - These files are managed by the codegen framework at [Azure/autorest.go][x].

### DIFF
--- a/services/network/mgmt/2021-08-01/network/models.go
+++ b/services/network/mgmt/2021-08-01/network/models.go
@@ -9,12 +9,13 @@ package network
 import (
 	"context"
 	"encoding/json"
+	"net/http"
+
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/date"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/Azure/go-autorest/tracing"
-	"net/http"
 )
 
 // The package's fully qualified name.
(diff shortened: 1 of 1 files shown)


tadelesh:
@rayterrill Our SDK is auto generated by service swagger definition. According to the [spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/stable/2021-08-01/privateEndpoint.json#L678), `resourceGuid` is not defined. Meanwhile, you are trying to change our previous SDK which has already been deprecated. After the service team fix the swagger, we could only update our new SDK which module path is `github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork`.

rayterrill:
@tadelesh is there an issue to get the swagger updated so the SDK can be regenerated with resourceGuid?

tadelesh:
> @tadelesh Chenjie Shi FTE is there an issue to get the swagger updated so the SDK can be regenerated with resourceGuid?

It seems not currently. Could you help to raise an issue in the swagger [repo](https://github.com/Azure/azure-rest-api-specs/issues)?

rayterrill:
@tadelesh Looks like there is at least one already, but it's been open since May: https://github.com/Azure/azure-rest-api-specs/issues/19111. Potentially this one as well: https://github.com/Azure/azure-rest-api-specs/issues/16693.

rayterrill:
@tadelesh we ended up using the azurerm terraform provider instead - it's a bummer that the main terraform-provider-azurerm uses the deprecated SDK (https://github.com/hashicorp/terraform-provider-azurerm/blob/c4ff420ccd94b60ce9b99be9ab5a67a35adc9fd0/internal/services/network/private_endpoint_resource.go#L11) - that's what let me down this path.

tadelesh:
I understand. The reason that terraform does not migrate to new SDK immediately is a long story. We are still discussing about it continuously. For this issue, the first thing we could do is to fix the swagger. As said in this [issue](https://github.com/Azure/azure-rest-api-specs/issues/19111#issuecomment-1282272934), we have already pushed the service team to look into it. I'll follow up with it.

Issue #18688, opened 2022-07-26
x-ms-version is not changed when using a different storage profile for Azure Stack Hub


ephrain1021:
### Bug Report

- import path of package in question: github.com/Azure/azure-sdk-for-go/profiles/2020-09-01/storage/mgmt/storage
- SDK version: latest
- output of `go version`:  go version go1.17.5 darwin/amd64

- What happened? 

I am trying to use azure-sdk-for-go/tree/main/sdk/storage/azblob to access container of storage account on Azure Stack Hub.
However, that does not work and returns following error:
```
===== RESPONSE ERROR (ErrorCode=InvalidHeaderValue) =====
Description=The value for one of the HTTP headers is not in the correct format.
RequestId:21b4164c-292b-09f6-4898-58e9e9552a82
Time:2022-07-01T03:35:49.7989476Z, Details: 
HeaderName: x-ms-version
HeaderValue: 2019-12-12
```

After digging into the source code of azblob, it looks like x-ms-version is hardcoded as 2019-12-12.
```
req.Raw().Header.Set("x-ms-version", "2019-12-12")
```

If I try to change the value 2019-12-12 to 2019-07-07, my testing program can access the container of storage account on Azure Stack Hub successfully.
However, this approach might not be a good solution since we patch the azblob sdk manually.

So, I followed another suggestion to use "profiles" that had been illustrated in the sample of https://github.com/Azure-Samples/Hybrid-Golang-Samples/tree/master/dataplane.
This sample used "profile", and an older (deprecated) version of azblob, azure-storage-blob-go/azblob.
The default profile used in the sample was "github.com/Azure/azure-sdk-for-go/profiles/2020-09-01/storage/mgmt/storage".
This sample was able to access the container of storage account on Azure Stack Hub.
However, I noticed that changing the "profile" only changed the api-version, not the x-ms-version, in the HTTP request.

1. profile 2017-03-09 => api-version 2016-01-01, x-ms-version 2019-02-02
2. profile 2018-03-01 => api-version 2016-01-01, x-ms-version 2019-02-02
3. profile 2019-03-01 => api-version 2017-10-01, x-ms-version 2019-02-02
4. profile 2020-09-01 => api-version 2017-10-01, x-ms-version 2019-02-02

- What did you expect or want to happen?

I expect that x-ms-version should be changed to align with the profile.
It should be natural to use an older x-ms-version with an older profile for storage account, right?

- How can we reproduce it?

[...]

PR #19363, opened 2022-10-14
Enable users to override client default API versions

chlowell:
...with code like this:
```go
armsubscription.NewSubscriptionsClient(cred, &arm.ClientOptions{
		APIVersion: "42",
	}
})
```
To get this API for ARM clients, users can simply upgrade `azcore`--no code gen required. For data plane clients, we'll need to release new versions that tell the new pipeline policy which header/query param to set. It's therefore possible for a user to upgrade `azcore` and set the new option for a data plane client that won't respect it. In this case I have the policy's `Do()` return an error (a constructor error isn't possible). That's bad, but I like it better than silently ignoring configuration. Let me know if you see a better alternative.

### DIFF
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -27,7 +27,7 @@ type Request = exported.Request
 // ClientOptions contains optional settings for a client's pipeline.
 // All zero-value fields will be initialized with default values.
 type ClientOptions struct {
-	// APIVersion overrides the client's default service version.
+	// APIVersion overrides the default version requested of the service. Set with caution as this package version has not been tested with arbitrary service versions.
 	APIVersion string
 
 	// Cloud specifies a cloud for the client. The default is Azure Public Cloud.
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -27,6 +27,9 @@ type Request = exported.Request
 // ClientOptions contains optional settings for a client's pipeline.
 // All zero-value fields will be initialized with default values.
 type ClientOptions struct {
+	// APIVersion overrides the client's default service version.
+	APIVersion string
+
 	// Cloud specifies a cloud for the client. The default is Azure Public Cloud.
 	Cloud cloud.Configuration
 
--- a/sdk/azcore/runtime/policy_api_version.go
+++ b/sdk/azcore/runtime/policy_api_version.go
@@ -1,0 +1,80 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package runtime
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+const (
+	apiVersionHeaderName = iota
+	apiVersionQueryParamName
+)
+
+// APIVersionName is the name of an API version query parameter or header.
+type APIVersionName interface {
+	fmt.Stringer
+	kind() int
+}
+
+// APIVersionHeaderName names the header a client should set with a request's API version.
+type APIVersionHeaderName string
+
+func (h APIVersionHeaderName) kind() int {
+	return apiVersionHeaderName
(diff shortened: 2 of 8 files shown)


sdk/azcore/runtime/policy_api_version.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+const (
+	apiVersionHeaderName = iota
+	apiVersionQueryParamName
+)
+
+// APIVersionName is the name of an API version query parameter or header.
```
JeffreyRichter, line 22: I think you could simplify this a lot. How about this:

type APIVersionLocation int

const (
	APIVersionLocationQueryParam = iota
	APIVersionLocationHeader     = 1
)

// NewAPIVersionPolicy constructs an APIVersionPolicy. If version is "", Do will be a no-op.
func NewAPIVersionPolicy(location APIVersionLocation, name string, version string) *APIVersionPolicy {
	return &APIVersionPolicy{location: location, name: name, version: version}
}

// APIVersionPolicy enables users to set the API version of every request a client sends.
type APIVersionPolicy struct {
	// location indicages if 'name' refers to a query parameter or header.
	location APIVersionLocation

	// name of the query param or header whose value should be overridden; provided by the client.
	name string

	// version is the value (provided by the user) that replaces the default version value.
	version string
}

// Do sets the request's API version, if the policy is configured to do so, replacing any prior value.
func (a *APIVersionPolicy) Do(req *policy.Request) (*http.Response, error) {
	if a.version != "" {
		if a.name == "" {
			// user set ClientOptions.APIVersion but the client ctor didn't set PipelineOptions.APIVersionName
			return nil, errors.New("this client doesn't support overriding its API version")
		}
		switch a.location {
		case APIVersionLocationHeader:
			req.Raw().Header.Set(a.name, a.version)
		case APIVersionLocationQueryParam:
			q := req.Raw().URL.Query()
			q.Set(a.name, a.version)
			req.Raw().URL.RawQuery = q.Encode()
		}
	}
	return req.Next()
}

sdk/azcore/policy/policy.go
```diff
@@ -27,6 +27,9 @@ type Request = exported.Request
 // ClientOptions contains optional settings for a client's pipeline.
 // All zero-value fields will be initialized with default values.
 type ClientOptions struct {
+	// APIVersion overrides the client's default service version.
```
JeffreyRichter, line 30:
```suggestion
	// APIVersion overrides the default version requested of the service. Set with caution as this package version has not been tested with arbitrary service versions.
```

sdk/azcore/runtime/policy_api_version.go
chlowell: I like it. The downside is that it `PipelineOptions` then needs two fields, one for the location and another for the name, and we won't be able to tell whether the caller set the location because its zero value will be useable. I suppose that's okay for an API intended for SDK developers.

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/784

sdk/azcore/runtime/policy_api_version.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Do sets the request's API version, if the policy is configured to do so, replacing any prior value.
+func (a *APIVersionPolicy) Do(req *policy.Request) (*http.Response, error) {
+	if a.version != "" {
+		if a.name == "" {
+			// user set ClientOptions.APIVersion but the client ctor didn't set PipelineOptions.APIVersionLocation
+			return nil, errors.New("this client doesn't support overriding its API version")
+		}
+		switch a.location {
+		case APIVersionLocationHeader:
```
jhendrixMSFT, line 53: Can you add a `default` clause in the unlikely event we get an unknown value?

jhendrixMSFT:
Also needs an entry in the changelog.

jhendrixMSFT:
This should also address https://github.com/Azure/azure-sdk-for-go/issues/18688

JeffreyRichter (approved):

jhendrixMSFT (approved):

PR #19381, opened 2022-10-18
Add service client accessor to azblob.Client

jhendrixMSFT:

### DIFF
--- a/sdk/storage/azblob/CHANGELOG.md
+++ b/sdk/storage/azblob/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features Added
 
 * Added [Set Blob Expiry API](https://learn.microsoft.com/rest/api/storageservices/set-blob-expiry).
+* Added method `ServiceClient()` to the `azblob.Client` type, allowing access to the underlying service client.
 
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
I know we had talked about naming this `NewServiceClient()` however I didn't for the following reasons.
- We have this same pattern for the specialized clients ([example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/appendblob/client.go#L91-L92))
- The method doesn't allocate a new client which is important to communicate (the other `New*Client()` ctors do)

JeffreyRichter:
I see you point. I can live with this but I do have a thought: We have a pattern all over the SDK of getting clients from clients via NewXxxClient methods. It feels to me that this method might be leaking an implementaiton detail in that it just happens to use a ServiceClient internally and perhaps someday it won't (unlikely I admit). Also, we could have this method allocate a new ServiceCelient and perhaps we should in case the design of clients every because mutable (also very unlikey). My point though, is that there are a bunch of assumptions built into the method name and it could tie our hands in the future.

vibhansa-msft (approved):

souravgupta-msft (approved):

siminsavani-msft (approved):
An example using ServiceClient would be great!

PR #19389, opened 2022-10-19
IMDS timeout increased to 10 seconds

realDeveloper45:
In case of stressed servers 1000ms is very aggressive time. This PR is to just trying to get some attention in to this intermittent context deadline issue. Error Im facing is _ManagedIdentityCredential: IMDS token request timed out_ error for long running jobs that makes many requests to azure secrets fetch data.

### DIFF
--- a/sdk/azidentity/default_azure_credential.go
+++ b/sdk/azidentity/default_azure_credential.go
@@ -67,7 +67,7 @@ func NewDefaultAzureCredential(options *DefaultAzureCredentialOptions) (*Default
 	msiCred, err := NewManagedIdentityCredential(o)
 	if err == nil {
 		creds = append(creds, msiCred)
-		msiCred.mic.imdsTimeout = time.Second
+		msiCred.mic.imdsTimeout =10 * time.Second
 	} else {
 		errorMessages = append(errorMessages, credNameManagedIdentity+": "+err.Error())
 		creds = append(creds, &defaultCredentialErrorReporter{credType: credNameManagedIdentity, err: err})


chlowell (changes requested):
Thanks, this is timely! We (Azure SDK authentication team) recently began discussing ways to improve IMDS timeout behavior for DefaultAzureCredential. Simply lengthening the timeout is one of the options but would regress performance in some scenarios, and I don't want to make any changes here until we've explored other options. Note that only DefaultAzureCredential imposes this timeout. If it's a problem in your application, you can avoid it by using [ManagedIdentityCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ManagedIdentityCredential) directly or with [ChainedTokenCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#ChainedTokenCredential).

To give some context on the timeout, we added it to limit the worst case running time of DefaultAzureCredential's first auth attempt. This is important when IMDS isn't available, for example in local development. The only way to learn whether IMDS is available is to try connecting to it, which can take longer than 10 seconds to time out on some platforms.

shashanksbs:
@chlowell @jhendrixMSFT @kenegozi 
 Pod Identity can take up to a couple minutes to get ready for token requests from new pods. In case of clients that have shorter timeouts, the retries can be terminated and the client will not receive a token in the first attempt. This feature flag can use to prevent timeouts due to that delay. we added https://azure.github.io/aad-pod-identity/docs/configure/feature_flags/#set-retry-after-header-in-nmi-response this check in azure pod identity helm chart.

But IMDS timeout issue still persists and sometimes we are getting _no default identity is assigned to this resource_ error.

these errors occur mostly on stressed servers, on normal day it works fine

Please suggest any resolution for this

jhendrixMSFT:
The short timeout that you're changing here is only part of `DefaultAzureCredential`.  If you use directly use `ManagedIdentityCredential` then there is no timeout (outside of how the default HTTP client works).

PR #19391, opened 2022-10-19
Set blob expiry

souravgupta-msft:
Added code for set blob expiry


sdk/storage/azblob/appendblob/client.go
```diff
@@ -200,6 +200,12 @@ func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.Set
 	return ab.BlobClient().SetTier(ctx, tier, o)
 }
 
+// SetExpiry operation sets an expiry time on an existing blob. This operation is only allowed on Hierarchical Namespace enabled accounts.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-expiry
+func (ab *Client) SetExpiry(ctx context.Context, expiryOptions blob.ExpiryOptions, o *blob.SetExpiryOptions) (blob.SetExpiryResponse, error) {
```
jhendrixMSFT, line 205: The `blob.ExpiryOptions` type isn't following the design guidelines. As a result, this API sig looks weird (i.e. why two options types). Can you please rename `blob.ExpiryOptions` to `blob.ExpiryType` and the const values to match?

jhendrixMSFT: Be sure to note the breaking change in the changelog.

sdk/storage/azblob/appendblob/examples_test.go
```diff
@@ -67,3 +70,32 @@ func Example_appendblob_Client() {
...
+	blobName := "test_append_blob_set_expiry.txt"
+	blobURL := fmt.Sprintf("https://%s.blob.core.windows.net/testcontainer/%s", accountName, blobName)
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	handleError(err)
+
+	appendBlobClient, err := appendblob.NewClient(blobURL, cred, nil)
+	handleError(err)
+
+	// set expiry on append blob to an absolute time which must be in RFC 1123 Format
```
jhendrixMSFT, line 90: The wire format is an implementation detail, we shouldn't make customers do it. Can you please update the API to properly format the time?

jhendrixMSFT (changes requested):
Rename constant and clean up time formatting.

sdk/storage/azblob/appendblob/client.go
souravgupta-msft: Done

sdk/storage/azblob/appendblob/examples_test.go
jhendrixMSFT: Given the polymorphic nature of the expressed value, I think we should use an interface instead of the const/string pairs.  Here's an example.
```go
type ExpiryType interface {
	format() (generated.ExpiryOptions, generated.BlobClientSetExpiryOptions)
}

type ExpiryTypeAbsolute time.Time

func (e ExpiryTypeAbsolute) format() (generated.ExpiryOptions, generated.BlobClientSetExpiryOptions) {
	return generated.ExpiryOptionsAbsolute, generated.BlobClientSetExpiryOptions{
		ExpiresOn: to.Ptr(time.Time(e).UTC().Format(http.TimeFormat)),
	}
}

type ExpiryTypeRelativeToNow time.Duration

func (e ExpiryTypeRelativeToNow) format() (generated.ExpiryOptions, generated.BlobClientSetExpiryOptions) {
	return generated.ExpiryOptionsRelativeToNow, generated.BlobClientSetExpiryOptions{
		ExpiresOn: to.Ptr(strconv.FormatInt(time.Duration(e).Milliseconds(), 10)),
	}
}
```
This lets callers naturally specify the desired time, and the conversion is performed by the underlying types.
```go
func (b *Client) SetExpiry(ctx context.Context, expiryType ExpiryType, o *SetExpiryOptions) (SetExpiryResponse, error) {
	opt, val := expiryType.format()
	resp, err := b.generated().SetExpiry(ctx, opt, &val)
	return resp, err
}
...
blobClient.SetExpiry(context.TODO(), blob.ExpiryTypeAbsolute(time.Now().Add(12*time.Hour)), nil)

blobClient.SetExpiry(context.TODO(), blob.ExpiryTypeRelativeToNow(10*time.Second), nil)
```

souravgupta-msft: Made the changes. Please review.

sdk/storage/azblob/blob/constants.go
```diff
@@ -171,12 +172,24 @@ func PossibleDeleteTypeValues() []DeleteType {
...
+type ExpiryTypeAbsolute time.Time
+
+// ExpiryTypeRelativeToNow defines the duration relative to now for the blob expiry
+type ExpiryTypeRelativeToNow time.Duration
+
+// ExpiryTypeRelativeToCreation defines the duration relative to creation for the blob expiry
+type ExpiryTypeRelativeToCreation time.Duration
+
+// ExpiryTypeNeverExpire defines that the blob will be set to never expire
+type ExpiryTypeNeverExpire struct {
```
jhendrixMSFT, line 190: Is this actually required?  The default is never expire, correct?

sdk/storage/azblob/appendblob/client.go
```diff
@@ -200,6 +200,12 @@ func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.Set
 	return ab.BlobClient().SetTier(ctx, tier, o)
 }
 
+// SetExpiry operation sets an expiry time on an existing blob. This operation is only allowed on Hierarchical Namespace enabled accounts.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-expiry
+func (ab *Client) SetExpiry(ctx context.Context, expiryType blob.ExpiryType, o *blob.SetExpiryOptions) (blob.SetExpiryResponse, error) {
```
jhendrixMSFT, line 205: Does the page blob client need this too?  I don't see it being added.

sdk/storage/azblob/blob/constants.go
souravgupta-msft: Removed this and made never expire as the default expiry type

sdk/storage/azblob/blob/models.go
```diff
@@ -244,6 +247,35 @@ func (o *SetTierOptions) format() (*generated.BlobClientSetTierOptions, *generat
 
 // ---------------------------------------------------------------------------------------------------------------------
 
+// SetExpiryOptions contains the optional parameters for the Client.SetExpiry method.
+type SetExpiryOptions struct {
+	// placeholder for future options
+}
+
+func (e ExpiryTypeAbsolute) format(o *SetExpiryOptions) (generated.ExpiryOptions, *generated.BlobClientSetExpiryOptions) {
```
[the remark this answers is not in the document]
souravgupta-msft: The `format()` method returns `*generated.BlobClientSetExpiryOptions` as well. In future if more options are added to it, they would come as input to the `format()` method through `o *SetExpiryOptions` parameter. This is the reason why I kept it as a parameter. Let me know your thoughts.

sdk/storage/azblob/appendblob/client.go
souravgupta-msft: Set expiry operation is only allowed on HNS enabled accounts.
HNS enabled accounts does not support page blobs. This is why I haven't added this to the page blob client.

sdk/storage/azblob/blob/client.go
```diff
@@ -158,6 +158,20 @@ func (b *Client) SetTier(ctx context.Context, tier AccessTier, o *SetTierOptions
...
 
+// SetExpiry operation sets an expiry time on an existing blob. This operation is only allowed on Hierarchical Namespace enabled accounts.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-expiry
+func (b *Client) SetExpiry(ctx context.Context, expiryType ExpiryType, o *SetExpiryOptions) (SetExpiryResponse, error) {
+	var et generated.ExpiryOptions
+	var opts *generated.BlobClientSetExpiryOptions
+	if expiryType != nil {
+		et, opts = expiryType.format(o)
+	} else {
+		et, opts = generated.ExpiryOptionsNeverExpire, nil
```
jhendrixMSFT, line 169: Do we expect customers to call this with a `nil` value though?  If one creates a blob, isn't the default that it never expires?

souravgupta-msft: Yes, the expiry type is never expire when a blob is created. But let's say that if an expiry type is set on a blob, then never expire option can be used to unset the expiry. This option has to be passed explicitly to the [SetExpiry](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-expiry#expiryoption) API.

I can think of two options here. First having `NeverExpire` as an empty struct which will implement the `ExpiryType` interface, similar to other expiry types.
```
type ExpiryTypeNeverExpire struct {
}

// implementation
func (e ExpiryTypeNeverExpire) format(o *SetExpiryOptions) (generated.ExpiryOptions, *generated.BlobClientSetExpiryOptions) {
	return generated.ExpiryOptionsNeverExpire, nil
}

// calling this
bbClient.SetExpiry(context.Background(), blob.ExpiryTypeNeverExpire{}, nil)
```

Second option is what we have currently that is having `NeverExpire` as the default expiry option.

jhendrixMSFT: I understand now.  I think your original solution of `ExpiryTypeNeverExpire` is the best (sorry for the hassle).

souravgupta-msft: Made the changes.
Please note that `nil` can still be passed as `ExpiryType` since it is an interface. In this case, I am setting `ExpiryType` to `ExpiryTypeNeverExpire` as a default option.

sdk/storage/azblob/blob/constants.go
jhendrixMSFT, line 190: You can remove the "Expire" suffix as it already starts with the word "Expiry".

siminsavani-msft (approved):

tasherif-msft (approved):

sdk/storage/azblob/blob/constants.go
souravgupta-msft: Done

vibhansa-msft (approved):

jhendrixMSFT (approved):

Issue #19192, opened 2022-09-23
default http client has way too low MaxIdleConnsPerHost


weinong:
### Bug Report

- What happened?
  - the [default http client](https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/runtime/transport_default_http_client.go#L26) that azure sdk uses has `MaxIdleConns: 100`. However, as it doesn't change `MaxIdleConnsPerHost`, the default is 2. This means for 100 connections it opens, it also closes 98 of them. [ref](https://github.com/OJ/gobuster/issues/127) and [ref](https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/274). It's particularly bad on Azure due to the infamous SNAT port exhaustion.
- What did you expect or want to happen?
  - change `MaxIdleConnsPerHost` to `100` as well

jhendrixMSFT:
I assume that on the AKS side there was some testing to arrive at this number.  Do you have any details on that?  Given that this would change the in-box default, I'd like to make sure there aren't any unwanted side effects.

weinong:
the reason I raised this issue was because I was fixing a connection leak bug and found some references (linked in the issue) that suggests to update the default `MaxIdleConnsPerHost`. The reasoning makes sense.
But I did some digging today and found I didn't see any change regardless `MaxIdleConnsPerHost` is set or not. Upon digging further, I found it only happened in HTTP 1.1, but not in HTTP/2. Since the SDK defaults to HTTP/2 with Azure Resource Manager, I think we can close it.

PR #19408, opened 2022-10-24
set MaxIdleConnsPerHost to 100 instead of 2 in defaultTransport

MartinForReal:

### DIFF
--- a/sdk/azcore/runtime/transport_default_http_client.go
+++ b/sdk/azcore/runtime/transport_default_http_client.go
@@ -22,8 +22,13 @@ func init() {
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
 		}).DialContext,
-		ForceAttemptHTTP2:     true,
-		MaxIdleConns:          100,
+		ForceAttemptHTTP2: true,
+		MaxIdleConns:      100,
+		//https://pkg.go.dev/net/http#Transport
+		// MaxIdleConnsPerHost, if non-zero, controls the maximum idle
+		// (keep-alive) connections to keep per-host. If zero,
+		// DefaultMaxIdleConnsPerHost(2) is used.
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
According to https://github.com/Azure/azure-sdk-for-go/issues/19192#issuecomment-1256848185, it was determined that this isn't required.  Can you please clarify?

MartinForReal:
From the code, https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/http/transport.go;l=975;drc=a1901f898bc05aac966edd247ff122f52fbb8d2e
this config will also be applied to http2 connections. and http2 connections can be shared among go routines but it also has limits

sdk/azcore/runtime/transport_default_http_client.go
```diff
@@ -22,8 +22,13 @@ func init() {
...
 		}).DialContext,
-		ForceAttemptHTTP2:     true,
-		MaxIdleConns:          100,
+		ForceAttemptHTTP2: true,
+		MaxIdleConns:      100,
+		//https://pkg.go.dev/net/http#Transport
+		// MaxIdleConnsPerHost, if non-zero, controls the maximum idle
+		// (keep-alive) connections to keep per-host. If zero,
+		// DefaultMaxIdleConnsPerHost(2) is used.
+		MaxIdleConnsPerHost:   100,
```
jhendrixMSFT, line 31: Do you think this is a reasonable default for _all_ clients (i.e. there are no negative side effects)?

You can create your own client transport and specify it [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/policy/policy.go#L46) when constructing clients.

MartinForReal: I think it's reasonable because we found this config will have huge impact on blob client when there are a lot of connections.

Issue #19280, opened 2022-10-04
Distributed Tracing


RickWinter:
Add distributed tracing support to Core
```[tasklist]
- [x] Azure.Core abstraction layer
- [x] CodeGen to light up feature
- [x] OTel adapter
- [ ] Beta Release
- [ ] Stable Release
```

richardpark-msft:
Looking at this from the lens of Service Bus we're missing some features (or it could be that they're not 1:1 with the Java/.NET layers)

- `Links`, which we can use to link multiple spans together, don't exist (yet). This is different than the parent/child relationship between spans. From @lmolkova's gist: [`builder.addLink`](https://gist.github.com/lmolkova/e4215c0f44a49ef824983382762e6b92#:~:text=EventData%20or%20EventDataBatch-,builder.addLink((SpanContext)msg.SpanContext)%3B,-%7D%20else%20%7B)

- Getting a SpanContext, which has fields like the TraceID, SpanID and trace options (unsure of this last one). [getDiagnosticId](https://gist.github.com/lmolkova/e4215c0f44a49ef824983382762e6b92#:~:text=spanContext.getTraceId().copyLowerBase16To(chars%2C%203)%3B)

Issue #7358, opened 2020-02-18
Azure-Core: Add OpenTelemetry Tracing Plugin Support for Track2 Go SDKs


joshfree:
Tracking issue for adding OpenTelemetry tracing support for azure-core, azure-identity, and Storage Go SDKs

- See https://opentelemetry.io/docs/golang/tracing/
- See https://github.com/Azure/azure-sdk/issues/501
- See https://gist.github.com/lmolkova/e4215c0f44a49ef824983382762e6b92

/cc @lmolkova @pakrym

jhendrixMSFT:
Codegen changes https://github.com/Azure/autorest.go/issues/501

PR #19416, opened 2022-10-25
Add tracing package

jhendrixMSFT:
This package contains the definitions needed to implement distributed tracing in SDKs and client-side providers.

https://apiview.dev/Assemblies/Review/4bc716d2a1de445ea75a731f427dec52#azcore/tracing

Fixes
- https://github.com/Azure/azure-sdk-for-go/issues/7358
- https://github.com/Azure/azure-sdk-for-go/issues/19280

### DIFF
--- a/sdk/azcore/tracing/constants.go
+++ b/sdk/azcore/tracing/constants.go
@@ -1,0 +1,41 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tracing
+
+// SpanKind represents the role of a Span inside a Trace. Often, this defines how a Span will be processed and visualized by various backends.
+type SpanKind int
+
+const (
+	// SpanKindInternal indicates the span represents an interal operation within an application.  This is the default value.
+	SpanKindInternal SpanKind = 1
+
+	// SpanKindServer indicates the span covers server-side handling of a request.
+	SpanKindServer SpanKind = 2
+
+	// SpanKindClient indicates the span describes a request to a remote service.
+	SpanKindClient SpanKind = 3
+
+	// SpanKindProducer indicates the span was created by a messaging producer.
+	SpanKindProducer SpanKind = 4
+
+	// SpanKindConsumer indicates the span was created by a messaging consumer.
+	SpanKindConsumer SpanKind = 5
+)
+
+// SpanStatus represents the status of a span.
+type SpanStatus int
+
+const (
+	// SpanStatusUnset is the default status code.
+	SpanStatusUnset SpanStatus = 0
+
+	// SpanStatusError indicates the operation contains an error.
+	SpanStatusError SpanStatus = 1
+
+	// SpanStatusOK indicates the operation completed successfully.
+	SpanStatusOK SpanStatus = 2
+)
--- a/sdk/azcore/tracing/constants.go
+++ b/sdk/azcore/tracing/constants.go
@@ -10,7 +10,7 @@ package tracing
 type SpanKind int
 
 const (
-	// SpanKindInternal indicates the span represents an interal operation within an application.  This is the default value.
+	// SpanKindInternal indicates the span represents an internal operation within an application.  This is the default value.
 	SpanKindInternal SpanKind = 1
 
 	// SpanKindServer indicates the span covers server-side handling of a request.
--- a/sdk/azcore/tracing/constants.go
+++ b/sdk/azcore/tracing/constants.go
@@ -10,7 +10,7 @@ package tracing
 type SpanKind int
 
 const (
-	// SpanKindInternal indicates the span represents an internal operation within an application.  This is the default value.
+	// SpanKindInternal indicates the span represents an internal operation within an application.
 	SpanKindInternal SpanKind = 1
 
 	// SpanKindServer indicates the span covers server-side handling of a request.
--- a/sdk/azcore/tracing/tracing.go
+++ b/sdk/azcore/tracing/tracing.go
@@ -17,26 +17,26 @@ type ProviderOptions struct {
 }
 
 // NewProvider creates a new Provider with the specified values.
-//   - impl is the underlying implementation for creating Tracer instances
+//   - newTracerFn is the underlying implementation for creating Tracer instances
 //   - options contains optional values; pass nil to accept the default value
-func NewProvider(impl func(name, version string) Tracer, options *ProviderOptions) Provider {
+func NewProvider(newTracerFn func(name, version string) Tracer, options *ProviderOptions) Provider {
 	return Provider{
-		impl: impl,
+		newTracerFn: newTracerFn,
 	}
 }
 
 // Provider is the factory that creates Tracer instances.
 // It defaults to a no-op provider.
 type Provider struct {
-	impl func(name, version string) Tracer
+	newTracerFn func(name, version string) Tracer
 }
 
 // NewTracer creates a new Tracer for the specified name and version.
 //   - name - the name of the tracer object, typically the fully qualified name of the service client
 //   - version - the version of the module in which the service client resides
 func (p Provider) NewTracer(name, version string) (tracer Tracer) {
-	if p.impl != nil {
-		tracer = p.impl(name, version)
+	if p.newTracerFn != nil {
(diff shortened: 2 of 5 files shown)


JeffreyRichter (approved):

sdk/azcore/tracing/constants.go
```diff
@@ -0,0 +1,41 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tracing
+
+// SpanKind represents the role of a Span inside a Trace. Often, this defines how a Span will be processed and visualized by various backends.
+type SpanKind int
+
+const (
+	// SpanKindInternal indicates the span represents an interal operation within an application.  This is the default value.
```
chlowell, line 13:
```suggestion
	// SpanKindInternal indicates the span represents an internal operation within an application.  This is the default value.
```
And should the value be 0, or does that indicate no setting?

lmolkova: my 2 cents: internal should be a default kind at least for OTel implementation - OTel requires kind to be set.

sdk/azcore/tracing/tracing.go
```diff
@@ -0,0 +1,168 @@
...
+//   - options contains optional values for the span, pass nil to accept any defaults
+func (t Tracer) Start(ctx context.Context, spanName string, options *SpanOptions) (context.Context, Span) {
+	if t.impl != nil {
+		return t.impl(ctx, spanName, options)
+	}
+	return ctx, Span{}
+}
+
+// SpanOptions contains optional settings for creating a span.
+type SpanOptions struct {
```
lmolkova, line 77: Will we be able to extend this struct and add more stuff later on? I recall it might be difficult in some cases.

sdk/azcore/tracing/tracing.go
```diff
@@ -0,0 +1,168 @@
...
+// Tracer is the factory that creates Span instances.
+type Tracer struct {
+	impl func(ctx context.Context, spanName string, options *SpanOptions) (context.Context, Span)
+}
+
+// Start creates a new span and a context.Context that contains it.
+//   - ctx is the parent context for this span. If it contains a Span, the newly created span will be a child of that span, else it will be a root span
+//   - spanName identifies the span within a trace, it's typically the fully qualified API name
+//   - options contains optional values for the span, pass nil to accept any defaults
+func (t Tracer) Start(ctx context.Context, spanName string, options *SpanOptions) (context.Context, Span) {
```
lmolkova, line 69: assuming we need to pass parent span context (e.g. in messaging scenarios, when we receive a message, we'll have a string with traceparent extracted from the message), we'll need to pass it to this method.

How do you envision we'll do it? Via context or options? Both are fine, just want to make sure we have a plan.

sdk/azcore/tracing/tracing.go
jhendrixMSFT: Adding more fields to this (and structs in general) is non-breaking.

The case you might be thinking of is adding new methods to an interface which is breaking (we discussed this early on).  It's why we use `SpanImpl`, a "struct witth funcs", as the abstraction layer.

sdk/azcore/tracing/tracing.go
jhendrixMSFT: By convention, the parent span is passed via the context.

sdk/azcore/tracing/constants.go
jhendrixMSFT: In the OTel adapter I do default to internal if no span kind was specified.

I left zero unspecified as I felt the above is implementation specific and should be left to the adapters.  However, if we (the SDK team) are comfortable with stating that unspecified always defaults to internal for any implementation I can change it.

And just for the record, changing the value post release would be a breaking change, so we should be confident in our decision.

jhendrixMSFT: In fact, having the default be impl-specific means I should also remove the "this is the default" comment.

sdk/azcore/tracing/tracing.go
lmolkova: thank you for the clarification!

sdk/azcore/tracing/tracing.go
lmolkova: span yes, but what about span-context or just a string traceparent header?

sdk/azcore/tracing/constants.go
jhendrixMSFT: If there are no objections, I'll leave this as-is and remove the comment about the default.

lmolkova: sounds good to me

sdk/azcore/tracing/tracing.go
```diff
@@ -0,0 +1,168 @@
...
+
+// ProviderOptions contains the optional values when creating a Provider.
+type ProviderOptions struct {
+	// for future expansion
+}
+
+// NewProvider creates a new Provider with the specified values.
+//   - impl is the underlying implementation for creating Tracer instances
+//   - options contains optional values; pass nil to accept the default value
+func NewProvider(impl func(name, version string) Tracer, options *ProviderOptions) Provider {
```
richardpark-msft, line 22:
```suggestion
func NewProvider(newTracerFn func(name, version string) Tracer, options *ProviderOptions) Provider {
```
It'd be nice if it was styled a bit like a "constructor" function.

sdk/azcore/tracing/tracing.go
```diff
@@ -0,0 +1,168 @@
...
+	return Provider{
+		impl: impl,
+	}
+}
+
+// Provider is the factory that creates Tracer instances.
+// It defaults to a no-op provider.
+type Provider struct {
+	impl func(name, version string) Tracer
```
richardpark-msft, line 31: Is there ever a possibility we might return an error when constructing a Tracer? Or is there another way to signal problems?

sdk/azcore/tracing/constants.go
jhendrixMSFT: Updated comment

richardpark-msft (commented):
Is the usage model here that you'd basically never use this API directly - you'd always be using some other package that implements it, and that package could be more convenient/concrete?

There are some areas that seem a little unclear to me but if you don't ever actually use them then it wouldn't matter much. I'd be more interested in our first implementation of it.

sdk/azcore/tracing/tracing.go
jhendrixMSFT: Good idea, fixed.

sdk/azcore/tracing/tracing.go
jhendrixMSFT: Good question.  This implementation was based on the [OTel for Go](https://pkg.go.dev/go.opentelemetry.io/otel) implementation where these are no-fail.  Our track 1 tracing implementation is based on Open Census and there too these operations are no-fail.

Conceptually I think this makes sense.  Given it's diagnostic tooling, would you want to fail your APIs and potentially bring down the process if your tracing provider is having a bad day?  @lmolkova probably has better insight here.

jhendrixMSFT: It looks like there's global error handling, at least for OTel.

https://pkg.go.dev/go.opentelemetry.io/otel#ErrorHandler

sdk/azcore/tracing/tracing.go
jhendrixMSFT: I see what you're getting at now.  I believe this would go into the options type.  However, I want to dig into this a bit more first.

sdk/azcore/tracing/tracing.go
richardpark-msft: I think it's perfectly okay to say "handle this in your implementation" and stay out of the game.

richardpark-msft (approved):
Approved on my end. @lmolkova,  you are the final gate.

Btw, where would we publish implementations of this? Will they just be in our mono repo?

sdk/azcore/tracing/tracing.go
lmolkova: perfect, since we can evolve options, I have no concern here

lmolkova (approved):
:shipit:

jhendrixMSFT:
For publishing an OTel adapter, I was discussing this with @JeffreyRichter and our current thinking is that we'd introduce a new module in this repo.

Issue #15755, opened 2021-10-08
Azure.Core code coverage improvements


RickWinter:

jhendrixMSFT:
At present, code coverage is at 88.8%.  Presumably the goal is to reach a higher value?

jhendrixMSFT:
Now at 92.3%

PR #19437, opened 2022-10-27
Fixed bug when combining pipeline and client allow lists

jhendrixMSFT:
Specify slice length when creating destination for copy operation.
Added additional tests to improve code coverage.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/15755

### DIFF
--- a/sdk/azcore/internal/exported/request_test.go
+++ b/sdk/azcore/internal/exported/request_test.go
@@ -8,9 +8,12 @@ package exported
 
 import (
 	"context"
+	"io"
 	"net/http"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 const testURL = "http://test.contoso.com/"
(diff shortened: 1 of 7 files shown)


sdk/azcore/internal/exported/request_test.go
```diff
@@ -81,9 +84,32 @@ func TestRequestBody(t *testing.T) {
 	if err := req.Close(); err != nil {
 		t.Fatal(err)
 	}
+	if req.Body() != nil {
+		t.Fatal("expected nil body")
+	}
```
chlowell, line 89:
```suggestion
	require.NotNil(t, req.Body())
```

chlowell (approved):

sdk/azcore/internal/exported/request_test.go
jhendrixMSFT: I'm going to make a separate pass to clean these things up.

PR #19447, opened 2022-10-28
Refactor KeyVaultChallengePolicy

chlowell:
This refactors Key Vault's auth policy to use the extension API added in azcore v1.3.0.

### DIFF
--- a/sdk/keyvault/internal/challenge_policy.go
+++ b/sdk/keyvault/internal/challenge_policy.go
@@ -65,9 +65,11 @@ func (k *keyVaultAuthorizer) authorize(req *policy.Request, authNZ func(policy.T
 			// body, authorize the request, and send it again.
 			rb := reqBody{body, req.Raw().Header.Get("content-type")}
 			req.SetOperationValue(rb)
-			req.SetBody(nil, "")
+			if err := req.SetBody(nil, ""); err != nil {
+				return err
+			}
 		}
-		// returning nil indicates the request is authorized and the bearer token policy should send it
+		// returning nil indicates the bearer token policy should send the request
 		return nil
 	}
 	// else we know the auth parameters and can authorize the request as normal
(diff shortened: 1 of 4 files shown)


sdk/keyvault/internal/challenge_policy.go
```diff
@@ -161,7 +123,7 @@ func (c *challengePolicyError) Unwrap() error {
 var _ errorinfo.NonRetriable = (*challengePolicyError)(nil)
 
 // sets the k.scope and k.tenantID from the WWW-Authenticate header
```
richardpark-msft, line 125: Looks like this comment needs updating.

Also, maybe it's "updateScopeAndTeneant", not "find" (which implies that it doesn't change anything).

sdk/keyvault/internal/challenge_policy.go
```diff
@@ -7,139 +7,101 @@
...
-		k.mainResource.Expire()
-
-		// Find the scope and tenant again in case they have changed
-		err := k.findScopeAndTenant(resp, req.Raw())
-		if err != nil {
-			// Error parsing challenge, doomed to fail. Return
-			return resp, cloneReqErr
+	// reattach the request's body, if it was removed by authorize()
+	var rb reqBody
+	if req.OperationValue(&rb) {
```
jhendrixMSFT, line 86: If this doesn't find the thing, what will happen?

chlowell: Then we send the request without a body. If it's e.g. a GET, no problem; if it's a POST, we get an error response from Key Vault. That should be a 400, so non-retriable and therefore an application error. Calling the client method again should lead to success because the policy won't remove the body again. Do you think we should check the verb to decide whether the request should have had a body so we can return a better error pointing to an azcore bug?

jhendrixMSFT: Hmm we could, but I suspect this code won't change much so maybe it's more effort than it's worth (I'll let you decide on that part).  It might be good to add a comment about this though for future maintainers.

chlowell: I added a comment because I don't want this policy to have an opinion about which requests should have a body.

PR #19448, opened 2022-10-28
Add optional auth challenge handling to BearerTokenPolicy

### LINKED ISSUE #17554: Add authentication challenge support to BearerTokenPolicy
Key Vault and Tables have vendored `BearerTokenPolicy` and `ExpiringResource` from `azcore` to implement their challenge protocols. It would be better if they could implement just their challenge handling logic and plug it in to `azcore/runtime.BearerTokenPolicy` with a callback, for example.

chlowell:
Closes #17554 by adding a couple of extension points SDK developers can use to insert custom logic around BearerTokenPolicy's default behavior. These allow clients to elicit challenges (to discover auth parameters) and to handle them, without having to copy and tweak BearerTokenPolicy, as we've done for Key Vault and ARM. This will simplify supporting CAE and ARM cross-tenant authentication. #19447 has a draft implementation for Key Vault showing this new API in use.

### DIFF
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -140,8 +140,6 @@ type BearerTokenOptions struct {
 }
 
 // AuthorizationHandler allows SDK developers to insert custom logic that runs when BearerTokenPolicy must authorize a request.
-// Errors returned by its functions should be have a NonRetriable() marker method to prevent [runtime.RetryPolicy] retrying
-// requests the AuthorizationHandler can't authorize.
 type AuthorizationHandler struct {
 	// OnRequest is called each time the policy receives a request. Its func parameter authorizes the request with a token
 	// from the policy's given credential. Implementations that need to perform I/O should use the Request's context,
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -7,6 +7,7 @@
 package policy
 
 import (
+	"net/http"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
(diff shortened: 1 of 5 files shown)


richardpark-msft (commented):
Seems reasonable to me. Leave the final judgement to @jhendrixMSFT, of course.

sdk/azcore/policy/policy.go
```diff
@@ -132,5 +133,28 @@ type TokenRequestOptions struct {
...
 type BearerTokenOptions struct {
-	// placeholder for future options
+	// AuthorizationHandler allows SDK developers to run client-specific logic when BearerTokenPolicy must authorize a request.
+	// When this field isn't set, the policy follows its default behavior of authorizing every request with a bearer token from
+	// its given credential.
+	AuthorizationHandler AuthorizationHandler
+}
+
+// AuthorizationHandler allows SDK developers to insert custom logic that runs when BearerTokenPolicy must authorize a request.
+// Errors returned by its functions should be have a NonRetriable() marker method to prevent [runtime.RetryPolicy] retrying
```
RickWinter, line 143: Any reason to not make this a formal requirement?
```suggestion
// Errors returned must have a NonRetriable() marker method to prevent [runtime.RetryPolicy] retrying
```

jhendrixMSFT: If we always want this to return a non-retriable error, we should just check for `NonRetriable()` and if it's not there then wrap it ourselves.

chlowell: I'll implement the wrapping approach. There may be a scenario in which retrying after an error from one of these functions is the right thing to do, but that would be an unusual scenario (let me know if you think of one; I haven't). It's more likely someone will trigger pointless retries by returning a plain old error from one of these functions.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -93,5 +94,31 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+		if _, ok := err.(errorinfo.NonRetriable); !ok {
+			err = btpError{err}
+		}
+	}
+	return err
+}
+
+// btpError is a wrapper that ensures RetryPolicy doesn't retry requests BearerTokenPolicy couldn't authorize
+type btpError struct {
+	err error
```
jhendrixMSFT, line 111: If you anonymously embed `error` then you can remove the `Error()` and `Unwrap()` methods.

We should put this into `sdk/internal` at some point as we have several copies of this floating around.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -93,5 +94,31 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
 			}
 		}
 	}
-	return res, err
+	return res, ensureNonRetriable(err)
+}
+
+func ensureNonRetriable(err error) error {
+	if err != nil {
+		if _, ok := err.(errorinfo.NonRetriable); !ok {
```
jhendrixMSFT, line 102: I think you should use `errors.As()` into an `errorinfo.NonRetriable` just in case `err` needs to be `Unwrap()`'ed.

sdk/azcore/runtime/policy_bearer_token.go
chlowell: Thanks for the embedding tip. I opened #19467 to track moving this into `internal`.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -42,23 +46,71 @@ func acquire(state acquiringResourceState) (newResource azcore.AccessToken, newE
...
-		req: req,
+	var err error
+	if b.authzHandler.OnRequest != nil {
+		err = b.authzHandler.OnRequest(req, b.authenticateAndAuthorize(req))
+	} else {
+		err = b.authenticateAndAuthorize(req)(policy.TokenRequestOptions{Scopes: b.scopes})
 	}
-	tk, err := b.mainResource.Get(as)
+	if err != nil {
+		return nil, ensureNonRetriable(err)
```
jhendrixMSFT, line 82: I think this call, and the call on line 98, need to be moved such that they only are call when the `On*()` methods are called yes?

chlowell: Well, it's an interesting question. I wrote it like this because I expect we want all errors from this policy to be nonretriable, and it seems like an oversight that we don't already wrap them to be sure. In practice, it usually doesn't matter because any errors originate in `azidentity`, whose auth errors are nonretriable. But if a user gives the policy a custom credential implementation?

jhendrixMSFT: I don't think we have enough context at this point though to tell if the error is transient, do we?  e.g. if one of these extension methods fails to make an HTTP request due to some transient condition, we should ensure that the retry policy will kick in.

chlowell: If an extension method needs to send a request, it brings its own transport and is therefore responsible for retrying that request. This is baked into the common case: `azidentity` credentials have their own pipelines and retry policies.

jhendrixMSFT: Oh yes of course, makes sense.

jhendrixMSFT (approved):

JeffreyRichter (approved):

PR #19462, opened 2022-10-31
[azquery] sovereign cloud support

gracewilcox:
This PR adds sovereign cloud support to Monitor Query for Go plus documentation.

No breaking changes, but updates internal implementation to allow for other clouds besides Azure Public Cloud.


sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,39 @@
...
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azquery
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+const monitorQueryLogs cloud.ServiceName = "azqueryLogs"
```
jhendrixMSFT, line 11: Nit, these should go in a `const` block.
```go
const(
    // put constants here
)
```

sdk/monitor/azquery/utils_test.go
```diff
@@ -61,10 +66,23 @@ func TestMain(m *testing.M) {
...
+		tenantID := lookupEnvVar("AZQUERY_TENANT_ID")
+		clientID := lookupEnvVar("AZQUERY_CLIENT_ID")
+		secret := lookupEnvVar("AZQUERY_CLIENT_SECRET")
+		credential, err = azidentity.NewClientSecretCredential(tenantID, clientID, secret, nil)
 		if err != nil {
 			panic(err)
 		}
+
+		if cloudEnv, ok := os.LookupEnv("AZQUERY_ENVIRONMENT"); ok {
+			if cloudEnv == "AzureUSGovernment" {
```
jhendrixMSFT, line 78: You should use [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold) for a case-insensitive comparison.

sdk/monitor/azquery/custom_client.go
```diff
@@ -27,28 +28,49 @@ type LogsClientOptions struct {
...
-	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{"https://api.loganalytics.io/.default"}, nil)
+
+	host := cloud.AzurePublic.Services[monitorQueryLogs].Endpoint
+	if c, ok := options.Cloud.Services[monitorQueryLogs]; ok {
+		host = c.Endpoint
+	}
+	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{host + "/.default"}, nil)
 	pl := runtime.NewPipeline(moduleName, version, runtime.PipelineOptions{PerRetry: []policy.Policy{authPolicy}}, &options.ClientOptions)
-	return &LogsClient{pl: pl}
+	return &LogsClient{host: host + "/v1", pl: pl}
```
jhendrixMSFT, line 57: Should the `/v1` go in the cloud config?

gracewilcox: Host is also used in line 55 to create the scope, and the scope doesn't use "/v1". It's easier and cleaner to add "/v1" in line 57 then having to strip the string of it in line 55.

jhendrixMSFT: I missed that, makes sense.

sdk/monitor/azquery/utils_test.go
gracewilcox: fixed in commit [d2eafa6](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/d2eafa65677cd3eeccdfa1f740ecf1ff6bd959f9)

sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,39 @@
...
+const monitorQueryMetrics cloud.ServiceName = "azqueryMetrics"
+
+func init() {
+	cloud.AzureChina.Services[monitorQueryLogs] = cloud.ServiceConfiguration{
+		Audience: "",
```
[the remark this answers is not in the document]
gracewilcox: fixed in commit [d2eafa6](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/d2eafa65677cd3eeccdfa1f740ecf1ff6bd959f9)

sdk/monitor/azquery/cloud_config.go
gracewilcox: fixed in commit [d2eafa6](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/d2eafa65677cd3eeccdfa1f740ecf1ff6bd959f9)

sdk/monitor/azquery/README.md
```diff
@@ -329,6 +331,7 @@ comments.
 [azure_monitor_create_using_portal]: https://docs.microsoft.com/azure/azure-monitor/logs/quick-create-workspace
 [azure_monitor_overview]: https://docs.microsoft.com/azure/azure-monitor/overview
 [context]: https://pkg.go.dev/context
+[cloud_documentation]: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/cloud/doc.go
```
chlowell, line 334:
```suggestion
[cloud_documentation]: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud
```

gracewilcox: Thanks!

sdk/monitor/azquery/custom_client.go
```diff
@@ -27,28 +28,49 @@ type LogsClientOptions struct {
...
 // NewLogsClient creates a client that accesses Azure Monitor logs data.
 func NewLogsClient(credential azcore.TokenCredential, options *LogsClientOptions) *LogsClient {
 	if options == nil {
 		options = &LogsClientOptions{}
 	}
-	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{"https://api.loganalytics.io/.default"}, nil)
+
+	host := cloud.AzurePublic.Services[monitorQueryLogs].Endpoint
+	if c, ok := options.Cloud.Services[monitorQueryLogs]; ok {
+		host = c.Endpoint
```
chlowell, line 53: This elides an error condition: a user may set `options.Cloud.Services[monitorQueryLogs]` but not that value's `Endpoint` field. In that case, it appears the user intends, but has failed, to provide private cloud configuration; what should the constructor do? This is why we return an error from ARM client constructors: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/runtime/pipeline.go#L57

@jhendrixMSFT another instance where we'd be in a jam if the module were GA already

sdk/monitor/azquery/utils_test.go
```diff
@@ -103,18 +122,39 @@ func startLogsTest(t *testing.T) *azquery.LogsClient {
...
+	var opts *azquery.MetricsClientOptions
+	if recording.GetRecordMode() == recording.LiveMode {
+		transport := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					Renegotiation: tls.RenegotiateOnceAsClient,
+				},
+			},
+		}
+		opts = &azquery.MetricsClientOptions{ClientOptions: azcore.ClientOptions{Transport: transport, Cloud: clientCloud}}
```
chlowell, line 139: Shouldn't we set the cloud in playback as well?

sdk/monitor/azquery/custom_client.go
```diff
@@ -27,28 +28,49 @@ type LogsClientOptions struct {
...
 func NewMetricsClient(credential azcore.TokenCredential, options *MetricsClientOptions) *MetricsClient {
 	if options == nil {
 		options = &MetricsClientOptions{}
 	}
-	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{"https://management.azure.com/.default"}, nil)
+	host := cloud.AzurePublic.Services[monitorQueryMetrics].Endpoint
+	if c, ok := options.Cloud.Services[monitorQueryMetrics]; ok {
+		host = c.Endpoint
+	}
+	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{host + "/.default"}, nil)
```
chlowell, line 69: I'm thinking these ctors should read `ServiceConfiguration.Audience` because although the `host + /.default` pattern works for well-known clouds, I've seen scopes for Azure Stack that don't follow this pattern.

sdk/monitor/azquery/custom_client.go
jhendrixMSFT: I was wondering about this but didn't look deep enough.  Makes me think that we should update the guidelines to have ctors return errors by default.

sdk/monitor/azquery/utils_test.go
gracewilcox: We could set it here, but it's not necessary. The cloud during testing is only changed from the default public setting in live mode (lines 78-85). The tests automatically fail in playback mode with different clouds. It can't identify the recording since the URLs are different.

sdk/monitor/azquery/custom_client.go
chlowell: Yeah. Some, but not all, constructors need to return errors, so we have to choose consistency or brevity. I keep forgetting whether Key Vault constructors return errors, so I think we should favor consistency. When last we discussed this, I believe @JeffreyRichter was open to returning errors across the SDK?

sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,35 @@
...
+		Endpoint: "https://api.loganalytics.io",
+	}
+	cloud.AzureChina.Services[monitorQueryMetrics] = cloud.ServiceConfiguration{
+		Endpoint: "https://management.chinacloudapi.cn/",
+	}
+	cloud.AzureGovernment.Services[monitorQueryMetrics] = cloud.ServiceConfiguration{
+		Endpoint: "https://management.usgovcloudapi.net/",
+	}
+	cloud.AzurePublic.Services[monitorQueryMetrics] = cloud.ServiceConfiguration{
+		Endpoint: "https://management.azure.com",
```
chlowell, line 33: MetricsClient talks to ARM, which makes me think we could simplify its constructor by calling `arm/runtime.NewPipeline()`. Come to think of it, I wonder whether we could simplify further by generating it as an ARM client 🤔

sdk/monitor/azquery/custom_client.go
JeffreyRichter: Yes, for client ctors we can always return an error.

sdk/monitor/azquery/custom_client.go
gracewilcox: fixed in commit [3772df1](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/3772df1fc545f3ba93ce4eb8c7ba45c582aff39d)

sdk/monitor/azquery/cloud_config.go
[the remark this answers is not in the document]
gracewilcox: Updated to give audience a value in commit [3772df1](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/3772df1fc545f3ba93ce4eb8c7ba45c582aff39d)

sdk/monitor/azquery/custom_client.go
gracewilcox: Updated in commit [3772df1](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/3772df1fc545f3ba93ce4eb8c7ba45c582aff39d) to set "/v1" in cloud config

sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,35 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azquery
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+const (
+	monitorQueryLogs    cloud.ServiceName = "azqueryLogs"
+	monitorQueryMetrics cloud.ServiceName = "azqueryMetrics"
```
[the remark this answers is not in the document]
gracewilcox: Fixed in commit [3772df1](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/3772df1fc545f3ba93ce4eb8c7ba45c582aff39d)

sdk/monitor/azquery/custom_client.go
gracewilcox: Updated constructors to return an error in commit [3772df1](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/3772df1fc545f3ba93ce4eb8c7ba45c582aff39d)

sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azquery
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+const (
+	MonitorQueryLogs    cloud.ServiceName = "azqueryLogs"
```
jhendrixMSFT, line 12: These will need doc comments.  In addition, I think this is the first time we've added constants related to cloud config (outside of `azcore`), so we don't have a well-established naming pattern.

jhendrixMSFT: CC @JeffreyRichter

gracewilcox: @chlowell Do we want to standardize the way we name the cloud.ServiceName? Also, I have `MonitorQueryLogs` and `MonitorQueryMetrics` declared in `azquery`, should we move it to the `cloud` package? That's where the arm cloud.ServiceName is declared.  
https://github.com/Azure/azure-sdk-for-go/blob/80bf31d4cb0db751479079cc4a1718adc7c8b547/sdk/azcore/cloud/cloud.go#L28

jhendrixMSFT: We definitely want them in `azquery`.  The thinking is that modules would declare these constants and add them to the global config as you did below.  That way, new modules don't have to update `azcore/cloud` with their constants (which might also introduce some naming complexities if they're all in the same package).

So really the question is, what's the naming pattern we want to establish for these per-module, cloud-config constants?

JeffreyRichter: If we move service names into the cloud pkg, any new service updates cloud. This is perhaps OK if no other pkg takes a dependency on the cloud pkg.

JeffreyRichter: ServiceName is a type like an enum type, right? So, we would name them like we name enums: ServiceNameResourceManager. ServiceNameMonitorQueryLogs?

I know they're long but what are you gonna do?

jhendrixMSFT: Good point about `ServiceName` being an enum type, so I agree we should follow the convention.  I would then suggest that we shorten the names.  Maybe something like this.
```go
ServiceNameLogs cloud.ServiceName = "azqueryLogs"
```
Making the fully qualified type name be `azquery.ServiceNameLogs` etc.

JeffreyRichter: Works for me!

sdk/monitor/azquery/cloud_config.go
gracewilcox: leaving metrics client as is for now

sdk/monitor/azquery/cloud_config.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azquery
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+const (
+	ServiceNameLogs    cloud.ServiceName = "azqueryLogs"
```
jhendrixMSFT, line 12: Missing doc comments.

gracewilcox: added in commit [be375c6](https://github.com/Azure/azure-sdk-for-go/pull/19462/commits/be375c658075097c3a95c800c5a584376381d5f4)

PR #19470, opened 2022-11-01
Expose tracing provider from the pipeline

jhendrixMSFT:
This simplifies client access to the configured tracing provider. The Pipeline type is no longer an alias, however its definition is identical to the internal version.

Example usage.
```go
func NewLRORetrysClient(moduleName, moduleVersion string, options *LRORetrysClientOptions) *LRORetrysClient {
	cp := LRORetrysClientOptions{}
	if options != nil {
		cp = *options
	}
	pl := runtime.NewPipeline(moduleName, moduleVersion, runtime.PipelineOptions{}, &cp.ClientOptions)
	client := &LRORetrysClient{
		pl: pl,
	}
	client.tracer = pl.TracingProvider().NewTracer(reflect.TypeOf(client).Elem().PkgPath(), moduleVersion)
	return client
}
```

### DIFF
--- a/sdk/azcore/runtime/pipeline.go
+++ b/sdk/azcore/runtime/pipeline.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/tracing"
 )
 
 // PipelineOptions contains Pipeline options for SDK developers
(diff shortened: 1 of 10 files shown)


sdk/azcore/runtime/pipeline.go
```diff
@@ -23,7 +24,20 @@ type PipelineOptions struct {
...
+	traceProv tracing.Provider
+}
+
+// Do implements the policy.Transporter interface for the Pipeline type.
+func (p Pipeline) Do(req *policy.Request) (*http.Response, error) {
+	return p.internal.Do(req)
+}
+
+// TracingProvider returns the tracing provider associated with this Pipeline.
+func (p Pipeline) TracingProvider() tracing.Provider {
```
jhendrixMSFT, line 38: Not sure how much I like this.  While it's convenient and solves the issue, the tracing provider doesn't really have anything to do with the pipeline at a conceptual level.

JeffreyRichter: In C++, the tracing provider is in the Client options. The ctor code looks for it being set and then adds the policy if necessary to the pipeline. The provider itself is another field in the client. So, the pipeline doesn't know anything about tracing - just another policy gets injected into it.

jhendrixMSFT: How will the client name get passed to [`CreateTracer`](https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/inc/azure/core/tracing/tracing.hpp#L31)?

PR #19471, opened 2022-11-02
[azeventhubs, azservicebus] stress eng script requires a specific version, not ~version's

richardpark-msft:
The stress deploy script is doing version checking, but it doesn't handle versions with `~` as the matcher, which I use exclusively.

Changing for now, issue filed for the stress script itself at https://github.com/Azure/azure-sdk-tools/issues/4590.


richardpark-msft:
(Looks like @benbp is already fixing the bug!)

PR #19478, opened 2022-11-03
AppConfig: Make `OnlyIfChanged`/`OnlyIfUnchanged` to work

### LINKED ISSUE #19297: [azappconfig] Issue with `getSettings` method's `OnlyIfChanged ` parameter
### Bug Report

Creating a different issue get more visibility, earlier I have mentioned this issue as part of a different issue here - https://github.com/Azure/azure-sdk-for-go/issues/19081

According to the document, we can set this parameter to true and response will come only if the value gets modified. I have the below code for `getSettings` - 

```
sentinelKeyResponse, err := r.client.GetSetting(
		ctx,
		"key",
		&azappconfig.GetSettingOptions{
			OnlyIfChanged: true,
			Label:         r.getLabelFromMetadata(req.Metadata),
		},
	)
```
But I always get the response irrespective of value has changed or not. I can see the *sentinelKeyResponse.Etag and can see its value.
My understanding is if the value gets updated, it should give the response only 1st time when accessed after that it should given some error response code.

Kindly share your thoughts and please let me know if I am supposed to use it differently. Thanks

antkmsft:
Closes #19297.


jhendrixMSFT (approved):

sdk/data/azappconfig/CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
 # Release History
 
-## 0.4.4 (Unreleased)
+## 0.5.0 (Unreleased)
```
jhendrixMSFT, line 3: Isn't there a constant that needs to be updated too?

antkmsft: You are right, and the CI just failed for that reason. I just updated it in the most recent commit.

PR #19486, opened 2022-11-03
Unexport APIVersionPolicy and constructor

chlowell:
Also adding an options bag.

### DIFF
--- a/sdk/azcore/runtime/policy_api_version.go
+++ b/sdk/azcore/runtime/policy_api_version.go
@@ -24,13 +24,10 @@ const (
 	APIVersionLocationHeader = 1
 )
 
-// apiVersionPolicyOptions contains optional parameters for apiVersionPolicy
-type apiVersionPolicyOptions struct{}
-
 // newAPIVersionPolicy constructs an APIVersionPolicy. name is the name of the query parameter or header and
 // version is its value. If version is "", Do will be a no-op. If version isn't empty and name is empty,
 // Do will return an error.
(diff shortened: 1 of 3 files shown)


sdk/azcore/runtime/policy_api_version.go
```diff
@@ -24,15 +24,18 @@ const (
 	APIVersionLocationHeader = 1
 )
 
-// NewAPIVersionPolicy constructs an APIVersionPolicy. name is the name of the query parameter or header and
+// apiVersionPolicyOptions contains optional parameters for apiVersionPolicy
+type apiVersionPolicyOptions struct{}
```
jhendrixMSFT, line 28: Personally, I think this is less important now that the constructor isn't exported.  It's fine to leave it too, up to you.

jhendrixMSFT (approved):

JeffreyRichter (approved):

jhendrixMSFT:
Does the changelog need to be updated?

sdk/azcore/runtime/policy_api_version.go
JeffreyRichter: Agreed, your call Charles.

chlowell:
No, the only part of this in the changelog is `ClientOptions.APIVersion`

sdk/azcore/runtime/policy_api_version.go
chlowell: ✂️✂️✂️

PR #19489, opened 2022-11-03
Consolidate API version policy options

chlowell:
Following from Joel's comment that it's awkward for `PipelineOptions` to have two fields for `APIVersionPolicy` options, and what if we later want to add a third, this PR consolidates them in a new type. The result is easier to extend but still awkward because although it seems best for the policy constructor to take this new type, its values are logically required. That is to say, it doesn't make sense to construct `APIVersionPolicy` without telling it where on to set the version on requests. Functionally speaking though, these values are optional because the policy is a no-op when they aren't set. (That's a workaround for `NewPipeline()` being unable to return an error.) So, please look it over and share your thoughts--maybe there's a tidier third way here.

### DIFF
--- a/sdk/azcore/runtime/policy_api_version.go
+++ b/sdk/azcore/runtime/policy_api_version.go
@@ -14,6 +14,12 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 )
 
+// APIVersionOptions contains options for API versions
+type APIVersionOptions struct {
+	Location APIVersionLocation
+	Name     string
+}
+
 // APIVersionLocation indicates which part of a request identifies the service version
 type APIVersionLocation int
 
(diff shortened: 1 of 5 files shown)


JeffreyRichter (approved):

sdk/azcore/runtime/policy_api_version.go
```diff
@@ -14,6 +14,12 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 )
 
+// APIVersionOptions contains options for API versions
+type APIVersionOptions struct {
+	Location APIVersionLocation
```
jhendrixMSFT, line 19: Doc comments for this and the `Name` field.

jhendrixMSFT (approved):
Add some doc comments but looks good otherwise.

PR #19492, opened 2022-11-04
[azservicebus/azeventhubs] Fix "$cbs node has already been opened" issue

### LINKED ISSUE #19347: [azeventhubs] stress: $cbs link opened more than once
The $cbs link is used for authentication. It's an exclusive link - you can only have one of these open for your connection.

In our code we make this happen using a mutex, which protects the code that actually creates (and disposes) the $cbs link. I recently got an error in one of our stress tests where this somehow did NOT work, but it's unclear what led to that:

> 

There were some small windows where it seemed like we might leak a link on certain kinds of errors. Those are being addressed in PR#19326, but we'll need to keep an eye out and see if the error comes back. For now, I'm also making it so all the tests are in verbose mode in some runs.

richardpark-msft:
When looking for leaks in negotiateClaim I missed the most "obvious" one - if you cancel the function we still need to close the link we opened. If we don't we leak it, and the service will then start failing future negotiations to open it.

This showed up in https://github.com/Azure/azure-sdk-for-go/issues/19347, originally, but a fix I made recently in SB to test attaching/detaching links made it even more obvious.

Fixes #19347

### DIFF
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -1,14 +1,10 @@
 # Release History
 
-## 0.2.1 (Unreleased)
-
-### Features Added
-
-### Breaking Changes
+## 0.2.1 (2022-11-08)
 
 ### Bugs Fixed
 
-### Other Changes
+- $cbs link is properly closed, even on cancellation (#19492)
 
 ## 0.2.0 (2022-10-17)
 
--- a/sdk/messaging/azeventhubs/internal/eh/stress/Chart.yaml
+++ b/sdk/messaging/azeventhubs/internal/eh/stress/Chart.yaml
@@ -8,5 +8,5 @@ annotations:
   namespace: 'go'
 dependencies:
 - name: stress-test-addons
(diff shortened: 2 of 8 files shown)


sdk/messaging/azeventhubs/internal/cbs.go
```diff
@@ -35,7 +35,7 @@ func NegotiateClaim(ctx context.Context, audience string, conn amqpwrap.AMQPClie
 	}
 
 	defer func() {
-		if err := link.Close(ctx); err != nil {
+		if err := link.Close(context.Background()); err != nil {
```
jhendrixMSFT, line 38: Should these have a timeout, or does this happen on a separate goroutine?

richardpark-msft: (just updating with offline conversation)

We're going to leave this as is, and double down and add this to more places. The proper cleanup and closing of links is important enough for the "health" of our connection that leaving it in an incomplete state can cause subtle issues that'd be best to avoid.

jhendrixMSFT (approved):

PR #19503, opened 2022-11-04
Add WorkloadIdentityCredential

chlowell:
This credential authenticates via workload identity on Kubernetes. It's part of the DefaultAzureCredential chain, in which it's configured by environment variables set by the platform.


sdk/azidentity/workload_identity.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// GetToken requests an access token from Azure Active Directory. Azure SDK clients call this method automatically.
+func (w *WorkloadIdentityCredential) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	return w.cred.GetToken(ctx, opts)
+}
+
+// getAssertion returns the specified file's content, which must be an appropriately formatted assertion (the platform is responsible for formatting)
+func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, error) {
+	w.mtx.RLock()
```
jhendrixMSFT, line 59: Would using the `ExpiringResource` from the `internal` module be advantageous here?

chlowell: I don't think so. `ExpiringResource` implements a fault-tolerant refresh window during which it attempts to update the resource before it expires. In this case we just want to ensure we read the file at most once every 5 minutes. The platform is responsible for keeping the file up to date, and we can't establish a refresh window around it because we don't know when the content expires.

jhendrixMSFT: Thanks for the explanation.  Might be good to add a comment about the platform keeping the file current, and/or a link to any docs that explains how this works.

chlowell: You mean, replace this lock upgrade/double check approach with waiting on a `sync.Cond`?

jhendrixMSFT: Correct.

chlowell: Here's what I came up with: [gist](https://gist.github.com/chlowell/3a7b9491f6c0b0aaf0c79695b0d3103c). I find it harder to read (and to write!). I also found it's marginally slower than the `RWMutex` approach, I guess because goroutines must acquire an exclusive lock before calling `Cond.Wait()`, and that overhead doesn't pay off because we're waiting on relatively fast I/O. Am I overlooking some other possible benefit in using `sync.Cond`?

jhendrixMSFT: My thought process was, if we have an algorithm for concurrently checking/acquiring/updating shared state (OAuth tokens), then the same algorithm should apply here too.  But perhaps this case is different enough (local file IO, no refresh window) that it doesn't matter as much.

Maybe @JeffreyRichter has an opinion on this.

sdk/azidentity/workload_identity.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// getAssertion returns the specified file's content, which is expected to be a Kubernetes service account token.
+// Kubernetes is responsible for updating the file as service account tokens expire.
+func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, error) {
+	w.mtx.RLock()
+	if w.expires.Before(time.Now()) {
+		// ensure only one goroutine at a time updates the assertion
+		w.mtx.RUnlock()
+		w.mtx.Lock()
```
jhendrixMSFT, line 65: If N goroutines are reading and they all see that the assertion has expired, then they all end up waiting here (typical double-checked locking pattern).  Maybe it's fine.  Seeing this is what started me thinking about if we should use expiring resource.

sdk/azidentity/workload_identity.go
jhendrixMSFT: I'll also add that what you have is correct and can always be changed later, so we don't have to block on this now if we want to get the feature out.

jhendrixMSFT (approved):

sdk/azidentity/workload_identity.go
chlowell: That makes sense. I agree it's better in principle for goroutines to suspend instead of spinning. It seems not to matter here because the critical section is fast but maybe it won't be in some scenario. I'm happy to start with the simpler approach and tweak later, if it becomes a perf problem.

Issue #18517, opened 2022-07-01
[azservicebus] Receiver indefinitely "stuck" after a long idle period


tkent:
### Bug Report

After a long period of inactivity, a receiver will stop receiving new messages. A "long period" is sometime more than 8 hours and less than 13 days, but exactly how long is unknown.

(The problem was originally brought up [in this comment](https://github.com/Azure/azure-sdk-for-go/issues/16088#issuecomment-1094537898))

This is very straightforward to demonstrate if you are willing to wait and setup a dedicated service bus instance.  It occurs frequently with infrastructure used for QA, since that will often not receive any activity over weekends and holidays. 

### SDK Versions Used

I have seen this behavior across many versions of the `azure-sdk-for-go`, but the most recent test was conducted using these versions:

```
github.com/Azure/azure-sdk-for-go/sdk/azcore v1.1.0
github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.1.0
github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v1.0.1
```

###  About the most recent time this was reproduced

We most recently reproduced this by running a small golang app in an AKS cluster using a managed identity assigned by `aad-pod-identity`.

In this test, we setup a dedicated azure service bus + managed identity (terraform below) and let the app run. After 13 days, we came back to it . No errors had been emitted, just the regular startup message for the app. I then entered a message into the bus. The receiver in the app did not pickup the message after waiting 30 minutes. We deleted the pod running the app and allowed it to be recreated by the deployment. The replacement pod immediately picked up the message.

### Workaround

We can work around this issue by polling for messages using a 10 minute timeout and restarting in a loop. Our workaround looks like this and is known to work for weeks without an issue.

```golang
rcvrCtxForSdkWorkaround, canceller := context.WithTimeout(ctx, 10*time.Minute)
messages, err := azsbReceiver.ReceiveMessages(rcvrCtxForSdkWorkaround, 1, nil)
canceller()
if err != nil && !errors.Is(err, context.DeadlineExceeded) {
	r.logger.Info(EvtNameErrRetrievingMsgs, map[string]string{
		"error": err.Error(),
	})
	continue
}
// This just means that the context was closed before any messages
// were picked up. This could have been the context (master or)
// rcvrCtxForSdkWorkaround, so a loop is required.
if len(messages) == 0 {
	continue
}
```

### The terraform for the test bus

The terraform below was used to setup the test bus and assign the app identity access to it.

[...]

PR #19506, opened 2022-11-05
[azservicebus] Make the client side idle timer work across multiple calls

richardpark-msft:
I added in a simple idle timer in #19465, which would expire the link if our internal message receive went longer than 5 minutes. This extends that to track it across multiple consecutive calls as well, in case the user calls and cancels multiple times in a row, eating up 5 minutes of wall-clock time.

This is actually pretty similar to the workaround applied by the customer here in #18517 but tries to take into account multiple calls and also recovers the link without exiting ReceiveMessages().


richardpark-msft:
Merge failure was just the changelog.

sdk/messaging/azservicebus/internal/idle_check.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+)
+
+type LocalIdleTracker struct {
+	// MaxDuration controls how long we'll wait, on the client side, for the first message
+	MaxDuration time.Duration
+
+	// IdleStart marks the first time the user cancelled in a string of cancellations.
+	// It gets reset any time there is a success or a non-cancel related failure.
+	// NOTE: this is public for some unit tests but isn't intended to be set by external code.
```
jhendrixMSFT, line 18: Maybe instead of exposing the field, you add a `Testing_IdleStart()` method or something like that?

richardpark-msft: I thought about it, I actually set and read it at this point for tests so it'd be me adding in getters and setters, which is probably worse.

jhendrixMSFT: Ah ok yeah that would be ugly then.

jhendrixMSFT (approved):

PR #19527, opened 2022-11-09
use the actual release date in the changelog

### LINKED ISSUE #19538: Modify the generator release-v2 tool to make sure it could use the actual release date in the changelog
Current the auto release pipeline will use the "today"'s date to modify the release date in changelog.
We could change it to use the release date on the release request instead so that we do not have to modify it manually.

Alancere:
Resolve #19538

### DIFF
--- a/eng/tools/generator/cmd/issue/request/request.go
+++ b/eng/tools/generator/cmd/issue/request/request.go
@@ -110,10 +110,7 @@ func NewReleaseRequestIssue(issue github.Issue) (*ReleaseRequestIssue, error) {
 	targetDate := regexp.MustCompile(`\d*-\d*-\d*`).FindString(contents[releaseDateKeyword])
 	releaseDate, err := time.Parse("2006-01-02", targetDate)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


eng/tools/generator/cmd/issue/request/request.go
```diff
@@ -103,11 +106,22 @@ func NewReleaseRequestIssue(issue github.Issue) (*ReleaseRequestIssue, error) {
 		linkKeyword, tagKeyword, releaseDateKeyword,
 	})
 
+	// get release date
+	targetDate := regexp.MustCompile(`\d*-\d*-\d*`).FindString(contents[releaseDateKeyword])
+	releaseDate, err := time.Parse("2006-01-02", targetDate)
+	if err != nil {
```
tadelesh, line 112: I'd prefer to use today as `targetDate` than throw an error.

Alancere: fix

tadelesh (approved):

PR #19529, opened 2022-11-09
azidentity multitenant authentication API

chlowell:
Usage looks like this:
```go
// cred will request tokens from the configured tenant unless another is specified in GetToken args
cred, err := azidentity.NewClientSecretCredential("primary-tenant", "client-id", "secret", &azidentity.ClientSecretCredentialOptions{
	// cred will (attempt to) authenticate in any requested tenant; can also list specific tenants here
	AdditionallyAllowedTenants: []string{"*"},
})
```
`GetToken()` returns an error when its arguments specify a tenant the credential isn't configured to allow.


sdk/azidentity/go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/Azure/azure-sdk-for-go/sdk/azidentity
 
 go 1.18
 
+replace github.com/Azure/azure-sdk-for-go/sdk/azcore => ../azcore
```
jhendrixMSFT, line 5: This change is dependent on some unreleased `azcore` functionality?

chlowell: Yes, `TokenRequestOptions.TenantID`

sdk/azidentity/azidentity.go
```diff
@@ -73,6 +75,41 @@ var getPublicClient = func(clientID, tenantID string, co *azcore.ClientOptions)
...
+	return cp
+}
+
+// resolveTenant returns the correct tenant for a token request given a credential's configuration
+func resolveTenant(defaultTenant, reqTenant string, allowedTenants []string) (string, error) {
+	if reqTenant == "" || reqTenant == defaultTenant {
+		return defaultTenant, nil
+	}
+	if defaultTenant == "adfs" {
+		return "", errors.New("the authority doesn't support tenants")
```
jhendrixMSFT, line 100: Should this error message include the term "adfs" someplace?

sdk/azidentity/azure_cli_credential.go
```diff
@@ -58,8 +63,9 @@ func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredent
 	}
 	cp.init()
 	return &AzureCLICredential{
-		tokenProvider: cp.tokenProvider,
-		tenantID:      cp.TenantID,
+		additionallyAllowedTenants: cp.AdditionallyAllowedTenants,
```
jhendrixMSFT, line 66: How come this one doesn't get the `resolveAdditionallyAllowedTenants()` treatment?

chlowell: I can't think of a reason for it not to, so I believe this is an oversight on the part of past me. Thanks for spotting it.

sdk/azidentity/azidentity.go
chlowell: Sure, that could help someone.

jhendrixMSFT (dismissed):

chlowell:
I changed base to `main` because azidentity betas are shipping from there and converted it to a draft pending the azcore beta release.

sdk/azidentity/environment_credential.go
```diff
@@ -75,9 +86,21 @@ func NewEnvironmentCredential(options *EnvironmentCredentialOptions) (*Environme
 	if clientID == "" {
 		return nil, errors.New("missing environment variable " + azureClientID)
 	}
+	// tenants set by NewDefaultAzureCredential() override the value of AZURE_ADDITIONALLY_ALLOWED_TENANTS
+	additionalTenants := options.additionallyAllowedTenants
+	if len(additionalTenants) == 0 {
```
jhendrixMSFT, line 91: Nit: Won't this always be true, given that the option isn't exported?

jhendrixMSFT (approved):

sdk/azidentity/environment_credential.go
chlowell: The option exists so `DefaultAzureCredentialOptions.AdditionallyAllowedTenants` can override the environment variable, preventing weirdness like the option having unexpected or inconsistent effects.

PR #19548, opened 2022-11-10
Update storage codeowners

jhendrixMSFT:
Remove legacy owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -44,7 +44,7 @@
 /sdk/messaging/azservicebus   @richardpark-msft @jhendrixMSFT
 
 # PRLabel: %Event Hubs
-/sdk/messaging/azeventhubs   @richardpark-msft @jhendrixMSFT
+/sdk/messaging/azeventhubs    @richardpark-msft @jhendrixMSFT
 
 # PRLabel: %Service Bus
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -56,7 +56,7 @@
 /sdk/resourcemanager/         @ArcturusZhang @lirenhe @tadelesh
 
 # PRLabel: %Storage
-/sdk/storage/                 @mohsha-msft @zezha-msft @jhendrixMSFT
+/sdk/storage/                 @siminsavani-msft @souravgupta-msft @jhendrixMSFT
```
siminsavani-msft, line 59: Could we also add tasherif-msft here please?

siminsavani-msft (changes requested):

.github/CODEOWNERS
jhendrixMSFT: Yes, is there a preferred order?

siminsavani-msft: Nope, there is no preference!

jhendrixMSFT: done

siminsavani-msft (approved):
Looks good! Thank you!

PR #19554, opened 2022-11-11
ACR data plane module

tadelesh:
This is the first preview version of ACR data plane package. Most majority of the operations is auto-generated. Customizations are done by autorest directive. Manual code includes auth policy, client CTOR and blob upload operations.


sdk/acr/azacr/build.go
```diff
@@ -0,0 +1,9 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+//go:generate autorest ./autorest.md
```
jhendrixMSFT, line 7: You should also add a line to format the code.
```
//go:generate gofmt -w .
```

jhendrixMSFT:
Should the path be something like this.
```
/sdk/containerregistry/azacr
```
IMO `acr` is too non-descript for the top-level path.

sdk/containers/azcontainerregistry/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry
+
+go 1.18
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.0.0
```
jhendrixMSFT, line 6: This should be updated to the latest.

sdk/containers/azcontainerregistry/go.mod
```diff
@@ -0,0 +1,27 @@
...
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.0
```
jhendrixMSFT, line 8: Also update to the latest (will be required for remote test assets).

jhendrixMSFT (changes requested):
New SDKs should use remote test assets and not check in the recordings.  Please see https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/documentation/asset-sync/README.md#im-a-dev-who-uses-the-test-proxy-currently-how-do-i-externalize-my-recordings for more details.

You can also see an example of this being enabled for `azqueue` in https://github.com/Azure/azure-sdk-for-go/pull/19756

tadelesh:
> New SDKs should use remote test assets and not check in the recordings. Please see https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/documentation/asset-sync/README.md#im-a-dev-who-uses-the-test-proxy-currently-how-do-i-externalize-my-recordings for more details.
> 
> You can also see an example of this being enabled for `azqueue` in #19756

Resolved.

tadelesh:
@jhendrixMSFT I've resolved all CI failure. Could you help to review again?

jhendrixMSFT (approved):

Issue #19539, opened 2022-11-10
Enhance changelog generation tool to make changelog content more meaningful


tadelesh:
Possible pattern:
1. enum alias
const: XXXValueA XXXValueB, type alias: XXX, function: PossibleXXXValues
-> Enum XXX
2. operation
function: Client.XXX, struct: ClientXXXOptions, ClientXXXResponse
Operation Client.XXX
3. LRO -> non-LRO
4. Other possible patterns

Try to make rule extensible and plan for the future.

tadelesh:
Please also help to sort changelog items in the meantime.

PR #19562, opened 2022-11-14
Changelog filter

Alancere:
Resolve: https://github.com/Azure/azure-sdk-for-go/issues/19539


ArcturusZhang (commented):
The changes look good, do we have some examples of how the chnagelog would change under this change?

eng/tools/generator/cmd/v2/common/testdata/new/enum/enumfileter.go
```diff
@@ -0,0 +1,15 @@
+package enumfilter_test
```
ArcturusZhang, line 1: This should be `package enum` because this is not a test file

eng/tools/generator/cmd/v2/common/testdata/new/func/funcfilter.go
```diff
@@ -0,0 +1,12 @@
+package funcfilter_test
```
ArcturusZhang, line 1: This should be `package func` (func is a keyword, maybe you need to rename this directory)

eng/tools/generator/cmd/v2/common/testdata/new/lro/lrofilter.go
```diff
@@ -0,0 +1,12 @@
+package lrofilter_test
```
ArcturusZhang, line 1: This should be `package lro`

ArcturusZhang (commented):
Some package in the testdata is not correct

eng/tools/generator/cmd/v2/common/testdata/old/enum/enumfileter.go
```diff
@@ -1,4 +1,4 @@
-package enumfilter_test
+package enumfilter
```
ArcturusZhang, line 1: package name should be aligned with the directory name
``` suggestion
package enum
```

eng/tools/generator/cmd/v2/common/testdata/old/lro/lrofilter.go
```diff
@@ -1,4 +1,4 @@
-package lrofilter_test
+package lrofilter
```
ArcturusZhang, line 1: package name should be aligned with the directory name
``` suggestion
package lro
```

eng/tools/generator/autorest/model/changelog.go
```diff
@@ -268,10 +268,23 @@ func getRemovedContent(removed *delta.Content) []string {
 	}
 	// write functions
 	if len(removed.Funcs) > 0 {
+		var lroItem []string
 		for _, k := range sortFuncItem(removed.Funcs) {
+			v := removed.Funcs[k]
+			if v.ReplacedBy != nil {
```
ArcturusZhang, line 274: out of curiosity, can we detected something is replaced by something else?

eng/tools/generator/cmd/v2/common/changelogProcessor.go
```diff
@@ -207,3 +210,136 @@ func removeMarshalUnmarshalFunc(funcs map[string]exports.Func) {
...
+				if strings.Contains(clientFunc[1], "Begin") {
+					clientFunc[1] = strings.ReplaceAll(clientFunc[1], "Being", "")
+					beginFunc = fmt.Sprintf("%s.%s", clientFunc[0], clientFunc[1])
+				} else {
+					beginFunc = fmt.Sprintf("%s.Begin%s", clientFunc[0], clientFunc[1])
+				}
+				if _, ok := changelog.Modified.AdditiveChanges.Funcs[beginFunc]; ok {
+					delete(changelog.Modified.AdditiveChanges.Funcs, beginFunc)
+					tmp := &beginFunc
+					v.ReplacedBy = tmp
```
ArcturusZhang, line 339: why we need a `tmp` here?

ArcturusZhang (approved):

eng/tools/generator/autorest/model/changelog.go
Alancere: lro to non-lro or non-lor to lro

Alancere: example: Operation `*AgentPoolsClient.AbortLatestOperation` has been changed to LRO, use `*ManagedClustersClient.BeginAbortLatestOperation` instead.

ReplacedBy save *ManagedClustersClient.BeginAbortLatestOperation

eng/tools/generator/cmd/v2/common/changelogProcessor.go
Alancere: removed

ArcturusZhang (commented):
Could we also add a test case for the new enum value accumulation?

Alancere:
> Could we also add a test case for the new enum value accumulation?

updated TestEnumFilter test

ArcturusZhang (approved):

PR #19581, opened 2022-11-16
[azservicebus] Removing the idle tracking code.

richardpark-msft:
Client-side idle tracking was added in v1.1.2, but it seems be causing issues for customers and is actually making their receiving _less_ reliable. I had previously removed it for just sessions, but I'm now also removing it for non-session receivers as well.

I'm leaving the idle check code, just not hooked up for now, because it still seems like a good defensive feature but it requires more testing and stressing to be fit for production.

This rolls back the changes in #19465 (simple idle timer), #19492 (ignore ctx on link close) and #19506 (track idle across multiple receives).


jhendrixMSFT (approved):

jhendrixMSFT (approved):

yulin-li:
I want to say the 1.1.2 makes our service down because this issue. Besides this fix, could you try to avoid using a path release for these big changes?

richardpark-msft:
> I want to say the 1.1.2 makes our service down because this issue. Besides this fix, could you try to avoid using a path release for these big changes?

Hi @yulin-li, you've got a good point about this change, although it was intended to simply complement the existing recovery code (but revealed a weak point). The next time I release this it'll go out as a minor rev bump.

PR #19598, opened 2022-11-18
Add LoginHint option for InteractiveBrowserCredential

### LINKED ISSUE #15599: login_hint support for InteractiveBrowserCredential
Seems sensible to finish #14920 first and let MSAL handle this but it's uncertain when that will be possible because MSAL doesn't yet allow setting login_hint (https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/271). Adding it to our own auth code flow would create a blocker for  transitioning to MSAL.

chlowell:
Closes #15599

### DIFF
--- a/sdk/azidentity/interactive_browser_credential.go
+++ b/sdk/azidentity/interactive_browser_credential.go
@@ -21,16 +21,18 @@ const credNameBrowser = "InteractiveBrowserCredentiall"
 type InteractiveBrowserCredentialOptions struct {
 	azcore.ClientOptions
 
-	// TenantID is the Azure Active Directory tenant the credential authenticates in. Defaults to the
-	// "organizations" tenant, which can authenticate work and school accounts.
-	TenantID string
 	// ClientID is the ID of the application users will authenticate to.
 	// Defaults to the ID of an Azure development application.
 	ClientID string
+	// LoginHint pre-populates the account prompt with a username. Users may choose to authenticate a different account.
+	LoginHint string
 	// RedirectURL is the URL Azure Active Directory will redirect to with the access token. This is required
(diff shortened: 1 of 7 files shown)


sdk/azidentity/interactive_browser_credential.go
```diff
@@ -73,11 +77,7 @@ func (c *InteractiveBrowserCredential) GetToken(ctx context.Context, opts policy
 		return azcore.AccessToken{Token: ar.AccessToken, ExpiresOn: ar.ExpiresOn.UTC()}, err
 	}
 
-	o := []public.InteractiveAuthOption{}
-	if c.options.RedirectURL != "" {
-		o = append(o, public.WithRedirectURI(c.options.RedirectURL))
-	}
-	ar, err = c.client.AcquireTokenInteractive(ctx, opts.Scopes, o...)
+	ar, err = c.client.AcquireTokenInteractive(ctx, opts.Scopes, public.WithLoginHint(c.options.LoginHint), public.WithRedirectURI(c.options.RedirectURL))
```
jhendrixMSFT, line 80: If `options.LoginHint` and `options.RedirectURL` are empty, these options are just no-op?

chlowell: Yes. The values end up as string fields on a struct, so an explicit empty string is the same as not setting these options. Is that weird? I'm happy to keep the conditionals if they prevent future maintainers from wondering about this.

jhendrixMSFT: I think it's fine and makes for simpler code on our side, assuming the contract for these APIs doesn't change on the MSAL side (seems unlikely and would show up in our unit testing).

jhendrixMSFT (approved):

PR #19601, opened 2022-11-21
retract sdk/resourcemanager/loadtestservice/armloadtestservice

Alancere:


sdk/resourcemanager/loadtestservice/armloadtestservice/go.mod
```diff
@@ -1,3 +1,4 @@
+// Deprecated: use github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/loadtesting/armloadtesting instead.
```
ArcturusZhang, line 1: Could we add a `please` here?
```suggestion
// Deprecated: please use github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/loadtesting/armloadtesting instead.
```

ArcturusZhang (approved):

PR #19608, opened 2022-11-22
Added parameter LRO in order to mark LRO

Alancere:


eng/tools/internal/exports/exports.go
```diff
@@ -54,6 +54,8 @@ type Func struct {
 
 	// a comma-delimited list of the return types
 	Returns *string `json:"returns,omitempty"`
+
```
tadelesh, line 57: Please add a comment about the param.

Alancere: comment: ”mark long running operation“ is it okay？

tadelesh: could you please provide some examples of this string?

Alancere: image

tadelesh (approved):

Issue #19612, opened 2022-11-23
azblob UploadStream produces panic: send on closed channel if input stream has error


ncw:
### Bug Report

If an error occurs on the input channel when using `Client.UploadStream` from `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob` then `UploadStream` panics in the background shortly **after** it returns the error.

Here is a 100% reliable reproducer - set `account` and `containerName` below and make sure your auth is in the environment.

```go
package main

import (
	"bytes"
	"context"
	"crypto/rand"
	"errors"
	"fmt"
	"io"
	"log"
	"time"

	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
)

const (
	account       = "https://XXXXX.blob.core.windows.net/"
	containerName = "XXXX"
	blobName      = "test.bin"
)

// ErrorReader wraps an error to return on Read
type ErrorReader struct {
	Err error
}

// Read always returns the error
func (er ErrorReader) Read(p []byte) (n int, err error) {
	return 0, er.Err
}

func main() {
	// authenticate with Azure Active Directory
	// this picks up the az login credentials
	cred, err := azidentity.NewDefaultAzureCredential(nil)
	if err != nil {
		log.Fatalf("auth failed: %v", err)
	}

	// create a client for the specified storage account
	client, err := azblob.NewClient(account, cred, nil)
	if err != nil {
		log.Fatalf("create client failed: %v", err)
	}

	var N int64 = 5 * 1024

	// Construct in so it reads N bytes then produce an error
	buf := make([]byte, N)
	_, _ = rand.Read(buf)
	in1 := bytes.NewReader(buf)
	er := &ErrorReader{Err: errors.New("potato")}
	in := io.MultiReader(in1, er)

	// It is necessary to pass in azblob.UploadStreamOptions to
	// trigger the bug.
	uploadStreamOptions := azblob.UploadStreamOptions{
		BlockSize:   4 * 1024 * 1024,
		Concurrency: 16,
	}

	fmt.Printf("Uploading blob %q, error after %d\n", blobName, N)

	// upload the file to the specified container with the specified blob name
	_, err = client.UploadStream(context.TODO(), containerName, blobName, in, &uploadStreamOptions)
	fmt.Printf("Result err=%v\n", err)

	// Note the panic happens in the background and we don't
	// always catch it unless we sleep here
[...]

Issue #19579, opened 2022-11-16
Controlling TransferManager in azblob 0.5.1


AsafMah:
Hey, we want to migrate from azblob 0.3 to 0.5.1.

For memory saving reasons, currently we expose a way for the client to decide if the transfermanager is "StaticBuffer" or "Syncpool".

It seems that in 0.5.1 though, TransferManager became an internal type, and it doesn't seem to be editable by the user.
https://github.com/Azure/azure-sdk-for-go/blob/72b362adee0e743c1cbb80870384044d0f855480/sdk/storage/azblob/blockblob/models.go

Is there a way for the user to control it? There are the concurrency and blocksize arguments, but they seem now to only instanciate a "staticbuffer" kind.

Thanks.

jhendrixMSFT:
When we did the refactor, this was intentional.  However, looks like we've regressed some functionality.

CC @JeffreyRichter

JeffreyRichter:
The Go SDK is the ONLY SDK that has a transfer manager. When I asked about its purpose, no one was able to explain why it was necessary. So, ideally, it would go away completely unless there is a great reason for it. To ship sooner and not think about it now, we made it internal.

jhendrixMSFT:
My understanding from looking at the code is the [`TransferManager`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/shared/transfer_manager.go#L18) interface is used to abstract different implementations for managing memory when uploading a blob, allowing apps to fine-tune how memory is allocated.  At present, it looks to only be used by [`copier`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blockblob/chunkwriting.go#L84).  We used to expose two different implementations, a [static buffer](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/shared/transfer_manager.go#L48) and [sync pool](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/shared/transfer_manager.go#L109) that customers could select, with the default being static buffer.

JeffreyRichter:
I found the publicly-exposed place where copyFromReader is called: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blockblob/client.go#L484

[...]

PR #19635, opened 2022-11-29
Use memory mapped files for streaming uploads

jhendrixMSFT:
This is to avoid extra allocations and heap fragmentation. Removed transfer manager and related code.

Fixes
- https://github.com/Azure/azure-sdk-for-go/issues/19579
- https://github.com/Azure/azure-sdk-for-go/issues/19612


sdk/storage/azblob/blockblob/chunkwriting.go
```diff
@@ -12,225 +12,306 @@ import (
...
 
-	// Send all our chunks until we get an error.
-	for {
-		if err = cp.sendChunk(); err != nil {
+	// This goroutine grabs a buffer, reads from the stream into the buffer,
+	// then creates a goroutine to upload/stage the block.
+	for blockNum := uint32(0); true; blockNum++ {
+		var buffer T
+		select {
+		case buffer = <-buffers.Acquire():
```
jhendrixMSFT, line 87: @JeffreyRichter I must point out that this algorithm will use _up to_ the number of gorutines as specified by `options.Concurrency` but it _does not guarantee_ that this number of goroutines will be used.  In this respect, I think the semantics of `options.Concurrency` are misleading.

jhendrixMSFT: In fact, the docs are pretty much wrong.
```go
// Concurrency defines the number of concurrent uploads to be performed to upload the file.
// Each concurrent upload will create a buffer of size BlockSize.  The default value is one.
```
If this is the algorithm we want, I'll need to make some adjustments.

jhendrixMSFT: It would also be interesting to measure both implementations to see how much of a difference there is in terms of throughput.

sdk/storage/azblob/CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
 
 ### Other Changes
 
-* Avoid allocations when storing/fetching a slice via sync.Pool.
+* The `UploadStream()` methods now use memory mapped files for buffers in order to reduce heap allocations/fragmentation.
```
vibhansa-msft, line 25: Few concerns here with memory mapping:
1. At process level we also have limitations on how many files we can memory map at a given time. If user application does too many upload/downloads in parallel that will be an issue here.
2. Memory mapping a file will increase our memory usage and we may put system under resource crunch with memory mapping huge files for upload/download.

JeffreyRichter: A memory mapping doesn't require a file descriptor, so these don't count towards the process' quota. Also, these are anonymous memory mappings so no for descriptor is even used at all. Also, we're not doing any file uploads or downloads here. We're just creating the anonymous memory mappings for this upload stream method. The number and size of the memory maps buffers is determined by the customer options so they are controlled directly by the application developer.

jhendrixMSFT: I've updated the text here to clarify that this uses anonymous memory mapped files.

jhendrixMSFT: Also, regarding point 2 above, the total memory allocated is identical compared to using byte slices (sans any differences in overhead imposed by the OS).  The difference is that the anonymous memory mapped files aren't part of the GC heap, so using them reduces GC pressure/fragmentation.

sdk/storage/azblob/blockblob/chunkwriting.go
```diff
@@ -12,225 +12,306 @@ import (
...
-	defer atomic.AddUint32(&id.num, 1)
+func (pool *mmbPool) Acquire() <-chan mmb {
+	return pool.buffers
+}
 
-	binary.BigEndian.PutUint32(id.u[len(uuid.UUID{}):], atomic.LoadUint32(&id.num))
-	str := base64.StdEncoding.EncodeToString(id.u[:])
-	id.all = append(id.all, str)
+func (pool *mmbPool) Grow() (int, error) {
+	if pool.count < pool.max {
```
JeffreyRichter, line 296: This doesn't need to be thread safe because Grow is only ever called from the 1 goroutine calling UploadStream, right?

sdk/storage/azblob/blockblob/mmf_unix.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"syscall"
+)
+
+// mmb is a memory mapped buffer
+type mmb []byte
+
+// newMMB creates a new memory mapped buffer with the specified size
+func newMMB(size int64) (mmb, error) {
+	prot, flags := syscall.PROT_READ|syscall.PROT_WRITE, syscall.MAP_ANONYMOUS|syscall.MAP_SHARED
```
JeffreyRichter, line 21: You don't want MAP_SHARED

sdk/storage/azblob/blockblob/mmf_unix.go
```diff
@@ -0,0 +1,36 @@
...
+	addr, err := syscall.Mmap(-1, 0, int(size), prot, flags)
+	if err != nil {
+		return nil, os.NewSyscallError("Mmap", err)
+	}
+	return mmb(addr), err
```
JeffreyRichter, line 26: I'd probably hard code nil for err here instead of using the err variable.

sdk/storage/azblob/blockblob/mmf_unix.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// delete cleans up the memory mapped buffer
+func (m *mmb) delete() {
+	err := syscall.Munmap(*m)
+	*m = nil
+	if err != nil {
+		panic(fmt.Sprintf("fatal error unmapping mmb: %v", err))
```
JeffreyRichter, line 34: Bold having panic here. I'm OK with this.

sdk/storage/azblob/blockblob/mmf_windows.go
```diff
@@ -0,0 +1,52 @@
...
+	"syscall"
+	"unsafe"
+)
+
+// mmb is a memory mapped buffer
+type mmb []byte
+
+// newMMB creates a new memory mapped buffer with the specified size
+func newMMB(size int64) (mmb, error) {
+	const InvalidHandleValue = ^uintptr(0)
```
JeffreyRichter, line 22: This should be -1; not 0

JeffreyRichter (approved):

sdk/storage/azblob/blockblob/chunkwriting.go
jhendrixMSFT: This was originally `MaxConcurrency` however we changed it due to arch board feedback to align with other languages.  I personally have no issues with changing it, I only bring this up for context.

sdk/storage/azblob/blockblob/chunkwriting.go
jhendrixMSFT: Correct.  I've added a comment to explain this.

sdk/storage/azblob/blockblob/mmf_windows.go
jhendrixMSFT: The `^` inverts the bits (see also [here](https://github.com/golang/go/blob/master/src/runtime/netpoll_windows.go#L14)).

sdk/storage/azblob/blockblob/chunkwriting.go
```diff
@@ -12,225 +12,306 @@ import (
...
+func copyFromReader[T ~[]byte](ctx context.Context, src io.Reader, dst blockWriter, options UploadStreamOptions, getBufferManager func(maxBuffers int, bufferSize int64) bufferManager[T]) (CommitBlockListResponse, error) {
+	options.setDefaults()
+
+	wg := sync.WaitGroup{}       // Used to know when all outgoing blocks have finished processing
+	errCh := make(chan error, 1) // contains the first error encountered during processing
+
+	buffers := getBufferManager(options.Concurrency, options.BlockSize)
+	defer buffers.Free()
+
+	// create initial buffer
```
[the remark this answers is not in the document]
jhendrixMSFT: Good point, removed.

sdk/storage/azblob/blockblob/mmf_unix.go
jhendrixMSFT: Fixed.

sdk/storage/azblob/blockblob/mmf_unix.go
jhendrixMSFT: That's what I thought too.  However, I had failure on Linux without it.  I will take another look.

sdk/storage/azblob/blockblob/mmf_unix.go
jhendrixMSFT: This was copied from track 1.5.  I'm on the fence about this, but I don't know how "fatal" failing to unmap the buffer is.  I suppose if we surface this as an `os.SyscallError` then callers _should_ be savvy enough to treat this as fatal.

jhendrixMSFT: Also, it makes it hard to reason about the API.  If we returned an error in this case, but the blob was uploaded, how is a customer supposed to figure this out?

sdk/storage/azblob/blockblob/mmf_unix.go
jhendrixMSFT: Oops my mistake, I had omitted it entirely.  We want `MAP_PRIVATE`.

sdk/storage/azblob/blockblob/mmf_windows.go
JeffreyRichter: I missed the ^. How about a comment // -1

sdk/storage/azblob/blockblob/chunkwriting.go
JeffreyRichter: I don't feel strongly one way or the other.  I don't think that a customer specifying concurrency expects that it **has to be** used and can't be less.

sdk/storage/azblob/blockblob/mmf_unix.go
JeffreyRichter: For the record, I'd also be OK with _ = syscall.Mumap(*m) and just throw the error away. In Win32 no one ever checked the true/false return from calling CloseHandle.

jhendrixMSFT: I suppose that in the event this happens, it manifests as a memory leak and wouldn't cause actual memory corruption.  Perhaps it would be better to just log that this happened and move on instead.

JeffreyRichter: How would we log it? I think we either keep the panic or make this method return an error (just like how Win32's CloseHandle returns a Boolean). The caller can then decide to propogate it or ignore it.

jhendrixMSFT: Through our SDK logging facility.  I already made this change, take a look to see what you think.

vibhansa-msft (approved):

sdk/storage/azblob/blockblob/mmf_unix.go
jhendrixMSFT: I can switch it to returning an error, but before I do that, we need to decide how to message this to the caller.  In this case, the upload succeeded, we just failed to clean up the buffer(s).  And since customers pay for data transfers, I think it would be disingenuous to present the error in such a way that they think they'd need to redo the upload.  So, we'd have this corner-case where the caller receives an `os.SyscallError` _but_ the operation actually succeeded.

jhendrixMSFT: You've convinced me that `panic` is the best choice for now, I've switched it back.

PR #19645, opened 2022-11-30
Prevent ExpiryType from being publicly implementable

jhendrixMSFT:
The type is exported but is not meant for external implementation.

### DIFF
--- a/sdk/storage/azblob/internal/exported/set_expiry.go
+++ b/sdk/storage/azblob/internal/exported/set_expiry.go
@@ -7,16 +7,18 @@
 package exported
 
 import (
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 	"net/http"
 	"strconv"
 	"time"
+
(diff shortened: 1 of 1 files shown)


sdk/storage/azblob/internal/exported/set_expiry.go
```diff
@@ -7,16 +7,18 @@
...
 	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 )
 
 // ExpiryType defines values for ExpiryType
 type ExpiryType interface {
 	Format(o *SetExpiryOptions) (generated.ExpiryOptions, *generated.BlobClientSetExpiryOptions)
+	notPubliclyImplementable()
```
vibhansa-msft, line 21: This is part of "internal" package so this will not be exposed outside for public use anyway. Is there any problem you see with the current implementation?

jhendrixMSFT: The convention is that types in the `internal/exported` package are exported as public surface area (unless comments indicate otherwise).  In this case, `ExpiryType` is exported from the `appendblob` and `blockblob` packages.

PR #19668, opened 2022-12-06
[azservicebus] Stress test improvements

richardpark-msft:
- Fixing the panic that happens if you're not using a topic!
- Adding in a chaos run for infiniteSendAndReceive that can be used as an example for others.


richardpark-msft:
(fixed a rather nagging issue with logging that would sometimes cause panics)

jhendrixMSFT (approved):

Issue #19614, opened 2022-11-23
azblob: blob.StartCopyFromURL doesn't work with UTF-8 characters in the source blob


ncw:
### Bug Report

If I attempt to call `blob.StartCopyFromURL` from `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob` on a source blob with UTF-8 characters then the call fails with the error.

```
--------------------------------------------------------------------------------
RESPONSE 400: 400 The metadata specified is invalid. It has characters that are not permitted.
ERROR CODE: InvalidMetadata
--------------------------------------------------------------------------------
﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidMetadata</Code><Message>The metadata specified is invalid. It has characters that are not permitted.
RequestId:ad893aa4-c01e-0073-6858-fff064000000
Time:2022-11-23T16:28:54.8170886Z</Message></Error>
--------------------------------------------------------------------------------
```

I think this is because the output of `blob.Client().URL()` is not correct. It might be that there is a better way of constructing the source URL for `StartCopyFromURL` - I don't know, and if there is, it should be in the docs!

Here is a reproducer - you will need to change `account` and `containerName` and make sure there are some credentials in the environment

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"log"

	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
)

const (
	account       = "https://XXXX.blob.core.windows.net/"
	containerName = "XXXX"
	// This is OK
	//blobName = `hello sausage/Hello/ " ' @ < > & +z.txt`
	// This fails
	blobName    = `世界.txt`
	newBlobName = "copytest2.bin"
)

func main() {
	ctx := context.TODO()

	// authenticate with Azure Active Directory
	// this picks up the az login credentials
	cred, err := azidentity.NewDefaultAzureCredential(nil)
	if err != nil {
		log.Fatalf("auth failed: %v", err)
	}

	// create a client for the specified storage account
	client, err := service.NewClient(account, cred, nil)
	if err != nil {
		log.Fatalf("create client failed: %v", err)
	}

	// Make data to upload
	var N int64 = 1024
	buf := make([]byte, N)
	in := bytes.NewReader(buf)

	fmt.Printf("Uploading blob %q, size %d\n", blobName, N)

	cnt := client.NewContainerClient(containerName)
[...]

Issue #19613, opened 2022-11-23
azblob: when using SharedKey credentials, can't reference some blob names with ? in


ncw:
### Bug Report

When using SharedKey credentials, some blob names cause  this error

```
RESPONSE 403: 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
ERROR CODE: AuthenticationFailed
```

Here is an example program to demonstrate - you'll need to fill in `account`, `key` and `containerName` to try it.

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
)

const (
	account    = "XXXX"
	serviceURL = "https://" + account + ".blob.core.windows.net/"
	key        = "XXXX"

	containerName = "XXXX"
	blobNameGood  = `hello? sausage/hello.txt`
	blobNameBad   = `hello? sausage/Hello.txt`
)

func main() {
	ctx := context.Background()

	// Make shared key cred
	cred, err := service.NewSharedKeyCredential(account, key)
	if err != nil {
		log.Fatalf("create new shared key credential failed: %w", err)
	}

	// Make client from that cred
	client, err := service.NewClientWithSharedKeyCredential(serviceURL, cred, nil)
	if err != nil {
		log.Fatalf("create client with shared key failed: %w", err)
	}

	container := client.NewContainerClient(containerName)

	// now get the blob properties
	// This works OK
	blb := container.NewBlobClient(blobNameGood)
	_, err = blb.GetProperties(ctx, nil)
	fmt.Printf("OK err = %v\n", err)

	// now get the blob properties
	// And this fails
	blb = container.NewBlobClient(blobNameBad)
	_, err = blb.GetProperties(ctx, nil)
	fmt.Printf("BAD err = %v\n", err)
}
```

When run this attempts to HEAD two objects. It doesn't matter whether they exist or not. The first HEAD request returns a `404` error as the blob does not exist. The second HEAD requests returns a 403 request stating that the `Authorization` header is formed incorrectly.

This is fine for the blob named `hello? sausage/hello.txt`

```
OK err = HEAD https://rclone.blob.core.windows.net/rclone/hello
--------------------------------------------------------------------------------
RESPONSE 404: 404 The specified blob does not exist.
ERROR CODE: BlobNotFound
--------------------------------------------------------------------------------
Response contained no body
--------------------------------------------------------------------------------
```

[...]

Issue #19475, opened 2022-11-02
Azure Blob Storage paths are not URL-escaped


dgolub:
### Bug Report

Import path: `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container`
SDK version: `v0.5.1`
Go version: `go1.18.5 windows/amd64`

When opening a blob on Azure Blob Storage with a path that contains a special character (e.g. `foo%5Cbar`), the path doesn't get properly URL-escaped.  It comes back with a not found error, even though a blob with the specified name actually exists.

For example, this fails:
```
blobClient := containerClient.NewBlobClient("foo%5Cbar")
```

But this succeeds:
```
path := url.PathEscape("foo%5Cbar")
blobClient := containerClient.NewBlobClient(path)
```

siminsavani-msft:
Hi @dgolub ! Thank you for reporting an issue! I just wanted to clarify that in your scenario, the blobname is 'foo%5Cbar', is this correct?

dgolub:
Yes, that's correct.

siminsavani-msft:
@dgolub Thank you for your prompt reply! We are working on this and will update this thread when we have additional updates.

siminsavani-msft:
Hi @dgolub ! I am not able to reproduce. Could you please share the failure/error you are getting?

ncw:
I'll just note that the rclone test suite picks up this bug also @siminsavani-msft - its a bug which has been present in Onedrive in the past too! Though I think it is an SDK bug this time since the original Azure blob SDK `github.com/Azure/azure-storage-blob-go/azblob` does **not** have this problem.

jhendrixMSFT:
The fix will roll out in `v0.6.0`.

PR #19670, opened 2022-12-07
URL-encode blob name parameters

jhendrixMSFT:
Fixes

- https://github.com/Azure/azure-sdk-for-go/issues/19475
- https://github.com/Azure/azure-sdk-for-go/issues/19613
- https://github.com/Azure/azure-sdk-for-go/issues/19614

### DIFF
--- a/sdk/storage/azblob/service/client.go
+++ b/sdk/storage/azblob/service/client.go
@@ -120,11 +120,8 @@ func (s *Client) URL() string {
 	return s.generated().Endpoint()
 }
 
-// NewContainerClient creates a new ContainerClient object by concatenating containerName to the end of
-// Client's URL. The new ContainerClient uses the same request policy pipeline as the Client.
-// To change the pipeline, create the ContainerClient and then call its WithPipeline method passing in the
-// desired pipeline object. Or, call this package's NewContainerClient instead of calling this object's
-// NewContainerClient method.
+// NewContainerClient creates a new container.Client object by concatenating containerName to the end of
+// this Client's URL. The new container.Client uses the same request policy pipeline as the Client.
 func (s *Client) NewContainerClient(containerName string) *container.Client {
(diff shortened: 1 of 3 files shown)


sdk/storage/azblob/service/client.go
```diff
@@ -120,11 +120,8 @@ func (s *Client) URL() string {
...
 }
 
-// NewContainerClient creates a new ContainerClient object by concatenating containerName to the end of
-// Client's URL. The new ContainerClient uses the same request policy pipeline as the Client.
-// To change the pipeline, create the ContainerClient and then call its WithPipeline method passing in the
-// desired pipeline object. Or, call this package's NewContainerClient instead of calling this object's
-// NewContainerClient method.
+// NewContainerClient creates a new container.Client object by concatenating containerName to the end of
+// this Client's URL. The new container.Client uses the same request policy pipeline as the Client.
 func (s *Client) NewContainerClient(containerName string) *container.Client {
```
JeffreyRichter, line 125: We need to do the same for container names too for consistency. Look at tiny blob methods too. Some of those methods take both a container & blob name

JeffreyRichter (approved):

siminsavani-msft (approved):

sdk/storage/azblob/service/client.go
jhendrixMSFT: @seanmcc-msft can you comment on this?  From the docs, I thought there are more stringent requirements on container names.  That said, if .NET is doing this, we can follow suit.

JeffreyRichter: I don't think you can do this. It is a breaking change to alter this behavior. URL Endpoint must be encoded by the customer before passing them to our ctor. But, methods that take high-level abstractions (like container or blob) we'll encode them before putting them into a URL - I think we just have to do this.

Issue #19673, opened 2022-12-07
[azeventhubs/azservicebus] User agent strings are missing runtime information and have an incorrect leading slash


richardpark-msft:
The user agent strings for azeventhubs and azservicebus are formatted incorrectly or missing information. The HTTP SDK's headers look like this:

HTTP:
```
User-Agent: azsdk-go-azidentity/v1.2.0 (go1.19.3; linux)
User-Agent: azsdk-go-azsecrets/v0.11.0 (go1.19.3; linux)
```

And SB and EH:
```
USER AGENT: /azsdk-go-azservicebus/v1.1.4
```
We need to add in the OS section, and _remove_ the leading '/' (no other SDKs appear to have it).

PR #19681, opened 2022-12-08
[internal] Export function to format user agent data so we can share it between non-HTTP and HTTP clients.

richardpark-msft:
Export user agent formatting code that used to be in azcore's policy_telemetry.go so it can be shared with non-HTTP clients (ie: azservicebus/azeventhubs).

Part of a fix for #19673


richardpark-msft:
NOTE: there'll be separate PRs to move azcore and azservicebus/azeventhubs to using this function for their user agents.

jhendrixMSFT (approved):

PR #19683, opened 2022-12-09
[azeventhubs] Fixing issues with connection and link robustness for Event Hubs

richardpark-msft:
Fixing issues with connection and link robustness for Event Hubs.
    
There were some issues with how we do recovery, which could make things
hang or even leave us in an inconsistent state.

1. We didn't have any bound on how long a Close() could take, but since
that requires a network hop it's possible for it to hang infinitely if
the service doesn't respond. This now has a max bound of 60 seconds.

2. If a Close() of a link is cancelled we signal that the connection needs
to be reset, since not doing that will leave us in an inconsistent state
where we believe a link is active, but the service does not.

3. We now ignore errors that come back from trying to close a link (except
for the previously mentioned cancellation). The error that comes back will
be redundant (ie, it'll be the same error that initiated the recovery in
the first place) or it'll be something we can't immediately troubleshoot
and will be dealt with on the next recovery round.

There are also a number of testing quality fixes that I made to validate
this change. I've got gomock now for mocks in newer tests (and will start
to phase out the previous uses of my custom mocks). I've also added some
more tests around recovery and improved logging to log more context per
line so it's easier to correlate activity between links.


jhendrixMSFT:
If I understand correctly, we add a timeout via context to the close operation, and if close times out (or is cancelled), then we consider the `amqp.Conn` to be in a bad state and recreate it.  Did I get that right?

sdk/messaging/azeventhubs/doc.go
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
jhendrixMSFT, line 2: Nit: Copyright header should be after build tags.

sdk/messaging/azeventhubs/consumer_client_internal_test.go
```diff
@@ -41,6 +41,7 @@ func TestConsumerClient_Recovery(t *testing.T) {
 
 	// trim the partition list down so the test executes in resonable time.
 	ehProps.PartitionIDs = ehProps.PartitionIDs[0:3] // min for testing is 3 partitions anyways
+	//ehProps.PartitionIDs = ehProps.PartitionIDs[0:1] // min for testing is 3 partitions anyways
```
jhendrixMSFT, line 44: Cruft?

sdk/messaging/azeventhubs/internal/utils/retrier.go
```diff
@@ -34,7 +34,7 @@ func (rf *RetryFnArgs) ResetAttempts() {
 
 // Retry runs a standard retry loop. It executes your passed in fn as the body of the loop.
 // It returns if it exceeds the number of configured retry options or if 'isFatal' returns true.
-func Retry(ctx context.Context, eventName log.Event, operation string, o exported.RetryOptions, fn func(ctx context.Context, callbackArgs *RetryFnArgs) error, isFatalFn func(err error) bool) error {
+func Retry(ctx context.Context, eventName log.Event, prefix func() string, o exported.RetryOptions, fn func(ctx context.Context, callbackArgs *RetryFnArgs) error, isFatalFn func(err error) bool) error {
```
jhendrixMSFT, line 37: Why does this need to be a func now?

richardpark-msft: So it can be dynamic - for instance, if we rotate out the link and connection ID in between retries it's good to update the logging prefix as well.

richardpark-msft: Example here: https://github.com/Azure/azure-sdk-for-go/pull/19683/files/bd9cc2a0e0f4aeb3dccb3f0a754681431bab0079#diff-5f99975d3c585da6e7dfb4726eca7c769eff91f91d10eaa57a0e6c69d12e30f0R170

sdk/messaging/azeventhubs/internal/cbs.go
```diff
@@ -23,7 +24,7 @@ const (
 )
 
 // NegotiateClaim attempts to put a token to the $cbs management endpoint to negotiate auth for the given audience
-func NegotiateClaim(ctx context.Context, audience string, conn amqpwrap.AMQPClient, provider auth.TokenProvider) error {
+func NegotiateClaim(ctx context.Context, audience string, conn amqpwrap.AMQPClient, provider auth.TokenProvider, contextWithTimeoutFn contextWithTimeoutFn) error {
```
jhendrixMSFT, line 27: What's the purpose of `contextWithTimeoutFn`?

richardpark-msft: Will do on the next push to this PR.

richardpark-msft: (done)

sdk/messaging/azeventhubs/doc.go
richardpark-msft: Done.

sdk/messaging/azeventhubs/consumer_client_internal_test.go
richardpark-msft: Yeah, I was debugging something. Removed.

sdk/messaging/azeventhubs/internal/links.go
```diff
@@ -264,7 +325,8 @@ func (l *Links[LinkT]) newLinkState(ctx context.Context, partitionID string) (*l
 	session, connID, err := l.ns.NewAMQPSession(ctx)
 
 	if err != nil {
-		_ = ls.Close(ctx)
+		azlog.Writef(exported.EventConn, "(%s): Failed to create AMQP session for partition ID '%s': %s", ls.String(), partitionID, err)
+		err = ls.Close(ctx)
```
jhendrixMSFT, line 330: This will overwrite the last error.  I'm guessing this is intentional but wanted to confirm.

sdk/messaging/azeventhubs/internal/links.go
```diff
@@ -274,16 +336,18 @@ func (l *Links[LinkT]) newLinkState(ctx context.Context, partitionID string) (*l
 	tmpLink, err := l.newLinkFn(ctx, session, l.entityPathFn(partitionID))
 
 	if err != nil {
+		azlog.Writef(exported.EventConn, "(%s): Failed to create link for partition ID '%s': %s", ls.String(), partitionID, err)
 		_ = ls.Close(ctx)
```
jhendrixMSFT, line 341: And here we _don't_ overwrite the last error?

sdk/messaging/azeventhubs/internal/links.go
richardpark-msft: I believe it's unintentional - I think we want the original error, not the Close() error. Will switch.

sdk/messaging/azeventhubs/internal/links.go
richardpark-msft: Right, I think the non-overwrite behavior is better. Leaving this one, changing the other to match the perfect model this one has provided.

sdk/messaging/azeventhubs/internal/links.go
```diff
@@ -371,9 +451,7 @@ func (l *Links[LinkT]) closePartitionLinkIfMatch(ctx context.Context, partitionI
 		return nil
 	}
 
-	current.cancelAuth()
```
jhendrixMSFT, line 374: Was this removed because it was too eagerly closing the auth goroutine?

richardpark-msft: No, although it was probably harmless to leave it in.

Just below that call is a call to `current.Close()` which will cancel the authentication, in the same way. So it was just redundant.

```go
func (ls *linkState[LinkT]) Close(ctx context.Context) error {
	if ls.cancelAuth != nil {
		ls.cancelAuth()
	}
```

jhendrixMSFT (approved):

PR #19712, opened 2022-12-15
[azeventhubs+azservicebus] Updating our user agent format so it matches HTTP

### LINKED ISSUE #19673: [azeventhubs/azservicebus] User agent strings are missing runtime information and have an incorrect leading slash

richardpark-msft:
User-Agent was formatted inconsistent with our HTTP-based clients which was making doing simple reporting difficult. 

This brings it inline with what we do for user agents in HTTP, including when a client specified appid is present.

Examples:
* No appID: `azsdk-go-azeventhubs/v0.3.1 (go1.19.3; linux)`
* With AppID: `userApplicationID azsdk-go-azeventhubs/v0.3.1 (go1.19.3; linux)`

Fixes #19673


sdk/messaging/azeventhubs/internal/namespace.go
```diff
@@ -457,7 +456,7 @@ func (ns *Namespace) GetEntityAudience(entityPath string) string {
 func (ns *Namespace) getUserAgent() string {
 	userAgent := rootUserAgent
 	if ns.userAgent != "" {
```
jhendrixMSFT, line 458: So `ns.userAgent` is the user-provided application ID?

richardpark-msft: Yeah, I never changed the internal field name but it maps from this exposed option: 
* (option) https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azeventhubs/consumer_client.go#L46
* (copying) https://github.com/Azure/azure-sdk-for-go/blob/1ad22a964ea16c159e7e3f28263c940667d61519/sdk/messaging/azeventhubs/consumer_client.go#L249

jhendrixMSFT (approved):

PR #19718, opened 2022-12-20
Changelog removed consts print

Alancere:


eng/tools/generator/autorest/model/changelog.go
```diff
@@ -283,8 +283,26 @@ func getRemovedContent(removed *delta.Content) []string {
...
+				}
+			}
+		}
+
+		for _, k := range sortChangeItem(removedConst) {
+			consts := ""
+			for _, cs := range removedConst[k] {
+				consts = fmt.Sprintf("%s`%s`, ", consts, cs)
+			}
+			line := fmt.Sprintf("Const %s from type alias `%s` has been removed", strings.TrimRight(strings.TrimSpace(consts), ","), k)
```
ArcturusZhang, line 305: Could we bulk the values removed from the same type alias?

ArcturusZhang (commented):
Also please add a test case for this new pattern

ArcturusZhang (approved):

PR #19732, opened 2022-12-28
[Feature] Queue Service Client + Key Credential Authentication

tasherif-msft:
Setting up tests will be in a separate PR based off of this branch.


sdk/storage/azqueue/service/client.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/http"
+)
+
+// ClientOptions contains the optional parameters when creating a Client.
+type ClientOptions struct {
+	azcore.ClientOptions
+}
+
+// Client represents a URL to the Azure Queue Storage service allowing you to manipulate queues.
+type Client base.Client[generated.ServiceClient]
```
jhendrixMSFT, line 21: Sorry if this has already been discussed, but can we not directly expose the generated code for queues?  My understanding is that the public surface area is quite small/simple so the generated code should just work.

CC @JeffreyRichter

sdk/storage/azqueue/service/models.go
```diff
@@ -0,0 +1,122 @@
...
+	HourMetrics *Metrics
+
+	// Azure Analytics Logging settings.
+	Logging *Logging
+
+	// a summary of request statistics grouped by API in hour or minute aggregates for queues
+	MinuteMetrics *Metrics
+}
+
+func (o *SetPropertiesOptions) format() (generated.StorageServiceProperties, *generated.ServiceClientSetPropertiesOptions) {
```
gapra-msft, line 89: Why are we returning *generated.ServiceClientSetPropertiesOptions if we always return nil? Is this for future use for another method?

tasherif-msft: currently its set to nil, this might change in the future for one reason or another. Rather than directly passing nil to the generated layer just abstract it here and change implementation here in the future if something changes.

sdk/storage/azqueue/service/client.go
```diff
@@ -0,0 +1,148 @@
+package service
```
jhendrixMSFT, line 1: Why is this in a `service` sub-package?  Maybe a better question is, what packages are intended to be part of this module, and what purposes do they serve?

tasherif-msft: Similar to blobs, queue has a service/account client https://learn.microsoft.com/en-us/rest/api/storageservices/operations-on-the-account--queue-service- -> this link contains operations on account level.
There will be two clients, service and queue, each with their respective sub-package.

sdk/storage/azqueue/service/client.go
tasherif-msft: this is pattern matching blobs

sdk/storage/azqueue/service/client.go
jhendrixMSFT: We did this for `azblob` due to the size/complexity of the public surface area.  My understanding is that queues is much smaller/simpler.  Do we need to separate the two?

sdk/storage/azqueue/service/client.go
jhendrixMSFT: Blobs is a special case due to the size and complexity of the public surface area.

sdk/storage/azqueue/service/client.go
tasherif-msft: we separate them in other languages as well (python, Java and .NET). Technically we're already merging one extra client (the MessageClient) into QueueClient, so I think having ServiceClient and QueueClient is reasonable.

JeffreyRichter: I think it best to have both clients ServiceClient and QueueClient in a single azqueue package.

sdk/storage/azqueue/service/client.go
JeffreyRichter: Joel, I think you mean you WANT to expose the code gen directly, right? (your first comment says "can we not directly expose..."). I do think merging the MessageClient into QueueClient is a good thing to do and I don't think the code gen will accomplish this, right? If it doesn' then wrapping may be necessary.

jhendrixMSFT: Correct, I was hoping we can expose the generated code and not wrap (which is always our default position).  If we can't for reasons, it would be great to enumerate them, so everybody is on the same page.

JeffreyRichter: Yes, I agree with you. In this case, the merging of clients may be the reason for wrapping the code-gen.

jhendrixMSFT: Is it as simple as the APIs being on the same client?  If so, we can fix this via an autorest transform.  I suspect there's more to it than this though.

JeffreyRichter: I dont' know all the things that transforms are capable of.  The URL for messages (https://myaccount.queue.core.windows.net/myqueue/messages)  is different from the URL for the queue (https://myaccount.queue.core.windows.net/myqueue) and so the code gen would have to append the "/messages" part to the queue URL. And the method names in swagger are not great (like PUT message instead of Enqueu, for example) but I think this can be fixed with a transform.

seanmcc-msft: We have a `QueueClient` and `QueueServiceClient` for the queues package in .NET, Java, Python, and JavaScript.  Why would we diverge from this pattern in Go?  I think it would be more confusing for the customer if we merge them.

JeffreyRichter: No, we code gen a MessageClient and this is what we want to merge into QueueClient so there are 2 clients; not 3.

seanmcc-msft: Makes sense, we don't have a MessageClient on the other platforms either.

sdk/storage/azqueue/internal/exported/shared_key_credential.go
```diff
@@ -0,0 +1,218 @@
+//go:build go1.18
```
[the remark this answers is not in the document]
tasherif-msft: yes shared key cred and account level sas is basically the same!

sdk/storage/azqueue/go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue
 
 go 1.19
```
jhendrixMSFT, line 3: This should be `go 1.18`

sdk/storage/azqueue/service/client.go
tasherif-msft: I don't think it's necessary to have transforms. Similar to the blob package, we can create a composite client and store an inner client of `messages` alongside `queues`. but unlike blobs there will not be user facing APIs for messages. Instead, the composite client will just directly call the messages generated code as well as the queues generated code. This will all be in a separate PR of course. 

TLDR; we will use the wrapper, which will wrap two clients service client and queue client (composite of queue, message_ID, messages)

In the meantime, if we're good with the service client please approve so I can build on top of it :) @jhendrixMSFT @JeffreyRichter

sdk/storage/azqueue/go.mod
tasherif-msft: done!

sdk/storage/azqueue/go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue
```
jhendrixMSFT, line 1: Please be sure to run `go mod tidy` and include the generated `go.sum` file.

sdk/storage/azqueue/service/client.go
jhendrixMSFT: @tasherif-msft does this mean that the merging of the clients means this one will change in a future PR and the `service` package goes away?

tasherif-msft: it will not change. It just means there will be a new package called queue (in addition to service) that will also be wrapped and will include messages and queue ops.

jhendrixMSFT: How large is the public surface area for each?  Are separate packages really required?  It adds an amount of complexity, both on us and the consumer, so we should be sure that the tradeoff is worth it.

sdk/storage/azqueue/go.mod
tasherif-msft: done and sum file is already in there

jhendrixMSFT: Oh yes some files have been previously committed, my mistake.

sdk/storage/azqueue/service/client.go
tasherif-msft: Yes, they are required to be separate, this has always been the standard on all languages, Python, Java and .NET the packages mentioned above are exactly the same pattern the other languages are following. Design wise its also preferable to separate account level operations from the queue level operations hence why it's done in all languages and all storage services.

tasherif-msft: offline discussion with @jhendrixMSFT: will go with keeping everything in queue package and having separate clients

sdk/storage/azqueue/queue/client.go
```diff
@@ -0,0 +1,148 @@
+package queue
```
jhendrixMSFT, line 1: These files would go directly in the `azqueue` package, no need for a `queue` package.

Also, this file, and `models.go`, are missing the build constraint and copyright header comments.

sdk/storage/azqueue/service_client.go
```diff
@@ -0,0 +1,154 @@
+//go:build go1.18
```
jhendrixMSFT, line 1: Nit: I'd name the file simply `client.go` so it matches the name of the client.  The other clients would have snake-cased file names to match.

jhendrixMSFT (approved):

sdk/storage/azqueue/service_client.go
tasherif-msft: done

Issue #19699, opened 2022-12-13
Requests to the Azure metadata service take minutes to time out


johananl:
### Bug Report

- import path of package in question: `"github.com/Azure/azure-sdk-for-go/sdk/azidentity"`
- SDK version: `v1.2.0`
  - Specify the exact commit if possible; one way to get this is the REVISION
    column output by `dep status "github.com/Azure/azure-sdk-for-go`.
- output of `go version`: `go version go1.19.1 linux/amd64`

- What happened?

When using a custom credential chain with a managed identity credential, requests to the Azure metadata take minutes to time out instead of 1 second. This seems to be caused by the fact that...

https://github.com/Azure/azure-sdk-for-go/blob/f8cb16d25c435299b972737b2021aa1854d95e07/sdk/azidentity/default_azure_credential.go#L100

...which is set when using the **default** chain isn't set when using a custom chain.

- What did you expect or want to happen?

I expected the metadata request to time out after 1 second when not running on Azure so that the next credential in the chain is attempted.

- How can we reproduce it?

```go
package main

import (
	"context"
	"fmt"

	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
)

func main() {
	clientID := azidentity.ClientID("abcd1234-...")
	subscriptionID := "foo"
	opts := azidentity.ManagedIdentityCredentialOptions{ID: clientID}

	managed, err := azidentity.NewManagedIdentityCredential(&opts)
	if err != nil {
		panic(err)
	}

	cred, err := azidentity.NewChainedTokenCredential([]azcore.TokenCredential{managed}, nil)
	if err != nil {
		panic(err)
	}

	client, err := armresources.NewResourceGroupsClient(subscriptionID, cred, nil)
	if err != nil {
		panic(err)
	}

	res, err := client.Get(context.TODO(), "foo", nil)
	if err != nil {
		panic(err)
	}

	fmt.Println(*res.ID)
}
```

The code above blocks for more than two minutes at `client.Get()` on my machine.

- Anything we should know about your environment.

I _think_ this problem is hard to reproduce if on your dev machine you get a **fast** network failure for `169.254.169.254`. On my machine, requests to this address are dropped silently, in which case it would make sense that the SDK would give up after a short amount of time and try another credential rather than hang for a long time.

[...]

PR #19733, opened 2022-12-29
Export credentialUnavailableError constructor

chlowell:
This error type exists so a credential that can't acquire a token due to some non-fatal reason can signal `ChainedTokenCredential` to try another credential. For example, `ManagedIdentityCredential` returns this error type when managed identity isn't available in the runtime environment. Today, because this type isn't exported, `ChainedTokenCredential` treats all errors from custom credential implementations as fatal and therefore never tries the next credential in the chain. https://github.com/Azure/azure-sdk-for-go/issues/19699#issuecomment-1352295710 describes a scenario blocked by that behavior.

I also updated `ChainedTokenCredential` to behave the same whether a `credentialUnavailableError` is wrapped or not, so custom credentials can propagate other errors more easily.


JeffreyRichter:
We need to start with a very super clear definition of what "non-fatal" means exactly. 
What I'm worried about is unpredictable behavior where a credential considered something non-fatal and so the chain picks the next one but this non-fatal error is successful in the future and now a different credential is in effect. It is critical that behavior be consistent & predictable all the time once the process starts. 

So, I assume a network error is non-fatal because another attempt may succeed? But some EnvironmentCredential always fails if the env var doesn't exist or has a bad value? But then, what if the process adds the env var or changes the value? Then this EnvCred in the chain will now work and be used in a new client?

Let's discuss this at our next scrum meeting.

chlowell:
> What I'm worried about is unpredictable behavior where a credential considered something non-fatal and so the chain picks the next one but this non-fatal error is successful in the future and now a different credential is in effect.
>
I agree that would be bad. Fortunately, the default behavior of a chain is to reuse the first successful credential for all subsequent authentications. If the chain is `A -> B` and `A` returns `CredentialUnavailableError` while `B` succeeds, the chain won't try `A` again.

> I assume a network error is non-fatal because another attempt may succeed?
>
No, network errors during authentication are fatal in that they cause credentials to return other error types. `CredentialUnavailableError` means a credential can't attempt to authenticate because the runtime environment lacks some necessary data or state.

> But some EnvironmentCredential always fails if the env var doesn't exist or has a bad value? But then, what if the process adds the env var or changes the value? Then this EnvCred in the chain will now work and be used in a new client?
>
This doesn't arise because only the constructor reads environment variables, and it returns an error when it doesn't find a complete set. If the configuration is complete but incorrect, the credential will attempt to authenticate--no `CredentialUnavailableError`--and ultimately return an Azure AD error.

JeffreyRichter (approved):

sdk/azidentity/errors.go
```diff
@@ -101,24 +101,26 @@ func (*AuthenticationFailedError) NonRetriable() {
...
-// credentialUnavailableError indicates a credential can't attempt
-// authentication because it lacks required data or state.
-type credentialUnavailableError struct {
-	credType string
-	message  string
+// CredentialUnavailableError indicates a credential can't attempt authentication because it lacks required
+// data or state. When [ChainedTokenCredential] receives this error it will try its next credential, if any.
+type CredentialUnavailableError struct {
+	// Message describes the cause of the error
+	Message string
```
jhendrixMSFT, line 108: I'm thinking this shouldn't be exported.  Is there a use-case for it to be?

chlowell: My goal was to allow user code to determine the string representation of the error, so whatever logs follow an error can be meaningful. Is there a better way to do that?

jhendrixMSFT: Not sure I follow.  Do you have an example?

chlowell: https://github.com/Azure/azure-sdk-for-go/issues/19699#issuecomment-1366258523 has an example. This error may cause authentication and a client request to fail entirely, in which case a string explaining what went wrong helps in debugging. I suppose we could have an `Err error` field instead?

jhendrixMSFT: Oh, is this for allowing custom credential implementations (`wrapper` in the example) to return a `CredentialUnavailableError` with a user-supplied message?  And code outside of `azidentity` doesn't actually need to `errors.As()` for this error type?  If so, it might be simpler to add a `NewCredentialUnavailableError(msg string) error` function so we can keep the error type internal.

jhendrixMSFT: Although, if we do add more fields then this is easier to cope with over time.

chlowell: Hmm that's a good point, we could provide the constructor instead. Really, this type exists only to signal `ChainedTokenCredential` to try another credential. I doubt the distinction between it and any other error type is interesting in other contexts. The only `errors.As()` scenario I can think of just now is a custom chained credential implementation. I don't know why anyone would want to build their own but if they did, it would need `errors.As()` to integrate well with in-box credential types.

jhendrixMSFT: A custom credential chain is an interesting point, but I agree that this seems highly unlikely.

Ultimately, it seems like what we need is a way for code external to `azidentity` be able to create an error that states "this is a non-fatal error in the context of credential chains".  I feel like a constructor is a bit cleaner in the sense that we don't have to expose the mechanism that achieves this.

chlowell: Works for me, though I think if we do end up exporting the type, we won't want the constructor.

sdk/azidentity/errors.go
```diff
@@ -101,24 +101,31 @@ func (*AuthenticationFailedError) NonRetriable() {
...
+	msg := fmt.Sprintf("%s: %s", credType, message)
+	return &credentialUnavailableError{msg}
 }
 
-func (e *credentialUnavailableError) Error() string {
-	return e.credType + ": " + e.message
+// NewCredentialUnavailableError constructs an error indicating a credential can't attempt authentication
+// because it lacks required data or state. When [ChainedTokenCredential] receives this error it will try
+// its next credential, if any.
+func NewCredentialUnavailableError(message string) error {
```
jhendrixMSFT, line 119: I know I had the parameter a `string` in my initial comment, but now I'm wondering if `error` might be better?  Of course, in the case of an error that prompts code to call this ctor, the caller can always pass `err.Error()` as the parameter.  Mostly I'm thinking about this in terms of versioning.  My gut tells me `error` but I could be convinced otherwise.  What do you think?

chlowell: What sort of versioning jam do you have in mind? I'm struggling to come up with a case in which we need more from an error than its string.

jhendrixMSFT: I don't have anything concrete at present and am probably just overthinking it.  Really, all we need to convey to the token chain is "log this then move along" and this does exactly that.

jhendrixMSFT (approved):

PR #19754, opened 2023-01-05
Request.SetBody() allows overwriting with an empty body

chlowell:
Today, sending a `Request` without its body (Key Vault clients do this) requires copying it and/or modifying the underlying `http.Request` because `Request.SetBody()` won't overwrite a non-empty body with an empty one. This is apparently an oversight--`SetBody` will overwrite a non-empty body or set an empty body, but it won't do both at once.

### DIFF
--- a/sdk/azcore/internal/exported/request.go
+++ b/sdk/azcore/internal/exported/request.go
@@ -102,24 +102,28 @@ func (req *Request) OperationValue(value interface{}) bool {
 
 // SetBody sets the specified ReadSeekCloser as the HTTP request body.
 func (req *Request) SetBody(body io.ReadSeekCloser, contentType string) error {
-	// Set the body and content length.
-	size, err := body.Seek(0, io.SeekEnd) // Seek to the end to get the stream's size
-	if err != nil {
-		return err
+	var err error
+	var size int64
+	if body != nil {
+		size, err = body.Seek(0, io.SeekEnd) // Seek to the end to get the stream's size
+		if err != nil {
+			return err
+		}
 	}
 	if size == 0 {
-		body.Close()
-		return nil
+		// treat an empty stream the same as a nil one: assign req a nil body
+		body = nil
+	} else {
+		_, err = body.Seek(0, io.SeekStart)
+		if err != nil {
+			return err
+		}
+		req.Raw().GetBody = func() (io.ReadCloser, error) {
+			_, err := body.Seek(0, io.SeekStart) // Seek back to the beginning of the stream
+			return body, err
+		}
 	}
-	_, err = body.Seek(0, io.SeekStart)
-	if err != nil {
-		return err
-	}
-	req.Raw().GetBody = func() (io.ReadCloser, error) {
-		_, err := body.Seek(0, io.SeekStart) // Seek back to the beginning of the stream
-		return body, err
-	}
-	// keep a copy of the original body.  this is to handle cases
+	// keep a copy of the body argument.  this is to handle cases
 	// where req.Body is replaced, e.g. httputil.DumpRequest and friends.
 	req.body = body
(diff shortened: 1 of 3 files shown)


sdk/azcore/internal/exported/request.go
```diff
@@ -102,24 +102,28 @@ func (req *Request) OperationValue(value interface{}) bool {
...
+		if err != nil {
+			return err
+		}
+		req.Raw().GetBody = func() (io.ReadCloser, error) {
+			_, err := body.Seek(0, io.SeekStart) // Seek back to the beginning of the stream
+			return body, err
+		}
 	}
-	// keep a copy of the original body.  this is to handle cases
+	// keep a copy of the body argument.  this is to handle cases
```
jhendrixMSFT, line 126: For the case where we're clearing the body, should we remove the content-type and content-length headers?  Or at the very least the content-type header seeing there's no content.  This might make the case for a separate `RemoveBody()` (or similarly named) API.

sdk/azcore/internal/exported/request.go
```diff
@@ -102,24 +102,28 @@ func (req *Request) OperationValue(value interface{}) bool {
 
 // SetBody sets the specified ReadSeekCloser as the HTTP request body.
```
jhendrixMSFT, line 103: If we decide to keep the functionality in `SetBody()` then we should document this part of the behavior.

sdk/azcore/internal/exported/request.go
chlowell: With this, we'll clear content-type by setting its value to "" and set content-length to "0". The length is redundant, but I believe still correct. We could of course remove it instead when the given stream has no content.

I went with this approach over adding a `RemoveBody()` because that method would be redundant alongside `SetBody()`. It seems intuitive, to me anyway, to remove a request's body by setting it empty. This approach also fixes the quirky current behavior of `SetBody()`, which silently ignores an empty stream input when the request already has a body. We should do something about that in any case. If we don't overwrite the current body, we should return an error.

jhendrixMSFT: Good points, works for me (we should doc this though).

jhendrixMSFT (approved):
Just needs a doc comment.

sdk/azcore/internal/exported/request.go
chlowell: Well, it got a little more interesting. RFC 9110 specifies Content-Length shouldn't be set when a request contains no content, so I now have `SetBody()` remove the header in that case. The RFC has no similar guidance for Content-Type. I see that Go will send a 0-byte value for Content-Type when the header is set to "". That's legal HTTP and shouldn't make a difference to the server, but it means the current API can express either "I want to unset Content-Type" or "I want to set Content-Type to `""`". I prefer the latter--`SetBody()` should continue setting whatever value the caller provides and not have an opinion about what that value means. If it turns out someone really needs to unset Content-Type, they can work around this by modifying the underlying `http.Request`.

jhendrixMSFT: Good catch on the Content-Length.

Regarding Content-Type, I get what you're saying and mostly agree with you.  I just think it's a bit strange that if we're saying "pass nil/empty body to remove the body" that it doesn't also remove the Content-Type header but sends a zero-value instead.  I get that in practice it shouldn't be a problem, so maybe I'm overthinking it.

chlowell: I've definitely overthought it, and you have me reconsidering again, so I'll leave this PR open until tomorrow. At this moment I don't mind sending an empty Content-Type because although that wastes a few bytes, it is valid and semantically correct, and I believe the resulting behavior is easier to reason about.

PR #19756, opened 2023-01-05
[Feature] SAS Support + Test Framework

tasherif-msft:
SAS support + sas parsing tests
Added first client basic tests


tasherif-msft:
@jhendrixMSFT please give this a review when you get a chance

sdk/storage/azqueue/testdata/recordings/Test/TestGetProperties.json
```diff
@@ -0,0 +1,32 @@
+{
```
jhendrixMSFT, line 1: There's been a change in the test proxy that removes the need to commit recordings.  We should use that for `azqueue` so we don't have to clean it up later (we need to do this for the other modules).

@scbedd is there a link to the docs on how to enable this?

jhendrixMSFT: Here's the doc which describes how to do this.

https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/documentation/asset-sync/README.md#im-a-dev-who-uses-the-test-proxy-currently-how-do-i-externalize-my-recordings

tasherif-msft: done!

jhendrixMSFT (approved):

PR #19757, opened 2023-01-05
Code generation for Azure Storage Files

souravgupta-msft:


sdk/storage/azfile/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Azure/azure-sdk-for-go/sdk/storage/azfile
+
+go 1.19
```
jhendrixMSFT, line 3: This should be `go 1.18`

jhendrixMSFT:
Will we be able to directly export the generated code?

CC @JeffreyRichter

sdk/storage/azfile/go.mod
souravgupta-msft: Done

JeffreyRichter:
We can probably expose all the models but I think we want some custom clients because there is are service, (sub)directory, and file clients that have relationships with each other similar to blob services, containers, and blobs.

sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -0,0 +1,171 @@
...
+            $[property]["parameters"] = $[property]["parameters"].filter(function(param) { return (typeof param['$ref'] === "undefined") || (false == param['$ref'].endsWith("#/parameters/ShareName") && false == param['$ref'].endsWith("#/parameters/DirectoryPath"))});
+        }
+        else if (property.includes('/{shareName}'))
+        {
+            $[property]["parameters"] = $[property]["parameters"].filter(function(param) { return (typeof param['$ref'] === "undefined") || (false == param['$ref'].endsWith("#/parameters/ShareName"))});
+        }
+    }
+```
+
+### Use strings for dates in responses
```
jhendrixMSFT, line 49: Why are these being changes to `string`?

sdk/storage/azfile/internal/generated/zz_directory_client.go
```diff
@@ -0,0 +1,744 @@
...
+		endpoint: endpoint,
+		pl:       pl,
+	}
+	return client
+}
+
+// Create - Creates a new directory under the specified share or parent directory.
+// If the operation fails it returns an *azcore.ResponseError type.
+// Generated from API version 2020-10-02
+// fileAttributes - If specified, the provided file attributes shall be set. Default value: ‘Archive’ for file and ‘Directory’
```
jhendrixMSFT, line 44: Parameters with default values should be optional.  Can you check if these are marked as required?

sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -0,0 +1,171 @@
...
+  where: $["x-ms-paths"]..responses..headers["x-ms-file-creation-time"]
+  transform: >
+    $.format = "str";
+- from: swagger-document
+  where: $.parameters.FileChangeTime
+  transform: >
+    $.format = "str";
+```
+
+### Change new SMB file parameters to use default values
```
jhendrixMSFT, line 71: Since these have default values they should also be optional.

sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -0,0 +1,171 @@
...
+  where: $["x-ms-paths"]["/{shareName}/{directory}/{fileName}?comp=metadata"]
+  transform: >
+    $.put.responses["200"].headers["Last-Modified"] = {
+        "type": "string",
+        "format": "date-time-rfc1123",
+        "description": "Returns the date and time the file was last modified. Any operation that modifies the file, including an update of the file's metadata or properties, changes the last-modified time of the file."
+    }
+```
+
+### Add Content-MD5 to Put Range from URL
```
jhendrixMSFT, line 109: This doesn't appear to be language specific.  While we can add it here now, seems like this should be in the swagger for all languages.

sdk/storage/azfile/internal/generated/autorest.md
souravgupta-msft: This transform was used in the track 1.5 azfile SDK. So, I have used it here in the same manner.
Will change it to `time.Time`, since that looks like a better type.

sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -0,0 +1,171 @@
...
+    $.ShareMetrics.properties.IncludeAPIs["x-ms-client-name"] = "IncludeApis";
+    $.ShareServiceProperties.xml = {"name": "StorageServiceProperties"};
+    $.ShareCorsRule.xml = {"name": "CorsRule"};
+- from: swagger-document
+  where: $.parameters
+  transform: >
+    $.StorageServiceProperties.name = "ShareServiceProperties";
+```
+
+### Rename FileHttpHeaders to ShareFileHTTPHeaders and remove file prefix from properties
```
tasherif-msft, line 152: can you elaborate more on this transform? and why do we remove the file prefix from properties?

souravgupta-msft: This transform is first renaming the `FileHTTPHeaders` struct to `ShareFileHTTPHeaders`.
Secondly, it is removing the file prefix from the individual parameters of the `FileHTTPHeaders` struct. For example, `FileCacheControl` to `CacheControl`, `FileContentDisposition` to `ContentDisposition`, etc.

sdk/storage/azfile/internal/generated/zz_directory_client.go
souravgupta-msft: These are required parameters as per,
https://learn.microsoft.com/en-us/rest/api/storageservices/create-directory#request-headers

sdk/storage/azfile/internal/generated/autorest.md
souravgupta-msft: These are also required parameters as mentioned in the below docs,
https://learn.microsoft.com/en-us/rest/api/storageservices/create-directory#request-headers
https://learn.microsoft.com/en-us/rest/api/storageservices/set-directory-properties#request-headers

sdk/storage/azfile/internal/generated/autorest.md
souravgupta-msft: Removed this transform.

sdk/storage/azfile/internal/generated/autorest.md
jhendrixMSFT: The pattern is that for parameters with a default value, we treat them as optional.  Do you know if they were being generated as optional without this transform?  If not, then it might be a codegen bug.

jhendrixMSFT:
Linter error is expected due to no contents in the root package.  This will eventually go away.

sdk/storage/azfile/internal/generated/autorest.md
souravgupta-msft: These parameters are generated as required params even without the transform.
I have removed this transform and they don't have any changes in the generated code.

tasherif-msft:
@souravgupta-msft just as a final check please refer to the azure storage rest api docs to make sure generated APIs required/optional fields are mapped correctly, other than that this looks good

tasherif-msft (commented):
@souravgupta-msft just as a final check please refer to the azure storage rest api docs to make sure generated APIs required/optional fields are mapped correctly, other than that this looks good

jhendrixMSFT (approved):

PR #19761, opened 2023-01-06
AzureCLICredential reports token expiration time in UTC

chlowell:
Currently it does so in local time, unlike all other credential types. This change doesn't affect comparisons through the `time` API because those are location-aware, however it is a breaking change for anyone who took a dependency on this value's specific location. I can't imagine why anyone would have done that.

### DIFF
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -162,15 +162,15 @@ func (c *AzureCLICredential) createAccessToken(tk []byte) (azcore.AccessToken, e
 		return azcore.AccessToken{}, err
 	}
 
-	// note the Azure CLI's "expiresOn" field is local time
-	tokenExpirationDate, err := time.ParseInLocation("2006-01-02 15:04:05.999999", t.ExpiresOn, time.Local)
+	// the Azure CLI's "expiresOn" is local time
+	exp, err := time.ParseInLocation("2006-01-02 15:04:05.999999", t.ExpiresOn, time.Local)
 	if err != nil {
 		return azcore.AccessToken{}, fmt.Errorf("Error parsing token expiration time %q: %v", t.ExpiresOn, err)
 	}
 
 	converted := azcore.AccessToken{
 		Token:     t.AccessToken,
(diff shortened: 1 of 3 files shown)


sdk/azidentity/azure_cli_credential.go
```diff
@@ -162,32 +162,17 @@ func (c *AzureCLICredential) createAccessToken(tk []byte) (azcore.AccessToken, e
...
+	exp, err := time.ParseInLocation("2006-01-02 15:04:05.999999", t.ExpiresOn, time.Local)
 	if err != nil {
-		return azcore.AccessToken{}, fmt.Errorf("Error parsing Token Expiration Date %q: %+v", t.ExpiresOn, err)
+		return azcore.AccessToken{}, fmt.Errorf("Error parsing token expiration time %q: %v", t.ExpiresOn, err)
 	}
 
 	converted := azcore.AccessToken{
 		Token:     t.AccessToken,
-		ExpiresOn: *tokenExpirationDate,
+		ExpiresOn: exp.UTC(),
```
jhendrixMSFT, line 173: How did this bug manifest?

chlowell: I noticed it while trying to write better unit tests. It's kind of marginal as a bug in that it probably has no impact for applications beyond how expiration times appear as strings.

jhendrixMSFT: I get that this unifies with the other credentials.  However, I can see why running something locally would return the expiration time in local time (vs asking Azure for a token which isn't local).  It probably doesn't matter either way, and if it improves testing then that's a win.

jhendrixMSFT (approved):

sdk/azidentity/azure_cli_credential.go
chlowell: I suppose that's the CLI's rationale for converting AAD's Unix timestamp to local time for display. `azidentity` is a library that generally serves internal purposes for applications though, and I believe software should use UTC internally wherever possible. I agree what we do in this case probably doesn't matter either way 😄

PR #19781, opened 2023-01-10
Added SAS missing permissions tests and fixed parsing logic

gapra-msft:

### DIFF
--- a/sdk/storage/azblob/common.go
+++ b/sdk/storage/azblob/common.go
@@ -24,6 +24,7 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 // NOTE: Changing any SAS-related field requires computing a new SAS signature.
 type URLParts = sas.URLParts
 
+// TODO : Can we delete below? Currently unused.
 // ParseURL parses a URL initializing URLParts' fields including any SAS-related & snapshot query parameters. Any other
 // query parameters remain in the UnparsedParams field. This method overwrites all fields in the URLParts object.
 func ParseURL(u string) (URLParts, error) {
--- a/sdk/storage/azblob/common.go
+++ b/sdk/storage/azblob/common.go
@@ -24,7 +24,6 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 // NOTE: Changing any SAS-related field requires computing a new SAS signature.
 type URLParts = sas.URLParts
 
-// TODO : Can we delete below? Currently unused.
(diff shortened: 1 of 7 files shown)


sdk/storage/azblob/sas/service.go
```diff
@@ -322,25 +333,29 @@ func (p *ContainerPermissions) String() string {
 			p.DeletePreviousVersion = true
 		case 'l':
 			p.List = true
-		case 't':
+		case 'f':
```
vibhansa-msft, line 336: Order of these fields matter in the sas, can we put some test cases around the valid and invalid sequences.

gapra-msft: Yep, This would be validated by the String tests and the tests that generate and use sas in tests

tasherif-msft (approved):

sdk/storage/azblob/common.go
```diff
@@ -24,6 +24,7 @@ func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredentia
 // NOTE: Changing any SAS-related field requires computing a new SAS signature.
 type URLParts = sas.URLParts
 
+// TODO : Can we delete below? Currently unused.
```
jhendrixMSFT, line 27: I thought we wanted this in `azblob` as part of SAS support.  Also, we export `URLParts` above.  So it would seem weird to export it and not `ParseURL()` from the same package.

tasherif-msft: It just seems like it's duplicate code, clients (container, pageblob, etc.) are using the parser (and URLParts) defined here https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blob/utils.go#L77
so this seems safe to remove along with URLParts from this common file, unless we want to use this instead of the one defined in utils @jhendrixMSFT

jhendrixMSFT: The original thinking was that the contents of the `azblob` package should address the 95% use-case (i.e. you should only have to import `azblob` and no other sub-packages), and SAS support was determined to be part of that 95%.

siminsavani-msft:
Need to add line in CHANGELOG; otherwise, LGTM!

siminsavani-msft (approved):

sdk/storage/azblob/common.go
tasherif-msft: @jhendrixMSFT I see, we can keep this then

jhendrixMSFT (approved):

Issue #19797, opened 2023-01-12
[azappconfig] AddSetting should handle content-type


Eviradnus:
### Get AppConfig settings Content-Type back

The v0.1.0 & v0.2.0 provided a way to add AppConfig settings with a Content-Type (using a generous `Setting` struct); it's now impossible (from v0.3.0 onwards, only the label is accessible in the `AddSettingOptions`).

It would be nice to get this feature back, typically for Azure KeyVault references.

jhendrixMSFT:
Fix will come out in this month's release.

PR #19800, opened 2023-01-12
[azappconfig] Handle setting content type in AddSetting and SetSetting (#19797)

Eviradnus:
Just added a `ContentType` near the `Label` in `*SettingOptions`, to be used as it was before the v0.3.0 breaking changes.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/19797

antkmsft (approved):
Thank you!

PR #19836, opened 2023-01-17
[Blob] Fix API signatures and tests

tasherif-msft:


siminsavani-msft (approved):

sdk/storage/azblob/lease/blob_client.go
```diff
@@ -102,19 +102,19 @@ func (c *BlobClient) BreakLease(ctx context.Context, o *BlobBreakOptions) (BlobB
 
 // ChangeLease changes the blob's lease ID.
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/lease-blob.
-func (c *BlobClient) ChangeLease(ctx context.Context, o *BlobChangeOptions) (BlobChangeResponse, error) {
+func (c *BlobClient) ChangeLease(ctx context.Context, proposedLeaseID string, o *BlobChangeOptions) (BlobChangeResponse, error) {
```
jhendrixMSFT, line 105: I thought that having an auto-generated lease ID, thus making the proposed lease ID optional, was the desired behavior.  I'm not pushing back on the change as I don't know enough about this scenario, but I want to confirm that this new behavior is what we want.

CC @JeffreyRichter

JeffreyRichter: There are times when a customer needs to specify a precise lease ID. But, we could make it optional; not mandatory.

jhendrixMSFT: Yes agreed, and that's what we previously had implemented.

jhendrixMSFT: I think part of the original confusion is that, while the proposed lease ID is required by the REST endpoint, we didn't doc that we set it to a random UUID on your behalf if you don't specify your own.  If we are to keep the previous behavior, then we need to doc it.

sdk/storage/azblob/pageblob/client.go
```diff
@@ -150,14 +150,18 @@ func (pb *Client) Create(ctx context.Context, size int64, o *CreateOptions) (Cre
 // This method panics if the stream is not at position 0.
 // Note that the http client closes the body stream after the request is sent to the service.
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/put-page.
-func (pb *Client) UploadPages(ctx context.Context, body io.ReadSeekCloser, options *UploadPagesOptions) (UploadPagesResponse, error) {
+func (pb *Client) UploadPages(ctx context.Context, body io.ReadSeekCloser, contentRange blob.HTTPRange, options *UploadPagesOptions) (UploadPagesResponse, error) {
```
jhendrixMSFT, line 153: From the internal discussion.

> We are taking ReadSeekCloser as input and calculating length of it in code. If no range is given then we are overwriting from 0 offset so shall be ok.
that's actually not correct I attempted to have 0 and the API fails - it requires a proper range, I have looked at python as well and range is expected to be provided explicitly - so the customer needs to have it handy

It's not clear to me if `HTTPRange` is really required or we have a bug in our math to calculate the range from the provided `ReadSeekCloser`.  Can somebody clarify?

CC @JeffreyRichter

JeffreyRichter: For UploadPages, a start offset MUST be specified by the caller. The stream has to be a multiple of 512 bytes and we could calculate the end by adding the stream length to the start offset.

jhendrixMSFT: I see.  Are there cases where the caller might want to specify an end that's shorter than the `ReadSeekCloser`?  If so, then specifying an explicit `HTTPRange` makes sense.

jhendrixMSFT: Let's just keep the `HTTPRange` considering this is the power-user scenario where complete control is best.

JeffreyRichter: Either customer specifies the end (and we don't read the whole stream) or the customer wraps the stream with a call to https://pkg.go.dev/io#LimitReader. I kind like the latter because the customer knows what to expect of the stream but LimitReader returns an io.Reader; not an io.ReadSeekCloser which we'd need.

jhendrixMSFT: That's unfortunate.  Maybe we need a `streaming.LimitReader` that's an `io.ReadSeeker` at some point.

sdk/storage/azblob/lease/blob_client.go
JeffreyRichter: If I remember correctly, the service doesn't "require" this but we always want to pass it because it makes the operation idempotent so we can retry it.

JeffreyRichter: Oh, you're right. I thought you were talking about AcquireLease; not ChangeLease. For ChangeLease, you could see the ID being required because if the customer wants to change the lease, they most-likely have a specific lease ID in mind to change it to. Otherwise, there is no real need to change the lease.

tasherif-msft: @jhendrixMSFT @JeffreyRichter yea the new lease id is required for the rest endpoint, as @JeffreyRichter said the point of this is operation is to change the id so it should be precise by the customer anyway. So just to confirm, this change looks ok?

sdk/storage/azblob/container/client.go
```diff
@@ -217,15 +217,12 @@ func (c *Client) GetAccessPolicy(ctx context.Context, o *GetAccessPolicyOptions)
...
 // SetAccessPolicy sets the container's permissions. The access policy indicates whether blobs in a container may be accessed publicly.
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/set-container-acl.
-func (c *Client) SetAccessPolicy(ctx context.Context, containerACL []*SignedIdentifier, o *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
-	accessPolicy, mac, lac := o.format()
-	for _, c := range containerACL {
-		err := formatTime(c)
-		if err != nil {
-			return SetAccessPolicyResponse{}, err
-		}
+func (c *Client) SetAccessPolicy(ctx context.Context, o *SetAccessPolicyOptions) (SetAccessPolicyResponse, error) {
```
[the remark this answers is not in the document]
tasherif-msft: done

vibhansa-msft (approved):

PR #19852, opened 2023-01-19
Queue Client + More Tests

tasherif-msft:
Queue Client:
- queue APIs
- message APIs

Tests:
- service client tests
- queue client tests

More tests to come in a separate PR.


sdk/storage/azblob/container/client.go
```diff
@@ -190,7 +190,7 @@ func (c *Client) Restore(ctx context.Context, deletedContainerVersion string, op
 // For more information, see https://docs.microsoft.com/rest/api/storageservices/get-container-metadata.
 func (c *Client) GetProperties(ctx context.Context, o *GetPropertiesOptions) (GetPropertiesResponse, error) {
 	// NOTE: GetMetadata actually calls GetProperties internally because GetProperties returns the metadata AND the properties.
-	// This allows us to not expose a GetProperties method at all simplifying the API.
+	// This allows us to not expose a GetMetadata method at all simplifying the API.
```
jhendrixMSFT, line 193: Was this meant to be changed?

sdk/storage/azqueue/client.go
```diff
@@ -18,59 +18,60 @@ import (
...
-// ClientOptions contains the optional parameters when creating a Client.
+// ClientOptions contains the optional parameters when creating a ServiceClient or QueueClient.
 type ClientOptions struct {
 	azcore.ClientOptions
 }
 
-// Client represents a URL to the Azure Queue Storage service allowing you to manipulate queues.
-type Client base.Client[generated.ServiceClient]
+// ServiceClient represents a URL to the Azure Queue Storage service allowing you to manipulate queues.
+type ServiceClient base.Client[generated.ServiceClient]
```
jhendrixMSFT, line 32: Given the name change, probably good to rename the file to match.

sdk/storage/azqueue/client.go
```diff
@@ -168,21 +166,47 @@ func (s *Client) NewListQueuesPager(o *ListQueuesOptions) *runtime.Pager[ListQue
...
+// this Client's URL. The new QueueClient uses the same request policy pipeline as the Client.
+func (s *ServiceClient) NewQueueClient(queueName string) *QueueClient {
+	queueName = url.PathEscape(queueName)
+	queueURL := runtime.JoinPaths(s.URL(), queueName)
+	return (*QueueClient)(base.NewQueueClient(queueURL, s.generated().Pipeline(), s.sharedKey()))
+}
+
+// CreateQueue creates a new queue within a storage account. If a queue with the same name already exists, the operation fails.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/create-queue4.
+func (s *ServiceClient) CreateQueue(ctx context.Context, queueName string, options *CreateOptions) (generated.QueueClientCreateResponse, error) {
```
jhendrixMSFT, line 179: This is returning an internal type.

sdk/storage/azqueue/client.go
```diff
@@ -81,46 +82,43 @@ func NewClientFromConnectionString(connectionString string, options *ClientOptio
...
 	return s.generated().Endpoint()
 }
 
-// TODO: CreateQueue
-// TODO: DeleteQueue
-
 // GetProperties - gets the properties of a storage account's Queue service, including properties for Storage Analytics
 // and CORS (Cross-Origin Resource Sharing) rules.
-func (s *Client) GetProperties(ctx context.Context, o *GetPropertiesOptions) (GetPropertiesResponse, error) {
+func (s *ServiceClient) GetProperties(ctx context.Context, o *GetPropertiesOptions) (GetPropertiesResponse, error) {
```
jhendrixMSFT, line 106: Can we rename the return type to `GetServicePropertiesResponse` so the pattern is consistent with the queues client.

sdk/storage/azqueue/queue_client.go
```diff
@@ -0,0 +1,250 @@
...
+	return resp, err
+}
+
+// UpdateMessage updates a message from the queue with the given popReceipt.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/update-message.
+func (q *QueueClient) UpdateMessage(ctx context.Context, messageID string, popReceipt string, content string, o *UpdateMessageOptions) (UpdateMessageResponse, error) {
+	opts := o.format()
+	message := generated.QueueMessage{MessageText: &content}
+	messageURL := runtime.JoinPaths(q.generated().Endpoint(), messageID)
+	messageClient := generated.NewMessageIDClient(messageURL, q.generated().Pipeline())
```
jhendrixMSFT, line 175: Shouldn't this use the internal `MessageIDClient`?

sdk/storage/azqueue/queue_client.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue/internal/generated"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue/internal/shared"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue/sas"
+	"time"
+)
+
+// QueueClient represents a URL to the Azure Queue Storage service allowing you to manipulate queues.
+type QueueClient base.CompositeClient[generated.QueueClient, generated.MessagesClient, generated.MessageIDClient]
+
+func (q *QueueClient) generated() *generated.QueueClient {
```
jhendrixMSFT, line 25: Maybe name this `queueClient()`?

sdk/storage/azqueue/models.go
```diff
@@ -120,7 +181,7 @@ func (o *SetPropertiesOptions) format() (generated.StorageServiceProperties, *ge
 
 // ---------------------------------------------------------------------------------------------------------------------
 
-// GetPropertiesOptions contains the optional parameters for the Client.GetProperties method.
+// GetPropertiesOptions contains the optional parameters for the ServiceClient.GetProperties method.
 type GetPropertiesOptions struct {
```
jhendrixMSFT, line 185: GetServicePropertiesOptions to match with queue.

sdk/storage/azblob/container/client.go
tasherif-msft: yea just a small correction in the docs

sdk/storage/azqueue/queue_client.go
tasherif-msft: that's what I'm doing here, I'm not sure I understand

sdk/storage/azqueue/client.go
tasherif-msft: done

sdk/storage/azqueue/queue_client.go
jhendrixMSFT: The call to `base.NewQueueClient()` in the client constructor already created a MessageIDClient, so you should be able to use that one instead.

tasherif-msft: ah right, I may be unable to use that since I have to append the message id to the URL, so I'll have to remove the message id client from the composite and keep this code as is

tasherif-msft: @jhendrixMSFT I removed messageID from the composite struct since we need to construct a message specific URL, now we'll just directly construct the generated ID client and perform the operation without exposing messageID client to the customer

jhendrixMSFT (approved):

PR #19856, opened 2023-01-20
Increase timeout for azkeys live tests

chlowell:
We need to allow more time because de/provisioning resources (MHSM) can be very slow (~50 minutes).

### DIFF
--- a/sdk/keyvault/azkeys/ci.yml
+++ b/sdk/keyvault/azkeys/ci.yml
@@ -24,6 +24,7 @@ pr:
 stages:
 - template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
   parameters:
+    TimeoutInMinutes: 120
     ServiceDirectory: 'keyvault/azkeys'
     RunLiveTests: true
     AdditionalMatrixConfigs:


gracewilcox (approved):

benbp:
@chlowell is the de-provisioning part of the test itself, or a post-test action?

chlowell:
It's a task following the unit tests ([for example](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2114194&view=logs&j=837a4e85-6685-5796-76b9-3e8f37298b2a&t=1e26a6c1-6e63-5948-c438-ccca8c0fd9d5)).

benbp:
Ah ok. We could also make a choice here to not block on removal, and depend on our live test cleanup jobs that run every four hours. It will incur a little more cost, but would reduce the time that a pipeline blocks a PR/release/dev/etc.

chlowell:
That would help with slow deprovisioning, but wouldn't we still need to increase the timeout to address slow provisioning (which is much more common)?

benbp:
@chlowell ah, I misread `de/provisioning` as `de-provisioning`. Yeah if slow provisioning is a factor, then we need the timeout bump regardless.

Issue #19721, opened 2022-12-20
Add support to set timeout in ManagedIdentityCredential


JorTurFer:
In [KEDA](https://github.com/kedacore/keda) we support managed identities with AAd-Pod-Identity and with WI. We'd like to introduce a custom chained token because in our use case, we support using a different clientId than the default value for a better permission segmentation. 

We have been working with different implementations depending on the podIdentity set by the user, but for the sake of the unification, and due to the new SDK allows us to unify both using chained credential, we want to unify them. Our problem is that we cannot set the timeout for `ManagedIdentityCredential` because it's an internal and the default value is too long (60 sec if I'm not wrong) for waiting before trying the next credential in the chain.

I have checked `DefaultAzureCredential` and it sets the timeout in 1 second 
```
msiCred, err := NewManagedIdentityCredential(o)
	if err == nil {
		creds = append(creds, msiCred)
		msiCred.mic.imdsTimeout = time.Second
	}
```
This is perfect and could solve our use case perfectly, but it's an internal setter and we don't have any change to modify that value. It'd be nice if `ManagedIdentityCredentialOptions` supports a new parameter to set this timeout instead of setting it inside `DefaultAzureCredential`. I guess that this could be solved just adding a new property there and configuring the timeout before returning `ManagedIdentityCredential`.

If you think that it's something feasible, I'm willing to contribute with the needed changes

chlowell:
You can use the `context.Context` parameter of client and credential methods to control timeouts and cancellation. #19699 has examples showing how to replicate this internal timeout (note some of those examples require `azidentity` to export a new error type, which I'm working on).

Your scenario sounds unusual though. Why does your application chain credentials to support custom client IDs? Can it simply take an optional ID as configuration and then construct a single credential with that value?

JorTurFer:
The problem we have faced is that KEDA stack a lot of permissions in its managed identity due to what KEDA is for (for getting autoscaling metrics, we need to have administrator permissions in azure services). 
Our users requested us a better permission segmentation to reduce the potential breach, and that's why we support the option to set a specific managed identity for a specific scaler instead of using the KEDA managed identity, with this approach, we can reduce the amount of stack permissions in the same identity, that's why we allow the option of giving another managed identity to use. 
At this point, maybe @tomkerkhove or @v-shenoy can give you more info about the reasons.

[...]

PR #19859, opened 2023-01-20
Replace internal IMDS timeout with public API workaround

chlowell:
A couple of users have asked us to export the internal API DefaultAzureCredential uses to impose a timeout on managed identity authentication, to avoid a long timeout when IMDS is unavailable (#19699, #19721). We don't want to do that because it means adding a redundant, esoteric cancellation mechanism that conflicts with cancellation via `context.Context`. This example shows how to reimplement this DefaultAzureCredential behavior with Context and a small wrapper.

I'm also tempted to replace the internal API with this wrapper. That would steer future readers of the source in the right direction, and I believe it would simplify some of the implementation as well. Thoughts?

### DIFF
--- a/sdk/azidentity/default_azure_credential.go
+++ b/sdk/azidentity/default_azure_credential.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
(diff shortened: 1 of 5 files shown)


jhendrixMSFT (approved):
I think your idea to replace the timeout with the wrapper makes sense.  And if people were finding the timeout, maybe they'll find the wrapper and do the same thing.

chlowell:
Well, the wrapper got a little more complicated when I put it to internal use. One behavior we have internally is that if managed identity authentication succeeds, DefaultAzureCredential stops applying this special timeout, to prevent future failures due to transient latency. I updated the example to behave the same; it's a bit ugly because `GetToken()` has a value receiver 🙄

sdk/azidentity/default_azure_credential.go
```diff
@@ -94,10 +95,9 @@ func NewDefaultAzureCredential(options *DefaultAzureCredentialOptions) (*Default
 	if haveClientID {
 		o.ID = ClientID(clientID)
 	}
-	msiCred, err := NewManagedIdentityCredential(o)
+	miCred, err := NewManagedIdentityCredential(o)
 	if err == nil {
-		creds = append(creds, msiCred)
-		msiCred.mic.imdsTimeout = time.Second
+		creds = append(creds, timeoutWrapper{mic: miCred, timeout: to.Ptr(time.Second)})
```
jhendrixMSFT, line 100: If you use `&timeoutWrapper{...}` you can change the receiver to byref.

PR #19869, opened 2023-01-23
Added basic default values for version parameter and cors rule + tests

tasherif-msft:


sdk/storage/azblob/service/models.go
```diff
@@ -188,6 +192,43 @@ func (o *SetPropertiesOptions) format() (generated.StorageServiceProperties, *ge
...
 	}
 
+	defaultVersion := to.Ptr[string]("1.0")
+	defaultAge := to.Ptr[int32](0)
+	emptyStr := to.Ptr[string]("")
+
+	if o.Cors != nil {
+		for i := 0; i < len(o.Cors); i++ {
+			if o.Cors[i].AllowedHeaders == nil {
+				o.Cors[i].AllowedHeaders = emptyStr
```
vibhansa-msft, line 202: How was this working earlier ?

souravgupta-msft: Can you share some reference for this?
If its documented somewhere in the Microsoft docs, then maybe we should also document this.

tasherif-msft: So if you reference the generated docs it clearly says all the fields are required if one is provided. However, the average customer will usually provide the allowedOrigins and allowedHeaders, the other parameters are not as intiutitive to keep as an empty str or set to 0, so I added these checks as quick validations. You can see this in python as well here https://github.com/Azure/azure-sdk-for-python/blob/8a4472205bf4fff0566a427e41888d0c0dc0fecf/sdk/storage/azure-storage-blob/azure/storage/blob/_models.py#L333-L335

vibhansa-msft (approved):

sdk/storage/azblob/service/client_test.go
```diff
@@ -157,6 +157,123 @@ func (s *ServiceUnrecordedTestsSuite) TestListContainersBasic() {
...
+	_require.NoError(err)
+
+	days := to.Ptr[int32](5)
+	enabled := to.Ptr(true)
+
+	loggingOpts := service.Logging{
+		Read: enabled, Write: enabled, Delete: enabled,
+		RetentionPolicy: &service.RetentionPolicy{Enabled: enabled, Days: days}}
+	opts := service.SetPropertiesOptions{Logging: &loggingOpts}
+	_, err = svcClient.SetProperties(context.Background(), &opts)
```
souravgupta-msft, line 172: I think you'll have to add some wait time after running SetProperties(). Make sure to run the live tests.

souravgupta-msft (approved):

sdk/storage/azblob/service/client_test.go
tasherif-msft: it ran fine without sleep

PR #19872, opened 2023-01-25
Add support for MSALGO's InstanceDiscoveryDisabled flag

jonnguyen1:


sdk/azidentity/azidentity.go
```diff
@@ -151,17 +163,17 @@ func (p pipelineAdapter) Do(r *http.Request) (*http.Response, error) {
 
 // enables fakes for test scenarios
 type confidentialClient interface {
-	AcquireTokenSilent(ctx context.Context, scopes []string, options ...confidential.AcquireTokenSilentOption) (confidential.AuthResult, error)
-	AcquireTokenByAuthCode(ctx context.Context, code string, redirectURI string, scopes []string, options ...confidential.AcquireTokenByAuthCodeOption) (confidential.AuthResult, error)
-	AcquireTokenByCredential(ctx context.Context, scopes []string) (confidential.AuthResult, error)
-	AcquireTokenOnBehalfOf(ctx context.Context, userAssertion string, scopes []string) (confidential.AuthResult, error)
+	AcquireTokenSilent(ctx context.Context, scopes []string, options ...confidential.AcquireSilentOption) (confidential.AuthResult, error)
+	AcquireTokenByAuthCode(ctx context.Context, code string, redirectURI string, scopes []string, options ...confidential.AcquireByAuthCodeOption) (confidential.AuthResult, error)
```
jonnguyen1, line 167: Changed to match MSAL interface

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -226,6 +227,33 @@ func TestClientCertificateCredential_Live(t *testing.T) {
 	}
 }
 
+func TestClientCertificateCredentialADFS_Live(t *testing.T) {
```
jonnguyen1, line 230: I ended up creating a new test after considering it. There is a different tenant id, client id, authhost, etc. Essentially every env variable would be different from the other run. Also, ADFS cert authentication needed a password, which was not on the previous test. Additionally, I only checked PFX, as I created my graph application in the ADFS stack with a PFX cert. While I could've created a PEM file aswell, I believe that is unnecessary, due to the instanceDiscoveryFlag not affecting the flow of certificate parsing.

sdk/azidentity/client_secret_credential_test.go
```diff
@@ -42,7 +43,11 @@ func TestClientSecretCredential_GetTokenSuccess(t *testing.T) {
 func TestClientSecretCredential_Live(t *testing.T) {
 	opts, stop := initRecording(t)
 	defer stop()
-	o := ClientSecretCredentialOptions{ClientOptions: opts}
+	disableID, err := strconv.ParseBool(disableInstanceDiscovery)
```
jonnguyen1, line 46: Could potentially move this logic into livetest, but I elected to put it here due to error handling if the env variable is empty (aka most cases), and it would just set disableID to be false (default value)

sdk/azidentity/live_test.go
```diff
@@ -62,7 +64,9 @@ const (
 	fakeUsername   = "fake@user"
 )
 
-var liveTestScope = "https://management.core.windows.net//.default"
+var liveTestScope = "https://management.core.windows.net//.default" // For ADFS tests, had to change this scope to openid
```
jonnguyen1, line 67: just a small note that this liveTestScope had to change for the ADFS tests. Might have to get this value from an environment variable in the future? Open to suggestions whether that is necessary.

sdk/azidentity/environment_credential_test.go
```diff
@@ -278,6 +283,7 @@ func TestEnvironmentCredential_InvalidClientSecretLive(t *testing.T) {
 
 func TestEnvironmentCredential_UserPasswordLive(t *testing.T) {
 	vars := map[string]string{
+		//for ADFS change clientID
```
jonnguyen1, line 286: The client ID for ADFS had to be changed to a specific client such as windows Powershell.

sdk/azidentity/environment_credential_test.go
```diff
@@ -278,6 +283,7 @@ func TestEnvironmentCredential_InvalidClientSecretLive(t *testing.T) {
...
 		azureClientID: developerSignOnClientID,
```
jonnguyen1, line 287: This developerSignOnClientID doesn't work with ADFS. For testing, I had to change this client ID to be someting that ADFS accepts, such as Windows Powershell.

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -228,12 +229,17 @@ func TestClientCertificateCredential_Live(t *testing.T) {
 }
 
 func TestClientCertificateCredentialADFS_Live(t *testing.T) {
-	certData, err := os.ReadFile(liveSP.pfxPath)
+	if recording.GetRecordMode() == recording.LiveMode {
+		if adfsLiveSP.certPass == "" || adfsLiveSP.clientID == "" || adfsLiveSP.pfxPath == "" || adfsLiveSP.scope == "" || adfsLiveSP.tenantID == "" {
+			t.Skip("this test requires manual recording and access to ADFS instance, and can't pass live in CI")
+		}
+	}
```
chlowell, line 236: A few comments here:
1. Live and recording mode are distinct but have the same configuration requirements i.e. the distinction is between playback mode and everything else
1. The skip message could point to what this test needs for a live run
1. Won't the tenant always be "adfs"?
```suggestion
	if recording.GetRecordMode() != recording.PlaybackMode {
		if adfsLiveSP.certPass == "" || adfsLiveSP.clientID == "" || adfsLiveSP.pfxPath == "" || adfsLiveSP.scope == "" {
			t.Skip("set ADFS_SP_* environment variables to run this test live")
		}
	}
```

sdk/azidentity/live_test.go
```diff
@@ -64,7 +62,35 @@ const (
...
-var liveTestScope = "https://management.core.windows.net//.default" // For ADFS tests, had to change this scope to openid
+var adfsLiveSP = struct {
+	tenantID string
+	clientID string
+	secret   string
+	pfxPath  string
+	certPass string
+	scope    string
+}{
+	tenantID: os.Getenv("ADFS_SP_TENANT_ID"),
```
chlowell, line 73: Won't this always be "adfs"?

sdk/azidentity/azidentity.go
chlowell: You can rebase this out, main now has equivalent changes

sdk/azidentity/live_test.go
```diff
@@ -111,6 +141,8 @@ func TestMain(m *testing.M) {
 			liveSP.tenantID:                                 fakeTenantID,
 			liveUser.tenantID:                               fakeTenantID,
 			liveUser.username:                               fakeUsername,
+			adfsLiveSP.tenantID:                             fakeTenantID,
+			adfsLiveUser.tenantID:                           fakeTenantID,
```
chlowell, line 145: We don't want to replace "adfs" as the tenant in recordings because we want to test any magic around this value, and we don't need to replace it to make playback work because it's a static string.

sdk/azidentity/interactive_browser_credential_test.go
```diff
@@ -92,4 +92,35 @@ func TestInteractiveBrowserCredential_Live(t *testing.T) {
...
+
+func TestInteractiveBrowserCredentialADFS_Live(t *testing.T) {
+	if !runManualBrowserTests {
+		t.Skip("set AZIDENTITY_RUN_MANUAL_BROWSER_TESTS to run this test")
+	}
+	if adfsLiveUser.clientID == "" {
+		t.Skip("set ADFS_IDENTITY_TEST_CLIENT_ID environment variables to run this test live")
+	}
+	//Redirect URL is necessary
+	url := "http://localhost:8405"
```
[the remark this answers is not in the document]
jonnguyen1: Stack ADFS requires a specific port. Wildcard ports also aren't allowed, so we will need to designate a port.

sdk/azidentity/azidentity.go
```diff
@@ -56,20 +57,31 @@ var getConfidentialClient = func(clientID, tenantID string, cred confidential.Cr
...
 	}
 	o = append(o, additionalOpts...)
+	if strings.ToLower(tenantID) == "adfs" {
+		o = append(o, confidential.WithInstanceDiscovery(false))
+	}
 	return confidential.New(clientID, cred, o...)
 }
 
-var getPublicClient = func(clientID, tenantID string, co *azcore.ClientOptions) (public.Client, error) {
+func getPublicClient(clientID, tenantID string, co *azcore.ClientOptions, additionalOpts ...public.Option) (public.Client, error) {
```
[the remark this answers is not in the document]
jonnguyen1: Ah this was probably just a merge issue

sdk/azidentity/client_certificate_credential_test.go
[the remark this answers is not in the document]
keystroke: ADFS rsa keypairs work same way as in AAD, only public key (cert) needs to be uploaded and associated with the application. Don't see the code that this comment is referencing anymore, perhaps it meant the pfx / pem file was password protected?

sdk/azidentity/live_test.go
keystroke: Also CC'ing @chlowell for some background here.

ADFS doesn't support oauth2 in the same way as is setup in Azure. There is no graph API etc. to go with it. For ASH the ADFS team built a custom graph service with limited capability that jury-rigged applications and service principals and some oauth2 stuff into the flow, but essentially working off of pre-authorization for applications. For example the ARM application was saved citing that any client application is authorized to request tokens for it without user consent, as if an admin had pre-consented to apps the way that can be done in Azure.

Additionally, ADFS does not translate the fake ".default" scope segment into scope values for consented permissions the way AAD does. Instead, it only supports literal scope values in a token request. In V1 token requests, there was a resource parameter that was used, alongside a scope value. In V2 flows, they combine these properties and still call it "scope" so the format becomes "resource/scope". There are also global scopes, like "openid" and "profile" and "email" which influence what claims STS include in the various tokens which aren't directly associated with any application.

In the AAD world, when you ask for a token, you provide resource/scope and AAD will look at the oauth2permissions exposed by the application described by the "resource" segment of the "scope" to then evaluate the "scope" segment of the string, with the exception of the magic ".default" scope value for SP auth to function as an alias for "all the scopes for which I have been authorized to acquire, give me the maximally powerful token" which is the design AAD went with for whatever reason.

Another exception (IIRC) is that the graph permission scopes are also "naked" so you don't need to request them with resource/scope you can use them like the global scopes.

The other relevant behavior here, is that most azure services don't use scope-based access controls, they use resource-based access controls, so only the aud claim is evaluated by them (for example, ARM doesn;t even look at the scp claim in tokens). So the scp value doesn't actaully matter for most Azure services.

Because of above, to make this Azure SDK flows work in ADFS setups, we needed ADFS to be able to resolve the target "scope" value always as a global scope value. It will still chop out the "resource" segment of a scope value to populate the aud claim, but it needs to recognize the remaining "scope" value as a global scope setup in ADFS.

Originally, we discussed with ADFS MSAL teams and agreed they would use "openid" as the scope segment for the resource/scope format so ADFS would recognize "openid" and add it as the value of the scp claim in the resulting access token.

The token "works" when calling ARM because the scope value is irrelevant, even in Azure where the token acquired would be for scope "https://management.azure.com//user_impersonation" resulting in aud claim of "https://management.azure.com/" and scp claim of "user_impersonation" (which is the default permission scope added to all applications be default and ignored for the most part).

However, for ASH stamps on recent / latest builds, we started adding "default" as a global scope built-in to ADFS so it could recognize that value on incoming calls, so code changes aren't required in the MSAL or app layers to toggle between using .default and openid based on identity provider. This results in ADFS tokens having ".default" as a literal scp claim value, instead of being translated into some permssion scope exposed by that app. This works only because the relevant azure services being called do not use scope-based auth / permissions.

Sorry for wall of text but there is some background on the behavior. For this comment specifically, please try the flow again as latest ASH ADFS environments should recognize ".default" scope value.

sdk/azidentity/interactive_browser_credential.go
```diff
@@ -30,6 +30,9 @@ type InteractiveBrowserCredentialOptions struct {
 	// only when setting ClientID, and must match a redirect URI in the application's registration.
 	// Applications which have registered "http://localhost" as a redirect URI need not set this option.
 	RedirectURL string
+
+	// DisableInstanceDiscovery allows disconnected cloud solutions to skip instance discovery for unknown authority hosts.
+	DisableInstanceDiscovery bool
```
chlowell, line 35: Can you please place this after `ClientID`? The struct definition appears verbatim in the package docs; alphabetizing the fields makes them easier to scan.

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -224,6 +225,53 @@ func TestClientCertificateCredential_Live(t *testing.T) {
 			testGetTokenSuccess(t, cred)
 		})
 	}
+	t.Run("disableID", func(t *testing.T) {
```
chlowell, line 228: Consistent with other tests, avoids further overloading "ID"
```suggestion
	t.Run("instance discovery disabled", func(t *testing.T) {
```

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -224,6 +225,53 @@ func TestClientCertificateCredential_Live(t *testing.T) {
...
+			t.Fatalf("failed to construct credential: %v", err)
+		}
+		testGetTokenSuccess(t, cred)
+	})
+}
+
+func TestClientCertificateCredentialADFS_Live(t *testing.T) {
+	if recording.GetRecordMode() != recording.PlaybackMode {
+		if adfsLiveSP.clientID == "" || adfsLiveSP.pemPath == "" || adfsLiveSP.scope == "" {
+			t.Skip("this test requires manual recording and access to ADFS instance, and can't pass live in CI")
```
chlowell, line 254:
```suggestion
			t.Skip("set ADFS_SP_* to run this test live")
```

sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -224,6 +225,53 @@ func TestClientCertificateCredential_Live(t *testing.T) {
...
+		}
+	}
+	//Set authority host to be ADFS authority
```
chlowell, line 257:
```suggestion
```

sdk/azidentity/live_test.go
```diff
@@ -56,12 +57,42 @@ var liveUser = struct {
...
+
+var adfsLiveSP = struct {
+	tenantID    string
+	clientID    string
+	secret      string
+	pemPath     string
+	scope       string
+	redirectURL string
+}{
+	tenantID:    "adfs",
```
chlowell, line 78: Since this value won't change, I would just use it directly in all the tests and forego this field. If you really want the field, please also add it to `adfsLiveUser` so future readers don't wonder why user auth tests use service principal configuration.

sdk/azidentity/live_test.go
```diff
@@ -56,12 +57,42 @@ var liveUser = struct {
...
+	fakeAdfsScope     = "fake.adfs.local/fake-scope/.default"
 )
 
+var adfsAuthority = os.Getenv("ADFS_AUTHORITY_HOST")
+
+var adfsLiveSP = struct {
+	tenantID    string
+	clientID    string
+	secret      string
+	pemPath     string
```
chlowell, line 74: PEM or PKCS12 will work
```suggestion
	certPath     string
```

sdk/azidentity/live_test.go
```diff
@@ -56,12 +57,42 @@ var liveUser = struct {
...
+	scope       string
+	redirectURL string
+}{
+	tenantID:    "adfs",
+	clientID:    os.Getenv("ADFS_SP_CLIENT_ID"),
+	secret:      os.Getenv("ADFS_SP_CLIENT_SECRET"),
+	pemPath:     os.Getenv("ADFS_SP_CERT_PEM"),
+	scope:       os.Getenv("ADFS_SCOPE"),
```
chlowell, line 82: I'm thinking this should be a package-level variable (`var adfsScope = os.Getenv("ADFS_SCOPE")`) or duplicated on `adfsfLiveUser`

sdk/azidentity/live_test.go
```diff
@@ -206,11 +253,11 @@ func testGetTokenSuccess(t *testing.T, cred azcore.TokenCredential) {
...
 		t.Fatal("ExpiresOn isn't UTC")
 	}
-	tk2, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{liveTestScope}})
+	tk2, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: scopes})
 	if err != nil {
 		t.Fatal(err)
 	}
 	if tk2.Token != tk.Token || tk2.ExpiresOn != tk.ExpiresOn {
-		t.Fatal("expected a cached token")
+		t.Fatalf("expected a cached token")
```
chlowell, line 261:
```suggestion
		t.Fatal("expected a cached token")
```

sdk/azidentity/live_test.go
```diff
@@ -111,6 +150,8 @@ func TestMain(m *testing.M) {
 			liveSP.tenantID:                                 fakeTenantID,
 			liveUser.tenantID:                               fakeTenantID,
 			liveUser.username:                               fakeUsername,
+			strings.TrimLeft(adfsLiveSP.scope, "htps:/"):    fakeAdfsScope,
+			strings.TrimLeft(adfsAuthority, "htps:/"):       fakeAdfsAuthority,
```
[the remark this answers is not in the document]
jonnguyen1: From what I've seen, it is necessary. For whatever reason, the test proxy isn't able to change the url in the token response when I make the replacement key https://url... It successfully replaces the request url, but not the response. Probably a bug worth looking into outside of this PR.

sdk/azidentity/testdata/recordings/TestDeviceCodeCredentialADFS_Live.json
```diff
@@ -0,0 +1,407 @@
...
+        "interval": 5,
+        "message": "redacted",
+        "user_code": "redacted",
+        "verification_uri": "https://fake.adfs.local/adfs/oauth2/deviceauth",
+        "verification_uri_complete": "https://fake.adfs.local/adfs/oauth2/deviceauth?user_code=TPWGLBSKM\u0026client-request-id=2a5bf874-bebf-462e-f814-0080010000ee",
+        "verification_url": "https://fake.adfs.local/adfs/oauth2/deviceauth"
+      }
+    },
+    {
+      "RequestUri": "https://fake.adfs.local/adfs/oauth2/token/",
```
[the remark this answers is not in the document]
jonnguyen1: Sounds good

sdk/azidentity/live_test.go
jonnguyen1: Changed to package level

chlowell (approved):

Issue #17906, opened 2022-05-12
Move keyvault under a security directory


jhendrixMSFT:
Per discussion we'll move all of the key vault modules.

Old location
`
github.com/Azure/azure-sdk-for-go/sdk/keyvault/...
`

New location
`
github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/...
`

gracewilcox:
Move completed

PR #19879, opened 2023-01-26
Copy keyvault/internal to security/keyvault/internal

chlowell:
Part of #17906


sdk/security/keyvault/internal/ci.keyvault.yml
```diff
@@ -0,0 +1,28 @@
...
+      - hotfix/*
+      - release/*
+  paths:
+    include:
+    - sdk/security/keyvault/internal
+
+stages:
+- template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
+  parameters:
+    ServiceDirectory: 'keyvault/internal'
```
jhendrixMSFT, line 27: I think this needs to be `security/keyvault/internal`

chlowell: Me too! I'm sure it was user error but I'll never trust find/replace again :suspect:

sdk/security/keyvault/internal/ci.keyvault.yml
```diff
@@ -0,0 +1,28 @@
+# NOTE: Please refer to https://aka.ms/azsdk/engsys/ci-yaml before editing this file.
```
jhendrixMSFT, line 1: Looks like the name of this file will need to be changed so its pipeline doesn't collide with the old one.  That, or we repurpose the old one to be used here.

CC @benbp

benbp:
@chlowell Let me know if you'd like me to update the pipeline generator tool to handle multiple `.` characters, e.g. `go - internal - security-keyvault` instead of `go - internal - securitykeyvault`.

benbp:
If we only capture the final segment, we run into name collisions in a few places. I think we can extend the regex to look for a third segment, but I need to do some more digging to figure out other components that would be affected by this. I think your workaround is ok for now.

eng/scripts/Language-Settings.ps1
```diff
@@ -37,7 +37,7 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>((?:resourcemanager|security)/)?([^/]+/)?(?<modName>[^/]+$)))")
```
weshaggard, line 40: Do we want to hardcode security here or do we want to make it more generic? 

If we do update this we also need to at least update https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227 otherwise new packages will not correctly get discovered for the package index.

benbp: https://github.com/Azure/azure-sdk-for-go/pull/19971
https://github.com/Azure/azure-sdk/pull/5534

jhendrixMSFT (approved):

PR #19882, opened 2023-01-26
APIView Comments for azblob GA

tasherif-msft:


sdk/storage/azblob/blob/models.go
```diff
@@ -34,11 +34,11 @@ type LeaseAccessConditions = exported.LeaseAccessConditions
 // ModifiedAccessConditions contains a group of parameters for specifying access conditions.
 type ModifiedAccessConditions = exported.ModifiedAccessConditions
 
-// CpkInfo contains a group of parameters for client provided encryption key.
-type CpkInfo = generated.CpkInfo
+// CPKInfo contains a group of parameters for client provided encryption key.
+type CPKInfo = generated.CpkInfo
```
jhendrixMSFT, line 38: I think the generated content needs to be transformed too in case there are exported fields with the pascal-cased names.

tasherif-msft: Should that change be on the autorest side as well? if this a rule we want to follow across GO, in the meantime I can introduce a transform

tasherif-msft: Changes applied to generated code.

jhendrixMSFT (approved):

souravgupta-msft (approved):

vibhansa-msft (approved):

siminsavani-msft (approved):

PR #19886, opened 2023-01-27
[azservicebus] Adding a timeout so link closures have a maximum of 60 seconds before they are cancelled.

richardpark-msft:
Adding a timeout so link closures have a maximum of 60 seconds before they time out. This fixes an issue that came up when I made a fix in the last release to not timeout on close (since that can leave us in an inconsistent state). This surfaced an interesting problem where customers do see Close() hang.

So this adds in a timeout instead, which will cause us to close the connection to avoid inconsistency and also to recover overall since it's likely there's a problem that will require a more thorough reset than a simple link detach/reattach would give us.

As part of this I also created some more thorough mocks that test more of the connection/open behavior as well as links, ensuring that we do cleanup and end up in the state that we want. It's not 100% complete but it's good enough for this kind of testing and can be expanded.


sdk/messaging/azservicebus/CHANGELOG.md
```diff
@@ -1,13 +1,15 @@
...
-## 1.1.5 (Unreleased)
+## 1.2.0 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
 
 ### Bugs Fixed
 
+- Links could hang when closing, preventing recovery from completing and making a link appear stalled. (PR#19886)
```
jhendrixMSFT, line 11: Do we know where/why the hang happens?  Is this a bug in go-amqp we need to fix?

sdk/messaging/azservicebus/go.mod
```diff
@@ -25,7 +24,10 @@ require (
 	nhooyr.io/websocket v1.8.6
 )
 
-require golang.org/x/net v0.0.0-20220425223048-2871e0cb64e4
+require (
+	github.com/golang/mock v1.6.0
+	github.com/google/uuid v1.1.1
```
jhendrixMSFT, line 29: Is this an indirect dependency?

richardpark-msft: Ah, I removed it. Sometimes it gets inserted when I'm trying to find the internal/uuid package.

sdk/messaging/azservicebus/CHANGELOG.md
richardpark-msft: Until recently I wasn't aware that we were getting stuck here. Interestingly, this came up when I removed the Close() cancellation behavior and saw that customers were actually getting blocked there. How we get into this state is less clear.

jhendrixMSFT: Do we at least have a call stack so we know where the hang happens?

richardpark-msft: I can't seem to repro this behavior locally. We had a couple of customers that were seeing hangs when we ignored the Close() context so it was pretty clear.

jhendrixMSFT: Is there enough info to know if the hang was within go-amqp?

richardpark-msft: No, we'd need to enable go-amqp's logging for that and just the azservicebus/azeventhubs package is verbose enough. We should talk, let's see if we can repro it more easily locally.

jhendrixMSFT (approved):
I mainly focused on the changed in amqpLinks.go as that's where the behavior changes are.

rokf:
Now that these changes have reached the `main` branch, when can we expect them to land in a new https://github.com/Azure/azure-sdk-for-go version?

richardpark-msft:
> Now that these changes have reached the `main` branch, when can we expect them to land in a new https://github.com/Azure/azure-sdk-for-go version?

Yes, it'll go out with the next regular release (next week).

PR #19888, opened 2023-01-27
Test DeviceCodeCredential with default tenant

chlowell:
The default tenant is "organizations", and MSAL may add special handling for it in the future.

### DIFF
--- a/sdk/azidentity/device_code_credential_test.go
+++ b/sdk/azidentity/device_code_credential_test.go
@@ -87,8 +87,8 @@ func TestDeviceCodeCredential_UserPromptError(t *testing.T) {
 
 func TestDeviceCodeCredential_Live(t *testing.T) {
 	mode := recording.GetRecordMode()
-	if mode != recording.PlaybackMode {
-		t.Skip("this test requires manual recording and can't pass live in CI")
+	if mode != recording.PlaybackMode && !runManualTests {
+		t.Skip("set AZIDENTITY_RUN_MANUAL_TESTS to run this test")
 	}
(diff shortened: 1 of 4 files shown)


sdk/azidentity/device_code_credential_test.go
```diff
@@ -86,18 +86,27 @@ func TestDeviceCodeCredential_UserPromptError(t *testing.T) {
...
-	o, stop := initRecording(t)
-	defer stop()
-	opts := DeviceCodeCredentialOptions{TenantID: liveUser.tenantID, ClientOptions: o}
-	if recording.GetRecordMode() == recording.PlaybackMode {
-		opts.UserPrompt = func(ctx context.Context, m DeviceCodeMessage) error { return nil }
-	}
-	cred, err := NewDeviceCodeCredential(&opts)
-	if err != nil {
-		t.Fatal(err)
+	for _, tenant := range []string{"", liveUser.tenantID} {
```
jhendrixMSFT, line 93: Based on the change description I was expecting a tenant named "organizations" to be in the list.

chlowell: I meant that when the tenant isn't specified, the credential uses "organizations". The empty string here results in `TenantID: ""` i.e., getting the default by explicitly zeroing the field. I did it this way to avoid duplicating the test body but would doing that instead help clarify what the test does?

jhendrixMSFT: I get it now, thanks for explaining.

jhendrixMSFT (approved):

Issue #19889, opened 2023-01-28
Poller executes extra GET request if original request is PUT/PATCH


gracewilcox:
When a poller PUT/PATCH request finishes and calls the poller async Result method, it'll execute a final GET request to the original URL. 

This causes an error in some cases. If you send a GET request to the URL I am currently working with, it will cause a `400 Bad Argument` error since the service was expecting a PUT request with a body. In this case, the code shouldn't send that final GET request.

Line of code in Poller: [here](https://github.com/Azure/azure-sdk-for-go/blob/73e2c0181e0c671247d301c5e2e1c9b8fe736f32/sdk/azcore/internal/pollers/async/async.go#L132)

jhendrixMSFT:
Per the [docs](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-long-running-operation) a `final-state-via` value of `azure-async-operation` does _not_ require a final GET.

tadelesh:
I've tested some API in control plane. One example [here](https://github.com/Azure/azure-rest-api-specs/blob/78ec1b99699a4bf44869bd13f1b0ed7d92a99c27/specification/network/resource-manager/Microsoft.Network/stable/2022-07-01/applicationSecurityGroup.json#L145). The return from the last call of `Azure-AsyncOperation` URI does not contain the resource info and it will cause empty return of `CreateOrUpdate`. Log as follows.

[Jan 30 13:06:08.412480] Request: ==> OUTGOING REQUEST (Try=1)
   PUT https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/chenjieshitest/providers/Microsoft.Network/applicationSecurityGroups/test-asg2?api-version=2022-07-01
   Accept: application/json
   Authorization: REDACTED
   Content-Length: 21
   Content-Type: application/json
   User-Agent: azsdk-go-armnetwork/v2.1.0 (go1.19; Windows_NT)
   --------------------------------------------------------------------------------
{"location":"eastus"}
   --------------------------------------------------------------------------------
[Jan 30 13:06:15.911467] Response: ==> REQUEST/RESPONSE (Try=1/7.4990073s, OpTime=7.4990589s) -- RESPONSE RECEIVED
   PUT https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/chenjieshitest/providers/Microsoft.Network/applicationSecurityGroups/test-asg2?api-version=2022-07-01
[...]

PR #19892, opened 2023-01-28
Fix bug in Azure-AsyncOperation polling

jhendrixMSFT:
Prefer the specified final-state-via when specified, falling back to heuristics per spec in its absense.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/19889


jhendrixMSFT:
Needs some additional testing.  It did uncover a bug in the autorest test server LRO implementation (https://github.com/Azure/autorest.testserver/pull/413).

PR #19895, opened 2023-01-30
remove item of interface{} to any

Alancere:
resolve: https://github.com/Azure/azure-sdk-for-go/issues/19830

### DIFF
--- a/eng/tools/generator/cmd/v2/common/changelogProcessor.go
+++ b/eng/tools/generator/cmd/v2/common/changelogProcessor.go
@@ -351,3 +351,19 @@ func LROFilter(changelog *model.Changelog) {
 		}
 	}
 }
+
+func InterfaceToAnyFilter(changelog *model.Changelog) {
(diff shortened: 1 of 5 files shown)


eng/tools/generator/cmd/v2/common/changelogProcessor.go
```diff
@@ -351,3 +351,19 @@ func LROFilter(changelog *model.Changelog) {
 		}
 	}
 }
+
+func InterfaceToAnyFilter(changelog *model.Changelog) {
+	if changelog.HasBreakingChanges() {
+		for structName, s := range changelog.Modified.BreakingChanges.Structs {
+			for k, v := range s.Fields {
+				if v.From == "interface{}" && v.To == "any" {
```
tadelesh, line 359: You should use fuzzy match instead of equal to do the comparison.

tadelesh (approved):

PR #19896, opened 2023-01-30
Blob Batch API Surface

souravgupta-msft:


sdk/storage/azblob/container/client.go
```diff
@@ -329,3 +329,40 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, expiry time.Tim
 
 	return endpoint, nil
 }
+
+// NewBatchBuilder creates an instance of BatchBuilder with the specified values.
+//   - cred - an Azure AD credential, typically obtained via the azidentity module
+//   - options - client options; pass nil to accept the default values
+func (c *Client) NewBatchBuilder(cred azcore.TokenCredential, options *ClientOptions) *BatchBuilder {
```
JeffreyRichter, line 336: I don't think we need ClientOptions for this since it is not really a client

sdk/storage/azblob/container/client.go
```diff
@@ -329,3 +329,40 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, expiry time.Tim
...
+
+// NewBatchBuilderWithSharedKeyCredential creates an instance of BatchBuilder with the specified values.
+//   - cred - a SharedKeyCredential created with the matching container's storage account and access key
+//   - options - client options; pass nil to accept the default values
+func (c *Client) NewBatchBuilderWithSharedKeyCredential(cred *SharedKeyCredential, options *ClientOptions) *BatchBuilder {
+	return nil
+}
+
+// NewBatchBuilderWithNoCredential creates an instance of BatchBuilder with the specified values.
+// This is used to perform batch operations anonymously or with a shared access signature (SAS) token.
```
JeffreyRichter, line 348: I don't think tis works "anonymously"; I think it has to be SAS.

sdk/storage/azblob/container/client.go
```diff
@@ -329,3 +329,40 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, expiry time.Tim
...
+
+// SetTier operation is used to add set tier sub-request to the batch builder.
+func (bb *BatchBuilder) SetTier(blobName string, accessTier blob.AccessTier, options *BatchSetTierOptions) {
+}
+
+// SubmitBatch operation allows multiple API calls to be embedded into a single HTTP request.
+// It builds the request body using the BatchBuilder object passed.
+// BatchBuilder contains the list of operations to be submitted. It supports up to 256 sub-requests in a single batch.
+// For more information, see https://docs.microsoft.com/rest/api/storageservices/blob-batch.
+func (c *Client) SubmitBatch(ctx context.Context, bb *BatchBuilder) (SubmitBatchResponse, error) {
```
JeffreyRichter, line 366: This method should probably take a SubmitBatchOptions for the last parameter for future-proofing.

sdk/storage/azblob/service/client.go
```diff
@@ -281,3 +281,40 @@ func (s *Client) FilterBlobs(ctx context.Context, where string, o *FilterBlobsOp
...
+
+// NewBatchBuilderWithSharedKeyCredential creates an instance of BatchBuilder with the specified values.
+//   - cred - a SharedKeyCredential created with the matching container's storage account and access key
+//   - options - client options; pass nil to accept the default values
+func (s *Client) NewBatchBuilderWithSharedKeyCredential(cred *SharedKeyCredential, options *ClientOptions) *BatchBuilder {
+	return nil
+}
+
+// NewBatchBuilderWithNoCredential creates an instance of BatchBuilder with the specified values.
+// This is used to perform batch operations anonymously or with a shared access signature (SAS) token.
```
JeffreyRichter, line 300: Anonymous?

sdk/storage/azblob/service/client.go
```diff
@@ -281,3 +281,40 @@ func (s *Client) FilterBlobs(ctx context.Context, where string, o *FilterBlobsOp
...
+
+// SetTier operation is used to add set tier sub-request to the batch builder.
+func (bb *BatchBuilder) SetTier(containerName string, blobName string, accessTier blob.AccessTier, options *BatchSetTierOptions) {
+}
+
+// SubmitBatch operation allows multiple API calls to be embedded into a single HTTP request.
+// It builds the request body using the BatchBuilder object passed.
+// BatchBuilder contains the list of operations to be submitted. It supports up to 256 sub-requests in a single batch.
+// For more information, see https://docs.microsoft.com/rest/api/storageservices/blob-batch.
+func (s *Client) SubmitBatch(ctx context.Context, bb *BatchBuilder) (SubmitBatchResponse, error) {
```
JeffreyRichter, line 318: Options param

sdk/storage/azblob/container/client.go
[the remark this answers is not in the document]
souravgupta-msft: BatchBuilder object has endpoint as one of its private fields which is used to create the delete or set tier sub-requests in the `BatchBuilder.Delete()` or `BatchBuilder.SetTier()` methods.
This endpoint is coming from the client.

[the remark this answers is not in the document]
jhendrixMSFT: So the created BatchBuilder can only be used to batch operations on blobs within this container, is that correct?  If so, then this should be included in the docs.

sdk/storage/azblob/container/client.go
souravgupta-msft: If public access is enabled on the storage account, then I think that the sub-requests won't need to have the authorization header. This part we can double check while implementing this.

sdk/storage/azblob/service/client.go
souravgupta-msft: Same as previous. This case can happen when public access is enabled on the storage account.
We will confirm this part while implementing this.

sdk/storage/azblob/container/client.go
JeffreyRichter: According to this https://learn.microsoft.com/en-us/rest/api/storageservices/create-container?tabs=azure-ad, public access means public READ access and so delete is not allowed and I doubt set tier is allowed as well.

sdk/storage/azblob/container/client.go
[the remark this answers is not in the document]
souravgupta-msft: Yes, the BatchBuilder can only be used to batch the delete or set tier sub-requests.
Will add relevant examples for this.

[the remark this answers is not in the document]
jhendrixMSFT: In addition to the examples, which people might not see, please be sure to update the docs with this info.

sdk/storage/azblob/service/client.go
```diff
@@ -281,3 +281,40 @@ func (s *Client) FilterBlobs(ctx context.Context, where string, o *FilterBlobsOp
 	resp, err := s.generated().FilterBlobs(ctx, where, serviceFilterBlobsOptions)
 	return resp, err
 }
+
+// NewBatchBuilder creates an instance of BatchBuilder with the specified values.
+//   - cred - an Azure AD credential, typically obtained via the azidentity module
+//   - options - client options; pass nil to accept the default values
+func (s *Client) NewBatchBuilder(cred azcore.TokenCredential, options *ClientOptions) *BatchBuilder {
```
jhendrixMSFT, line 288: Similar to the other constructor, add docs about the behavior.

sdk/storage/azblob/container/client.go
jhendrixMSFT: Presumably we'd have `BatchBuilderOptions` though yes?

sdk/storage/azblob/container/client.go
souravgupta-msft: Done. Renamed the method to `NewBatchBuilderWithSAS`.

sdk/storage/azblob/container/client.go
souravgupta-msft: Added.

sdk/storage/azblob/service/client.go
souravgupta-msft: Done. Renamed the method to `NewBatchBuilderWithSAS`.

sdk/storage/azblob/service/client.go
souravgupta-msft: Done.

sdk/storage/azblob/service/client.go
souravgupta-msft: Added.

sdk/storage/azblob/container/client.go
JeffreyRichter: We could but I don't think it's necessary. I'm struggling to think of anything that could show up here in the future. The BatchBuilder is really just concatenating items together - it never talks to a service. We'd have to have an option that applies to the entire batch (not individual items).

jhendrixMSFT: I see.  I suppose we also have an escape hatch; we could add a method to the BB to configure it post-construction if necessary.

sdk/storage/azblob/container/client.go
JeffreyRichter: This name implies that the SAS is passed to the ctor and never passed anywhere else. Actually, how would a SAS be passed? A SAS for each item? If so, how? If the SAS is the same for all items, then it needs to be passed to this ctor as a parameter. Would it be a string or a structure with fields set?

souravgupta-msft: The method `NewBatchBuilderWithSAS` does not take any parameters. It gets the SAS from the client's endpoint. Here the customer has to make sure that the client is created using SAS (added this as comment). We can rename this method to `NewBatchBuilderWithNoCredential` if we don't want to pass the SAS as a parameter.

If we do want to pass SAS, we can pass it to the ctor as a parameter and it will be of type string since it can be account, service or user delegation SAS.

JeffreyRichter: If you do it this way, there are some inconsistencies. First, all the other BB ctors take creds used for the items and this ctor would not. Second, when this ctor is called, we don't know if the client was created with SAS or not so we can't guarantee success. We could rename this ctor to emphasise the inconsistency something like NewBuilderUsingClientSAS (I don't like this but we can think about it more) and have this ctor return a (BB, error) with a non-nil error if the client doesn't have a SAS.

JeffreyRichter: Or, maybe we want to be consistent and let the ctor take SAS parameters which can be different than the SAS on the client?

souravgupta-msft: I like the second option where we can take SAS as a parameter which will be of `string` type.

jhendrixMSFT: What about `sas.QueryParameters` instead of `string`?  Would that be better?

sdk/storage/azblob/container/client.go
jhendrixMSFT: Agreed.  Then I worry that omitting an options type might pidgeon-hole us in the future.

sdk/storage/azblob/container/client.go
JeffreyRichter: Yes, I think this is good

souravgupta-msft: Added `sas.QueryParameters` as a parameter to the ctor `NewBatchBuilderWithSAS`.

sdk/storage/azblob/service/client.go
souravgupta-msft: Same as container client.

sdk/storage/azblob/container/client.go
vibhansa-msft: @JeffreyRichter: In case user has a custom user-agent string that is set using these client-options and same is case for custom pipeline policies. If we do not take the client-options we lose this flexibility. For e.g. user-agent will be marked only on the outer request and not at every sub-request level. This means when FE iterates the batch and executes individual sub-request user-agent will not be present and this might skew some of the metrics collected for billing and analysis purpose (not sure each individual sub-request is billed separately or not, but assuming that will be the case).

JeffreyRichter: The BatchBuilder's pipeline is not for sending a request to a service; it is for building an in-memory buffer composed of several batch items that ultimately SubmitBatch will send to a service There is no such concept of a user-agent string for the BatchBuilder. For the same reason, there is no need for customer customization of the BatchBuilder's pipeline. 

The purpose of the UA string is for the service to know what SDK is making the request; it doesn't make sense for each item to be generated by different SDKs. The customer is billed for each SubmitBatch operation; they are NOT billed for each item in the batch. But, this also has nothing to do with user-agent headers. 

As a side note: A service can NEVER trust data provided by a customer. If the UA string were used for billing, then customers would set this to whatever they want for their advantage - services can never be designed to trust this.

sdk/storage/azblob/internal/base/clients.go
```diff
@@ -7,14 +7,16 @@
...
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/exported"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 )
 
 type Client[T any] struct {
-	inner     *T
-	sharedKey *exported.SharedKeyCredential
+	inner      *T
+	sharedKey  *exported.SharedKeyCredential
+	authPolicy policy.Policy
```
souravgupta-msft, line 19: We should have a `credential any` field here instead of `sharedKey` and `authPolicy` fields.
cc: @JeffreyRichter, @jhendrixMSFT

jhendrixMSFT: Good idea, I like it.

sdk/storage/azblob/container/client.go
```diff
@@ -329,3 +333,40 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, expiry time.Tim
...
+func (c *Client) NewBatchBuilder() (*BatchBuilder, error) {
+	conOptions := new(ClientOptions)
+	if c.authPolicy() != nil {
+		conOptions.PerRetryPolicies = append(conOptions.PerRetryPolicies, c.authPolicy())
+	}
+
+	// TODO: Use an empty transport so requests aren't sent
+	conOptions.Transport = nil
+
+	pl := runtime.NewPipeline(exported.ModuleName, exported.ModuleVersion, runtime.PipelineOptions{}, &conOptions.ClientOptions)
```
jhendrixMSFT, line 348: I suspect that you don't really need to create a full pipeline as it contains a bunch of policies that aren't useful for this.  You should just be able to call `Policy.Do()` in order to add the necessary auth content to the fake request.

sdk/storage/azblob/internal/base/clients.go
jhendrixMSFT: One thing I will point out is that with this solution, you don't know how the underlying credential policy will mutate the fake HTTP request with the credential info.  So you might just be delaying the switch/case to after the writing instead of before.

If you have the concrete credential type up front, then you know the type and how it writes its authentication info.

sdk/storage/azblob/container/client.go
souravgupta-msft: Do you mean we should not create a new pipeline for the BatchBuilder and directly use `Policy.Do()`. If this is the case, we don't need `pipeline` field in the BatchBuilder. It will look like this,
```
type BatchBuilder struct {
	endpoint    string
	subRequests []*policy.Request
}
```

If we do want to have a separate pipeline for the BatchBuilder, can you suggest some way to create a pipeline having just the auth policy (we can add a dummy transport policy to it if required).

jhendrixMSFT: There is no way to create an empty pipeline.  We _could_ add this if necessary but I'm not convinced we need it.

sdk/storage/azblob/container/responses.go
```diff
@@ -42,3 +43,12 @@ type GetAccessPolicyResponse = generated.ContainerClientGetAccessPolicyResponse
 
 // SetAccessPolicyResponse contains the response from method Client.SetAccessPolicy.
 type SetAccessPolicyResponse = generated.ContainerClientSetAccessPolicyResponse
+
+// SubmitBatchResponse contains the response from method Client.SubmitBatch.
+type SubmitBatchResponse struct {
+	generated.ContainerClientSubmitBatchResponse
```
jhendrixMSFT, line 49: We typically don't embed unexported types in public surface area.  Can't the generated type be used and exported?

sdk/storage/azblob/internal/shared/blob_batch.go
```diff
@@ -0,0 +1,286 @@
...
+	// remove x-ms-version header from the request header
+	for k := range req.Raw().Header {
+		if strings.EqualFold(k, HeaderXmsVersion) {
+			delete(req.Raw().Header, k)
+		}
+	}
+}
+
+// BlobBatchResponse contains the response for the individual sub-requests.
+type BlobBatchResponse struct {
```
jhendrixMSFT, line 168: It looks like this, and maybe other types, are directly exported via type alias.  In this case, the content should be moved to the `exported` package so it's clear this is part of public surface area.  The `shared` package is for internal-only content that's shared across packages.

sdk/storage/azblob/internal/shared/log_events.go
```diff
@@ -1,7 +1,10 @@
+//go:build go1.18
+// +build go1.18
+
 // Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
+// Licensed under the MIT License. See License.txt in the project root for license information.
 
-package exported
+package shared
```
jhendrixMSFT, line 7: This should go back to the `exported` package after the batching building content has been moved.

jhendrixMSFT (changes requested):
Some internal content needs to be moved to the `exported` package.

sdk/storage/azblob/container/responses.go
souravgupta-msft: The response body that comes from the generated code is like [this](https://learn.microsoft.com/en-us/rest/api/storageservices/blob-batch#sample-response).
Here I am parsing this response body and storing the responses of the individual sub-requests. The reason why I have kept `generated.ContainerClientSubmitBatchResponse` is because if the parsing of the sub-request's response results in error due to unknown reasons, the users can always refer the raw response from the `generated.ContainerClientSubmitBatchResponse`.

If we want to export the generated response, we can do something like this.
```
type BatchRawResponse = generated.ContainerClientSubmitBatchResponse

// SubmitBatchResponse contains the response from method Client.SubmitBatch.
type SubmitBatchResponse struct {
	BatchRawResponse
	Responses []*BatchSubResponse
}
```

sdk/storage/azblob/internal/shared/blob_batch.go
souravgupta-msft: Will move this to `exported` package.

sdk/storage/azblob/container/responses.go
jhendrixMSFT: If the parsing fails, I would expect a zero-value `SubmitBatchResponse` to be returned a long with the appropriate `error`.

sdk/storage/azblob/container/client_test.go
```diff
@@ -2222,3 +2224,786 @@ func (s *ContainerUnrecordedTestsSuite) TestSASContainerClient() {
...
+			}
+		}
+	}
+	_require.Equal(ctrHot, 0)
+	_require.Equal(ctrCool, 10)
+}
+
+/*
+func (s *ContainerUnrecordedTestsSuite) TestContainerBlobBatchDeleteUsingTokenCredential() {
+	_require := require.New(s.T())
```
siminsavani-msft, line 2355: Is there a reason for leaving token credential tests out?

siminsavani-msft (approved):

tasherif-msft (approved):

sdk/storage/azblob/container/client_test.go
souravgupta-msft: I commented this part because we currently don't have the OAuth credentials in our DevOps pipeline.
Will uncomment these tests once it is added.

sdk/storage/azblob/container/responses.go
souravgupta-msft: Here we are talking about parsing of the individual responses for the sub-requests. If parsing fails for one of the sub-request's response, then we should not fail the entire batch IMO. For these cases, we will store the error in the individual response object(`BatchSubResponse`) in the `Error` field. Users can always refer the raw response stored in the `BatchRawResponse` field of the `SubmitBatchResponse`.
```
// BatchSubResponse contains the response for the individual sub-requests.
type BatchSubResponse struct {
	ContentID     *int
	ContainerName *string
	BlobName      *string
	Response      *http.Response
	Error         error
}
```

vibhansa-msft (approved):

sdk/storage/azblob/internal/shared/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/internal/shared/log_events.go
souravgupta-msft: Done

sdk/storage/azblob/container/models.go
```diff
@@ -329,3 +332,79 @@ func formatTime(c *SignedIdentifier) error {
 
 	return nil
 }
+
+// ---------------------------------------------------------------------------------------------------------------------
+
+// BatchBuilder is used for creating the batch operations list. It contains the list of either delete or set tier sub-requests.
+// NOTE: All sub-requests in the batch must be of the same type, either delete or set tier.
+type BatchBuilder struct {
```
jhendrixMSFT, line 340: The convention is to keep the type definition and its methods together in the same file.

Since there's a decent amount of code for `BatchBuilder` I would suggest placing it all into a `batch_builder.go` file.

sdk/storage/azblob/container/models.go
```diff
@@ -329,3 +332,79 @@ func formatTime(c *SignedIdentifier) error {
...
+
+	leaseAccessConditions, modifiedAccessConditions := exported.FormatBlobAccessConditions(o.AccessConditions)
+	return &basics, leaseAccessConditions, modifiedAccessConditions
+}
+
+// BatchSetTierOptions contains the optional parameters for the BatchBuilder.SetTier method.
+type BatchSetTierOptions struct {
+	VersionID *string
+	Snapshot  *string
+	*blob.SetTierOptions
```
jhendrixMSFT, line 385: The convention is to place anonymously embedded types as the first field.

Also, this doesn't need to be pointer-to-type.

sdk/storage/azblob/container/responses.go
jhendrixMSFT: In Go, when a func/method returns `(Foo, error)`, and the `error` value is not `nil`, callers must _not_ inspect the value of `Foo` as it's unspecified.  It's best to make this the zero-value so we don't inadvertently let callers violate this principle.

If `SubmitBatch()` fails to parse the response, we want to directly return the parsing error so it's clear we have a bug that needs to be fixed (we do the same thing for all of the other APIs).

Regarding the raw response, we've removed this from the success case.  If callers need access to the raw HTTP response, they must use [`runtime.WithCaptureResponse`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.3.1/runtime#WithCaptureResponse) to get it.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,287 @@
...
+		return "", err
+	}
+
+	return BatchIdPrefix + batchID.String(), nil
+}
+
+// getBatchRequestDelimiter is used for creating the batch boundary
+// e.g. --batch_357de4f7-6d0b-4e02-8cd2-6361411a9525
+// last line of the request body: --batch_357de4f7-6d0b-4e02-8cd2-6361411a9525--
+func getBatchRequestDelimiter(batchID *string, prefixDash bool, postfixDash bool) string {
```
jhendrixMSFT, line 54: Take a look at the [`multipart`](https://pkg.go.dev/mime/multipart) package in the standard library.  You can use it for reading/writing multipart form data.

sdk/storage/azblob/container/models.go
souravgupta-msft: Done

sdk/storage/azblob/container/models.go
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/container/client.go
```diff
@@ -329,3 +342,68 @@ func (c *Client) GetSASURL(permissions sas.ContainerPermissions, expiry time.Tim
...
+	switch cred := c.credential().(type) {
+	case *azcore.TokenCredential:
+		authPolicy = runtime.NewBearerTokenPolicy(*cred, []string{shared.TokenScope}, nil)
+	case *SharedKeyCredential:
+		authPolicy = exported.NewSharedKeyCredPolicy(cred)
+	case nil:
+		// for authentication using SAS
+		authPolicy = nil
+	default:
+		panic(fmt.Sprintf("unrecognised authentication type %T", cred))
```
jhendrixMSFT, line 361: This should return an error, not panic.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	// remove x-ms-version header from the request header
+	for k := range req.Raw().Header {
+		if strings.EqualFold(k, shared.HeaderXmsVersion) {
+			delete(req.Raw().Header, k)
+		}
+	}
+}
+
+// BlobBatchSubResponse contains the response for the individual sub-requests.
+type BlobBatchSubResponse struct {
```
jhendrixMSFT, line 158: How about we call this `BatchResponseItem` instead?

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+const (
+	BatchIdPrefix                                    = "batch_"
+	HttpVersion                                      = "HTTP/1.1"
+	HttpNewline                                      = "\r\n"
+	BatchDeleteOperationType  BlobBatchOperationType = "delete"
+	BatchSetTierOperationType BlobBatchOperationType = "set tier"
+)
+
+type BlobBatchBuilder struct {
+	Endpoint    *string
```
jhendrixMSFT, line 39: Does this really need to be a pointer?

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	AuthPolicy  policy.Policy
+	SubRequests []*policy.Request
```
jhendrixMSFT, line 41: Interface types can be `nil` so this shouldn't need to be a pointer.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+
+	partHeaders := make(textproto.MIMEHeader)
+	partHeaders["Content-Type"] = []string{"application/http"}
+	partHeaders["Content-Transfer-Encoding"] = []string{"binary"}
+	var partWriter io.Writer
+
+	for i, req := range bb.SubRequests {
+		if bb.AuthPolicy != nil {
+			resp, err := bb.AuthPolicy.Do(req)
+			if err != nil && resp != nil {
```
jhendrixMSFT, line 118: I don't think this is correct as the returned `resp` and `err` are mutually exclusive.

Also, if `err` isn't `nil`, shouldn't the error be returned here?

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	val, err := strconv.Atoi(strings.TrimSpace(*contentID))
+	if err != nil {
+		return nil, err
+	}
+	return &val, nil
+}
+
+func getResponseHeader(key string, headers map[string][]string) *string {
```
jhendrixMSFT, line 191: I don't think this is necessary.  Can't you use the `Get()` method on `http.Header`?

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+type BlobBatchSubResponse struct {
+	ContentID     *int
+	ContainerName *string
+	BlobName      *string
+	RequestID     *string
+	Version       *string
+	Error         error // nil error indicates that the batch sub-request operation is successful
+}
+
+func getResponseBoundary(contentType *string) (string, error) {
```
jhendrixMSFT, line 167: See `mime.ParseMediaType()` to get the response boundary.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	respReader := multipart.NewReader(respBody, boundary)
+	var responses []*BlobBatchSubResponse
+
+	for {
+		part, err := respReader.NextPart()
+		if err != nil {
```
jhendrixMSFT, line 211: This can be simplified.
```go
if errors.Is(err, io.EOF) {
	break
} else if err != nil {
	return nil, err
}
```
Generally speaking, to check for sentinel error types, use `errors.Is()`.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	Error         error // nil error indicates that the batch sub-request operation is successful
+}
+
+func getResponseBoundary(contentType *string) (string, error) {
+	if contentType == nil {
+		return "", fmt.Errorf("Content-Type returned in SubmitBatch response is nil")
+	}
+	boundaryIdx := strings.Index(*contentType, "batchresponse_")
+	if boundaryIdx == -1 {
+		return "", fmt.Errorf("batch boundary not present in Content-Type header of the SubmitBatch response.\nContent-Type: %v\n", *contentType)
```
jhendrixMSFT, line 173: The trailing `\n` can be removed.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		if batchSubResponse.ContentID != nil {
+			path := strings.Trim(subRequests[*batchSubResponse.ContentID].Raw().URL.Path, "/")
+			p := strings.Split(path, "/")
+			batchSubResponse.ContainerName = to.Ptr(p[0])
+			batchSubResponse.BlobName = to.Ptr(strings.Join(p[1:], "/"))
+		}
+
+		respBytes, err := ioutil.ReadAll(part)
```
jhendrixMSFT, line 232: The `ioutil` package has been deprecated, use `io` instead (I'm surprised the linter isn't complaining about this).

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+		respBytes = append(respBytes, byte('\n'))
+		buf := bytes.NewBuffer(respBytes)
+		resp, err := http.ReadResponse(bufio.NewReader(buf), nil)
+		// sub-response parsing error
+		if err != nil {
+			return nil, err
+		}
+
+		if resp != nil {
```
jhendrixMSFT, line 244: Would this ever be `nil` if `err` above is not `nil`?  If it was `nil`, then it would appear that there would be an empty `BlobBatchSubResponse` entry in the slice.

benbp:
@jhendrixMSFT looking into the inner correlation ID, it's a quota issue:

> "statusMessage":"{"error":{"code":"InvalidTemplateDeployment","message":"The template deployment 't09f8def0cc4f4419' is not valid according to the validation procedure. The tracking id is 'c67e6b9d-d9cf-4655-8117-701b697f88ee'. See inner errors for details.","details":[{"code":"PreflightValidationCheckFailed","message":"Preflight validation failed. Please refer to the details for the specific errors.","details":[{"code":"MaxStorageAccountsCountPerSubscriptionExceeded","target":"2cd617ea-1866-46b1-90e3-fffb087ebf9b",**"message":"Subscription 2cd617ea-1866-46b1-90e3-fffb087ebf9b already contains 244 storage accounts with Standard Dns endpoints in location westus and the maximum allowed is 250."**}]}]}}"

Right now the go storage tests have a leak because they create immutable containers. We need to clean these up automatically on our end or ideally in the live test side, but let me go and do a sweep to clean some of the existing ones up.

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: `Get(key)` method on `http.Header` canonicalizes the key. This was why I have the method `getResponseHeader()` in place of that. I will check if `Get(key)` can used instead of this method.

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Renamed to `BatchResponseItem`

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,269 @@
...
+	"net/http"
+	"net/textproto"
+	"strconv"
+	"strings"
+)
+
+type BlobBatchOperationType string
+
+const (
+	BatchIdPrefix                                    = "batch_"
```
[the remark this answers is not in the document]
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Removed the `Endpoint` field since it is not being used in `exported/blob_batch.go`.

sdk/storage/azblob/service/client.go
```diff
@@ -280,3 +293,69 @@ func (s *Client) FilterBlobs(ctx context.Context, where string, o *FilterBlobsOp
...
+	switch cred := s.credential().(type) {
+	case *azcore.TokenCredential:
+		authPolicy = runtime.NewBearerTokenPolicy(*cred, []string{shared.TokenScope}, nil)
+	case *SharedKeyCredential:
+		authPolicy = exported.NewSharedKeyCredPolicy(cred)
+	case nil:
+		// for authentication using SAS
+		authPolicy = nil
+	default:
+		panic(fmt.Sprintf("unrecognised authentication type %T", cred))
```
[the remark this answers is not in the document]
souravgupta-msft: Done. Returning error.

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Updated the logic to use `Get()` method of `http.Header` here.

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Removed

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Updated from `ioutil` to `io`

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Removed the `nil` check for `resp`

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: The auth policy is just executed here which has the following code: `res, err := req.Next()`
The `Next()` method has this check,
```
if len(req.policies) == 0 {
	return nil, errors.New("no more policies")
}
```
So, `bb.AuthPolicy.Do(req)` will always return the error `errors.New("no more policies")`, since the auth policy is directly executed here.

We can update the logic to do something like this,
```
if err != nil && !errors.Is(err, errors.New("no more policies")) {
	if log.Should(EventSubmitBatch) {
		log.Writef(EventSubmitBatch, "failed to authorize sub-request for %v.\nError: %v", req.Raw().URL.Path, err.Error())
	}
}
```

@jhendrixMSFT , should we fail the entire batch here if the auth policy is failing for a single sub-request?

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,271 @@
...
+	"strings"
+)
+
+type BlobBatchOperationType string
+
+const (
+	batchIdPrefix                                    = "batch_"
+	httpVersion                                      = "HTTP/1.1"
+	httpNewline                                      = "\r\n"
+	BatchDeleteOperationType  BlobBatchOperationType = "delete"
```
jhendrixMSFT, line 35: Nit.  The convention is to separate logically grouped constants into their own `const` block.  And since these are strongly typed, the block should go under the respective type.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,271 @@
...
+	"io"
+	"mime"
+	"mime/multipart"
+	"net/http"
+	"net/textproto"
+	"strconv"
+	"strings"
+)
+
+type BlobBatchOperationType string
```
jhendrixMSFT, line 29: Since this isn't actually exported, but is required by exported content, it would be good to add a comment explaining that.  In fact. if there's other content in here that's not exported, it would be best to group them all together at the end of the file ([example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/exported/request.go#L173-L182)).

sdk/storage/azblob/internal/exported/blob_batch.go
jhendrixMSFT: I think it should fail.  The reason is because in the singular case, if the auth policy fails the request fails, and I think having this be consistent is good.

sdk/storage/azblob/internal/exported/blob_batch.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	return reqBody.Bytes(), batchID, nil
+}
+
+// UpdateSubRequestHeaders updates the sub-request headers.
+// Removes x-ms-version header.
+func UpdateSubRequestHeaders(req *policy.Request) {
+	// remove x-ms-version header from the request header
+	for k := range req.Raw().Header {
```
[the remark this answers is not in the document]
souravgupta-msft: The reason why I am doing `range` over `map` is because of the canonicalization of the HTTP headers.
When the request is created, the `x-ms-version` is stored as,
```
req.Raw().Header["x-ms-version"] = []string{"2020-10-02"}
```
So, if we do `req.Header.Get("x-ms-version")`, it will always return empty string because internally it will canonicalize the `x-ms-version` to `X-Ms-Version` for which we don't have an entry in the `Header` map.

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
souravgupta-msft: Done

sdk/storage/azblob/internal/exported/blob_batch.go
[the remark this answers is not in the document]
jhendrixMSFT: Oh yes I forgot we changed the codegen to not use `Header.Set()`.

jhendrixMSFT (approved):
This is looking great.

Can you please update the APIView with the latest changes?  Once this has been updated, @JeffreyRichter and I will take another look.
Also, we need to decide if you want to release this as a beta or not.  If you do, we'll need to create a beta branch.

PR #19899, opened 2023-01-31
Support ARM cross-tenant authentication

chlowell:
Copying #19309 to the azcore beta branch


chlowell:
azappconfig and aztables live tests are failing against main for the same reasons so, this PR doesn't break them.

jhendrixMSFT (approved):

Issue #14931, opened 2021-06-29
Add support for CAE 


catalinaperalta:
Use Azure Identity crew's CAE onboarding page for guidance in the OneNote.

chlowell:
Blocked on https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/263

baywet:
related #14284

baywet:
Additionally to passing the claim, azure identity should set the capabilities [in places like these](https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azidentity/client_certificate_credential.go#L72) (`WithClientCapabilities(new [] {"cp1"})`) [source](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-resilience-continuous-access-evaluation).

It seems that there's no way today for somebody using azure identity to set that capability.

baywet:
see the equivalent for dotnet https://github.com/Azure/azure-sdk-for-net/blob/c7f68fb15a2dc8dace6e61f78b5b0f7425ec2d4c/sdk/identity/Azure.Identity/src/MsalPublicClient.cs#L46

chlowell:
This support is available now: with azidentity v1.3.0-beta.3 and azcore v1.4.0-beta.1, credentials default to sending CP1 and ARM clients internally handle claims challenges.

PR #19902, opened 2023-01-31
CAE support for ARM clients

chlowell:
This is the azcore side of #14931. I'm targeting the beta branch because the end-to-end scenario requires an azidentity beta. ARM is the only SDK service that supports CAE (continuous access evaluation) today. Briefly, it works like this: ARM evaluates security policies for the access token attached to each request. When a token doesn't satisfy policy, ARM responds 401 with an authentication challenge. The client must retry the request with a new access token having the claims specified by the challenge. So, this PR adds challenge handling to ARM's auth policy, and a `TokenRequestOptions.Claims` field that enables the policy to request claims when calling `GetToken()`.

### DIFF
--- a/sdk/azcore/arm/runtime/policy_bearer_token.go
+++ b/sdk/azcore/arm/runtime/policy_bearer_token.go
@@ -74,10 +74,16 @@ func NewBearerTokenPolicy(cred azcore.TokenCredential, opts *armpolicy.BearerTok
 
 func (b *BearerTokenPolicy) onChallenge(req *azpolicy.Request, res *http.Response, authNZ func(azpolicy.TokenRequestOptions) error) error {
 	challenge := res.Header.Get(shared.HeaderWWWAuthenticate)
-	if claims := parseChallenge(challenge); claims != "" {
+	claims, err := parseChallenge(challenge)
+	if err != nil {
+		// the challenge contains claims we can't parse
+		return err
+	} else if claims != "" {
+		// request a new token having the specified claims, send the request again
 		return authNZ(azpolicy.TokenRequestOptions{Claims: claims, Scopes: b.scopes})
 	}
-	return fmt.Errorf("failed to parse authentication challenge %q", challenge)
+	// auth challenge didn't include claims, so this is a simple authorization failure
+	return azruntime.NewResponseError(res)
 }
 
 // onRequest authorizes requests with one or more bearer tokens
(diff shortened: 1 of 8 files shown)


sdk/azcore/arm/runtime/policy_bearer_token.go
```diff
@@ -99,3 +109,26 @@ func (b *BearerTokenPolicy) onRequest(req *azpolicy.Request, authNZ func(azpolic
 func (b *BearerTokenPolicy) Do(req *azpolicy.Request) (*http.Response, error) {
 	return b.btp.Do(req)
 }
+
+// parseChallenge parses claims from an authentication challenge so a client can request a token that will satisfy
+// conditional access policies. Returns an empty string when no claims are found or claims aren't in ARM's format.
+// This function isn't universally applicable because RPs differ in challenge format and content.
```
jhendrixMSFT, line 115: What are the implications of this this, does it mean that this might fail in some way for certain RPs?

chlowell: Yes. This function can't parse a Storage challenge, for example. The implication is that we can't simply copy or export this function when another RP adds CAE support.

jhendrixMSFT: Does it work for all of ARM though?  The fact that it's part of ARM's bearer token policy makes me think yes.

chlowell: It should. At least, that's my assumption given no one on the ARM or Identity side mentioned partial support when we first implemented this a couple years ago.

jhendrixMSFT: ok I get it now, thanks for explaining.

sdk/azcore/arm/runtime/policy_bearer_token.go
```diff
@@ -99,3 +109,26 @@ func (b *BearerTokenPolicy) onRequest(req *azpolicy.Request, authNZ func(azpolic
...
+		if _, after, found := strings.Cut(param, "claims="); found {
+			if claims != "" {
+				// The header contains multiple challenges, at least two of which specify claims. The specs allow this
+				// but it's unclear what a client should do in this case and there's as yet no concrete example of it.
+				return ""
+			}
+			// trim stuff that would get an error from RawURLEncoding; claims may or may not be padded
+			claims = strings.Trim(after, `\"=`)
+			// we don't return the error because when not nil it's something unhelpful like "illegal base64 data at input byte 42"
+			if b, err := base64.RawURLEncoding.DecodeString(claims); err == nil {
```
jhendrixMSFT, line 128: If this does fail, is the assumption that the claim isn't encoded?

chlowell: Yes. That would be surprising because all claims from ARM should be base64 encoded.

After reading this again, I made some changes to distinguish failure to parse claims from absence of claims and enable the policy to return a `ResponseError` in the latter case, as clients do today. (ARM includes claims in the challenge only when it rejects a request due to a conditional access policy.)

jhendrixMSFT (approved):

PR #19906, opened 2023-01-31
Blobs GA Final APIView comments

tasherif-msft:
- Rename CorsRule -> CORSRule
- Remove Account Services


sdk/storage/azblob/appendblob/client_test.go
```diff
@@ -1771,7 +1771,6 @@ func (s *AppendBlobRecordedTestsSuite) TestAppendBlockPermanentDelete() {
 		Protocol:      sas.ProtocolHTTPS,                    // Users MUST use HTTPS (not HTTP)
 		ExpiryTime:    time.Now().UTC().Add(48 * time.Hour), // 48-hours before expiration
 		Permissions:   to.Ptr(sas.AccountPermissions{Read: true, List: true, PermanentDelete: true}).String(),
-		Services:      to.Ptr(sas.AccountServices{Blob: true}).String(),
```
gapra-msft, line 1774: why are we deleting the services from the account sas? Isnt this required?

tasherif-msft: since its the blobs package, we should only support the blob service (hardcode it), 
tables used to be here too, but it no longer is - it would be a breaking change to remove it. Better to not have it and add it later if demanded than break and remove it..

jhendrixMSFT (approved):

souravgupta-msft (approved):

PR #19920, opened 2023-02-01
restore all url with pkg.go.dev for ACR readme

tadelesh:
restore all url with pkg.go.dev as the package has already been indexed

### DIFF
--- a/sdk/containers/azcontainerregistry/README.md
+++ b/sdk/containers/azcontainerregistry/README.md
@@ -10,7 +10,7 @@ Use the client library for Azure Container Registry to:
 - Delete images and artifacts, repositories and tags
 - Upload and download images
 
-[Source code](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/containers/azcontainerregistry) | [REST API documentation](https://docs.microsoft.com/rest/api/containerregistry/) | [Product documentation](https://docs.microsoft.com/azure/container-registry/)
+[Source code](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/containers/azcontainerregistry) | [Package (pkg.go.dev)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry) | [REST API documentation](https://docs.microsoft.com/rest/api/containerregistry/) | [Product documentation](https://docs.microsoft.com/azure/container-registry/)
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Strange that it's been indexed before tagging.  The version is a hash, so I wonder if somebody grabbed it pre-release?

tadelesh:
> Strange that it's been indexed before tagging. The version is a hash, so I wonder if somebody grabbed it pre-release?

I grabbed it to check the indexed content, also I want to add the pkg url in readme for the first preview.

sdk/containers/azcontainerregistry/README.md
```diff
@@ -38,11 +38,11 @@ az acr create --name MyContainerRegistry --resource-group MyResourceGroup --loca
...
-This credential type works in both local development and production environments. 
+This document demonstrates using [azidentity.NewDefaultAzureCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#NewDefaultAzureCredential) to authenticate.
+This credential type works in both local development and production environments.
 We recommend using a [managed identity](https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/overview) in production.
 
-Client and BlobClient accepts any [azidentity][https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity] credential. 
+[Client](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry#Client) and [BlobClient](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry#BlobClient) accepts any [azidentity][https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity] credential.
 See the [azidentity](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) documentation for more information about other credential types.
```
xboxeer, line 46: can we paste env var that needed to make the default credential work quickly?

tadelesh: I don't prefer to do so. It is related with auth process that should be documented in `azidentity`.

chlowell: :+1: it's best to use `DefaultAzureCredential` as simply as possible in client library docs. We don't want to imply that environment variables are the only or the best way to configure authentication, or to describe the other options here.

xboxeer (approved):
LGTM

sdk/containers/azcontainerregistry/README.md
xboxeer: Since this is the overall policy across data plane, I'm ok we have a link to azidentity

Issue #19883, opened 2023-01-26
[azeventhubs] Move to the latest azblob surface


richardpark-msft:
The blob package is nearing release, which means we can STOP vendoring azblob internally and start targeting the newer package. This also eliminates another problem that we have where there are some credential types that aren't exported for blobs - we won't need to do this anymore.

This will end up "fixing" #19842, since the customer will have full access to the storage package, and can use any credential types in there.

richardpark-msft:
Done as of of the latest release. 

https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fmessaging%2Fazeventhubs%2Fv0.5.0

PR #19928, opened 2023-02-01
[azeventhubs] Update to latest azblob SDK

### LINKED ISSUE #19842: Event Hubs: Missing exported method "NewBlobStoreFromSharedKey" for creating checkpointer
### Bug Report

- import path of package in question: `github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/checkpoints`
- SDK version: `0.4.0`
- output of `go version`: `1.19.4`

### Description

When creating a checkpointer using Azure Blob Storage, the package `github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/checkpoints` offers two methods:

- `NewBlobStore` -> Uses Azure AD
- `NewBlobStoreFromConnectionString` -> Supports Azure Storage connection strings

A missing feature is being able to create a new blob store checkpointer from a shared key (a pair of account name and key).

I see that in the internal package that `checkpoints` references, `github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs/internal/blob`, there is already a method [`NewContainerClientWithSharedKey`](https://github.dev/Azure/azure-sdk-for-go/blob/2a983c938d002d4f27b2d277fc352b636b4d8eb5/sdk/messaging/azeventhubs/internal/blob/zc_container_client.go#L53-L54) which does precisely what we're looking for (and it depends on `NewSharedKeyCredential` which is another method in the same internal package). However, seems that there's no exported "higher level" method that leverages `NewContainerClientWithSharedKey`

richardpark-msft:
azblobs is approaching GA and we can now remove our vendored copy of azblobs and use the proper package. Our code footprint will get smaller and it'll also solve #19842, where we weren't exporting all the possible authentication for blobs.

Fixes #19842 #19883


richardpark-msft:
NOTE: this uses the pre-release commit for azblobs so we can get close to what we need. We'll want to wait to merge this and release until after the GA release for azblobs goes out.

sdk/messaging/azeventhubs/checkpoint_store.go
```diff
@@ -31,9 +33,9 @@ type Ownership struct {
 	FullyQualifiedNamespace string
 	PartitionID             string
 
-	OwnerID          string    // the owner ID of the Processor
-	LastModifiedTime time.Time // used when calculating if ownership has expired
-	ETag             string    // the ETag, used when attempting to claim or update ownership of a partition.
+	OwnerID          string      // the owner ID of the Processor
+	LastModifiedTime time.Time   // used when calculating if ownership has expired
+	ETag             azcore.ETag // the ETag, used when attempting to claim or update ownership of a partition.
```
jhendrixMSFT, line 38: Does this need to be `*azcore.ETag`?  I know we use that in other places as I believe that an empty-string ETag is semantically different from no ETag.

richardpark-msft: Ah, interesting. I'm good with being explicit, I'll switch it over.

jhendrixMSFT: You might want to double-check how this is used in other places, but I'm pretty sure you'll want this change.

richardpark-msft: Swapped over, seemed clean enough. I don't intend to interpret this value, so I think it's best to just completely stay out of that game on the client.

richardpark-msft:
Interesting test failure - it's in an area completely untouched by my change. Going to retry.

sdk/messaging/azeventhubs/go.mod
```diff
@@ -3,10 +3,11 @@ module github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs
 go 1.18
 
 require (
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.1.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.1.0
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.1.2
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/eventhub/armeventhub v1.0.0
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v0.6.2-0.20230201191316-b8690e0d2585
```
jhendrixMSFT, line 10: v1.0.0 is now available.

jhendrixMSFT (approved):

PR #19942, opened 2023-02-03
CAE support for azidentity

chlowell:
This is the azidentity side of #14931. Credentials must do two things to enable the end-to-end feature: specify client capability "CP1" to indicate the client can handle claims challenges, and pass claims through `GetToken` to MSAL. We send CP1 by default so applications are secure by default. Doing so has no effect when the RP doesn't support CAE or the tenant isn't configured for it. However, we provide an environment variable opt out for applications that can't handle claims challenges, for example because they need to upgrade a client library first.

### DIFF
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## 1.3.0-beta.3 (Unreleased)
 
 ### Features Added
+* By default, credentials set client capability "CP1" to enable support for
+  [Continuous Access Evaluation (CAE)](https://docs.microsoft.com/azure/active-directory/develop/app-resilience-continuous-access-evaluation).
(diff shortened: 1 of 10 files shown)


jhendrixMSFT (approved):

JeffreyRichter (commented):
I see a lot of WithXxx functions (Claims, SilentAccount, TenantID). We don't usually have this WithXxx pattern in Go (except maybe for Context). Do customers use this or is this internal implementation for us and it could be changed in a non-breaking way?

Do these WithXxx functions treat the input as immutable and return immutable results or modify the receiver of the WithXxx function? I'm OK with the former (what Context does) and not OK with the latter.

chlowell:
The With* functions are MSAL's options API. None is part of our public API. Are you still curious about their implementation?😆

JeffreyRichter:
No, not anymore. Thanks.

PR #19945, opened 2023-02-03
Adding nil check for GetSASURLOptions.format()

souravgupta-msft:
1. Added nil check in `GetSASURLOptions.format()` method.
2. Fixed Account SAS creation.

### DIFF
--- a/sdk/storage/azblob/service/client_test.go
+++ b/sdk/storage/azblob/service/client_test.go
@@ -693,20 +693,20 @@ func (s *ServiceUnrecordedTestsSuite) TestSASServiceClient() {
 		Delete: true,
 	}
 
-	start := time.Now().Add(-time.Hour)
-	expiry := start.Add(time.Hour)
+	expiry := time.Now().Add(time.Hour)
 
-	opts := service.GetSASURLOptions{StartTime: &start}
-	sasUrl, err := serviceClient.GetSASURL(resources, permissions, expiry, &opts)
+	sasUrl, err := serviceClient.GetSASURL(resources, permissions, expiry, nil)
 	_require.Nil(err)
 
 	svcClient, err := service.NewClientWithNoCredential(sasUrl, nil)
 	_require.Nil(err)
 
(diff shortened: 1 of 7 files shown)


sdk/storage/azblob/service/client_test.go
```diff
@@ -693,20 +693,20 @@ func (s *ServiceUnrecordedTestsSuite) TestSASServiceClient() {
...
 	svcClient, err := service.NewClientWithNoCredential(sasUrl, nil)
 	_require.Nil(err)
 
-	// mismatched container name
-	_, err = svcClient.CreateContainer(context.Background(), containerName+"002", nil)
-	_require.Error(err)
-	testcommon.ValidateBlobErrorCode(_require, err, bloberror.AuthenticationFailed)
+	// create container using SAS
+	_, err = svcClient.CreateContainer(context.Background(), containerName, nil)
+	_require.Nil(err)
```
jhendrixMSFT, line 706: It's preferable to use `_require.NoError(error)`.  You don't have to change it now, but we should clean this up at some point (it's used everywhere).

jhendrixMSFT (approved):

siminsavani-msft (approved):
LGTM!

tasherif-msft:
@souravgupta-msft when we start the release process is there an easy way to target changes prior to this? since we want this change in post GA, should we have some sort of a dev branch instead of directly merging to main?

jhendrixMSFT:
Why do we not want this change as part of GA?  Looks like a potential panic that we should fix.

tasherif-msft:
> Why do we not want this change as part of GA? Looks like a potential panic that we should fix.

my bad I thought this was a PR for something completely different :sweat_smile: (thought we were adding a new api) but yes this should be in GA!

tasherif-msft (approved):

vibhansa-msft (approved):

vibhansa-msft:
These changes shall be part of GA.

PR #19955, opened 2023-02-06
Queues Live Pipeline + APIViews Feedback

tasherif-msft:
Remove `item` suffix from generated and client code.
Add live pipeline.

### DIFF
--- a/sdk/storage/azqueue/service_client_test.go
+++ b/sdk/storage/azqueue/service_client_test.go
@@ -281,7 +281,7 @@ func (s *RecordedTestSuite) TestServiceListQueuesWithMetadata() {
 	for pager.More() {
 		resp, err := pager.NextPage(context.Background())
 		_require.Nil(err)
-		for _, queue := range resp.QueuesList {
+		for _, queue := range resp.Queues {
 			_require.NotNil(queue.Name)
 			if *queue.Name == queueName {
 				_require.NotNil(queue.Metadata)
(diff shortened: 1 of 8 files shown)


tasherif-msft:
@benbp I'm trying to create a live pipeline for the `azqueue` package, is the addition of the arm template (test_resources.json) the only step required? please advise

sdk/storage/azqueue/service_client_test.go
```diff
@@ -281,7 +281,7 @@ func (s *RecordedTestSuite) TestServiceListQueuesWithMetadata() {
 	for pager.More() {
 		resp, err := pager.NextPage(context.Background())
 		_require.Nil(err)
-		for _, queue := range resp.QueueItems {
+		for _, queue := range resp.QueuesList {
```
vibhansa-msft, line 284: QueueItems sound more intutive then QueueList (does not clearly tell what it means)

tasherif-msft: I'm ok with either or here, @jhendrixMSFT what's your input on this?

jhendrixMSFT: Is the suffix even necessary, can it just be `Queue`?

tasherif-msft: @jhendrixMSFT so `Queues`? I'm ok with that as well.

siminsavani-msft (approved):

PR #19958, opened 2023-02-07
More Service Client and Queue Client Tests

tasherif-msft:
Added:
- more service client tests
- more queue client tests (primarily message(s) related APIs)


sdk/storage/azqueue/models.go
```diff
@@ -369,10 +387,13 @@ type UpdateMessageOptions struct {
 }
 
 func (o *UpdateMessageOptions) format() *generated.MessageIDClientUpdateOptions {
+	defaultVT := to.Ptr(int32(0))
```
vibhansa-msft, line 390: setting it to just 'nil' won't work?

vibhansa-msft (approved):

sdk/storage/azqueue/models.go
tasherif-msft: no :(

PR #19969, opened 2023-02-07
Trigger `aztables` when proxy files are updated

scbedd:
So that I can avoid manually triggering this pipeline before merging an update to `eng/common/testproxy/target_version.txt`.

### DIFF
--- a/sdk/data/aztables/ci.yml
+++ b/sdk/data/aztables/ci.yml
@@ -21,6 +21,7 @@ pr:
     include:
     - sdk/data/aztables
     - eng/common/testproxy
+    # adding eng/common/testproxy as aztables is a good exercise of the test-proxy features. if a new proxy version works on this build, we can be confident updating to it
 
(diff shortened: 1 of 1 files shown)


benbp:
Might be good to add a comment alongside the path inclusion so the purpose is clear and people don't copy/paste the ci.yml files along with the proxy pipeline trigger.

benbp (approved):

sdk/data/aztables/ci.yml
```diff
@@ -20,6 +20,7 @@ pr:
   paths:
     include:
     - sdk/data/aztables
+    - eng/common/testproxy
```
scbedd, line 23:
```suggestion
    - eng/common/testproxy
    # adding eng/common/testproxy as aztables is a good exercise of the test-proxy features. if a new proxy version works on this build, we can be confident updating to it
```

PR #19971, opened 2023-02-08
Update package regex to detect arbitrary length directory trees

benbp:
Fix for https://github.com/Azure/azure-sdk-for-go/pull/19879#discussion_r1093893548 to support packages like `security/keyvault/internal` for #19879 

Needs to go in with https://github.com/Azure/azure-sdk/pull/5534

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -37,10 +37,12 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227
+  # The serviceName named capture group is unused but used in azure-sdk, so it's kept here for parity
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(.*?(?<serviceName>[^/]+)/)?(?<modName>[^/]+$)))")
   {
     $modPath = $matches["modPath"]
-    $modName = $matches["modName"] # We may need to start readong this from the go.mod file if the path and mod config start to differ
+    $modName = $matches["modName"] # We may need to start reading this from the go.mod file if the path and mod config start to differ
     $serviceDir = $matches["serviceDir"]
     $sdkType = "client"
     if ($modName.StartsWith("arm") -or $modPath.Contains("resourcemanager")) { $sdkType = "mgmt" }


eng/scripts/Language-Settings.ps1
```diff
@@ -37,10 +37,10 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?(.+/)?(?<modName>[^/]+$)))")
```
chlowell, line 40: We should have a comment about the other regex this one must stay in sync with

eng/scripts/Language-Settings.ps1
```diff
@@ -37,10 +37,11 @@ function Get-GoModuleVersionInfo($modPath)
 function Get-GoModuleProperties($goModPath)
 {
   $goModPath = $goModPath -replace "\\", "/"
-  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?([^/]+/)?(?<modName>[^/]+$)))")
+  # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227
+  if ($goModPath -match "(?<modPath>sdk/(?<serviceDir>(resourcemanager/)?(.+/)?(?<modName>[^/]+$)))")
```
weshaggard, line 41: The last time I updated this I didn't make it this generic because it started picking up other go mods for test projects and such so making it this open ended may start discovering more then you want and you might start treating test projects as client libraries which will cause them to start failing.

chlowell: That's a good point. Could we just exclude `sdk/samples`? It's the only directory which contains no releasing module.

benbp: @weshaggard @chlowell `sdk/samples` and a few other projects like perf match the regex as well with this change, but they get excluded from the package properties read out because they don't contain the version information we expect in `constants.go` and/or `version.go` (see [here](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L8-L25)).

I think it will be more flexible to use a metadata-style convention over a directory style convention anyways, rather than continuously needing to tune the regex. Right now there is no diff on what package metadata is found before/after this regex change due to the version search criteria.

chlowell (approved):

eng/scripts/Language-Settings.ps1
weshaggard: While we might not get more packages we will get a lot more warnings (https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/Language-Settings.ps1#L33). I'm not sure how we clean-up those warnings while still providing diagnostics for folks that are trying to add a new module but doing something wrong.

weshaggard (approved):

PR #19992, opened 2023-02-10
[azservicebus] Validate that credits are within limits (not over or under)

### LINKED ISSUE #19743: [azservicebus] Asking for a negative count of messages results doesn't return an error
This should return a validation error because the number of requested messages is negative:

```go
// (blocks today)
receiver.ReceiveMessages(context.TODO(), -1, nil)
```

This is a definite error case, should return a validation error. It is possible to request negative credits in AMQP, but that's not something we're trying to enable here and it's possible it doesn't even do that at the moment with our stack.

### LINKED ISSUE #19965: azservice bus can over-issue receiving link credit
Over-issuing of credit can cause a receiver to hang in some circumstances.  This has been fixed in later versions of `go-amqp` to return an error, so SB will need to update to handle this (unsure about EH).

richardpark-msft:
Fixing issues where we could over-request credit (#19965) or allow for negative/zero credits (#19743), both of which could cause issues with go-amqp..

Fixes #19965
Fixes #19743

### DIFF
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -81,7 +81,7 @@ type ReceiverOptions struct {
 	SubQueue SubQueue
 }
 
-const defaultLinkRxBuffer = 10000
+const defaultLinkRxBuffer uint32 = 10000
 
 func applyReceiverOptions(receiver *Receiver, entity *entity, options *ReceiverOptions) error {
 	if options == nil {
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -81,7 +81,10 @@ type ReceiverOptions struct {
 	SubQueue SubQueue
 }
 
-const defaultLinkRxBuffer uint32 = 10000
+// defaultLinkRxBuffer is the maximum number of transfer frames we can handle
+// on the Receiver. This matches the current default window size that go-amqp
+// uses for sessions.
+const defaultLinkRxBuffer uint32 = 5000
 
 func applyReceiverOptions(receiver *Receiver, entity *entity, options *ReceiverOptions) error {
 	if options == nil {
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -45,22 +45,18 @@ const (
 
 // Receiver receives messages using pull based functions (ReceiveMessages).
 type Receiver struct {
-	receiveMode ReceiveMode
-	entityPath  string
-
-	settler        settler
-	retryOptions   RetryOptions
-	cleanupOnClose func()
-
-	lastPeekedSequenceNumber int64
 	amqpLinks                internal.AMQPLinks
-
-	mu        sync.Mutex
-	receiving bool
-
+	cancelReleaser           *atomic.Value
+	cleanupOnClose           func()
 	defaultTimeAfterFirstMsg time.Duration
-
-	cancelReleaser *atomic.Value
+	entityPath               string
+	lastPeekedSequenceNumber int64
+	maxAllowedCredits        uint32
+	mu                       sync.Mutex
+	receiveMode              ReceiveMode
+	receiving                bool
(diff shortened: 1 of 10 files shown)


sdk/messaging/azservicebus/receiver.go
```diff
@@ -85,7 +81,7 @@ type ReceiverOptions struct {
 	SubQueue SubQueue
 }
 
-const defaultLinkRxBuffer = 2048
+const defaultLinkRxBuffer = 10000
```
jhendrixMSFT, line 84: Make sure this isn't larger than the incoming window.  While it doesn't really matter today, it might in the future.

jhendrixMSFT: It also seems quite large.  Do you really think we need to change it?  How do we measure the impact at any value?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -364,6 +361,14 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
 	cancelReleaser := r.cancelReleaser.Swap(emptyCancelFn).(func() string)
 	_ = cancelReleaser()
 
+	if maxMessages <= 0 {
+		return nil, internal.NewErrNonRetriable("maxMessages should be greater than 0")
+	}
+
+	if maxMessages > int(r.maxAllowedCredits) {
```
jhendrixMSFT, line 368: Does this mean that if the link has N number of credits and I call `Receive()` with N+M, `Receive()` will return an error?

richardpark-msft: No, if you look a little below it takes into account your existing credits when calculating how many to issue.

In other words, if you ask for 10000 we'll just make sure that 10000 credits are issued, overall.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: > Make sure this isn't larger than the incoming window. While it doesn't really matter today, it might in the future.

How can I determine the incoming window value? Or maybe the other way - shouldn't the incoming window value be calculated based on my max link credit?

> It also seems quite large. Do you really think we need to change it? How do we measure the impact at any value?

I'd much rather not set it at all and have it be dynamic, based on whatever the outstanding issued credit is. In the meantime I have to pick some ceiling. What would you suggest?

sdk/messaging/azservicebus/receiver.go
jhendrixMSFT: That part I get.  I'm looking at the caller above us `Receiver.ReceiveMessages()`.  It looks to me that if I pass a value for `maxMessages` that's greater than the max link credit it returns an error.

sdk/messaging/azservicebus/receiver.go
jhendrixMSFT: From this conversation, I think it's misleading to call the option `ReceiverOptions.MaxCredit`.  It's not the maximum amount of credit on the link (that's essentially 2^32).  It's the maximum amount of prefetched messages.  I think the field needs to be renamed.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Talked more with @jhendrixMSFT about this and it's more interesting than I think either of us considered. We'll let this stand (for now) and then hit it again when we change how windowing+max credits works in go-amqp.  This'll protect the current implementation from possibly hanging when the underlying Receiver's channel is full.

We'd like to make a change to go-amqp to eliminate this restriction. I've filed an issue for the discussion here: https://github.com/Azure/go-amqp/issues/240

sdk/messaging/azservicebus/receiver.go
richardpark-msft: (from other discussion, I think we might end up eliminating this field altogether)

jhendrixMSFT (approved):

jhendrixMSFT:
Once we get the undelrying issue in `go-amqp` fixed we should be able to back this out.

PR #20003, opened 2023-02-13
AzQueue Public Preview APIVIew Feedback

tasherif-msft:


souravgupta-msft (approved):

jhendrixMSFT (approved):

vibhansa-msft (approved):

siminsavani-msft (approved):

sdk/storage/azqueue/models.go
```diff
@@ -80,6 +80,9 @@ type Queue = generated.Queue
 // AccessPolicy - An Access policy
 type AccessPolicy = generated.AccessPolicy
 
+// GeoReplicationStatus - The status of the secondary location
+type GeoReplicationStatus = generated.GeoReplicationStatus
```
jhendrixMSFT, line 84: The pattern we've been following is to place all of the `const` types and their respective values in a `constants.go` file.

tasherif-msft: that makes sense, will make that change

JeffreyRichter (approved):

sdk/storage/azqueue/constants.go
```diff
@@ -0,0 +1,12 @@
+package azqueue
```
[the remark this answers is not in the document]
tasherif-msft: Yea, I scrubbed it once more, we should be all good

Issue #19984, opened 2023-02-09
Consolidate test utility


tadelesh:
Try to remove duplication of `sdk/resourcemanager/internal` compared with `sdk/internal` and modify all existing usage in mgmt. live test.

PR #20009, opened 2023-02-14
Add GenerateAlphaNumericID for test-proxy to recording package

tadelesh:
1. Add GenerateAlphaNumericID function.
2. Store random seed in each recording file.
3. This change is to move the similar function under `sdk/resourcemanger/internal` as it should be part of our recording system. 

Part of #19984

### DIFF
--- a/sdk/internal/recording/recording.go
+++ b/sdk/internal/recording/recording.go
@@ -237,38 +237,8 @@ func (r *Recording) UUID() uuid.UUID {
 // GenerateAlphaNumericID will generate a recorded random alpha numeric id
 // if the recording has a randomSeed already set, the value will be generated from that seed, else a new random seed will be used
 func (r *Recording) GenerateAlphaNumericID(prefix string, length int, lowercaseOnly bool) (string, error) {
-
-	if length <= len(prefix) {
-		return "", errors.New("length must be greater than prefix")
-	}
-
 	r.initRandomSource()
-
-	sb := strings.Builder{}
-	sb.Grow(length)
-	sb.WriteString(prefix)
-	i := length - len(prefix) - 1
-	// A src.Int63() generates 63 random bits, enough for letterIdxMax characters!
-	for cache, remain := r.src.Int63(), letterIdxMax; i >= 0; {
-		if remain == 0 {
-			cache, remain = r.src.Int63(), letterIdxMax
-		}
-		if lowercaseOnly {
-			if idx := int(cache & letterIdxMask); idx < len(alphanumericLowercaseBytes) {
-				sb.WriteByte(alphanumericLowercaseBytes[idx])
-				i--
-			}
-		} else {
-			if idx := int(cache & letterIdxMask); idx < len(alphanumericBytes) {
-				sb.WriteByte(alphanumericBytes[idx])
-				i--
-			}
-		}
-		cache >>= letterIdxBits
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
@tadelesh can you please try running the recorded tests for `azblob` with this change to ensure that nothing is broken?

sdk/internal/recording/recording.go
```diff
@@ -814,7 +801,76 @@ func Stop(t *testing.T, options *RecordingOptions) error {
...
+		testStruct.recordingSeed = seed
+		testStruct.recordingRandomSource = source
+		testSuite.Store(t.Name(), testStruct)
+	}
+
+	return source
+}
+
+// GenerateAlphaNumericID will generate a recorded random alpha numeric id.
+// When playback mode, a new random seed will be used, otherwise, the random seed is stable and will be stored in recording file.
```
jhendrixMSFT, line 836: I think this comment is wrong and should be the same as `Recording.GenerateAlphaNumericID()` yes?

sdk/internal/recording/recording.go
```diff
@@ -496,22 +466,27 @@ func init() {
...
 
 type recordedTest struct {
-	recordingId string
-	liveOnly    bool
-	variables   map[string]interface{}
+	recordingId           string
+	liveOnly              bool
+	variables             map[string]interface{}
+	recordingSeed         int64
+	recordingRandomSource rand.Source
```
jhendrixMSFT, line 489: Can we shorten the name of this field a bit?  Maybe `recordingRndSrc`?

sdk/internal/recording/recording.go
```diff
@@ -778,6 +758,12 @@ func Stop(t *testing.T, options *RecordingOptions) error {
 			// test should only be run live, don't want to generate recording
 			return nil
 		}
+		if testStruct.recordingSeed != 0 {
```
jhendrixMSFT, line 761: How did this used to work, or was it broken?

sdk/internal/recording/recording.go
```diff
@@ -496,22 +466,27 @@ func init() {
...
-	IDHeader          = "x-recording-id"
-	ModeHeader        = "x-recording-mode"
-	UpstreamURIHeader = "x-recording-upstream-base-uri"
+	RecordingMode                   = "record"
+	PlaybackMode                    = "playback"
+	LiveMode                        = "live"
+	IDHeader                        = "x-recording-id"
+	ModeHeader                      = "x-recording-mode"
+	UpstreamURIHeader               = "x-recording-upstream-base-uri"
+	recordingRandomSeedVariableName = "recordingRandomSeed"
```
jhendrixMSFT, line 481: Can we shorten this name?

sdk/internal/recording/recording.go
tadelesh: For the test-proxy testing framework, we don't have utility to support generating random string. So, we need to store the random seed in the option's variable and stop with this option to store the random seed. For now, we add such feature, so we need to do the seed store for user.

tadelesh:
> @tadelesh can you please try running the recorded tests for `azblob` with this change to ensure that nothing is broken?

I've tested `azblob` and all tests passed.

jhendrixMSFT (approved):

PR #20016, opened 2023-02-15
PutBlobFromUrl API

siminsavani-msft:
Adding the PutBlobFromUrl API


sdk/storage/azblob/blockblob/client.go
```diff
@@ -165,6 +165,20 @@ func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *U
 	return resp, err
 }
 
+// PutBlobFromURL - The Put Blob from URL operation creates a new Block Blob where the contents of the blob are read from
+// a given URL. Partial updates are not supported with Put Blob from URL; the content of an existing blob is overwritten
+// with the content of the new blob. To perform partial updates to a block blob’s contents using a source URL, use the Put
+// Block from URL API in conjunction with Put Block List.
+// For more information, see https://learn.microsoft.com/rest/api/storageservices/put-blob-from-url
+func (bb *Client) PutBlobFromURL(ctx context.Context, contentLength int64, copySource string, options *PutBlobFromURLOptions) (PutBlobFromURLResponse, error) {
```
souravgupta-msft, line 173: The [documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url#request-headers) states the following for the `Content-Length` header.
> <html>
<body>


Content-Length | Required. Specifies the number of bytes being transmitted in the request body. The value of this header must be set to 0. When the length isn't 0, the operation fails with the status code 400 (Bad Request).
-- | --


</body>
</html>

If this is the case, we should not have this as a required parameter and always pass its value as 0 to the generated code.

siminsavani-msft: Thank you for pointing this out! Fixed now!

sdk/storage/azblob/blockblob/client.go
```diff
@@ -165,6 +165,21 @@ func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *U
...
 
+// PutBlobFromURL - The Put Blob from URL operation creates a new Block Blob where the contents of the blob are read from
+// a given URL. Partial updates are not supported with Put Blob from URL; the content of an existing blob is overwritten
+// with the content of the new blob. To perform partial updates to a block blob’s contents using a source URL, use the Put
+// Block from URL API in conjunction with Put Block List.
+// For more information, see https://learn.microsoft.com/rest/api/storageservices/put-blob-from-url
+func (bb *Client) PutBlobFromURL(ctx context.Context, copySource string, options *PutBlobFromURLOptions) (PutBlobFromURLResponse, error) {
+	opts, httpHeaders, leaseAccessConditions, cpkInfo, cpkSourceInfo, modifiedAccessConditions, sourceModifiedConditions := options.format()
+
+	contentLength := int64(0)
```
vibhansa-msft, line 176: We can directly pass on 0 instead of creating a variable here.

sdk/storage/azblob/blockblob/client.go
```diff
@@ -165,6 +165,21 @@ func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *U
 	return resp, err
 }
 
+// PutBlobFromURL - The Put Blob from URL operation creates a new Block Blob where the contents of the blob are read from
+// a given URL. Partial updates are not supported with Put Blob from URL; the content of an existing blob is overwritten
+// with the content of the new blob. To perform partial updates to a block blob’s contents using a source URL, use the Put
+// Block from URL API in conjunction with Put Block List.
+// For more information, see https://learn.microsoft.com/rest/api/storageservices/put-blob-from-url
+func (bb *Client) PutBlobFromURL(ctx context.Context, copySource string, options *PutBlobFromURLOptions) (PutBlobFromURLResponse, error) {
+	opts, httpHeaders, leaseAccessConditions, cpkInfo, cpkSourceInfo, modifiedAccessConditions, sourceModifiedConditions := options.format()
```
vibhansa-msft, line 174: Is this our standard practice in azblob to return so many values from format method ()? Returning a structure will look much simpler in terms of code.

siminsavani-msft: Yes, this is the current practice seen throughout azblob. Changing the current practice would be inconsistent.

sdk/storage/azblob/blockblob/models.go
```diff
@@ -70,6 +70,59 @@ func (o *UploadOptions) format() (*generated.BlockBlobClientUploadOptions, *gene
...
+
+	// Optional. Indicates the tier to be set on the blob.
+	Tier *blob.AccessTier
+
+	// Specify the transactional md5 for the body, to be validated by the service.
+	TransactionalContentMD5 []byte
+
+	HTTPHeaders                    *blob.HTTPHeaders
+	AccessConditions               *blob.AccessConditions
+	CPKInfo                        *blob.CPKInfo
```
tasherif-msft, line 98: we should add tests for these

sdk/storage/azblob/blockblob/client.go
```diff
@@ -165,6 +165,20 @@ func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *U
 	return resp, err
 }
 
+// PutBlobFromURL - The Put Blob from URL operation creates a new Block Blob where the contents of the blob are read from
+// a given URL. Partial updates are not supported with Put Blob from URL; the content of an existing blob is overwritten
+// with the content of the new blob. To perform partial updates to a block blob’s contents using a source URL, use the Put
+// Block from URL API in conjunction with Put Block List.
+// For more information, see https://learn.microsoft.com/rest/api/storageservices/put-blob-from-url
+func (bb *Client) PutBlobFromURL(ctx context.Context, copySource string, options *PutBlobFromURLOptions) (PutBlobFromURLResponse, error) {
```
tasherif-msft, line 173: `CopyBlobFromURL` is a better name here

sdk/storage/azblob/blockblob/models.go
```diff
@@ -70,6 +70,59 @@ func (o *UploadOptions) format() (*generated.BlockBlobClientUploadOptions, *gene
...
+	SourceContentMD5 []byte
+
+	// Optional. Indicates the tier to be set on the blob.
+	Tier *blob.AccessTier
+
+	// Specify the transactional md5 for the body, to be validated by the service.
+	TransactionalContentMD5 []byte
+
+	HTTPHeaders                    *blob.HTTPHeaders
+	AccessConditions               *blob.AccessConditions
```
tasherif-msft, line 97: please add more tests to test more of these headers

sdk/storage/azblob/blockblob/responses.go
```diff
@@ -16,6 +16,9 @@ import (
 // UploadResponse contains the response from method Client.Upload.
 type UploadResponse = generated.BlockBlobClientUploadResponse
 
+// PutBlobFromURLResponse contains the response from the method Client.PutBlobFromURL
+type PutBlobFromURLResponse = generated.BlockBlobClientPutBlobFromURLResponse
```
tasherif-msft, line 20: rename like above to `Copy`

sdk/storage/azblob/blockblob/client.go
souravgupta-msft: `CopyBlobFromURL` sounds like this, https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url.
[Put Blob](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob) operation is named [`Upload`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blockblob/client.go#L156). Similarly we can name [Put Blob from URL](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url) as `UploadBlobFromURL`.

siminsavani-msft: Done!

sdk/storage/azblob/blockblob/models.go
```diff
@@ -70,6 +70,59 @@ func (o *UploadOptions) format() (*generated.BlockBlobClientUploadOptions, *gene
 
 // ---------------------------------------------------------------------------------------------------------------------
 
+// UploadBlobFromURLOptions contains the optional parameters for the Client.UploadBlobFromURL method.
+type UploadBlobFromURLOptions struct {
+	// Optional. Used to set blob tags in various blob operations.
+	Tags map[string]string
+
+	// Only Bearer type is supported. Credentials should be a valid OAuth access token to copy source.
+	CopySourceAuthorization *string
```
jhendrixMSFT, line 79: How is this used?

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -586,6 +586,119 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+	_require.NoError(err)
+
+	containerName := testcommon.GenerateContainerName(testName)
+	containerClient := testcommon.CreateNewContainer(context.Background(), _require, containerName, svcClient)
+	defer testcommon.DeleteContainer(context.Background(), _require, containerClient)
+
+	srcBlob := testcommon.GenerateBlobName(testName)
+	srcBBClient := testcommon.CreateNewBlockBlob(context.Background(), _require, srcBlob, containerClient)
+
+	dest := testcommon.GenerateBlobName(testName)
```
souravgupta-msft, line 602: `srcBlob` and `dest` are same here. These should be different, right?

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -586,6 +586,119 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
```
souravgupta-msft, line 653: `srcBlob` and `dest` should be different

sdk/storage/azblob/blockblob/models.go
siminsavani-msft: It's an optional parameter that uses bearer token for authorization.

sdk/storage/azblob/blockblob/client.go
```diff
@@ -165,6 +165,20 @@ func (bb *Client) Upload(ctx context.Context, body io.ReadSeekCloser, options *U
...
 
+// UploadBlobFromURL - The Put Blob from URL operation creates a new Block Blob where the contents of the blob are read from
+// a given URL. Partial updates are not supported with Put Blob from URL; the content of an existing blob is overwritten
+// with the content of the new blob. To perform partial updates to a block blob’s contents using a source URL, use the Put
+// Block from URL API in conjunction with Put Block List.
+// For more information, see https://learn.microsoft.com/rest/api/storageservices/put-blob-from-url
+func (bb *Client) UploadBlobFromURL(ctx context.Context, copySource string, options *UploadBlobFromURLOptions) (UploadBlobFromURLResponse, error) {
+	opts, httpHeaders, leaseAccessConditions, cpkInfo, cpkSourceInfo, modifiedAccessConditions, sourceModifiedConditions := options.format()
+
+	resp, err := bb.generated().PutBlobFromURL(ctx, int64(0), copySource, opts, httpHeaders, leaseAccessConditions,
```
vibhansa-msft, line 176: we might need trasacational md5/crc or setting up md5 sum feature here as well.

sdk/storage/azblob/blockblob/models.go
jhendrixMSFT: Do we expect customers to populate this with the actual bearer token string, or should this be an `azcore.TokenCredential`?  What do other languages do?

gapra-msft: @jhendrixMSFT I looked into the other languages, so for Java and .NET, they have a special HttpAuthorization type provided by Azure Core
https://learn.microsoft.com/en-us/dotnet/api/azure.httpauthorization.-ctor?view=azure-dotnet#azure-httpauthorization-ctor(system-string-system-string)
https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java/staging/apidocs/com/azure/core/http/HttpAuthorization.html

For Python, looks like it's just a string https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L583

jhendrixMSFT: I see.  Is it the case that we're authenticating with an endpoint outside of Azure?

seanmcc-msft: https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url?tabs=azure-ad#request-headers

> Only scheme bearer is supported for Azure Active Directory.

sdk/storage/azblob/blockblob/client.go
seanmcc-msft: In this case, we will need to follow the same pattern as https://github.com/Azure/azure-sdk-for-go/pull/20356.  Keep in mind the public interface needs to be future proofed, as we will be adding the ability for the SDK to calculate checksums in the future

seanmcc-msft: We just need to support `x-ms-source-content-md5`, and it will not be possible for the SDK to pre-calculate this checksum.

https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url?tabs=azure-ad#request-headers

sdk/storage/azblob/blockblob/models.go
seanmcc-msft: @jhendrixMSFT, `CopySourceAuthorization` is a `*string` everywhere else in AzBlob.  For example - https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blockblob/models.go#L100.  Put Block from URL also only supports authenticating with AAD - https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-from-url?tabs=azure-ad#request-headers

sdk/storage/azblob/blockblob/client.go
seanmcc-msft: Hold off on this for now, we need to figure out what we are doing with checksuming in AzCopy.  Consider this comment non-blocking.

sdk/storage/azblob/blockblob/models.go
siminsavani-msft: Added tests covering all headers!

siminsavani-msft:
Hi @scbedd ! 

I have one recorded test in this PR and recorded it according (1. turned on test-proxy, recorded the test 2. ran the test-proxy command to get a new tag with updated recorded test). The CI is failing because other (unrelated to my PR) recorded tests that were merged are showing this error: 

`Recording ID was not returned by the response. Response body: {"Message":"Recording file path /mnt/vss/_work/1/s/.assets/bN3eZCureo/go/sdk/storage/azblob/testdata/recordings/Test/TestBlockGetAccountInfo.json does not exist.","Status":"NotFound"}`

Do you know why is this happening and how I can fix this without having to recorded the recorded tests that are not related to PR?

siminsavani-msft:
Closing this PR and creating a new fresh PR here: https://github.com/Azure/azure-sdk-for-go/pull/20558

Issue #20017, opened 2023-02-15
Handle more cases in changelog generation tools


tadelesh:
1. Refine changelog content when an operation change to pageable or non-pageable.
https://github.com/Azure/azure-sdk-for-go/pull/19912/commits/5f3cc7f5c1db6dc2db0bf5a8068a9b6497ac231e#diff-ed0d79a6efd2c2dca8679a33a2ac0a9e4ff293c192d6ea8e0fa050cd9c8f33edR6-R16
2. Move any type change to any to `Features Added` part. This should support array and map.
https://github.com/Azure/azure-sdk-for-go/pull/19917/commits/6286a5d31ff81a3ff876c382e9a283ca134bf19e#diff-9cdac115969df075aeeab25b834ad6467e29cbb39585bb9da84fb4f672995fefR6
Advanced example: []string -> []any, map[string]string -> map[string]any

PR #20021, opened 2023-02-15
changelog generator tool add `PageableFilter` and `type to any changelog print`

Alancere:
https://github.com/Azure/azure-sdk-for-go/issues/20017


eng/tools/generator/autorest/model/changelog.go
```diff
@@ -265,8 +269,10 @@ func getSignatureChangeItems(b *report.BreakingChanges) []string {
 		for _, k := range sortChangeItem(b.Structs) {
 			v := b.Structs[k]
 			for f, d := range v.Fields {
-				line := fmt.Sprintf("Type of `%s.%s` has been changed from `%s` to `%s`", k, f, d.From, d.To)
-				items = append(items, line)
+				if !strings.Contains(d.To, "any") {
```
tadelesh, line 272: Could you please remove duplication of these code?

eng/tools/generator/autorest/model/changelog.go
```diff
@@ -419,3 +440,24 @@ func removePattern(funcName string, returnValue string) string {
...
+	if b.IsEmpty() {
+		return nil
+	}
+
+	var items []string
+	if len(b.Structs) > 0 {
+		for _, k := range sortChangeItem(b.Structs) {
+			v := b.Structs[k]
+			for f, d := range v.Fields {
+				if strings.Contains(d.To, "any") {
```
tadelesh, line 454: I'm a little bit struggling about this condition. For most situation, is could be right, but I wonder if there is any corner case that will break it, e.g., `map[A]string` -> `map[B]any`. To be safe, perhaps we only support exact equal at first.

eng/tools/generator/cmd/v2/common/changelogProcessor_test.go
```diff
@@ -79,6 +79,28 @@ func TestLROFilter(t *testing.T) {
 	assert.Equal(t, excepted, changelog.ToCompactMarkdown())
 }
 
+func TestPageableFilter(t *testing.T) {
```
tadelesh, line 82: You may add some tests for "to any" changelog.

eng/tools/generator/autorest/model/changelog.go
Alancere: Duplicate code means duplication with function typeToAny?

tadelesh: > Duplicate code means duplication with function typeToAny?

Yes.

tadelesh (approved):

PR #20028, opened 2023-02-16
Show SDK log for CI test

tadelesh:
To make the debug of CI occasionally failure easier.


benbp (approved):

benbp:
I think this should be ok, but there's a chance the logging volume adds processing delay to our pipeline log queries that we use on the engsys side. I'll follow up if it becomes an issue, but at the moment it doesn't look too bad.

Issue #20007, opened 2023-02-14
Fix occasional CI failure for azcontainerregistry


tadelesh:
The failure could happen in every Update test case: `UpdateTag`, `UpdateManifest`, and the error always is `CONTENT_LEN_INVALID`. See [log](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2184663&view=logs&j=9289e972-0a8d-5259-5124-44db723f911e&t=7e9a595d-1588-5d0f-d134-1dae779a2ee4).

PR #20034, opened 2023-02-17
Fix test failure because of "CONTENT_LEN_INVALID"

tadelesh:
Two problems:
1. When cloning challenge request in the auth policy, the previous code will erase the original request's "Content-Length"
2. When retry, the previous code will do another redundant challenge request

Previous failure example: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2194977&view=logs&j=a20abffc-7e85-5d60-a753-b6154d45fe99&t=b38a3d44-54a5-5a70-e115-78df002fe247.
Live test result after fix: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2198625&view=results

Fix: https://github.com/Azure/azure-sdk-for-go/issues/20007

### DIFF
--- a/sdk/containers/azcontainerregistry/authentication_policy.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy.go
@@ -66,9 +66,6 @@ func (p *authenticationPolicy) Do(req *policy.Request) (*http.Response, error) {
 	if req.Raw().Header.Get(headerAuthorization) != "" {
 		// retry request could do the request with existed token directly
 		resp, err = req.Next()
-		if err != nil {
-			return nil, err
-		}
 	} else if p.accessTokenCache != "" {
 		// if there is a previous access token, then we try to use this token to do the request
 		req.Raw().Header.Set(
@@ -76,9 +73,6 @@ func (p *authenticationPolicy) Do(req *policy.Request) (*http.Response, error) {
 			fmt.Sprintf("%s%s", bearerHeader, p.accessTokenCache),
 		)
 		resp, err = req.Next()
-		if err != nil {
-			return nil, err
-		}
 	} else {
 		// do challenge process for the initial request
 		challengeReq, err := p.getChallengeRequest(*req)
@@ -86,9 +80,9 @@ func (p *authenticationPolicy) Do(req *policy.Request) (*http.Response, error) {
 			return nil, err
 		}
 		resp, err = challengeReq.Next()
-		if err != nil {
-			return nil, err
-		}
+	}
+	if err != nil {
+		return nil, err
 	}
 
 	// if 401 response, then try to get access token
--- a/sdk/containers/azcontainerregistry/authentication_policy.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy.go
@@ -7,13 +7,10 @@
 package azcontainerregistry
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"net/http"
 	"strings"
@@ -52,10 +49,7 @@ func newAuthenticationPolicy(cred azcore.TokenCredential, scopes []string, authC
 
 func (p *authenticationPolicy) Do(req *policy.Request) (*http.Response, error) {
 	// send a copy of the original request without body content
-	challengeReq, err := p.getChallengeRequest(*req)
-	if err != nil {
-		return nil, err
-	}
+	challengeReq := p.getChallengeRequest(*req)
 	resp, err := challengeReq.Next()
 	if err != nil {
 		return nil, err
@@ -148,27 +142,11 @@ func (p *authenticationPolicy) findServiceAndScope(resp *http.Response) error {
 	return nil
 }
 
-func (p authenticationPolicy) getChallengeRequest(orig policy.Request) (*policy.Request, error) {
-	req, err := runtime.NewRequest(orig.Raw().Context(), orig.Raw().Method, orig.Raw().URL.String())
-	if err != nil {
-		return nil, err
-	}
-
-	req.Raw().Header = orig.Raw().Header
-	req.Raw().Header.Set("Content-Length", "0")
-	req.Raw().ContentLength = 0
-
-	copied := orig.Clone(orig.Raw().Context())
-	copied.Raw().Body = req.Body()
-	copied.Raw().ContentLength = 0
-	copied.Raw().Header.Set("Content-Length", "0")
-	err = copied.SetBody(streaming.NopCloser(bytes.NewReader([]byte{})), "application/json")
-	if err != nil {
-		return nil, err
-	}
+func (p authenticationPolicy) getChallengeRequest(oriReq policy.Request) *policy.Request {
+	copied := oriReq.Clone(oriReq.Raw().Context())
+	copied.SetBody(nil, "")
 	copied.Raw().Header.Del("Content-Type")
-
-	return copied, err
+	return copied
 }
 
 type acquiringResourceState struct {
--- a/sdk/containers/azcontainerregistry/authentication_policy.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy.go
@@ -29,67 +29,100 @@ const (
 type authenticationPolicyOptions struct {
 }
 
+// authenticationPolicy is a policy to do the challenge-based authentication for container registry service. The authorization flow is as follows:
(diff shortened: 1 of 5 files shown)


tadelesh:
Explanation for the CI failure:
When the first request try failed with a retriable return code (e.g., 504 Gateway Time-out), then it will start a retry request. Then the auth policy will use the original req to do another challenge request, but this request has an additional `Authorization` header that caused the service 400 error.

sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -51,38 +48,42 @@ func newAuthenticationPolicy(cred azcore.TokenCredential, scopes []string, authC
...
 			return nil, err
 		}
-
-		accessToken, err := p.getAccessToken(req)
+		resp, err := challengeReq.Next()
 		if err != nil {
 			return nil, err
 		}
+		// do challenge process
+		if resp.StatusCode == 401 {
```
jhendrixMSFT, line 63: Nit: use `http.StatusUnauthorized`

jhendrixMSFT: Also, I think you could clean up some of this nesting.
```go
if err != nil {
    return nil, err
} else if resp.StatusCode != http.StatusUnauthorized {
    // the request failed for some other reason, don't try any further
    return resp, nil
}
```

jhendrixMSFT (approved):
Just some minor cleanup.

chlowell:
I'm not familiar with the protocol details here, so please forgive me if I'm missing something, but it looks like each client call entails 3 requests. Do we really need to get a new challenge and a new ACR token before every request?

Also, something for a future PR: the bearer auth policy in azcore now has an extension API for challenges that could simplify this. Here are the [docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.3.1/policy#AuthorizationHandler) and an implementation for [Key Vault](https://github.com/Azure/azure-sdk-for-go/blob/82977624dd73caeb69c989a2a814bb0581cd082b/sdk/security/keyvault/internal/challenge_policy.go#L52).

tadelesh:
> I'm not familiar with the protocol details here, so please forgive me if I'm missing something, but it looks like each client call entails 3 requests. Do we really need to get a new challenge and a new ACR token before every request?
> 
> Also, something for a future PR: the bearer auth policy in azcore now has an extension API for challenges that could simplify this. Here are the [docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.3.1/policy#AuthorizationHandler) and an implementation for [Key Vault](https://github.com/Azure/azure-sdk-for-go/blob/82977624dd73caeb69c989a2a814bb0581cd082b/sdk/security/keyvault/internal/challenge_policy.go#L52).

I added a simple access token cache to the policy. But I think cache will be hit only when continuously calling same API. The reason is the access token is generated with scope returned by challenge request and the scope will change for different API and API parameter. e.g., the scope for get property of repository ubuntu is `repository:ubuntu:metadata_read` and the scope for change property of that repository is `repository:ubuntu:metadata_write`. However, the cache will be useful when doing upload by chunk, because there will be several same API calls. And based on that, I do not add any synchronized logic to this cache, just a simple string, lock will be a waste.

I also went through the challenge handler logic recently added to the bearer token policy. I'm afraid it is hard to build ACR auth policy based on it. The main reason is the token fetch and cache logic are totally different, write a clean new logic is easier.

@chlowell @jhendrixMSFT Could you help to review the change ASAP? I want to catch up with the March release for this fix because the issue may happen for every request's retry.

sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -32,93 +29,129 @@ const (
...
-	}
-	resp, err := challengeReq.Next()
-	if err != nil {
-		return nil, err
+	var resp *http.Response
+	var err error
+	if req.Raw().Header.Get(headerAuthorization) != "" {
+		// retry request could do the request with existed token directly
+		resp, err = req.Next()
+		if err != nil {
```
jhendrixMSFT, line 69: Nit: you could move this and the other `err` checks outside the main `if` clause to consolidate them.

sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -32,93 +29,129 @@ const (
...
-	challengeReq, err := p.getChallengeRequest(*req)
-	if err != nil {
-		return nil, err
-	}
-	resp, err := challengeReq.Next()
-	if err != nil {
-		return nil, err
+	var resp *http.Response
+	var err error
+	if req.Raw().Header.Get(headerAuthorization) != "" {
```
jhendrixMSFT, line 66: When is this branch hit?

tadelesh: When retrying request, this branch will be hit.

chlowell:
> The reason is the access token is generated with scope returned by challenge request and the scope will change for different API and API parameter. e.g., the scope for get property of repository ubuntu is `repository:ubuntu:metadata_read` and the scope for change property of that repository is `repository:ubuntu:metadata_write`.
> 

Does this mean a token acquired for `repository:ubuntu:metadata_read` isn't valid for `repository:ubuntu:metadata_write`? If so, isn't it unsafe to simply cache the last token received, because there's no way to know whether the token will be valid for the next call?

tadelesh:
> > The reason is the access token is generated with scope returned by challenge request and the scope will change for different API and API parameter. e.g., the scope for get property of repository ubuntu is `repository:ubuntu:metadata_read` and the scope for change property of that repository is `repository:ubuntu:metadata_write`.
> 
> Does this mean a token acquired for `repository:ubuntu:metadata_read` isn't valid for `repository:ubuntu:metadata_write`? If so, isn't it unsafe to simply cache the last token received, because there's no way to know whether the token will be valid for the next call?

Yes. You are right. Do a request with an invalid token is similar with do a clear challenge request. The response will all include the scope needed for the request.

sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -32,93 +29,120 @@ const (
...
+		resp, err = challengeReq.Next()
 	}
-	resp, err := challengeReq.Next()
 	if err != nil {
 		return nil, err
 	}
 
-	// do challenge process
+	// if 401 response, then try to get access token
 	if resp.StatusCode == 401 {
```
jhendrixMSFT, line 90: Nit replace `401` with `http.StatusUnauthorized`

sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -32,93 +29,120 @@ const (
...
 			return nil, err
 		}
-
-		accessToken, err := p.getAccessToken(req)
-		if err != nil {
+		if accessToken, err = p.getAccessToken(req, service, scope); err != nil {
 			return nil, err
 		}
-
+		p.accessTokenCache = accessToken
```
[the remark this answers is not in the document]
jhendrixMSFT: I'm probably missing something here, but shouldn't the access token already be cached in the `temporal.Resource`?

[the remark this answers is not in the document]
chlowell: That caches the ACR refresh token. The protocol is (@tadelesh please correct my mistakes)
1. acquire AAD access token
1. exchange AAD access token for ACR refresh token
1. exchange ACR refresh token for ACR access token
1. authorize request with ACR access token

`p.accessTokenCache` has the ACR access token. I suppose another `temporal.Resource` could synchronize around that.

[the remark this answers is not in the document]
jhendrixMSFT: Thanks for clarifying.

Can the ACR access token be "pushed down" into the existing `temporal.Resource`, or do we need to cache both?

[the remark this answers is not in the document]
chlowell: I suppose that's possible, however could be more complex than it's worth. First complication that comes to mind is that the access and refresh token could expire separately.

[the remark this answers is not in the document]
tadelesh: Let me add some context. Each container registry shares one refresh token, so one policy created by one client (one client operate one registry) could share and cache this refresh token. With this refresh token, we could get the access token, and DIFFERENT operation (download/upload blob, get/update repository properties, get/update tag, etc.) has DIFFERENT access token. Only when you call two same operation (always happen when you upload blob chunk by chunk, or do some batch operations), the access token can be reused. Meanwhile, refresh token and access token expire separately. So, I use `temporal.Resource` to cache refresh token as it is shared, meanwhile, I use a simple string to cache access token as it is no need to take care of access token's expiration because the access token is large probably invalid for the next request (as I described below). I've added synchronization.

jhendrixMSFT (approved):

chlowell (approved):

PR #20045, opened 2023-02-22
[azeventhubs] Client identifier support

### LINKED ISSUE #15074: Event Hubs: Client Identifier Support
# Summary

When creating one of the Event Hubs client types, it can be useful for a host application to be able to uniquely identify the client instance.  The Event Hubs service also supports setting the identifier of an AMQP link, which it will associate with some error messages to make it easier to correlate errors with a given client instance.  

The event processor types currently support an informational Identifier, passed in via their respective options at creation time.  The remaining Event Hubs clients should be extended to follow this pattern.

# Scope of Work

- The options for each Event Hubs client type should be extended with a string-based `Identifier` property, defaulting to `null` if not supplied by callers.

- The  Event Hubs client types should be extended with an `Identifier` property, mapping to the `Identifer` received as part of their client options.   If the options do not specify an identifier, a unique one should be generated.

- When opening an AMQP link for receiving, the identifier should be added to a link property named `com.microsoft:receiver-name` so that it can be used by the service with error messages.   _([example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpConnectionScope.cs#L672))_

- When opening an AMQP link for receiving, the identifier should be used as the `Target` in the link settings if doing so us supported by the transport.  _([example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpConnectionScope.cs#L661))_

- When opening an AMQP link for sending, the identifier should be used as the `Source` in the link settings if doing so is supported by the transport.  _([example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpConnectionScope.cs#L771))_

- There should be no change to the logic when opening an AMQP `RequestResponse` link for management purposes; the identifier is not supported in this context.

# Success Criteria

- The API changes have been informally reviewed by the language architects and approved.

- The client options identified in the scope have been enhanced to include an `Identifier` property.

- The client types identifier in the scope have been enhanced to accept the identifier from the options, expose it via an `Identifier` property, and make use of it when creating links.

- The tests necessary for ensuring the feature works as intended have been created or adjusted and pass reliably.

- Existing tests continue to produce deterministic results and pass reliably.

richardpark-msft:
Event Hubs has an optional parameter you can pass that lets you identify a consumer link. This name will come back in error messages like when link ownership is contested. Given the distributed nature of consumers this identifier can help diagnose issues where consumers are battling over links.

Fixes #15074

### DIFF
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.5.1 (Unreleased)
+## 0.6.0 (Unreleased)
 
 ### Features Added
 
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-- Added the `ConsumerClientOptions.Identifier` field. This optional field can enhance error messages from 
+- Added the `ConsumerClientOptions.InstanceID` field. This optional field can enhance error messages from 
   Event Hubs. For example, error messages related to ownership changes for a partition will contain the 
   name of the link that has taken ownership, which can help with traceability.
 
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Features Added
 
+- Added the `ConsumerClientOptions.Identifier` field. This optional field can enhance error messages from 
+  Event Hubs. For example, error messages related to ownership changes for a partition will contain the 
+  name of the link that has taken ownership, which can help with traceability.
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Breaking Changes
 
+- `ConsumerClient.ID()` renamed to `ConsumerClient.InstanceID()`.
+
 ### Bugs Fixed
 
 ### Other Changes
--- a/sdk/messaging/azeventhubs/consumer_client.go
+++ b/sdk/messaging/azeventhubs/consumer_client.go
@@ -65,10 +65,10 @@ type ConsumerClient struct {
 	consumerGroup string
 	eventHub      string
 
-	// identifier is a customer supplied identifier that can be passed to Event Hubs.
+	// instanceID is a customer supplied instanceID that can be passed to Event Hubs.
 	// It'll be returned in error messages and can be useful for customers when
 	// troubleshooting.
-	identifier string
+	instanceID string
 
 	links        *internal.Links[amqpwrap.AMQPReceiverCloser]
 	namespace    *internal.Namespace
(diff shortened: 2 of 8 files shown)


richardpark-msft:
APIView pending for this feature.

sdk/messaging/azeventhubs/consumer_client.go
```diff
@@ -39,31 +39,40 @@ const (
...
-	TLSConfig *tls.Config
-
-	// Application ID that will be passed to the namespace.
+	// ApplicationID is used as the identifier when setting the User-Agent property.
 	ApplicationID string
 
+	// Identifier is a unique name used to identify the consumer. This can help with
+	// diagnostics as this name will be returned in error messages. By default,
+	// an identifier will be automatically generated.
+	Identifier string
```
jhendrixMSFT, line 48: What if we called this `InstanceID` instead (I made the same note on the APIView).  I think this helps separate it from `ApplicationID` and clarifies the usage.

richardpark-msft: I like that. 

There's a corresponding `ID()` function on ConsumerClient that we should change to match.

richardpark-msft: https://apiview.dev/Assemblies/Review/8245669789dd4453bea087a9afaf2705?diffRevisionId=6ec7cf3099724b8bad89e1ce538b32d9&diffOnly=False&revisionId=c445bbc6810a4e2e95565b97f7e72434&doc=False

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
...
-## 0.5.1 (Unreleased)
+## 0.6.0 (Unreleased)
 
 ### Features Added
 
+- Added the `ConsumerClientOptions.InstanceID` field. This optional field can enhance error messages from 
+  Event Hubs. For example, error messages related to ownership changes for a partition will contain the 
+  name of the link that has taken ownership, which can help with traceability.
+
 ### Breaking Changes
```
jhendrixMSFT, line 11: Need to add a note about the rename of `ID()`.

sdk/messaging/azeventhubs/partition_client.go
```diff
@@ -204,20 +203,24 @@ func (pc *PartitionClient) getEntityPath(partitionID string) string {
...
 	}
 
 	receiverOptions := &amqp.ReceiverOptions{
 		SettlementMode: to.Ptr(amqp.ModeFirst),
 		Filters: []amqp.LinkFilter{
 			amqp.LinkFilterSelector(pc.offsetExpression),
 		},
-		Properties: props,
+		Properties:    props,
+		TargetAddress: pc.identifier,
```
jhendrixMSFT, line 223: Curious why this is needed.

richardpark-msft: It doesn't appear to be critical (in my limited testing) but it's part of the spec for the feature request.

> When opening an AMQP link for receiving, the identifier should be used as the Target in the link settings if doing so us supported by the transport. ([example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpConnectionScope.cs#L661))

sdk/messaging/azeventhubs/CHANGELOG.md
richardpark-msft: Done.

richardpark-msft:
Just a note, the CI failures don't appear to be related to this PR.

jhendrixMSFT (approved):

PR #20054, opened 2023-02-23
azfile: SAS

souravgupta-msft:


sdk/storage/azfile/sas/account.go
```diff
@@ -0,0 +1,159 @@
...
+
+// AccountResourceTypes type simplifies creating the resource types string for an Azure Storage Account SAS.
+// Initialize an instance of this type and then call its String method to set AccountSASSignatureValues' ResourceTypes field.
+type AccountResourceTypes struct {
+	Service, Container, Object bool
+}
+
+// String produces the SAS resource types string for an Azure Storage account.
+// Call this method to set AccountSASSignatureValues' ResourceTypes field.
+func (rt *AccountResourceTypes) String() string {
```
gapra-msft, line 147: why do we have a parseAccountPermission but no parseAccountResourceTypes?

sdk/storage/azfile/sas/account_test.go
```diff
@@ -0,0 +1,92 @@
...
+			List:   true,
+			Create: true,
+		}, input: "rwdlc"},
+		{expected: AccountPermissions{
+			Read:   true,
+			Write:  true,
+			Delete: true,
+			List:   true,
+			Create: true,
+		}, input: "rwldc"},
```
gapra-msft, line 60: randomize the order here to test that it works

sdk/storage/azfile/sas/service.go
```diff
@@ -0,0 +1,228 @@
...
+	Version            string    `param:"sv"`  // If not specified, this defaults to Version
+	Protocol           Protocol  `param:"spr"` // See the Protocol* constants
+	StartTime          time.Time `param:"st"`  // Not specified if IsZero
+	ExpiryTime         time.Time `param:"se"`  // Not specified if IsZero
+	SnapshotTime       time.Time
+	Permissions        string  `param:"sp"` // Create by initializing a SharePermissions or FilePermissions and then call String()
+	IPRange            IPRange `param:"sip"`
+	Identifier         string  `param:"si"`
+	ShareName          string
+	FilePath           string // Ex: "directory/FileName" or "FileName". Use "" to create a Share SAS.
```
gapra-msft, line 31: Do directories in file share have SAS too? if so would path be more generic a name to cover that case too?

gapra-msft: also if they dont support it now, maybe they might in the future, so might be good to think about how to future proof this?

gapra-msft (commented):
left a few comments

tasherif-msft (approved):

sdk/storage/azfile/sas/account.go
souravgupta-msft: Added `parseAccountResourceTypes()` method.

sdk/storage/azfile/sas/account_test.go
souravgupta-msft: Done

sdk/storage/azfile/sas/service.go
souravgupta-msft: I think SAS is supported for directories as well. I checked the .NET SDK and they do have method to generate SAS in the directory client. What we can do is rename this parameter to `DirectoryOrFilePath`. Thoughts?

gapra-msft (approved):

Issue #20051, opened 2023-02-23
runtime.NewPollerFromResumeToken doesn't use custom handler


gracewilcox:
In NewPollerFromResumeTokenOptions, I specified a custom handler, but NewPollerFromResumeToken uses the async handler instead. This is because the custom handler is last in the if/else block.

Line of code: https://github.com/Azure/azure-sdk-for-go/blob/39eb62497de93d1fa1805f29c6fc83d9062a011a/sdk/azcore/runtime/poller.go#L149

jhendrixMSFT:
Fix will go out in `azcore@v1.3.2`

PR #20286, opened 2023-02-23
Favor custom PollingHandler when rehydrating a poller

jhendrixMSFT:
This ensures that the custom handler is used in the event that the custom poller has the same shape as an in-box poller.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/20051


richardpark-msft (approved):
Change looks good.

PR #20291, opened 2023-02-24
Some minor refactoring of azcore

jhendrixMSFT:
Moved ARM resource ID types and helpers to an internal package so they can be leveraged elsewhere.  The public versions are type-aliased to the matching internal implementations.
policyFunc has moved to the shared package so it can be used elsewhere. Sanitizing of URLs has been split into its own function. Moved a func-local constant to the shared package.


jhendrixMSFT:
The tracing work will build on these changes, and I wanted to keep them separate to simplify the review.  There should be no functional changes.

sdk/azcore/internal/exported/request.go
```diff
@@ -169,3 +169,14 @@ func (req *Request) Clone(ctx context.Context) *Request {
 	r2.req = req.req.Clone(ctx)
 	return &r2
 }
+
+// not exported but dependent on Request
```
chlowell, line 173: Then could it somewhere else instead, like `arm/internal`?

chlowell (approved):

sdk/azcore/internal/exported/request.go
jhendrixMSFT: Right now it's only used in `runtime`, but it will also be used in `arm/runtime` as part of the tracing work.

chlowell: Would it create an import cycle in `shared`? (I can live with it being here if there's no better alternative)

jhendrixMSFT: Unfortunately yes as that's where I originally had placed it.

PR #20350, opened 2023-03-07
Add RetryOptions.Predicate field

jhendrixMSFT:
Allows for finer-grained control over when to retry.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.4.1 (Unreleased)
 
 ### Features Added
+* Added `Predicate` to `policy.RetryOptions` for finer-grained control over when to retry.
 
 ### Breaking Changes
 
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 1.4.1 (Unreleased)
 
 ### Features Added
-* Added `Predicate` to `policy.RetryOptions` for finer-grained control over when to retry.
+* Added `ShouldRetry` to `policy.RetryOptions` for finer-grained control over when to retry.
 
 ### Breaking Changes
 
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -113,6 +113,14 @@ type RetryOptions struct {
 	// Specifying values will replace the default values.
 	// Specifying an empty slice will disable retries for HTTP status codes.
 	StatusCodes []int
+
+	// Predicate evaluates if the retry policy should retry the request.
+	// When present, the Predicate overrides comparison against the list of
+	// HTTP status codes and error checking within the retry policy. Context
+	// and NonRetriable errors remain evaluated before the Predicate.
+	// The *http.Response and error parameters are mutually exclusive.
+	// A return value of true means the retry policy should retry.
+	Predicate func(*http.Response, error) bool
 }
 
 // TelemetryOptions configures the telemetry policy's behavior.
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -114,13 +114,14 @@ type RetryOptions struct {
 	// Specifying an empty slice will disable retries for HTTP status codes.
 	StatusCodes []int
 
-	// Predicate evaluates if the retry policy should retry the request.
-	// When present, the Predicate overrides comparison against the list of
+	// ShouldRetry evaluates if the retry policy should retry the request.
+	// When specified, the function overrides comparison against the list of
 	// HTTP status codes and error checking within the retry policy. Context
-	// and NonRetriable errors remain evaluated before the Predicate.
-	// The *http.Response and error parameters are mutually exclusive.
+	// and NonRetriable errors remain evaluated before calling ShouldRetry.
+	// The *http.Response and error parameters are mutually exclusive, i.e.
+	// if one is nil, the other is not nil.
 	// A return value of true means the retry policy should retry.
-	Predicate func(*http.Response, error) bool
+	ShouldRetry func(*http.Response, error) bool
 }
 
 // TelemetryOptions configures the telemetry policy's behavior.
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -161,9 +161,9 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 			return
 		}
 
-		if options.Predicate != nil {
+		if options.ShouldRetry != nil {
 			// a non-nil Predicate overrides our HTTP status code check
-			if !options.Predicate(resp, err) {
+			if !options.ShouldRetry(resp, err) {
 				// predicate says we shouldn't retry
 				log.Write(log.EventRetryPolicy, "exit due to Predicate")
 				return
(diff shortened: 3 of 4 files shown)


sdk/azcore/policy/policy.go
```diff
@@ -113,6 +113,14 @@ type RetryOptions struct {
 	// Specifying values will replace the default values.
 	// Specifying an empty slice will disable retries for HTTP status codes.
 	StatusCodes []int
+
+	// Predicate evaluates if the retry policy should retry the request.
+	// When present, the Predicate overrides comparison against the list of
+	// HTTP status codes and error checking within the retry policy. Context
+	// and NonRetriable errors remain evaluated before the Predicate.
+	// The *http.Response and error parameters are mutually exclusive.
```
JeffreyRichter, line 121: "mutually exclusive" isn't clear enough to me. If response is nil, then error is not and vice versa, right?
You could say: If one is nil, the other is not nil.

jhendrixMSFT: Correct, I'll update the wording.

sdk/azcore/policy/policy.go
```diff
@@ -113,6 +113,14 @@ type RetryOptions struct {
...
+	// A return value of true means the retry policy should retry.
+	Predicate func(*http.Response, error) bool
```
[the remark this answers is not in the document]
jhendrixMSFT: I like the name, will change.

The request's context can be retrieved from the response (`resp.Request.Context()`).  However, the way I've currently implemented this is to evaluate the context (and NonRetriable error) first, so if the context was canceled/expired, the retry policy will return _before_ passing control to the predicate.  Are you suggesting we not do this?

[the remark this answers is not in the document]
jhendrixMSFT: I can see arguments both ways, and if we want ultimate control, then I agree that passing the context and _not_ checking any error conditions before the predicate makes sense.

sdk/azcore/runtime/policy_retry.go
```diff
@@ -165,6 +161,19 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 			return
 		}
 
+		if options.ShouldRetry != nil {
```
vibhansa-msft, line 164: Does this mean that if ShouldRetry is configured by user then SDK will not check its own list of retriable errors and responsibility is passed down to application to define what all to retry?

jhendrixMSFT: Users can't configure this, only SDK authors can.

The only errors that remain checked first are context errors and errors marked as NonRetriable.  There's discussion in the above comments if we want to change this so that `ShouldRetry` includes those errors as well.  Do you have an opinion?

sdk/azcore/policy/policy.go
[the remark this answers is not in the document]
JeffreyRichter: The more I think about this, the more I think context should NOT be passed. The retry policy should just handle it correctly and ShouldRetry should just use the response/error to make its decision. ShouldRetry should be simple and not have to consider too many other things. The context can always be canceled by some other customer code if that's what they want to do.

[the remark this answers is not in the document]
JeffreyRichter: > I like the name, will change.
> 
> The request's context can be retrieved from the response (`resp.Request.Context()`). However, the way I've currently implemented this is to evaluate the context (and NonRetriable error) first, so if the context was canceled/expired, the retry policy will return _before_ passing control to the predicate. Are you suggesting we not do this?

I like your approach here.

sdk/azcore/runtime/policy_retry.go
jhendrixMSFT: @vibhansa-msft the POR is to evaluate context/NonRetriable errors before the ShouldRetry handler.  Please let us know if this will be problematic for azcopy.

chlowell: Is this supposed to be inaccessible to users? RetryOptions is part of ClientOptions.

gapra-msft: I'm a little confused by "Users can't configure this, only SDK authors can." Since ShouldRetry is exposed in the RetryOptions, why can't a User configure this themselves?

jhendrixMSFT: You're correct, I misspoke earlier.  Conceptually, I was thinking that this might make more sense to expose only to SDK authors, but that's clearly not how it's implemented in this PR.

sdk/azcore/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.4.1 (Unreleased)
 
 ### Features Added
+* Added `Predicate` to `policy.RetryOptions` for finer-grained control over when to retry.
```
gapra-msft, line 6:
```suggestion
* Added `ShouldRetry` to `policy.RetryOptions` for finer-grained control over when to retry.
```

jhendrixMSFT: fixed

sdk/azcore/runtime/policy_retry.go
vibhansa-msft: So we are exposing this callback which user can register while creating the client and in case of failure, callback will be hit where user application can decide whether to retry or not.
My earlier concern here was SDK has a set of retriable error codes. We are going to check those first and then hit the callback in case of no match, or if the callback is registered then we skip the SDK check and rely on user to decide when to retry and when to fail.

jhendrixMSFT: When `ShouldRetry` is not `nil`, it will be called instead of checking against the configured set of HTTP status codes.  However, the retry policy will evaluate context/NonRetriable errors before the `ShouldRetry` handler is called.

gapra-msft (approved):

sdk/azcore/runtime/policy_retry_test.go
```diff
@@ -672,6 +672,45 @@ func TestRetryPolicySuccessWithPerTryTimeoutNoRetryWithBodyDownload(t *testing.T
...
+func TestRetryPolicyWithShouldRetryRetry(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	srv.AppendResponse(mock.WithStatusCode(http.StatusRequestTimeout))
+	srv.AppendResponse()
+
+	pl := exported.NewPipeline(srv, NewRetryPolicy(&policy.RetryOptions{
+		RetryDelay: time.Millisecond,
+		ShouldRetry: func(r *http.Response, err error) bool {
+			return r.StatusCode == http.StatusRequestTimeout
```
chlowell, line 703: The policy retries StatusRequestTimeout by default, so I might add a bool to track whether ShouldRetry was called or use a code the policy wouldn't retry.

chlowell (approved):

sdk/azcore/runtime/policy_retry.go
```diff
@@ -165,6 +161,19 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
...
+			// a non-nil Predicate overrides our HTTP status code check
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

sdk/azcore/runtime/policy_retry.go
```diff
@@ -165,6 +161,19 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
...
+			if !options.ShouldRetry(resp, err) {
+				// predicate says we shouldn't retry
+				log.Write(log.EventRetryPolicy, "exit due to Predicate")
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

Issue #20092, opened 2023-02-23
[azblob] file upload validation


vijayrajah:
### Bug Report
I'm trying to upload a file to azure storage (v2) using 
`github.com/Azure/azure-sdk-for-go/sdk/storage/azblob`

it is not clear (at least to me ) how to do upload validation using MD5.

I have added this in code
```
opts := azblob.UploadFileOptions{
		TransactionalContentMD5: md5sum,
		AccessTier:              to.Ptr(blob.AccessTierCool),
	}
```

I'm computing MD5 using
```
h := md5.New()
	if _, err := io.Copy(h, fil); err != nil {
		return errors.Wrap(err, "unable to compute MD5 for the file "+f)
	}
md5sum := h.Sum(nil)
```

The Md5 from service and my locally computed MD5 do not match, but the upload still proceeds, when it should fail. 

Am I missing something? is there a code sample that I can refer to?

siminsavani-msft:
Hi @vijayrajah ! If it is possible, could you please provide your code from start to finish?

vijayrajah:
Here is the function that does the upload

```
import (
	"context"
	"crypto/md5"
	"encoding/hex"
	"fmt"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
	"github.com/pkg/errors"
	"io"
	"os"
	"strings"
	"time"
)

func uploadFileToStorage(ctx context.Context, f string) error {

	fil, err := os.Open(f)
	if err != nil {
		log.Error().Err(err).Msgf("unable to open file %s", f)
		return err
	}

	fn := strings.Split(f, "/")
	blobname := fn[len(fn)-1]

	// calculate md5 sum
	h := md5.New()
	if _, err := io.Copy(h, fil); err != nil {
		return errors.Wrap(err, "unable to compute MD5 for the file "+f)
	}
	md5sum := h.Sum(nil)

	opts := azblob.UploadFileOptions{
		TransactionalContentMD5: md5sum,
		AccessTier:              to.Ptr(blob.AccessTierCool),
	}

[...]

PR #20356, opened 2023-03-07
Adding MD5 from UploadFileOptions

siminsavani-msft:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/20092

Adds MD5 from UploadFileOptions to UploadOptions and CommitBlockListOptions. 

### DIFF
--- a/sdk/storage/azblob/blockblob/models.go
+++ b/sdk/storage/azblob/blockblob/models.go
@@ -228,24 +228,26 @@ func (o *uploadFromReaderOptions) getStageBlockOptions() *StageBlockOptions {
 
 func (o *uploadFromReaderOptions) getUploadBlockBlobOptions() *UploadOptions {
 	return &UploadOptions{
-		Tags:             o.Tags,
-		Metadata:         o.Metadata,
-		Tier:             o.AccessTier,
-		HTTPHeaders:      o.HTTPHeaders,
-		AccessConditions: o.AccessConditions,
-		CPKInfo:          o.CPKInfo,
-		CPKScopeInfo:     o.CPKScopeInfo,
+		Tags:                    o.Tags,
+		Metadata:                o.Metadata,
+		Tier:                    o.AccessTier,
+		TransactionalContentMD5: o.TransactionalContentMD5,
+		HTTPHeaders:             o.HTTPHeaders,
+		AccessConditions:        o.AccessConditions,
+		CPKInfo:                 o.CPKInfo,
+		CPKScopeInfo:            o.CPKScopeInfo,
 	}
 }
 
 func (o *uploadFromReaderOptions) getCommitBlockListOptions() *CommitBlockListOptions {
 	return &CommitBlockListOptions{
-		Tags:         o.Tags,
-		Metadata:     o.Metadata,
-		Tier:         o.AccessTier,
-		HTTPHeaders:  o.HTTPHeaders,
-		CPKInfo:      o.CPKInfo,
-		CPKScopeInfo: o.CPKScopeInfo,
+		Tags:                    o.Tags,
+		Metadata:                o.Metadata,
+		Tier:                    o.AccessTier,
+		TransactionalContentMD5: o.TransactionalContentMD5,
+		HTTPHeaders:             o.HTTPHeaders,
+		CPKInfo:                 o.CPKInfo,
+		CPKScopeInfo:            o.CPKScopeInfo,
 	}
 }
 
--- a/sdk/storage/azblob/client_test.go
+++ b/sdk/storage/azblob/client_test.go
@@ -8,6 +8,8 @@ package azblob_test
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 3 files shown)


jhendrixMSFT (approved):
Might be worth doing a scrub to see if there are other places where we missed propagating option values.

sdk/storage/azblob/client_test.go
```diff
@@ -335,6 +337,78 @@ func (s *AZBlobUnrecordedTestsSuite) TestUploadAndDownloadFileNonZeroOffsetAndCo
...
+
+}
+
+func (s *AZBlobUnrecordedTestsSuite) TestMD5LargeFileUpload() {
+	_require := require.New(s.T())
+	fileSize := 257 * 1024 * 1024 // 257 MB file, 256 MB is the max for calculate md5 and service md5 won't match
+
+	md5sum, resp, err := performUploadWithMD5Check(s.T(), _require, s.T().Name(), fileSize)
+
+	_require.ErrorContains(err, "400 The MD5 value specified in the request did not match with the MD5 value calculated by the server.")
```
souravgupta-msft, line 360: Why are we getting this error for large file uploads?

sdk/storage/azblob/blockblob/models.go
```diff
@@ -228,24 +228,26 @@ func (o *uploadFromReaderOptions) getStageBlockOptions() *StageBlockOptions {
...
-		Metadata:         o.Metadata,
-		Tier:             o.AccessTier,
-		HTTPHeaders:      o.HTTPHeaders,
-		AccessConditions: o.AccessConditions,
-		CPKInfo:          o.CPKInfo,
-		CPKScopeInfo:     o.CPKScopeInfo,
+		Tags:                    o.Tags,
+		Metadata:                o.Metadata,
+		Tier:                    o.AccessTier,
+		TransactionalContentMD5: o.TransactionalContentMD5,
```
vibhansa-msft, line 234: There are two features, one is to just set the MD5 sum in the blob and other one is transactional MD5. For transactional options crc64 is also supported. Transactional fields are only for validation and will not set any properties in the blob. Just double check that we are good with all these combinations as I do not see transactional CRC here, just had doubts on we are still missing some things.

seanmcc-msft: We need to be careful about the way we chose to expose transactional MD-5 and CRC-64 to the customer.  We will be implement end-to-end content validation in the Go SDK this fall, after the FE has implement trailing CRC-64 header.

This work will look something like this - https://github.com/Azure/azure-sdk-for-net/pull/29778

We do not need to calculate CRC-64 or MD5 within the Go SDK at this time, but we need to design our public interfaces in a way that will support this option in the future.

seanmcc-msft: This is how UploadOptions looks in .NET - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/Models/BlobUploadOptions.cs

It has UploadTransferValidationOptions as a property - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Common/src/UploadTransferValidationOptions.cs

I recommend we do something similar in Go to future proof out public interfaces.

jaschrep-msft: Agree with Sean. In .NET we had to retire our `byte[] contentHash` arguments and replace them with a whole configuration setup. It would be best to avoid writing "instant legacy" APIs and this probably merits further offline discussion of how to balance immediate needs with medium-term plans for transactional content integrity.

seanmcc-msft: In .NET, if the customer populates `TransferValidationOptions.PrecalculatedChecksum` and `TransferValidationOptions.ChecksumAlgorithm`, the SDK uses their pre-calculated checksum.  If the customer only populates `TransferValidationOptions.ChecksumAlgorithm`, the SDK calculates the checksum for the user.

We don't want to implement the SDK calculating the checksum for the user until ~6 months from now when the service supports trailing checksum footers, but we need to design our public interface to be compatible with this functionality.

sdk/storage/azblob/client_test.go
jaschrep-msft: a transactional md5 will always be calculated by the service. If the service is failing the transactional md5 check it's because the check is actually failing. I imagine this is because you've configured the transactional md5 on put blocklist. A transactional md5 for that API is literally checking the content integrity of the blocklist itself (it's one of the only places .NET SDK currently doesn't calculate one).

Transactional md5 validates individual requests and nothing more, meaning it would need to validate individual blocks. If one large md5 is being passed in for a partitioned upload, the client should fail fast since the checksum cannot be split.

sdk/storage/azblob/blockblob/models.go
jaschrep-msft: We should fail fast. If a customer provides a transactional checksum that the SDK cannot make use of, we shouldn't pretend an integrity check has occurred when it actually hasn't.

In .NET, before we had SDK calculation on caller behalf, we didn't accept the parameter on any API that *had the possibility* of splitting an upload, as we didn't want customers to write code that suddenly broke once they passed a file size threshold they were not aware of. Customers had to use direct to REST apis if they wanted this feature.

PR #20367, opened 2023-03-08
Synchronize goroutines within GetToken

### LINKED ISSUE #20044: Sharing a TokenCredential shouldn't result in excessive refresh requests
If N clients are sharing the same `azcore.TokenRequest` with the same scopes, when the token needs to be refreshed, it should only happen once, not once per client.

This might already be the case today, but needs confirmation and possibly documentation.

chlowell:
Sharing a credential instance among goroutines can produce redundant token requests because goroutines may race to populate the credential's cache. This PR refactors `GetToken()` into two callbacks, one for silent auth (checking the cache) and one for requesting a new token, and adds a new `syncer` type to synchronize them. (Credentials that don't distinguish checking the cache from requesting a token use the same method for both callbacks.) Credentials will allow only one goroutine to authenticate at a time, and waiting goroutines will check the cache again when they wake up. Closes #20044

### DIFF
--- a/sdk/azidentity/device_code_credential.go
+++ b/sdk/azidentity/device_code_credential.go
@@ -8,7 +8,6 @@ package azidentity
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
@@ -72,11 +71,10 @@ type DeviceCodeMessage struct {
 // If a web browser is available, InteractiveBrowserCredential is more convenient because it
 // automatically opens a browser to the login page.
 type DeviceCodeCredential struct {
-	account                    public.Account
-	additionallyAllowedTenants []string
-	client                     publicClient
-	tenant                     string
-	userPrompt                 func(context.Context, DeviceCodeMessage) error
+	account public.Account
+	client  publicClient
+	s       *syncer
+	prompt  func(context.Context, DeviceCodeMessage) error
 }
 
 // NewDeviceCodeCredential creates a DeviceCodeCredential. Pass nil to accept default options.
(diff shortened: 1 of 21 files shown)


sdk/azidentity/device_code_credential.go
```diff
@@ -90,50 +88,44 @@ func NewDeviceCodeCredential(options *DeviceCodeCredentialOptions) (*DeviceCodeC
...
-		return azcore.AccessToken{}, errors.New(credNameDeviceCode + ": GetToken() requires at least one scope")
-	}
-	tenant, err := resolveTenant(c.tenant, opts.TenantID, c.additionallyAllowedTenants)
+	return c.s.GetToken(ctx, opts)
+}
+
+func (c *DeviceCodeCredential) requestToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	dc, err := c.client.AcquireTokenByDeviceCode(ctx, opts.Scopes, public.WithClaims(opts.Claims), public.WithTenantID(opts.TenantID))
 	if err != nil {
 		return azcore.AccessToken{}, err
```
jhendrixMSFT, line 105: How come we no longer do `newAuthenticationFailedErrorFromMSALError()` in this error case?

sdk/azidentity/managed_identity_credential.go
```diff
@@ -107,17 +110,17 @@ func (c *ManagedIdentityCredential) GetToken(ctx context.Context, opts policy.To
...
-	if err != nil {
-		return azcore.AccessToken{}, err
-	}
-	logGetTokenSuccess(c, opts)
+	opts.Scopes = []string{strings.TrimSuffix(opts.Scopes[0], defaultSuffix)}
+	return c.s.GetToken(ctx, opts)
+}
+
+func (c *ManagedIdentityCredential) requestToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	ar, err := c.client.AcquireTokenByCredential(ctx, opts.Scopes, confidential.WithClaims(opts.Claims), confidential.WithTenantID(opts.TenantID))
```
jhendrixMSFT, line 118: Looks like some new options are being passed here?

sdk/azidentity/device_code_credential.go
jhendrixMSFT: Looks like this was moved to `syncer.GetToken()`.

sdk/azidentity/managed_identity_credential.go
chlowell: And shouldn't be, thanks for noticing!

jhendrixMSFT (approved):

Issue #20342, opened 2023-03-06
DetailedCostReportClient.BeginCreateOperation LRO returns wrong terminal success status


jhendrixMSFT:
It's returning `Completed` when per ARM RPC spec it should return `Succeeded`.  This causes `PollUntilDone()` to essentially be in an infinite loop.

jhendrixMSFT:
Since the service might not be able to change this due to back-compat concerns, we've added `Completed` to the list of terminal success status codes.  Fix will be included in the next update of `azcore` post `v1.4.0`.

jhendrixMSFT:
On further investigation, it appears the service is returning a provisioning state of `Cancelled` to indicate that the reservation has been cancelled, _not_ that the LRO has been cancelled.  Provisioning state is used to communicate the status of an LRO, not the state of the resource.

jhendrixMSFT:
Ping @ccmixpdevs

flanakin:
Can you remove the `Cost Management - Query` label? This is not using Query API.

PR #20370, opened 2023-03-08
Added some additional LRO terminal state values

jhendrixMSFT:
While they're non-conformant per the RPC spec, we've seen these in the wild and changing them now would be breaking for REST clients.

Fixes
- https://github.com/Azure/azure-sdk-for-go/issues/19209
- https://github.com/Azure/azure-sdk-for-go/issues/20342

### DIFF
--- a/sdk/azcore/internal/pollers/util_test.go
+++ b/sdk/azcore/internal/pollers/util_test.go
@@ -22,8 +22,10 @@ import (
 func TestIsTerminalState(t *testing.T) {
 	require.False(t, IsTerminalState("Updating"), "Updating is not a terminal state")
 	require.True(t, IsTerminalState("Succeeded"), "Succeeded is a terminal state")
+	require.True(t, IsTerminalState("Completed"), "Completed is a terminal state")
 	require.True(t, IsTerminalState("failed"), "failed is a terminal state")
 	require.True(t, IsTerminalState("canceled"), "canceled is a terminal state")
+	require.True(t, IsTerminalState("cancelled"), "cancelled is a terminal state")
 }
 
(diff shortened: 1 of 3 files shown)


sdk/azcore/internal/pollers/util_test.go
```diff
@@ -22,8 +22,10 @@ import (
 func TestIsTerminalState(t *testing.T) {
 	require.False(t, IsTerminalState("Updating"), "Updating is not a terminal state")
 	require.True(t, IsTerminalState("Succeeded"), "Succeeded is a terminal state")
+	require.True(t, IsTerminalState("Completed"), "Completed is a terminal state")
 	require.True(t, IsTerminalState("failed"), "failed is a terminal state")
 	require.True(t, IsTerminalState("canceled"), "canceled is a terminal state")
+	require.True(t, IsTerminalState("cancelled"), "cancelled is a terminal state")
```
chlowell, line 28: Should we also test "Cancelled"?

chlowell (approved):

sdk/azcore/internal/pollers/util_test.go
jhendrixMSFT: We're supposed to, and do at present, perform a case-insensitive comparison.  Perhaps instead of testing both permutations, what if we used input values that don't match either permutation (e.g. `cancELLed`)?

chlowell: Sounds good, I like it 👍

jhendrixMSFT: Fixed

JeffreyRichter (approved):

PR #20379, opened 2023-03-09
Adding checks during SAS creation

souravgupta-msft:


sdk/storage/azblob/sas/account.go
```diff
@@ -29,15 +30,19 @@ type AccountSignatureValues struct {
...
 	IPRange       IPRange   `param:"sip"`
-	ResourceTypes string    `param:"srt"` // Create by initializing AccountSASResourceTypes and then call String()
+	ResourceTypes string    `param:"srt"` // Create by initializing AccountResourceTypes and then call String()
 }
 
 // SignWithSharedKey uses an account's shared key credential to sign this signature values to produce
 // the proper SAS query parameters.
 func (v AccountSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKeyCredential) (QueryParameters, error) {
 	// https://docs.microsoft.com/en-us/rest/api/storageservices/Constructing-an-Account-SAS
+	if sharedKeyCredential == nil {
```
jhendrixMSFT, line 42: The convention is to not perform `nil` checks on required parameters.

sdk/storage/azblob/sas/service.go
```diff
@@ -51,7 +53,11 @@ func getDirectoryDepth(path string) string {
 // SignWithSharedKey uses an account's SharedKeyCredential to sign this signature values to produce the proper SAS query parameters.
 func (v BlobSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKeyCredential) (QueryParameters, error) {
 	if sharedKeyCredential == nil {
```
jhendrixMSFT, line 55: This should just be removed.

sdk/storage/azblob/sas/account.go
souravgupta-msft: `SignWithSharedKey` is an exposed method and can be directly called as `sas.SignWithSharedKey()`. So, I added this check since `nil` can be passed as a `*SharedKeyCredential` parameter.

jhendrixMSFT: Understood.  However, having this panic due to a missing required parameter is ok and is the convention (if the panic would occur at some other point post this method call then checking for `nil` is ok).

souravgupta-msft: Done. Removed this check from queues package as well.

sdk/storage/azblob/sas/service.go
souravgupta-msft: Done

vibhansa-msft (approved):

jhendrixMSFT (approved):

gapra-msft (approved):

sdk/storage/azblob/sas/service.go
```diff
@@ -50,8 +51,8 @@ func getDirectoryDepth(path string) string {
 
 // SignWithSharedKey uses an account's SharedKeyCredential to sign this signature values to produce the proper SAS query parameters.
 func (v BlobSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKeyCredential) (QueryParameters, error) {
-	if sharedKeyCredential == nil {
-		return QueryParameters{}, fmt.Errorf("cannot sign SAS query without Shared Key Credential")
+	if v.ExpiryTime.IsZero() || v.Permissions == "" {
```
tasherif-msft, line 54: I think this should be an or not `||`, both are required, are they not?

stotz89:
Hi @souravgupta-msft,
This PR has introduced a check within the `SignWithSharedKey` method:

```
       if v.ExpiryTime.IsZero() || v.Permissions == "" {
		return QueryParameters{}, errors.New("service SAS is missing at least one of these: ExpiryTime or Permissions")
	}
```

When the SAS should be associated with stored access policy the [documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#specify-the-access-policy) says expiry and permissions must be omitted if those information are specified in the associated stored access policy. 

Currently this change leads to an error during signing which has worked previously. My code looks like
```
sasQueryParams, err := sas.BlobSignatureValues{
		Protocol:      sas.ProtocolHTTPS,
		StartTime:     *new(time.Time),
		ExpiryTime:    *new(time.Time),
		Identifier:    policyName,
		ContainerName: containerName,
	}.SignWithSharedKey(credential)

```

PR #20399, opened 2023-03-13
Append Blob Tests

siminsavani-msft:
Adding missing tests for APIs in AppendBlob

Ignored/removed APIs:
1. SetTier - deprecated from appendblob client, this not an api for appendblob
2. CopyFromURL - deprecated from appendblob client, this not api for appendblob
3. AbortCopyFromURL - existing blob base client tests
4. DownloadFile - existing blob base client tests

Written tests for:
1. SetMetadata
2. SetHTTPHeaders
3. SetTags

### DIFF
--- a/sdk/storage/azblob/appendblob/client.go
+++ b/sdk/storage/azblob/appendblob/client.go
@@ -257,9 +257,9 @@ func (ab *Client) SetLegalHold(ctx context.Context, legalHold bool, options *blo
 }
 
 // SetTier
-// Deprecated: SetTier only works for page blob in premium storage account and block blob in block storage account.
+// Deprecated: SetTier only works for page blob in premium storage account and block blob in blob storage account.
 func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.SetTierOptions) (blob.SetTierResponse, error) {
-	return blob.SetTierResponse{}, errors.New("Operation will not work on this blob type. SetTier only works for page blob in premium storage account and block blob in block storage account.")
+	return blob.SetTierResponse{}, errors.New("Operation will not work on this blob type. SetTier only works for page blob in premium storage account and block blob in blob storage account.")
 }
 
 // SetExpiry operation sets an expiry time on an existing blob. This operation is only allowed on Hierarchical Namespace enabled accounts.
--- a/sdk/storage/azblob/appendblob/client.go
+++ b/sdk/storage/azblob/appendblob/client.go
@@ -8,6 +8,7 @@ package appendblob
 
 import (
 	"context"
+	"errors"
 	"io"
 	"os"
 	"time"
@@ -255,15 +256,10 @@ func (ab *Client) SetLegalHold(ctx context.Context, legalHold bool, options *blo
 	return ab.BlobClient().SetLegalHold(ctx, legalHold, options)
 }
 
+// SetTier
 // Deprecated: SetTier only works for page blob in premium storage account and block blob in block storage account.
-// SetTier operation sets the tier on a blob. The operation is allowed on a page
-// blob in a premium storage account and on a block blob in a blob storage account (locally
-// redundant storage only). A premium page blob's tier determines the allowed size, IOPS, and
-// bandwidth of the blob. A block blob's tier determines Hot/Cool/Archive storage type. This operation
-// does not update the blob's ETag.
-// For detailed information about block blob level tiering see https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers.
 func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.SetTierOptions) (blob.SetTierResponse, error) {
-	return ab.BlobClient().SetTier(ctx, tier, o)
+	return blob.SetTierResponse{}, errors.New("Operation will not work on this blob type. SetTier only works for page blob in premium storage account and block blob in block storage account.")
 }
 
 // SetExpiry operation sets an expiry time on an existing blob. This operation is only allowed on Hierarchical Namespace enabled accounts.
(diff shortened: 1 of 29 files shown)


sdk/storage/azblob/appendblob/client_test.go
```diff
@@ -2381,3 +2382,573 @@ func (s *AppendBlobUnrecordedTestsSuite) TestAppendBlobSetExpiryToAbsolute() {
...
+	_require.NoError(err)
+
+	containerClient := testcommon.CreateNewContainer(context.Background(), _require, testcommon.GenerateContainerName(testName), svcClient)
+	defer testcommon.DeleteContainer(context.Background(), _require, containerClient)
+
+	abClient := getAppendBlobClient(testcommon.GenerateBlobName(testName), containerClient)
+	_, err = abClient.Create(context.Background(), nil)
+	_require.Nil(err)
+
+	proposedLeaseIDs := []*string{to.Ptr("c820a799-76d7-4ee2-6e15-546f19325c2c"), to.Ptr("326cc5e1-746e-4af8-4811-a50e6629a8ca")}
```
gapra-msft, line 2921: do we need this to be an array?

sdk/storage/azblob/appendblob/client.go
```diff
@@ -255,16 +255,6 @@ func (ab *Client) SetLegalHold(ctx context.Context, legalHold bool, options *blo
 	return ab.BlobClient().SetLegalHold(ctx, legalHold, options)
 }
 
-// SetTier operation sets the tier on a blob. The operation is allowed on a page
-// blob in a premium storage account and on a block blob in a blob storage account (locally
-// redundant storage only). A premium page blob's tier determines the allowed size, IOPS, and
-// bandwidth of the blob. A block blob's tier determines Hot/Cool/Archive storage type. This operation
-// does not update the blob's ETag.
-// For detailed information about block blob level tiering see https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers.
-func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.SetTierOptions) (blob.SetTierResponse, error) {
```
gapra-msft, line 264: Was this released already?

sdk/storage/azblob/appendblob/client.go
jhendrixMSFT, line 1: Can we get more info on the APIs that are being removed?  Do they just not work?  Something else?  It's a breaking change and it's highly likely that we can't take it and would just have to `// Deprecate` them with an explanation why.

jhendrixMSFT (changes requested):

sdk/storage/azblob/appendblob/client.go
```diff
@@ -326,12 +316,6 @@ func (ab *Client) GetTags(ctx context.Context, o *blob.GetTagsOptions) (blob.Get
 	return ab.BlobClient().GetTags(ctx, o)
 }
 
-// CopyFromURL synchronously copies the data at the source URL to a block blob, with sizes up to 256 MB.
-// For more information, see https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url.
-func (ab *Client) CopyFromURL(ctx context.Context, copySource string, o *blob.CopyFromURLOptions) (blob.CopyFromURLResponse, error) {
```
vibhansa-msft, line 331: Any reason to remove this? AppendBlockFromURL is supported on AppendBlob so do we have a seperate API for that.

sdk/storage/azblob/appendblob/client.go
vibhansa-msft: "The Set Blob Tier operation sets the access tier on a blob. The operation is allowed on a page blob in a premium storage account and on a block blob in a blob storage or general purpose v2 account. "
Appears its not supported on AppendBlob as such.

jhendrixMSFT: This is unfortunate.  My suggestion is to add a `// Deprecated: <message>` comment explaining this instead of removing it.

siminsavani-msft: I can add a comment instead of removing it if that is the general consensus.

sdk/storage/azblob/appendblob/client.go
siminsavani-msft: > SetTier - removed from appendblob client, this not an api for appendblob
> CopyFromURL - removed from appendblob client, this not api for appendblob

Both of these APIs cannot be used for appendblob. Technically, the service will return an error that states that the API won't work on this particular blobtype. I felt that it was misleading having these APIs on appendblob when they don't actually apply.

sdk/storage/azblob/appendblob/client.go
jhendrixMSFT: I'd also remove all the doc comments (you might have to keep a small `// SetTier` doc comment) and have the API return an error stating that this API doesn't work for append blobs.

sdk/storage/azblob/appendblob/client.go
siminsavani-msft: AppendBlockFromURL is a separate API

vibhansa-msft (approved):

gapra-msft (approved):

sdk/storage/azblob/appendblob/client.go
```diff
@@ -326,6 +327,7 @@ func (ab *Client) GetTags(ctx context.Context, o *blob.GetTagsOptions) (blob.Get
 	return ab.BlobClient().GetTags(ctx, o)
 }
 
+// Deprecated: CopyFromURL works only with block blob
```
jhendrixMSFT, line 330: The `// Deprecated` comment must be the last doc comment line.

I think you can also shorten the entire doc comment so the deprecation is more noticeable.  Maybe something like this.
```go
// CopyFromURL
// Deprecated: CopyFromURL works only with block blob
```

siminsavani-msft: fixed!

jhendrixMSFT (approved):

sdk/storage/azblob/appendblob/client.go
```diff
@@ -255,14 +256,10 @@ func (ab *Client) SetLegalHold(ctx context.Context, legalHold bool, options *blo
...
-// blob in a premium storage account and on a block blob in a blob storage account (locally
-// redundant storage only). A premium page blob's tier determines the allowed size, IOPS, and
-// bandwidth of the blob. A block blob's tier determines Hot/Cool/Archive storage type. This operation
-// does not update the blob's ETag.
-// For detailed information about block blob level tiering see https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers.
+// SetTier
+// Deprecated: SetTier only works for page blob in premium storage account and block blob in block storage account.
 func (ab *Client) SetTier(ctx context.Context, tier blob.AccessTier, o *blob.SetTierOptions) (blob.SetTierResponse, error) {
-	return ab.BlobClient().SetTier(ctx, tier, o)
+	return blob.SetTierResponse{}, errors.New("Operation will not work on this blob type. SetTier only works for page blob in premium storage account and block blob in block storage account.")
```
souravgupta-msft, line 262:
```suggestion
	return blob.SetTierResponse{}, errors.New("Operation will not work on this blob type. SetTier only works for page blob in premium storage account and block blob in blob storage account.")
```

PR #20403, opened 2023-03-14
genertor tool add --original-commit flag

Alancere:
Whether to update the commit id in the autorest.md file

### DIFF
--- a/eng/tools/generator/cmd/v2/common/generation.go
+++ b/eng/tools/generator/cmd/v2/common/generation.go
@@ -20,14 +20,14 @@ import (
 )
 
 type GenerateContext struct {
-	SDKPath          string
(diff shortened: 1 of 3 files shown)


eng/tools/generator/cmd/v2/common/generation.go
```diff
@@ -27,6 +27,7 @@ type GenerateContext struct {
 	SpecReadmeFile   string
 	SpecReadmeGoFile string
 	SpecRepoURL      string
+	OriginalCommit   bool
```
tadelesh, line 30:
```suggestion
	UpdateSpecVersion   bool
```

You also need to change the bool logic.

tadelesh (approved):

PR #20404, opened 2023-03-14
Use underlying error when checking for context errors

jhendrixMSFT:
There is a race between error propagation and closing a context's
deadline channel.  The result is that while the API that takes a context
returns the correct error, context.Err() might return nil.
    
See https://github.com/golang/go/issues/31863 for more info.

### DIFF
--- a/sdk/azidentity/default_azure_credential.go
+++ b/sdk/azidentity/default_azure_credential.go
@@ -182,11 +182,11 @@ func (w *timeoutWrapper) GetToken(ctx context.Context, opts policy.TokenRequestO
 		c, cancel := context.WithTimeout(ctx, w.timeout)
 		defer cancel()
 		tk, err = w.mic.GetToken(c, opts)
-		if err == nil {
+		if isAuthFailedDueToContext(err) {
+			err = newCredentialUnavailableError(credNameManagedIdentity, "managed identity timed out")
+		} else {
 			// some managed identity implementation is available, so don't apply the timeout to future calls
 			w.timeout = 0
-		} else if isAuthFailedDueToContext(err) {
-			err = newCredentialUnavailableError(credNameManagedIdentity, "managed identity timed out")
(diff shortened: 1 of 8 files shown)


RickWinter (approved):

sdk/azidentity/default_azure_credential.go
```diff
@@ -182,14 +182,26 @@ func (w *timeoutWrapper) GetToken(ctx context.Context, opts policy.TokenRequestO
 		c, cancel := context.WithTimeout(ctx, w.timeout)
 		defer cancel()
 		tk, err = w.mic.GetToken(c, opts)
-		if ce := c.Err(); errors.Is(ce, context.DeadlineExceeded) {
-			err = newCredentialUnavailableError(credNameManagedIdentity, "managed identity timed out")
-		} else {
+		if err == nil {
```
chlowell, line 185: We want to remove the timeout when a managed identity token service is available, even if we can't get a token from it. Some errors indicate availability (e.g. IMDS may respond 400), so we should keep the timeout only when the error is related to our Context.

sdk/azidentity/default_azure_credential_test.go
```diff
@@ -171,7 +171,14 @@ type delayPolicy struct {
 }
 
 func (p *delayPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
-	time.Sleep(p.delay)
+	if p.delay > 0 {
+		select {
+		case <-req.Raw().Context().Done():
+			return nil, req.Raw().Context().Err()
```
chlowell, line 177: 🤞😁🤞

sdk/azidentity/default_azure_credential.go
jhendrixMSFT: Fixed.

chlowell (approved):

Issue #20355, opened 2023-03-07
Support bearer challenge for identity based managed disks


jhendrixMSFT:
OAuth with managed disks, a special kind of page blob, needs to support bearer challenge.

vibhansa-msft:
This goes in Identity package or azidentity alredy has the support and we need to add it to azblob sdk.

jhendrixMSFT:
The bearer token policy has extensibility points for handling challenges.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/policy/policy.go#L136

Here's an example of how keyvault uses it.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/keyvault/internal/challenge_policy.go#L54-L55

PR #20418, opened 2023-03-15
Added support for storage managed disk challenge

gapra-msft:

Resolves https://github.com/Azure/azure-sdk-for-go/issues/20355

### DIFF
--- a/sdk/storage/azblob/appendblob/client.go
+++ b/sdk/storage/azblob/appendblob/client.go
@@ -34,7 +34,7 @@ type Client base.CompositeClient[generated.BlobClient, generated.AppendBlobClien
 //   - cred - an Azure AD credential, typically obtained via the azidentity module
 //   - options - client options; pass nil to accept the default values
 func NewClient(blobURL string, cred azcore.TokenCredential, options *ClientOptions) (*Client, error) {
-	authPolicy := runtime.NewBearerTokenPolicy(cred, []string{shared.TokenScope}, nil)
+	authPolicy := shared.NewStorageChallengePolicy(cred)
 	conOptions := shared.GetClientOptions(options)
 	conOptions.PerRetryPolicies = append(conOptions.PerRetryPolicies, authPolicy)
 	pl := runtime.NewPipeline(exported.ModuleName,
--- a/sdk/storage/azblob/internal/shared/challenge_policy.go
+++ b/sdk/storage/azblob/internal/shared/challenge_policy.go
@@ -8,7 +8,6 @@ package shared
 
 import (
 	"errors"
-	"fmt"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
@@ -79,7 +78,6 @@ func parseTenant(url string) string {
 
 func (s *storageAuthorizer) parseChallenge(resp *http.Response) error {
 	authHeader := resp.Header.Get("WWW-Authenticate")
-	fmt.Println(authHeader)
 	if authHeader == "" {
 		return &challengePolicyError{err: errors.New("response has no WWW-Authenticate header for challenge authentication")}
 	}
@@ -101,7 +99,6 @@ func (s *storageAuthorizer) parseChallenge(resp *http.Response) error {
 	}
 
 	s.tenantID = parseTenant(vals["authorization_uri"])
-	fmt.Println(s.tenantID)
 
 	scope := vals["resource_id"]
 	if scope == "" {
@@ -112,6 +109,5 @@ func (s *storageAuthorizer) parseChallenge(resp *http.Response) error {
 		scope += "/.default"
 	}
 	s.scopes = []string{scope}
-	fmt.Println(s.scopes)
 	return nil
 }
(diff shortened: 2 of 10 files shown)


sdk/storage/azblob/appendblob/client.go
```diff
@@ -34,7 +34,7 @@ type Client base.CompositeClient[generated.BlobClient, generated.AppendBlobClien
 //   - cred - an Azure AD credential, typically obtained via the azidentity module
 //   - options - client options; pass nil to accept the default values
 func NewClient(blobURL string, cred azcore.TokenCredential, options *ClientOptions) (*Client, error) {
-	authPolicy := runtime.NewBearerTokenPolicy(cred, []string{shared.TokenScope}, nil)
+	authPolicy := shared.NewStorageChallengePolicy(cred)
```
jhendrixMSFT, line 37: Does this need to be swapped out?  I thought we only needed this for page blobs.

gapra-msft: Eventually we will want to take advantage of the ability to correct the tenant ID from the challenge (once the latest identity beta is GAed). That will be needed for all the clients anyway, so should I remove this now and add it back once we upgrade?

jhendrixMSFT: Ah I see, makes sense to keep this then.

sdk/storage/azblob/internal/shared/challenge_policy_test.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func TestParseTenant(t *testing.T) {
+	actual := parseTenant("")
+	require.Empty(t, actual)
+
+	expected := "00000000-0000-0000-0000-000000000000"
+	sampleURL := "https://login.microsoftonline.com/" + expected
+	actual = parseTenant(sampleURL)
+	require.Equal(t, expected, actual, "tenant was not properly parsed, got %s, expected %s", actual, expected)
```
jhendrixMSFT, line 85: When using `require.Equal()`, if the values don't match, the function will print out got/want for you.

gapra-msft: Gotcha, will modify this message

jhendrixMSFT:
Is it possible to add a live/recorded test at least for page blobs to test this?

sdk/storage/azblob/internal/shared/challenge_policy.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// parses Tenant ID from auth challenge
+// https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/authorize
+func parseTenant(url string) string {
+	if url == "" {
+		return ""
+	}
+	parts := strings.Split(url, "/")
+	tenant := parts[3]
```
vibhansa-msft, line 74: Is is safe to assume this or we shall have a check on length of parts as well, may cause panic if its not met.

sdk/storage/azblob/appendblob/client.go
```diff
@@ -34,7 +34,7 @@ type Client base.CompositeClient[generated.BlobClient, generated.AppendBlobClien
 //   - cred - an Azure AD credential, typically obtained via the azidentity module
 //   - options - client options; pass nil to accept the default values
 func NewClient(blobURL string, cred azcore.TokenCredential, options *ClientOptions) (*Client, error) {
-	authPolicy := runtime.NewBearerTokenPolicy(cred, []string{shared.TokenScope}, nil)
```
vibhansa-msft, line 37: does it make sense to make this TokenScope also a configurable option. Not sure in what condition user will need it but having it as an option in the API means there are cases where user want to override this.

sdk/storage/azblob/internal/shared/challenge_policy.go
souravgupta-msft: You can try `url.Parse(u)`.

sdk/storage/azblob/appendblob/client.go
gapra-msft: @jhendrixMSFT is that 

> Is it possible to add a live/recorded test at least for page blobs to test this?

A live/recorded test might be hard since the Disk needs a specific role (not sure how to do this programmatically) and then we have to invoke the compute management plane SDK. I've added a comment of my test and what is required for the test to work. The mock tests should be sufficient

gapra-msft: I think this might go against the original intent of the Identity SDK. It seems like exposing TokenScope is a decision the Identity SDK should consider, not us.

sdk/storage/azblob/internal/shared/challenge_policy.go
gapra-msft: I added a check

vibhansa-msft (approved):

souravgupta-msft (approved):

sdk/storage/azblob/appendblob/client.go
vibhansa-msft: Fact that we are able to pass this as input while creating new policy, I guess identity already has exposed it. We are not giving user a choice to pass this as an argument and hardcoding it here.

siminsavani-msft (approved):

sdk/storage/azblob/appendblob/client.go
gapra-msft: Identity/core has to expose it to SDK devs, which they did. 

Additionally, none of the other platforms support setting the token scope, so I don't think we need to be reinventing the wheel here.

PR #20429, opened 2023-03-17
Cleanup after SetImmutabilityPolicy() and SetLegalHold() tests

souravgupta-msft:
- Delete containers having immutability policies and legal holds in live test runs
- Add OAuth in azblob pipeline

### DIFF
--- a/sdk/storage/azblob/internal/testcommon/clients_auth.go
+++ b/sdk/storage/azblob/internal/testcommon/clients_auth.go
@@ -46,6 +46,8 @@ const (
 	AccountNameEnvVar           = "AZURE_STORAGE_ACCOUNT_NAME"
 	AccountKeyEnvVar            = "AZURE_STORAGE_ACCOUNT_KEY"
 	DefaultEndpointSuffixEnvVar = "AZURE_STORAGE_ENDPOINT_SUFFIX"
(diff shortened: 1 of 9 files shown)


sdk/storage/azblob/internal/testcommon/clients_auth.go
```diff
@@ -364,3 +368,27 @@ func GetAccountSAS(permissions sas.AccountPermissions, resourceTypes sas.Account
 
 	return sasQueryParams.Encode(), nil
 }
+
+func DeleteContainerUsingManagementClient(_require *require.Assertions, accountType TestAccountType, containerName string) {
```
gapra-msft, line 372: Do we need to use the management client? We could also list containers in the account with container name prefix or get container properties and see if a lease is held on the client, break lease if needed and then delete

vibhansa-msft (approved):

sdk/storage/azblob/internal/testcommon/clients_auth.go
souravgupta-msft: In this case, Immutability policy is set on the account level. So, the container deletion fails using the client created from the azblob package. This is why I am using the management client here.
Get properties on the container returns `available` in the lease state and `unlocked` in the lease status. So, the container doesn't have lease here.

PR #20444, opened 2023-03-21
Remove references to the previous GO SDK.

RickWinter:
Update dependencies in tooling


eng/tools/internal/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/Azure/azure-sdk-for-go/eng/tools/internal
 go 1.18
 
 require (
-	github.com/Azure/azure-sdk-for-go v54.2.1+incompatible
+	//github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
```
jhendrixMSFT, line 6: Remove?

eng/tools/apidiff/go_mod_tidy_hack.go
```diff
@@ -1,13 +0,0 @@
-// +build modhack
```
jhendrixMSFT, line 1: Removing these might be problematic.  Unfortunately, it's also hard to properly test.

jhendrixMSFT (approved):

Issue #16112, opened 2021-11-08
Enable race detection configuration


RickWinter:
### Feature Request

Go has a built in race detector:
https://golang.org/doc/articles/race_detector

Enable this as a configuration in the pipelines

PR #20467, opened 2023-03-24
Enable Go race detector in CI

jhendrixMSFT:

### DIFF
--- a/sdk/messaging/azeventhubs/internal/namespace_test.go
+++ b/sdk/messaging/azeventhubs/internal/namespace_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -117,7 +118,7 @@ func TestNamespaceNegotiateClaimRenewal(t *testing.T) {
 	}
 
 	var errorsLogged []error
-	nextRefreshDurationChecks := 0
+	var nextRefreshDurationChecks atomic.Int32
 
 	ns.newClientFn = func(ctx context.Context) (amqpwrap.AMQPClient, error) {
 		return &amqpwrap.AMQPClientWrapper{Inner: &amqp.Client{}}, nil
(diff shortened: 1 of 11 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/16112

sdk/messaging/azeventhubs/internal/namespace_test.go
```diff
@@ -118,7 +118,7 @@ func TestNamespaceNegotiateClaimRenewal(t *testing.T) {
 	}
 
 	var errorsLogged []error
-	var nextRefreshDurationChecks atomic.Int32
+	var nextRefreshDurationChecks int32
```
richardpark-msft, line 121: Oh, I know this "downgrade" feeling. They added some nice stuff in atomics in 1.19.

jhendrixMSFT: Yeah I was a bit bummed about having to make this switch.

jhendrixMSFT:
@benbp enabling race detection has required certain tests to be excluded due to the tests failing with out-of-memory errors (this isn't surprising due to how the race detector works and the tests allocating large amounts of memory), and since we want race detection on by default, it means the impacted tests won't run in CI which isn't ideal.

What I'm thinking is that we add some extra flag in `ci.yml` files that indicates "please also run tests in the enumerated directories without race detection enabled".  Does that seem reasonable/doable?  Open to other ideas.

jhendrixMSFT:
Here's an example.

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2661578&view=logs&j=a4d06d16-879e-5ba3-ceb1-b266dbe8989a&t=35eb6bb5-29cf-53ca-cab6-34e1cf74a298

It's easiest to see this on the Windows runs, here's the relevant text (for Linux, the tests appear to just die without any info).
```
2023-03-28T19:15:13.8639015Z ==5880==ERROR: ThreadSanitizer failed to allocate 0x000274000000 (10536091648) bytes at 0x040477000000 (error code: 1455)
```

Error code 1455 means the [page file is too small to complete the operation](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1300-1699-).

The exclusions are at the test level by way of a build tag, see file `client_norace_test.go` in this PR.  So all tests in this file will be excluded from running with the race detector.

I think the easiest/safest thing to do would be to run tests in the specified directories without the race detector.  So in this case, we'd run the tests in `blockblob` twice.  While it does mean there will be some duplication, I think going through and annotating every other test file (and remembering to do so in the future) just makes things too complicated.

benbp:
@jhendrixMSFT Yeah the annotation is not ideal but I also don't think that doubling the test time is great either. Is this just the blob tests (e.g. they consume more memory because they upload/download large blobs and that affects the race detector) or is it more widespread?

jhendrixMSFT:
At present, it's just the handful of tests I moved to `client_norace_test.go` as they allocate large chunks of memory for large blobs.

PR #20482, opened 2023-03-28
rename profile module

tadelesh:
Since the previous module name collides with the semantic versioning, we decide to rename `v20200901` to `p20200901`.


tadelesh:
@jhendrixMSFT Could you help to review and approve?

jhendrixMSFT (approved):
Looks like the `ModuleName` constant's value doesn't match the name of the module.  Is that intentional?

tadelesh:
> Looks like the `ModuleName` constant's value doesn't match the name of the module. Is that intentional?

Good catch. I've consolidated it.

Issue #18073, opened 2022-05-17
Consolidate Key Vault test helper code


chlowell:
Key Vault modules contain duplicated test helper code that we could move into the `internal` module for easier maintenance.

RickWinter:
This a good one to improve the maintainability of KeyVault.  Charles can provide details.

PR #20500, opened 2023-03-29
[keyvault] update azsecrets to api version 7.4

gracewilcox:
- update autorest go version to 4.0.0-preview.46
- updated certificate api version to 7.4

### DIFF
--- a/sdk/security/keyvault/azsecrets/utils_test.go
+++ b/sdk/security/keyvault/azsecrets/utils_test.go
@@ -37,7 +37,9 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	vaultURL = strings.TrimSuffix(os.Getenv("AZURE_KEYVAULT_URL"), "/")
+	if recording.GetRecordMode() != recording.PlaybackMode {
+		vaultURL = strings.TrimSuffix(os.Getenv("AZURE_KEYVAULT_URL"), "/")
+	}
 	if vaultURL == "" {
 		if recording.GetRecordMode() != recording.PlaybackMode {
 			panic("no value for AZURE_KEYVAULT_URL")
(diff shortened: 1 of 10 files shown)


sdk/security/keyvault/azsecrets/utils_test.go
```diff
@@ -37,7 +37,9 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	vaultURL = strings.TrimSuffix(os.Getenv("AZURE_KEYVAULT_URL"), "/")
+	if recording.GetRecordMode() != recording.PlaybackMode {
+		vaultURL = strings.TrimSuffix(os.Getenv("AZURE_KEYVAULT_URL"), "/")
```
chlowell, line 41: Did this cause a problem in playback? Asking because at first glance it looks like it wouldn't make any difference, in which case I'd rather not have the conditional, if only to avoid making some future reader wonder about this 😄

gracewilcox: Yeah, it was causing an error on my end since I had the keyvault url set as an environment variable. Without the if statement, it'll pick up the real keyvault url instead of the fake one for the recordings. Guess we don't technically need it, but it's an easy issue to run into.

chlowell: I see, yes that would be a problem. You can tell I didn't let environment variables linger while I was working on this 🤭 

How about this:
```go
vaultURL = fakeVaultURL
if recording.GetRecordMode() != recording.PlaybackMode {
	if u, ok := os.LookupEnv("AZURE_KEYVAULT_URL"); ok {
		vaultURL = strings.TrimSuffix(u, "/")
	} else {
		panic("no value for AZURE_KEYVAULT_URL")
	}
}
```
(this reminds me of #18073)

chlowell (approved):

PR #20549, opened 2023-04-04
Client for Cognitive Services Anomaly Detector

mikekistler:
This PR is an experiment with a "fully generated" client from OpenAPI generated from TypeSpec.

The autorest.md (51 lines) and build.go (11 lines) were hand-written, as was the custom_client.go (104 lines) and version.go (11 lines), for a total of 177 lines of hand-written code. All the remaining code, 2438 lines, is completely generated.

About 30 lines of custom_client.go is implementing a KeyCredential auth policy, which won't be necessary once that's implemented in core.

I think the remainder of custom_client.go could probably be generated -- specifically for this service but also for most services with standard auth policies. This would also eliminate the need for most of the transforms currently implemented in autorest.md -- reducing the amount of hand-written code even further.

I've written a simple test of one method (listMultivariateModels) and it works as expected.


sdk/cognitiveservices/azanomalydetector/autorest.md
```diff
@@ -0,0 +1,51 @@
...
+input-file:
+- https://github.com/azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AnomalyDetector/stable/v1.1/openapi.json
+output-folder: ../azanomalydetector
+clear-output-folder: false
+module: github.com/Azure/azure-sdk-for-go/sdk/cognitiveservices/azanomalydetector
+license-header: MICROSOFT_MIT_NO_VERSION
+openapi-type: data-plane
+go: true
+use: "@autorest/go@4.0.0-preview.47"
+export-clients: true
```
jhendrixMSFT, line 15: `true` should now be the default.

sdk/cognitiveservices/azanomalydetector/custom_client.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Clients
+
+// UnivariateClient contains the methods for the Univariate group.
+// Don't use this type directly, use a constructor function instead.
+type UnivariateClient struct {
+	internal *azcore.Client
+	endpoint string
+	pl       runtime.Pipeline
```
jhendrixMSFT, line 58: The internal client contains a pipeline so this can be removed.

sdk/cognitiveservices/azanomalydetector/version.go
```diff
@@ -0,0 +1,11 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azanomalydetector
+
+const (
+	version = "v3.0.0-beta6"
```
jhendrixMSFT, line 10: Initial version would be `v0.1.0`

sdk/cognitiveservices/azanomalydetector/custom_client.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// UnivariateClientOptions contains optional settings for UnivariateClient.
+type UnivariateClientOptions struct {
+	azcore.ClientOptions
+}
+
+// NewUnivariateClient creates a client that accesses Azure Monitor metrics data.
+func NewUnivariateClient(endpoint string, credential KeyCredential, options *UnivariateClientOptions) (*UnivariateClient, error) {
```
jhendrixMSFT, line 67: I think we'd name this `NewUnivariateClientWithKeyCredential()` and save the "good name" for if/when the service supports AAD.

sdk/cognitiveservices/azanomalydetector/version.go
mikekistler: I made this consistent with the version used in all the other languages, but I'm not opposed to using a go-specific version.

sdk/cognitiveservices/azanomalydetector/custom_client.go
```diff
@@ -0,0 +1,104 @@
...
+
+// APIKeyPolicy authorizes requests with an API key acquired from a KeyCredential.
+type APIKeyPolicy struct {
+	header string
+	cred   KeyCredential
+}
+
+// NewAPIKeyPolicy creates a policy object that authorizes requests with an API Key.
+// cred: a KeyCredential implementation.
+func NewAPIKeyPolicy(cred KeyCredential, header string) *APIKeyPolicy {
```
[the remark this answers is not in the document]
mikekistler: True but this is only temporary until we can get this implemented in Azure core.

sdk/cognitiveservices/azanomalydetector/version.go
jhendrixMSFT: Having a major version greater than `1` activates [semantic import versioning](https://github.com/golang/go/wiki/Modules#semantic-import-versioning) and would require us to release the module as `v3`.
```
module github.com/Azure/azure-sdk-for-go/sdk/cognitiveservices/azanomalydetector/v3
```
I don't think we want to go down this path just to align with the other languages (that's been our position so far anyways).

JeffreyRichter (approved):
This looks great

mikekistler:
@jr-MS would you be interested in a Go client for MVAD?

PR #20550, opened 2023-04-04
Transaction Validation - Blob Client

siminsavani-msft:
Single test contains positive & negative check for user provided MD5 (CopyBlobFromURL has a parameter called SourceContentMD5)

### DIFF
--- a/sdk/storage/azblob/blob/client_test.go
+++ b/sdk/storage/azblob/blob/client_test.go
@@ -191,6 +191,67 @@ func waitForCopy(_require *require.Assertions, copyBlobClient *blockblob.Client,
 	}
 }
 
+func (s *BlobUnrecordedTestsSuite) TestCopyBlockBlobFromURL() {
+	_require := require.New(s.T())
+	testName := s.T().Name()
+	svcClient, err := testcommon.GetServiceClient(s.T(), testcommon.TestAccountDefault, nil)
+	if err != nil {
+		s.Fail("Unable to fetch service client because " + err.Error())
+	}
+
+	containerName := testcommon.GenerateContainerName(testName)
+	containerClient := testcommon.CreateNewContainer(context.Background(), _require, containerName, svcClient)
+	defer testcommon.DeleteContainer(context.Background(), _require, containerClient)
+
+	const contentSize = 8 * 1024 // 8 KB
+	content := make([]byte, contentSize)
+	contentMD5 := md5.Sum(content)
+	body := bytes.NewReader(content)
+
+	srcBlob := containerClient.NewBlockBlobClient("srcblob")
+	destBlob := containerClient.NewBlockBlobClient("destblob")
+
+	// Prepare source bbClient for copy.
+	_, err = srcBlob.Upload(context.Background(), streaming.NopCloser(body), nil)
+	_require.Nil(err)
+
+	expiryTime, err := time.Parse(time.UnixDate, "Fri Jun 11 20:00:00 UTC 2049")
+	_require.Nil(err)
+
+	credential, err := testcommon.GetGenericSharedKeyCredential(testcommon.TestAccountDefault)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


sdk/storage/azblob/blob/client_test.go
```diff
@@ -191,6 +191,67 @@ func waitForCopy(_require *require.Assertions, copyBlobClient *blockblob.Client,
 	}
 }
 
+func (s *BlobUnrecordedTestsSuite) TestCopyBlockBlobFromURL() {
```
seanmcc-msft, line 194: Is this a live-only test, or where is the test recording?

siminsavani-msft: CopyFromURL uses SAS to authenticate and currently there is no way to record SAS related tests.

vibhansa-msft (approved):

sdk/storage/azblob/blob/client_test.go
seanmcc-msft: Makes sense.  Do we have an issue open with Azure SDK to enable us to record SAS tests?

seanmcc-msft (approved):

sdk/storage/azblob/blob/client_test.go
seanmcc-msft, line 194:
```suggestion
func (s *BlobUnrecordedTestsSuite) TestCopyBlockBlobFromUrlSourceContentMD5() {
```

sdk/storage/azblob/blob/client_test.go
```diff
@@ -191,6 +191,67 @@ func waitForCopy(_require *require.Assertions, copyBlobClient *blockblob.Client,
...
+	_require.Nil(err)
+
+	srcBlobParts, _ := blob.ParseURL(srcBlob.URL())
+	srcBlobParts.SAS = sasQueryParams
+	srcBlobURLWithSAS := srcBlobParts.String()
+
+	// Invoke CopyFromURL.
+	sourceContentMD5 := contentMD5[:]
+	resp, err := destBlob.CopyFromURL(context.Background(), srcBlobURLWithSAS, &blob.CopyFromURLOptions{
+		SourceContentMD5: sourceContentMD5,
```
souravgupta-msft, line 242: Shouldn't we use the interface type `blob.SourceContentValidationType` here?
Is it that CopyFromURL() only supports `x-ms-source-content-md5` now, so we directly have the `[]byte` type for `SourceContentMD5`.
cc: @seanmcc-msft , @vibhansa-msft

seanmcc-msft: @souravgupta-msft, no, I don't think so.  Can you provide a link to this interface?

Because this is an MD5 on source data that the SDK never downloads, it is not possible for the SDK to calculate a checksum.  In addition, there are currently no plans to add CRC-64 support for source checksums in the storage service.

souravgupta-msft: This is the `SourceContentValidationType` interface.
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blob/constants.go#L202

It has two implementations, `SourceContentValidationTypeCRC64` and `SourceContentValidationTypeMD5`. We do not calculate checksum for source content for this interface.

seanmcc-msft: CRC64 is not supported by the service, and there are no plans to do so as far as I know.  I think we should leave this as a byte[].

siminsavani-msft: CopyFromURL only accepts `x-ms-source-content-md5` so using the SourceContentValidationType wouldn't work as it supports both CRC64 and MD5. (https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url?tabs=azure-ad#request-headers)

PR #20558, opened 2023-04-05
PutBlobFromURL/UploadBlobFromURL API 

siminsavani-msft:
Adding the PutBlobFromURL API, named UploadBlobFromURL.

Fresh PR for https://github.com/Azure/azure-sdk-for-go/pull/20016

### DIFF
--- a/sdk/storage/azblob/blockblob/client_test.go
+++ b/sdk/storage/azblob/blockblob/client_test.go
@@ -14,6 +14,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"hash/crc64"
 	"io"
 	"math/rand"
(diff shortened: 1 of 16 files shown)


sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -587,6 +588,624 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+}
+
+func (s *BlockBlobRecordedTestsSuite) TestPutBlobFromURLCopySourceAuth() {
+	_require := require.New(s.T())
+	testName := s.T().Name()
+	svcClient, err := testcommon.GetServiceClient(s.T(), testcommon.TestAccountDefault, nil)
+	_require.NoError(err)
+
+	// Getting AAD Authentication
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
souravgupta-msft, line 1172: The recording for this test is failing because,

1. There is no environment variable **AZURE_TENANT_ID** in the recording. This environment variable is used by the `azidentity.NewDefaultAzureCredential()` method to return the token credential. You can try a workaround like [this](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/testcommon/clients_auth.go#L148).
2. `cred.GetToken()` also returns different token strings which is passed in the `x-ms-copy-source-authorization` request header. So, the test fails in playback mode since it is unable to find the recording for the given CopySourceAuthorization header. Maybe something like [this](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/testcommon/common.go#L185) can help here.

sdk/storage/azblob/internal/testcommon/clients_auth.go
```diff
@@ -57,6 +57,12 @@ const (
 	FakeStorageURL     = "https://fakestorage.blob.core.windows.net"
 )
 
+var (
+	CoolAccessTier    = blob.AccessTierCool
```
vibhansa-msft, line 61: Do we need to redefine here again for testing? we could have used the predefined types itself.

siminsavani-msft: Updated! Was having issues with this, but I've made it a local variable which works fine.

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -587,6 +587,662 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+	destBBClient := testcommon.CreateNewBlockBlob(context.Background(), _require, dest, containerClient)
+
+	// Upload some data to source
+	contentSize := 4 * 1024 // 4KB
+	r, sourceData := testcommon.GetRandomDataAndReader(contentSize)
+	_, err := srcBBClient.Upload(context.Background(), streaming.NopCloser(r), nil)
+	_require.Nil(err)
+
+	// Create SAS for source and get SAS URL
+	expiryTime, err := time.Parse(time.UnixDate, "Fri Jun 11 20:00:00 UTC 2049")
```
souravgupta-msft, line 607: Any reason why the expiry time is so long. Since, this is an unrecorded test, we can have the expiry time as some small value like,
`expiryTime := time.Now().UTC().Add(15 * time.Minute)`

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -587,6 +587,662 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+	_require.Nil(err)
+
+	credential, err := testcommon.GetGenericSharedKeyCredential(testcommon.TestAccountDefault)
+	if err != nil {
```
souravgupta-msft, line 611: This can be used instead.
```suggestion
	_require.NoError(err)
```

sdk/storage/azblob/blockblob/client_test.go
siminsavani-msft: Good point, fixed!

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -587,6 +587,674 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+	_require.Nil(err)
+
+	srcBlobParts, _ := blob.ParseURL(srcBBClient.URL())
+	srcBlobParts.SAS = sasQueryParams
+	srcBlobURLWithSAS := srcBlobParts.String()
+
+	return containerClient, srcBBClient, destBBClient, srcBlobURLWithSAS, sourceData
+}
+
+func (s *BlockBlobUnrecordedTestsSuite) TestPutBlobFromURL() {
```
gapra-msft, line 628: should these all be recorded?

siminsavani-msft: PutBlobFromURL uses SAS to authenticate and at the moment, there is no way to record SAS related tests.

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -587,6 +587,674 @@ func (s *BlockBlobRecordedTestsSuite) TestUploadBlockWithImmutabilityPolicy() {
...
+	svcClient, err := testcommon.GetServiceClient(s.T(), testcommon.TestAccountDefault, nil)
+	_require.NoError(err)
+
+	containerClient, _, destBlob, _, _ := setUpPutBlobFromURLTest(testName, _require, svcClient)
+	defer testcommon.DeleteContainer(context.Background(), _require, containerClient)
+
+	// Create Blob with CPK
+	bbName := testcommon.GenerateBlobName(testName)
+	srcBlob := testcommon.CreateNewBlockBlobWithCPK(context.Background(), _require, bbName, containerClient, nil, &encryptionScope)
+	expiryTime, err := time.Parse(time.UnixDate, "Fri Jun 11 20:00:00 UTC 2049")
```
souravgupta-msft, line 738: Same as previous for this as well as the error check after `testcommon.GetGenericSharedKeyCredential`.

siminsavani-msft: Updated!

sdk/storage/azblob/blockblob/client_test.go
gapra-msft: SAS as source or destination? I think we can use regex to remove the last bit of the URL

sdk/storage/azblob/blockblob/models.go
```diff
@@ -70,6 +70,56 @@ func (o *UploadOptions) format() (*generated.BlockBlobClientUploadOptions, *gene
...
+
+	// Optional, default is true. Indicates if properties from the source blob should be copied.
+	CopySourceBlobProperties *bool
+
+	// Optional. Specifies a user-defined name-value pair associated with the blob.
+	Metadata map[string]*string
+
+	// Optional. Specifies the md5 calculated for the range of bytes that must be read from the copy source.
+	SourceContentMD5 []byte
+
```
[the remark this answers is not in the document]
seanmcc-msft: We don't want to include `TransactionalContentMD5`.  This is the MD5 for the Put Blob From URL Request itself, which the SDK is generating.  The customer would have no way to calculate the hash of the request.

sdk/storage/azblob/blockblob/client_test.go
gapra-msft: this is non blocking, we can resolve this now.

gapra-msft (approved):

PR #20564, opened 2023-04-05
[azservicebus] Removing code that handled session/link cancellation

richardpark-msft:
This removes the special cancellation handling we added after the latest go-amqp update. 

The problems mostly stem from how many thorny cases pop up when you try to handle this outside of the protocol stack. The first pass at this actually resulted in too many connection resets, which made the common use-case for AcceptNextSession() hazardous as every cancellation there would result in the connection resetting, and all existing sessions losing their link.

Even in non-session cases it could result in unnecessary interrupts for Receivers.

While looking at this I also found some bugs in our cleanup code for RPCLink. I fixed those and optimized it a bit to just close the session, rather than waiting and attempting to close each link individually.

### DIFF
--- a/sdk/messaging/azservicebus/internal/amqpLinks.go
+++ b/sdk/messaging/azservicebus/internal/amqpLinks.go
@@ -31,8 +31,6 @@ type RetryWithLinksFn func(ctx context.Context, lwid *LinksWithID, args *utils.R
 // stub things out for tests.
 type contextWithTimeoutFn func(parent context.Context, timeout time.Duration) (context.Context, context.CancelFunc)
 
-const defaultCloseTimeout = time.Minute
-
 type AMQPLinks interface {
 	EntityPath() string
 	ManagementPath() string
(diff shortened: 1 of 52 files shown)


sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -176,12 +173,7 @@ func (links *AMQPLinksImpl) recoverLink(ctx context.Context, theirLinkRevision L
 	}
 
 	// best effort close, the connection these were built on is gone.
-	if err := links.closeWithoutLocking(ctx, false); err != nil {
-		if GetRecoveryKind(err) == RecoveryKindConn {
-			return err
-		}
-	}
-
+	_ = links.closeWithoutLocking(ctx, false)
```
jhendrixMSFT, line 176: Does `closeWithoutLocking` need to return an error?  Looks like it's ignored in other places too.

richardpark-msft: _need_ is probably a good question but I think it's best to return relevant info and then let the caller sort it out. (just my personal preference)

richardpark-msft:
The one test that was failing is related to the go-amqp behavior of leaving the connection in an unstable state after cancelling creation of a session (in this case, it's the RPC link creation that's getting cancelled). 

This'll go away after we upgrade to the new go-amqp.

jhendrixMSFT (approved):

PR #20566, opened 2023-04-06
[azeventhubs] Removing code that handled session/link cancellation

richardpark-msft:
This removes the special cancellation handling we added after the latest go-amqp update.

The problems mostly stem from how many thorny cases pop up when you try to handle this outside of the protocol stack. go-amqp is going to be changed to handle most of these cases internally, by making sure we don't re-use handles when we're in ambiguous states and waiting eventual DETACH/END calls.

Companion to #20564

### DIFF
--- a/sdk/messaging/azeventhubs/internal/rpc.go
+++ b/sdk/messaging/azeventhubs/internal/rpc.go
@@ -35,9 +35,8 @@ type (
 		sessionID     *string
 		id            string
 
-		responseMu              sync.Mutex
-		startResponseRouterOnce *sync.Once
-		responseRouterClosed    chan struct{}
+		responseMu           sync.Mutex
+		responseRouterClosed chan struct{}
 
 		responseMap      map[string]chan rpcResponse
 		rpcLinkCtx       context.Context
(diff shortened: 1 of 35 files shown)


sdk/messaging/azeventhubs/internal/rpc.go
```diff
@@ -101,7 +91,7 @@ func NewRPCLink(ctx context.Context, args RPCLinkArgs) (*rpcLink, error) {
 
 	linkID, err := uuid.New()
 	if err != nil {
-		closeOrLog("session", session)
+		_ = session.Close(ctx)
```
jhendrixMSFT, line 94: `Close()` could return an "interesting" error, e.g. if the parent `Conn` died during closing.  Maybe you want to defer handling that until the next time `Conn` is used?  At any rate, just pointing it out.

richardpark-msft: My thinking was along the lines of your second point - that it'd just be more descriptive the next time we attempt to use the Conn, so we didn't have to get into too complicated an error return strategy here.

jhendrixMSFT (approved):

jhendrixMSFT (approved):

PR #20615, opened 2023-04-12
OAuth based tests for blob batch

souravgupta-msft:
Uncommenting the OAuth based tests for blob batch


sdk/storage/azblob/container/client_test.go
```diff
@@ -2371,28 +2371,13 @@ func (s *ContainerUnrecordedTestsSuite) TestContainerBlobBatchSetTierPartialFail
...
-	}
-	clientSecret, ok := os.LookupEnv("AZURE_STORAGE_CLIENT_SECRET")
-	if !ok {
-		panic("AZURE_STORAGE_CLIENT_SECRET could not be found")
-	}
+	accountName, _ := testcommon.GetGenericAccountInfo(testcommon.TestAccountDefault)
+	_require.Greater(len(accountName), 0)
 
-	cred, err := azidentity.NewClientSecretCredential(tenantID, clientID, clientSecret, nil)
+	cred, err := testcommon.GetGenericTokenCredential()
```
gapra-msft, line 2380: are we able to record these now with GetGenericTokenCred?

siminsavani-msft (approved):

tasherif-msft (approved):

sdk/storage/azblob/container/client_test.go
souravgupta-msft: All tests for blob batch are unrecorded currently whether we use shared key or OAuth credential. This is because in blob batch a random uuid is passed in the [Content-Type](https://learn.microsoft.com/en-us/rest/api/storageservices/blob-batch#request-headers) request header. For example, multipart/mixed; boundary=batch_**a81786c8-e301-4e42-a729-a32ca24ae252**.
This uuid is also passed in the request body as a batch boundary. I did try to add sanitizers for both the header and the body. But it didn't seem to resolve this. So, I have these tests as unrecorded.

Issue #20651, opened 2023-04-18
data/azcosmos: query value will cause error


chiyoi:
### Bug Report

- import path of package in question: `github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos`
- SDK version: `azcosmos@v0.3.3`
- output of `go version`: `go version go1.20.2 darwin/amd64`

- What happened?

When querying values like `select value c.id from c`, cosmos endpoint will return array of values (for example strings), but the SDK tried to unmarshal it into `map[string]any`, causing:
```
unmarshalling type *azcosmos.queryServiceResponse: json: cannot unmarshal string into Go struct field queryServiceResponse.Documents of type map[string]interface {}
```

- What did you expect or want to happen?

`cosmos_query_response.go: line 61`: 
change ...`[]map[string]interface{}`...
into ...`[]any`... or ...`[]interface{}`...

- How can we reproduce it?

```go
func reproduce() {
	client, _ := ...// ...: get a container client
	pager := client.NewQueryItemsPager("select value c.id from c", ..., nil) // ...: choose the right partition key
	for pager.More() {
		_, err := pager.NextPage(context.TODO())
		fmt.Println(err) // unmarshalling type *azcosmos.queryServiceResponse: json: cannot unmarshal string into Go struct field queryServiceResponse.Documents of type map[string]interface {}
	}
}
```

- Anything we should know about your environment.

/

ealsur:
@chiyoi please see https://github.com/Azure/azure-sdk-for-go/pull/20656

PR #20656, opened 2023-04-18
Cosmos DB: Fixes value queries

ealsur:
Value queries like `select value c.id from c` fail to be executed with the error:

```
unmarshalling type *azcosmos.queryServiceResponse: json: cannot unmarshal string into Go struct field queryServiceResponse.Documents of type map[string]interface {}
```

- [X] Updates to [CHANGELOG.md][] are included.

Closes https://github.com/Azure/azure-sdk-for-go/issues/20651

### DIFF
--- a/sdk/data/azcosmos/cosmos_query_response.go
+++ b/sdk/data/azcosmos/cosmos_query_response.go
@@ -58,7 +58,7 @@ func newQueryResponse(resp *http.Response) (QueryItemsResponse, error) {
 }
 
 type queryServiceResponse struct {
-	Documents []map[string]interface{} `json:"Documents,omitempty"`
+	Documents []any `json:"Documents,omitempty"`
 }
 
 // QueryContainersResponse contains response from the container query operation.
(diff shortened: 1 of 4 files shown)


sdk/data/azcosmos/cosmos_query_response.go
```diff
@@ -58,7 +58,7 @@ func newQueryResponse(resp *http.Response) (QueryItemsResponse, error) {
 }
 
 type queryServiceResponse struct {
-	Documents []map[string]interface{} `json:"Documents,omitempty"`
+	Documents []any `json:"Documents,omitempty"`
```
jhendrixMSFT, line 61: In later versions of codegen, we've changed exposing raw formats like this to just `[]byte` so that callers can unmarshal the JSON into a structure of their own.  I don't know if that makes sense here, just pointing it out.

Also, since this is a breaking change, be sure to note that in the changelog.

ealsur: This is an internal type though, not public, how is this a breaking change ?

jhendrixMSFT: Oh sorry I thought this was a return type exposed to callers.

jhendrixMSFT (approved):

PR #20658, opened 2023-04-18
Fix issue where tags header is passed if map is empty

gapra-msft:
HNS accounts do not currently support tags. If a customer passes nil OR and empty map, we should not set the x-ms-tags header.

### DIFF
--- a/sdk/storage/azblob/internal/shared/shared.go
+++ b/sdk/storage/azblob/internal/shared/shared.go
@@ -130,9 +130,6 @@ func ParseConnectionString(connectionString string) (ParsedConnectionString, err
 
 // SerializeBlobTags converts tags to generated.BlobTags
 func SerializeBlobTags(tagsMap map[string]string) *generated.BlobTags {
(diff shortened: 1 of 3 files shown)


sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -133,6 +133,9 @@ func SerializeBlobTags(tagsMap map[string]string) *generated.BlobTags {
 	if tagsMap == nil {
 		return nil
 	}
+	if len(tagsMap) == 0 {
```
jhendrixMSFT, line 136: Nit: with this change you can remove the above `nil` check as a `nil` map has a length of zero.

jhendrixMSFT (approved):

sdk/storage/azblob/internal/shared/shared.go
gapra-msft: fixed

souravgupta-msft (approved):

PR #20674, opened 2023-04-20
Rename DisableInstanceDiscovery

chlowell:
Architects requested a new name that calls out the option's security implications.

### DIFF
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Features Added
 
 ### Breaking Changes
+> These changes affect only code written against a beta version such as v1.3.0-beta.5
+* Renamed `DisableInstanceDiscovery` to `DisableAuthorityValidationAndInstanceDiscovery`
 
 ### Bugs Fixed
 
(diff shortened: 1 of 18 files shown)


sdk/azidentity/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Features Added
 
 ### Breaking Changes
+> These changes affect only code written against a beta version such as v1.3.0-beta.5
+* Renamed `DisableInstanceDiscovery` to `DisableAuthorityValidationAndInstanceDiscovery`
```
jhendrixMSFT, line 9: The length of the name hurts my head, but I don't think we can shorten it in any meaningful way.

jhendrixMSFT: Well, what if we named it `DisableAuthorityValidation` or do we lose some important meaning?

jhendrixMSFT (approved):

sdk/azidentity/CHANGELOG.md
chlowell: I suppose it could confuse someone wanting to disable instance discovery who doesn't know authority validation implies instance discovery. One goal of this new name is to make that relationship more explicit. It is a daunting name though 😞

PR #20696, opened 2023-04-25
AzFile File Client Samples

tasherif-msft:


sdk/storage/azfile/file/examples_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+func generateData(sizeInBytes int) (io.ReadSeekCloser, []byte) {
+	data := make([]byte, sizeInBytes)
+	_len := len(random64BString)
+	if sizeInBytes > _len {
+		count := sizeInBytes / _len
+		if sizeInBytes%_len != 0 {
+			count = count + 1
+		}
+		copy(data[:], strings.Repeat(random64BString, count))
```
gapra-msft, line 40: instead of repeating a static string for each test can we generate random data based on a seed (that is generated from the test name)? That way we are testing different data for each test. There should be an example in blobs of this.

gapra-msft: But I guess these are just samples, maybe it doesnt matter?

tasherif-msft: it samples, so hardcoded size is fine, I can however make that change for tests

gapra-msft (approved):

sdk/storage/azfile/file/examples_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	dirClient := shareClient.NewDirectoryClient("testDir")
+	fmt.Println(dirClient.URL())
+
+	fileClient := dirClient.NewFileClient("testFile")
+	fmt.Println(fileClient.URL())
+
+}
+
+func Example_directory_NewClientFromConnectionString() {
```
souravgupta-msft, line 75:
```suggestion
func Example_file_NewClientFromConnectionString() {
```

sdk/storage/azfile/file/examples_test.go
```diff
@@ -0,0 +1,281 @@
...
+	rangeList2, err := fileClient.GetRangeList(context.Background(), nil)
+	handleError(err)
+
+	fmt.Println(rangeList2.Ranges, 0)
+	_, err = fileClient.Delete(context.Background(), nil)
+	handleError(err)
+
+	_, err = shareClient.Delete(context.Background(), nil)
+	handleError(err)
+}
```
souravgupta-msft, line 281: I see that the examples are missing for these methods. Can you please add those.
DownloadStream, DownloadBuffer, DownloadFile
UploadFile
GetSASURL
UploadRangeFromURL
Resize, SetHTTPHeaders
StartCopyFromURL, AbortCopy

PR #20699, opened 2023-04-25
Remove ClientOptions from clients

tasherif-msft:


sdk/storage/azblob/appendblob/client.go
```diff
@@ -15,26 +15,22 @@ import (
...
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/base"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/exported"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/generated"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared"
 )
 
-// ClientOptions contains the optional parameters when creating a Client.
```
gapra-msft, line 25: We can't break APIs, so as Jeff suggested we should turn all the other ClientOptions structures into type aliases to the 1 ClientOptions structure. And I think Sourav mentioned we should be able to resolve the cyclical import by defining it in the internal/base package.

gapra-msft (approved):

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #20301, opened 2023-02-28
Add supporting features to enable distributed tracing

jhendrixMSFT:
This includes new internal pipeline policies and other supporting types. See the changelog for a full description.
Added some missing doc comments.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/7358

PR #20708, opened 2023-04-27
Restore supporting features to enable distributed tracing

jhendrixMSFT:
For the next `azcore` beta.
Previously reviewed in https://github.com/Azure/azure-sdk-for-go/pull/20301

### DIFF
--- a/sdk/azcore/runtime/policy_http_trace.go
+++ b/sdk/azcore/runtime/policy_http_trace.go
@@ -1,0 +1,117 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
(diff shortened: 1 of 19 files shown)


sdk/azcore/runtime/policy_http_trace.go
```diff
@@ -0,0 +1,117 @@
...
+
+	attrAZClientReqID  = "az.client_request_id"
+	attrAZServiceReqID = "az.service_request_id"
+
+	attrNetPeerName = "net.peer.name"
+)
+
+// newHTTPTracePolicy creates a new instance of the httpTracePolicy.
+//   - allowedQueryParams contains the user-specified query parameters that don't need to be redacted from the trace
+func newHTTPTracePolicy(allowedQueryParams []string) exported.Policy {
```
chlowell, line 35: What's the rationale for not exporting this?

jhendrixMSFT: Since the policy is always added to the pipeline and is a no-op if tracing isn't enabled, there's no need to export it for a custom pipeline.

This highlights a bit of a design flaw.  Originally, we let callers create an empty pipeline so they could add their own policies piecemeal; this required policy constructors to be exported.  We later removed the ability to create an empty pipeline but forgot to unexport all of the policy constructors.

I think the rule should be that if a policy is always added to the pipeline, there's no need to export its constructor.  IFF the policy is not in the pipeline by default should its constructor be exported (seems unlikely that we'd create such a policy, but I suppose it's possible).

@JeffreyRichter what are your thoughts on this?

JeffreyRichter:
Sure, I can agree to this guideline.

-- Jeffrey

PR #20710, opened 2023-04-27
[azservicebus, azeventhubs] Stress test and logging improvement

richardpark-msft:
azeventhubs and azservicebus consistently uses very little CPU and memory, so we can reduce the resources for our tests to be a bit more considerate to other users in the stress cluster.

In addition, the azservicebus package internal logging has been improved - mostly just trimming out some of the repetitive logging and making sure that we have a consistent prefix with link IDs.

### DIFF
--- a/sdk/messaging/azservicebus/internal/stress/tests/finite_peeks.go
+++ b/sdk/messaging/azservicebus/internal/stress/tests/finite_peeks.go
@@ -28,11 +28,13 @@ func FinitePeeks(remainingArgs []string) {
 	sender, err := client.NewSender(queueName, nil)
 	sc.PanicOnError("failed to create sender", err)
 
+	log.Printf("Sending a single message")
 	err = sender.SendMessage(sc.Context, &azservicebus.Message{
 		Body: []byte("peekable message"),
 	}, nil)
 	sc.PanicOnError("failed to send message", err)
 
+	log.Printf("Closing sender")
 	_ = sender.Close(sc.Context)
 
 	receiver, err := client.NewReceiverForQueue(queueName, nil)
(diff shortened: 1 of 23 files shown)


sdk/messaging/azservicebus/internal/stress/tests/finite_peeks.go
```diff
@@ -51,9 +53,13 @@ func FinitePeeks(remainingArgs []string) {
 	sc.PanicOnError("failed to abandon message",
 		receiver.AbandonMessage(sc.Context, tmp[0], nil))
 
-	for i := 0; i < 10000; i++ {
-		log.Printf("Sleeping for 1 second before iteration %d", i)
-		time.Sleep(time.Second)
+	maxPeeks := 10000
```
jhendrixMSFT, line 56: constants?

sdk/messaging/azservicebus/internal/utils/logger.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"sync/atomic"
+
+	azlog "github.com/Azure/azure-sdk-for-go/sdk/internal/log"
+)
+
+type Logger struct {
+	prefix *atomic.Value
```
jhendrixMSFT, line 14: Is this actually used concurrently?

richardpark-msft: It's possible, yes.  This logger is used in all locations where the Receiver (or Sender) does any opeations.

For instance, the user can do a parallel message completion like what we do in send_receive_drain.go stress test.

sdk/messaging/azservicebus/internal/stress/tests/finite_peeks.go
richardpark-msft: I don't know why I didn't do that already...

sdk/messaging/azservicebus/internal/utils/logger.go
jhendrixMSFT: Is this for tests only?

richardpark-msft: It's for production code.

jhendrixMSFT: Understood per offline discussion.

jhendrixMSFT (approved):

PR #20749, opened 2023-05-05
Fixing upload from reader

### LINKED ISSUE #20735: azblob BlockBlobClient.UploadFile fails with broken connection errors for certain files
### Bug Report

- importing `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob`
- using `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.0.0` in `go.mod`
- `go version go1.20.3 linux/amd64`

- What happened?

BlockBlobClient fails uploads only for certain files, I believe based on their length. I've managed to capture an example file from our production environment and have isolated some minimal code that can reproduce it. The `Progress` function reports that the full size of the file is transferred before an error is returned. 

The errors come in 2 flavors -- one that reports a `ContentLength` of ~9 exabytes:

```
Put "https://MyStorageAccount.blob.core.windows.net/MyStorageContainer/MyTestFile.txt?blockid=NjMwYWZiMTItMGNkMi00NmI0LTU1MWItYjhmZjIxZGIxZDk0&comp=block": net/http: HTTP/1.x transport connection broken: http: ContentLength=9223372036456316927 with Body length 0
```

and another that reports a smaller `ContentLength`:

```
Put "https://MyStorageAccount.blob.core.windows.net/MyStorageContainer/MyTestFile.txt?blockid=ODcyNDBkZGItOGRhOC00YzYyLTc0YzItNDk5ZmVmZDdkNGJk&comp=block": net/http: HTTP/1.x transport connection broken: http: ContentLength=4194304 with Body length 4193321
```

I expect the file to upload to blob storage successfully

- How can we reproduce it?

Generate a file with an unfortunate size -- for the case I'm examining, the file size is `398457897`. You can do a `fallocate -l 398457897 testfile.txt` if you'd like.

Here's some (messy) code that reproduces the issue reliably for me (you'll want to update the constants as appropriate for your testing): 

```
package main

import (
	"context"
	fmt "fmt"
	"log"
	"os"

	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
)

const (
	filepath = "/home/ben.east/workspace/testfile.txt"
	storageAccountName = "MyStorageAccount"
	storageContainerName = "MyStorageContainer"
)

func Upload(ctx context.Context, bucket string, key string, source string) (error) {
	if err != nil {
		return fmt.Errorf("unable to get credential: %w", err)
	}

	storageAccountURL := fmt.Sprintf("https://%s.blob.core.windows.net/", storageAccountName)
	client, err := azblob.NewClient(storageAccountURL, credential, nil)
	if err != nil {
		return fmt.Errorf("failed to get storage account client: %s. %w", storageAccountName, err)
	}

	containerClient := client.ServiceClient().NewContainerClient(bucket)
	blobClient := containerClient.NewBlockBlobClient(key)

	sourceFile, err := os.Open(source)
	if err != nil {
		return fmt.Errorf("unable to open file: %w", err)
	}
	defer sourceFile.Close()

	uploadOptions := &azblob.UploadFileOptions{
		Progress: func(bytesTransferred int64) {
			log.Printf("transferred bytes %d", bytesTransferred)
		},
	}
	_, err = blobClient.UploadFile(ctx, sourceFile, uploadOptions)
	if err != nil {
		return fmt.Errorf("fail to upload blob: %w", err)
	}
	return nil
}

func main() {
	err := Upload(context.Background(), storageContainerName, "test/testfile.txt", filepath)
	if err != nil {
		log.Fatalf("failed to upload: %v", err)
	}
	log.Println("succeeded")
}
```

execute `go run main.go` and you should see the error

This seems to occur only for certain files, and I believe it's related to the size of the file. I've been able to reproduce with a real artifact grabbed from our production environment, and again with a generated file of the same size.

souravgupta-msft:
Fixes #20735 


sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -4560,6 +4560,41 @@ func (s *BlockBlobUnrecordedTestsSuite) TestLargeBlockBufferedUploadInParallel()
 	_require.Equal(*(committed[1].Size), largeBlockSize)
 }
 
+func (s *BlockBlobUnrecordedTestsSuite) TestUploadFromReader() {
+	_require := require.New(s.T())
```
gapra-msft, line 4564: Are there any other content sizes we need to test here?

gapra-msft (approved):

sdk/storage/azblob/blockblob/client_test.go
souravgupta-msft: I can think of some content sizes and some tests for that.

souravgupta-msft: Added tests for few more content sizes.

PR #20754, opened 2023-05-05
[azeventhubs,azservicebus] Some API cleanup, renames

richardpark-msft:
Some changes for GA:
- CheckpointStore is a struct with func() fields instead of an interface. Just generally safer for expansion.
- Renamed CheckpointStore.UpdateCheckpoint() to SetCheckpoint() since it's more of a data layer and it reflects what it does (it creates/updates).
- Changed ReceivedEventData.Offset from a pointer to a int64 field. It's not optional when we receive the data.

Also, removed the vendored go-amqp and using the GA version.


sdk/messaging/azeventhubs/checkpoints/blob_store.go
```diff
@@ -36,9 +37,16 @@ type BlobStoreOptions struct {
 // NewBlobStore creates a checkpoint store that stores ownership and checkpoints in
 // Azure Blob storage.
 // NOTE: the container must exist before the checkpoint store can be used.
-func NewBlobStore(containerClient *container.Client, options *BlobStoreOptions) (*BlobStore, error) {
-	return &BlobStore{
+func NewBlobStore(containerClient *container.Client, options *BlobStoreOptions) (*azeventhubs.CheckpointStore, error) {
```
jhendrixMSFT, line 40: Does this have to return a `CheckpointStore` type?  I thought that abstraction was only for inputs.

richardpark-msft: It has to since we need to pass the result of this into the Processor() when it's created (and that type is of `CheckpointStore`.

jhendrixMSFT: I see.  It just seems weird that the name of the func is `NewBlobStore()` but it returns something other than a `BlobStore` (I get why).

Does this mean then that the `BlobStore` type doesn't need to be exported?

richardpark-msft: You're very correct, I will unexport the `BlobStore` type.

sdk/messaging/azeventhubs/checkpoint_store.go
```diff
@@ -11,19 +11,19 @@ import (
 )
 
 // CheckpointStore is used by multiple consumers to coordinate progress and ownership for partitions.
-type CheckpointStore interface {
+type CheckpointStore struct {
```
jhendrixMSFT, line 14: Thinking out loud, given the below conversation around `NewBlobStore()`, I'm reconsidering if this change is worth it (I didn't know that we return types that satisfy this interface).  With this change, it means that the `BlobStore` type can't expose any APIs other than what `CheckPointStore` exposes, since we can't return an actual `BlobStore` instance; it essentially tightly couples `BlobStore` (and other future types) to the `CheckpointStore` type which seems unfortunate.

CC @JeffreyRichter to weigh in on this as well.

richardpark-msft: It's a valid concern, for sure.

I think we can ignore this, however, because the majority of the type this is just an opaque type, so we wouldn't offer additional functionality on the `CheckpointStore` implementors. 

I could see us thinking about extending the overall CheckpointStore definition at some point, however, so having it be a struct with func()'s instead of an interface should cover what I hope is the actual common extension point.

richardpark-msft: Reverted!

jhendrixMSFT (approved):

Issue #20718, opened 2023-04-30
NewClientFromConnectionString doesn't support SAS tokens with custom endpoints


jlaundry:
### Bug Report

- SDK: `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.0.0 h1:u/LLAOFgsMv7HmNL4Qufg58y+qElGOt5qv0z1mURkRY=`
- `go version go1.20.2 darwin/amd64`

### Reproduction Steps

1. Start Azurite
2. Create a SAS token (example below uses the default Azurite access key)
3. Try to create a client with NewClientFromConnectionString:

```go
connectionString := "AccountName=devstoreaccount1;SharedAccessSignature=sig=PTTuImbY50brWEx%2FEBwY9czhUpSYzi5IGVg6utbiD2A%3D&sp=raw&sr=c&sv=2020-02-10;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"
containerName := "89d52d33-166d-4e69-8c0b-39e5c1e7940f"
client, err := azblob.NewClientFromConnectionString(connectionString, nil)
content := []byte("hello")
_, err = client.UploadBuffer(context.TODO(), containerName, "test.txt", content, nil)
```

### Expected

Client uploads a file to `http://127.0.0.1:10000/devstoreaccount1/89d52d33-166d-4e69-8c0b-39e5c1e7940f/test.txt`

### Actual

Client errors, as the upload was attempted to `https://devstoreaccount1.blob.core.windows.net/89d52d33-166d-4e69-8c0b-39e5c1e7940f/test.txt`

siminsavani-msft:
Hi @jlaundry ! Could you please share the client errors you are getting?

jlaundry:
Hi @siminsavani-msft - the 'error' is `connection string is either blank or malformed. The expected connection string should contain key value pairs separated by semicolons. For example 'DefaultEndpointsProtocol=https;AccountName=<accountName>;AccountKey=<accountKey>;EndpointSuffix=core.windows.net'`

But this error message occurs **after** the client has tried to connect to `devstoreaccount1.blob.core.windows.net` instead of `127.0.0.1:10000`...

I've had a chance to look at this more, the issue is

https://github.com/Azure/azure-sdk-for-go/blob/8849196dfa482cef8345590dc7bd12c8ad3c5de1/sdk/storage/azblob/internal/shared/shared.go#L101-L103

[...]

PR #20757, opened 2023-05-05
AzDatalake APIView

tasherif-msft:
SAS APIs will be in separate PR

In this pr you'll notice categories of (in the comments):
dfs1 -> meaning its an HNS op using a dfs endpoint
dfs/blob2 -> meaning its an HNS op using a blob endpoint
blob3 -> meaning its purely a blob endpoint op (applies to flat and HNS) and uses the underlying blob client

This PR also includes:
- API Design for path, service, dir, file and fs clients
- Lease API Design + Implementation (most of it not all)
- Generator upgrade 
- Generated Code fixes with transforms


sdk/storage/azdatalake/common.go
```diff
@@ -0,0 +1,81 @@
...
+const SnapshotTimeFormat = "2006-01-02T15:04:05.0000000Z07:00"
+
+// FilesystemAccessConditions identifies container-specific access conditions which you optionally set.
+type FilesystemAccessConditions struct {
+	ModifiedAccessConditions *ModifiedAccessConditions
+	LeaseAccessConditions    *LeaseAccessConditions
+}
+
+// FormatContainerAccessConditions formats FilesystemAccessConditions into container's LeaseAccessConditions and ModifiedAccessConditions.
+func FormatContainerAccessConditions(b *FilesystemAccessConditions) *container.AccessConditions {
```
jhendrixMSFT, line 24: Can these be moved to an internal package so they don't need to be exported?  I assume these kinds of functions aren't intended to be directly called by customers.

tasherif-msft: Makes sense, moved.

sdk/storage/azdatalake/internal/shared/challenge_policy.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"errors"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"net/http"
+	"strings"
+)
+
+type storageAuthorizer struct {
```
gapra-msft, line 18: Do we have a challenge policy in other SDKs too? This was added to blob specifically for managed disks, so I dont think it is relevant to HNS accounts. Thoughts?

tasherif-msft: I assumed if it was under shared in blob it applied to the other packages. Haven't looked closely at this yet. Can be removed if this is the case.

sdk/storage/azdatalake/file/client.go
```diff
@@ -0,0 +1,99 @@
...
+	return RenameResponse{}, nil
+}
+
+// SetExpiry operation sets an expiry time on an existing file.
+func (f *Client) SetExpiry(ctx context.Context, expiryType ExpiryType, o *SetExpiryOptions) (SetExpiryResponse, error) {
+	return SetExpiryResponse{}, nil
+}
+
+// Upload uploads data to a file.
+func (f *Client) Upload(ctx context.Context) {
```
gapra-msft, line 82: Shouldn't these have some inputs and outputs?

sdk/storage/azdatalake/filesystem/constants.go
```diff
@@ -0,0 +1,17 @@
...
+
+package filesystem
+
+import "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+
+// PublicAccessType defines values for AccessType - private (default) or file or filesystem.
+type PublicAccessType = azblob.PublicAccessType
+
+const (
+	File       PublicAccessType = azblob.PublicAccessTypeBlob
```
gapra-msft, line 15: should this be Path?

gapra-msft (approved):

sdk/storage/azdatalake/common.go
```diff
@@ -0,0 +1,31 @@
...
+
+package azdatalake
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/internal/generated"
+)
+
+// ClientOptions contains the optional parameters when creating a Client.
+type ClientOptions struct {
```
souravgupta-msft, line 15: Can this be moved to `internal/base` package. Just to be uniform with azblob and azfile.

sdk/storage/azdatalake/directory/client.go
```diff
@@ -0,0 +1,73 @@
...
+//   - options - client options; pass nil to accept the default values
+func NewClientWithNoCredential(serviceURL string, options *azdatalake.ClientOptions) (*Client, error) {
+	return nil, nil
+}
+
+// NewClientWithSharedKeyCredential creates an instance of Client with the specified values.
+//   - serviceURL - the URL of the storage account e.g. https://<account>.dfs.core.windows.net/
+//   - cred - a SharedKeyCredential created with the matching storage account and access key
+//   - options - client options; pass nil to accept the default values
+func NewClientWithSharedKeyCredential(serviceURL string, cred *exported.SharedKeyCredential, options *azdatalake.ClientOptions) (*Client, error) {
```
souravgupta-msft, line 42: You should do typedef of `exported.SharedKeyCredential` in the directory package (same applies to other packages).
For the `ClientOptions` also you should do typedef in my opinion.

sdk/storage/azdatalake/filesystem/constants.go
```diff
@@ -0,0 +1,17 @@
...
+	Filesystem PublicAccessType = azblob.PublicAccessTypeContainer
```
souravgupta-msft, line 16:
```suggestion
	FileSystem PublicAccessType = azblob.PublicAccessTypeContainer
```

sdk/storage/azdatalake/internal/exported/version.go
```diff
@@ -0,0 +1,12 @@
...
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package exported
+
+const (
+	ModuleName    = "azdatalake"
+	ModuleVersion = "v0.0.0"
```
souravgupta-msft, line 11:
```suggestion
	ModuleVersion = "v0.1.0"
```

sdk/storage/azdatalake/file/client.go
tasherif-msft: In the APIView I mentioned that this will be designed when implemeneted, just left it empty for now.

sdk/storage/azdatalake/filesystem/constants.go
tasherif-msft: It should be File, not path its basically anonymous access to data in a blob/file

sdk/storage/azdatalake/filesystem/constants.go
tasherif-msft: Should be Filesystem https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/filesystem

sdk/storage/azdatalake/directory/client.go
tasherif-msft: ClientOptions should be in the common since it does not cause any circular deps here.

sdk/storage/azdatalake/internal/shared/shared.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	for _, stringPart := range splitString {
+		parts := strings.SplitN(stringPart, "=", 2)
+		if len(parts) != 2 {
+			return ParsedConnectionString{}, errConnectionString
+		}
+		connStrMap[parts[0]] = parts[1]
+	}
+
+	accountName, ok := connStrMap["AccountName"]
```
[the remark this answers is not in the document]
souravgupta-msft: This was the issue fyi, https://github.com/Azure/azure-sdk-for-go/issues/20718

souravgupta-msft (approved):

PR #20759, opened 2023-05-06
Fix #20718

### LINKED ISSUE #20718: NewClientFromConnectionString doesn't support SAS tokens with custom endpoints

jlaundry:
As per #20718, this fixes support for Blob and Queue's NewClientFromConnectionString where the connection string uses both a SAS token and (Blob|Queue)Endpoint parameters.

I've also created `azqueue/internal/shared/shared_test.go`, using the azblob equivalent as a baseline


sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -98,9 +98,18 @@ func ParseConnectionString(connectionString string) (ParsedConnectionString, err
 		if !ok {
 			return ParsedConnectionString{}, errors.New("connection string missing AccountKey and SharedAccessSignature")
 		}
-		return ParsedConnectionString{
-			ServiceURL: fmt.Sprintf("%v://%v.blob.%v/?%v", defaultScheme, accountName, defaultSuffix, sharedAccessSignature),
-		}, nil
+
+		blobEndpoint, ok := connStrMap["BlobEndpoint"]
```
souravgupta-msft, line 102: I created the connection string from SAS using the portal. Its format is like this,

`BlobEndpoint=https://xxxxxxxx.blob.core.windows.net/;QueueEndpoint=https://xxxxxxxxx.queue.core.windows.net/;FileEndpoint=https://xxxxxxxx.file.core.windows.net/;TableEndpoint=https://xxxxxxxx.table.core.windows.net/;SharedAccessSignature=xxxxxxxxxxxxxxxx`

There is no `AccountName` here. So, the above code for checking account name also needs a fix.
```
accountName, ok := connStrMap["AccountName"]
if !ok {
	return ParsedConnectionString{}, errors.New("connection string missing AccountName")
}
```

souravgupta-msft: If you want, we can update this logic and fix this in some other PR.
I have done a similar fix in `azfile` package.
https://github.com/Azure/azure-sdk-for-go/pull/20798

jlaundry: Hi @souravgupta-msft, you're right.

I've added further tests, and then refactored the ParseConnectionString because the logic was too confusing and duplicated.

jlaundry: Resolving thread as code is now outdated

siminsavani-msft (approved):
Thank you for your contribution!

tasherif-msft (approved):

jlaundry:
Hey @siminsavani-msft @tasherif-msft - is there anything left to do before this can be merged?

sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -113,19 +97,35 @@ func ParseConnectionString(connectionString string) (ParsedConnectionString, err
...
+	if has_accountName && has_accountKey {
 		return ParsedConnectionString{
-			ServiceURL:  blobEndpoint,
+			ServiceURL:  serviceURL,
 			AccountName: accountName,
 			AccountKey:  accountKey,
 		}, nil
+	} else if has_sharedAccessSignature {
+		return ParsedConnectionString{
+			ServiceURL: fmt.Sprintf("%v/?%v", serviceURL, sharedAccessSignature),
```
souravgupta-msft, line 123: Can you add a check above the return statement for checking whether the serviceURL has trailing slash. If not, then add it and update this statement.
```suggestion
			ServiceURL: fmt.Sprintf("%v?%v", serviceURL, sharedAccessSignature),
```

sdk/storage/azqueue/internal/shared/shared.go
```diff
@@ -104,19 +88,34 @@ func ParseConnectionString(connectionString string) (ParsedConnectionString, err
...
+	if has_accountName && has_accountKey {
 		return ParsedConnectionString{
-			ServiceURL:  queueEndpoint,
+			ServiceURL:  serviceURL,
 			AccountName: accountName,
 			AccountKey:  accountKey,
 		}, nil
+	} else if has_sharedAccessSignature {
+		return ParsedConnectionString{
+			ServiceURL: fmt.Sprintf("%v/?%v", serviceURL, sharedAccessSignature),
```
souravgupta-msft, line 114: Same as previous. Please make these two changes. Everything else looks good.

sdk/storage/azblob/internal/shared/shared.go
jlaundry: Done - had to modify test data to do this though...

sdk/storage/azqueue/internal/shared/shared.go
jlaundry: Done, same

souravgupta-msft (approved):

sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -113,19 +97,40 @@ func ParseConnectionString(connectionString string) (ParsedConnectionString, err
...
 
-	if blobEndpoint, ok := connStrMap["BlobEndpoint"]; ok {
+	blobEndpoint, has_blobEndpoint := connStrMap["BlobEndpoint"]
+	accountName, has_accountName := connStrMap["AccountName"]
+
+	var serviceURL string
+	if has_blobEndpoint {
+		serviceURL = blobEndpoint
+	} else if has_accountName {
+		serviceURL = fmt.Sprintf("%v://%v.blob.%v/", protocol, accountName, suffix)
```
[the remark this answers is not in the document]
jlaundry: I don't think so: appending a trailing / only happens after serviceURL is set, and only if !strings.HasSuffix(serviceURL, "/")...

Can you please share the failed test data, or share a unit test that recreates the issue?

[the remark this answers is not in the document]
jlaundry: Alternatively, we could go back to not appending a trailing /, which as far as I can tell, has no material impact?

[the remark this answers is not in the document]
souravgupta-msft: If the `suffix` in the above suggested change is **core.windows.net/** (having trailing slash), then the current code which is,
`serviceURL = fmt.Sprintf("%v://%v.blob.%v/", protocol, accountName, suffix)`
will create, `serviceURL="https://accountName.blob.core.windows.net//"` (two slashes at the end).

So, the trailing slash while creating the serviceURL can be removed as suggested in the change above.
Lines 112-114 anyways takes care of checking and adding the trailing slash.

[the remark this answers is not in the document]
jlaundry: Ahhhh, ok I understand now. Done, and added further unit tests.

siminsavani-msft (approved):

Issue #20747, opened 2023-05-05
Increase azcontianerregistry test coverage to 90%+


tadelesh:
One of the GA criteria.

PR #20762, opened 2023-05-06
increase acr test coverage over 90%

tadelesh:
Resolve: #20747

### DIFF
--- a/sdk/containers/azcontainerregistry/autorest.md
+++ b/sdk/containers/azcontainerregistry/autorest.md
@@ -457,3 +457,25 @@ directive:
     transform: >
         $.ArtifactOperatingSystem.description = "The artifact platform's operating system.";
 ```
+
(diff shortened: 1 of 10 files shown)


sdk/containers/azcontainerregistry/autorest.md
```diff
@@ -457,3 +457,25 @@ directive:
     transform: >
         $.ArtifactOperatingSystem.description = "The artifact platform's operating system.";
 ```
+
+### Remove useless Marshal method
```
jhendrixMSFT, line 461: Why aren't these needed?

tadelesh: These models will only be used in responses. Two ways to increase test coverage for the serialization part:
1. add useless marshal test
2. remove marshal function
And I choose second.

PR #20828, opened 2023-05-10
[azeventhubs] Fixing govcloud live tests

richardpark-msft:
We're now setting the `Cloud` in the client options when using armeventhubs. This only affected our tests.


richardpark-msft:
Merged a separate PR to handle the Cloud field. There's a persistent connectivity issue between our CI machines and the remote cloud that cause failures beyond our (pretty reasonable) retry policies. That's being investigated, but for now our tests are at least correct.

PR #20843, opened 2023-05-11
Retry policy will always clone the *http.Request

jhendrixMSFT:
This ensures that the entirety of the request is restored on retries.

### DIFF
--- a/sdk/azcore/runtime/policy_retry_test.go
+++ b/sdk/azcore/runtime/policy_retry_test.go
@@ -810,6 +810,47 @@ func TestRetryableRequestBodyWithCloser(t *testing.T) {
 	require.True(t, tr.closeCalled)
 }
 
+func TestRetryPolicySuccessWithRetryPreserveHeaders(t *testing.T) {
+	srv, close := mock.NewServer()
+	defer close()
+	srv.AppendResponse(mock.WithStatusCode(http.StatusRequestTimeout))
+	srv.AppendResponse()
+	pl := exported.NewPipeline(srv, NewRetryPolicy(testRetryOptions()), exported.PolicyFunc(challengeLikePolicy))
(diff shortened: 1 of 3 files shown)


sdk/azcore/runtime/policy_retry_test.go
```diff
@@ -810,6 +810,47 @@ func TestRetryableRequestBodyWithCloser(t *testing.T) {
...
+	if req.Body() == nil {
+		return nil, errors.New("request body wasn't restored")
+	}
+	if req.Raw().Header.Get("content-type") != "text/plain" {
+		return nil, errors.New("content-type header wasn't restored")
+	}
+
+	if body := req.Body(); body != nil {
+		rb := reqBody{body, req.Raw().Header.Get("content-type")}
+		req.SetOperationValue(rb)
```
chlowell, line 846: My first thought was, the actual policy does this but it isn't necessary for this test. Now I'm wondering whether the OperationValue feature needs additional testing. The actual policy's logic is:
1. set: `req.SetOperationValue()`
2. send: `req.Next()`
3. get: `req.OperationValue()`

It appears every retry runs on its own clone of `req`, and a policy never gets `req` itself. So step 1 sets the value on an ephemeral clone; if there's a retry during step 2, is that value lost?

chlowell: I see. Then I'm back to wondering why this test sets the value 😆. Does it have a side effect?

chlowell (approved):

sdk/azcore/runtime/policy_retry_test.go
jhendrixMSFT: No, it's just copy/pasta from the key vault challenge policy.  I'll clean this up a bit.

jhendrixMSFT: Done.

Issue #20823, opened 2023-05-10
Resource provider not registred automatically for `Microsoft.Quota/quotas`


ms-henglu:
With track2 sdk, creating of a Microsoft.Quota/quotas@2021-03-15-preview resource is failing to a newly created subscription without a registered Microsoft.Quota/quotas provider. 

```bash
GET https://management.azure.com/subscriptions/{subscription_id}/providers/Microsoft.Compute/locations/northeurope/providers/Microsoft.Quota/quotas
│ --------------------------------------------------------------------------------
│ RESPONSE 400: 400 Bad Request
│ ERROR CODE: MissingRegistrationForResourceProvider
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "MissingRegistrationForResourceProvider",
│     "message": "The subscription is not in registered state for the resource provider: Microsoft.Quota."
│   }
│ }
│ --------------------------------------------------------------------------------
```

This might be caused by the fact that the automatic resource registration is checking for the error code `MissingSubscriptionRegistration` but the API is returning `MissingRegistrationForResourceProvider`

https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/arm/runtime/policy_register_rp.go#L83

Issue #19900, opened 2023-01-31
Resource provider not registred automatically for `Microsoft.Security/pricings`


ms-henglu:
With track2 sdk, creating of a Microsoft.Security/pricings@2022-03-01 resource is failing to a newly created subscription without a registered Microsoft.Security provider. 

```bash
│ PUT https://management.azure.com/subscriptions/[subscription-id]/providers/Microsoft.Security/pricings/VirtualMachines?api-version=2022-03-01
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE: Subscription Not Registered
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "Subscription Not Registered",
│     "message": "Please register to Microsoft.Security in order to view your security status"
│   }
│ }
│ --------------------------------------------------------------------------------
│ 
│ 
╵
```

This might be caused by the fact that the automatic resource registration is checking for the error code `MissingSubscriptionRegistration` but the API is returning `Subscription Not Registered`

https://github.com/Azure/azure-sdk-for-go/blob/b386adc15c2e60e3812956f68d244b608330c479/sdk/azcore/arm/runtime/policy_register_rp.go#L83

tadelesh:
@eliagrady I wonder if the error is returned from ARM or your service which is not consistent with the [guide](https://github.com/Azure/azure-resource-manager-rpc/blob/13e38b290b7bdde85285139271143bb876ec0c14/v1.0/subscription-lifecycle-api-reference.md#subscription-states). Could you help to take a look?

PR #20848, opened 2023-05-12
Handle more error codes when an RP isn't registered

jhendrixMSFT:
There's more than one error code returned from unregistered RPs.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/20823

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Bugs Fixed
 * Retry policy always clones the underlying `*http.Request` before invoking the next policy.
+* Added `MissingRegistrationForResourceProvider` to the list of error codes for unregistered resource providers.
 
 ### Other Changes
 
--- a/sdk/azcore/arm/runtime/pipeline_test.go
+++ b/sdk/azcore/arm/runtime/pipeline_test.go
@@ -104,7 +104,7 @@ func TestDisableAutoRPRegistration(t *testing.T) {
 	srv, close := mock.NewServer()
 	defer close()
 	// initial response that RP is unregistered
-	srv.SetResponse(mock.WithStatusCode(http.StatusConflict), mock.WithBody([]byte(rpUnregisteredResp)))
+	srv.SetResponse(mock.WithStatusCode(http.StatusConflict), mock.WithBody([]byte(rpUnregisteredResp1)))
 	opts := &armpolicy.ClientOptions{DisableRPRegistration: true, ClientOptions: policy.ClientOptions{Transport: srv}}
 	req, err := azruntime.NewRequest(context.Background(), http.MethodGet, srv.URL())
 	if err != nil {
--- a/sdk/azcore/arm/runtime/policy_register_rp.go
+++ b/sdk/azcore/arm/runtime/policy_register_rp.go
@@ -80,7 +80,6 @@ func (r *rpRegistrationPolicy) Do(req *azpolicy.Request) (*http.Response, error)
 		// policy is disabled
 		return req.Next()
 	}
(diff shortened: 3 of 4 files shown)


jhendrixMSFT:
We could also add a string slice to [`RegistrationOptions`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/policy/policy.go#L29) to allow callers to add additional non-standard error codes to unblock themselves.  Thoughts?

CC @JeffreyRichter

JeffreyRichter:
I'm sure I'm missing context but when I look at #20823, it looks like the error code string was just wrong and that this bug should be fixed.

I can see the slight desire to look for error code strings (which should be compared case-sensitively, by the way - I mention this because I saw a call to EqualFold in #20823 but it looks that was may have been removed). But, since error codes are related to HTTP status code, the thing to check on would be BOTH status code AND error code string.
But, do we really have to add this additional infrastructure?

jhendrixMSFT:
We do already check the HTTP status code and if it's a 409 then we look for this additional information.

Most services return `MissingSubscriptionRegistration` (I believe _all services_ are supposed to use this error code when the RP is unregistered).  However, the Quota service is returning `MissingRegistrationForResourceProvider` and I'm sure we can't change that at this point for compat reasons (I have an email out to Roopesh and you about this).

Adding a slice of additional error codes would let customers unblock themselves if/when a service returns a non-standard error code.  That said, it's probably rare enough that the extra machinery isn't worth it (I don't feel strongly about it, just throwing the idea out there).

JeffreyRichter:
Who owns the code that is doing this comparison?
If it's us (the SDK), then we can check for both strings.
If it's the customer, then they can check for the right (or both) strings.

JeffreyRichter:
OK, so if you want to check for both strings, that's OK with me. Add a comment.

chlowell:
> Adding a slice of additional error codes would let customers unblock themselves if/when a service returns a non-standard error code. That said, it's probably rare enough that the extra machinery isn't worth it (I don't feel strongly about it, just throwing the idea out there).

👍 if there's one standard code, it's reasonable to assume the number of variants is small; this may be the only one. We can patch again if another variant turns up and add API if we come to believe there are many unknown variants.

chlowell (approved):

ms-henglu:
Hi @jhendrixMSFT ,

Thank you for fixing the issue, and I have another similar case: https://github.com/Azure/azure-sdk-for-go/issues/19900, would you please also check it? Thanks!

The error is 
```json
 {
   "error": {
     "code": "Subscription Not Registered",
     "message": "Please register to Microsoft.Security in order to view your security status"
   }
 }
```

ms-henglu:
@jhendrixMSFT - Sure, it's already in this issue: https://github.com/Azure/azure-sdk-for-go/issues/19900

With track2 sdk, creating of a Microsoft.Security/pricings@2022-03-01 resource is failing to a newly created subscription without a registered Microsoft.Security provider. 

```bash
│ PUT https://management.azure.com/subscriptions/[subscription-id]/providers/Microsoft.Security/pricings/VirtualMachines?api-version=2022-03-01
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE: Subscription Not Registered
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "Subscription Not Registered",
│     "message": "Please register to Microsoft.Security in order to view your security status"
│   }
│ }
│ --------------------------------------------------------------------------------
│ 
│ 
╵
```

jhendrixMSFT:
My mistake, I confused this with issue https://github.com/Azure/azure-sdk-for-go/issues/20823

PR #20852, opened 2023-05-14
Cosmos DB: add support for getting partition key ranges for a container

juruen:
This PR adds support for getting partition key ranges for a container. It's based on the [REST API documentation](https://learn.microsoft.com/en-us/rest/api/cosmos-db/get-partition-key-ranges).

We are planning to use this feature as part of our work to have more insights on the state of the partitions from the client application.

I've tried to follow the the existing style for marshalling and unmarshalling the responses.

Thank you! ❤️ 

### DIFF
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -562,6 +562,34 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
 	return newTransactionalBatchResponse(azResponse)
 }
 
+// PartitionKeyRanges obtains the information for the partition key ranges in a Cosmos container
+// ctx - The context for the request.
+func (c *ContainerClient) PartitionKeyRanges(ctx context.Context) (PartitionKeyRangesResponse, error) {
+	operationContext := pipelineRequestOptions{
+		resourceType:    resourceTypePartitionKeyRange,
+		resourceAddress: c.link,
+	}
+
+	path, err := generatePathForNameBased(resourceTypeCollection, c.link, false)
+	if err != nil {
+		return PartitionKeyRangesResponse{}, err
+	}
+
+	path += "/pkranges"
+
+	azResponse, err := c.database.client.sendGetRequest(
+		path,
+		ctx,
+		operationContext,
+		nil,
+		nil)
+	if err != nil {
+		return PartitionKeyRangesResponse{}, err
+	}
+
+	return newPartitionKeyRangeResponse(azResponse)
+}
+
 func (c *ContainerClient) getRID(ctx context.Context) (string, error) {
 	containerResponse, err := c.Read(ctx, nil)
 	if err != nil {
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -575,7 +575,7 @@ func (c *ContainerClient) PartitionKeyRanges(ctx context.Context) (PartitionKeyR
 		return PartitionKeyRangesResponse{}, err
 	}
 
-	path += "/pkranges"
+	path += "/" + pathSegmentPartitionKeyRange
 
 	azResponse, err := c.database.client.sendGetRequest(
 		path,
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+* Added support for getting partition key ranges for a container
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/data/azcosmos/cosmos_container_test.go
+++ b/sdk/data/azcosmos/cosmos_container_test.go
@@ -649,3 +649,106 @@ func TestContainerPatchItem(t *testing.T) {
 		t.Errorf("Expected url to be %s, but got %s", "/dbs/databaseId/colls/containerId/docs/doc1", verifier.requests[0].url.RequestURI())
 	}
 }
+
+func TestContainerPartitionKeyRanges(t *testing.T) {
+	nowAsUnix := time.Unix(time.Now().Unix(), 0)
+
+	etag := azcore.ETag("etag")
+	properties := &PartitionKeyRangesProperties{
+		ResourceID: "someResourceId",
+		Count:      1,
+		PartitionKeyRanges: []PartitionKeyRange{
+			{
+				ID:           "someId",
+				ETag:         &etag,
+				SelfLink:     "someSelfLink",
+				LastModified: nowAsUnix,
+				MinInclusive: "someMinInclusive",
+				MaxExclusive: "someMaxExclusive",
+			},
+		},
+	}
+	jsonString, err := json.Marshal(properties)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	srv, close := mock.NewTLSServer()
+	defer close()
+	srv.SetResponse(
+		mock.WithBody(jsonString),
+		mock.WithHeader(cosmosHeaderEtag, "someEtag"),
+		mock.WithHeader(cosmosHeaderActivityId, "someActivityId"),
+		mock.WithStatusCode(200))
+
+	pl := azruntime.NewPipeline("azcosmostest", "v1.0.0", azruntime.PipelineOptions{}, &policy.ClientOptions{Transport: srv})
+	client := &Client{endpoint: srv.URL(), pipeline: pl}
+
+	database, _ := newDatabase("databaseId", client)
+	container, _ := newContainer("containerId", database)
+
+	if container.ID() != "containerId" {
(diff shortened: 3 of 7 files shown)


sdk/data/azcosmos/cosmos_container.go
```diff
@@ -562,6 +562,34 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
...
+		resourceType:    resourceTypePartitionKeyRange,
+		resourceAddress: c.link,
+	}
+
+	path, err := generatePathForNameBased(resourceTypeCollection, c.link, false)
+	if err != nil {
+		return PartitionKeyRangesResponse{}, err
+	}
+
+	path += "/" + pathSegmentPartitionKeyRange
```
juruen, line 578: Is it OK to build the path like this or I should an available function?

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -562,6 +562,34 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
...
+
+	azResponse, err := c.database.client.sendGetRequest(
+		path,
+		ctx,
+		operationContext,
+		nil,
```
juruen, line 584: I'm sure it's OK to set cosmosRequestOptions to `nil` here.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -562,6 +562,34 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
 	return newTransactionalBatchResponse(azResponse)
 }
 
+// PartitionKeyRanges obtains the information for the partition key ranges in a Cosmos container
+// ctx - The context for the request.
+func (c *ContainerClient) PartitionKeyRanges(ctx context.Context) (PartitionKeyRangesResponse, error) {
```
ealsur, line 567: @juruen This API is not even available in other Track2 SDKs and not available in the .NET SDKs by design. Partition level information can be obtained from the Azure Monitor metrics/APIs and there are no plans on the Go SDK to use any form of public type or API that let's users directly send operations to a Physical partition.

ealsur (changes requested):

sdk/data/azcosmos/cosmos_container.go
ealsur: Could you describe the scenario where knowing this information on a Data Plane SDK is required? None of the information contained in `PartitionKeyRangesResponse` can be used as input in any API. If the goal is to have more insights on the partitions, then using Azure Monitor or potentially the metrics in `armcosmos` which is a Management Plane SDK.

ealsur: I see, thanks for clarifying.

You are correct in your understanding of the usage of the hashing to route requests, but this is only for SDKs with Direct connectivity (.NET and Java). Those SDKs however use this all as internal routing information. And it's not the only way to route a request either. They don't make this public.

In Go, having this information (the PKRange Min/Max ranges) won't be useful as it's not nor will be an SDK with Direct support, there is no hashing algorithm. So it would be impossible to use those values, you'd have to take the request's PK value, apply the hashing and then use hashing compare to find the right partition, and this is all internal implementation that is subject to change (I mentioned before that this is not the only way to route a request).

I do understand what you are trying to achieve, basically a smarter client-side throttling by identifying which groups of data are causing it, but in that case, the 429 response in itself is basically saying that at least for that partition RUs are exhausted (maybe not on other partitions and those requests will flow fine).

kirankumarkolli: Unless the queue it-self is partitioned using the same hash algo as Cosmos, there is not much locality -> any PkRange usage counters needs to be global across nodes/processes at second granularity. 

Few options to be proactive about a PkRange throttling.
- As @ealsur already mentioned pro-active monitoring through Azure monitoring and scale appropriately
- Autoscale: ref: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-provision-autoscale-throughput?tabs=api-async
- Burst capacity: https://learn.microsoft.com/en-us/azure/cosmos-db/burst-capacity

juruen: > In Go, having this information (the PKRange Min/Max ranges) won't be useful as it's not nor will be an SDK with Direct support, there is no hashing algorithm. So it would be impossible to use those values, you'd have to take the request's PK value, apply the hashing and then use hashing compare to find the right partition, and this is all internal implementation that is subject to change (I mentioned before that this is not the only way to route a request). We would not have the hashing API public either if we ever decide to include it here, so there would be no way for you to use the PKRange information.

@ealsur, I totally understand why you wouldn't want to expose the PKRange information. I decided to give this PR a chance, basically because a)  the endpoint is part of the [public REST API](https://learn.microsoft.com/en-us/rest/api/cosmos-db/get-partition-key-ranges) and b) for convenience for us, to build a feature on top of it, not in the SDK itself.

Just for completeness, as I already had the operation to fetch pkranges working, I just took a stab at identifying the pkRange a given pk belongs to by looking at the Java SDK. Its  partition key hashing v2 implementation seems to work OK  with my cluster and it was trivial to port to Go. For simplicity,  getting the effective partition key for a string looks like something like this in Go:

```Go
func toEffectivePartitionKeyString(pk string) string {
	var buf bytes.Buffer
	buf.WriteByte(0x8) // String PK type
	buf.Write([]byte(pk))
	buf.WriteByte(0xFF)

	hash := murmur128(buf)
	hash = reverse(hash)
	hash[0] &= 0x3f
	return strings.ToUpper(hex.EncodeToString(hash))
}
```

by using  the result of that function, finding the pkRange it belongs to is just a matter of finding the range where this value is included.

This seems to work and is good enough for us to experiment with having a smarter throttling mechanism. As we can build it ourselves, there's no need for us to push for having any of this in the SDK.

> in that case, the 429 response in itself is basically saying that at least for that partition RUs are exhausted (maybe not on other partitions and those requests will flow fine). If 1 partition is hot, then all requests will get throttled and all of them will be for that partition (or subset of partitions). You don't really need to identify which partitions they are though.

It's not that using 429s doesn't provide information about a hot partition key(s), my point is that mapping the partition keys to partition range IDs provides richer information to act on, obviously, at the expense of relying on implementation details that are subject to change.

Let me illustrate this with an example:

Say we are ingesting a sequence of documents where their PKs have high cardinality. If we process this sequentially from a distributed queue, If there's a subset of these documents that end up exhausting a physical partition, relying on 429s works OK when this subset is small: a small bump in 429s and everything goes back to normal. However, if this subset is large, the impact on ingestion becomes a bit more critical and relying on 429s has its own issues, namely: you can't know if a given PK is hitting a hot partition until you actually try to write to it and get a 429. I But getting a 429 is not cheap, and it involves some retries, and when you have many different PKs that hit the hot partition, this problem compounds  and might have an impact on the overall latency.

Mapping PKs to pkRanges allows us to skip the _let's see if we get a 429_ step. Because when we  mark a pkRange as hot in our client, there's no need to try to write to Cosmos, we know the PKs that belong to the pkRange  and we can throttle or handle them differently. E.g: moving them into a slower queue and therefore, we can reduce the impact of hot partition has on the overall system.

There's an obvious trade-off here that we'll evaluate, is it worth adding this complexity to try to get better reliability?. We'll see! We'll experiment and make a decision based on that.

Going back to the SDK, I just want to say that I've been only working with it and with Cosmos DB for a couple of months now and am very happy with it. Kudos to team!

Thanks! ❤️

ealsur: > Mapping PKs to pkRanges allows us to skip the let's see if we get a 429 step. Because when we mark a pkRange as hot in our client, there's no need to try to write to Cosmos, we know the PKs that belong to the pkRange and we can throttle or handle them differently. E.g: moving them into a slower queue and therefore, we can reduce the impact that a hot partition has on the overall system.

This is a really nice idea and I understand why you'd want to approach it. It is correct that it is relying on internal implementations and things that might change (like with the new [burst capacity](https://learn.microsoft.com/en-us/azure/cosmos-db/burst-capacity) or [redistribution of RU across partitions](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/distribute-throughput-across-partitions)).

Thanks a lot for sharing the hashing implementation, while not valid for all containers it's great to have it handy.

Thanks also for your contribution, I feel bad for having to hold the PR because of the work you put into it, please don't feel discouraged to contribute in the future. Maybe one approach we can take next time is first open an Issue and discuss the idea, and we can see if it can ship.

juruen: > Thanks also for your contribution, I feel bad for having to hold the PR because of the work you put into it, please don't feel discouraged to contribute in the future

Don't feel bad at all! This is something that we are going to use internally to experiment, and it has also helped me to look into the SDK closely. All in all, it's been super useful!

If we see an opportunity to contribute in the future, I'll make sure I open an issue to discuss it first.

Cheers!

PR #20860, opened 2023-05-16
Add another non-standard error code for RP registration

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/19900

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 ### Bugs Fixed
 * Retry policy always clones the underlying `*http.Request` before invoking the next policy.
-* Added `MissingRegistrationForResourceProvider` to the list of error codes for unregistered resource providers.
+* Added some non-standard error codes to the list of error codes for unregistered resource providers.
 
 ### Other Changes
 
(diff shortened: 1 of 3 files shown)


chlowell (approved):

felipebbc:
Hi @jhendrixMSFT 

The issue #19900 with the Resource Provider registration has not been resolved by adding the non-standard error code `Subscription Not Registered`.

The sdk also checks if the response has a 409 error code:
https://github.com/Azure/azure-sdk-for-go/blob/6387cfc7d4421e4df032c64fc7fab4cf36f7c622/sdk/azcore/arm/runtime/policy_register_rp.go#L90-L92

The error message from the Microsoft.Security/pricings@2022-03-01 call returns a **404 error code**, so the provider registration doesn't move forward.

Also, the provider registration requires that the error message contain the provider name in the `error.details.target` field

https://github.com/Azure/azure-sdk-for-go/blob/6387cfc7d4421e4df032c64fc7fab4cf36f7c622/sdk/azcore/arm/runtime/policy_register_rp.go#L202-L207

The error message from the Microsoft.Security/pricings@2022-03-01 does not contain the `error.details.target` field, it only contains `error.code` and `error.message`:

```
│ PUT https://management.azure.com/subscriptions/[subscription-id]/providers/Microsoft.Security/pricings/VirtualMachines?api-version=2022-03-01
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE: Subscription Not Registered
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "Subscription Not Registered",
│     "message": "Please register to Microsoft.Security in order to view your security status"
│   }
│ }
│ -------------------------------------------------------------------------------- 
```

I tested out a solution adding the 404 code on line 94 and adding a regex to extract the resource provider name from the error.message. It works but I understand this is not a good solution. 

Looking for you to see how we can allow automatic resource provider registration in this scenario.

jhendrixMSFT:
@felipebbc could you please open a new issue for this so we can track it?

PR #20862, opened 2023-05-17
Samples for article for azeventhubs GA

richardpark-msft:
These are four samples I plan to publish as part of the Event Hubs GA blog post. I used only the standard library and azeventhubs+azidentity, with the idea that users could just copy/paste the code and build the utility themselves, customizing it as they see fit. No binaries are distributed.

If there are no obstacles I'd also like to check them into the monorepo as well, as I think they're useful for myself and they can be easily built and installed using `go install`.


sdk/messaging/azeventhubs/internal/tools/ehlagcheck/ehlagcheck.go
jhendrixMSFT, line 1: Maybe these files should go under `/sdk/samples/azeventhubs`?

jhendrixMSFT: I think the answer is yes.

richardpark-msft: Yeah, looking in there that looks like a perfect home for them. I'll move them.

richardpark-msft: (it also makes me feel less bad about using some other packages like `cobra`)

richardpark-msft: Moved, also seeing if I can resurrect the `go - samples` pipeline.

sdk/samples/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+      - main
+      - feature/*
+      - hotfix/*
+      - release/*
+  paths:
+    include:
+    - sdk/samples
+
+stages:
+- template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
```
jhendrixMSFT, line 25: I think this should be archetype-sdk-client-samples.yml

richardpark-msft: That looks like it might be better. I'd like to add in tests too but this is better than what's going on now.

sdk/samples/azeventhubs/ehlagcheck/internal/constants.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package internal
+
+// Version is the semantic version number
+const Version = "v0.0.1"
```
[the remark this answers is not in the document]
richardpark-msft: I can actually see this being useful if we incorporate it somewhere in the output. It doesn't get tagged but users could get some idea of what version of the source they copied.

richardpark-msft:
@benbp, this is the PR I had where I attempted to get the samples pipeline bootstrapped. I want to come back and hit this again, but if there's a pre-built archetype I should use let me know.

richardpark-msft:
You don't control me, bot.

PR #20866, opened 2023-05-17
Refinement of fake package per API review

jhendrixMSFT:
Added an HTTP status code parameter to to methods that return a response.  For methods that already had such a paraemter, it was relocated to be the first parameter.
Removed Responder[T].SetHeader method, replacing the functionality with an http.Header field in SetResponseOptions.
Cleaned up some wording in the server API doc comments.


jhendrixMSFT:
Needs more test coverage in the `server` package.

sdk/azcore/fake/fake.go
```diff
@@ -43,507 +31,110 @@ type TokenCredential struct {
...
-}
+type AddNonTerminalResponseOptions = exported.AddNonTerminalResponseOptions
 
 // SetTerminalResponseOptions contains the optional values for PollerResponder[T].SetTerminalResponse.
-type SetTerminalResponseOptions struct {
-	// place holder for future optional values
-}
-
-/////////////////////////////////////////////////////////////////////////////////////////////////////////////
-// the following APIs are intended for use by fake servers
```
jhendrixMSFT, line 186: These APIs were moved into the `server` package.

sdk/azcore/fake/fake.go
```diff
@@ -43,507 +31,110 @@ type TokenCredential struct {
...
-}
-
-// SetHeader sets the specified header key/value pairs to be returned.
-// Call multiple times to set multiple headers.
-func (r *Responder[T]) SetHeader(key, value string) {
-	if r.content.Header == nil {
-		r.content.Header = http.Header{}
-	}
-	r.content.Header.Set(key, value)
+func (r *Responder[T]) SetResponse(httpStatus int, resp T, o *SetResponseOptions) {
```
jhendrixMSFT, line 54: I have trepidations about the `httpStatus` parameter.  Given that `SetResponse()` is _always_ for the success case, we've now opened the possibility of the customer specifying the wrong HTTP status code, either the wrong 2xx status code or some other code that doesn't make sense for the success case.

jhendrixMSFT: Per discussion, we will leave this as-is.  To help push customers into the pit of success, we will codegen comments containing the list of successful HTTP status codes and return a descriptive error message when there's a mismatch.

JeffreyRichter (approved):

PR #20867, opened 2023-05-18
[azeventhubs] Changes to make errors and low level types carry connectionID and linkID

richardpark-msft:
Make it so the lower level primitives and errors carry more information about the connection ID, rather than requiring us to shuttle it around externally.

- Adding (AMQPReceiver|AMQPSender|AMQPSession).ConnID() fields (and AMQPClient.ID()) fields.
- Errors are now wrapped immediately when returned from the underlying `amqpwrap` types. This is prep for another issue I found where we'd fail to recover a connection in some cases because we didn't have enough context to do it.
- Moved out some universally useful helpers for gomock into the `test` package (basically, splitting up mock_helpers.go into two files).
- Regenerating mocks

### DIFF
--- a/sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
+++ b/sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
@@ -27,7 +27,7 @@ func (e Error) Is(target error) bool {
 	return errors.Is(e.Err, target)
 }
 
-func NewError(err error, connID uint64, linkName string) error {
+func WrapError(err error, connID uint64, linkName string) error {
 	if err == nil {
 		return nil
 	}
--- a/sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
+++ b/sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package amqpwrap
+
+import (
+	"errors"
+)
+
+// Error is a wrapper that has the context of which connection and
+// link the error happened with.
+type Error struct {
+	ConnID   uint64
+	LinkName string
+	Err      error
+}
+
+func (e Error) Error() string {
+	return e.Err.Error()
+}
+
+func (e Error) As(target any) bool {
+	return errors.As(e.Err, target)
+}
+
+func (e Error) Is(target error) bool {
+	return errors.Is(e.Err, target)
+}
+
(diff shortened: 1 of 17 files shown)


sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
```diff
@@ -0,0 +1,44 @@
...
+
+package amqpwrap
+
+import (
+	"errors"
+)
+
+// Error is a wrapper that has the context of which connection and
+// link the error happened with.
+type Error struct {
```
jhendrixMSFT, line 12: Is this only used internally, or will this be returned from APIs?

sdk/messaging/azeventhubs/internal/amqpwrap/amqpwrap.go
```diff
@@ -3,6 +3,7 @@
 
 // Package amqpwrap has some simple wrappers to make it easier to
 // abstract the go-amqp types.
+
```
jhendrixMSFT, line 6: If there's no space, then the above comment will be used as the doc comment for this package.  Is this intentional?

sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (e Error) As(target any) bool {
+	return errors.As(e.Err, target)
+}
+
+func (e Error) Is(target error) bool {
+	return errors.Is(e.Err, target)
+}
+
+func NewError(err error, connID uint64, linkName string) error {
```
jhendrixMSFT, line 30: I wonder if this would be better named something like `WrapError` since it might not do anything if `err` is `nil` ("New" makes me think it always creates something).  Relatively minor considering this is internal.

sdk/messaging/azeventhubs/internal/errors.go
```diff
@@ -257,94 +249,13 @@ func GetRecoveryKind(err error) RecoveryKind {
 	return RecoveryKindConn
 }
 
-type (
```
jhendrixMSFT, line 260: Was this just dead code, or was it replaced by something else?

jhendrixMSFT (approved):
Just a few questions.

sdk/messaging/azeventhubs/internal/amqpwrap/amqp_error.go
richardpark-msft: Internally only (and we're in an internal package).

PR #20871, opened 2023-05-18
azfile: upgrade to STG 87

souravgupta-msft:


sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -307,3 +312,14 @@ directive:
       return $.
         replace(/ShareUsageBytes\s+\*int32/g, `ShareUsageBytes *int64`);
 ```
+
+### Convert StringEncoded to string type
```
gapra-msft, line 316: this will not quite work fully to resolve that encoded bug, but we can fix this later I think. We will need the latest generator for that.

souravgupta-msft: I tried updating the code generator to _4.0.0-preview.49_ version. But there are some issues with it. I am in discussion with @jhendrixMSFT regarding that. You can check [this commit](https://github.com/souravgupta-msft/azure-sdk-for-go/commit/639dc2663d23e3037dcffe5f78e35b66048bd0cb#diff-1967425b657f4121ac3847ecc36f26075211a0f0a8d60b95bff8b8794e062f51) for reference.

souravgupta-msft: Also the fields where the `StringEncoded` field is used, are all response types. So, this transform in my opinion is safe.
The service also returns these fields as `string` type instead of what `StringEncoded` represents.

souravgupta-msft: I have updated the code generator and added the unmarshal methods as discussed.
Will add more tests for these cases.

gapra-msft (approved):

sdk/storage/azfile/directory/client.go
```diff
@@ -93,23 +104,23 @@ func (d *Client) URL() string {
...
+	return (*Client)(base.NewDirectoryClient(subDirectoryURL, d.generated().InternalClient(), d.sharedKey(), d.getClientOptions()))
 }
 
 // NewFileClient creates a new file.Client object by concatenating fileName to the end of this Client's URL.
 // The new file.Client uses the same request policy pipeline as the Client.
 func (d *Client) NewFileClient(fileName string) *file.Client {
 	fileName = url.PathEscape(fileName)
 	fileURL := runtime.JoinPaths(d.URL(), fileName)
-	return (*file.Client)(base.NewFileClient(fileURL, d.generated().Pipeline(), d.sharedKey()))
+	return (*file.Client)(base.NewFileClient(fileURL, d.generated().InternalClient(), d.sharedKey(), d.getClientOptions()))
```
jhendrixMSFT, line 115: The file client will end up with the wrong name using in distributed tracing (see line 52 above).

souravgupta-msft: Can we use the name "azfile.Client" instead of "directory.Client", "file.Client", etc. for all packages.
Creating new azcore client will create new pipelines. So, if file client is created from directory client, both will have different pipelines. Correct me if I am wrong here. 
Ideal scenario will be the child and parent clients sharing the same pipeline like it is currently with the old code generator.

souravgupta-msft: @jhendrixMSFT, can you confirm if the above change looks good.

jhendrixMSFT: IMO the client names need to be unique so that when looking at a trace it's clear what the spans apply to (especially if any clients have the same method names).  Including @lmolkova for comments.

As I recall, sharing the pipeline was to simplify creating the "child" client with respect to credentials.  Other than that there's no need to share them right?

jhendrixMSFT (changes requested):
Since an `*azcore.Client` contains instance-specific data, we cannot reuse them for "child" clients.

jhendrixMSFT (changes requested):
Since an `*azcore.Client` contains instance-specific data, we cannot reuse them for "child" clients.

jhendrixMSFT (changes requested):
Since an `*azcore.Client` contains instance-specific data, we cannot reuse them for "child" clients.

sdk/storage/azfile/directory/client.go
jhendrixMSFT: Per today's discussion, we can create a new `*azcore.Client` with the correct client name.

I've opened https://github.com/Azure/azure-sdk-for-go/issues/20973 to track optimizing this so we can continue to share pipelines across child clients.

souravgupta-msft: @jhendrixMSFT, I have made the changes. Can you please review it?

sdk/storage/azfile/internal/shared/shared.go
```diff
@@ -62,6 +62,13 @@ const (
 	FileAttributesDirectory = "Directory"
 )
 
+const (
+	ServiceClient   = "service.Client"
```
jhendrixMSFT, line 66: Can you please update to `azcore@v1.6.1` and then add the prefix `azfile/` to all of these constants?  At present, this will generate the wrong package name `service` in the telemetry string.

sdk/storage/azfile/internal/base/clients.go
```diff
@@ -8,19 +8,22 @@ package base
...
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/internal/exported"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/internal/generated"
 )
 
 // ClientOptions contains the optional parameters when creating a Client.
 type ClientOptions struct {
 	azcore.ClientOptions
+	AllowTrailingDot       *bool
+	FileRequestIntent      *generated.ShareTokenIntent
```
jhendrixMSFT, line 19: `generated.ShareTokenIntent` and its accompanying values will need to be exported as a type alias.

sdk/storage/azfile/directory/client.go
```diff
@@ -45,9 +50,12 @@ func NewClientWithSharedKeyCredential(directoryURL string, cred *SharedKeyCreden
 	authPolicy := exported.NewSharedKeyCredPolicy(cred)
 	conOptions := shared.GetClientOptions(options)
 	conOptions.PerRetryPolicies = append(conOptions.PerRetryPolicies, authPolicy)
```
jhendrixMSFT, line 52: This really should be.
```go
	plOpts := runtime.PipelineOptions{
		PerRetry: []policy.Policy{authPolicy},
	}
	azClient, err := azcore.NewClient(shared.DirectoryClient, exported.ModuleVersion, plOpts, &conOptions.ClientOptions)
```
It was missed in previous reviews.  I first noticed this in `azblob` earlier this week.  If you don't want to fix it now, can you please open a tracking issue?

sdk/storage/azfile/directory/client.go
jhendrixMSFT: Looks great!

sdk/storage/azfile/internal/shared/shared.go
souravgupta-msft: Done

sdk/storage/azfile/internal/base/clients.go
souravgupta-msft: Exported the `generated.ShareTokenIntent` type and its accompanying values.

sdk/storage/azfile/directory/client.go
souravgupta-msft: Added the auth policy in the `PerRetry` field of `runtime.PipelineOptions`.

souravgupta-msft: @jhendrixMSFT, I have resolved the other comments. Can you please review and let me know if there are more changes needed.

jhendrixMSFT (approved):

Issue #20870, opened 2023-05-18
Duplicate digest calculation when retrying


tadelesh:
The body will be read twice when retrying which will cause duplicate calculation for digest.

PR #20872, opened 2023-05-18
do not re-calculate digest when retry

tadelesh:
fix: https://github.com/Azure/azure-sdk-for-go/issues/20870


tadelesh:
@jhendrixMSFT could you help to review?

Issue #20836, opened 2023-05-11
Add digest validation to all related operations for azcontainerregistry


tadelesh:
1. check whether digest calculator could be applied to all related operation
2. add hero scenario sample code and prepare for service review

PR #20887, opened 2023-05-23
add digest validation utils and examples

tadelesh:
resolve: https://github.com/Azure/azure-sdk-for-go/issues/20836


tadelesh:
@jhendrixMSFT @JeffreyRichter please help to review the digest validation utils feature. You could refer all example files for the usage. A brief explanation:
1. `GetManifest`, `GetBlob`, `GetChunk` all use `DigestValidationReader` to read stream while validate.
2. if `UploadManifest` by tag, we could not know the sig alg before service respond, so user need to do the read twice with `DigestValidationReader` to validate digest.
3. `UploadChunk` keep the same with previous version, only some code tidy.

sdk/containers/azcontainerregistry/blob_client_example_test.go
```diff
@@ -47,21 +51,82 @@ func ExampleBlobClient_DeleteBlob() {
 }
 
 func ExampleBlobClient_GetBlob() {
-	res, err := blobClient.GetBlob(context.TODO(), "prod/bash", "sha256:16463e0c481e161aabb735437d30b3c9c7391c2747cc564bb927e843b73dcb39", nil)
+	digest := "sha256:16463e0c481e161aabb735437d30b3c9c7391c2747cc564bb927e843b73dcb39"
```
jhendrixMSFT, line 54: Nit, should be a `const`.

sdk/containers/azcontainerregistry/blob_client_example_test.go
```diff
@@ -47,21 +51,82 @@ func ExampleBlobClient_DeleteBlob() {
...
+	}
+	f, err := os.Create("blob_file")
+	if err != nil {
+		log.Fatalf("failed to create blob file: %v", err)
+	}
+	defer f.Close()
+	buf := make([]byte, 1024*1024)
+	for {
+		_, err := reader.Read(buf)
+		if err == io.EOF {
```
jhendrixMSFT, line 71: Use `errors.Is()` instead of comparison.

sdk/containers/azcontainerregistry/utils.go
jhendrixMSFT, line 1: I think we should give the file a more meaningful name, maybe `digest_validation.go` or something like that.

tadelesh: Since it contains both digest validator and digest calculator, I rename it to `digest_helper.go`.

tadelesh:
@JeffreyRichter The latest API view for `azcontainerregistry` could be found [here](https://apiview.dev/Assemblies/Review/518fb41eeb5447ea850c25d28cf9e7dc?diffRevisionId=450d4204342a4d1eae5c5f21b80111af&diffOnly=False&revisionId=0862316330cd4ef08a6b77fdb66b2bc4&doc=False). The change include removal of useless marshal method and new-added utils for digest validation. Do I need to have a former meeting review with you?

JeffreyRichter:
I had just 1 comment about hash algorithms

tadelesh:
> I had just 1 comment about hash algorithms

Thanks you the quick review. I've replied the comment. The hash algorithm is considered already.

jhendrixMSFT (approved):

sdk/containers/azcontainerregistry/example_download_image_test.go
```diff
@@ -0,0 +1,98 @@
...
+		log.Fatalf("failed to create validation reader: %v", err)
+	}
+	config, err := io.ReadAll(reader)
+	if err != nil {
+		log.Fatalf("failed to read config data: %v", err)
+	}
+	fmt.Printf("config: %s\n", config)
+
+	// Get layers
+	layers := manifestJSON["layers"].([]interface{})
```
jhendrixMSFT, line 74: Nit, replace `interface{}` with `any`.

sdk/containers/azcontainerregistry/blob_custom_client_example_test.go
jhendrixMSFT, line 1: Files that contain examples should have `example_` as the prefix.

sdk/containers/azcontainerregistry/digest_helper.go
```diff
@@ -0,0 +1,160 @@
...
+	ErrDigestAlgNotSupported = errors.New("digest algorithm not supported")
+)
+
+type digestValidator interface {
+	io.Writer
+	validate(digest string) error
+}
+
+func parseDigestValidator(digest string) (digestValidator, error) {
+	alg := digest[:strings.Index(digest, ":")]
```
shizhMSFT, line 31: There might be an "index out of range" issue if `digest` does not contain a `:`.

sdk/containers/azcontainerregistry/digest_helper.go
```diff
@@ -0,0 +1,160 @@
...
+	if v, ok := validatorCtors[alg]; ok {
+		return v(), nil
+	} else {
+		return nil, ErrDigestAlgNotSupported
+	}
```
shizhMSFT, line 36: In general, you may want to read https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow to remove the `else`.

shizhMSFT: It applies to all occurrences in this file.

sdk/containers/azcontainerregistry/example_blob_custom_client_test.go
```diff
@@ -42,9 +43,32 @@ func ExampleBlobClient_CompleteUpload() {
...
+	}
+	chunkSize := int64(5)
+	current := int64(0)
+	for {
+		end := current + chunkSize
+		if end > size {
+			end = size
+		}
+		chunkReader := io.NewSectionReader(f, current, end-current)
+		uploadResp, err := blobClient.UploadChunk(context.TODO(), location, chunkReader, calculator, &azcontainerregistry.BlobClientUploadChunkOptions{RangeStart: to.Ptr(int32(current)), RangeEnd: to.Ptr(int32(end - 1))})
```
shizhMSFT, line 63: Just curious if `calculator` can be reused in another process or not.

sdk/containers/azcontainerregistry/digest_helper.go
tadelesh: Will fix before GA.

sdk/containers/azcontainerregistry/digest_helper.go
tadelesh: Will fix before GA.

PR #20888, opened 2023-05-23
[azservicebus] Add support for `DefaultRule` when creating a Subscription

StrawbrryFlurry:
Hi there! c:

We've recently migrated from a [tool](https://docs.particular.net/transports/azure-service-bus/operational-scripting) that used the Azure SDK for dotnet to manage Service Bus Topics and Subscriptions to the Azure Resource Manager Terraform Provider. Since then, we're struggling with an issue that newly created subscriptions automatically consume all messages in their topic. It seems like there currently is no way to specify a default rule (filter) when creating a subscription with the go api client, which means this option is also not available in the Terraform Provider using this SDK. The dotnet SDK supports this option which is why we didn't have this issue before.

It looks like this option was considered to also be part of the go SDK, which is which is why the option was initially added, [but was left commented out for some reason at that time](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/admin/admin_client_subscription.go#L383).

To me it looks like whatever held back the implementation back then is now no longer an issue, which is why I was able to add the necessary changes to make it work. I've tried to keep my changes in the same style as the existing code as well as adding tests for the new feature. I'd love to get any feedback on the changes and am dediacted to work out any kinks, you might not be happy with! Additionally, I could also imagine not exposing a `RuleProperties` object through `SubscriptionProperties` but instead just offer a flag, `DoNotConsumeMessages`, which sets a `FalseFilter` instead of the `TrueFilter` in the default rule.

Changes:
- Add `DefaultRule` field to `SubscriptionProperties`
- Extract conversion functions from `createOrUpdateRule` to make them available when creating a new subscription
- Create a `DefaultRuleDescription` using `SubscriptionProperties`, when creating a  `SubscriptionEnvelope`
- Extend `DefaultRuleDescription` with `Action` and change `Filter` to be a pointer rather than a value to be consistent with the optional `Action`
- Add tests for `Client.CreateSubscription` with a `DefaultRule` in `SubscriptionProperties`

### DIFF
--- a/sdk/messaging/azservicebus/admin/admin_client_rules.go
+++ b/sdk/messaging/azservicebus/admin/admin_client_rules.go
@@ -208,7 +208,7 @@ func (ac *Client) GetRule(ctx context.Context, topicName string, subscriptionNam
 		return mapATOMError[GetRuleResponse](err)
 	}
 
-	props, err := ac.newRuleProperties(ruleEnv)
+	props, err := ac.newRulePropertiesFromEnvelope(ruleEnv)
 
 	if err != nil {
 		return nil, err
@@ -240,7 +240,7 @@ func (ac *Client) NewListRulesPager(topicName string, subscriptionName string, o
 	}
 
 	ep := &entityPager[atom.RuleFeed, atom.RuleEnvelope, RuleProperties]{
-		convertFn:    ac.newRuleProperties,
+		convertFn:    ac.newRulePropertiesFromEnvelope,
 		baseFragment: fmt.Sprintf("/%s/Subscriptions/%s/Rules/", topicName, subscriptionName),
 		maxPageSize:  pageSize,
 		em:           ac.em,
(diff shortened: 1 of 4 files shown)


StrawbrryFlurry:
@microsoft-github-policy-service agree company="Digitec Galaxus AG"

richardpark-msft (approved):
Looks really good, thank you for doing this!

The reason we left it out, initially, was because we didn't have working rule support but that did get added later for the other Rule related functions but I must have missed this one.

I'll add you to the changelog when I prep the release. 

Would you like me to credit you with your github alias or some other name?

StrawbrryFlurry:
Great! Thank you for the very positive feedback! 😊 If it's all the same to you, I would prefer the GitHub handle. 

Is there anything else you would need me to do - possibly to fix the test coverage that is failing the tests?

richardpark-msft:
> Great! Thank you for the very positive feedback! 😊 If it's all the same to you, I would prefer the GitHub handle.

Awesome, just wanted to make sure you got to choose, really appreciate that you did this work!

> Is there anything else you would need me to do - possibly to fix the test coverage that is failing the tests?

The test coverage is a bit of an ornery problem, I'll tackle that separately. I've created this PR(https://github.com/Azure/azure-sdk-for-go/pull/20902) if you want to merge with it which will lower the required coverage level.

> Also, could you possibly give me a rough estimate, when this release will be happening?

Yes, if all goes according to plan we'll release this on 2023-06-06.

sdk/messaging/azservicebus/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.3.1 (2023-06-06)
```
chlowell, line 3:
```suggestion
## 1.4.0 (2023-06-06)
```

PR #20892, opened 2023-05-23
[keyvault] azkeys renames

gracewilcox:
Renames for azkeys based on feedback

### DIFF
--- a/sdk/security/keyvault/azkeys/CHANGELOG.md
+++ b/sdk/security/keyvault/azkeys/CHANGELOG.md
@@ -25,7 +25,8 @@
 * `JSONWebKeySignatureAlgorithm` to `SignatureAlgorithm`
 * `JSONWebKeyType` to `KeyType`
 * Removed `DeletionRecoveryLevel` type
-* Deleted `SignatureAlgorithmRSNULL` constant
+* Removed `SignatureAlgorithmRSNULL` constant
+* Removed `KeyOperationExport` constant
 * Removed `MaxResults` option
 
 ### Other Changes
--- a/sdk/security/keyvault/azkeys/CHANGELOG.md
+++ b/sdk/security/keyvault/azkeys/CHANGELOG.md
@@ -1,6 +1,31 @@
 # Release History
 
-## 1.0.0 (2023-05-11)
+## 1.0.0 (unreleased)
+
+### Breaking Changes
+* Renamed `GetRandomBytesRequest` to `GetRandomBytesParameters`
+* `ListDeletedKey` to `ListDeletedKeyProperties`
+* `ListKeys` to `ListKeyProperties`
+* `DeletedKeyBundle` to `DeletedKey`
+* `KeyBundle` to `KeyVaultKey`
+* `RestoreKeyParameters.KeyBundleBackup` to `RestoreKeyParameters.KeyBackup`
+* `DeletedKeyItem` to `DeletedKeyProperties`
+* `KeyItem` to `KeyProperties`
+* `DeletedKeyListResult` to `DeletedKeyPropertiesListResult`
+* `KeyListResult` `KeyPropertiesListResult`
+* `KeyOperationsParameters` to `KeyOperationParameters`
+* `KeyOps` field to `KeyOperations`
+* Changed `JSONWebKey.KeyOperations` from type []*string to []*KeyOperation
+* `KeyEncryptionAlgorithm` to `KeyExportEncryptionAlgorithm`
+* `ReleaseParameters.Enc` to `ReleaseParameters.Algorithm`
+* `KeyOperationParameters.AAD` to `KeyOperationParameters.AdditionalAuthenticationData`
+* `KeyOperationParameters.Tag` to `KeyOperationParameters.AuthenticationTag`
+* `JSONWebKeyOperation` to `KeyOperation`
+* `JSONWebKeyCurveName` to `KeyCurveName`
+* `JSONWebKeyEncryptionAlgorithm` to `EncryptionAlgorithm`
+* `JSONWebKeySignatureAlgorithm` to `SignatureAlgorithm`
+* `JSONWebKeyType` to `KeyType`
+* Removed `DeletionRecoveryLevel` type
 
 ### Other Changes
 * Updated dependencies
--- a/sdk/security/keyvault/azkeys/CHANGELOG.md
+++ b/sdk/security/keyvault/azkeys/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## 1.0.0 (unreleased)
 
 ### Breaking Changes
+
+*Breaking Changes from last preview. Going forward, the module is now stable and will not have breaking api changes.*
+
 * Renamed `GetRandomBytesRequest` to `GetRandomBytesParameters`
 * `ListDeletedKey` to `ListDeletedKeyProperties`
 * `ListKeys` to `ListKeyProperties`
@@ -14,11 +17,9 @@
 * `DeletedKeyListResult` to `DeletedKeyPropertiesListResult`
 * `KeyListResult` `KeyPropertiesListResult`
 * `KeyOperationsParameters` to `KeyOperationParameters`
-* `KeyOps` field to `KeyOperations`
 * Changed `JSONWebKey.KeyOperations` from type []*string to []*KeyOperation
-* `KeyEncryptionAlgorithm` to `KeyExportEncryptionAlgorithm`
 * `ReleaseParameters.Enc` to `ReleaseParameters.Algorithm`
-* `KeyOperationParameters.AAD` to `KeyOperationParameters.AdditionalAuthenticationData`
+* `KeyOperationParameters.AAD` to `KeyOperationParameters.AdditionalAuthenticatedData`
 * `KeyOperationParameters.Tag` to `KeyOperationParameters.AuthenticationTag`
 * `JSONWebKeyOperation` to `KeyOperation`
 * `JSONWebKeyCurveName` to `KeyCurveName`
--- a/sdk/security/keyvault/azkeys/CHANGELOG.md
+++ b/sdk/security/keyvault/azkeys/CHANGELOG.md
@@ -1,10 +1,8 @@
 # Release History
 
-## 1.0.0 (unreleased)
+## 0.12.0 (unreleased)
 
 ### Breaking Changes
-
-*Breaking Changes from last preview. Going forward, the module is now stable and will not have breaking api changes.*
 
 * Renamed `GetRandomBytesRequest` to `GetRandomBytesParameters`
 * `ListDeletedKey` to `ListDeletedKeyProperties`
@@ -27,6 +25,8 @@
 * `JSONWebKeySignatureAlgorithm` to `SignatureAlgorithm`
 * `JSONWebKeyType` to `KeyType`
 * Removed `DeletionRecoveryLevel` type
+* Deleted `SignatureAlgorithmRSNULL` constant
+* Removed `MaxResults` option
 
 ### Other Changes
 * Updated dependencies
--- a/sdk/security/keyvault/azkeys/autorest.md
+++ b/sdk/security/keyvault/azkeys/autorest.md
@@ -180,6 +180,11 @@ directive:
     where: $
     transform: return $.replace(/.*(\bSignatureAlgorithmRSNULL\b).*/g, "");
 
+   # delete KeyOperationExport
+  - from: constants.go
+    where: $
+    transform: return $.replace(/.*(\bKeyOperationExport\b).*/g, "");
+
   # delete unused error models
   - from: models.go
     where: $
--- a/sdk/security/keyvault/azkeys/autorest.md
+++ b/sdk/security/keyvault/azkeys/autorest.md
@@ -70,9 +70,6 @@ directive:
       from: DeletedKeyBundle
       to: DeletedKey
   - rename-model:
-      from: KeyBundle
-      to: KeyVaultKey
-  - rename-model:
       from: KeyItem
       to: KeyProperties
   - rename-model:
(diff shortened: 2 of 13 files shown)


sdk/security/keyvault/azkeys/constants.go
```diff
@@ -9,244 +9,191 @@
...
+		KeyTypeOctHSM,
+		KeyTypeRSA,
+		KeyTypeRSAHSM,
 	}
 }
 
-// KeyEncryptionAlgorithm - The encryption algorithm to use to protected the exported key material
-type KeyEncryptionAlgorithm string
+// KeyExportEncryptionAlgorithm - The encryption algorithm to use to protected the exported key material
+type KeyExportEncryptionAlgorithm string
```
chlowell, line 183: I like the old name better because these are simply algorithms for encrypting keys. They aren't algorithms specifically for encrypting keys for export, though that may be Key Vault's only use of them today.

sdk/security/keyvault/azkeys/constants.go
```diff
@@ -9,244 +9,191 @@
 
 package azkeys
 
-// DeletionRecoveryLevel - Reflects the deletion recovery level currently in effect for keys in the current vault. If it contains
-// 'Purgeable' the key can be permanently deleted by a privileged user; otherwise, only the system
-// can purge the key, at the end of the retention interval.
-type DeletionRecoveryLevel string
+// KeyCurveName - Elliptic curve name. For valid values, see JsonWebKeyCurveName.
+type KeyCurveName string
```
chlowell, line 13: How about just `CurveName`?

gracewilcox: I prefer `KeyCurveName` since it matches the other languages. This type is also used in `azcertificates` and the Key prefix adds clarity on what the type is for.

sdk/security/keyvault/azkeys/autorest.md
```diff
@@ -74,17 +50,144 @@ directive:
...
+      from: LifetimeActionsType
+      to: KeyRotationLifetimeActionsType
+
+  # rename KeyItem and KeyBundle
+  - rename-model:
+      from: DeletedKeyBundle
+      to: DeletedKey
+  - rename-model:
+      from: KeyBundle
+      to: KeyVaultKey
```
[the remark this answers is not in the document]
gracewilcox: Ok, down with `Key` name. Only thing now is users will have to always include the struct name when using the result. `resp.Key.KID` beomes `resp.Key.Key.KID`

sdk/security/keyvault/azkeys/constants.go
gracewilcox: Added the Export part to help differentiate it from type `Encryption algorithm`

sdk/security/keyvault/azkeys/autorest.md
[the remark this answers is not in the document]
chlowell: I'm lost 😄 what type is `resp`?

sdk/security/keyvault/azkeys/constants.go
chlowell: But **Key**EncryptionAlgorithm is already different and model definitions make it obvious where to use each type

sdk/security/keyvault/azkeys/constants.go
chlowell: Matching other languages isn't important because developers read our docs when writing Go. I could live with `KeyCurveName`, but I do think `CurveName` is better, and precise in the context of crypto keys. (If you're really worried about a vague name here, `EllipticCurveName` removes all doubt.)

I don't think the "Key" prefix helps developers figure out how to use this type. Developers don't list a package's types and try to intuit by their names how to use them. They start with method calls, which lead them to the relevant models, whose fields lead them to the relevant types, whose names aren't so important because the compiler points out the right one.

sdk/security/keyvault/azkeys/autorest.md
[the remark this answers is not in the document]
gracewilcox: Sorry! I was just talking about how the renaming will affect how users will handle the response in the code. You can see the difference in the examples/tests.

sdk/security/keyvault/azkeys/constants.go
heaths: Right, the "Key" prefix in azcertificates was to indicate it was the curve name of the certificate backing key. It was significant there, but seemingly redundant here in azkeys.

heaths:
Does go's pipeline not automatically create APIViews for PRs? I don't see one. Would be easier for just a public API review. In lieu of that, though, I'll review here. Might be something to ping whoever your lang's EngSys person is.

sdk/security/keyvault/azkeys/autorest.md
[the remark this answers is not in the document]
chlowell: Ah, I see, we've created this interesting case of struct embedding:
```diff
---type KeyBundle struct {
+++type Key struct {
	Key *JSONWebKey
	Tags map[string]*string
}

type GetKeyResponse struct {
---	KeyBundle
+++	Key
}
```
Now `GetKeyResponse.Tags` is okay, that's a promoted field, and `GetKeyResponse.Key.Tags` is a qualified reference to the same field. `GetKeyResponse.Key` is ambiguous though: is `Key` the embedded type or that type's `Key` field? The qualified and promoted field syntax can't coexist, so only the qualified one is valid and everyone must write `resp.Key.Key.Foo`. That's no good, let's not do this 😆

I suppose our two options are to name the type something other than `Key` or rename its `Key` field. I like the simpler option: stick with `KeyBundle`.

heaths (commented):
Would it be possible to upload to APIView to view a diff of new/old there?

sdk/security/keyvault/azkeys/constants.go
gracewilcox: Keeping as `KeyEncryptionAlgorithm`. Fixed in [6eb8c22](https://github.com/Azure/azure-sdk-for-go/pull/20892/commits/6eb8c22e41825a190c2df4fdd2f2361e26f93280)

sdk/security/keyvault/azkeys/constants.go
gracewilcox: Will keep as `CurveName`. Will also have it as `CurveName` in `azcertificates` for consistency.

sdk/security/keyvault/azkeys/constants.go
heaths: I missed the context here, but in other languages we call this `KeyWrapAlgorithm` because it coincides with the `WrapKey` method. It just sounded better the other way.

sdk/security/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
...
+* `DeletedKeyItem` to `DeletedKeyProperties`
+* `KeyItem` to `KeyProperties`
+* `DeletedKeyListResult` to `DeletedKeyPropertiesListResult`
+* `KeyListResult` `KeyPropertiesListResult`
+* `KeyOperationsParameters` to `KeyOperationParameters`
+* `KeyOps` field to `KeyOperations`
+* Changed `JSONWebKey.KeyOperations` from type []*string to []*KeyOperation
+* `KeyEncryptionAlgorithm` to `KeyExportEncryptionAlgorithm`
+* `ReleaseParameters.Enc` to `ReleaseParameters.Algorithm`
+* `KeyOperationParameters.AAD` to `KeyOperationParameters.AdditionalAuthenticationData`
```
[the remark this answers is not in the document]
gracewilcox: Typo in Changelog, fixed in 
0e05cc2

sdk/security/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
...
+* `ListKeys` to `ListKeyProperties`
+* `DeletedKeyBundle` to `DeletedKey`
+* `KeyBundle` to `KeyVaultKey`
+* `RestoreKeyParameters.KeyBundleBackup` to `RestoreKeyParameters.KeyBackup`
+* `DeletedKeyItem` to `DeletedKeyProperties`
+* `KeyItem` to `KeyProperties`
+* `DeletedKeyListResult` to `DeletedKeyPropertiesListResult`
+* `KeyListResult` `KeyPropertiesListResult`
+* `KeyOperationsParameters` to `KeyOperationParameters`
+* `KeyOps` field to `KeyOperations`
```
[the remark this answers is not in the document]
gracewilcox: Changed back to original KeyOps in 
0e05cc2

sdk/security/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,6 +1,31 @@
 # Release History
 
-## 1.0.0 (2023-05-11)
+## 1.0.0 (unreleased)
+
+### Breaking Changes
```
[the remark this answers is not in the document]
gracewilcox: Added warning message in 
0e05cc2

gracewilcox:
> Does go's pipeline not automatically create APIViews for PRs? I don't see one. Would be easier for just a public API review. In lieu of that, though, I'll review here. Might be something to ping whoever your lang's EngSys person is.

Thanks for reviewing! Apiview just created an automatic diff once things are merged. I'll ask @benbp about it. I created a manual one [here](https://apiview.dev/Assemblies/Review/afdff47c3cd940c193ab5df0611d0264?diffRevisionId=14d60e038e784dfb8c6b675a2a2ffbb9&diffOnly=False&revisionId=5611e64b98174872a92bc4ea52fdbc9c&doc=False)

heaths:
> > Does go's pipeline not automatically create APIViews for PRs? I don't see one. Would be easier for just a public API review. In lieu of that, though, I'll review here. Might be something to ping whoever your lang's EngSys person is.
> 
> Thanks for reviewing! Apiview just created an automatic diff once things are merged. I'll ask @benbp about it. I created a manual one [here](https://apiview.dev/Assemblies/Review/afdff47c3cd940c193ab5df0611d0264?diffRevisionId=14d60e038e784dfb8c6b675a2a2ffbb9&diffOnly=False&revisionId=5611e64b98174872a92bc4ea52fdbc9c&doc=False)

Thanks! I left a few comments in there. Among those, we do have separate `KeyWrapAlgorithm` and `KeyExportEncryptionAlgorithm` enums which are distinct and for `WrapKey` and `Export`, respectively.

heaths:
> @heaths It's currently set to create reviews on main CI only. I do not remember the context on why it got set this way though.

Any chance we can have it do PRs like I know .NET does, at least?

sdk/security/keyvault/azkeys/CHANGELOG.md
```diff
@@ -1,11 +1,9 @@
 # Release History
 
-## 1.0.0 (unreleased)
+## 0.12.0 (unreleased)
```
heaths, line 3: I thought this was finally GA'ing? Or are you doing one final RC?

heaths (approved):

sdk/security/keyvault/azkeys/CHANGELOG.md
gracewilcox: Talked it over with Rick today and want to do one final beta since we're changing so much of the surface area.

heaths (approved):

sdk/security/keyvault/azkeys/utils_test.go
```diff
@@ -245,3 +246,28 @@ func getMarshalledReleasePolicy(attestationURL string) []byte {
...
+	err = model.UnmarshalJSON(data)
+	require.NoError(t, err)
+
+	// testing unmarshal error scenarios
+	var data2 []byte
+	err = model.UnmarshalJSON(data2)
+	require.Error(t, err)
+
+	m := regexp.MustCompile(":.*$")
+	modifiedData := m.ReplaceAllString(string(data), ":[\"test\", \"test1\", \"test2\"]}")
```
chlowell, line 267: Just FYI, you can use a raw string literal (backticks) to avoid escaping quotes:
```suggestion
	modifiedData := m.ReplaceAllString(string(data), `:["test", "test1", "test2"]}`)
```

But do you really need a regex replace? It looks like the intent is to make the data invalid JSON so unmarshal is certain to fail; if so, you could do something simpler like `data = append(data, 0x42)`

sdk/security/keyvault/azkeys/autorest.md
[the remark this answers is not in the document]
gracewilcox: Charles convinced me offline to change back to `KeyBundle`. :)
Changed in dc700d5

chlowell (approved):

sdk/security/keyvault/azkeys/utils_test.go
```diff
@@ -245,3 +246,28 @@ func getMarshalledReleasePolicy(attestationURL string) []byte {
...
+func testSerde[T serdeModel](t *testing.T, model T) {
+	data, err := model.MarshalJSON()
+	require.NoError(t, err)
+	err = model.UnmarshalJSON(data)
+	require.NoError(t, err)
+
+	// testing unmarshal error scenarios
+	var data2 []byte
+	err = model.UnmarshalJSON(data2)
+	require.Error(t, err)
```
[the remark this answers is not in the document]
gracewilcox: fixed in [517ecd7](https://github.com/Azure/azure-sdk-for-go/pull/20892/commits/517ecd72ed62fa4fa6c5384fb482635240ab2ee4)

sdk/security/keyvault/azkeys/utils_test.go
gracewilcox: There are two error scenarios in the JSON Unmarshaller. The first is where the JSON is completely invalid. This tests the second scenario where it's a valid json, but it unmarshals the field into the wrong type.  So, it might be expecting a string for a specific field, but I modified the JSON to have an array instead which causes an error like:

`unmarshalling type *azkeys.KeyAttributes: json: cannot unmarshal array into Go value of type map[string]json.RawMessage"}`

Testing these two error scenarios improves code coverage from 79.9% to 83%. 😎

PR #20916, opened 2023-05-30
Initial implementation of the azotel module

jhendrixMSFT:
It contains a complete end-to-end example of creating an OTel TracerAdapter that exports spans to Jaeger.

Addresses https://github.com/Azure/azure-sdk-for-go/issues/7358


sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+
+// TracingProviderOptions contains the optional values for NewTracingProvider.
+type TracingProviderOptions struct {
+	// for future expansion
+}
+
+// NewTracingProvider creates a new tracing.Provider that wraps the specified OpenTelemetry TracerProvider.
+//   - tracerProvider - the TracerProvider to wrap
+//   - opts - optional configuration. pass nil to accept the default values
+func NewTracingProvider(tracerProvider *tracesdk.TracerProvider, opts *TracingProviderOptions) tracing.Provider {
```
richardpark-msft, line 29: Small thing - given the similarity in names between our types and theirs perhaps we can alias 'tracesdk' as 'oteltrace' or something?

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+			var attrs []attribute.KeyValue
+			if options != nil {
+				kind = options.Kind
+				attrs = convertAttributes(options.Attributes)
+			}
+			ctx, span := tracer.Start(ctx, spanName, trace.WithSpanKind(convertSpanKind(kind)), trace.WithAttributes(attrs...))
+			return ctx, convertSpan(span)
+		}, &tracing.TracerOptions{
+			SpanFromContext: func(ctx context.Context) (tracing.Span, bool) {
+				return convertSpan(trace.SpanFromContext(ctx)), true
```
richardpark-msft, line 44: Minor - would it be possible to just have `convertSpan` match the signature needed for `SpanFromContext`? We could avoid one closure/anon func if so.

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+func convertAttributes(attrs []tracing.Attribute) []attribute.KeyValue {
+	if len(attrs) == 0 {
+		return nil
+	}
+	keyvals := []attribute.KeyValue{}
+	for _, kv := range attrs {
+		switch vv := kv.Value.(type) {
+		case int:
+			keyvals = append(keyvals, attribute.Int(kv.Key, vv))
+		case int64:
```
richardpark-msft, line 81: Is there any "integer promotion" that needs to happen here for int32, uint, etc..?

sdk/tracing/azotel/otel_test.go
```diff
@@ -0,0 +1,259 @@
...
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func TestNewTracingProvider(t *testing.T) {
+	exporter := &testExporter{}
+
+	otelTP := tracesdk.NewTracerProvider(tracesdk.WithBatcher(exporter))
+
+	options := azcore.ClientOptions{}
```
richardpark-msft, line 32: I briefly wondered - did Joel do this because using a literal initialize would actually use one _extra_ line? 

And I have to say, I couldn't know for sure.

sdk/tracing/azotel/otel_test.go
```diff
@@ -0,0 +1,259 @@
...
+	span.AddError(io.EOF)
+	assert.Equal(t, io.EOF, ts.recordedError)
+
+	const eventName = "event"
+	eventAttr := tracing.Attribute{
+		Key:   "key",
+		Value: "value",
+	}
+	span.AddEvent(eventName, eventAttr)
+	assert.EqualValues(t, eventName, ts.eventName)
```
richardpark-msft, line 78: Question - what's your internal rule for using 'assert' vs 'require'? I always just end up using 'require' but that's just what I'm used to.

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+				return convertSpan(trace.SpanFromContext(ctx)), true
+			},
+		})
+
+	}, nil)
+}
+
+func convertSpan(traceSpan trace.Span) tracing.Span {
+	impl := tracing.SpanImpl{
+		End: func() {
```
richardpark-msft, line 53: NIT: Signature matches for `End` and `traceSpan.End`?

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+			traceSpan.End()
+		},
+		SetAttributes: func(attrs ...tracing.Attribute) {
+			traceSpan.SetAttributes(convertAttributes(attrs)...)
+		},
+		AddEvent: func(name string, attrs ...tracing.Attribute) {
+			traceSpan.AddEvent(name, trace.WithAttributes(convertAttributes(attrs)...))
+		},
+		AddError: func(err error) {
```
richardpark-msft, line 62: NIT: signature matches for AddError and traceSpan.RecordError?

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+		SetStatus: func(code tracing.SpanStatus, desc string) {
+			traceSpan.SetStatus(convertStatus(code), desc)
+		},
+	}
+	return tracing.NewSpan(impl)
+}
+
+func convertAttributes(attrs []tracing.Attribute) []attribute.KeyValue {
+	if len(attrs) == 0 {
```
richardpark-msft, line 73: NIT: I'm always iffy on these. I mean technically the 'attrs' below probably falls through but...yeah. I'd probably do this too.

sdk/tracing/azotel/otel_test.go
```diff
@@ -0,0 +1,259 @@
...
+			value: "hello",
+			validate: func(t *testing.T, kv attribute.KeyValue) {
+				require.EqualValues(t, "STRING", kv.Value.Type().String())
+				assert.EqualValues(t, "hello", kv.Value.AsString())
+			},
+		},
+		{
+			label: "float32",
+			key:   "float32_key",
+			value: float32(3.14159),
```
richardpark-msft, line 152: I added a comment for this where you do this conversion, but it seems like we can provide a straight promotion for these to the "larger" version of the type that we do support. Although I know sometimes float promotion can be a double-edged sword.

richardpark-msft (approved):
Looks good to me, I'm a bit late to the game though so let me know if you've already talked about some of the issues I brought up.

sdk/tracing/azotel/otel.go
jhendrixMSFT: The clash here is due to `go.opentelemetry.io/otel` having two packages named `trace`, `go.opentelemetry.io/otel/sdk/trace` and `go.opentelemetry.io/otel/trace`.  I'm not opposed to changing it, but I chose `tracesdk` as I've seen this in other OTel samples for Go.

sdk/tracing/azotel/otel_test.go
jhendrixMSFT: I usually use `require` as well and in hindsight, perhaps a little too liberally.  Since a failed `require` will fail the test at that point, the rest of the test isn't executed.  In some cases this is desirable, e.g. `require.NotNil(t, widget)` before validating fields within `widget`.  Here, a failed `assert` shouldn't prevent the rest of the test from executing.

sdk/tracing/azotel/otel.go
jhendrixMSFT: Good question.  I'm tempted to say no because OTel doesn't natively support other types (see [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/attribute/kv.go)).  And we [document](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/tracing/tracing.go#L198) the natively supported types, falling back to `fmt.Sprintf()` as required.

@lmolkova do you have an opinion on this?

sdk/tracing/azotel/otel.go
richardpark-msft: I see, makes sense.

It looks like you could rename `tracesdk` to `otel` or something since the other `trace` package isn't exported via the public surface here.

jhendrixMSFT: Good idea.  It could end up colliding with the `otel` package in the future, but we'll deal with it then (if that ever happens).

jhendrixMSFT: Maybe `otelsdk`.

sdk/tracing/azotel/otel.go
jhendrixMSFT: I noticed this too.  I've opened https://github.com/Azure/azure-sdk-for-go/issues/20924 to track it.

sdk/tracing/azotel/otel.go
jhendrixMSFT: Done.

sdk/tracing/azotel/otel_test.go
jhendrixMSFT: Turns out, using a literal was one line shorter than what I had.  Save the new-lines!

sdk/tracing/azotel/otel.go
jhendrixMSFT: You're right though, it is superfluous.  Removed.

sdk/tracing/azotel/otel.go
jhendrixMSFT: @lmolkova do we want optional values for `End()`?  I don't recall discussing this.

sdk/tracing/azotel/otel.go
jhendrixMSFT: @lmolkova do we want optional values for `AddError()`?  I don't recall discussing this.

sdk/tracing/azotel/otel.go
lmolkova: only as an extensibility point in case we want to set status along with end or if OTel decides do add another end method with parameters - there were some talks about it.

sdk/tracing/azotel/otel.go
lmolkova: only as an extensibility point in case there could be errors not represented with `error` object.  BTW, are we going to use `AddError` method at all somewhere?

lmolkova (approved):

sdk/tracing/azotel/otel.go
```diff
@@ -0,0 +1,117 @@
...
+type TracingProviderOptions struct {
+	// for future expansion
+}
+
+// NewTracingProvider creates a new tracing.Provider that wraps the specified OpenTelemetry TracerProvider.
+//   - tracerProvider - the TracerProvider to wrap
+//   - opts - optional configuration. pass nil to accept the default values
+func NewTracingProvider(tracerProvider *otelsdk.TracerProvider, opts *TracingProviderOptions) tracing.Provider {
+	return tracing.NewProvider(func(namespace, version string) tracing.Tracer {
+		tracer := tracerProvider.Tracer(namespace, trace.WithInstrumentationVersion(version))
```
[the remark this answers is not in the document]
jhendrixMSFT: So this is where it gets a little clunky.  We can certainly set the schema URL here (1.7.0 IIRC).  However, when creating an OTel `TracerProvider` with a resource, a schema URL must also be specified (see the example in `examples_test.go`).  This means that customers must import the matching schema URL package (`go.opentelemetry.io/otel/semconv/v1.7.0`).

sdk/tracing/azotel/otel.go
jhendrixMSFT: Looks like this is not being used at present.  Should it be removed?

sdk/tracing/azotel/otel.go
jhendrixMSFT: We don't use these helpers externally.  Since `Value` is exposed as `any` type from `azcore`, we check the type here before calling the appropriate `attribute.Int()` etc helpers.  If we need to support other types that what we [documented](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/tracing/tracing.go#L197-L200) then we need to check for them here.

sdk/tracing/azotel/otel.go
jhendrixMSFT: Is it enough to call `SetStatus()` before calling `End()` (that's how we currently do it in `azcore`).

sdk/tracing/azotel/otel.go
lmolkova: yep, let's remove it - we don't really capture full exceptions in other languages and delegate it to logging which aligns well with otel vision.

sdk/tracing/azotel/otel.go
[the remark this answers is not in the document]
lmolkova: I mean we (when instrumenting HTTP or messaging calls) set attributes with compliance with specific version, e.g. `https://opentelemetry.io/schemas/1.20.0`. Attribute names we use are relevant for this version and might not be available  for other versions.

We should set it to allow backends to interpret telemetry properly.

for us, the schema is a hardcoded string (we can in future support more than one version for back-compat story and in the same way as we allow multiple service API versions). The [format of the string is documented](https://github.com/open-telemetry/opentelemetry-specification/blob/e66cfeb34fbfebf973579624d664cb8df49d0824/specification/schemas/README.md#how-schemas-work) and is used in different tooling across otel. There is no need to depend on anything to stamp it.

As for resource schema URL - it describes schema of the resource attributes users want to populate - they can use a hardcoded string or depend on semconv package - their choice.

sdk/tracing/azotel/otel.go
jhendrixMSFT: Opened https://github.com/Azure/azure-sdk-for-go/issues/20933 to track this.

sdk/tracing/azotel/otel.go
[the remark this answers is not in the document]
jhendrixMSFT: I've opened https://github.com/Azure/azure-sdk-for-go/issues/20934 to track this for the next beta.

sdk/tracing/azotel/otel_test.go
jhendrixMSFT: Agreed on rough edges WRT floats.  I'm going to leave this as-is for now which will require callers to perform the cast themselves if they don't want the `fmt.Sprintf()` fallback.  We can revisit later based on feedback.

sdk/tracing/azotel/otel.go
jhendrixMSFT: Going to leave this as-is for now.  If callers don't want the `fmt.Sprintf()` fallback they can cast the values themselves (we can adjust accordingly per any feedback).

sdk/tracing/azotel/examples_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
```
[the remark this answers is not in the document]
jhendrixMSFT: Looks like 1.17.0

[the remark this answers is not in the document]
lmolkova: yep, we should currently be using 1.17.0

PR #20932, opened 2023-06-01
[azservicebus] Some cleanup in prep for release

richardpark-msft:
Some tests fixes.

- Some of the new DefaultRule tests were defer'ing their cleanup a bit late.
- The ScheduleMessages() tests were inconsistent and relying on way too much internal behavior/timing. I've changed them to be more reliable and only use public behavior.

### DIFF
--- a/sdk/messaging/azservicebus/sender_test.go
+++ b/sdk/messaging/azservicebus/sender_test.go
@@ -251,74 +251,14 @@ func messageFromReceivedMessage(t *testing.T, receivedMessage *ReceivedMessage) 
 }
 
 func Test_Sender_ScheduleAMQPMessages(t *testing.T) {
-	ctx := context.Background()
-
-	client, cleanup, queueName := setupLiveTest(t, nil)
-	defer cleanup()
-
-	receiver, err := client.NewReceiverForQueue(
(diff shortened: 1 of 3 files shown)


sdk/messaging/azservicebus/sender_test.go
```diff
@@ -334,50 +274,83 @@ func Test_Sender_ScheduleMessages(t *testing.T) {
...
+					"x-opt-scheduled-enqueue-time": &farFuture,
+				},
+			}, nil)
+		require.NoError(t, err)
+	}
 
-	// cancel one of the ones scheduled using `ScheduleMessages`
-	err = sender.CancelScheduledMessages(ctx, []int64{sequenceNumbers[0]}, nil)
-	require.NoError(t, err)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
jhendrixMSFT, line 335: Looks like this `cancel()` isn't called.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/sender_test.go
richardpark-msft: Added.

PR #20941, opened 2023-06-02
Replace pow util function with bit operation

MartinForReal:
Replace pow util function with bit operation in runtime package

### DIFF
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -59,15 +59,7 @@ func setDefaults(o *policy.RetryOptions) {
 }
 
 func calcDelay(o policy.RetryOptions, try int32) time.Duration { // try is >=1; never 0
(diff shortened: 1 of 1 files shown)


sdk/azcore/runtime/policy_retry.go
```diff
@@ -59,15 +59,11 @@ func setDefaults(o *policy.RetryOptions) {
 }
 
 func calcDelay(o policy.RetryOptions, try int32) time.Duration { // try is >=1; never 0
-	pow := func(number int64, exponent int32) int64 { // pow is nested helper function
-		var result int64 = 1
-		for n := int32(0); n < exponent; n++ {
-			result *= number
-		}
-		return result
+	pow := func(exponent int32) uint64 { // pow is nested helper function
```
jhendrixMSFT, line 62: Since this is no longer really `pow()` but power of two, and it's only used in one place, we should just remove the helper func.

jhendrixMSFT (approved):

PR #20978, opened 2023-06-06
[keyvault] azcertificates renames

gracewilcox:
Renames for azcertificates based on feedback.

This pull request just contains renames for the module. Exploring the more structural suggestions now for a later pull request.

### DIFF
--- a/sdk/security/keyvault/azcertificates/autorest.md
+++ b/sdk/security/keyvault/azcertificates/autorest.md
@@ -53,28 +53,83 @@ directive:
   # rename paged operations from Get* to List*
   - rename-operation:
       from: GetCertificates
-      to: ListCertificates
+      to: ListCertificateProperties
   - rename-operation:
       from: GetCertificateIssuers
-      to: ListCertificateIssuers
+      to: ListIssuerProperties
   - rename-operation:
       from: GetCertificateVersions
-      to: ListCertificateVersions
+      to: ListCertificatePropertiesVersions
   - rename-operation:
       from: GetDeletedCertificates
-      to: ListDeletedCertificates
+      to: ListDeletedCertificateProperties
+  - rename-model:
+      from: CertificateListResult
+      to: CertificatePropertiesListResult
+  - rename-model:
+      from: DeletedCertificateListResult
+      to: DeletedCertificatePropertiesListResult
 
-  # Maxresults -> MaxResults
-  - from: swagger-document
-    where: $.paths..parameters..[?(@.name=='maxresults')]
-    transform: $["x-ms-client-name"] = "MaxResults"
+  # remove redunant "certificate" from operation name
+  - rename-operation:
+      from: SetCertificateContacts
+      to: SetContacts
+  - rename-operation:
+      from: GetCertificateContacts
+      to: GetContacts
+  - rename-operation:
+      from: DeleteCertificateContacts
+      to: DeleteContacts
+  - rename-operation:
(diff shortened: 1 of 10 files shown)


heaths:
Could you create an apiview for this as well? Much easier to diff across languages and see the final result.

gracewilcox:
> Could you create an apiview for this as well? Much easier to diff across languages and see the final result.

Api view here: https://apiview.dev/Assemblies/Review/66481607611444989da6998d76c414d6?diffRevisionId=b4b48fc855284881971c5cf94426954e&diffOnly=False&revisionId=6811d1794b714328a2da849ae047ade4&doc=False

sdk/security/keyvault/azcertificates/autorest.md
```diff
@@ -119,7 +206,7 @@ directive:
   # (specifying models because others have "ID" fields whose values aren't cert IDs)
   - from: models.go
     where: $
-    transform: return $.replace(/(type (?:Deleted)?Certificate(?:Bundle|Item) struct \{(?:\s.+\s)+\sID \*)string/g, "$1ID")
+    transform: return $.replace(/(type (?:Deleted)?Certificate(?:\s|Properties\s)struct \{(?:\s.+\s)+\sID \*)string/g, "$1ID")
```
chlowell, line 209: This can be simpler because there will always be a space between the type name and `struct`:
```suggestion
    transform: return $.replace(/(type (?:Deleted)?Certificate(?:Properties)? struct \{(?:\s.+\s)+\sID \*)string/g, "$1ID")
```

heaths (changes requested):
All my review comments are in APIView, though I recommend reading all of them first. Given the nature of how I was reviewing code, how APIView orders the ToC vs. the content area, etc., some comments build on others and may not make sense read alone. I'd be happy to go over feedback in person as well, if that'd be easier.

sdk/security/keyvault/azcertificates/autorest.md
gracewilcox: fixed in 361ff3f

sdk/security/keyvault/azcertificates/models.go
```diff
@@ -461,20 +433,8 @@ type ImportCertificateParameters struct {
 	Tags map[string]*string `json:"tags,omitempty"`
 }
 
-// IssuerAttributes - The attributes of an issuer managed by the Key Vault service.
```
heaths, line 464: Commenting close: `ImportCertificateParmaeters should be `[]byte` for `Base64EncodedCertificate`, which you agreed to fix in APIView: https://apiview.dev/Assemblies/Review/66481607611444989da6998d76c414d6?diffRevisionId=b4b48fc855284881971c5cf94426954e&diffOnly=False&revisionId=6811d1794b714328a2da849ae047ade4&doc=False#4643f49b0d4c46be81f3ddd4b0e3d88d

heaths (approved):
There are still a few pending comments on https://apiview.dev/Assemblies/Review/66481607611444989da6998d76c414d6 after reviewing this, but otherwise LGTM.

sdk/security/keyvault/azcertificates/models.go
gracewilcox: Investigated further and talked with Charles, it turns out changing the type to []bytes is not a good idea because it would require a nuanced convenience layer there that doesn't translate well to Go and doesn't provide enough value to users to justify the change.

PR #20995, opened 2023-06-08
Remove autorest core version from build scripts

jhendrixMSFT:
The code generator MUST be the ONLY source of this value.


richardpark-msft (approved):

tadelesh:
IIRC I added this config because of some weird behavior when generation. The autorest have some chances not use the core version config in the codegen readme. I don't know whether the root cause is cache or something else. But anyway, I'm OK with removing it and I'll try to find the root cause if I fall into the problem again.

tadelesh (approved):

PR #21015, opened 2023-06-14
sdk/resourcemanager/cognitiveservices/armcognitiveservices live test

Alancere:
0.57(24/42)


tadelesh:
@Alancere cognitive service is important. Could you increase the coverage for it?

PR #21019, opened 2023-06-14
[Datalake] SAS Support

tasherif-msft:


sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+type DatalakeSignatureValues struct {
+	Version              string    `param:"sv"`  // If not specified, this defaults to Version
+	Protocol             Protocol  `param:"spr"` // See the Protocol* constants
+	StartTime            time.Time `param:"st"`  // Not specified if IsZero
+	ExpiryTime           time.Time `param:"se"`  // Not specified if IsZero
+	Permissions          string    `param:"sp"`  // Create by initializing FilesystemPermissions, FilePermissions or DirectoryPermissions and then call String()
+	IPRange              IPRange   `param:"sip"`
+	Identifier           string    `param:"si"`
+	FilesystemName       string
+	FilePath             string // Use "" to create a Filesystem SAS
```
gapra-msft, line 31: this isn't totally accurate right? for Filesystem SAS don't FilePath and DirectoryPath need to be ""

sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+	DirectoryPath        string // Not nil for a directory SAS (ie sr=d)
```
gapra-msft, line 32: Can we add a comment here that file path will be set to "" if this is passed?

sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// SignWithSharedKey uses an account's SharedKeyCredential to sign this signature values to produce the proper SAS query parameters.
+func (v DatalakeSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKeyCredential) (QueryParameters, error) {
+	if v.ExpiryTime.IsZero() || v.Permissions == "" {
+		return QueryParameters{}, errors.New("service SAS is missing at least one of these: ExpiryTime or Permissions")
+	}
+
+	//Make sure the permission characters are in the correct order
+	perms, err := parsePathPermissions(v.Permissions)
```
gapra-msft, line 59: are we just parsing as path permissions since it is a super set of file system permissions?

sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+		default:
+			return FilesystemPermissions{}, fmt.Errorf("invalid permission: '%v'", r)
+		}
+	}
+	return p, nil
+}
+
+// FilePermissions type simplifies creating the permissions string for an Azure Storage blob SAS.
+// Initialize an instance of this type and then call its String method to set BlobSignatureValues' Permissions field.
+type FilePermissions struct {
```
gapra-msft, line 335: Should this be Path Permissions?

sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+		udkExpiry,
+		*udk.SignedService,
+		*udk.SignedVersion,
+		v.AuthorizedObjectID,
+		v.UnauthorizedObjectID,
+		v.CorrelationID,
+		v.IPRange.String(),
+		string(v.Protocol),
+		v.Version,
+		resource,
```
gapra-msft, line 194: same here does encryption scope and snapshot need empty spaces here?

gapra-msft (approved):
left a few comments

sdk/storage/azdatalake/sas/service.go
tasherif-msft: that's correct I'll leave a comment on dir as well!

sdk/storage/azdatalake/sas/service.go
tasherif-msft: I call the method pathPermissions because both Dir and File permissions are identical, so this internal method is just used for both

gapra-msft: this also parses filesystem permissions right?

sdk/storage/azdatalake/sas/service.go
gapra-msft: I dont see a DirectoryPermissions here, am I missing something?

sdk/storage/azdatalake/sas/service.go
```diff
@@ -0,0 +1,407 @@
...
+	stringToSign := strings.Join([]string{
+		v.Permissions,
+		startTime,
+		expiryTime,
+		getCanonicalName(sharedKeyCredential.AccountName(), v.FilesystemName, v.FilePath, v.DirectoryPath),
+		signedIdentifier,
+		v.IPRange.String(),
+		string(v.Protocol),
+		v.Version,
+		resource,
```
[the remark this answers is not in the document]
tasherif-msft: neither is supported atm you just want a comment having them for the future?

sdk/storage/azdatalake/sas/service.go
tasherif-msft: same as above

sdk/storage/azdatalake/sas/service.go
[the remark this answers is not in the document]
gapra-msft: even if they arent supported they need to be encoded in the signature properly, right? otherwise any request will fail.

sdk/storage/azdatalake/sas/service.go
tasherif-msft: whoa, idk how I deleted it before pushing... let me add it again haha

gapra-msft: but yeah anyway, Java uses the PathPermissions type and .NET calls it DatalakePermissions, so ultimately it is a design decision here

sdk/storage/azdatalake/sas/service.go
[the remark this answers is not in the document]
tasherif-msft: yes, you're correct, I forgot we're using the blob endpoint and this needs to be encoded.

sdk/storage/azdatalake/sas/service.go
tasherif-msft: I split the parser for paths and fs for future proofing, but yea the two essentially do the same thing

tasherif-msft: I split the parsers of paths and fs just for future proofing, but yea they essentially do the same thing

sdk/storage/azdatalake/sas/service.go
tasherif-msft: I'll go for PathPermissions actually.

PR #21027, opened 2023-06-16
Adding in the CloudEvent type

### LINKED ISSUE #20970: Create a `CloudEvent` type into azcore, for use with messaging SDKs
The `CloudEvent` type is used heavily in the upcoming EventGrid SDK for Go, with potential future integration into other messaging SDKs like webpubsub, azeventhubs, and azservicebus.

The basic type is pretty simple:

```go
type CloudEvent struct {
  Data any               // JSON serializable data

  ID *string

  // defaults to "1.0" (TODO: unsure if empty string is considered a valid version)
  SpecVersion *string
}
```

The `ID` and `SpecVersion` fields will get default values

Some other bits I need to flesh out:

1. CloudEvent has a concept of 'extensions', which allow you to plug in properties that are serialized at the root of the object.
2. The actual model type has a property called data_base64, but the libraries all hide it. It's used internally for transport, but not exposed to the user directly.

richardpark-msft:
The `CloudEvent` type implements the CloudEvent from https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md. This same format (and type) is used in the upcoming EventGrid SDK and is also intended to be usable with other messaging packages as well.

Fixes #20970


sdk/azcore/messaging/cloudevent.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+
+	return nil
+}
+
+func updateFieldFromValue(ce *CloudEvent, k string, raw json.RawMessage) error {
+	switch k {
+	//
+	// required attributes
```
jhendrixMSFT, line 205: Should this func return an error if any required attributes are absent?

sdk/azcore/messaging/cloudevent.go
```diff
@@ -0,0 +1,268 @@
...
+		Data: data,
+	}
+
+	if options != nil {
+		ce.DataContentType = options.DataContentType
+		ce.DataSchema = options.DataSchema
+		ce.Extensions = options.Extensions
+		ce.Subject = options.Subject
+
+		if options.Time != nil {
```
jhendrixMSFT, line 119: I think this can be removed.

richardpark-msft: Removed.

sdk/azcore/messaging/cloudevent.go
richardpark-msft: Added (in a few spots)

sdk/azcore/messaging/cloudevent.go
jhendrixMSFT, line 1: Maybe snake-case the file name?

jhendrixMSFT (approved):

PR #21028, opened 2023-06-17
[azeventgrid] Use the common azcore/messaging/CloudEvent model

richardpark-msft:
Switch over to using the CloudEvent model from azcore, rather than the generated one. Also, removing some more auto-generated dead code (all in autorest.md).

Completes the entire round-trip for #20970

### DIFF
--- a/sdk/messaging/azeventgrid/autorest.md
+++ b/sdk/messaging/azeventgrid/autorest.md
@@ -34,6 +34,19 @@ directive:
   - from: swagger-document
     where: $.definitions.CloudEvent.properties.dataschema
     transform: $["x-ms-client-name"] = "DataSchema"
+  # mark models as external so they're just omitted
+  - from: swagger-document
+    where: $.definitions.CloudEvent
+    transform: $["x-ms-external"] = true
+  - from: swagger-document
+    where: $.definitions.["Azure.Core.Foundations.Error"]
+    transform: $["x-ms-external"] = true
+  - from: swagger-document
+    where: $.definitions.["Azure.Core.Foundations.ErrorResponse"]
+    transform: $["x-ms-external"] = true
+  - from: swagger-document
+    where: $.definitions.["Azure.Core.Foundations.InnerError"]
+    transform: $["x-ms-external"] = true
   # make the endpoint a parameter of the client constructor
   - from: swagger-document
     where: $["x-ms-parameterized-host"]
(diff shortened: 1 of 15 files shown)


sdk/messaging/azeventgrid/autorest.md
```diff
@@ -54,11 +45,25 @@ directive:
...
       - response_types.go
     where: $
-    transform: return $.replace(/AzureCoreFoundations/g, "");
+    transform: return $.replace(/\*CloudEvent/g, "messaging.CloudEvent");
+
+  # delete some models that we don't need.
+  - from:
+      - models.go
+    where: $
+    transform: return $.replace(/\/\/ (AzureCoreFoundation|CloudEvent).+?\n}/sg, "");
```
jhendrixMSFT, line 60: It might be simpler to mark the applicable types as [`x-ms-external`](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-external) so they're omitted from codegen altogether.

sdk/messaging/azeventgrid/response_types.go
```diff
@@ -1,5 +1,3 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: Seems like this shouldn't be removed?

sdk/messaging/azeventgrid/time_rfc3339.go
```diff
@@ -1,5 +1,3 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT, line 1: Another removed license header.

jhendrixMSFT (approved):

sdk/messaging/azeventgrid/response_types.go
richardpark-msft: This is odd.

image

The diff is showing it's deleted but it's not in the branch...

sdk/messaging/azeventgrid/autorest.md
richardpark-msft: Done, this was great!

sdk/messaging/azeventgrid/time_rfc3339.go
richardpark-msft: I'm looking at the actual file in the branch and it's not removed there. I think this might just be a PR cosmetics error. 

(if it gets removed after merge I'll add it back)

sdk/messaging/azeventgrid/response_types.go
richardpark-msft: I'm looking at the actual file in the branch and it's not removed there. I think this might just be a PR cosmetics error. 

(if it gets removed after merge I'll add it back)

richardpark-msft (commented):
Looks like the PR is showing diffs that either don't exist, or are some sort of odd display error. I'll reconcile it afterwards if it turns out they're real.

richardpark-msft:
Just looping back, looks like it was just a display error. `main` looks good.

PR #21042, opened 2023-06-20
[azopenai] Fixing example's typo, creating a constructor for KeyCredential

richardpark-msft:
Fixing a few things based on feedback from @JeffreyRichter.
- Example had a typo and a hardcoded calculation. 
- Adding a constructor for KeyCredential.

### DIFF
--- a/sdk/cognitiveservices/azopenai/examples_client_test.go
+++ b/sdk/cognitiveservices/azopenai/examples_client_test.go
@@ -18,7 +18,7 @@ import (
 func ExampleNewClientForOpenAI() {
 	// NOTE: this constructor creates a client that connects to the public OpenAI endpoint.
 	// To connect to an Azure OpenAI endpoint, use azopenai.NewClient() or azopenai.NewClientWithyKeyCredential.
-	keyCredential, err := azopenai.NewKeyCredential("open-ai-apikey", nil)
+	keyCredential, err := azopenai.NewKeyCredential("<OpenAI-APIKey>", nil)
 
 	if err != nil {
 		panic(err)
@@ -55,7 +55,7 @@ func ExampleNewClient() {
 func ExampleNewClientWithKeyCredential() {
 	// NOTE: this constructor creates a client that connects to an Azure OpenAI endpoint.
 	// To connect to the public OpenAI endpoint, use azopenai.NewClientForOpenAI
-	keyCredential, err := azopenai.NewKeyCredential("Azure OpenAI apikey", nil)
+	keyCredential, err := azopenai.NewKeyCredential("<Azure-OpenAI-APIKey>", nil)
 
 	if err != nil {
 		panic(err)
--- a/sdk/cognitiveservices/azopenai/examples_client_test.go
+++ b/sdk/cognitiveservices/azopenai/examples_client_test.go
@@ -18,8 +18,10 @@ import (
 func ExampleNewClientForOpenAI() {
 	// NOTE: this constructor creates a client that connects to the public OpenAI endpoint.
 	// To connect to an Azure OpenAI endpoint, use azopenai.NewClient() or azopenai.NewClientWithyKeyCredential.
-	keyCredential := azopenai.KeyCredential{
-		APIKey: "open-ai-apikey",
+	keyCredential, err := azopenai.NewKeyCredential("open-ai-apikey", nil)
+
+	if err != nil {
+		panic(err)
 	}
 
 	client, err := azopenai.NewClientForOpenAI("https://api.openai.com/v1", keyCredential, nil)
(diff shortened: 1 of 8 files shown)


sdk/cognitiveservices/azopenai/examples_client_test.go
```diff
@@ -53,8 +55,10 @@ func ExampleNewClient() {
 func ExampleNewClientWithKeyCredential() {
 	// NOTE: this constructor creates a client that connects to an Azure OpenAI endpoint.
 	// To connect to the public OpenAI endpoint, use azopenai.NewClientForOpenAI
-	keyCredential := azopenai.KeyCredential{
-		APIKey: "Azure OpenAI apikey",
+	keyCredential, err := azopenai.NewKeyCredential("Azure OpenAI apikey", nil)
```
JeffreyRichter, line 58: All your examples use qa different string parameter: "open-ai-apikey" or "Azure OpenAI apikey". Please be consistent and if the customer must replace this value, then put it in angle brackets or something, like:<OpenAI-APIKey> or <Azure-OpenAI-APIKey>

sdk/cognitiveservices/azopenai/examples_client_test.go
```diff
@@ -102,7 +108,7 @@ func ExampleClient_GetCompletionsStream() {
 		entry, err := resp.CompletionsStream.Read()
 
 		if errors.Is(err, io.EOF) {
-			fmt.Printf("More more completions")
+			fmt.Printf("No more completions")
```
JeffreyRichter, line 111:
```suggestion
			fmt.Printf("\n *** No more completions ***\n")
```

sdk/cognitiveservices/azopenai/policy_apikey.go
```diff
@@ -14,8 +14,19 @@ import (
...
+}
+
+// KeyCredentialOptions contains the optional parameters for the [NewKeyCredential] method.
+type KeyCredentialOptions struct {
+	// For future expansion
+}
+
+// NewKeyCredential creates a KeyCredential containing an API key for
+// either Azure OpenAI or OpenAI.
+func NewKeyCredential(apiKey string, options *KeyCredentialOptions) (KeyCredential, error) {
```
JeffreyRichter, line 28: I don't think we offer options params for credential types. @chlowell should review this.

richardpark-msft: Just spot checking, but it looks like we do:

```go
func NewAzureCLICredential(options *AzureCLICredentialOptions) (*AzureCLICredential, error)
func NewDefaultAzureCredential(options *DefaultAzureCredentialOptions) (*DefaultAzureCredential, error)
func NewDeviceCodeCredential(options *DeviceCodeCredentialOptions) (*DeviceCodeCredential, error)

```

sdk/cognitiveservices/azopenai/examples_client_test.go
richardpark-msft: Fixed.

sdk/cognitiveservices/azopenai/policy_apikey.go
jhendrixMSFT: For `azblob` we don't https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/common.go#L19

richardpark-msft: TBH, with this library, I'd want to make it as "future flexible" as I can. Feels like adding it is minimal fuss (like any of our `options` taking functions) and gives us some insulation.

jhendrixMSFT: Richard and I were having a side conversation about this, I'll summarize.

The credential types in `azidentity` are higher-level/more complex, covering a wide range of scenarios.  Thus, the need for options types makes sense.

For service-specific key credential types, I would think that we know the shape of the credential which would obviate the need for any options.

Given the rapid change in the OpenAI endpoint, how likely is it that the shape of the key credential will change?

richardpark-msft: For now I'll just remove it, but we should keep it in mind. 

This also is just a beta, we have time to consider it.

sdk/cognitiveservices/azopenai/policy_apikey.go
```diff
@@ -14,8 +14,19 @@ import (
...
 // KeyCredential is used when doing APIKey-based authentication.
 type KeyCredential struct {
-	// APIKey is the api key for the client.
-	APIKey string
+	// apiKey is the api key for the client.
+	apiKey string
+}
+
+// KeyCredentialOptions contains the optional parameters for the [NewKeyCredential] method.
+type KeyCredentialOptions struct {
```
jhendrixMSFT, line 22: Remove?

richardpark-msft: Done.

jhendrixMSFT (approved):

PR #21054, opened 2023-06-24
[azopenai] Adding in image generation, as exported by the current .tsp

richardpark-msft:
Using the latest tsp we now have image generation, so now you can create all the cat images you want.

Some interesting bits:
- The current Azure OpenAI image generation endpoint is an LRO, but that's temporary. So for now we use the poller under the covers, but don't expose it. This also matches what the OpenAI endpoint does for image generation. 
- There's a bug I'm working around in the latest preview service in Azure OpenAI where it returns an empty entry when streaming chat completions/completions (unrelated to the work I did here, but related to the fact that we had to reference a later service version). I've filed that as https://github.com/Azure/azure-sdk-for-go/issues/21086 and marked it as blocking-for-release.


sdk/cognitiveservices/azopenai/examples_client_test.go
```diff
@@ -121,3 +125,44 @@ func ExampleClient_GetCompletionsStream() {
...
+		return
+	}
+
+	keyCredential, err := azopenai.NewKeyCredential(azureOpenAIKey)
+
+	if err != nil {
+		panic(err)
+	}
+
+	client, err := azopenai.NewClientWithKeyCredential(azureOpenAIEndpoint, keyCredential, "", nil)
```
richardpark-msft, line 146: @JeffreyRichter - image generation doesn't take a model deployment. It's making me think we might want to consider moving it per-request instead, and just modifying the URL dynamically on each request, rather than requiring it up-front.

What do you think about that?

richardpark-msft:
NOTE: the tests are just failing because my recordings haven't sanitized/genericized enough content. They pass with live testing so please feel free to review the code :)

sdk/cognitiveservices/azopenai/examples_client_test.go
glecaros: I would think this needs to be a per request param, otherwise we would be requiring the developer customer to create, for example, a client for chat completions and another client for embeddings (which may be ok, but I'm not sure if it is the experience we want to provide)

sdk/cognitiveservices/azopenai/constants.go
```diff
@@ -8,6 +8,28 @@
 
 package azopenai
 
+// AzureOpenAIOperationState - The state of a job or item.
+type AzureOpenAIOperationState string
```
jhendrixMSFT, line 12: Since we're not exposing the poller, can these be removed?

sdk/cognitiveservices/azopenai/models.go
```diff
@@ -320,11 +420,11 @@ type Embeddings struct {
...
 // strings and are commonly used for search, clustering, recommendations, and other similar scenarios.
 type EmbeddingsOptions struct {
-	// REQUIRED; Input text to get embeddings for, encoded as a string. To get embeddings for multiple inputs in a single request,
-	// pass an array of strings. Each input must not exceed 2048 tokens in length.
+	// REQUIRED; Input texts to get embeddings for, encoded as a an array of strings. Each input must not exceed 2048 tokens in
+	// length.
 	// Unless you are embedding code, we suggest replacing newlines (\n) in your input with a single space, as we have observed
 	// inferior results when newlines are present.
-	Input any
+	Input []*string
```
jhendrixMSFT, line 427: I believe we now want slices of scalar types to not be pointer-to-type.

jhendrixMSFT: That will likely have to wait for a code generator update to opt into the behavior though.

jhendrixMSFT: https://github.com/Azure/autorest.go/pull/986

sdk/cognitiveservices/azopenai/constants.go
richardpark-msft: It's still referenced since we're calling into the poller underneath so I've unexported it.

sdk/cognitiveservices/azopenai/models.go
richardpark-msft: Issue filed: https://github.com/Azure/azure-sdk-for-go/issues/21101

sdk/cognitiveservices/azopenai/examples_client_test.go
richardpark-msft: Filed here: https://github.com/Azure/azure-sdk-for-go/issues/21102

I'll get some discussion going. I believe there's some discussion about how this is represented, even in C# (@trrwilson?) where the parameter for the 'model' is 'modelOrDeployment'.

sdk/cognitiveservices/azopenai/constants.go
```diff
@@ -8,6 +8,28 @@
...
+const (
+	azureOpenAIOperationStateCanceled   azureOpenAIOperationState = "canceled"
+	azureOpenAIOperationStateFailed     azureOpenAIOperationState = "failed"
+	azureOpenAIOperationStateNotRunning azureOpenAIOperationState = "notRunning"
+	azureOpenAIOperationStateRunning    azureOpenAIOperationState = "running"
+	azureOpenAIOperationStateSucceeded  azureOpenAIOperationState = "succeeded"
+)
+
+// PossibleazureOpenAIOperationStateValues returns the possible values for the azureOpenAIOperationState const type.
+func PossibleazureOpenAIOperationStateValues() []azureOpenAIOperationState {
```
jhendrixMSFT, line 23: This can also be unexported.

sdk/cognitiveservices/azopenai/custom_client_image.go
```diff
@@ -0,0 +1,84 @@
...
+	// placeholder for future optional parameters
+}
+
+// CreateImageResponse contains the response from method Client.CreateImage.
+type CreateImageResponse struct {
+	ImageGenerations
+}
+
+// CreateImage creates an image using the Dall-E API.
+func (client *Client) CreateImage(ctx context.Context, body ImageGenerationOptions, options *CreateImageOptions) (CreateImageResponse, error) {
```
jhendrixMSFT, line 27: A new nits here (can be fixed later).
`ImageGenerationOptions` needs a better name, maybe something like `ImageGenerationSettings` (or perhaps `CreateImageSettings` so it aligns with `CreateImageOptions`).  And ideally, it would contain all required settings and the optional ones moved to `CreateImageOptions`.

sdk/cognitiveservices/azopenai/custom_models.go
```diff
@@ -29,3 +29,9 @@ type GetChatCompletionsStreamResponse struct {
 	// ChatCompletionsStream returns the stream of completions. Token limits and other settings may limit the number of chat completions returned by the service.
 	ChatCompletionsStream *EventReader[ChatCompletions]
 }
+
+// ImageGenerationsDataItem holds the result of image generation.
```
jhendrixMSFT, line 33: Add a comment that the fields are mutually exclusive?

sdk/cognitiveservices/azopenai/custom_client_image.go
richardpark-msft: Filed: https://github.com/Azure/azure-sdk-for-go/issues/21106

We should match, on data type name, with the other SDKs as much as possible.

jhendrixMSFT: I mostly agree, unless the other SDKs got it pretty wrong.

jhendrixMSFT (approved):

sdk/cognitiveservices/azopenai/custom_client_image.go
richardpark-msft: There's a discussion about it although the context of that was whether we wanted to make parameters positional or not - the main issue there being that there's quite a few. Splitting into options and non-options feels like a good split but I think the other languages have different criteria on whether "things named options" only contain optional parameters.

Issue #21057, opened 2023-06-25
Expose PolicyFunc util to public


MartinForReal:
### Feature Request

PolicyFunc  https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.6.1/internal/exported#PolicyFunc is useful when policy can be defined in one function. 

I'm wondering if we can add a type definition in policy package.

Thanks!

jhendrixMSFT:
Closing per discussion on the PR.

PR #21059, opened 2023-06-25
Expose Policy.Func to public

MartinForReal:
close: #21057 

### DIFF
--- a/sdk/azcore/policy/policy.go
+++ b/sdk/azcore/policy/policy.go
@@ -19,6 +19,9 @@ import (
 // Request and react to the received Response.
 type Policy = exported.Policy
 
+// Func is a type that implements the Policy interface. Use this type when implementing a stateless policy as a first-class function.
+type Func = exported.PolicyFunc
+
 // Transporter represents an HTTP pipeline transport used to send HTTP requests and receive responses.
 type Transporter = exported.Transporter
 
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - `messaging/CloudEvent` allows you to serialize/deserialize CloudEvents, as described in the CloudEvents 1.0 specification: [link](https://github.com/cloudevents/spec)
 * Added method `WithClientName()` to type `azcore.Client` to support shallow cloning of a client with a new name used for tracing.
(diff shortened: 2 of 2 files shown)


MartinForReal:
https://github.com/Azure/azure-sdk-for-go/blob/main/CHANGELOG.md is a dead link. Which change log should I update?

jhendrixMSFT:
Each module has its own changelog.  The one for `azcore` is [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/CHANGELOG.md).

MartinForReal:
change log has been updated.

sdk/azcore/policy/policy.go
```diff
@@ -19,6 +19,9 @@ import (
 // Request and react to the received Response.
 type Policy = exported.Policy
 
+// PolicyFunc is a type that implements the Policy interface. Use this type when implementing a stateless policy as a first-class function.
+type PolicyFunc = exported.PolicyFunc
```
jhendrixMSFT, line 23: This type name stutters, it should probably be `Func`, making the fully qualified type name `policy.Func`.

MartinForReal: I think it will be ambiguous because we don't know from its name what Func does.

jhendrixMSFT: The fully qualified name will be `policy.Func`.  The name will never be just `Func` unless it's referenced from within the same package.  Compared to `policy.PolicyFunc` which stuttered (i.e. the word "policy" is repeated).

MartinForReal: updated.

jhendrixMSFT:
@JeffreyRichter any concerns about exporting this?

jhendrixMSFT:
Customers would use it the exact same way we do internally; to create a stateless pipeline policy from a function (example [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/pipeline.go#L67)).

Customers can achieve the same results by defining their own `pipeline.Policy` type.

JeffreyRichter:
> Customers can achieve the same results by defining their own `pipeline.Policy` type.
Exactly, so I don't see a reason for us to public expose this rarely used and completely optional thing.

Issue #20973, opened 2023-06-06
Add API to azcore.Client for shallow copying with new client name


jhendrixMSFT:
I think it would look something like this.
```go
// WithClientName returns a shallow copy of the Client with its tracing client name changed to clientName.
//   - clientName - the fully qualified name of the client ("package.Client"); this is used by the tracing provider when creating spans
func (c *Client) WithClientName(clientName string) *Client { /*...*/ }
```
This is similar to how [`*http.Request.WithContext()`](https://pkg.go.dev/net/http#Request.WithContext) works.

JeffreyRichter:
This is a public API exposed to customers?

jhendrixMSFT:
The API is on the `azcore.Client` type which is not typically exposed to customers (SDK clients use internally use this).  However, since this client is our raw HTTP client "escape hatch" mechanism, in those scenarios, customers would see this API.

Since this is an exported type from `azcore` there's no way to avoid making the API exported.  However, we could probably do some type-aliasing magic so that the API shows up in the `runtime` package if we think exposing it in the "escape hatch" scenario is unwanted.

PR #21065, opened 2023-06-26
Add support for shallow cloning azcore.Client instances

jhendrixMSFT:
This allows for multiple clients to share the same underlying pipeline while having the correct client name string in traces.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/20973

### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -73,6 +73,11 @@ type ClientOptions = policy.ClientOptions
 type Client struct {
 	pl runtime.Pipeline
 	tr tracing.Tracer
+
+	// cached on the client to support shallow copying with new values
+	tp tracing.Provider
+	mv string
+	ns string
 }
 
 // NewClient creates a new Client instance with the provided values.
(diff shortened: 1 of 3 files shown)


richardpark-msft (approved):
Looks good. From the links it sounds like the intended use case here is that you have a parent client and want to create some child clients that share everything but want to be more specific with their client name (ie, instead of just being BlobClient, they can be BlockBlobClient). 

Is that correct?

jhendrixMSFT:
That's correct.  This way, and spans created by the child client(s) will have the correct client name.

sdk/azcore/core.go
```diff
@@ -73,6 +73,11 @@ type ClientOptions = policy.ClientOptions
 type Client struct {
 	pl runtime.Pipeline
 	tr tracing.Tracer
+
+	// cached on the client to support shallow copying with new values
+	tp tracing.Provider
+	mv string
```
JeffreyRichter, line 79: I can tell that pl is a Pipeline and tr is a Tracer but I can't tell what mv and ns are for. I suggest descriptive names or at least comments.

jhendrixMSFT: Fixed up the names.

JeffreyRichter (approved):

chlowell (approved):

PR #21081, opened 2023-06-29
[azeventgrid] Removing the empty `Interface any` from `PublishCloudEventsResponse`

richardpark-msft:


sdk/messaging/azeventgrid/response_types.go
```diff
@@ -16,8 +16,6 @@ type AcknowledgeCloudEventsResponse struct {
 
 // PublishCloudEventsResponse contains the response from method Client.PublishCloudEvents.
 type PublishCloudEventsResponse struct {
-	// Anything
-	Interface any
```
jhendrixMSFT, line 20: I'm not opposed to this but I think the doc comment will need to be updated to omit the part about "In case of success, the server responds with an HTTP
// 200 status code with an empty JSON object in response"

sdk/messaging/azeventgrid/autorest.md
```diff
@@ -65,4 +65,16 @@ directive:
       - response_types.go
     where: $
     transform: return $.replace(/\*CloudEvent/g, "messaging.CloudEvent");
+
+  # remove the 'Interface any' that's generated for an empty response object.
```
jhendrixMSFT, line 69: You could also use a transform to change the definition of the 200 response to not include a schema (not sure which is easier).

jhendrixMSFT (approved):

sdk/messaging/azeventgrid/response_types.go
richardpark-msft: Fixed by just removing the 'schema' attribute, as you suggested. MUCH cleaner.

richardpark-msft:
Live tests are failing because we ran into a quota limit for Event Grid namespaces in a single subscription.

PR #21098, opened 2023-07-05
Add support for shallow cloning azcore.Client instances (#21065)

jhendrixMSFT:
This allows for multiple clients to share the same underlying pipeline while having the correct client name string in traces.

This is a cherry-pick of 1d8ba7bcb350a85f1f86b388b1847a1fadb9d9e3 with a slightly modified CHANGELOG to reflect the current sequence of releases.  The feature's implementation has also been slightly changed as the current incarnation in `main` was built on features added in the `v1.7.0-beta` series.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -1,11 +1,54 @@
 # Release History
 
+## 1.7.0 (Unreleased)
+
+### Features Added
+* Added method `WithClientName()` to type `azcore.Client` to support shallow cloning of a client with a new name used for tracing.
+
+### Breaking Changes
+> These changes affect only code written against beta versions v1.7.0-beta.1 or v1.7.0-beta.2
+* The beta features for CAE, tracing, and fakes have been omitted for this release.
+
+### Bugs Fixed
+
+### Other Changes
+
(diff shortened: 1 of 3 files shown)


richardpark-msft (approved):

jhendrixMSFT:
We have a bit of a problem here.  The latest preview of `armmonitor` is using `azcore@v1.7.0-beta.2` for fakes and tracing.  Removing that content breaks the dependency check.  While we can override it here, I don't believe we can in the release pipeline.

I see a few options.
- Fix dependency checking.  This is ideal but seems difficult with release week coming up.
- Add a flag to CI to allow skipping of dependency checking.
- Release `azcore` as `v1.6.2` instead of `v1.7.0`.

Frankly, releasing as `v1.6.2` simplifies the upcoming releases but that violates semver rules.  So I think the only option at present is bullet 2 unless folks thing that fixing dependency checking to account for this scenario can happen in the next few days.  Thoughts?

jhendrixMSFT:
Opened https://github.com/Azure/azure-sdk-for-go/issues/21116 to track the CI issue.

jhendrixMSFT:
@benbp what's the correct way for me to rerun this with the flag `Skip.DependencyCheck` set to `true`?

PR #21126, opened 2023-07-10
Always Restore Before Invoking Tests

scbedd:
@gracewilcox this is a good test. It shouldn't need to check out anything new.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -54,7 +54,7 @@ steps:
   - ${{ if eq(parameters.TestProxy, true) }}:
     - template: /eng/common/testproxy/test-proxy-tool.yml
 
-  - pwsh: PowerShell@2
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -54,6 +54,14 @@ steps:
   - ${{ if eq(parameters.TestProxy, true) }}:
     - template: /eng/common/testproxy/test-proxy-tool.yml
 
+  - task: PowerShell@2
```
weshaggard, line 57: Aren't we handling this in other languages by doing it in the test framework? If this is easier for now that is fine but it would be nice if this were integrated so that it would also work for folks running the tests locally.

scbedd: @weshaggard it definitely shouldn't be necessary. Closing this PR for now.

scbedd:
Unnecessary.

PR #21138, opened 2023-07-11
[Release] sdk/resourcemanager/storagemover/armstoragemover/2.0.0-beta.1

Alancere:
https://github.com/Azure/sdk-release-request/issues/4333


Alancere:
The breaking changes are introduced and approved in [this PR](https://github.com/Azure/azure-rest-api-specs/pull/24390).

raych1 (approved):

jhendrixMSFT:
It looks to me that the type changed from non polymorphic to polymorphic which is a breaking change for Go.

tadelesh:
> It looks to me that the type changed from non polymorphic to polymorphic which is a breaking change for Go.

Yes. Type has changed to polymorphic. Other languages either use union (JS) or use inheritance (Java/.Net/Python) could handle it as non-breaking. Seems only Go will be a breaking change.

PR #21155, opened 2023-07-13
[monitor] azingest

gracewilcox:
Initial PR for Azure Monitor Ingestion module


richardpark-msft (approved):
Looks good for a beta. The process of setting up a DCE/DCR seems a bit complicated, but that's not something we can fix.

PR #21168, opened 2023-07-14
Auto-install and run standalone test proxy server per test package

### LINKED ISSUE #21129: Utilize Standalone Proxy
## Context

This issue is actually prompted by what the real solution to [this other discussion](https://github.com/Azure/azure-sdk-tools/issues/6483) that I was having with @gracewilcox  and @benbp .

FIrst, some backstory. @gracewilcox was seeing some really wonky errors in the `azadmin` build. She mentioned that this also occurred in `azfile` and `azblob`. These issues started June 28th, the same day that I touched the test-proxy version. This immediately makes me suspect the proxy is the source of the problems, so I did a bit of digging.

Turns out, the test-proxy fix _did_ introduce this `* is not a valid pathspec` item. It's because we made the proxy _aware_ of what had already been checked out. If the proxy already has a targeted tag checked out, it will invoke a `clean` operation prior to restoring that tag. [This `clean` operation is what was failing, and is the reason for the `proxy changed, error began`](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2892005&view=logs&j=810fa30a-8c2a-57fb-9928-75f3b8cc56e1&t=6b4caddc-c1f8-510e-b712-58500968fa03&l=181).

Given that we're in the middle of the release week, I tried [an alternative](https://github.com/Azure/azure-sdk-for-go/pull/21126) that merely removes the possibility of a race condition.

After I added this, [azadmin continued failing](https://dev.azure.com/azure-sdk/public/_build/results?buildId=2906105&view=logs&j=7c2e6c09-e0bf-5c90-9850-b294a30ca1c6&t=3c7ace2f-c6fe-5b77-76a8-c8cf3bcdc698&l=143).  This new failure revolves around the fact that `azadmin` is actually _multiple_ packages, and as such has multiple test suites. Given there is a _single_ proxy running for all of these packages, and each test-suite has no idea about the others, it is easy for them to stomp on each other. In this case, the new error is because as each test suite in the `azadmin` module finishes, it fires an `Admin/Reset` to clear out any of its customizations. Given that there are _4_ test suites running, there is a VERY high chance that when test suite A fires `Admin/Reset`, test suite B is still running a test. When this happens, the test-proxy rightfully kicks back the reset with `400`, as once can't change a session level sanitizer _while a playback is happening_. 

The part that I'm still scratching my head at is why we haven't seen this at all prior to June 28th. I'd definitely expect there to be consistent failures given this.

I spoke to @gracewilcox about this not reproing locally, and the reasoning there is because given that these are _different packages_ they are never run locally as a single unit like they are in CI!

Why do we see this new error? This is one piece of data that I _don't_ understand.

## Prospective Solution

I think that the test-suite should swap to the standalone test-proxy, and utilize either a dynamic port or a configured port/protocol for each test suite.

At the beginning of each test suite, the proxy should be downloaded to `.proxy` at repo root, then utilized from there. If each test suite uses it's _own proxy and port_,  then a session-level sanitizer CANNOT stomp on another test run. They'll be totally split from each other!

benbp:
This update brings the go test recording behavior in line with other languages by removing the need to manually install/run the test proxy, and resolves some multi-module test run conflicts we had been seeing with single server test proxy sessions. See https://github.com/Azure/azure-sdk-tools/issues/6483#issuecomment-1629899174 for a description of the issue.

Now on test start we will check if the latest proxy version is downloaded, install if missing, and then run on a random port local to the test process.

Fixes #21129


sdk/internal/recording/recording.go
```diff
@@ -524,14 +524,16 @@ var client = http.Client{
...
 	GroupForReplace string
 	Variables       map[string]interface{}
 	TestInstance    *testing.T
 }
 
 func defaultOptions() *RecordingOptions {
 	return &RecordingOptions{
-		UseHTTPS: true,
+		UseHTTPS:  true,
+		ProxyPort: os.Getpid()%10000 + 20000,
```
scbedd, line 536: ohhh this is smart. They're each individual process ids so they _can't_ stomp on each other.

benbp: There could still be a port conflict that may be worth checking. @scbedd the other option I looked at was connecting to `localhost:0` which makes the OS assign a random port, and then closing the listener and using that as the proxy port, but this code gets called from a few places with a null value for the options, so it was more annoying to refactor everything to propagate the state of the initial port, rather than pick something deterministic like this.

sdk/internal/recording/server.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// NOTE: the lockfile will not be removed on ctrl-c during download.
+		// Go test seems to send an os.Interrupt signal on test setup completion, so if we
+		// call os.Exit(1) on ctrl-c the tests will never run. If we don't call os.Exit(1),
+		// the tests cannot be canceled.
+		// Therefore, if ctrl-c is pressed during download, the user will have to manually
+		// remove the lockfile in order to get the tests running again.
+		defer func() {
+			os.Remove(lockFile)
+			lock.Close()
```
benbp, line 162:
```suggestion
			lock.Close()
			os.Remove(lockFile)
```

sdk/internal/recording/recording.go
richardpark-msft: Does the test-proxy print out the port when you launch it? Seems like a nice to have for cases like this.

sdk/internal/recording/server.go
```diff
@@ -0,0 +1,329 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"runtime"
+	"time"
+)
+
+func getTestProxyDownloadFile() (string, error) {
+	if runtime.GOOS == "windows" {
```
richardpark-msft, line 28: Why is windows excluded from the switch statements below? Not good enough?

sdk/internal/recording/server.go
```diff
@@ -0,0 +1,329 @@
...
+	default:
+		return "", fmt.Errorf("unsupported OS/Arch combination: %s/%s", runtime.GOOS, runtime.GOARCH)
+	}
+}
+
+func extractTestProxyZip(archivePath string, outputDir string) error {
+    // Open the zip file
+    r, err := zip.OpenReader(archivePath)
+    if err != nil {
+        panic(err)
```
richardpark-msft, line 50: `return err`?

sdk/internal/recording/server.go
```diff
@@ -0,0 +1,329 @@
...
+	case runtime.GOOS == "darwin" && runtime.GOARCH == "amd64":
+		return "test-proxy-standalone-osx-x64.zip", nil
+	case runtime.GOOS == "darwin" && runtime.GOARCH == "arm64":
+		return "test-proxy-standalone-osx-arm64.zip", nil
+	default:
+		return "", fmt.Errorf("unsupported OS/Arch combination: %s/%s", runtime.GOOS, runtime.GOARCH)
+	}
+}
+
+func extractTestProxyZip(archivePath string, outputDir string) error {
```
richardpark-msft, line 46: Imagine if the test-proxy were just a single binary...

sdk/internal/recording/server.go
benbp: We don't build an ARM binary for windows, so we can just return x64 right away.

richardpark-msft: You don't _have_ to compare the arch in your case statement?

sdk/internal/recording/server.go
benbp: Ahh, missed one (copilot...)

sdk/internal/recording/server.go
benbp: It is a single binary, but we publish the release as an archive since it contains other things like debug symbols. @scbedd I wonder if we should also make the binaries themselves available on the releases page alongside the archives?

sdk/internal/recording/server.go
```diff
@@ -0,0 +1,329 @@
...
+func extractTestProxy(archivePath string, outputDir string) error {
+	if strings.HasSuffix(archivePath, ".zip") {
+		return extractTestProxyZip(archivePath, outputDir)
+	} else {
+		return extractTestProxyArchive(archivePath, outputDir)
+	}
+}
+
+func ensureTestProxyInstalled(proxyVersion string, proxyPath string, proxyDir string) error {
+	lockFile := filepath.Join(os.TempDir(), "test-proxy-install.lock")
```
richardpark-msft, line 144: It'd be good to print this in case something goes wrong and we have to manually delete it.

benbp: I print it here: https://github.com/Azure/azure-sdk-for-go/pull/21168/files/6f3a5d3f2127b1b59426da58f59919f0a083eaf0#diff-90f9ae1ef4aacd672ce8f7a44bcc77a6f8550e2321aefe9d1111896bade770d9R169

I tried to get handling for os.Interrupt working as well in case someone hits ctrl-c on the download, but the go test command seems to send that itself so it made things more complicated.

sdk/internal/recording/recording.go
benbp: Yep! The logs look like this:

> 2023/07/14 20:36:11 Test proxy version 1.0.0-dev.20230706.1 already installed
> 2023/07/14 20:36:11 Running test proxy command: /home/ben/sdk/azure-sdk-for-go/.proxy/Azure.Sdk.Tools.TestProxy start --storage-location /home/ben/sdk/azure-sdk-for-go -- --urls=https://localhost:29432
> 2023/07/14 20:36:11 Test proxy log location: /tmp/testproxy.log.uycawv
> 2023/07/14 20:36:13 Started test proxy instance (PID 19459) on https://localhost:29432
> 2023/07/14 20:36:15 Stopping test proxy instance (PID 19459) on https://localhost:29432

sdk/internal/recording/server.go
benbp: you're so right, I think originally it was written a bit differently with a helper which is why I had it this way. Updated to just use the switch.

sdk/internal/recording/server.go
scbedd: I think that's a bit of work for not very much benefit.

sdk/internal/recording/server.go
benbp: I also updated to print the lock file we're waiting on at the start, in case people ctrl-c out thinking it's hung.

benbp:
@chlowell @jhendrixMSFT let me know if you want to put any eyes on this.

sdk/internal/recording/README.md
```diff
@@ -16,9 +16,33 @@ After you've set the `AZURE_RECORD_MODE`, set the `PROXY_CERT` environment varia
...
+	proxy, err := recording.StartTestProxy(nil)
+	if err != nil {
+		panic(err)
+	}
+
+    ... all other test code, including proxy recording setup ...
+
+	code := m.Run()
+
+	recording.StopTestProxy(proxy)
```
chlowell, line 34:
```suggestion
	err = recording.StopTestProxy(proxy)
```

chlowell (approved):

PR #21169, opened 2023-07-14
[azopenai] Updating to the 2023-07-01 API surface

### LINKED ISSUE #21086: [azopenai] Azure OpenAI first entry is empty when streaming chat completions or completions
Using `2023-06-01-preview`, the first entry that comes back from the _service_ is empty. Skipping it allows things to work but this isn't needed for OpenAI, just Azure OpenAI.

This has to be fixed before release.

richardpark-msft:
Changes to get things to work against the latest 2023-07-01 surface:
- Adding in functions support and example.
- Added in accomodation for content filtering info. 
- Make it so we can use separate service instances for some tests so we can test against the latest upcoming fixes/changes.

Fixes #21086


richardpark-msft:
NOTE: readme update is coming in the next PR, so this is just snapping to the latest typespec revision in azure-rest-api-specs.

sdk/cognitiveservices/azopenai/custom_models.go
```diff
@@ -43,3 +51,47 @@ type ImageGenerationsDataItem struct {
...
+// ContentFilterResponseError is an error as a result of a request being filtered.
+type ContentFilterResponseError struct {
+	azcore.ResponseError
+
+	// ContentFilterResults contains Information about the content filtering category, if it has been detected.
+	ContentFilterResults *ContentFilterResults
+}
+
+// Unwrap returns the inner error for this error.
+func (e ContentFilterResponseError) Unwrap() error {
```
jhendrixMSFT, line 64: Should be pointer-to-type.

jhendrixMSFT (approved):

sdk/cognitiveservices/azopenai/custom_models.go
richardpark-msft: Done.

PR #21176, opened 2023-07-14
Add Service Labels to CODEOWNERS

JimSuplizio:
The FabricBot rule for @ mentioning people based upon the service label when the **Service Attention** label was added should have been generated from the CODEOWNERS file for each repository. As it turns out, for Go, this wasn't the case. These entries were copied from either Java, JS, Net or Python (don't know, don't care, it was incorrect either way).  github-event-processor uses the CODEOWNERS file to get this information which, because of the aforementioned FabricBot rule copy, was never actually in the CODEOWNERS file. This PR just takes the IssueRouting entries, from the FabricBot.json file that was removed from this repository at the time github-event-processor was enabled and adds them as NotInRepo entries which is the interim fix.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -65,3 +65,549 @@
 
 # PRLabel: %EngSys
 /sdk/template/                @benbp @weshaggard
+
+# ServiceLabel: %AAD
+#/<NotInRepo>/          @adamedx
+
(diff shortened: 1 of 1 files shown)


weshaggard (approved):

JimSuplizio:
Override check enforcer as there's no pipeline to actually test anything in this PR

PR #21177, opened 2023-07-14
azCosmos Global Endpoint Manager 

jay-most:
### Description
This pull request introduces the Global Endpoint Manager, which provides a centralized mechanism for managing endpoints in the Cosmos DB instance. The package includes a `globalEndpointManager` struct that handles endpoint registration, retrieval, availability tracking, and updates.

### Motivation
The current system lacks a centralized approach for managing endpoints, leading to duplicated code, inconsistencies, and difficulties in scaling and maintenance. The Global Endpoint Manager addresses these issues by providing a unified solution for managing and tracking endpoints.

### Changes Made
1. **Global Endpoint Manager**: Implemented the `globalEndpointManager` struct that serves as the central endpoint manager. It includes fields for the client reference, preferred regions, location cache, refresh ticker, and synchronization mutex.
2. **Endpoint Retrieval**: Added functions to retrieve write and read endpoints from the location cache.
3. **Account Properties**: Implemented functionality to retrieve account properties from the Cosmos DB instance.
4. **Endpoint Availability**: Added methods to mark endpoints as unavailable for read or write operations.
5. **Endpoint Updates**: Implemented functions to update the location cache and the client's default endpoint based on the provided account properties.
6. **Background Refresh**: Added mechanisms to start and stop background refreshing of the location cache.
7. **Endpoint Refresh**: Implemented functionality to trigger a refresh of stale endpoints in the location cache.
8. **Endpoint Status Check**: Added a method to check if an endpoint is marked as unavailable for requested operations.
9. **Multiple Write Locations**: Implemented a function to determine if multiple write locations can be used.
10. **Test Methods**: Added comprehensive test methods to cover the newly introduced functionalities. The tests cover different scenarios, error conditions, integration testing, and concurrent access to ensure correctness and robustness.

### Benefits
1. **Centralized Endpoint Management**: The Global Endpoint Manager provides a centralized approach to managing endpoints, reducing code duplication and ensuring consistency across the system.
2. **Enhanced Maintenance**: With a unified endpoint manager, it becomes easier to update, deprecate, and maintain endpoints throughout the system.
3. **Improved Scalability**: The manager facilitates efficient scaling by distributing and load balancing requests across available endpoints.
4. **Endpoint Availability Tracking**: The manager allows tracking the availability of endpoints for read and write operations, enabling better handling of failures or temporary unavailability.
5. **Background Refresh**: The addition of background refreshing ensures that the location cache stays up-to-date and avoids stale endpoint data.

### Testing
Comprehensive unit tests have been added to cover the new functionalities introduced by the Global Endpoint Manager. Test cases include various scenarios, edge cases, error conditions, integration testing, and concurrent access to ensure the correctness, robustness, and performance of the code.


sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/url"
+	"sync"
+	"time"
+
+	azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
+)
+
+const DefaultBackgroundRefreshInterval = 5 * time.Minute
```
ealsur, line 16: Casing defines accessor in Go. This is an internal/private const, use lower camel case:

```suggestion
const globalEndpointManagerDefaultBackgroundRefreshInterval = 5 * time.Minute
```

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	client           *Client
+	preferredRegions []string
+	locationCache    *locationCache
+	refreshTicker    *time.Ticker
+	refreshDone      chan struct{}
+	sync.Mutex
+}
+
+// NewGlobalEndpointManager creates a new global endpoint manager.
+func NewGlobalEndpointManager(client *Client, preferredRegions []string) (*globalEndpointManager, error) {
```
ealsur, line 28: Same accessor comment for this. This is a constructor, but it's public, instead of internal:

```suggestion
func newGlobalEndpointManager(client *Client, preferredRegions []string) (*globalEndpointManager, error) {
```

ealsur, line 28: Consider adding the refresh interval as an input parameter, which if not set, defaults to the constant.

That way you can create Unit Tests that verify that the refresh happens but don't take 5 minutes to execute.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	properties, err := parseAccountProperties(azResponse)
+	if err != nil {
+		return accountProperties{}, fmt.Errorf("failed to parse account properties: %w", err)
+	}
+
+	return properties, nil
+}
+
+// parseAccountProperties parses the account properties from the HTTP response.
+func parseAccountProperties(azResponse *http.Response) (accountProperties, error) {
```
ealsur, line 77: Because this method creates an accountProperties object, the ideal naming convention would be:

```suggestion
func newAccountProperties(azResponse *http.Response) (accountProperties, error) {
```

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+
+	gem.stopBackgroundRefresh()
+	gem.startBackgroundRefresh(accountProperties)
+
+	endpoints, err := gem.locationCache.writeEndpoints()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve write endpoints: %w", err)
+	}
+	if len(endpoints) > 0 {
+		gem.client.endpoint = endpoints[0].String()
```
ealsur, line 115: This is not needed, the endpoint for the client should not be modified by the Global endpoint manager

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+// Update updates the location cache and the client's default endpoint based on the provided account properties.
+func (gem *globalEndpointManager) Update(accountProperties accountProperties) error {
+	if err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations); err != nil {
+		return fmt.Errorf("failed to update location cache: %w", err)
+	}
+
+	gem.stopBackgroundRefresh()
+	gem.startBackgroundRefresh(accountProperties)
+
+	endpoints, err := gem.locationCache.writeEndpoints()
```
ealsur, line 110: The code including and after this line can be removed as it's not a logic that is needed

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if len(endpoints) > 0 {
+		gem.client.endpoint = endpoints[0].String()
+	}
+
+	return nil
+}
+
+func (gem *globalEndpointManager) startBackgroundRefresh(accountProperties accountProperties) {
+	gem.Lock()
```
ealsur, line 122: @jhendrixMSFT Could you help us with your expertise in Go for this part?

The goal is to have a background process that runs at a periodic cadence, is this the right pattern? I am not sure about the need for a lock or the use of `go func()` based on the Guidelines.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Update updates the location cache and the client's default endpoint based on the provided account properties.
+func (gem *globalEndpointManager) Update(accountProperties accountProperties) error {
+	if err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations); err != nil {
+		return fmt.Errorf("failed to update location cache: %w", err)
+	}
+
+	gem.stopBackgroundRefresh()
+	gem.startBackgroundRefresh(accountProperties)
```
ealsur, line 108: The background refresh should start when the GEM is created and be independent from the update.

The goal is that the refresh is the one that:

1. Calls gem.GetAccountProperties
2. Uses the obtained value to call Update
3. Goes back to wait for the timer.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+		return fmt.Errorf("failed to retrieve write endpoints: %w", err)
+	}
+	if len(endpoints) > 0 {
+		gem.client.endpoint = endpoints[0].String()
+	}
+
+	return nil
+}
+
+func (gem *globalEndpointManager) startBackgroundRefresh(accountProperties accountProperties) {
```
ealsur, line 121: Based on this comment: https://github.com/Azure/azure-sdk-for-go/pull/21177/files#r1264096480

The signature of this function should have no input parameters, because the expectation is that the background work will issue the request (GetAccountProperties) and then use the result to call Update?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+	defer gem.Unlock()
+
+	gem.refreshTicker = time.NewTicker(DefaultBackgroundRefreshInterval)
+	gem.refreshDone = make(chan struct{})
+
+	go func() {
+		for {
+			select {
+			case <-gem.refreshTicker.C:
+				err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations)
```
ealsur, line 132: There is nothing calling GetAccountProperties to obtain a newer version of the account properties when the timer is up, shouldn't it be a call to `newAccountProperties:= gem.GetAccountProperties()` here? And then call `gem.Update(newAccountProperties)` which should internall handle the `gem.locationCache.update`?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,170 @@
...
+			case <-gem.refreshDone:
+				gem.refreshTicker.Stop()
+				close(gem.refreshDone)
+				return
+			}
+		}
+	}()
+}
+
+func (gem *globalEndpointManager) stopBackgroundRefresh() {
```
ealsur, line 145: What is the use case for stopping the background refresh?

sdk/data/azcosmos/emulator_cosmos_gem_test.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	fmt.Printf("gem: %v", gem)
+
+	t.Run("TestGetAccountProperties", func(t *testing.T) {
+		accountProperties, err := gem.GetAccountProperties()
+		assert.NoError(t, err, "GetAccountProperties should not return an error")
+		assert.NotNil(t, accountProperties, "accountProperties should not be nil")
+
+		accountProperties.EnableMultipleWriteLocations = true
+		assert.Equal(t, accountProperties.EnableMultipleWriteLocations, true, "EnableMultipleWriteLocations should be true")
```
ealsur, line 39: This is testing a flag that was set on the previous line, what is the case where this might fail?

sdk/data/azcosmos/emulator_cosmos_gem_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+		accountProperties.EnableMultipleWriteLocations = false
```
ealsur, line 41: This is testing a flag that was set on the previous line, what is the case where this might fail?

sdk/data/azcosmos/emulator_cosmos_gem_test.go
```diff
@@ -0,0 +1,236 @@
...
+		ReadRegions := accountProperties.ReadRegions
+		assert.NotNil(t, ReadRegions, "ReadRegions should not be nil")
+		assert.Len(t, ReadRegions, 2, "ReadRegions should contain 2 regions")
+
+		WriteRegion := accountProperties.WriteRegions
+		assert.NotNil(t, WriteRegion, "WriteRegion should not be nil")
+		assert.Len(t, WriteRegion, 1, "WriteRegion should contain 1 region")
+	})
+
+	t.Run("TestLocationCache", func(t *testing.T) {
```
ealsur, line 53: Location cache already has a test file, no need to add coverage here

sdk/data/azcosmos/emulator_cosmos_gem_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"os"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGlobalEndpointManager(t *testing.T) {
+	emulatorTests := newEmulatorTests(t)
```
ealsur, line 17: This is laid down as an Emulator tests, but there are cases (like multiple write locations) that really cannot be tested with the Emulator.

The Global Endpoint Manager can be tested through unit tests, without the Emulator, you can mock HTTP responses (for example, mock an HTTP response that contains multiple write and read regions and has the EnableMultipleWriteLocations property true or false) and verify that the GEM is loading it correctly.

For example: 

https://github.com/Azure/azure-sdk-for-go/blob/c39ac95a0bce6890c1f3c4403b8b0e2a5d6be2f0/sdk/data/azcosmos/cosmos_client_test.go#L564-L587

You can create mock TLS Server with an specific HTTP JSON response and use it to create a pipeline and a client to inject into the GEM constructor:

```
srv, close := mock.NewTLSServer()
defer close()
srv.AppendResponse(
	mock.WithBody("...the raw account properties json..."),
	mock.WithStatusCode(200))

pl := azruntime.NewPipeline("azcosmostest", "v1.0.0", azruntime.PipelineOptions{}, &policy.ClientOptions{Transport: srv})
client := &Client{endpoint: srv.URL(), pipeline: pl}
```

And use that to verify the methods, and even the GetAccountProperties calls.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jay-most: Ah okay

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jay-most: I see.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jay-most: The stopBackgroundRefresh method is called when you explicitly invoke it on an instance of the globalEndpointManager. It is typically used to stop the background refresh of the location cache when it is no longer needed or when you want to manually control the refresh process.

The stopBackgroundRefresh method is called within the newGlobalEndpointManager function, right after creating a new instance of the globalEndpointManager. It ensures that any existing background refresh process is stopped before starting a new one.

Additionally, the stopBackgroundRefresh method is designed to be called at any point in your application's lifecycle when you want to stop the background refresh process. For example, you might call it when shutting down your application or when specific conditions are met.

By calling stopBackgroundRefresh, you stop the periodic refresh of the location cache performed by the background goroutine, as specified by the refreshTicker. This helps conserve resources and prevent unnecessary updates to the location cache when they are not required.

sdk/data/azcosmos/emulator_cosmos_gem_test.go
jay-most: okay, do we want an emulator test at all?

ealsur: I think it's not a matter of want, but rather, would emulator tests cover all the possible scenarios? If the answer is no, then probably that clarifies if we should be using emulator or unit tests. I was able to at least find some scenarios where the emulator test would not cover it, unless you can find a way for the emulator to yield an account with multiple write locations (I personally don't know how but I'm open if that's possible).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -25,17 +25,24 @@ type globalEndpointManager struct {
...
 		client:           client,
 		preferredRegions: preferredRegions,
 		locationCache:    newLocationCache(preferredRegions, *endpoint),
 	}, nil
+	if err != nil {
+		return nil, err
+	}
+	gem.stopBackgroundRefresh()
+	gem.startBackgroundRefresh(refreshInterval)
+	defer gem.stopBackgroundRefresh()
```
ealsur, line 43: Won't this `defer` stop the background refresh when the constructor exists/returns?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	gem, err := &globalEndpointManager{
+		client:           client,
+		preferredRegions: preferredRegions,
+		locationCache:    newLocationCache(preferredRegions, *endpoint),
+	}, nil
+	if err != nil {
+		return nil, err
+	}
+	gem.stopBackgroundRefresh()
```
ealsur, line 41: What is the purpose of stopping the background refresh if it's not started?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: Sorry, this comment was resolved but the code is still the original one. How is this triggering a new HTTP request and an update with the response?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+const globalEndpointManagerDefaultBackgroundRefreshInterval = 5 * time.Minute
+
+type globalEndpointManager struct {
+	client           *Client
+	preferredRegions []string
+	locationCache    *locationCache
+	refreshTicker    *time.Ticker
+	refreshDone      chan struct{}
```
ealsur, line 23: Could you help me understand how is this `refreshDone` used?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: I put together an example gist [here](https://gist.github.com/jhendrixMSFT/0bf567852e83bff788b61d847da6fbcc).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT, line 1: As there appears to be a mix of internal and private APIs on this type, I suggest moving it to an internal package.
```
/sdk/data/azcosmos/internal/gem
```
This would provide better encapsulation, and keeps private APIs private.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,186 @@
...
+// GetAccountProperties retrieves account properties from the Cosmos DB instance.
+func (gem *globalEndpointManager) GetAccountProperties() (accountProperties, error) {
+	operationContext := pipelineRequestOptions{
+		resourceType:    resourceTypeDatabaseAccount,
+		resourceAddress: "",
+	}
+
+	path, err := generatePathForNameBased(resourceTypeDatabaseAccount, "", false)
+	if err != nil {
+		return accountProperties{}, fmt.Errorf("failed to generate path for name-based request: %w", err)
```
jhendrixMSFT, line 67: Using `%w` will wrap the error, meaning it can be unwrapped via `errors.Is()` and/or `errors.As()`.  As a result, the wrapped error is publicly accessible which is usually undesirable.  It would be better to use `%v` instead, or just remove the `failed to generate...` text altogether and jus return `err`.

Comment applies to all places where `%w` is used.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Write(azlog.EventResponse, fmt.Sprintf("Background refresh error: %v", err))
+		return
+	}
+	go func() {
+		for {
+			select {
+			case <-gem.refreshTicker.C:
+				err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations)
+				if err != nil {
+					log.Write(azlog.EventResponse, fmt.Sprintf("Background refresh error: %v", err))
```
jhendrixMSFT, line 144: Having a background goroutine silently swallow errors seems dubious.  How is a customer to know if an error occurred, and what is the recovery path?

jay-most: this is a siently error. it should not disrupt the connection rather us the lastest location cache.

jhendrixMSFT: It sounds to me that it's not really an error then and should be removed.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,186 @@
...
+			case <-gem.refreshDone:
+				gem.refreshTicker.Stop()
+				close(gem.refreshDone)
+				return
+			}
+		}
+	}()
+}
+
+func (gem *globalEndpointManager) stopBackgroundRefresh() {
```
kirankumarkolli, line 155: Background refresh lifetime needs to be tied to client. I.e. when client is released the backgroundRefresh. 

Gist sample @jhendrixMSFT shared passing context to manage/control lifetime, is context passing from client needed here?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
kirankumarkolli: On new instance refresh background task is not yet started, its not necessary right?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: I'm still unclear on the value of this.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+					return
+				}
+				return
+			}
+		}
+	}()
+	cancel()
+
+	if err := <-errChan; err != nil {
+		panic(err)
```
jhendrixMSFT, line 146: We shouldn't be explicitly calling `panic` as this will terminate the user's process.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+				err := gem.Update()
+				if err != nil {
+					log.Write(azlog.EventResponse, fmt.Sprintf("Failed to update location cache: %v", err))
+					return
+				}
+				return
+			}
+		}
+	}()
+	cancel()
```
jhendrixMSFT, line 143: This is going to cause your refresh goroutine to terminate right away.  Either the `context.Context` should be a parameter to `startBackgroundRefresh`, or you use a channel on `gem` to control the lifetime (that might be simpler).  Let me know if you have questions on that.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
+package azcosmos
```
jhendrixMSFT, line 1: File is missing the MIT license header.  You can copy it from [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/core.go#L4-L5) (note the newline between the header text and the package declaration).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: Looks like this isn't feasible as gem has other dependencies (Client and locationCache).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+				err := gem.Update()
+				if err != nil {
+					return
+				}
+				return
+			case <-time.After(refreshInterval):
+				err := gem.Update()
+				if err != nil {
+					log.Write(azlog.EventResponse, fmt.Sprintf("Failed to update location cache: %v", err))
+					return
```
jhendrixMSFT, line 137: This will terminate the refresh gorouotine.  What's the impact?  How does it get restarted?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+		return fmt.Errorf("failed to retrieve account properties: %v", err)
+	}
+
+	if err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations); err != nil {
+		return fmt.Errorf("failed to update location cache: %v", err)
+	}
+	return nil
+}
+
+func (gem *globalEndpointManager) startBackgroundRefresh(refreshInterval time.Duration) {
```
jhendrixMSFT, line 119: I think it would help to have a comment explaining the lifetime of the background refresh goroutine.  When does it start?  When should it stop?  If it stops, does it automatically get restarted?  etc...

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: It's unused and is currently being flagged by the linter.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, err
+	}
+	gem.startBackgroundRefresh(refreshInterval)
+
+	return gem, nil
+}
+
+// GetWriteEndpoints returns the write endpoints from the location cache.
+func (gem *globalEndpointManager) GetWriteEndpoints() ([]url.URL, error) {
```
jhendrixMSFT, line 47: Since we can't move `globalEndpointManager` to an internal package, we should use methods that start with an uppercase letter to denote APIs that are expected to be called outside of `gem`.  Methods that are internal to `gem` should start with a lowercase letter.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	errChan := make(chan error)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				err := gem.Update()
+				if err != nil {
```
jhendrixMSFT, line 129: You can simplify this all to just the following.
```go
_ = gem.Update()
return
```

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	client           *Client
+	preferredRegions []string
+	locationCache    *locationCache
+	refreshTicker    *time.Ticker
+	refreshDone      chan struct{}
+	sync.Mutex
+}
+
+// NewGlobalEndpointManager creates a new global endpoint manager.
+func newGlobalEndpointManager(client *Client, preferredRegions []string, refreshInterval time.Duration) (*globalEndpointManager, error) {
```
jhendrixMSFT, line 28: From where will this be called?  I think we need to know the answer to that in order to help control the lifetime of the refresh goroutine.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	if err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations); err != nil {
+		return fmt.Errorf("failed to update location cache: %v", err)
+	}
+	return nil
+}
+
+func (gem *globalEndpointManager) startBackgroundRefresh(refreshInterval time.Duration) {
+	errChan := make(chan error)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
jhendrixMSFT, line 122: The cancellation mechanism will need to come from outside this method.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+		resourceType:    resourceTypeDatabaseAccount,
+		resourceAddress: "",
+	}
+
+	path, err := generatePathForNameBased(resourceTypeDatabaseAccount, "", false)
+	if err != nil {
+		return accountProperties{}, fmt.Errorf("failed to generate path for name-based request: %v", err)
+	}
+
+	azResponse, err := gem.client.sendGetRequest(path, context.Background(), operationContext, nil, nil)
```
jhendrixMSFT, line 68: For internal I/O calls, it's best practice to use a context with a timeout.  This is to prevent inadvertent hangs due to I/O failure.  Something like this (I just made up the 10 seconds, use whatever value is appropriate).
```go
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
azResponse, err := gem.client.sendGetRequest(path, ctx, operationContext, nil, nil)
cancel()
```

jhendrixMSFT:
I have some concerns WRT managing the lifetime of the refresh goroutine.  Since there is nothing like the `IDisposable` pattern in Go, callers will need to explicitly cancel the background refresh when they're done with a client (there might be tricks we can do with finalization, but I'd prefer to avoid that if possible).  The typical pattern is to return a cancel function (like [context.WithCancel](https://pkg.go.dev/context#WithCancel) does).  It means that Cosmos client constructors would end up looking like this.
```go
client, cancelFn, err := azcosmos.NewClient(...)
// do stuff with client
cancelFn()
```
While this provides a clear contract on cancellation, it's a bit clunky and looks weird compared to other SDK clients.

Another option would be to add an API to the `azcosmos.Client` type.
```go
client, err := azcosmos.NewClient(...)
// do stuff with client
client.Close()
```
Not as clear as the first, but easier to reason about and is pretty typical.  Callers just need to ensure that they don't forget to call `Close()` else they'll leak goroutines.  This can likely be mitigated with good docs and examples.

But I wonder if we really need a background refresh at all?  What if the client maintained a timestamp of the last refresh, and if the interval has elapsed during an API call, then `gem` is refreshed?  We do something similar for how AAD token credentials are refreshed on a client.  We might even be able to do this as a pipeline policy but would need some experimentation.

As an aside, what's the behavior if a client API call is made while a refresh is in progress?  Does it block until the refresh completes, or does it use the stale values?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+
+const globalEndpointManagerDefaultBackgroundRefreshInterval = 5 * time.Minute
+
+type globalEndpointManager struct {
+	client           *Client
+	preferredRegions []string
+	locationCache    *locationCache
+	refreshTicker    *time.Ticker
+	refreshDone      chan struct{}
+	sync.Mutex
```
kirankumarkolli, line 24: New bee clarification: How is sync.Mutex used? It it needs to be named?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
kirankumarkolli: Same for refreshTicker too right?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+func (gem *globalEndpointManager) startBackgroundRefresh(refreshInterval time.Duration) {
+	errChan := make(chan error)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				err := gem.Update()
```
kirankumarkolli, line 128: Can you please share context on why call to gem.Update() is needed before dispose/closing of GlobalEndpointManager?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
kirankumarkolli: As discussed the goal is to keep retrying looking for changes to account even after failures. its lifetime needs to be closely tied to ComsosClient lifetime.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+					return
+				}
+				return
+			case <-time.After(refreshInterval):
+				err := gem.Update()
+				if err != nil {
+					log.Write(azlog.EventResponse, fmt.Sprintf("Failed to update location cache: %v", err))
+					return
+				}
+				return
```
kirankumarkolli, line 139: Won't return here also terminate the refresh go-routine?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+			}
+		}
+	}()
+	cancel()
+
+	if err := <-errChan; err != nil {
```
kirankumarkolli, line 145: Isn't it blocking the caller? 

Scenario requirement is to run the refreshTask in background/asynchronously till the lifetime of the CosmosClient. It might be related to the comment from @jhendrixMSFT  https://github.com/Azure/azure-sdk-for-go/pull/21177/files#r1269993978

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+// MarkEndpointUnavailableForWrite marks an endpoint as unavailable for write operations.
+func (gem *globalEndpointManager) MarkEndpointUnavailableForWrite(endpoint url.URL) error {
+	return gem.locationCache.markEndpointUnavailableForWrite(endpoint)
+}
+
+// Update updates the location cache and the client's default endpoint based on the provided account properties.
+func (gem *globalEndpointManager) Update() error {
+	accountProperties, err := gem.GetAccountProperties()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve account properties: %v", err)
```
kirankumarkolli, line 110: @jhendrixMSFT what's the recommended pattern to return errors? 

In this specific case all the error context is from gem.GetAccountProperties() and above fmt.Errorf is not adding more context.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	if err := gem.locationCache.update(accountProperties.WriteRegions, accountProperties.ReadRegions, gem.preferredRegions, &accountProperties.EnableMultipleWriteLocations); err != nil {
```
kirankumarkolli, line 113: So-far the pattern is assignment and evaluation (if below) are done in two separate lines. 
Its does improve the code readability.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	return gem.locationCache.markEndpointUnavailableForRead(endpoint)
+}
+
+// MarkEndpointUnavailableForWrite marks an endpoint as unavailable for write operations.
+func (gem *globalEndpointManager) MarkEndpointUnavailableForWrite(endpoint url.URL) error {
+	return gem.locationCache.markEndpointUnavailableForWrite(endpoint)
+}
+
+// Update updates the location cache and the client's default endpoint based on the provided account properties.
+func (gem *globalEndpointManager) Update() error {
```
kirankumarkolli, line 107: Is this the final form the caller (ex: ClientRetryPolicy) is expected to use? 
If not does it need to be public? (Please correct me if I am wrong that capitalization is public)

kirankumarkolli: Number of times Update reads accounts info needs to be limited/throttles. Unbounded will impact Gateway service. 

Please refer: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5ec7b4bcd7e2e0698736cd374f2196682227c913/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs#L574

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return properties, nil
+}
+
+// newAccountProperties parses the account properties from the HTTP response.
+func newAccountProperties(azResponse *http.Response) (accountProperties, error) {
+	properties := accountProperties{}
+	err := azruntime.UnmarshalAsJSON(azResponse, &properties)
+	if err != nil {
```
kirankumarkolli, line 85: Is if block needed? 
Lines 85 & 88 can be removed right?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	properties, err := newAccountProperties(azResponse)
+	if err != nil {
+		return accountProperties{}, fmt.Errorf("failed to parse account properties: %v", err)
+	}
+
+	return properties, nil
+}
+
+// newAccountProperties parses the account properties from the HTTP response.
+func newAccountProperties(azResponse *http.Response) (accountProperties, error) {
```
kirankumarkolli, line 82: error as argument is not used.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	return gem.locationCache.writeEndpoints()
+}
+
+// GetReadEndpoints returns the read endpoints from the location cache.
+func (gem *globalEndpointManager) GetReadEndpoints() ([]url.URL, error) {
+	return gem.locationCache.readEndpoints()
+}
+
+// GetAccountProperties retrieves account properties from the Cosmos DB instance.
+func (gem *globalEndpointManager) GetAccountProperties() (accountProperties, error) {
```
kirankumarkolli, line 57: Captial case => pubic right? If so is it expected?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
kirankumarkolli: @jay-most please cover expected interactions with GlobalEnpointManager part of PR description. Few ones of interest
- How is GlobalEnpointManager created and lifetime managed
- How is GlobalEnpointManager leveraged in request-flows 
> - First region choice of routing 
> - How are regions status managed/maintained
> - On-demand account refresh scenarios

kirankumarkolli:
> I have some concerns WRT managing the lifetime of the refresh goroutine. Since there is nothing like the `IDisposable` pattern in Go, callers will need to explicitly cancel the background refresh when they're done with a client (there might be tricks we can do with finalization, but I'd prefer to avoid that if possible). The typical pattern is to return a cancel function (like [context.WithCancel](https://pkg.go.dev/context#WithCancel) does). It means that Cosmos client constructors would end up looking like this.
> 
> ```go
> client, cancelFn, err := azcosmos.NewClient(...)
> // do stuff with client
> cancelFn()
> ```
> 
> While this provides a clear contract on cancellation, it's a bit clunky and looks weird compared to other SDK client constructors.
> 
> Another option would be to add an API to the `azcosmos.Client` type.
> 
> ```go
> client, err := azcosmos.NewClient(...)
> // do stuff with client
> client.Close()
> ```
> 
> The contract isn't as clear as the first, but easier to reason about and is pretty typical. Callers just need to ensure that they don't forget to call `Close()` else they'll leak a goroutine. This can likely be mitigated with good docs and examples.
> 
> But I wonder if we really need a background refresh at all? What if the client maintained a timestamp of the last refresh, and if the interval has elapsed during an API call, then `gem` is refreshed? We do something similar for how AAD token credentials are refreshed on a client. We might even be able to do this as a pipeline policy but would need some experimentation.
> 
> As an aside, what's the behavior if a client API call is made while a refresh is in progress? Does it block until the refresh completes, or does it use the stale values?

Great points @jhendrixMSFT. 

Common understanding on the GlobalEndpoint usage will help for this conversation Covered it part of the comment  https://github.com/Azure/azure-sdk-for-go/pull/21177#discussion_r1271225013

@jay-most lets please connect and go-over it once above comment is addressed.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: I was discussing this with Jeff and clients must _not_ have a close method.  This also goes against our client [design guidelines](https://azure.github.io/azure-sdk/golang_introduction.html#client-immutability) that state clients must not create any goroutines (this point isn't very clear, we will get this fixed).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: In this case, since the type isn't exported, the method also isn't exported.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: By default, we should not return errors with extra wording unless it will improve the ability to diagnose the error (e.g. disambiguate cases where the same error might be returned from different branches).

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -23,6 +26,9 @@ func TestGlobalEndpointManager_GetWriteEndpoints(t *testing.T) {
 	verifier := headerPoliciesVerify{}
 	pl := azruntime.NewPipeline("azcosmostest", "v1.0.0", azruntime.PipelineOptions{PerCall: []policy.Policy{headerPolicy, &verifier}}, &policy.ClientOptions{Transport: srv})
 	req, err := azruntime.NewRequest(context.Background(), http.MethodGet, srv.URL())
+	if err != nil {
+		return
+	}
```
Pilchie, line 31: Should these blocks in the tests actually just be `assert.NoError(t, err)`?  Otherwise, the tests will appear to pass if there is an error, right?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,171 @@
...
+			// If the context is canceled (e.g., due to stopBackgroundRefresh being called), perform an update
+			// to refresh the location cache and cancel the goroutine.
+			case <-ctx.Done():
+				_ = gem.Update() // Perform the update to refresh the location cache.
+				cancel()         // Cancel the context to stop the goroutine.
+				return           // Return from the goroutine.
+			// After the specified refreshInterval, perform an update to refresh the location cache and cancel the goroutine.
+			case <-time.After(refreshInterval):
+				_ = gem.Update() // Perform the update to refresh the location cache.
+				return           // Return from the goroutine.
```
kirankumarkolli, line 152: Assuming 'return' is to exit the go-function won't it stop background refresh?

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,378 @@
...
+
+	err = gem.MarkEndpointUnavailableForRead(*endpoint)
+	assert.NoError(t, err)
+
+	// Check if the endpoint is marked as unavailable for read
+	unavailable := gem.IsEndpointUnavailable(*endpoint, 1)
+	assert.True(t, unavailable)
```
kirankumarkolli, line 202: Also validate that unavailable region is excluded from readEndpoints afterwards.

kirankumarkolli: Same for below TestGlobalEndpointManager_MarkEndpointUnavailableForWrite test

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,378 @@
...
+	// Create a new global endpoint manager with preferred regions and refresh interval
+	gem, err := newGlobalEndpointManager(mockClient, []string{"westus", "eastus"}, 5*time.Minute)
+	assert.NoError(t, err)
+
+	// Check if multiple write locations can be used
+	canUseMultipleWriteLocs := gem.CanUseMultipleWriteLocations()
+	assert.False(t, canUseMultipleWriteLocs)
+}
+
+func TestGlobalEndpointManagerBackgroundRefresh(t *testing.T) {
```
kirankumarkolli, line 351: Is the test expected to validate that backgroundRefresh is up and running in background and doing account refreshes? 

If so the test needs to validate below
- background refresh continuing even after current refresh failure
- Between retries refreshInterval

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
Pilchie: This doesn't seem to be fixed?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: The goroutine lifetime is tied to the client lifetime, it should run until the client instance is destroyed/disposed.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: Given that the lifetime of the goroutine is the client lifetime, I guess then that the cancellation should be passed as an input that is tied with the client instance lifetime (this is not controlled by users)?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: In Cosmos SDKs (this is common to all languages), we have a background task that every 5 minutes refreshes the account information. The goal is to proactively detect any changes in account information (Available Regions, Consistency) that will affect how the client routes requests.

One common scenario is: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-sdk-availability#adding-a-region-to-an-account

If for whatever reason the refresh fails, it is not critical. Depending on the error, we might decide to do a retry (for example, for timeouts) or in some cases, we can even make it better and use other regional endpoints (there is a global DNS and there are regional endpoints with a different DNS each).

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: We met last week to discuss this, and we don't want clients to have the need to be closed.  The current thinking is to use a pipeline policy that will start the refresh during an API call if it's been 5 minutes from the last refresh.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: error is not an argument but rather a return type, in Go `(accountProperties, error)` means the function either returns an instance of accountProperties or an error (there is no throwing exceptions), so the caller needs to inspect the `error`.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
jhendrixMSFT: There needs to be a space between the license header comment and the package name definition else the license comment shows up as a doc comment which we don't want.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,173 @@
...
+	if refreshInterval == 0 {
+		refreshInterval = defaultBackgroundRefreshInterval
+	}
+
+	gem, err := &globalEndpointManager{
+		client:           client,
+		preferredRegions: preferredRegions,
+		locationCache:    newLocationCache(preferredRegions, *endpoint),
+	}, nil
+	if err != nil {
```
ealsur, line 44: How is `err != nil` if it's being set to `nil` on the previous line? Should this line be removed?

sdk/data/azcosmos/cosmos_global_endpoint_policy.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// globalEndpointManagerPolicies contains policies related to caching and the global endpoint manager.
+type globalEndpointManagerPolicies struct {
+	endpointMgr *globalEndpointManager // endpointMgr is a pointer to the global endpoint manager instance.
+}
+
+// NewGlobalEndpointManagerPolicies creates a new instance of globalEndpointManagerPolicies.
+func NewGlobalEndpointManagerPolicies(endpointMgr *globalEndpointManager) *globalEndpointManagerPolicies {
```
ealsur, line 19: This should be lowe camel case, otherwise it's public.

```suggestion
func newGlobalEndpointManagerPolicy(endpointMgr *globalEndpointManager) *globalEndpointManagerPolicies {
```

sdk/data/azcosmos/cosmos_global_endpoint_policy.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// globalEndpointManagerPolicies contains policies related to caching and the global endpoint manager.
+type globalEndpointManagerPolicies struct {
```
ealsur, line 14: Regarding naming: Should this be singular (policy)?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: Got it. In this case then, we should remove the code for the background refresh

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,173 @@
...
+	return gem.locationCache.markEndpointUnavailableForRead(endpoint)
+}
+
+// MarkEndpointUnavailableForWrite marks an endpoint as unavailable for write operations.
+func (gem *globalEndpointManager) MarkEndpointUnavailableForWrite(endpoint url.URL) error {
+	return gem.locationCache.markEndpointUnavailableForWrite(endpoint)
+}
+
+// Update updates the location cache and the client's default endpoint based on the provided account properties.
+func (gem *globalEndpointManager) Update() error {
```
ealsur, line 116: If we move forward with the Policy approach, this method needs to be guarded for concurrency. Because there could be multiple requests going through the Policy, detecting an old account information, and triggering an update concurrently.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: This is not public. The public/private/internal is defined by the **type**. In this case, the type is `globalEndpointManager`, which is not public. The casing of this method is correct.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: > Please correct me if I am wrong that capitalization is public
Capitalization is correct. Public/private is defined at the type level, in this case `globalEndpointManager`, which is internal/private.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,173 @@
...
+		client:           client,
+		preferredRegions: preferredRegions,
+		locationCache:    newLocationCache(preferredRegions, *endpoint),
+	}, nil
+	if err != nil {
+		return nil, err
+	}
+
+	// Start the background refresh process with the specified refreshInterval.
+	gem.startBackgroundRefresh(refreshInterval)
```
ealsur, line 49: Background refresh should be removed after the last meetings we had where we decided this will be done through the Update call from the policy.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
ealsur: After our offline sync: The background refresh should be removed.

PR #21197, opened 2023-07-18
Filter Tags API on Container

siminsavani-msft:

### DIFF
--- a/sdk/storage/azblob/container/client.go
+++ b/sdk/storage/azblob/container/client.go
@@ -423,7 +423,7 @@ func (c *Client) SubmitBatch(ctx context.Context, bb *BatchBuilder, options *Sub
 }
 
 // FilterBlobs operation finds all blobs in the container whose tags match a given search expression.
-// https://docs.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags
+// https://docs.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags-container
 // eg. "dog='germanshepherd' and penguin='emperorpenguin'"
 func (c *Client) FilterBlobs(ctx context.Context, where string, o *FilterBlobsOptions) (FilterBlobsResponse, error) {
 	containerClientFilterBlobsOptions := o.format()
--- a/sdk/storage/azblob/container/client.go
+++ b/sdk/storage/azblob/container/client.go
@@ -422,11 +422,9 @@ func (c *Client) SubmitBatch(ctx context.Context, bb *BatchBuilder, options *Sub
 	}, nil
 }
 
-// FilterBlobs operation finds all blobs in the storage account whose tags match a given search expression.
-// Filter blobs searches across all containers within a storage account but can be scoped within the expression to a single container.
+// FilterBlobs operation finds all blobs in the container whose tags match a given search expression.
 // https://docs.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags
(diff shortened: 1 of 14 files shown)


sdk/storage/azblob/container/client.go
```diff
@@ -421,3 +421,12 @@ func (c *Client) SubmitBatch(ctx context.Context, bb *BatchBuilder, options *Sub
 		Version:     resp.Version,
 	}, nil
 }
+
+// FilterBlobs operation finds all blobs in the container whose tags match a given search expression.
+// https://docs.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags
```
souravgupta-msft, line 426: The URL needs to be corrected.
```suggestion
// https://docs.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags-container
```

sdk/storage/azblob/container/models.go
```diff
@@ -397,3 +397,31 @@ type SubmitBatchOptions struct {
...
+	// operation did not return all containers remaining to be listed with the current page. The NextMarker value can be used
+	// as the value for the marker parameter in a subsequent call to request the next
+	// page of list items. The marker value is opaque to the client.
+	Marker *string
+	// Specifies the maximum number of containers to return. If the request does not specify maxresults, or specifies a value
+	// greater than 5000, the server will return up to 5000 items. Note that if the
+	// listing operation crosses a partition boundary, then the service will return a continuation token for retrieving the remainder
+	// of the results. For this reason, it is possible that the service will
+	// return fewer results than specified by maxresults, or than the default of 5000.
+	MaxResults *int32
```
souravgupta-msft, line 416: `Include` should also be added in the options?
https://github.com/Azure/azure-sdk-for-go/blob/feature/azblob-stg85/86/sdk/storage/azblob/internal/generated/zz_models.go#L878-L880

sdk/storage/azblob/container/models.go
```diff
@@ -397,3 +397,31 @@ type SubmitBatchOptions struct {
 func (o *SubmitBatchOptions) format() *generated.ContainerClientSubmitBatchOptions {
 	return nil
 }
+
+// ---------------------------------------------------------------------------------------------------------------------
+
+// FilterBlobsOptions provides set of options for Client.FindBlobsByTags.
```
souravgupta-msft, line 403:
```suggestion
// FilterBlobsOptions provides set of options for Client.FilterBlobs.
```

sdk/storage/azblob/container/models.go
seanmcc-msft: Is the link to the auto-generated internal type?  I think there might be a typo in the swagger, the "include" parameter is not in the public rest docs - https://learn.microsoft.com/en-us/rest/api/storageservices/find-blobs-by-tags-container?tabs=azure-ad

souravgupta-msft: The link is to the auto-generated internal type where `Include` is present in the options. The [request creation code](https://github.com/Azure/azure-sdk-for-go/blob/feature/azblob-stg85/86/sdk/storage/azblob/internal/generated/zz_container_client.go#L515-L517) also has this in the query parameter. 
If it is a typo in the swagger, then this comment can be resolved.

siminsavani-msft: I believe there is an error in the swagger because the documentation doesn't include `Include` and it is not present in the the .NET or Python SDK. Will resolve this comment!

souravgupta-msft (approved):

PR #21219, opened 2023-07-19
AzureCLICredential and OnBehalfOfCredential return errors immediately on failure

chlowell:
While refactoring `syncer` to add an internal options bag for an upcoming feature, I noticed a bug that would cause `AzureCLICredential.GetToken()` and `OnBehalfOfCredential.GetToken()` to make a second authentication attempt immediately after their first attempt fails. They should instead return an error because there's no reason to believe a second attempt would succeed and in the case of the `AzureCLICredential`, trying again is particularly expensive.

### DIFF
--- a/sdk/azidentity/azure_cli_credential_test.go
+++ b/sdk/azidentity/azure_cli_credential_test.go
@@ -29,6 +29,28 @@ var (
 	}
 )
 
+func TestAzureCLICredential_Error(t *testing.T) {
+	// GetToken shouldn't invoke the CLI a second time after a failure
+	authNs := 0
+	o := AzureCLICredentialOptions{
+		tokenProvider: func(context.Context, string, string) ([]byte, error) {
+			authNs++
+			return nil, errors.New("it didn't work")
+		},
+	}
+	cred, err := NewAzureCLICredential(&o)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = cred.GetToken(context.Background(), testTRO)
+	if err == nil {
+		t.Fatal("expected an error")
+	}
+	if authNs != 1 {
+		t.Fatalf("expected 1 authN, got %d", authNs)
+	}
+}
+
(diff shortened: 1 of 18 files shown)


sdk/azidentity/syncer.go
```diff
@@ -41,21 +46,24 @@ func newSyncer(name, tenant string, additionalTenants []string, reqToken, silent
...
+			return at, err
 		}
 		opts.TenantID = tenant
 	}
-	at, err := s.silent(ctx, opts)
-	if err != nil {
+	var err error
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.silent == nil {
```
jhendrixMSFT, line 64: This is the magic right?  Since CLI and OBO credentials don't populate `s.silent` they only attempt once.

chlowell: Right. Previously, `s.silent` and `s.reqToken` were the same func for these credentials, so when `silent` returned an error like "you need to log in to the CLI", we'd waste time doing the same thing again.

jhendrixMSFT (approved):

sdk/azidentity/azure_cli_credential_test.go
```diff
@@ -29,6 +29,28 @@ var (
...
+	authNs := 0
+	o := AzureCLICredentialOptions{
+		tokenProvider: func(context.Context, string, string) ([]byte, error) {
+			authNs++
+			return nil, errors.New("it didn't work")
+		},
+	}
+	cred, err := NewAzureCLICredential(&o)
+	if err != nil {
+		t.Fatal(err)
```
RickWinter, line 43: Output that we were expecting a failure but got success

RickWinter (approved):

sdk/azidentity/azure_cli_credential_test.go
chlowell: This error should be nil because it's from the constructor

PR #21222, opened 2023-07-19
[AzDatalake] Cleanup

tasherif-msft:


sdk/storage/azdatalake/common.go
```diff
@@ -34,3 +35,47 @@ func ParseURL(u string) (URLParts, error) {
 // ending at offset+count. A zero-value HTTPRange indicates the entire resource. An HTTPRange
 // which has an offset but no zero value count indicates from the offset to the resource's end.
 type HTTPRange = exported.HTTPRange
+
+// ===================================== LEASE CONSTANTS ============================================================
+
+// StatusType defines values for StatusType
+type StatusType = lease.StatusType
```
tasherif-msft, line 42: moved to common since this applies to all clients

vibhansa-msft: There was a lease client in AzBlob. If we have something similar here this shall be part of that.

vibhansa-msft (approved):

souravgupta-msft (approved):

sdk/storage/azdatalake/common.go
tasherif-msft: circular dependency :/

PR #21223, opened 2023-07-19
[azopenai] Make the deployment ID a per-request parameter, rather than a constructor parameter.

richardpark-msft:
A rather big change.
    
Prior to this we required the deployment ID to be passed as a parameter when constructing the client. This had some advantages in that it matched the underlying physical properties for Azure (it's a route parameter, not a request parameter), but it broke the de-facto conventions established by OpenAI, where the `model` (ex: "gpt-4") is considered a request parameter.
    
This is now changed, with the one concession being that the `Model` field has been renamed to `DeploymentID`, to match with the other Azure SDKs.


sdk/cognitiveservices/azopenai/custom_client.go
```diff
@@ -239,7 +239,21 @@ func (client *Client) newError(resp *http.Response) error {
...
 
 type clientData struct {
-	endpoint     string
-	baseEndpoint string
-	azure        bool
+	endpoint string
+	azure    bool
+}
+
+func getDeploymentID[T ChatCompletionsOptions | CompletionsOptions | EmbeddingsOptions | ImageGenerationOptions](v T) string {
```
JeffreyRichter, line 246: This is a clever way of accomplishing this!

sdk/cognitiveservices/azopenai/custom_client.go
```diff
@@ -239,7 +239,21 @@ func (client *Client) newError(resp *http.Response) error {
...
+	switch a := any(v).(type) {
+	case ChatCompletionsOptions:
+		return a.DeploymentID
+	case CompletionsOptions:
+		return a.DeploymentID
+	case EmbeddingsOptions:
+		return a.DeploymentID
+	case ImageGenerationOptions:
+		return ""
+	default:
```
JeffreyRichter, line 256: You shouldn't need this.

sdk/cognitiveservices/azopenai/models.go
```diff
@@ -237,9 +237,8 @@ type ChatCompletionsOptions struct {
 	// The maximum number of tokens to generate.
 	MaxTokens *int32
 
-	// The model name to provide as part of this completions request. Not applicable to Azure OpenAI, where deployment information
-	// should be included in the Azure resource URI that's connected to.
-	Model *string
+	// DeploymentID specifies the name of the deployment (for Azure OpenAI) or model (for OpenAI) to use for this request.
+	DeploymentID string
```
JeffreyRichter, line 241: I guess the lack of * is OK here since it really becomes mandatory. But maybe not for images?

JeffreyRichter (approved):

sdk/cognitiveservices/azopenai/models.go
jhendrixMSFT: Presumably it's because the empty string and `nil` are equivalent in this context?

jhendrixMSFT (approved):

sdk/cognitiveservices/azopenai/custom_client.go
richardpark-msft: I finally got to use generics :)

sdk/cognitiveservices/azopenai/models.go
richardpark-msft: Images doesn't have the model/deploymentID parameter, so we're safe there.

It's a required parameter, so you get bad request from OpenAI and deployment not found from Azure OpenAI if it's empty. TBH, we might want to check this on the client side but the service-side errors aren't horrible either.

sdk/cognitiveservices/azopenai/custom_client.go
richardpark-msft: I wish I had a way to enforce that - the problem is that I can't type switch on the generic `T`, so I have to cast to `any` and do the type switch on that, which means that there's no checking that I haven't fallen through a default path.

I could change this to also return an 'err', which wouldn't be horrible, but this was just something I was trying to do to be expedient. :|

sdk/cognitiveservices/azopenai/models.go
jhendrixMSFT: Oh this is the options type that isn't our synthesized options type right?  Does this need a comment that it's required for OpenAI?

Issue #21048, opened 2023-06-23
Escape single quotes on partition/row keys


davidandradeduarte:
### Bug Report

**Reported version:**
github.com/Azure/azure-sdk-for-go/sdk/data/aztables v1.0.1

The current table storage SDK doesn't escape single quotes in partition and row keys. This means an API Bad Request whenever we use single quotes in these fields.

I would expect that the API could handle this, instead of clients escaping every string they send to PK/RK, because most times you wouldn't know.

I can try to raise a PR to fix this.

Here's a repro: https://github.com/davidandradeduarte/azurite-escape-odata-bug (this was initially meant to test azurite, but you can also reproduce the issue with a real storage account)

PR #21232, opened 2023-07-20
Escape single-quote characters in partition and row keys

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/21048


sdk/data/aztables/client_test.go
```diff
@@ -216,7 +217,7 @@ func TestInsertEntity(t *testing.T) {
 			defer delete()
 
 			// 1. Create Basic Entity
-			entityToCreate := createSimpleEntity(1, "partition")
+			entityToCreate := createSimpleEntityWithRowKey(1, "parti'tion", "one'")
```
RickWinter, line 220: Lets beef up the test suite around  this.  I know the current fix makes it moot, but if that impl ever changed it would be good to know we have a strong escaping suite.

Some ideas:
Single quote at beginning of key/values
Single quote at end of key/values
Multiple single quotes in a row
Single quote only

jhendrixMSFT: Added tests (thanks for reminding me).

PR #21235, opened 2023-07-20
azidentity test cleanup

chlowell:
Simplifying mocking, improving manual test output

### DIFF
--- a/sdk/azidentity/client_certificate_credential_test.go
+++ b/sdk/azidentity/client_certificate_credential_test.go
@@ -17,7 +17,6 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
 )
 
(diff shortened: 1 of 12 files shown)


sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -165,14 +157,8 @@ func TestClientCertificateCredential_NoCertificate(t *testing.T) {
...
 	test := allCertTests[0]
-	srv, close := mock.NewTLSServer()
-	defer close()
-	srv.AppendResponse(mock.WithBody(accessTokenRespSuccess))
-	options := ClientCertificateCredentialOptions{}
-	options.Cloud.ActiveDirectoryAuthorityHost = srv.URL()
-	options.Transport = srv
 	var key crypto.PrivateKey
-	_, err := NewClientCertificateCredential(fakeTenantID, fakeClientID, test.certs, key, &options)
+	_, err := NewClientCertificateCredential(fakeTenantID, fakeClientID, test.certs, key, nil)
```
jhendrixMSFT, line 161: How come this one doesn't need a `&mockSTS{}` in the options?

jhendrixMSFT: Oh perhaps because the inputs cause the constructor to fail.

jhendrixMSFT (approved):

sdk/azidentity/client_certificate_credential_test.go
chlowell: The constructor doesn't do any I/O and should return an error when the private key is nil

PR #21238, opened 2023-07-20
Update aztables to use latest code generator

jhendrixMSFT:


RickWinter:
Can you add some context to the description of this PR.  Why are we switching to the zz_ naming pattern?

jhendrixMSFT:
The `zz_` prefix is used to distinguish generated code from handwritten code as the `internal` directory contains both.

sdk/data/aztables/client.go
```diff
@@ -23,7 +23,6 @@ type Client struct {
 	service *ServiceClient
 	cred    *SharedKeyCredential
 	name    string
-	con     *generated.Connection
```
RickWinter, line 26: Removing this would be a breaking change.  If it wasn't usable as it stood before, then its ok to remove.  Otherwise we should deprecate it instead.

sdk/data/aztables/internal/zz_service_client.go
```diff
@@ -0,0 +1,230 @@
...
+	"strconv"
+	"time"
+)
+
+// ServiceClient contains the methods for the Service group.
+// Don't use this type directly, use a constructor function instead.
+type ServiceClient struct {
+	internal *azcore.Client
+	endpoint string
+	version  Enum0
```
RickWinter, line 26: Can we get a better name on Enum0

sdk/data/aztables/service_client.go
```diff
@@ -22,19 +22,26 @@ type ServiceClient struct {
 	client  *generated.TableClient
 	service *generated.ServiceClient
 	cred    *SharedKeyCredential
-	con     *generated.Connection
```
RickWinter, line 25: See comment on table client

sdk/data/aztables/internal/zz_service_client.go
jhendrixMSFT: I can do that in a separate PR so that the cleanup doesn't get lost in the noise of regeneration.

jhendrixMSFT: It is internal only though, so having a better name isn't a huge deal IMO.

RickWinter (approved):
Run an APIView diff before checking in.  Otherwise LGTM

Issue #16643, opened 2021-12-14
User authentication and cache persistence API


chlowell:
Add an `Authenticate` method to user credentials, and persistent token caching options. Looks like this in Python: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/samples/user_authentication.py

~~Blocked on a persistent token cache implementation from MSAL for Go (https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/58).~~

camara-tech:
The mentioned blocker is now implemented here: https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-extensions-for-go/cache.

I'm interested in seeing this moved forward so that tools like AZCopy can support newer authentication mechanisms like  federated credentials

chlowell:
Thanks for pointing this out. I updated the metadata. This feature is almost ready to ship--#21244 and #21245 have the implementation minus a Linux storage implementation which is almost complete but not quite ready for release.

PR #21244, opened 2023-07-21
Persistent caching on macOS and Windows

chlowell:
This adds a new module supporting persistent caching on macOS and Windows. I won't ship this module without Linux support, which I'll add in another PR. With this module, an application can securely store authentication data and with #21245, reuse it across executions so users don't need to authenticate every time they run the application. Usage looks like this:
```go
// importing the cache module registers the implementation for the current platform
import _ "github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache"

cred, err := azidentity.NewDefaultAzureCredential(&azidentity.DefaultAzureCredentialOptions{
	// Non-nil TokenCachePersistenceOptions enables persistent caching with default options.
	TokenCachePersistenceOptions: &azidentity.TokenCachePersistenceOptions{},
})
```
Part of #16643


sdk/azidentity/cache/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache
```
jhendrixMSFT, line 1: Since this is a new module, it will require a `LICENSE.txt` file at minimum (a `README.md` file might be good too).

sdk/azidentity/example_cache_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azidentity_test
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+
+	// importing the cache module registers the cache implementation for the current platform
```
jhendrixMSFT, line 12: Might be good to augment this comment stating that Linux variants aren't supported at present.

sdk/azidentity/cache/cache.go
```diff
@@ -0,0 +1,47 @@
...
+		p, err := internal.CacheFilePath(cp.Name)
+		if err != nil {
+			return nil, err
+		}
+		return extcache.New(a, p)
+	}
+	internal.CacheFilePath = func(name string) (string, error) {
+		dir, err := cacheDir()
+		if err != nil {
+			return "", fmt.Errorf("couldn't create a cache file due to error %q", err)
```
jhendrixMSFT, line 43: I suppose that on Linux, the lack of support will trip this error.  Still, it might be more clear to have an explicit error on Linux platforms stating that it's not supported.

jhendrixMSFT: Hmm, but on Linux shouldn't the `storage()` and `cacheDir()` funcs not even exist?  I'm probably overlooking something, but the fact that the build succeeded on Linux in CI has me confused.

sdk/azidentity/cache/darwin.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return a, err
+}
+
+func tryAccessor() error {
+	a, err := accessor.New("azidentity-test-cache")
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
```
jhendrixMSFT, line 54: Typically, for contexts that aren't caller provided, it's best practice to create one with a timeout (or deadline) to avoid the potential for a hang.

sdk/azidentity/cache/darwin.go
```diff
@@ -0,0 +1,64 @@
...
+
+func storage(o internal.TokenCachePersistenceOptions) (accessor.Accessor, error) {
+	name := o.Name
+	if name == "" {
+		name = defaultName
+	}
+	var a accessor.Accessor
+	err := tryAccessor()
+	if err != nil {
+		msg := fmt.Sprintf("cache encryption is impossible because the keychain isn't usable: %s", err)
```
jhendrixMSFT, line 32: What causes us to go down this path?

sdk/azidentity/azidentity.go
```diff
@@ -50,8 +51,12 @@ var (
 	disableCP1 = strings.ToLower(os.Getenv("AZURE_IDENTITY_DISABLE_CP1")) == "true"
 )
 
+// TokenCachePersistenceOptions contains options for persistent token caching
+type TokenCachePersistenceOptions = internal.TokenCachePersistenceOptions
```
jhendrixMSFT, line 55: Missing a `CHANGELOG` entry with this feature.

sdk/azidentity/example_cache_test.go
chlowell: Sorry, I should have written in the description that I won't ship this without Linux support. My goal with this PR is to review everything else, to avoid a last minute review bomb or at least make it a smaller bomb.

sdk/azidentity/cache/cache.go
chlowell: I excluded this file on Linux with a build constraint.

sdk/azidentity/cache/darwin.go
chlowell: Anything that prevents us from writing, reading, deleting a keychain item. I imagine the most common things will be running this code in an SSH session and the user not granting keychain access.

sdk/azidentity/cache/cache.go
jhendrixMSFT: Oh yes I totally missed that.

sdk/azidentity/cache/darwin.go
chlowell: 👍 good idea. I set a generous timeout because we'll be waiting on user input.

jhendrixMSFT (approved):

PR #21251, opened 2023-07-24
Suppress nested SDK method spans

jhendrixMSFT:
The inner SDK span must be omitted.  However, any REST span it creates must be a child of the outer SDK method span.

https://azure.github.io/azure-sdk/general_implementation.html#distributed-tracing

### DIFF
--- a/sdk/azcore/runtime/policy_http_trace.go
+++ b/sdk/azcore/runtime/policy_http_trace.go
@@ -103,17 +103,23 @@ func StartSpan(ctx context.Context, name string, tracer tracing.Tracer, options 
 	if !tracer.Enabled() {
 		return ctx, func(err error) {}
 	}
+	const newSpanKind = tracing.SpanKindInternal
 	if activeSpan := ctx.Value(ctxActiveSpan{}); activeSpan != nil {
 		// per the design guidelines, if a SDK method Foo() calls SDK method Bar(),
 		// then the span for Bar() must be suppressed. however, if Bar() makes a REST
 		// call, then Bar's HTTP span must be a child of Foo's span.
-		return ctx, func(err error) {}
+		// however, there is an exception to this rule. if the SDK method Foo() is a
+		// messaging producer/consumer, and it takes a callback that's a SDK method
+		// Bar(), then the span for Bar() must _not_ be suppressed.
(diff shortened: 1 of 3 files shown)


chlowell (approved):

sdk/azcore/runtime/policy_http_trace.go
```diff
@@ -103,10 +103,17 @@ func StartSpan(ctx context.Context, name string, tracer tracing.Tracer, options
 	if !tracer.Enabled() {
 		return ctx, func(err error) {}
 	}
+	if activeSpan := ctx.Value(ctxActiveSpan{}); activeSpan != nil {
+		// per the design guidelines, if a SDK method Foo() calls SDK method Bar(),
+		// then the span for Bar() must be suppressed. however, if Bar() makes a REST
+		// call, then Bar's HTTP span must be a child of Foo's span.
+		return ctx, func(err error) {}
```
lmolkova, line 110: this does not need to be addressed right away, but there is an unobvious exception:

e.g. ServiceBus processor callback creates a span for processing and then the user calls into some SDK method like
- ServiceBus.process
  - BlobClient.upload


In this case no suppression is needed for `BlobClient.upload` because user explicitly called it. Span kind can be used as a heuristic to distinguish those (i.e. only suppress `client` or `internal` spans that happen in the scope of other `client` or `internal` spans.

lmolkova (approved):
LGTM for current scenarios just left a comment that should be addressed to make messaging instrumentation work

sdk/azcore/runtime/policy_http_trace.go
lmolkova: /cc @richardpark-msft

richardpark-msft: @lmolkova, just to check - do _I_ need an equivalent policy/check inside of Service Bus and Event Hubs to avoid nested spans?

lmolkova: I'd expect suppression to be done in core and should never be a library concern - this is what we do in other languages. 

It should probably happen inside `tracer.Start`.

jhendrixMSFT: This should now be fixed.  Only suppress the span if the in-flight span is a `client` or `internal` span.

PR #21254, opened 2023-07-24
Enable spans in azkeys

jhendrixMSFT:
Regenerated code with inject-spans enabled.


gracewilcox (approved):
LGTM

jhendrixMSFT:
@benbp have you seen this?

benbp:
go - azkeys - security - ci

benbp:
@jhendrixMSFT @gracewilcox I think this is because of the pipeline rename [here](https://github.com/Azure/azure-sdk-for-go/pull/21652) but we need to re-generate some pipelines to point back to the new yaml files. That's why the triggers are failing.

gracewilcox:
This is strange. I updated the pipelines to have the new name go - azkeys - ci, so it should be using that new name

gracewilcox:
> @jhendrixMSFT @gracewilcox I think this is because of the pipeline rename [here](https://github.com/Azure/azure-sdk-for-go/pull/21652) but we need to re-generate some pipelines to point back to the new yaml files. That's why the triggers are failing.

Strange that we need to regenerate the pipelines. Thought I did it in my pull request, but might have been wrong.

benbp:
@gracewilcox It's possible the generator won't forcibly change the location of the yaml files once it's been set. Let me see about manually changing it.

benbp:
It's trickier because of how we determine the naming convention based on the file path, and the fact that there is already a pipeline from the older non-security yaml files that exists.

gracewilcox:
Kk! I checked the live pipelines and at least they have been running correctly.

jhendrixMSFT:
Can we delete the old pipelines for non-security KV stuffs now?

gracewilcox:
> Can we delete the old pipelines for non-security KV stuffs now?

Deleted all the old pipelines. The proper pipeline should be go - azkeys - ci

PR #21258, opened 2023-07-24
Clean up constants in aztables

jhendrixMSFT:
Older versions of autorest implicitly sealed single-value enums, so enable the legacy behavior to avoid having to pass constant values. Updated the perf test to use the current version of aztables.


sdk/data/aztables/internal/zz_service_client.go
```diff
@@ -53,19 +50,19 @@ func (client *ServiceClient) GetProperties(ctx context.Context, restype Enum5, c
...
-	reqQP.Set("restype", string(restype))
-	reqQP.Set("comp", string(comp))
+	reqQP.Set("restype", "service")
+	reqQP.Set("comp", "properties")
 	if options != nil && options.Timeout != nil {
 		reqQP.Set("timeout", strconv.FormatInt(int64(*options.Timeout), 10))
 	}
 	req.Raw().URL.RawQuery = reqQP.Encode()
-	req.Raw().Header["x-ms-version"] = []string{string(client.version)}
+	req.Raw().Header["x-ms-version"] = []string{"2019-02-02"}
```
RickWinter, line 65: Why do we use a constant here rather than client.version?

sdk/data/aztables/internal/zz_service_client.go
```diff
@@ -118,19 +113,19 @@ func (client *ServiceClient) GetStatistics(ctx context.Context, restype Enum5, c
...
-	reqQP.Set("restype", string(restype))
-	reqQP.Set("comp", string(comp))
+	reqQP.Set("restype", "service")
+	reqQP.Set("comp", "stats")
 	if options != nil && options.Timeout != nil {
 		reqQP.Set("timeout", strconv.FormatInt(int64(*options.Timeout), 10))
 	}
 	req.Raw().URL.RawQuery = reqQP.Encode()
-	req.Raw().Header["x-ms-version"] = []string{string(client.version)}
+	req.Raw().Header["x-ms-version"] = []string{"2019-02-02"}
```
RickWinter, line 128: Ditto about constant

sdk/data/aztables/internal/zz_table_client.go
```diff
@@ -68,11 +66,11 @@ func (client *TableClient) createCreateRequest(ctx context.Context, dataServiceV
 		reqQP.Set("$format", string(*queryOptions.Format))
 	}
 	req.Raw().URL.RawQuery = reqQP.Encode()
-	req.Raw().Header["x-ms-version"] = []string{string(client.version)}
+	req.Raw().Header["x-ms-version"] = []string{"2019-02-02"}
 	if options != nil && options.RequestID != nil {
 		req.Raw().Header["x-ms-client-request-id"] = []string{*options.RequestID}
 	}
-	req.Raw().Header["DataServiceVersion"] = []string{string(dataServiceVersion)}
+	req.Raw().Header["DataServiceVersion"] = []string{"3.0"}
```
RickWinter, line 73: Why doesn't this come from the variable anymore?  Should user be able to configure?

sdk/data/aztables/testdata/perf/insert_entity.go
```diff
@@ -164,8 +164,8 @@ func (g *insertEntityTestGlobal) NewPerfTest(ctx context.Context, options *perf.
 }
 
 func (d *insertEntityPerfTest) Run(ctx context.Context) error {
-	_, err := d.tableClient.InsertEntity(ctx, d.entity, &aztables.InsertEntityOptions{
-		UpdateMode: aztables.EntityUpdateModeMerge,
+	_, err := d.tableClient.UpsertEntity(ctx, d.entity, &aztables.UpsertEntityOptions{
```
RickWinter, line 167: UpsertEntity is the preferred name in all languages?

RickWinter (approved):

sdk/data/aztables/testdata/perf/insert_entity.go
jhendrixMSFT: I don't know if it's the preferred name, but it's what we GA'ed with so we can't change it now.

sdk/data/aztables/internal/zz_service_client.go
jhendrixMSFT: Historically, autorest treated enums with a single value as a constant.  This default behavior was later changed due to treating enums as extensible by default if not specified.  I set the flag to enable the old behavior which is what we want as these values are constant and not user-configurable.

sdk/data/aztables/internal/zz_table_client.go
jhendrixMSFT: This is not user configurable.

PR #21259, opened 2023-07-25
[storage][azblob]DownloadFile: Download large files serially

nakulkar-msft:

============================================================
The existing DownloadFile method writes blocks of a large blob to the file in random order, which is not performant. This change will make DownloadFile method to write serially instead. The blocks are still downloaded parallelly. 
This does not change public interface of the API. Also, this change does not impact performance of small blobs (Or blobs which fit in single chunk).

Below table makes a note of time taken to transfer a 10GiB file with existing and proposed change.
```
Concurrency    Current	        Proposed
Default = 5     3m22.915s	1m8.111s
8               2m22.286s       47.138s
16	        2m19.675s	47.377s
```


sdk/storage/azblob/blob/client.go
```diff
@@ -381,6 +382,157 @@ func (b *Client) download(ctx context.Context, writer io.WriterAt, o downloadOpt
...
+			}
+
+			buff, err := aquireBuffer()
+			if err != nil {
+				return err
+			}
+
+			_, err = io.ReadFull(body, buff[:count])
+			if err != nil {
+				return err
```
jhendrixMSFT, line 516: `body` won't be closed in this case.

sdk/storage/azblob/blob/client.go
```diff
@@ -381,6 +382,157 @@ func (b *Client) download(ctx context.Context, writer io.WriterAt, o downloadOpt
...
+		Concurrency:   o.Concurrency,
+		Operation: func(ctx context.Context, chunkStart int64, count int64) error {
+			body, err := getBodyForRange(ctx, chunkStart, count)
+			if err != nil {
+				return nil
+			}
+
+			buff, err := aquireBuffer()
+			if err != nil {
+				return err
```
jhendrixMSFT, line 511: `body` won't be closed in this case.

sdk/storage/azblob/blob/constants.go
```diff
@@ -18,6 +18,8 @@ const (
 
 	// DefaultDownloadBlockSize is default block size
 	DefaultDownloadBlockSize = int64(4 * 1024 * 1024) // 4MB
+
+	defaultConcurrency = 5
```
jhendrixMSFT, line 22: We use the same value in `/internal/shared/batch_transfer.go`.  Can we move this there and consolidate them?

sdk/storage/azblob/blob/client.go
nakulkar-msft: I would not want to use `defer body.Close()`, because that will have to wait till the block is sent to the writer channel. That should be okay right

sdk/storage/azblob/blob/client.go
```diff
@@ -381,6 +383,168 @@ func (b *Client) download(ctx context.Context, writer io.WriterAt, o downloadOpt
 	return count, nil
 }
 
+// downloadFile downloads an Azure blob to a Writer. The blocks are downloaded parallely,
+// but written to file serially
+func (b *Client) downloadFile(ctx context.Context, writer io.Writer, o downloadOptions) (int64, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	if o.BlockSize == 0 {
```
vibhansa-msft, line 391: As per our test results new logic works better for larger files, shall we have a check so that for smaller files old logic is triggered and for large files new is triggered based on some minimum file size like 1GB or so.

sdk/storage/azblob/blob/constants.go
```diff
@@ -18,6 +19,8 @@ const (
 
 	// DefaultDownloadBlockSize is default block size
 	DefaultDownloadBlockSize = int64(4 * 1024 * 1024) // 4MB
+
+	DefaultConcurrency = shared.DefaultConcurrency
```
jhendrixMSFT, line 23: Needs a doc comment.

sdk/storage/azblob/internal/shared/batch_transfer.go
```diff
@@ -11,10 +11,14 @@ import (
...
 )
 
+const (
+	DefaultConcurrency = 5
+)
 // BatchTransferOptions identifies options used by doBatchTransfer.
 type BatchTransferOptions struct {
 	TransferSize  int64
 	ChunkSize     int64
+	NumChunks     uint16
```
jhendrixMSFT, line 21: Why did this become configurable?  It's internal only so no public surface area change, but it looks like in both cases the same value is used.

sdk/storage/azblob/blob/client.go
nakulkar-msft: I've put a short for files which fit in single block. Also for medium sized files, the new logic works as good as the old one. We can skip the old logic in all cases.

sdk/storage/azblob/internal/shared/batch_transfer.go
nakulkar-msft: In downloadFile() method, we use `numChunks` to define the length of the array which holds these channels. In BatchTransfer, numChunks is used to split the file into so many chunks. They are the same parameter, and computing it twice (once in downloadFile and once in doBatchTransfer()) may cause inconsistencies.

sdk/storage/azblob/blob/client.go
```diff
@@ -381,6 +383,168 @@ func (b *Client) download(ctx context.Context, writer io.WriterAt, o downloadOpt
 	return count, nil
 }
 
+// downloadFile downloads an Azure blob to a Writer. The blocks are downloaded parallely,
+// but written to file serially
+func (b *Client) downloadFile(ctx context.Context, writer io.Writer, o downloadOptions) (int64, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
nakulkar-msft: This logic is already in place for azcopy and copytool. The user still has a way to force parallel writing using DownloadBuffer() method. This will also add new parameter to the options. Do we still need this?

vibhansa-msft (approved):

souravgupta-msft (approved):

sdk/storage/azblob/internal/shared/mmf_windows.go
```diff
@@ -14,11 +14,11 @@ import (
 	"unsafe"
 )
 
-// mmb is a memory mapped buffer
-type mmb []byte
+// Mmb is a memory mapped buffer
+type Mmb []byte
 
-// newMMB creates a new memory mapped buffer with the specified size
-func newMMB(size int64) (mmb, error) {
```
tasherif-msft, line 21: do we want to expose these to the customer?

nakulkar-msft: These were un-exported always. We can discuss this, and take it in a new review if we decide to change..

souravgupta-msft: I think this is an internal implementation and we should not expose this to the customer.

jhendrixMSFT (approved):

Issue #21260, opened 2023-07-25
[azopenai] Rename package


richardpark-msft:
The current package is under `cognitiveservices/azopenai`, but we want to move it to `ai/azopenai`

The process:
1. Deprecate the old package
2. Get OpenAI to update the link to our SDK on their libraries page.
3. Incorrectly submit issue to delist old package from pkg.go.dev: [link](https://github.com/golang/go/issues/61616)
4. Retract the package: https://github.com/Azure/azure-sdk-for-go/pull/21350
5. Delete the old source code after releasing one more time.

PR #21264, opened 2023-07-25
[azopenai] Move azopenai from `cognitiveservices/azopenai` to `ai/azopenai`

richardpark-msft:
Part of the fix for #21260.

- Moved the code (to preserve history) into `ai/azopenai`
- Marked `cognitiveservices/azopenai`, prepping to release one last release to deprecate it.

We still need to submit an issue to de-list `cognitiveservices/azopenai` but I believe this takes care of our end.


sdk/cognitiveservices/azopenai/README.md
```diff
@@ -19,7 +19,7 @@ Azure OpenAI Service provides access to OpenAI's powerful language models includ
 Install the `azopenai` and `azidentity` modules with `go get`:
 
 ```bash
-go get github.com/Azure/azure-sdk-for-go/sdk/cognitiveservices/azopenai
+go get github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai
```
jhendrixMSFT, line 22: I don't think we need to make any of these changes in the old module.

sdk/cognitiveservices/azopenai/go.mod
```diff
@@ -1,3 +1,4 @@
+// Deprecated: This package has moved to github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai
```
jhendrixMSFT, line 1: Nit, usually we just say "use <foo> instead"

sdk/cognitiveservices/azopenai/README.md
richardpark-msft: I'm actually tempted to leave this one. Just in case someone cuts and pastes from some randomly found readme.

sdk/cognitiveservices/azopenai/go.mod
richardpark-msft: So "This has package has moved. Use github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai instead."?

jhendrixMSFT: I think you can omit the first sentence.  At any rate, take a look at what we did for the old keyvault modules and use the same verbiage so we're consistent.

sdk/cognitiveservices/azopenai/README.md
jhendrixMSFT: It just seems weird that the readme for module A says "go get module B".

richardpark-msft: I ended up just clearing the readme out to make it obvious this is the old module (ie, we're really only publishing this so users get a pointer to the new module anyways).

sdk/cognitiveservices/azopenai/go.mod
richardpark-msft: Changed to match what we put in keyvault.

sdk/cognitiveservices/azopenai/autorest.md
```diff
@@ -9,7 +9,7 @@ input-file:
 
 output-folder: ../azopenai
 clear-output-folder: false
-module: github.com/Azure/azure-sdk-for-go/sdk/cognitiveservices/azopenai
+module: github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai
```
jhendrixMSFT, line 12: Is this more `git mv` shenanigans or does this need to be copied?

jhendrixMSFT (approved):

sdk/cognitiveservices/azopenai/autorest.md
richardpark-msft: Accidental search and replace. :|

PR #21282, opened 2023-07-27
[azcore/azeventgrid] Remove json.RawMessage usage in the public API

richardpark-msft:
This avoids any future customer pain if the `json` types or API were to change.

### DIFF
--- a/sdk/azcore/messaging/cloud_event.go
+++ b/sdk/azcore/messaging/cloud_event.go
@@ -44,12 +44,12 @@ type CloudEvent struct {
 	// Data is the payload for the event.
 	// * []byte will be serialized and deserialized as []byte.
 	// * Any other type will be serialized to a JSON object and deserialized into
-	//   a [json.RawMessage].
+	//   a []byte, containing the JSON text.
 	//
-	// To deserialize a [json.RawMessage] into your chosen type:
+	// To deserialize into your chosen type:
 	//
 	//   var yourData *YourType
-	//   json.Unmarshal(cloudEvent.Data.(json.RawMessage), &yourData)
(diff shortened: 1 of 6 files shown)


sdk/azcore/messaging/cloud_event.go
```diff
@@ -44,12 +44,12 @@ type CloudEvent struct {
...
+	//   a []byte, containing the JSON text.
 	//
-	// To deserialize a [json.RawMessage] into your chosen type:
+	// To deserialize into your chosen type:
 	//
 	//   var yourData *YourType
-	//   json.Unmarshal(cloudEvent.Data.(json.RawMessage), &yourData)
+	//   json.Unmarshal(cloudEvent.Data.([]byte), &yourData)
 	//
 	Data any
```
jhendrixMSFT, line 54: We don't have to fix this now, but I think this contract needs tightening up.  If I correctly understand the description, it's either a (base-64 encoded) sequence of bytes or a JSON payload.  And the response is always a `[]byte` (I guess the caller knows if it's JSON or a base-64 encoded sequence?).  The conversion just seems a little magical.  Can't we make it a `[]byte` and require the caller to pass what they want without the magic?

richardpark-msft: Filed this here: https://github.com/Azure/azure-sdk-for-go/issues/21289

So despite this being a bit open I don't think it's actually that bad and it does make the type a bit easier to use since we guide you into one of two paths - either a []byte or a JSON-serializable value (not the JSON bytes). 

`any` feels appropriate to me but it's still worth a discussion now that we're getting more familiar with the type.

jhendrixMSFT (approved):

PR #21285, opened 2023-07-28
Major version upgrade when processing fake

Alancere:
The major version was upgraded when the fake was generated, and the module in fake _server.go needs to be replaced.


tadelesh:
I don't think we should do it specifically for fake. We could try to handle all kinds of major version import.

tadelesh (approved):

PR #21288, opened 2023-07-28
Make recording.proxyManualStartEnv private

benbp:
Marking this well known test proxy variable private to lower public api surface area.

### DIFF
--- a/sdk/internal/perf/recording_test.go
+++ b/sdk/internal/perf/recording_test.go
@@ -15,9 +15,11 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
 )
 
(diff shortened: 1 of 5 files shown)


sdk/internal/perf/recording_test.go
```diff
@@ -15,9 +15,11 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
 )
 
+const proxyManualStartEnv = "PROXY_MANUAL_START"
```
benbp, line 18: This will be removed fully after I refactor the rest of the tests to auto-start the proxy, and remove the proxy start step from the CI yaml.

jhendrixMSFT (approved):

PR #21294, opened 2023-07-31
Encryption Scope SAS

siminsavani-msft:
Adding encryption sas to blob, account, and identity sas

### DIFF
--- a/sdk/storage/azblob/CHANGELOG.md
+++ b/sdk/storage/azblob/CHANGELOG.md
@@ -6,7 +6,8 @@
 * Added support for [Cold tier](https://learn.microsoft.com/azure/storage/blobs/access-tiers-overview?tabs=azure-portal).
 * Added `CopySourceTag` option for `UploadBlobFromURLOptions`
 * Added [FilterBlobs](https://learn.microsoft.com/rest/api/storageservices/find-blobs-by-tags-container) by tags API for container client.
-* Added `System` option to `ListContainersInclude` to allow listing of system containers. 
+* Added `System` option to `ListContainersInclude` to allow listing of system containers.
+* Updated the SAS Version to `2020-12-06` and added `Encryption Scope` to Account SAS, Blob SAS, and User Delegation SAS 
 
 ### Breaking Changes
 
--- a/sdk/storage/azblob/sas/account.go
+++ b/sdk/storage/azblob/sas/account.go
@@ -32,7 +32,6 @@ type AccountSignatureValues struct {
 	Permissions     string    `param:"sp"`  // Create by initializing AccountPermissions and then call String()
 	IPRange         IPRange   `param:"sip"`
 	ResourceTypes   string    `param:"srt"` // Create by initializing AccountResourceTypes and then call String()
-	Identifier      string    `param:"si"`
 	EncryptionScope string    `param:"ses"`
 }
 
--- a/sdk/storage/azblob/sas/account.go
+++ b/sdk/storage/azblob/sas/account.go
@@ -70,7 +70,8 @@ func (v AccountSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKey
 		v.IPRange.String(),
 		string(v.Protocol),
 		v.Version,
-		v.EncryptionScope},
+		v.EncryptionScope,
+		""},
 		"\n")
 
 	signature, err := exported.ComputeHMACSHA256(sharedKeyCredential, stringToSign)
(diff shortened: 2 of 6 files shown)


sdk/storage/azblob/sas/account.go
```diff
@@ -25,13 +25,15 @@ type UserDelegationCredential = exported.UserDelegationCredential
...
-	IPRange       IPRange   `param:"sip"`
-	ResourceTypes string    `param:"srt"` // Create by initializing AccountResourceTypes and then call String()
+	Version         string    `param:"sv"`  // If not specified, this format to SASVersion
+	Protocol        Protocol  `param:"spr"` // See the SASProtocol* constants
+	StartTime       time.Time `param:"st"`  // Not specified if IsZero
+	ExpiryTime      time.Time `param:"se"`  // Not specified if IsZero
+	Permissions     string    `param:"sp"`  // Create by initializing AccountPermissions and then call String()
+	IPRange         IPRange   `param:"sip"`
+	ResourceTypes   string    `param:"srt"` // Create by initializing AccountResourceTypes and then call String()
+	Identifier      string    `param:"si"`
```
souravgupta-msft, line 35: I don't think Signed Identifiers(si) are used in Account SAS creation. Also, this field is not used anywhere in the code below. So, it should be removed.

sdk/storage/azblob/sas/query_params.go
```diff
@@ -23,7 +23,7 @@ const (
 
 var (
 	// Version is the default version encoded in the SAS token.
-	Version = "2020-02-10"
+	Version = "2020-12-06"
```
souravgupta-msft, line 26: The version should be the STG 85/86 version in my opinion.
```suggestion
	Version = "2021-12-02"
```

sdk/storage/azblob/sas/service.go
```diff
@@ -103,7 +104,8 @@ func (v BlobSignatureValues) SignWithSharedKey(sharedKeyCredential *SharedKeyCre
 		string(v.Protocol),
 		v.Version,
 		resource,
-		snapshotTime,         // signed timestamp
+		snapshotTime, // signed timestamp
+		v.EncryptionScope,
 		v.CacheControl,       // rscc
 		v.ContentDisposition, // rscd
 		v.ContentEncoding,    // rsce
```
souravgupta-msft, line 111: As per the [documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#version-2020-12-06-and-later), `v.ContentType // rsct` is removed from version 2020-12-06 and later for creating signature string. You can check if this is actually the case or is it a documentation bug.

sdk/storage/azblob/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 * Added support for [Cold tier](https://learn.microsoft.com/azure/storage/blobs/access-tiers-overview?tabs=azure-portal).
 * Added `CopySourceTag` option for `UploadBlobFromURLOptions`
 * Added [FilterBlobs](https://learn.microsoft.com/rest/api/storageservices/find-blobs-by-tags-container) by tags API for container client.
-* Added `System` option to `ListContainersInclude` to allow listing of system containers. 
+* Added `System` option to `ListContainersInclude` to allow listing of system containers.
+* Updated the SAS Version to `2020-12-06` and added `Encryption Scope` to Account SAS, Blob SAS, and User Delegation SAS
```
souravgupta-msft, line 10:
```suggestion
* Updated the SAS Version to `2021-12-02` and added `Encryption Scope` to Account SAS, Service SAS, and User Delegation SAS 
```

seanmcc-msft: We should make sure we are sending the 2021-12-02 SAS version to the service, it should match the x-ms-version.

sdk/storage/azblob/sas/query_params.go
siminsavani-msft: That's right! Thanks for the call out!

sdk/storage/azblob/sas/account.go
seanmcc-msft: I agree with @souravgupta-msft, I am not seeing Signed Identifiers in the REST docs for account SAS - https://learn.microsoft.com/en-us/rest/api/storageservices/create-account-sas

sdk/storage/azblob/sas/service.go
siminsavani-msft: I believe that is a bug, I tried to remove it and get 'Authentication Failed'. I double checked the string to sign in the .NET SDK and it matches with what we have here.

https://github.com/Azure/azure-sdk-for-net/blob/e8f203346cbc8135aac9ce250853fd4cbed9d75d/sdk/storage/Azure.Storage.Blobs/src/Sas/BlobSasBuilder.cs#L350

souravgupta-msft (approved):

Issue #21291, opened 2023-07-30
Incorrect nextURL link for  armcompute.VirtualMachinesClient with filter of virtualMachineScaleSet (VMSS flex)


linguo555:
### Bug Report

	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4"
	"github.com/Azure/azure-sdk-for-go/services/resourcegraph/mgmt/2019-04-01/resourcegraph"

We are currently using armcompute.VirtualMachinesClient to listinstances. At beginning we added a filter for virtualMachineScaleSets, see below: 
`filter := fmt.Sprintf("'virtualMachineScaleSet/id' eq "+
		"'/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/virtualMachineScaleSets/%s'",
		c.subscription, c.resourceGroupName, vmssName)
	option := armcompute.VirtualMachinesClientListOptions{
		&filter,
	}
	pager := client.NewListPager(c.resourceGroupName, &option)`
	
When the VM list has more than 50 vms and needs Pagination, the `nextLink` is proved to be an invalid url which returned with response 400  BAD Request. 
Once we moved the option, everything goes ok. 
We manually take the url out and verified it indeed a bad url:

`400 Bad Request
ERROR CODE: BadRequest
{
  "error": {
    "code": "BadRequest",
    "message": "The request URL is not valid."
  }
}`

There is a bug in the SDK with the VMSS flex option. 

Ps: the VMSS we use is VMSS flex. 

Thanks,
Lin

jhendrixMSFT:
Can you please enable [logging](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore#hdr-Built_in_Logging) and provide the URL that's causing the 400 to be returned?

zhuweirongdb:
I think the issue is not the NextLink URL returned by the server, but how the SDK handles the NextLink. 

An example of the returned NextLink looks like

```
[...]

PR #21296, opened 2023-07-31
Validate and encode query parameters

jhendrixMSFT:
If the endpoint passed to runtime.NewRequest contains query parameters they must be validated and encoded.  The typical case is when a paged operation's nextLink value contains query parameters.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21291


chlowell:
I get that we can ship this much faster than a codegen update, but it's odd that callers would be responsible for encoding everything except query params. Shouldn't azcore encode everything or nothing?

jhendrixMSFT:
I was (am) struggling with that too.  I agree that ideally it should be all or nothing.  These are the points that swayed me to go down this path.
- I believe we don't have enough context at this point to know how properly encode the path (e.g. a `/` should be URL encoded within a path segment, but obviously not the ones that separate segments). Please correct me if I'm wrong on this one.
- In the presence of query params, we could split, encode, and reconstruct the path.  It just seems a bit redundant though given that this already happens in `http.NewRequestWithContext`.
- Having a centralized fix is appealing, though not a requirement (in hindsight, I should have tried to move more of this logic into `azcore`).

Totally open to other suggestions.

jhendrixMSFT:
In some ways, you could make the argument that this is how the `url` APIs (although they are a bit clunky).

https://play.golang.com/p/4d17WGIC452

Here, `RawQuery` contains the unencoded query parameters which is strange considering the docs state that it should contain the "encoded query values".

chlowell:
If we need to support path segments containing `/`, I agree azcore can't encode the whole endpoint. But we could (continue to) expect callers to encode it. Considering the documented and actual behavior of `NewRequest`, isn't any bug here in the calling code?

jhendrixMSFT:
You've convinced me.  I will retool this such that the caller encodes the query parameters (just like we do for path segments).

jhendrixMSFT:
I've moved this to be specific to pagers by adding a helper when creating a `PagingHandler[T].Fetcher`.  The generated code would be updated as per below.  I like this as it keeps the changes targeted to handling of next links and consolidates the fetcher so if we can centralize future updates.
```go
Fetcher: func(ctx context.Context, page *PagingClientGetMultiplePagesResponse) (PagingClientGetMultiplePagesResponse, error) {
	ctx = context.WithValue(ctx, runtime.CtxAPINameKey{}, "PagingClient.NewGetMultiplePagesPager")
	nextLink := ""
	if page != nil {
		nextLink = *page.NextLink
	}
	resp, err := runtime.FetcherHelper(ctx, client.internal.Pipeline(), nextLink, func(ctx context.Context) (*policy.Request, error) {
		return client.getMultiplePagesCreateRequest(ctx, options)
	})
	if err != nil {
		return PagingClientGetMultiplePagesResponse{}, err
	}
	return client.getMultiplePagesHandleResponse(resp)
},
```

jhendrixMSFT:
Unfortunately this falls apart for paged methods that use a next link operation instead of calling `runtime.NewRequest`.  It might just be better to solve this on the codegen side.

sdk/azcore/runtime/pager.go
```diff
@@ -88,3 +90,25 @@ func (p *Pager[T]) NextPage(ctx context.Context) (T, error) {
 func (p *Pager[T]) UnmarshalJSON(data []byte) error {
 	return json.Unmarshal(data, &p.current)
 }
+
+// FetcherHelper is a helper containing boilerplate code to simplify creating a PagingHandler[T].Fetcher.
+func FetcherHelper(ctx context.Context, pl Pipeline, nextLink string, createReq func(context.Context) (*policy.Request, error)) (*http.Response, error) {
```
jhendrixMSFT, line 95: Mabye we should just name this `Fetcher`?  I'm not super thrilled by the suffix but can live with it.

chlowell: It's an unfortunate name but it does accurately describe this thing. I know you looked at this already, but I don't have a firm opinion on where to draw the line between generating code and consolidating it into azcore, so I still wonder whether it's better to generate this (once per package?) instead of adding it to the public API. If it's just the increase in generated LOC that's troubling, could we generate a generic `Fetcher` factory?

sdk/azcore/CHANGELOG.md
```diff
@@ -15,7 +17,6 @@
 ### Features Added
 
 * Added function `SanitizePagerPollerPath` to the `server` package to centralize sanitization and formalize the contract.
-* Added `TokenRequestOptions.EnableCAE` to indicate whether to request a CAE token.
```
chlowell, line 18: Merge oops?

jhendrixMSFT: D'oh, fixed.

sdk/azcore/runtime/pager.go
jhendrixMSFT: > where to draw the line between generating code and consolidating it into azcore

This is a great point and probably something we should try to formalize.

The way I think about it is that `azcore` contains general-purpose building blocks and the generated code assembles them in a domain-specific way.  So, I would argue that if we have repetitive chunks of general-purpose code being generated, we should move those chunks to `azcore`.

@JeffreyRichter what are your thoughts on this?

jhendrixMSFT: Back to naming, perhaps this should be `FetcherForNextLink` (or something like that) as it only works for paged operations that use the "next link" pattern.

JeffreyRichter: My immediate thought is that this will be in the public surface area for the runtime package, but no customer should ever use it directly, right?
So, the docs for it would say to not use it and there would be no example code for it either, right?
And, as always, if we code-gen it today, then we could move it here tomorrow, but we can never the reverse. So, I'm prefer the conservative approach.

jhendrixMSFT: I renamed this to `FetcherForNextLink` in hopes that one day we'll also have a `FetcherForContinuationToken` if/when we support modeling this.

Jeff and I discussed when something should go in `azcore/runtime` vs generated code.  Here are the guiding principles we came up with.  Not really a prescriptive list, but things to keep in mind when weighing the decision.
- Generated code
  - Pros: immune to breaking changes, decentralized code insulates packages from regressions, no new public surface area
  - Cons: duplicates code, rolling out a bug-fix requires fixing/releasing the code generator, then regenerating the world
- Code in `azcore/runtime`
  - Pros: no duplicated code, bug fixes are centralized
  - Cons: regressions in shared code can break everybody, no breaking changes means potential for deprecated code

As an aside, the fact that `runtime` contains some APIs intended for use by customers was a mistake and will be cleaned up in https://github.com/Azure/azure-sdk-for-go/pull/21406.

chlowell (approved):

sdk/azcore/runtime/pager.go
```diff
@@ -88,3 +90,25 @@ func (p *Pager[T]) NextPage(ctx context.Context) (T, error) {
...
 	return json.Unmarshal(data, &p.current)
 }
+
+// FetcherForNextLink is a helper containing boilerplate code to simplify creating a PagingHandler[T].Fetcher from a next link URL.
+func FetcherForNextLink(ctx context.Context, pl Pipeline, nextLink string, createReq func(context.Context) (*policy.Request, error)) (*http.Response, error) {
+	var req *policy.Request
+	var err error
+	if nextLink == "" {
+		req, err = createReq(ctx)
+	} else if nextLink, err = EncodeQueryParams(nextLink); err == nil {
```
JeffreyRichter, line 100: This all looks good to me.
One thing that is supposed to happen and I can't tell if it is with this code is that the api-version query parameter value should be changed to the same value used by the client and not remain with the same value as was in the nextLink URL.
The scenario is that a client using api-version=v1 initiates the paging and then a client using api-version=v2 gets the next page.
This is also how LROs should work. The Poller does a GET to a URL and the api-version value should reflect that of the current client used when creating the poller.

jhendrixMSFT: We don't do that at present.  I've opened https://github.com/Azure/azure-sdk-for-go/issues/21408 to track that.

JeffreyRichter (approved):

sdk/azcore/runtime/pager.go
JeffreyRichter: ok

yumingqiao:
Hi @jhendrixMSFT, it seems the fix was made in v1.8.0, though we still experience the same issue with azcore v1.11.1. I see the server returned nextLink was not escaped in the second page's request. In the example below. The first response is truncated for brevity, at the end, you can see it returns `"nextLink"`, which is a unescaped URL, i.e. `$filter='virtualMachineScaleSet/id' eq '/subscriptions/'`. Then in the request of the next page, the SDK send the nextLink as it without escaping, as opposed to the bug fix, see `Request URL: https://management.azure.com:443/subscriptions/xxxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachines?api-version=2022-11-01&$skiptoken=83afce64-955b-47b3-8188-db0a7cf9c530!/Subscriptions/xxxxx/ResourceGroups/xxxx/VMs/C3D2001D933B-789C_AADFBC6A&$filter='virtualMachineScaleSet/id' eq '/subscriptions/xxxxx/resourceGroups/dev-azure-wu-nephos-0-yvjzkx/providers/Microsoft.Compute/virtualMachineScaleSets/c3d2001d933b-789c'`

Could you let us know if this is expected? We have many VMs in a resource group, if we are not able to query by VMSS, we will have to get all VMs of the entire resource group and filter by VMSS ourselves, this will consume much more API quota. Thank you.

```
        "diagnosticsProfile": {
          "bootDiagnostics": {
            "enabled": true
          }
        },
        "priority": "Regular",
        "timeCreated": "2024-07-19T05:05:33.4935443+00:00"
      }
    }
  ],
  "nextLink": "https://management.azure.com:443/subscriptions/xxxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachines?api-version=2022-11-01&$skiptoken=83afce64-955b-47b3-8188-db0a7cf9c530!/Subscriptions/xxxxx/ResourceGroups/xxxx/VMs/C3D2001D933B-789C_AADFBC6A&$filter='virtualMachineScaleSet/id' eq '/subscriptions/xxxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachineScaleSets/c3d2001d933b-789c'"
}
I0719 05:12:07.979935       1 http_client_utils.go:71] Request URL: https://management.azure.com:443/subscriptions/xxxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachines?api-version=2022-11-01&$skiptoken=83afce64-955b-47b3-8188-db0a7cf9c530!/Subscriptions/xxxxx/ResourceGroups/xxxx/VMs/C3D2001D933B-789C_AADFBC6A&$filter='virtualMachineScaleSet/id' eq '/subscriptions/xxxxx/resourceGroups/dev-azure-wu-nephos-0-yvjzkx/providers/Microsoft.Compute/virtualMachineScaleSets/c3d2001d933b-789c'
I0719 05:12:07.979952       1 http_client_utils.go:72] Request Method: GET
I0719 05:12:07.979955       1 http_client_utils.go:73] Request Headers: map[User-Agent:[azsdk-go-armcompute.VirtualMachinesClient/v4.2.0 (go1.22.4 X:boringcrypto,nocoverageredesign; linux)]]
I0719 05:12:07.980568       1 http_client_utils.go:93] Response Status: 400 Bad Request
I0719 05:12:07.980581       1 http_client_utils.go:94] Response Headers: map[Content-Length:[311] Content-Type:[text/html; charset=us-ascii] Date:[Fri, 19 Jul 2024 05:12:07 GMT]]
I0719 05:12:07.980594       1 http_client_utils.go:101] Response Body: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Bad Request</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
<BODY><h2>Bad Request</h2>
<hr><p>HTTP Error 400. The request is badly formed.</p>
</BODY></HTML>
E0719 05:12:07.980655       1 compute_client.go:112] "Failed to list virtual machines with vmss filter via VM API." err=<
	GET https://management.azure.com:443/subscriptions/xxxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachines
	RESPONSE 400: 400 Bad Request
	ERROR CODE UNAVAILABLE
	<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
	<HTML><HEAD><TITLE>Bad Request</TITLE>
	<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
	<BODY><h2>Bad Request</h2>
	<hr><p>HTTP Error 400. The request is badly formed.</p>
	</BODY></HTML>
	
 >
```

jhendrixMSFT:
The fix is comprised of a helper in `azcore/runtime` for escaping the next link URL and some changes to the generated SDKs to use it. So, it will not be enough to just update to the latest `azcore` to get the fix. You must also upgrade your SDKs to a later version.

yumingqiao:
@jhendrixMSFT the latest VM client [armconpute is v4.2.1](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4), which is released in 2023 April before your fix. Do you have another version I can try?

jhendrixMSFT:
Latest version is `v5.7.0` https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5

PR #21297, opened 2023-07-31
[AzDatalake] Lease Clients Implementation

tasherif-msft:


sdk/storage/azdatalake/lease/models.go
```diff
@@ -115,6 +138,14 @@ type PathBreakOptions struct {
 }
 
 func (o *PathBreakOptions) format() *lease.BlobBreakOptions {
+	if o == nil || o.ModifiedAccessConditions == nil {
+		if o != nil {
```
gapra-msft, line 142: nit picking but this code just seems a little weird to me

gapra-msft (approved):

vibhansa-msft (approved):

PR #21298, opened 2023-07-31
[AzDatalake] Pipelines + cleanup

tasherif-msft:
TODO:
figure out unmarshaljson issue


vibhansa-msft (approved):

benbp:
Is this copied over from net? I know in net the arm template is kind of a parent template for all their services, so it deploys everything. In go that's not the case, so we shouldn't be deploying all these extra resources in the template included here unless necessary. Similarly, I don't know if the go tests will require anything set up in https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/test-resources-post.ps1 or not?

PR #21333, opened 2023-08-08
Implement stable CAE design

### LINKED ISSUE #19215: Add Continuous Access Evaluation (CAE) support for service principals
Blocked on https://github.com/Azure/azure-sdk-for-go/issues/14931.

chlowell:
The big change from the beta design is that CAE becomes a token request option that's off by default. This allows clients to decide whether to acquire CAE tokens, which makes sense because clients are responsible for handling CAE challenges. MSAL will continue to support CAE only as a client option, so our per-request option requires credentials to maintain two MSAL clients. Most of the diff here is adding wrappers to manage that.

Closes #19215


sdk/azidentity/azidentity.go
```diff
@@ -41,73 +41,18 @@ const (
...
 	defaultSuffix           = "/.default"
-	tenantIDValidationErr   = "invalid tenantID. You can locate your tenantID by following the instructions listed here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names"
 )
 
 var (
 	// capability CP1 indicates the client application is capable of handling CAE claims challenges
-	cp1        = []string{"CP1"}
-	disableCP1 = strings.ToLower(os.Getenv("AZURE_IDENTITY_DISABLE_CP1")) == "true"
+	cp1                = []string{"CP1"}
+	errInvalidTenantID = errors.New("invalid tenantID. You can locate your tenantID by following the instructions listed here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names")
```
scottaddie, line 49:
```suggestion
	errInvalidTenantID = errors.New("invalid tenantID. You can locate your tenantID by following the instructions listed here: https://learn.microsoft.com/partner-center/find-ids-and-domain-names")
```

sdk/azidentity/confidential_client.go
```diff
@@ -0,0 +1,153 @@
...
+		host:     host,
+		name:     name,
+		noCAEMu:  &sync.Mutex{},
+		opts:     opts,
+		tenantID: tenantID,
+	}, nil
+}
+
+// GetToken requests an access token from MSAL, checking the cache first.
+func (c *confidentialClient) GetToken(ctx context.Context, tro policy.TokenRequestOptions) (azcore.AccessToken, error) {
```
jhendrixMSFT, line 66: Correct me if I'm wrong, but it appears that all of the per-credential logic for `GetToken()` has been moved out of the concrete token types and consolidated here.  In effect, the polymorphic nature of `azcore.TokenCredential` has been lost.  Assuming this is the case, my concern is that, over time, this function will become a large number of `if/else` statements to do the correct thing based on the options set and/or the calling token type.

I'm still reviewing, but IMO it would be ideal if each token type can maintain its own implementation of `GetToken()` so we don't end up with this monolithic implementation.  Not sure if you went down that rabbit hole (or how far).  What would it take to achieve this?

sdk/azidentity/azure_cli_credential.go
```diff
@@ -53,8 +55,8 @@ func (o *AzureCLICredentialOptions) init() {
 
 // AzureCLICredential authenticates as the identity logged in to the Azure CLI.
 type AzureCLICredential struct {
-	s             *syncer
-	tokenProvider azureCLITokenProvider
+	mu   *sync.Mutex
```
jhendrixMSFT, line 58: Why did `syncer` go away for CLI cred?

jhendrixMSFT: Derp, `syncer` was deleted.

chlowell: For more context, I deleted `syncer` because it doesn't do much work and the CLI cred is the only one using it now that every other credential uses an MSAL client wrapper. Other non-MSAL creds could use it, however we don't plan to have many of those.

sdk/azidentity/syncer.go
```diff
@@ -1,120 +0,0 @@
...
-
-// resolveAdditionalTenants returns a copy of tenants, simplified when tenants contains a wildcard
-func resolveAdditionalTenants(tenants []string) []string {
-	if len(tenants) == 0 {
-		return nil
-	}
-	for _, t := range tenants {
-		// a wildcard makes all other values redundant
-		if t == "*" {
-			return []string{"*"}
```
jhendrixMSFT, line 114: I can't find where this went.  Was it inadvertently removed?

sdk/azidentity/confidential_client.go
```diff
@@ -0,0 +1,153 @@
...
+	AdditionallyAllowedTenants []string
+	// Assertion for on-behalf-of authentication
+	Assertion                         string
+	DisableInstanceDiscovery, SendX5C bool
+}
+
+// confidentialClient wraps the MSAL confidential client
+type confidentialClient struct {
+	cae, noCAE               msalConfidentialClient
+	caeMu, clientMu, noCAEMu *sync.Mutex
```
jhendrixMSFT, line 36: Nit, keep the ordering the same as the clients.

sdk/azidentity/public_client.go
```diff
@@ -0,0 +1,175 @@
...
+	LoginHint, RedirectURL     string
+	Username, Password         string
+}
+
+// publicClient wraps the MSAL public client
+type publicClient struct {
+	account                  public.Account
+	cae, noCAE               msalPublicClient
+	clientID, tenantID       string
+	clientMu, caeMu, noCAEMu *sync.Mutex
```
jhendrixMSFT, line 37: Nit, keep the ordering the same as the clients and move one line up.

sdk/azidentity/confidential_client.go
```diff
@@ -0,0 +1,153 @@
...
+	if len(tro.Scopes) < 1 {
+		return azcore.AccessToken{}, fmt.Errorf("%s.GetToken() requires at least one scope", c.name)
+	}
+	client, mu, err := c.client(ctx, tro)
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+	// we don't resolve the tenant for managed identities because they acquire tokens only from their home tenants
+	if c.name != credNameManagedIdentity {
+		tenant, err := c.resolveTenant(tro.TenantID)
```
jhendrixMSFT, line 76: Nit: unless the ordering is important, it might be nice to `resolveTenant` before calling `c.client` so the sequence of events is the same as the public client, plus we don't create the client if the `TenantID` is invalid.

sdk/azidentity/public_client.go
```diff
@@ -0,0 +1,175 @@
...
+	} else {
+		res := getResponseFromError(err)
+		err = newAuthenticationFailedError(p.name, err.Error(), res, err)
+	}
+	return azcore.AccessToken{Token: ar.AccessToken, ExpiresOn: ar.ExpiresOn.UTC()}, err
+}
+
+// resolveTenant returns the correct tenant for a token request given the client's
+// configuration, or an error when that configuration doesn't allow the specified tenant
+func (m *publicClient) resolveTenant(specified string) (string, error) {
```
jhendrixMSFT, line 173: Receiver name should be `p`.

sdk/azidentity/public_client.go
```diff
@@ -0,0 +1,175 @@
...
+			Message:         dc.Result.Message,
+			UserCode:        dc.Result.UserCode,
+			VerificationURL: dc.Result.VerificationURL,
+		})
+		if err == nil {
+			ar, err = dc.AuthenticationResult(ctx)
+		}
+	case m.opts.Username != "" && m.opts.Password != "":
+		ar, err = c.AcquireTokenByUsernamePassword(ctx, tro.Scopes, m.opts.Username, m.opts.Password, public.WithClaims(tro.Claims), public.WithTenantID(tenant))
+	default:
```
jhendrixMSFT, line 113: What's the risk of having `InteractiveBrowserCredential` be the default case?

sdk/azidentity/public_client.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// reqToken requests a token from the MSAL public client. It's separate from GetToken() to enable Authenticate() to bypass the cache.
+func (m *publicClient) reqToken(ctx context.Context, c msalPublicClient, tro policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	tenant, err := m.resolveTenant(tro.TenantID)
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+	var ar public.AuthResult
+	switch {
```
jhendrixMSFT, line 97: Generally speaking, would it make more sense to switch on the credential name?  Might make it clear which flow is being used.

jhendrixMSFT: I don't feel super-strong about it but might help future expansion.

sdk/azidentity/syncer.go
chlowell: Thanks for noticing! I'll restore this. I may have persuaded myself we don't need it anymore but we do still need to copy the slice of allowed tenants because the application owns it

sdk/azidentity/public_client.go
chlowell: Good idea, I'll do that.

sdk/azidentity/public_client.go
chlowell: We block here when a credential's constructor doesn't set the right options. I'll change this to switch on the credential name, which is a good idea anyway and would cause tests to fail faster.

jhendrixMSFT (approved):

PR #21348, opened 2023-08-10
[AzDatalake] APIView Feedback + Samples + Doc + CI/Live issues cleanup

tasherif-msft:


sdk/storage/azdatalake/directory/responses.go
```diff
@@ -10,11 +10,14 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/internal/path"
 )
 
-// RenameResponse contains the response fields for the Create operation.
-type RenameResponse struct {
-	Response           CreateResponse
-	NewDirectoryClient *Client
-}
+//// RenameResponse contains the response fields for the Create operation.
```
jhendrixMSFT, line 13: Nit: I think can be removed.

sdk/storage/azdatalake/file/responses.go
```diff
@@ -26,15 +26,21 @@ type AppendDataResponse = generated.PathClientAppendDataResponse
 // FlushDataResponse contains the response from method Client.FlushData.
 type FlushDataResponse = generated.PathClientFlushDataResponse
 
-// RenameResponse contains the response fields for the Create operation.
-type RenameResponse struct {
-	Response      CreateResponse
-	NewFileClient *Client
-}
+// RenameResponse contains the response fields for the Rename operation.
+type RenameResponse = CreateResponse
```
jhendrixMSFT, line 30: Same as directory, add a new type instead of using an alias in the event we want to include the new client.

sdk/storage/azdatalake/directory/client.go
```diff
@@ -249,7 +249,7 @@ func (d *Client) renamePathInURL(newName string) (string, string, string) {
 	return parsedNewURL.Path, newPathURL, newBlobURL
 }
 
-// Rename renames a directory (dfs1)
+// Rename renames a directory.
```
jhendrixMSFT, line 252: After the rename, is the current client no longer valid due to the name change?  If so, we should document this (applies to the other `Rename` APIs).

sdk/storage/azdatalake/filesystem/models.go
```diff
@@ -143,15 +152,18 @@ func (o *ListPathsOptions) format() generated.FileSystemClientListPathsOptions {
 		Continuation: o.Marker,
 		MaxResults:   o.MaxResults,
 		Path:         o.Prefix,
-		Upn:          o.Upn,
+		Upn:          o.UPN,
 	}
 }
 
-// ListDeletedPathsOptions contains the optional parameters for the Filesystem.ListDeletedPaths operation. PLACEHOLDER
+// ListDeletedPathsOptions contains the optional parameters for the FileSystem.ListDeletedPaths operation. PLACEHOLDER
```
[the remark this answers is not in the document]
tasherif-msft: will remove, forgot to remove it

sdk/storage/azdatalake/file/examples_test.go
```diff
@@ -0,0 +1,331 @@
...
+	handleError(err)
+
+	data, err := io.ReadAll(dResp.Body)
+	handleError(err)
+
+	downloadedMD5Value := md5.Sum(data)
+	downloadedContentMD5 := downloadedMD5Value[:]
+
+	// compare the hashes
+	fmt.Println(downloadedContentMD5, contentMD5)
```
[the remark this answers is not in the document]
tasherif-msft: Given that these are not constants, it won't be possible to use this feature in the case of variable params, I'll just keep it as print for now

gapra-msft (approved):

PR #21351, opened 2023-08-11
Include response body in IMDS 400 error message

### LINKED ISSUE #19510: Improve IMDS error diagnosability
`ManagedIdentityCredential.GetToken()` returns a `credentialUnavailableError` when IMDS returns a 400 response. `credentialUnavailableError` doesn't carry the raw response, preventing applications from programmatically inspecting it. The error type is correct--a 400 from IMDS mustn't be fatal when iterating a chain of credentials--but the response, if any, should be exposed to the application.

chlowell:
This makes the message more helpful, especially when the client gets a response from something other than IMDS. Closes #19510

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -179,7 +180,12 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKi
 		if id != nil {
 			return azcore.AccessToken{}, newAuthenticationFailedError(credNameManagedIdentity, "the requested identity isn't assigned to this resource", resp, nil)
 		}
-		return azcore.AccessToken{}, newCredentialUnavailableError(credNameManagedIdentity, "no default identity is assigned to this resource")
+		msg := "failed to authenticate a system assigned identity"
+		if body, err := io.ReadAll(resp.Body); err == nil && len(body) > 0 {
+			msg += fmt.Sprintf(". The endpoint responded with %s", body)
+		}
+		resp.Body.Close()
```
richardpark-msft, line 187: I believe using this:

https://github.com/Azure/azure-sdk-for-go/blob/514985a4c9d311c90b86c74c6ff779abcec1ffcd/sdk/azcore/runtime/response.go#L24

Will make sure the response body is still available, which is useful if they're using [`WithCaptureResponse`](https://github.com/Azure/azure-sdk-for-go/blob/514985a4c9d311c90b86c74c6ff779abcec1ffcd/sdk/azcore/runtime/policy_include_response.go#L32).

richardpark-msft (approved):
Minor feedback, to use Payload() to read the data so users can still read it again later if they want to.

sdk/azidentity/managed_identity_client.go
chlowell: Thanks, that's definitely better

jhendrixMSFT (approved):

PR #21353, opened 2023-08-11
Fix typo in doc comment for the TokenFilePath field in workload_identity

ahsonkhan:


sdk/azidentity/workload_identity.go
```diff
@@ -47,7 +47,7 @@ type WorkloadIdentityCredentialOptions struct {
 	DisableInstanceDiscovery bool
 	// TenantID of the service principal. Defaults to the value of the environment variable AZURE_TENANT_ID.
 	TenantID string
-	// TokenFilePath is the path a file containing the workload identity token. Defaults to the value of the
+	// TokenFilePath is the path of a file containing the workload identity token. Defaults to the value of the
```
chlowell, line 50: Hmm "workload identity token" isn't a thing, might as well clarify that while we're here
```suggestion
	// TokenFilePath is the path of a file containing a Kubernetes service account token. Defaults to the value of the
```

chlowell (approved):

PR #21367, opened 2023-08-14
ARM auth policy opts in to CAE

chlowell:
CAE is now disabled by default, so clients for RPs like ARM which support CAE must declare explicitly opt in. I renamed `acquire` to clarify that it's called only for auxiliary tenants. This would be an important distinction if it turns out we shouldn't request CAE tokens from auxiliary tenants.

### DIFF
--- a/sdk/azcore/arm/runtime/policy_bearer_token.go
+++ b/sdk/azcore/arm/runtime/policy_bearer_token.go
@@ -28,12 +28,12 @@ type acquiringResourceState struct {
 	tenant string
 }
 
-// acquire acquires or updates the resource; only one
-// thread/goroutine at a time ever calls this function
-func acquire(state acquiringResourceState) (newResource azcore.AccessToken, newExpiration time.Time, err error) {
+// acquireAuxToken acquires a token from an auxiliary tenant. Only one thread/goroutine at a time ever calls this function.
+func acquireAuxToken(state acquiringResourceState) (newResource azcore.AccessToken, newExpiration time.Time, err error) {
 	tk, err := state.p.cred.GetToken(state.ctx, azpolicy.TokenRequestOptions{
-		Scopes:   state.p.scopes,
-		TenantID: state.tenant,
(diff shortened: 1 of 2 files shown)


sdk/azcore/arm/runtime/policy_bearer_token.go
```diff
@@ -28,12 +28,12 @@ type acquiringResourceState struct {
...
 
-// acquire acquires or updates the resource; only one
-// thread/goroutine at a time ever calls this function
-func acquire(state acquiringResourceState) (newResource azcore.AccessToken, newExpiration time.Time, err error) {
+// acquireAuxToken acquires a token from an auxiliary tenant. Only one thread/goroutine at a time ever calls this function.
+func acquireAuxToken(state acquiringResourceState) (newResource azcore.AccessToken, newExpiration time.Time, err error) {
 	tk, err := state.p.cred.GetToken(state.ctx, azpolicy.TokenRequestOptions{
-		Scopes:   state.p.scopes,
-		TenantID: state.tenant,
+		EnableCAE: true,
```
jhendrixMSFT, line 34: To make sure I understand, when requesting a token from an auxiliary tenant, we always want CAE enabled even if it wasn't enabled for the client?

jhendrixMSFT: Oh wait, `EnableCAE` isn't set at the client level.

jhendrixMSFT, line 34: Does this warrant a changelog entry or will it be transparent to callers?

jhendrixMSFT (approved):

sdk/azcore/arm/runtime/policy_bearer_token.go
chlowell: It will be transparent. The prior azidentity enables CAE for all clients by default so even if someone upgrades azcore only, the feature will still work even though azidentity would ignore this field.

PR #21369, opened 2023-08-14
[keyvault] azadmin fix tests

gracewilcox:
Having test pipeline failures with the KeyVault Administration service. Not getting consistent error codes from the service, so changing to check that an error was thrown instead of the specific error message.

### DIFF
--- a/sdk/security/keyvault/azadmin/settings/client_test.go
+++ b/sdk/security/keyvault/azadmin/settings/client_test.go
@@ -48,8 +48,6 @@ func TestGetSetting_InvalidSettingName(t *testing.T) {
 	require.Nil(t, res.Value)
 	var httpErr *azcore.ResponseError
 	require.ErrorAs(t, err, &httpErr)
-	require.Equal(t, "UnknownError", httpErr.ErrorCode)
-	require.Equal(t, 500, httpErr.StatusCode)
 }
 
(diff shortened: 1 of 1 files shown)


sdk/security/keyvault/azadmin/settings/client_test.go
```diff
@@ -48,8 +48,6 @@ func TestGetSetting_InvalidSettingName(t *testing.T) {
 	require.Nil(t, res.Value)
 	var httpErr *azcore.ResponseError
 	require.ErrorAs(t, err, &httpErr)
-	require.Equal(t, "UnknownError", httpErr.ErrorCode)
-	require.Equal(t, 500, httpErr.StatusCode)
```
chlowell, line 52: 500 is a server error--which should be transient 😅--so it makes sense that MHSM now responds differently. We could probably just update this to require the new error on the assumption that MHSM's new behavior is the norm, however I think this is the right change because the specifics of the error aren't important in testing the client.

chlowell (approved):

sdk/security/keyvault/azadmin/settings/client_test.go
gracewilcox: Agree, specific error isn't important so we can just test if there was an error.

PR #21374, opened 2023-08-15
[azopenai] Add test for different formats in chat completions streaming 

richardpark-msft:
Also, updating CODEOWNERS to add @jhendrixMSFT to the `ai` folder.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -18,6 +18,9 @@
 # PRLabel: %Azure.Identity
 /sdk/azidentity/              @chlowell @jhendrixMSFT @rickwinter @Azure/azure-sdk-write-identity
 
+# PRLable: %OpenAI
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -18,6 +18,9 @@
 # PRLabel: %Azure.Identity
 /sdk/azidentity/              @chlowell @jhendrixMSFT @rickwinter @Azure/azure-sdk-write-identity
 
+# PRLable: %OpenAI
+/sdk/ai                       @jhendrixMSFT @richardpark-msft
```
jhendrixMSFT, line 22: I believe the convention here is the first entry is the primary owner and all others are secondary/backup owners.

jhendrixMSFT (approved):

.github/CODEOWNERS
richardpark-msft: Updated.

Also added in EventGrid.

Issue #21255, opened 2023-07-24
Module /sdk/security/keyvault/internal should be 1.0.0


jhendrixMSFT:
We have GA'ed versions of the keyvault modules that depend on a prerelease version of `/sdk/security/keyvault/internal`.  We shouldn't have GA versions with pre-GA dependencies.

PR #21376, opened 2023-08-15
[keyvault] upgraded dependencies

gracewilcox:
Upgraded all KeyVault modules to use the stable version of the KeyVault internal module.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21255


chlowell (approved):
It's just a test dependency but all the modules may as well use the latest azidentity, v1.3.1

PR #21378, opened 2023-08-15
Refactor handling of setting file expiry policy

jhendrixMSFT:
Switch to using a struct containing explicit fields instead of an interface to improve discovery and self-documentation.


sdk/storage/azdatalake/file/client_test.go
```diff
@@ -472,9 +474,11 @@ func (s *UnrecordedTestSuite) TestCreateFileWithExpiryAbsolute() {
 	defer testcommon.DeleteFilesystem(context.Background(), _require, fsClient)
 
 	expiryTimeAbsolute := time.Now().Add(8 * time.Second)
-	expiry := file.CreationExpiryTypeAbsolute(expiryTimeAbsolute)
 	createFileOpts := &file.CreateOptions{
-		Expiry: expiry,
+		Expiry: file.CreateExpiryValues{
+			ExpiresOn:  time.Now().Add(8 * time.Second).UTC().Format(http.TimeFormat),
```
jhendrixMSFT, line 479: Note that this does put the onus on the developer to get this correct.  We could add helper funcs to create a well-formed instance.  Maybe we do this later based on customer feedback.

tasherif-msft (approved):

tasherif-msft:
I'm ok with these changes. Please ignore the ci issues, these are resolved in another PR. I have gone ahead and approved

sdk/storage/azdatalake/file/client_test.go
tasherif-msft: Yea, let's see customer feedback. I will include examples in this in my PR after we merge this PR.

JeffreyRichter: I would just reverse the initialization order because the mental model is:
1. Select an expiry type first and then
2. Set ExpiresOn (if necessary)

sdk/storage/azdatalake/file/models.go
```diff
@@ -61,6 +59,18 @@ type CreateOptions struct {
 	ACL *string
 }
 
+// CreateExpiryValues describes when a newly created file should expire.
+// A zero-value indicates the file has no expiration date.
+type CreateExpiryValues struct {
+	// ExpiresOn contains the time the file should expire.
```
JeffreyRichter, line 65: I would just reverse the order of the fields in the struct because the mental model is:
1. Select an expiry type first and then
2. Set ExpiresOn (if necessary)

sdk/storage/azdatalake/file/models.go
```diff
@@ -491,65 +499,28 @@ func (o *DownloadFileOptions) format() *blob.DownloadFileOptions {
...
-
-func (e CreationExpiryTypeAbsolute) Format() (generated.ExpiryOptions, *string) {
-	return generated.ExpiryOptionsAbsolute, to.Ptr(time.Time(e).UTC().Format(http.TimeFormat))
+// SetExpiryValues describes when a file should expire.
+// A zero-value indicates the file has no expiration date.
+type SetExpiryValues struct {
+	// ExpiresOn contains the time the file should expire.
+	// The value will either be an absolute UTC time in RFC1123 format or an integer expressing a number of milliseconds.
+	// NOTE: when ExpiryType is SetExpiryTypeNeverExpire, this value is ignored.
+	ExpiresOn string
```
JeffreyRichter, line 508: Reverse order

JeffreyRichter (approved):

sdk/storage/azdatalake/file/models.go
jhendrixMSFT: Fixed in both places.

PR #21407, opened 2023-08-21
Fixed issue where customers could not capture response of directory and file GetProperties calls

gapra-msft:
This is duplication of the azcore includeResponsePolicy, this is necessary for some APIs like directory/file
GetProperties. Under the hood, these APIs grab the raw response to construct the datalake properties
(owner, group, permission) in addition to the blob properties. If we use the includeResponsePolicy, this can
result in the customer not being able to retrieve the response themselves.

### DIFF
--- a/sdk/storage/azdatalake/CHANGELOG.md
+++ b/sdk/storage/azdatalake/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed an issue where customers could not capture the raw HTTP response of directory and file GetProperties operations.
 
 ### Other Changes
 
--- a/sdk/storage/azdatalake/directory/client.go
+++ b/sdk/storage/azdatalake/directory/client.go
@@ -56,6 +56,11 @@ func NewClient(directoryURL string, cred azcore.TokenCredential, options *Client
 	if options == nil {
 		options = &ClientOptions{}
 	}
+	perCallPolicies := []policy.Policy{shared.NewIncludeBlobResponsePolicy()}
+	if options.ClientOptions.PerCallPolicies != nil {
+		perCallPolicies = append(perCallPolicies, options.ClientOptions.PerCallPolicies...)
+	}
+	options.ClientOptions.PerCallPolicies = perCallPolicies
 	blobClientOpts := blockblob.ClientOptions{
 		ClientOptions: options.ClientOptions,
 	}
@@ -84,6 +89,12 @@ func NewClientWithNoCredential(directoryURL string, options *ClientOptions) (*Cl
 	if options == nil {
 		options = &ClientOptions{}
 	}
+	perCallPolicies := []policy.Policy{shared.NewIncludeBlobResponsePolicy()}
+	if options.ClientOptions.PerCallPolicies != nil {
+		perCallPolicies = append(perCallPolicies, options.ClientOptions.PerCallPolicies...)
+	}
+	options.ClientOptions.PerCallPolicies = perCallPolicies
+
 	blobClientOpts := blockblob.ClientOptions{
 		ClientOptions: options.ClientOptions,
 	}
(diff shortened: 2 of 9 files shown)


sdk/storage/azdatalake/service/client.go
```diff
@@ -54,6 +54,11 @@ func NewClient(serviceURL string, cred azcore.TokenCredential, options *ClientOp
 	if options == nil {
 		options = &ClientOptions{}
 	}
+	perCallPolicies := []policy.Policy{shared.NewIncludeBlobResponsePolicy()}
```
siminsavani-msft, line 57: Is it possible to test the changes in the service/filesystem as well?

siminsavani-msft (approved):

sdk/storage/azdatalake/service/client.go
gapra-msft: yes, I added some tests that test that path but within the same directory/file test

gapra-msft: added tests within directory/file to test construction from each level

sdk/storage/azdatalake/directory/client.go
```diff
@@ -56,6 +56,11 @@ func NewClient(directoryURL string, cred azcore.TokenCredential, options *Client
 	if options == nil {
 		options = &ClientOptions{}
 	}
+	perCallPolicies := []policy.Policy{shared.NewIncludeBlobResponsePolicy()}
```
jhendrixMSFT, line 59: I believe this should go in the `plOpts.PerCall` field instead.

sdk/storage/azdatalake/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed an issue where customers could not capture the raw HTTP response of directory and file GetProperties operations.
```
jhendrixMSFT, line 10: Could I get more info on how this is broken?  Mostly I'm wondering if this is something we need to fix in `azcore`.

gapra-msft: So some datalake APIs call into the blob API. GetProperties on the blob endpoint returns both blob properties and some HNS properties like owner, group, etc. Since those properties are HNS only, they arent in the BlobGetPropertiesResponse, so it cant be directly converted to or aliased as the File/DirGetPropertiesResponse. The way we currently implement this is we call blob get properties and capture the raw response (with includeRawResponse context) to grab the specific HNS property headers. But if a customer also passes this context, we effectively override the same context key resulting in the raw response to be captured internally, and the customer is left with nil as the raw response. Basically we need to be able to handle multiple context keys to prevent this overriding, one for internal use and one for customer use.

sdk/storage/azdatalake/directory/client.go
gapra-msft: I dont think I have access to pipeline options here. It's in the internal package.

jhendrixMSFT: See above, line 47 where the auth policy is added to `PerRetry`.

Generally speaking, SDK authors should put their policies in `runtime.PipelineOptions`, leaving `ClientOptions` for calling code.

sdk/storage/azdatalake/CHANGELOG.md
jhendrixMSFT: Got it, thanks for the explanation.

sdk/storage/azdatalake/directory/client.go
gapra-msft: the plOpts field above is for the datalake client. We need the custom blob response policy in the blob client code (which we are constructing as a caller)

jhendrixMSFT: Oh yes I see now, makes sense.

jhendrixMSFT (approved):

sdk/storage/azdatalake/directory/client.go
```diff
@@ -230,7 +247,7 @@ func (d *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 func (d *Client) GetProperties(ctx context.Context, options *GetPropertiesOptions) (GetPropertiesResponse, error) {
 	opts := path.FormatGetPropertiesOptions(options)
 	var respFromCtx *http.Response
-	ctxWithResp := runtime.WithCaptureResponse(ctx, &respFromCtx)
+	ctxWithResp := shared.WithCaptureBlobResponse(ctx, &respFromCtx)
 	resp, err := d.blobClient().GetProperties(ctxWithResp, opts)
```
souravgupta-msft, line 251: Why are we calling GetProperties on the blob client here, instead we could call it for the generated PathClient. This way we will get the HNS properities in the resp struct.
This might not be related to your PR changes, but we should discuss on this design.

PR #21409, opened 2023-08-21
Update AzBlob Service Version to 2023-08-03

siminsavani-msft:
Created an enum `ServiceVersion` to make it easier to update the service version.

`ServiceVersion` replaces all instances of `[]string{"2021-12-02"}` in blob, appendblob, blockblob, pageblob, container, and service client.

### DIFF
--- a/sdk/storage/azblob/blockblob/client_test.go
+++ b/sdk/storage/azblob/blockblob/client_test.go
@@ -5498,7 +5498,7 @@ func newServiceVersionTestPolicy() policy.Policy {
 func (m serviceVersionTest) Do(req *policy.Request) (*http.Response, error) {
 	const versionHeader = "x-ms-version"
 
-	currentVersion := req.Raw().Header[versionHeader]
+	currentVersion := map[string][]string(req.Raw().Header)[versionHeader]
 	if currentVersion[0] != generated.ServiceVersion {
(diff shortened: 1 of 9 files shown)


gapra-msft (approved):

jhendrixMSFT (approved):

sdk/storage/azblob/blockblob/client_test.go
```diff
@@ -5486,3 +5487,39 @@ func TestRequestIDGeneration(t *testing.T) {
...
+}
+
+func (m serviceVersionTest) Do(req *policy.Request) (*http.Response, error) {
+	const versionHeader = "x-ms-version"
+
+	currentVersion := map[string][]string(req.Raw().Header)[versionHeader]
+	if currentVersion[0] != generated.ServiceVersion {
+		return nil, fmt.Errorf(currentVersion[0] + " service version doesn't match expected version: " + generated.ServiceVersion)
+	}
+	return req.Next()
```
gapra-msft, line 5505: You don't really have to return req.Next(). That will make the test actually hit the service wont it?

What you can also do is just call GetProperties, that way we don't have to create data unnecessarily.

siminsavani-msft: That's true, I changed it to a fake response being returned.

PR #21410, opened 2023-08-21
[keyvault] fix doc comments

gracewilcox:
When we changed some of the names, not all the doc comments were changed to match the new names. This pull request that issue and updates the doc comments


chlowell:
What do you think of rewriting the comments to remove the redundant type name e.g. "// Elliptic curve name. ~~For valid values, see CurveName.~~"? I'm thinking, people reading the comment have the type or its name in front of them, and in the rendered docs type names link to definitions

chlowell (approved):

PR #21426, opened 2023-08-23
[azopenai] Adding in support for using Azure Chat Extensions with external data sources

### LINKED ISSUE #21373: [azopenai] Restructure ChatCompletionOptions to allow for Azure extensions
We're adding in a specific area to make it clear when options are specific to Azure OpenAI vs OpenAI. Just need to restructure our options to match it as the REST endpoint does not currently match it.

richardpark-msft:
Adding in support for chat extensions, based on the pre-release .tsp that @trrwilson created. 

This'll be close or exactly the same as what gets merged and I'll create a second PR to update when it does.

Fixes #21373


richardpark-msft:
(have to add recordings real quick and add some extra live config variables)

richardpark-msft:
(rerunning live build - I didn't update the remote variable group properly)

sdk/ai/azopenai/autorest.md
```diff
@@ -294,4 +303,32 @@ directive:
       return $
         .replace(/populate\(objectMap, "model", (c|e).Model\)/g, 'populate(objectMap, "model", &$1.Deployment)')
         .replace(/err = unpopulate\(val, "Model", &(c|e).Model\)/g, 'err = unpopulate(val, "Model", &$1.Deployment)');
+
+  # - from: client.go
```
jhendrixMSFT, line 307: Can this be deleted or is it temporary?

richardpark-msft: Yeah, you're right. The directives below handle the real renaming now (there was an awkward intermediate state here).

sdk/ai/azopenai/models.go
```diff
@@ -8,6 +8,114 @@
...
 
+import "time"
+
+// AzureChatExtensionConfiguration - A representation of configuration data for a single Azure OpenAI chat extension. This
+// will be used by a chat completions request that should use Azure OpenAI chat extensions to augment the response
+// behavior. The use of this configuration is compatible only with Azure OpenAI.
+type AzureChatExtensionConfiguration struct {
+	// REQUIRED; The configuration payload used for the Azure chat extension. The structure payload details are specific to the
+	// extension being configured. Azure chat extensions are only compatible with Azure OpenAI.
+	Parameters any
```
jhendrixMSFT, line 19: We've been trying to use `[]byte` for arbitrary JSON (there's a codegen switch for it).  Do you recall why we didn't enable that for this SDK?

richardpark-msft: In this case I think modeling it as a []byte would feel like I'm just pushing extra work on the user. It's really a discriminated union with a set of types that we provide.

Usage looks like this:

https://github.com/Azure/azure-sdk-for-go/pull/21426/files#diff-98ceffdea7d8f77a889cca1f345f7640290095f575388611bb09f996e10366f6R69

```go
{
  Type: to.Ptr(azopenai.AzureChatExtensionTypeAzureCognitiveSearch),
  Parameters: azopenai.AzureCognitiveSearchChatExtensionConfiguration{
    Endpoint:  &searchEndpoint,
    IndexName: &searchIndex,
    Key:       &searchAPIKey,
},
```

So we are providing the type they can use there, and will probably add more as different items are allowed as data sources.

jhendrixMSFT: Is it always limited to the types we generate?

richardpark-msft: "Yes" (in that that's what it's modeled as today, and I've heard nothing about new data source types yet)

jhendrixMSFT: I see.  At the very least, it seems like there should be a comment indicating that the value needs to be one of the defined types.  You could almost envision this type being generic, with a type constraint for the generated types.  Of course, that would mean that one couldn't use other types that become available before the SDK is updated.

richardpark-msft: > You could almost envision AzureChatExtensionConfiguration being generic, with a type constraint for the generated types.

YES except that it'd force the entire client to have to accept that type constraint as well since this is a "method".

I'll update the docs, which apparently I'll have to do by hand. This might be a nice case to consider in the generator, if we can preserve something when we do tsp directly to Go (if it knows the domain of known types).

richardpark-msft: Looking at this I think I'll add the doc about the type to `AzureChatExtensionType`'s description since that's how you'd drive it - you pick the 'type' first, then you know what underlying parameter type should be used (at least, that's what I know as of now).

jhendrixMSFT: Good point WRT making `AzureChatExtensionConfiguration` generic.  Hopefully this restriction will go away in the future.

Given your description, the type should go first in the struct.  Might be tricky to achieve given that this is generated code.  Also, if the tsp fixed its docs then we'd get the list of applicable values for free.

richardpark-msft: Fixed. More autorest.md hackery.

jhendrixMSFT (approved):

PR #21429, opened 2023-08-23
[keyvault] change managed hsms to run weekly

gracewilcox:
Closes https://github.com/Azure/azure-sdk-for-go/issues/20733

Changes live tests on managed HSM's to only run once a week.


sdk/security/keyvault/azadmin/ci.yml
```diff
@@ -31,4 +31,10 @@ stages:
...
         Selection: sparse
-        GenerateVMJobs: true
\ No newline at end of file
+        GenerateVMJobs: true
+
+    ${{ if not(contains(variables['Build.DefinitionName'], 'tests-weekly')) }}:
+        # Due to the high cost of Managed HSMs, which keyvault-admin requires, we only want to run
+        # the live tests weekly.
+        MatrixFilters:
+          - ArmTemplateParameters=^(?!.*enableHsm.*true)
```
chlowell, line 40: Could this expression be simpler e.g. just set `enableHsm=false`? Also, what happens to the daily internal pipelines when enableHsm is false? Do they skip the live job or does it fail because environment variables aren't set? Since azadmin requires a MHSM, should we just cancel its daily pipelines?

gracewilcox: Can try to refactor so that enableHsm=false, copied this statement over from .NET and js yml files.

When it's false, it just skips that live job. ([example pipeline run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3035153&view=results)) I think we should still have it set to run daily because it'll run the recorded tests.

gracewilcox: Talked to @timovv, and we need to leave `ArmTemplateParameters=^(?!.*enableHsm.*true)` as is. It is actually a regex expression that filters out matrix configs  where hsm's are enabled like so: https://github.com/Azure/azure-sdk-for-go/blob/dfff8e9837c0fe0b35fc1c2a94f6bbcb12c0d115/sdk/security/keyvault/azkeys/platform-matrix.json#L13

chlowell, line 40:
```suggestion
          - ArmTemplateParameters=^(?!.*enableHsm.*true)

```

chlowell (approved):

PR #21432, opened 2023-08-24
Handle Docker Desktop 403 response from IMDS endpoint

### LINKED ISSUE #21417: IMDS endpoint within Docker Desktop from a Linux container fails with 403
Tracking issue to check your language for this Docker Desktop edge case and apply a fix similar to Azure Identity for .NET

https://github.com/Azure/azure-sdk-for-net/issues/38218#issuecomment-1683085316

Here is a curl command to the IMDS endpoint run on my repro machine across 3 different environments

`curl -I http://169.254.169.254`

On Windows:
`curl: (7) Failed to connect to 169.254.169.254 port 80 after 0 ms: Couldn't connect to server`

On WSL2 Ubuntu:
`curl: (28) Failed to connect to 169.254.169.254 port 80: Connection timed out`

On Docker Desktop from a Linux container:
`HTTP/1.0 403 connecting to 169.254.169.254:80: connecting to 169.254.169.254:80: dial tcp 169.254.169.254:80: connectex: A socket operation was attempted to an unreachable network.`
`Connection: close`

Fix: https://github.com/Azure/azure-sdk-for-net/pull/38321

chlowell:
We want credential chains to try their next credential, if any, when a managed identity isn't available i.e., we want `ManagedIdentityCredential` to return `credentialUnavailableError`. This requires a special case for Docker Desktop containers because some component of that infrastructure responds 403 Forbidden to IMDS token requests. Closes #21417

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -175,15 +175,25 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKi
 		return c.createAccessToken(resp)
 	}
 
-	if c.msiType == msiTypeIMDS && resp.StatusCode == 400 {
-		if id != nil {
-			return azcore.AccessToken{}, newAuthenticationFailedError(credNameManagedIdentity, "the requested identity isn't assigned to this resource", resp, nil)
+	if c.msiType == msiTypeIMDS {
+		switch resp.StatusCode {
+		case http.StatusBadRequest:
+			if id != nil {
+				return azcore.AccessToken{}, newAuthenticationFailedError(credNameManagedIdentity, "the requested identity isn't assigned to this resource", resp, nil)
+			}
+			msg := "failed to authenticate a system assigned identity"
+			if body, err := runtime.Payload(resp); err == nil && len(body) > 0 {
+				msg += fmt.Sprintf(". The endpoint responded with %s", body)
+			}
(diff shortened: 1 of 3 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -175,15 +175,25 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKi
...
+			msg := "failed to authenticate a system assigned identity"
+			if body, err := runtime.Payload(resp); err == nil && len(body) > 0 {
+				msg += fmt.Sprintf(". The endpoint responded with %s", body)
+			}
+			return azcore.AccessToken{}, newCredentialUnavailableError(credNameManagedIdentity, msg)
+		case http.StatusForbidden:
+			// Docker Desktop runs a proxy that responds 403 to IMDS token requests. If we got that response,
+			// we return credentialUnavailableError so credential chains continue to their next credential
+			body, err := runtime.Payload(resp)
+			if err == nil && strings.HasSuffix(string(body), "connectex: A socket operation was attempted to an unreachable network.") {
```
jhendrixMSFT, line 193: I wonder if `HasSuffix` is too specific.  Should we do something like what the .NET folks did.
```
if (message.Contains("A socket operation was attempted to an unreachable network"))
```

christothes, line 193: In my local testing, the last bit out output would be "Connection: close", so perhaps we should do a contains rather than suffix?

RickWinter (approved):

PR #21451, opened 2023-08-28
Transport dialer: setting nil for wasm build

magodo:
This PR is to make this SDK to work as an WASM module. The implementation is refered from https://github.com/golang/go/blob/071aed2aaa0ed819582c5bff44b70d43c61f504a/src/net/http/transport.go#L45.

### DIFF
--- a/sdk/azcore/runtime/transport_default_dialer_other.go
+++ b/sdk/azcore/runtime/transport_default_dialer_other.go
@@ -1,0 +1,12 @@
+//go:build !wasm
+
+package runtime
+
+import (
+	"context"
+	"net"
+)
+
+func defaultTransportDialContext(dialer *net.Dialer) func(context.Context, string, string) (net.Conn, error) {
+	return dialer.DialContext
+}
--- a/sdk/azcore/runtime/transport_default_dialer_wasm.go
+++ b/sdk/azcore/runtime/transport_default_dialer_wasm.go
@@ -1,0 +1,12 @@
+//go:build (js && wasm) || wasip1
+
+package runtime
+
+import (
+	"context"
+	"net"
+)
+
+func defaultTransportDialContext(dialer *net.Dialer) func(context.Context, string, string) (net.Conn, error) {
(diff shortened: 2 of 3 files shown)


jhendrixMSFT:
Can we get more info on what this fixes and why it's needed please?

magodo:
@jhendrixMSFT I'm trying to build some toolings that are based on this SDK to WASM to be able to run in the browser. Say I have the following simplistic code:

```go
func main() {
	tenantID := os.Getenv("ARM_TENANT_ID")
	clientID := os.Getenv("ARM_CLIENT_ID")
	secret := os.Getenv("ARM_CLIENT_SECRET")

	cred, err := azidentity.NewClientSecretCredential(tenantID, clientID, secret, nil)
	if err != nil {
		log.Fatal(err)
	}
	tk, err := cred.GetToken(context.TODO(), policy.TokenRequestOptions{
		Scopes: []string{"https://management.core.windows.net//.default"},
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(tk)
}
```

It failed with following error:

```
2023/08/29 09:20:38 ClientSecretCredential: unable to resolve an endpoint: server response error:
 Get "https://login.microsoftonline.com/xxxx/v2.0/.well-known/openid-configuration": dial tcp: lookup login.microsoftonline.com: Protocol not available
```

The reason is that the browser APIs do not allow arbitrary TCP/IP connections and therefore WASM is equally limited. The Go standard `net/http` library hence conditionally build the default client to mitigate this issue.

So with this change, we can then successfully send above request. Though, in order to have successfully get the token, it will need some [other tricks](https://github.com/golang/go/issues/61889) to make it work. It is another story off this topic then.

jhendrixMSFT (approved):

magodo:
Can we settle done this PR soon? As the CI sends a ton of notifications for some weired faliure: https://github.com/magodo/azure-sdk-for-go/actions/runs/6048886404

Issue #21436, opened 2023-08-24
Upgrade the module version used in live test when the sdk has major version bump


raych1:
### Feature Request

When sdk has major version bump, the live test needs to update the module referenced in the code. Not every service has live test, but if it exists then it needs to update manually. We would like to have this to be supported by the automatic tool set.

This is an example PR: https://github.com/Azure/azure-sdk-for-go/pull/21434.

// @tadelesh

PR #21464, opened 2023-08-30
Upgrade the module version used in live test

Alancere:
https://github.com/Azure/azure-sdk-for-go/issues/21436

### DIFF
--- a/eng/tools/generator/cmd/v2/common/fileProcessor.go
+++ b/eng/tools/generator/cmd/v2/common/fileProcessor.go
@@ -566,3 +566,28 @@ func replaceModuleImport(path, rpName, namespaceName, previousVersion, currentVe
 		return nil
 	})
 }
+
+func existSuffixFile(path, suffix string) bool {
+
+	existed := false
+	err := filepath.WalkDir(path, func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		if strings.HasSuffix(d.Name(), suffix) {
+			existed = true
+			return nil
+		}
+		return nil
+	})
+	if err != nil {
+		return false
+	}
+
+	return existed
+}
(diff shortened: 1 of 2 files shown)


eng/tools/generator/cmd/v2/common/generation.go
```diff
@@ -306,6 +306,15 @@ func (ctx *GenerateContext) GenerateForSingleRPNamespace(generateParam *Generate
 			}
 		}
 
+		// When sdk has major version bump, the live test needs to update the module referenced in the code.
+		if changelog.HasBreakingChanges() && existSuffixFile(packagePath, "_live_test.go") {
```
raych1, line 310: Does `HasBreakingChanges` function only return true when having major version bump with breaking changes?

Alancere: It will only get whether there is a BreakingChange in the Changelog

eng/tools/generator/cmd/v2/common/fileProcessor.go
```diff
@@ -566,3 +566,28 @@ func replaceModuleImport(path, rpName, namespaceName, previousVersion, currentVe
...
+			return err
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		if strings.HasSuffix(d.Name(), suffix) {
+			existed = true
+			return nil
```
raych1, line 584: How about removing this extra line of code?

eng/tools/generator/cmd/v2/common/generation.go
raych1: In that case of minor version update, do we need to update live test module version?

eng/tools/generator/cmd/v2/common/fileProcessor.go
Alancere: removed

eng/tools/generator/cmd/v2/common/generation.go
Alancere: Yes, hasBreakingChange here only determines that there may be a major version bump. The details of upgrading the live test module are in the replaceModuleImport function.

tadelesh (approved):

raych1 (approved):

Issue #21346, opened 2023-08-10
Outbound calls timing out with `context deadline exceeded` when TCP connection is been drop


julienstroheker:
### Bug Report

Opening this for visibility in case other folks are having this issue !

TLDR : https://github.com/golang/go/issues/59690

- What happened?

When using HTTP/2, which is defaulted everywhere now, if the keep-alive connection is been dropped somehow, the sdk clients cannot recover causing some timeout `context deadline exceeded` until the connection been closed by the host.

- What did you expect or want to happen ?

When using a HTTP/2 server, CosmosDB for example, the SDK transport [should PING the server to identify is the connection is not stale](https://pkg.go.dev/golang.org/x/net/http2#PingFrame).

- How can we reproduce it?

I'm able to reproduce this using CosmosDB sample code such as :

```golang
package main

import (
	"context"
	"crypto/tls"
	"log"
	"net"
	"net/http"
	"time"

	logger "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
)

const (
	cosmosDbEndpoint = "https://julienstestcosmos.documents.azure.com"
	cosmosDbKey      = "REDACTED"
	dbName           = "julienstest"
	containerName    = "Subscriptions"
	pk               = "c88b61d7-85aa-415f-8eda-4e73e5e811d5"
	id               = "c88b61d7-85aa-415f-8eda-4e73e5e811d5"
)

func main() {
	ctx := context.Background()
	opt := &azcosmos.ClientOptions{ClientOptions: policy.ClientOptions{
		// we call cosmosDB more frequent and CosmosDB's latency is much less than ARM
		// so we use difference retry policy than ARM
		Retry: policy.RetryOptions{
			// Create/Update/Delete option will get ItemAlreadyExists error in second retry if first one already arrive server.
			StatusCodes: []int{
				// azcore defaults, see RetryOptions doc
				http.StatusRequestTimeout,      // 408
				http.StatusTooManyRequests,     // 429
				http.StatusInternalServerError, // 500
				http.StatusBadGateway,          // 502
				http.StatusServiceUnavailable,  // 503
				http.StatusGatewayTimeout,      // 504
				// adding 403 to work around random 403 from cosmosdb
				// see https://github.com/Azure/azure-sdk-for-go/issues/19785
[...]

PR #21468, opened 2023-08-30
Enable health checks for HTTP/2 connections

jhendrixMSFT:
If the health check fails, the connection will be closed and a new one created.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21346

### DIFF
--- a/sdk/azcore/runtime/transport_default_http_client.go
+++ b/sdk/azcore/runtime/transport_default_http_client.go
@@ -39,8 +39,8 @@ func init() {
 	if http2Transport, err := http2.ConfigureTransports(defaultTransport); err == nil {
 		// if the connection has been idle for 10 seconds, send a ping frame for a health check
 		http2Transport.ReadIdleTimeout = 10 * time.Second
-		// if there's no response to the ping within 2 seconds, close the connection
-		http2Transport.PingTimeout = 2 * time.Second
+		// if there's no response to the ping within the timeout, the connection will be closed
+		http2Transport.PingTimeout = 5 * time.Second
 	}
 	defaultHTTPClient = &http.Client{
 		Transport: defaultTransport,
--- a/sdk/azcore/runtime/transport_default_http_client.go
+++ b/sdk/azcore/runtime/transport_default_http_client.go
@@ -11,6 +11,8 @@ import (
 	"net"
 	"net/http"
 	"time"
+
+	"golang.org/x/net/http2"
 )
 
 var defaultHTTPClient *http.Client
@@ -31,6 +33,12 @@ func init() {
 			MinVersion:    tls.VersionTLS12,
 			Renegotiation: tls.RenegotiateFreelyAsClient,
 		},
+	}
+	if http2Transport, err := http2.ConfigureTransports(defaultTransport); err == nil {
+		// if the connection has been idle for 10 seconds, send a ping frame for a health check
+		http2Transport.ReadIdleTimeout = 10 * time.Second
+		// if there's no response to the ping within 2 seconds, close the connection
+		http2Transport.PingTimeout = 2 * time.Second
 	}
 	defaultHTTPClient = &http.Client{
 		Transport: defaultTransport,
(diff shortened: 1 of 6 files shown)


sdk/azcore/runtime/transport_default_http_client.go
```diff
@@ -32,6 +34,12 @@ func init() {
 			Renegotiation: tls.RenegotiateFreelyAsClient,
 		},
 	}
+	if http2Transport, err := http2.ConfigureTransports(defaultTransport); err == nil {
+		// if the connection has been idle for 10 seconds, send a ping frame for a health check
+		http2Transport.ReadIdleTimeout = 10 * time.Second
+		// if there's no response to the ping within 2 seconds, close the connection
+		http2Transport.PingTimeout = 2 * time.Second
```
julienstroheker, line 41: Just mentioning that this assumes that the server supports this. 
Since this defaulted transport is used across SDK for Control Plane and DP services (KV, EventGrid, cosmos etc...) are we sure they are all supporting this ? I assume yes (at least ARM and cosmosdb for sure) but just mentioning it.

julienstroheker: Also, might out of scope of that PR, but bumping the MaxIddleConnsPerHost from 2 (default) to a bit more is giving some slack to the transport multiplexing. 

Let's say you do only connections to ARM endpoint, then you'll have more than 2 connections handled in the pool that reduce the number of stale connections that can happens.

jhendrixMSFT: That's a good point about idle connections, I can bump that up a bit.

Agreed on the risk WRT RP support and am trying to perform more live validation to gain confidence.  Worse case, we enable this for the select few clients that have been bumping into this problem.

We're also following up with the KV team about this as our clients have reliably worked for years, so it's a bit odd that this suddenly started happening.

julienstroheker: For us, this is happening for more than 3 years at random time. It can happens for so many reasons where packet is been dropped : server not able to answer because of load, connections not been gracefully shutdown, NLB dropping connections, Unix kernel, etc... 
Until now, we were forcing http/1 with retries but this is a bummer.

jhendrixMSFT:
@tadelesh what live tests do we have for ARM SDKs that I can run with this change?

sdk/azcore/runtime/transport_default_http_client.go
jhendrixMSFT: Thanks for the context.  It is odd that this suddenly started happening for KV though.

FWIW the `http.Transport` used here invokes the bundled version of `http2.ConfigureTransports()` so the code paths are almost the same, we're just performing this up front and configuring some knobs.

Testing is still in progress.

julienstroheker: I think it mainly depends how heavy you are using the connection: Eg: You interact with KV DP once a day vs every sec ? The more interaction are happening the more chance you'll have to hit this issue.

jhendrixMSFT: By not supported, do you mean that the service might not respond to pings?

julienstroheker: Yep

jhendrixMSFT: Ugh I see.  So, I suppose we could have a perfectly good connection and if the host doesn't respond to pings, then the transport thinks it's dead and recycles it.  I will follow up with the ARM folks about this to get clarification.  Obviously, this could extend beyond ARM though.  Makes me think that perhaps this should be an opt-in feature.

julienstroheker: ARM does support H2 ping and "my team" never hit an issue like this on any control plane RP operation. This is mainly happening on any other DP endpoints where ARM is not in the middle. 
I assume ARM does a good job at cleaning connection. Might be interesting to ask them for sure.

chlowell (approved):

weinong:
@jhendrixMSFT does this change apply to azidentity?

PR #21471, opened 2023-08-30
Execute go mod tidy for perf tests when using live azcore

jhendrixMSFT:
This ensures that any updated dependencies are propagated.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -83,12 +83,10 @@ steps:
 
   - task: PowerShell@2
     displayName: 'Build Performance Tests'
-    variables:
-      useAzcoreFromMain: eq(variables['Use.AzcoreFromMain'], 'true')
     inputs:
       targetType: 'filePath'
       filePath: ./eng/scripts/Build_Perf.ps1
-      arguments: '${{ parameters.ServiceDirectory }} $(useAzcoreFromMain)'
+      arguments: '${{ parameters.ServiceDirectory }} $(Use.AzcoreFromMain)'
       pwsh: true
 
   - ${{ if eq(parameters.TestProxy, true) }}:
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -86,7 +86,7 @@ steps:
     inputs:
       targetType: 'filePath'
       filePath: ./eng/scripts/Build_Perf.ps1
-      arguments: '${{ parameters.ServiceDirectory }} $$(Use.AzcoreFromMain)'
+      arguments: '${{ parameters.ServiceDirectory }} $[variables.Use.AzcoreFromMain]'
       pwsh: true
 
   - ${{ if eq(parameters.TestProxy, true) }}:
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -86,7 +86,7 @@ steps:
     inputs:
       targetType: 'filePath'
       filePath: ./eng/scripts/Build_Perf.ps1
-      arguments: '${{ parameters.ServiceDirectory }} ${{ variables['Use.AzcoreFromMain'] }}'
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -86,7 +86,7 @@ steps:
     inputs:
       targetType: 'filePath'
       filePath: ./eng/scripts/Build_Perf.ps1
-      arguments: '${{ parameters.ServiceDirectory }}'
+      arguments: '${{ parameters.ServiceDirectory }} ${{ variables['Use.AzcoreFromMain'] }}'
```
benbp, line 89: The parameters vs. variables syntax is annoying. You have to do it this way to delay resolution of the symbol, otherwise it will try to calculate the variable before it is set when it renders the yaml to pass to the worker agent. You also need to add an extra `$` so it renders as valid powershell (`$true` vs. `true`).

```suggestion
      arguments: '${{ parameters.ServiceDirectory }} $$(Use.AzcoreFromMain)'
```

However, you also will then need to default it to false [here](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/variables/globals.yml). Otherwise devops will fail to find a variable and will pass the literal `$(<var>)` which will fail powershell execution due to syntax...

Yeah devops is special...we typically try to limit use of variables for this reason, but we have a good reason to add it here. The other approach you could consider is just checking for the existence of an env variable in the perf script instead, but since the variable has a `.` you have to do more fancy things to look it up (or remove the `.` from the var name)...

jhendrixMSFT: Thanks, I've made the change.

I like that the perf script takes it as a parameter so it's somewhat clear this is an option.

jhendrixMSFT: Should I remove the default value from the build_perf.ps1?

benbp: I would keep it there so it can be run in other contexts without the extra param, if needed.

eng/pipelines/templates/variables/globals.yml
```diff
@@ -8,3 +8,6 @@ variables:
   
   # Disable CodeQL injections except for where we specifically enable it
   Codeql.SkipTaskAutoInjection: true
+
+  # Set the default value so that 'Build Performance Tests' step can resolve it
+  Use.AzcoreFromMain: false
```
jhendrixMSFT, line 13: Looks like this value isn't getting overwritten when I define it queueing up a pipeline run.

benbp: It's possible my suggested fix isn't quite working. Do you have a link to the pipeline run? It doesn't seem to get linked to the PR (azure pipelines has been dropping the ball here lately).

jhendrixMSFT: [Here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3046116&view=results) is the run from last night.

I'm looking at [these docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#runtime-expression-syntax) and it seems like the `$[variables.foo]` syntax is what we want.  I tried that and it gets us closer but still not completely there.  Latest run [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3048169&view=logs&j=730d91e9-ccca-5f92-9343-3b1124b19442&t=57014ab0-c321-5206-21d1-de9826f53695).

jhendrixMSFT: Looks like defining a var in the YAML [won't allow it to be overridden](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#allow-at-queue-time) at queue time.

jhendrixMSFT:
Finally got this working, [here's](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3048469&view=results) the run from the internal pipeline.

benbp (approved):

PR #21474, opened 2023-08-31
Rename "send_and_receive drain.go" to remove space

flyn-org:
The space in "send_and_receive drain.go" made it more difficult to work on the source code using various Unix tools, such as combinations of `find` and other utilities.


jhendrixMSFT (approved):

richardpark-msft (approved):
Thanks for that, clearly a typo :)

PR #21496, opened 2023-09-05
[azservicebus] Fixing issue where the body was populated incorrectly, making it empty on retries since it can't be rewound.

### LINKED ISSUE #21489: [azservicebus] Throttled HTTP requests don't include a body when retried
### Bug Report

SDK version:
github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.0
github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0
github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus v1.4.0

output of `go version`:
go version go1.20.4 windows/amd64

### What happened?
Hey there! 😺
We're using the `azservicebus` SDK in a [terraform provider](https://github.com/DigitecGalaxus/terraform-provider-dg-servicebus) to manage Queues and Subscriptions for our Service Bus instances. After deploying it, we noticed very odd, hard to reproduce issues when creating Service Bus resources. Create and update operations fail, seemingly at random, in our production environment, while the same issues never occur during local development.

The errors logged from the SDK client are also not very helpful, because the failure reason provided by the Azure Management API does not fit with what we were doing. Most errors come from either of these responses:
- 400 Bad Request "The specified resource description is invalid"
- 401 Unauthorized "Azure Active Directory RBAC is not supported for entity type 'Unknown'"
> We also encountered some 409 Conflict "Another conflicting operation my be in progress" errors, though I'm not sure if those should be handled by the SDK

The latter of which made no sense, as other requests using the same instance of the admin client, on the same resource, worked fine.

We eventually managed to reproduce the issue locally by reading and creating a bunch of resources, much like in production and managed capture the http requests leading to one of the errors mentioned above:

#### Request

```http
PUT https://some-service-bus.servicebus.windows.net/some-servicebus-queue?api-version=2021-05 HTTP/1.1
Host: some-service-bus.servicebus.windows.net
User-Agent: azsdk-go-azsbadmin/v1.4.0 (go1.19.12; Windows_NT)
Transfer-Encoding: chunked
Authorization: eyxxx
Content-Type: application/atom+xml;type=entry;charset=utf-8
Accept-Encoding: gzip

1d8
<entry xmlns="http://www.w3.org/2005/Atom"><content type="application/xml"><QueueDescription xmlns="http://schemas.microsoft.com/netservices/2010/10/servicebus/connect"><LockDuration>PT5M</LockDuration><MaxSizeInMegabytes>5120</MaxSizeInMegabytes><MaxDeliveryCount>2147483647</MaxDeliveryCount><EnableBatchedOperations>true</EnableBatchedOperations><AuthorizationRules></AuthorizationRules><EnablePartitioning>true</EnablePartitioning></QueueDescription></content></entry>
0
```

#### Response

```http
HTTP/1.1 503 Service Unavailable
Transfer-Encoding: chunked
Content-Type: application/xml; charset=utf-8
Server: Microsoft-HTTPAPI/2.0
Strict-Transport-Security: max-age=31536000
Date: Thu, 31 Aug 2023 13:09:31 GMT

1DE
<Error><Code>503</Code><Detail>SubCode=50004. The request was terminated because the namespace some-service-bus is being throttled. Error code : 50009. Please wait 2 seconds and try again. To know more visit https://aka.ms/sbResourceMgrExceptions and https://aka.ms/ServiceBusThrottling TrackingId:a57a3354-92fa-4d01-8480-59358b873c88_G67_B5, SystemTracker:some-service-bus:Queue:some-servicebus-queue~111, Timestamp:2023-08-31T13:09:31</Detail></Error>
0
```

#### Retry

```http
PUT https://some-service-bus.servicebus.windows.net/some-servicebus-queue?api-version=2021-05 HTTP/1.1
Host: some-service-bus.servicebus.windows.net
User-Agent: azsdk-go-azsbadmin/v1.4.0 (go1.19.12; Windows_NT)
Transfer-Encoding: chunked
Authorization: eyxxx
Content-Type: application/atom+xml;type=entry;charset=utf-8
Accept-Encoding: gzip

0
```

#### Retry response

```http
HTTP/1.1 401 Unauthorized
Transfer-Encoding: chunked
Content-Type: application/xml; charset=utf-8
Server: Microsoft-HTTPAPI/2.0
Strict-Transport-Security: max-age=31536000
Date: Thu, 31 Aug 2023 13:09:33 GMT

F2
<Error><Code>401</Code><Detail>Generic: Azure Active Directory RBAC is not supported for entity type 'Unknown'. TrackingId:3bc5400b-561e-4e68-b6df-6084882a558b_G29, SystemTracker:NoSystemTracker, Timestamp:2023-08-31T13:09:33</Detail></Error>
0
```
---

As you can see, the initial request was rejected because the Service Bus is being throttled. This should be fine, as the azservicebus client retries the request in those cases. The subsequent retry request however, does not seem to contain any body. The resulting error also seems rather strange, though that is not an issue with the SDK. This behavior would also explain why arbitrary requests result in a 400 Bad Request, because an empty body is of course "an invalid resource description". ^^"

### How can we reproduce it?
The issue seems to occur, when the Azure Service Bus runs out of MUs for its quota and therefore throttles requests. If, during this time, the azservicebus admin client sends a request containing a body, this body seems to not be included in the retry request. This issue was very hard to reproduce on our end and we ended up wrapping the admin client requests with custom retry logic, which has fixed the issue.

I'm happy to provide any additional information :)

richardpark-msft:
Previously we were setting request.Raw().Body directly, instead of using the request.SetBody() function. 

This made it so the request body itself was no longer rewindable, which made subsequent requests send an empty body instead.

Fixes #21489

### DIFF
--- a/sdk/messaging/azservicebus/internal/atom/entity_manager.go
+++ b/sdk/messaging/azservicebus/internal/atom/entity_manager.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/auth"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/conn"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/sbauth"
@@ -159,15 +160,6 @@ type ExecuteOptions struct {
 	ForwardToDeadLetter *string
 }
 
-// nopCloser copied from azblob.
-type nopCloser struct {
-	io.ReadSeeker
-}
-
-func (n nopCloser) Close() error {
-	return nil
-}
-
 func (em *entityManager) execute(ctx context.Context, method string, entityPath string, body io.ReadSeeker, options *ExecuteOptions) (*http.Response, error) {
 	url := em.Host + strings.TrimPrefix(entityPath, "/")
 
@@ -184,7 +176,7 @@ func (em *entityManager) execute(ctx context.Context, method string, entityPath 
 	req.Raw().URL.RawQuery = q.Encode()
 
 	if body != nil {
-		if err := req.SetBody(nopCloser{ReadSeeker: body}, "application/atom+xml;type=entry;charset=utf-8"); err != nil {
+		if err := req.SetBody(streaming.NopCloser(body), "application/atom+xml;type=entry;charset=utf-8"); err != nil {
 			return nil, err
 		}
 	}
--- a/sdk/messaging/azservicebus/internal/atom/entity_manager.go
+++ b/sdk/messaging/azservicebus/internal/atom/entity_manager.go
@@ -121,7 +121,7 @@ func NewEntityManager(ns string, tokenCredential azcore.TokenCredential, version
 
 // Get performs an HTTP Get for a given entity path, deserializing the returned XML into `respObj`
 func (em *entityManager) Get(ctx context.Context, entityPath string, respObj any) (*http.Response, error) {
-	resp, err := em.execute(ctx, http.MethodGet, entityPath, http.NoBody, nil)
+	resp, err := em.execute(ctx, http.MethodGet, entityPath, nil, nil)
 	defer CloseRes(ctx, resp)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


sdk/messaging/azservicebus/internal/atom/entity_manager.go
```diff
@@ -151,15 +151,24 @@ func (em *entityManager) Put(ctx context.Context, entityPath string, body any, r
...
 }
 
 type ExecuteOptions struct {
 	ForwardTo           *string
 	ForwardToDeadLetter *string
 }
 
-func (em *entityManager) execute(ctx context.Context, method string, entityPath string, body io.Reader, options *ExecuteOptions) (*http.Response, error) {
+// nopCloser copied from azblob.
+type nopCloser struct {
```
jhendrixMSFT, line 163: See `azcore/streaming.NopCloser`

richardpark-msft: Done.

jhendrixMSFT (approved):

PR #21505, opened 2023-09-06
[azservicebus] Test china cloud with regional affinity pool

benbp:
Proof of concept for regional affinity testing with the China cloud


richardpark-msft (approved):
Okay with experimenting on this, since it only affects the weekly pipelines and those are failing on this stage already.

PR #21510, opened 2023-09-06
Skip getting module properties for **/testdata/*

chlowell:
`Get-GoModuleVersionInfo` logs a warning for modules lacking version info, such as our perf test harnesses. This produces many irrelevant warnings:

image

These are irrelevant because we don't ship anything in `**/testdata/*`. It seems excessive to add whatever artifacts are needed to satisfy the script(s), so how about having the script(s) simply skip over these modules?

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -39,7 +39,7 @@ function Get-GoModuleProperties($goModPath)
   $goModPath = $goModPath -replace "\\", "/"
   # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227
   # The serviceName named capture group is unused but used in azure-sdk, so it's kept here for parity
-  if ($goModPath -match "(?<modPath>(sdk|profile)/(?<serviceDir>(.*?(?<serviceName>[^/]+)/)?(?<modName>[^/]+$)))")
+  if (!$goModPath.Contains("testdata") -and $goModPath -match "(?<modPath>(sdk|profile)/(?<serviceDir>(.*?(?<serviceName>[^/]+)/)?(?<modName>[^/]+$)))")
   {
     $modPath = $matches["modPath"]
     $modName = $matches["modName"] # We may need to start reading this from the go.mod file if the path and mod config start to differ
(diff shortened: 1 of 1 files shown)


jhendrixMSFT (approved):

eng/scripts/Language-Settings.ps1
```diff
@@ -39,7 +39,7 @@ function Get-GoModuleProperties($goModPath)
   $goModPath = $goModPath -replace "\\", "/"
   # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227
   # The serviceName named capture group is unused but used in azure-sdk, so it's kept here for parity
-  if ($goModPath -match "(?<modPath>(sdk|profile)/(?<serviceDir>(.*?(?<serviceName>[^/]+)/)?(?<modName>[^/]+$)))")
+  if (!$goModPath.Contains("testdata") -and $goModPath -match "(?<modPath>(sdk|profile)/(?<serviceDir>(.*?(?<serviceName>[^/]+)/)?(?<modName>[^/]+$)))")
```
benbp, line 42: This will need to be updated in azure-sdk as well: https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L238

eng/scripts/Language-Settings.ps1
```diff
@@ -39,7 +39,7 @@ function Get-GoModuleProperties($goModPath)
   $goModPath = $goModPath -replace "\\", "/"
   # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L227
```
benbp, line 40: If you don't mind updating the line number in the comment here as part of this:

```suggestion
  # We should keep this regex in sync with what is in the azure-sdk repo at https://github.com/Azure/azure-sdk/blob/main/eng/scripts/Query-Azure-Packages.ps1#L238
```

eng/scripts/Language-Settings.ps1
chlowell: That script discovers modules by iterating the repo's tags. It therefore already filters out `testdata/*` because none of those modules has an associated tag.

Issue #20541, opened 2023-04-03
azcore should support api-key authentication scheme


mikekistler:
### Feature Request

Some Azure services use an api-key, usually in a request header, for authentication.  Azure Core for Go should include a class for defining this form of authentication to allow Go clients to support services that use api-key authentication.

An example is the Python AzureKeyCredential class:

https://azuresdkdocs.blob.core.windows.net/$web/python/azure-core/latest/azure.core.html#azure.core.credentials.AzureKeyCredential

richardpark-msft:
If we add KeyCredential it seems logical to add in AzureNamedKeyCredential as well.

https://azuresdkdocs.blob.core.windows.net/$web/python/azure-core/latest/azure.core.html#azure.core.credentials.AzureNamedKeyCredential

jhendrixMSFT:
Possibly also AzureSASCredential.

https://azuresdkdocs.blob.core.windows.net/$web/python/azure-core/latest/azure.core.html#azure.core.credentials.AzureSasCredential

jhendrixMSFT:
Will likely punt on NamedKeyCredential for now since AFAIK only storage uses these and they have their own SharedKeyCredential.

jhendrixMSFT:
Will go out in `azcore@v1.8.0-beta.4`

PR #21521, opened 2023-09-08
[azeventgrid] Adding in a Publisher client to use with Event Grid topics

richardpark-msft:
Event Grid has two separate products at the moment:
- Event Grid namespaces, which enables pull delivery and MQTT. This service is in preview.
- Event Grid topics and Event Grid subscriptions - this service is GA.

This PR adds a new client (`publisher.Client`) that you can use to talk with the existing Event Grid topic service. This is primarily useful for system event publishers, like Microsoft.


sdk/messaging/azeventgrid/publisher/client_custom.go
```diff
@@ -0,0 +1,126 @@
...
+	return &Client{
+		internal: azc,
+		endpoint: endpoint,
+	}, nil
+}
+
+var tokenScopes = []string{"https://eventgrid.azure.net/.default"}
+
+// NewClient creates a [Client] that authenticates using a TokenCredential.
+func NewClient(endpoint string, tokenCredential azcore.TokenCredential, options *ClientOptions) (*Client, error) {
```
jhendrixMSFT, line 81: Nit:  the ctors should be ordered such that the "good" one (AAD) is first.  Probably put the SAS one as the last.

sdk/messaging/azeventgrid/publisher/client_custom.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return &Client{
+		internal: azc,
+		endpoint: endpoint,
+	}, nil
+}
+
+// NewClientWithSharedKeyCredential creates a [Client] using a shared key.
+func NewClientWithSharedKeyCredential(endpoint string, key string, options *ClientOptions) (*Client, error) {
```
jhendrixMSFT, line 52: `key` should probably be a `SharedKeyCredential` type and would fix your TODO below.  See what we did in `azblob` for an example.

sdk/messaging/azeventgrid/publisher/autorest.md
```diff
@@ -0,0 +1,68 @@
...
+    - https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/eventgrid/data-plane/Microsoft.EventGrid/stable/2018-01-01/EventGrid.json
+license-header: MICROSOFT_MIT_NO_VERSION
+openapi-type: "data-plane"
+output-folder: ../publisher
+override-client-name: Client
+security: "AADToken"
+use: "@autorest/go@4.0.0-preview.52"
+version: "^3.0.0"
+slice-elements-byval: true
+remove-non-reference-schema: true
```
jhendrixMSFT, line 19: Does `rawjson-as-bytes: true` need to be enabled as well or is that not applicable here?

sdk/messaging/azeventgrid/example_test.go
```diff
@@ -4,20 +4,139 @@
...
+		log.Fatalf("ERROR: %s", err)
 	}
 
 	_ = client // ignore
+
+	// Output:
+}
+
+func ExampleClient_PublishCloudEvents() {
+	client := getEventGridClient()
```
jhendrixMSFT, line 38: I wonder how this will render on `pkg.go.dev`.  It might end up being that the examples aren't copy/pasteable.

sdk/messaging/azeventgrid/publisher/autorest.md
richardpark-msft: I don't think it's the right thing because the underlying []byte slice is either the raw bytes _or_ JSON bytes, depending on what they send. 

I did notice that we're generating it in a way that I believe we don't want to do, however. I filed a bug for it : https://github.com/Azure/autorest.go/issues/1022

sdk/messaging/azeventgrid/example_test.go
richardpark-msft: I use pkgsite - they render nicely! It actually brings in the function and the called function in the file.

richardpark-msft: I should mention - I could have sworn it didn't used to work well either, so I was pleasantly surprised.

sdk/messaging/azeventgrid/publisher/autorest.md
richardpark-msft: Reopening for a second. EventGrid doesn't do the fancy encoding-switch that CloudEvent does, which means we don't end up in a situation where the underlying value might just be a raw stream of bytes or JSON. It's always just data encoded into the body of the message.

I don't think that changes what we have today - I think it's preferable to allow the customer to just use an object or primitive here and let JSON encoding do the work when we send it, rather than relying on them serializing the data to []bytes first. This is one area where I think C#/Java's BinaryData makes this representation easier. 

Also, of note - we don't deserialize the Data in this SDK - there's an entire constellation of "System Events" that will do it but we don't yet have that as part of this package. We might want a different approach there, depending on how that works.

sdk/messaging/azeventgrid/publisher/client_custom.go
richardpark-msft: Done.

sdk/messaging/azeventgrid/publisher/client_custom.go
richardpark-msft: (per our offline discussion)

Let's delay this change and create an actual credential type for this in azcore (tracked with this issue: #20541). I'd want to use this same key type in the root package of this module as well, and there's even a few other places like azopenai.

(not sure if the ship has sailed for azblob, but I'm guessing we can work it out with a type alias).

jhendrixMSFT (approved):

PR #21522, opened 2023-09-08
armstorageimportexport - clean up client constructors

jhendrixMSFT:
We don't allow optional client constructor parameters for ARM. Use an autorest transform until the OpenAPI has been fixed.


tadelesh (approved):
@jhendrixMSFT do you mean it is illegal to have optional client param for arm? There is another service with required client param: https://github.com/Azure/azure-sdk-for-go/blob/8166b03403f7d87203702589fb65ddf148611540/sdk/resourcemanager/apicenter/armapicenter/services_client.go#L36, it is legal?

jhendrixMSFT:
The `*arm.ClientOptions` type is legal and expected.  In your example, `serviceName` looks to be required?

jhendrixMSFT:
I wonder though, should `serviceName` actually be a client parameter?  It's almost always the wrong choice.

tadelesh:
> I wonder though, should `serviceName` actually be a client parameter? It's almost always the wrong choice.

I'm also has such concern. Go just followed other languages. I'll try to confirm with service team.

PR #21526, opened 2023-09-08
Some refactoring of azappconfig

jhendrixMSFT:
Moved options and response types to their respective files. Moved methods on Setting to its respective file.
Inlined various from* conversion funcs to their call sites. Renamed paged response types to have the Response suffix per guidelines. Fixed an issue parsing the time value returned in the Last-Modified header. The expected format was incorrect and the error was previously silently dropped.

### DIFF
--- a/sdk/data/azappconfig/options.go
+++ b/sdk/data/azappconfig/options.go
@@ -1,0 +1,78 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azappconfig
+
+import (
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// AddSettingOptions contains the optional parameters for the AddSetting method.
+type AddSettingOptions struct {
+	// Configuration setting label.
+	Label *string
+
+	// Configuration setting content type.
+	ContentType *string
+}
+
+// DeleteSettingOptions contains the optional parameters for the DeleteSetting method.
+type DeleteSettingOptions struct {
+	// Configuration setting label.
+	Label *string
+
+	// If set, and the configuration setting exists in the configuration store,
+	// delete the setting if the passed-in ETag is the same as the setting's ETag in the configuration store.
+	OnlyIfUnchanged *azcore.ETag
+}
+
+// GetSettingOptions contains the optional parameters for the GetSetting method.
+type GetSettingOptions struct {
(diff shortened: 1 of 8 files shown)


richardpark-msft (approved):
Looks good.

Are we going to try to switch over to full code generation at some point? I know there's a few things that would need to be autorest'd into consistency, at the very least.

jhendrixMSFT:
We had looked into a codegen solution but IIRC there were just too many oddities with the generated APIs plus it would be a pretty massive breaking change (I thought there was a draft PR with this, but I can't find it).

I will take another look, but I suspect we'll be sticking with our convenience layer.

sdk/data/azappconfig/options.go
```diff
@@ -0,0 +1,78 @@
...
+	Label *string
+
+	// If set, and the configuration setting exists in the configuration store,
+	// delete the setting if the passed-in ETag is the same as the setting's ETag in the configuration store.
+	OnlyIfUnchanged *azcore.ETag
+}
+
+// GetSettingOptions contains the optional parameters for the GetSetting method.
+type GetSettingOptions struct {
+	// Configuration setting label.
```
chlowell, line 36: Aren't we supposed to lint for doc comments that don't start with the name they document?

richardpark-msft: Hm. Perhaps it doesn't fire off for fields. I've only never noticed it with top level types/functions.

jhendrixMSFT: Hmm.  It could be that the tool only looks at the exported types and not their exported fields.  :(

richardpark-msft: This would be a rather big change - the doc comments that come from swagger/typespec don't match the Go format. I'm guessing most of our generated code isn't in compliance.

jhendrixMSFT: Yeah it would have to be opt-in else the repo would explode.

chlowell: Ah, good point, swagger comments are all over the place and we would need an LLM or an extremely tedious manual process to fix them all

jhendrixMSFT: FWIW we used to do this in the code generator by adding a prefix "<FieldName> -" to the doc strings.  We removed it during track 2 development as the strings got clunky.  Plus, it doesn't solve the case when fields don't have doc strings.

jhendrixMSFT: Maybe that's why we decided not to lint the fields (assuming that was intentional, I don't remember).

PR #21538, opened 2023-09-11
Some azappconfig cleanup

jhendrixMSFT:
Improved docs on some APIs.
Simplify handling of optional params type.
Use context.Background() in tests.
Added test for setting with nil value.
Fix sort order of fields in options and response types.

### DIFF
--- a/sdk/data/azappconfig/client_test.go
+++ b/sdk/data/azappconfig/client_test.go
@@ -25,7 +25,10 @@ func TestClient(t *testing.T) {
 	value := "value"
 	client := NewClientFromConnectionString(t)
 
-	addResp, err2 := client.AddSetting(context.TODO(), key, &value, &azappconfig.AddSettingOptions{Label: to.Ptr(label), ContentType: &contentType})
+	addResp, err2 := client.AddSetting(context.Background(), key, &value, &azappconfig.AddSettingOptions{
+		Label:       to.Ptr(label),
+		ContentType: &contentType,
+	})
 	require.NoError(t, err2)
 	require.NotEmpty(t, addResp)
 	require.NotNil(t, addResp.Key)
@@ -37,7 +40,9 @@ func TestClient(t *testing.T) {
 	require.Equal(t, contentType, *addResp.ContentType)
 	require.Equal(t, value, *addResp.Value)
 
-	getResp, err3 := client.GetSetting(context.TODO(), key, &azappconfig.GetSettingOptions{Label: to.Ptr(label)})
+	getResp, err3 := client.GetSetting(context.Background(), key, &azappconfig.GetSettingOptions{
+		Label: to.Ptr(label),
+	})
 	require.NoError(t, err3)
 	require.NotEmpty(t, getResp)
 	require.NotNil(t, getResp.Key)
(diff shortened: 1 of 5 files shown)


sdk/data/azappconfig/client_test.go
```diff
@@ -32,7 +32,11 @@ func TestClient(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEmpty(t, client)
 
-	addResp, err2 := client.AddSetting(context.TODO(), key, &value, &AddSettingOptions{Label: &label, ContentType: &contentType})
+	addResp, err2 := client.AddSetting(context.TODO(), key, &AddSettingOptions{
```
richardpark-msft, line 35: I use context.Background() typically. Would context.TODO() be a better choice? I typically only use those in examples.

sdk/data/azappconfig/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Breaking Changes
 * Response types `ListRevisionsPage` and `ListSettingsPage` now have the suffix `Response` in their names.
+* The optional `value *string` parameter in methods `AddSetting` and `SetSetting` has been moved to the respective options types.
```
richardpark-msft, line 10: This is correct, but I wonder if, in practice, this is a hassle since you're unlikely to want to Add or Set a setting with a null value.

richardpark-msft (commented):
Some questions.

sdk/data/azappconfig/client_test.go
jhendrixMSFT: While the end result is the same, typically, it's `context.Background()` for tests and `context.TODO()` for examples.  I will clean this up.

sdk/data/azappconfig/CHANGELOG.md
jhendrixMSFT: I see your point.  The problem here is that this directly conflicts with our design guidelines and looks odd compared to other APIs.  Perhaps as a compromise we can take `value` by value and if it's the empty string we omit it.

@JeffreyRichter do you recall if this was discussed at some point?  I didn't find an APIView with any conversation history.

sdk/data/azappconfig/client_test.go
jhendrixMSFT: Done.

sdk/data/azappconfig/CHANGELOG.md
jhendrixMSFT: Empirically, the answer appears to be yes.  Need to confirm with the service team though.

jhendrixMSFT: Well, actually I think this means that we just take it be value and we do nothing special on the client.

jhendrixMSFT: I will say though that when retrieving such a key, the value comes back as the empty string and not a `nil` field.

JeffreyRichter: I don't like this change. AddSetting should take a key AND a value as formal parameters.

jhendrixMSFT: Value is optional.  Do we leave it as pointer-to-type then?

jhendrixMSFT: Per offline discussion will leave this param as-is and document the behavior.

richardpark-msft (approved):

PR #21553, opened 2023-09-13
Add KeyCredential and SASCredential types

jhendrixMSFT:
Includes supporting pipeline policies, config options, etc.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/20541

### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -27,7 +27,7 @@ type KeyCredential = exported.KeyCredential
 
 // NewKeyCredential creates a new instance of [KeyCredential] with the specified values.
 //   - key is the authentication key
-func NewKeyCredential(key string) (*KeyCredential, error) {
+func NewKeyCredential(key string) *KeyCredential {
 	return exported.NewKeyCredential(key)
 }
 
@@ -36,7 +36,7 @@ type SASCredential = exported.SASCredential
 
 // NewSASCredential creates a new instance of [SASCredential] with the specified values.
 //   - sas is the shared access signature
-func NewSASCredential(sas string) (*SASCredential, error) {
+func NewSASCredential(sas string) *SASCredential {
 	return exported.NewSASCredential(sas)
 }
 
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -22,6 +22,24 @@ type AccessToken = exported.AccessToken
 // TokenCredential represents a credential capable of providing an OAuth token.
 type TokenCredential = exported.TokenCredential
 
+// KeyCredential contains an authentication key used to authenticate to an Azure service.
+type KeyCredential = exported.KeyCredential
+
+// NewKeyCredential creates a new instance of [KeyCredential] with the specified values.
+//   - key is the authentication key
+func NewKeyCredential(key string) (*KeyCredential, error) {
+	return exported.NewKeyCredential(key)
+}
+
+// SASCredential contains a shared access signature used to authenticate to an Azure service.
+type SASCredential = exported.SASCredential
+
+// NewSASCredential creates a new instance of [SASCredential] with the specified values.
+//   - sas is the shared access signature
+func NewSASCredential(sas string) (*SASCredential, error) {
+	return exported.NewSASCredential(sas)
+}
+
 // holds sentinel values used to send nulls
 var nullables map[reflect.Type]interface{} = map[reflect.Type]interface{}{}
 
--- a/sdk/azcore/internal/exported/exported.go
+++ b/sdk/azcore/internal/exported/exported.go
@@ -121,12 +121,8 @@ type KeyCredential struct {
 
 // NewKeyCredential creates a new instance of [KeyCredential] with the specified values.
 //   - key is the authentication key
-func NewKeyCredential(key string) (*KeyCredential, error) {
-	cred, err := newKeyCredential(key)
-	if err != nil {
-		return nil, err
-	}
-	return &KeyCredential{cred: cred}, nil
+func NewKeyCredential(key string) *KeyCredential {
+	return &KeyCredential{cred: newKeyCredential(key)}
 }
 
 // Update replaces the existing key with the specified value.
@@ -142,12 +138,8 @@ type SASCredential struct {
 
 // NewSASCredential creates a new instance of [SASCredential] with the specified values.
 //   - sas is the shared access signature
-func NewSASCredential(sas string) (*SASCredential, error) {
-	cred, err := newKeyCredential(sas)
-	if err != nil {
-		return nil, err
-	}
-	return &SASCredential{cred: cred}, nil
+func NewSASCredential(sas string) *SASCredential {
+	return &SASCredential{cred: newKeyCredential(sas)}
 }
 
 // Update replaces the existing shared access signature with the specified value.
@@ -169,13 +161,10 @@ type keyCredential struct {
 	key atomic.Value // string
 }
 
-func newKeyCredential(key string) (*keyCredential, error) {
-	if key == "" {
-		return nil, errors.New("key cannot be empty")
-	}
+func newKeyCredential(key string) *keyCredential {
 	keyCred := keyCredential{}
 	keyCred.key.Store(key)
-	return &keyCred, nil
+	return &keyCred
 }
 
 func (k *keyCredential) Get() string {
--- a/sdk/azcore/internal/exported/exported.go
+++ b/sdk/azcore/internal/exported/exported.go
@@ -9,9 +9,11 @@ package exported
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"sync/atomic"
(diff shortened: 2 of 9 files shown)


jhendrixMSFT:
Still need to add tests (this is why the CI fails).

sdk/azcore/core.go
```diff
@@ -22,6 +22,24 @@ type AccessToken = exported.AccessToken
 // TokenCredential represents a credential capable of providing an OAuth token.
 type TokenCredential = exported.TokenCredential
 
+// KeyCredential contains an authentication key used to authenticate to an Azure service.
+type KeyCredential = exported.KeyCredential
+
+// NewKeyCredential creates a new instance of [KeyCredential] with the specified values.
+//   - key is the authentication key
+func NewKeyCredential(key string) (*KeyCredential, error) {
```
jhendrixMSFT, line 30: Should these have options?

chlowell: Famous last words but I think we'll be okay without options because this is just a container for a string. The policy does all the interesting stuff.

jhendrixMSFT: I agree and kind of hate having to add two more options placeholder types (this is one place where variadic config has the edge).  I just want to make sure we're all in agreement.

sdk/azcore/internal/exported/exported.go
```diff
@@ -110,3 +112,80 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
...
+	return &SASCredential{cred: cred}, nil
+}
+
+// Update replaces the existing shared access signature with the specified value.
+func (k *SASCredential) Update(sas string) error {
+	return k.cred.Update(sas)
+}
+
+// KeyCredentialGet returns the key for cred.
+func KeyCredentialGet(cred *KeyCredential) string {
```
chlowell, line 159: I suppose the getters aren't exported, and don't have method receivers, to prevent applications from calling them?

sdk/azcore/core.go
richardpark-msft: Normally I want them but this is such a limited case...

sdk/azcore/runtime/policy_sas_credential.go
```diff
@@ -0,0 +1,39 @@
...
+// SASCredentialPolicyOptions contains the optional values configuring [SASCredentialPolicy].
+type SASCredentialPolicyOptions struct {
+	// placeholder for future optional values
+}
+
+// NewSASCredentialPolicy creates a new instance of [SASCredentialPolicy].
+//   - cred is the [azcore.SASCredential] used to authenticate with the service
+//   - header is the name of the HTTP request header in which the shared access signature is placed
+//   - options contains optional configuration, pass nil to accept the default values
+func NewSASCredentialPolicy(cred *exported.SASCredential, header string, options *SASCredentialPolicyOptions) *SASCredentialPolicy {
```
richardpark-msft, line 28: Hm...I don't think we want `exported` to show up in the public API surface since it's internal (can users create it?).

It might be that we need to put the credentials into a sibling package to runtime because what I've done below might be circular.

```suggestion
func NewSASCredentialPolicy(cred *azcore.SASCredential, header string, options *SASCredentialPolicyOptions) *SASCredentialPolicy {
```

sdk/azcore/runtime/policy_key_credential.go
```diff
@@ -0,0 +1,50 @@
...
+	// Format is used if the key needs special formatting (e.g. a prefix) before it's inserted into the HTTP request.
+	// The value passed to the callback is the raw key and the return value is the augmented key.
+	Format func(string) string
+}
+
+// NewKeyCredentialPolicy creates a new instance of [KeyCredentialPolicy].
+//   - cred is the [azcore.KeyCredential] used to authenticate with the service
+//   - header is the name of the HTTP request header in which the key is placed
+//   - options contains optional configuration, pass nil to accept the default values
+func NewKeyCredentialPolicy(cred *exported.KeyCredential, header string, options *KeyCredentialPolicyOptions) *KeyCredentialPolicy {
```
richardpark-msft, line 31: (similar question here as to SASCredentialPolicy - I don't think we're supposed to  use `exported`.

richardpark-msft (changes requested):
A couple of questions about `exported` use in the public surface.

sdk/azcore/runtime/policy_sas_credential.go
jhendrixMSFT: For better or worse, we follow the same pattern elsewhere (e.g. `NewBearerTokenPolicy()`) for this exact reason (i.e. avoid a circular dependency).  Also keep in mind that `exported.SASCredential` is type-aliased as `azcore.SASCredential` so it's publicly available (this is included in the doc comment).

I don't think we want to put these credentials into a separate package.  `TokenCredential` is in `azcore` and IMO we should keep them all together.

sdk/azcore/internal/exported/exported.go
jhendrixMSFT: Correct.  Although, to be frank, I waffled on this a bit trying to decide if we're just overengineering it.  We did apply the same pattern elsewhere (e.g. `SharedKeyCredential` in `azblob`) so there is some prior art.  And I do think there's value in not exporting APIs that aren't interesting to customers.  I will point out though that this pattern prevents SDKs from creating their own pipeline policies for these credential types.  If we think supporting that is import, then this needs to be refactored (we can always do it later in a non-breaking way).

sdk/azcore/runtime/policy_sas_credential.go
richardpark-msft: I see. I have a QoL issue to discuss this at some point - it's not a great experience because your first auto-complete option in VSCode, for instance, is to the exported symbol. So you always have to fix it - in this case I think it's not so bad because you have to use a constructor + error and so you're never going to run into this particular symbol (hopefully).

sdk/azcore/runtime/policy_key_credential.go
richardpark-msft: (answered in another comment)

sdk/azcore/internal/exported/exported.go
richardpark-msft: It seems fine - they literally pass and set the key they're updating or initializing with. Get() doesn't really provide them much value. 

Of course if someone complains we can always implement it later.

richardpark-msft (approved):
Looks fine by me, @chlowell has a few unresolved questions, I believe, although I think we were just waiting for @JeffreyRichter to weigh in.

sdk/azcore/core.go
JeffreyRichter: Since the only parameter is a key whose format/values is defined by the service, the client cannot validate it and therefore I think returning an error is unnecessary.

jhendrixMSFT: The only reason it returns an error at present is if the key is the empty string.

Will remove per discussion.

jhendrixMSFT:
@chlowell any other concerns?

@JeffreyRichter does this look good now?

sdk/azcore/internal/exported/exported.go
```diff
@@ -110,3 +112,69 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
...
+	return cred.cred.Get()
+}
+
+// SASCredentialGet returns the shared access sig for cred.
+func SASCredentialGet(cred *SASCredential) string {
+	return cred.cred.Get()
+}
+
+type keyCredential struct {
+	key atomic.Value // string
```
JeffreyRichter, line 161: You could use this: https://pkg.go.dev/sync/atomic#Pointer with *string for T

sdk/azcore/internal/exported/exported.go
```diff
@@ -110,3 +112,69 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
...
+	keyCred := keyCredential{}
+	keyCred.key.Store(key)
+	return &keyCred
+}
+
+func (k *keyCredential) Get() string {
+	return k.key.Load().(string)
+}
+
+func (k *keyCredential) Update(key string) error {
```
JeffreyRichter, line 174: Remove error, right?

JeffreyRichter (approved):

sdk/azcore/internal/exported/exported.go
jhendrixMSFT: Good catch, fixed.

sdk/azcore/internal/exported/exported.go
jhendrixMSFT: We could, but it would require setting the minimum Go version to `1.19`.  I'm not against it, but it will create extra work to update the minimum version across all our modules.

chlowell (approved):

sdk/azcore/internal/exported/exported.go
JeffreyRichter: That's fine then - leave it as is.

PR #21580, opened 2023-09-18
Don't recreate body after reading

jhendrixMSFT:
Rewind the body instead. And if it was nil, leave it as nil. Removed some unnecessary local vars.

### DIFF
--- a/sdk/data/azappconfig/assets.json
+++ b/sdk/data/azappconfig/assets.json
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "go",
   "TagPrefix": "go/data/azappconfig",
-  "Tag": "go/data/azappconfig_560d04663e"
+  "Tag": "go/data/azappconfig_88f23302b7"
 }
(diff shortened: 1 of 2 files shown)


sdk/data/azappconfig/assets.json
```diff
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "go",
   "TagPrefix": "go/data/azappconfig",
-  "Tag": "go/data/azappconfig_560d04663e"
+  "Tag": "go/data/azappconfig_88f23302b7"
```
chlowell, line 5: How did this change recordings?

chlowell (approved):

sdk/data/azappconfig/assets.json
jhendrixMSFT: Previously, we were [always adding back the body](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azappconfig/internal/auth/policy_hmac_auth.go#L59) even if it was `nil`.

Issue #21585, opened 2023-09-19
azappconfig: incorrect sync-token value sent in requests


jhendrixMSFT:
Example received Sync-Token.
```
Sync-Token: zAJw6V16=Mzo1IzE0Njc1OTI0;sn=14675924
```

The sync token policy is incorrectly formatting the value.  Here's what's being sent over the wire.
```
Sync-Token: zAJw6V16
```

The value is missing (sequence number isn't sent).

jhendrixMSFT:
In addition, the current implementation isn't goroutine safe updating the map of sync tokens.

PR #21589, opened 2023-09-19
Refactor handling of sync tokens

jhendrixMSFT:
Added internal package synctoken that contains the sync token cache and pipeline policy implementations.
Renamed UpdateSyncToken to SetSyncToken so it follows the design guidelines.  In addition, it returns an error on malformed sync tokens. Removed setting sync token from test as it's not necessary (the policy already does this).

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21585


sdk/data/azappconfig/client.go
```diff
@@ -60,24 +61,25 @@ func newClient(endpoint string, authPolicy policy.Policy, options *ClientOptions
...
 	}, nil
 }
 
-// UpdateSyncToken sets an external synchronization token to ensure service requests receive up-to-date values.
-func (c *Client) UpdateSyncToken(token string) {
-	c.syncTokenPolicy.addToken(token)
+// SetSyncToken is used to set a sync token from an external source.
+// Sync tokens are required to be in the format "<id>=<value>;sn=<sn>".
+// See https://learn.microsoft.com/en-us/azure/azure-app-configuration/rest-api-consistency for more information on sync tokens.
+func (c *Client) SetSyncToken(syncToken string) error {
```
jhendrixMSFT, line 81: While this does make us deviate from the other SDKs, they got it wrong.

CC @JeffreyRichter

sdk/data/azappconfig/internal/synctoken/cache.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// Cache contains a collection sync token values.
```
chlowell, line 17:
```suggestion
// Cache contains a collection of sync token values.
```

chlowell (approved):

sdk/data/azappconfig/client.go
```diff
@@ -60,24 +61,25 @@ func newClient(endpoint string, authPolicy policy.Policy, options *ClientOptions
...
+		cache:           cache,
 	}, nil
 }
 
-// UpdateSyncToken sets an external synchronization token to ensure service requests receive up-to-date values.
-func (c *Client) UpdateSyncToken(token string) {
-	c.syncTokenPolicy.addToken(token)
+// SetSyncToken is used to set a sync token from an external source.
+// Sync tokens are required to be in the format "<id>=<value>;sn=<sn>".
+// See https://learn.microsoft.com/en-us/azure/azure-app-configuration/rest-api-consistency for more information on sync tokens.
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/data/azappconfig/internal/synctoken/cache.go
jhendrixMSFT: fixed

sdk/data/azappconfig/client.go
JeffreyRichter: OK. Do we want to introduce a "type SyncToken string" with some helper methods for better type-safety, splitting, and maybe determining with version is later?

jhendrixMSFT: We could add a `type SyncToken string` and replace all instances of `SyncToken *string` with it.  However, I don't think any of the splitting methods need to be public as only the `synctoken.Cache` cares about that (I don't see customers ever needing the ability to perform these operations; plus, we could always make them public later).

As a follow-up, how do you feel about making `SyncToken` fields _not_ pointer-to-type?  According to the docs, _all_ operations return a Sync-Token value so it would never be `nil`.  Plus, Sync-Token can never be an empty string, so I think an empty check instead of a `nil` check is not ambiguous.

JeffreyRichter: Fine with me. The SyncToken type-safety improves the API. The methods could be added later (if we ever need them) and removing pointer is also OK with me if we're 100% sure that there is never a chance of nil today or in the future.

jhendrixMSFT: Done.

PR #21593, opened 2023-09-20
Fix SetImmutabilityPolicyOptions.format() NPE

KastenMike:
The `o.format()` pattern is used as a convenient way to let people pass `nil` for options but there's an NPE in this case since `.format()` returns `nil` but [SetImmutabilityPolicy](https://github.com/Azure/azure-sdk-for-go/blob/6bcc14f7abbda96fd6fef863f6eaa82083c160f4/sdk/storage/azblob/blob/client.go#L250) then tries setting `blobSetImmutabilityPolicyOptions.ImmutabilityPolicyExpiry`. Returning an empty struct so this can be set without having an NPE.


KastenMike:

vibhansa-msft (approved):

PR #21607, opened 2023-09-22
Add user authentication example

chlowell:
Making the example from #21245 part of the docs

### DIFF
--- a/sdk/azidentity/example_user_auth_test.go
+++ b/sdk/azidentity/example_user_auth_test.go
@@ -1,0 +1,73 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azidentity_test
+
+import (
+	"context"
+	"encoding/json"
+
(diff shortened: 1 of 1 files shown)


sdk/azidentity/example_user_auth_test.go
```diff
@@ -0,0 +1,73 @@
...
+		DisableAutomaticAuthentication: true,
+
+		// By default, credentials cache in memory. Set TokenCachePersistenceOptions to enable persistent caching.
+		TokenCachePersistenceOptions: &azidentity.TokenCachePersistenceOptions{
+			// optionally set Name to isolate this credential's cache from other applications
+			Name: "myapp",
+		},
+	})
+	if err != nil {
+		// TODO: handle error
```
joshfree, line 33: I would put the error handling in the sample, rather than a TODO.

joshfree (approved):

sdk/azidentity/example_user_auth_test.go
jhendrixMSFT: This is the pattern we've adopted in all of our samples.  Makes it clear that if one copy/pastes there's more work to be done in order to properly integrate it.

PR #21625, opened 2023-09-26
[azopenai] Be more resilient to messages coming back with just content filtering…

richardpark-msft:
It's always possible for the Azure OpenAI to return chat messages that do not contain content (they carry content-filtering info instead).

### DIFF
--- a/sdk/ai/azopenai/example_client_getchatcompletions_test.go
+++ b/sdk/ai/azopenai/example_client_getchatcompletions_test.go
@@ -79,8 +79,26 @@ func ExampleClient_GetChatCompletions() {
 	for _, choice := range resp.Choices {
 		gotReply = true
 
+		if choice.ContentFilterResults != nil {
+			fmt.Fprintf(os.Stderr, "Content filter results\n")
+
+			if choice.ContentFilterResults.Error != nil {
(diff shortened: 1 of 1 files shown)


sdk/ai/azopenai/example_client_getchatcompletions_test.go
```diff
@@ -78,7 +78,10 @@ func ExampleClient_GetChatCompletions() {
 
 	for _, choice := range resp.Choices {
 		gotReply = true
-		fmt.Fprintf(os.Stderr, "Content[%d]: %s\n", *choice.Index, *choice.Message.Content)
+
+		if choice.Message != nil && choice.Message.Content != nil {
```
jhendrixMSFT, line 82: Would it be possible to add a comment here explaining when a response might contain only filtering info.

jhendrixMSFT (approved):

sdk/ai/azopenai/example_client_getchatcompletions_test.go
richardpark-msft: I've adjusted it. The current behavior isn't ideal so it's best to not assume it's one or the other. I did change it so it's clear what data is what, however.

PR #21630, opened 2023-09-27
Persistent caching on Linux

chlowell:
This implementation stores encrypted cache data in a file and makes the encryption key available to all the user's processes via the kernel key retention facility. There's a bit of DIY here to implement PKCS#7 padding and the required subset of JWE, but the actual encryption is via the standard library.


sdk/azidentity/cache/linux_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	p, err := internal.CacheFilePath(o.Name)
+	require.NoError(t, err)
+	require.NoFileExists(t, p)
+	o.AllowUnencryptedStorage = true
+	k, err := storage(o)
+	require.NoError(t, err)
+	require.NoError(t, k.Write(ctx, []byte("data")))
+	require.FileExists(t, p)
+	os.Remove(p)
```
heaths, line 160: Since you're using `require`, shouldn't this be either a `t.Cleanup` or `defer`? A lot can panic between assignment of `p` and now.

heaths (approved):
Nice!

sdk/azidentity/cache/cache.go
```diff
@@ -1,6 +1,6 @@
-//go:build go1.18 && (darwin || windows)
+//go:build go1.18 && (darwin || linux || windows)
 // +build go1.18
-// +build darwin windows
+// +build darwin linux windows
```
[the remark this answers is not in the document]
chlowell: The old fashioned constraints are a legacy of sharing this repo with track 1. I believe we still need them despite having removed track 1 from main, but I've forgotten the scenario that's broken without them.

sdk/azidentity/cache/internal/aescbc/aescbc.go
```diff
@@ -0,0 +1,118 @@
...
+		encKey: k[len(k)/2:],
+		hasher: hasher,
+		macKey: k[:len(k)/2],
+		tLen:   len(k) / 2,
+	}, nil
+}
+
+func (a *AESCBCHMACSHA2) Decrypt(iv, ciphertext, additionalData, tag []byte) ([]byte, error) {
+	expected := a.tag(iv, ciphertext, additionalData)
+	if !hmac.Equal(tag, expected) {
```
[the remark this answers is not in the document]
chlowell: It is, according to the docs at least.

sdk/azidentity/cache/linux_test.go
chlowell: `k.Write()` creates the file, so if an assertion fails before this line, there shouldn't be a file to remove. But deferring this doesn't hurt and I suppose it's conceivable a bug could cause `Write` to create the file *and* return an error.

sdk/azidentity/cache/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0
-	github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.0
+	github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.1-0.20230928163952-361da6dd79f8
```
jhendrixMSFT, line 8: When will `v0.1.1` be released?

sdk/azidentity/cache/internal/aescbc/aescbc.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, errors.New("decryption failed")
```
jhendrixMSFT, line 56: Does this error get surfaced outside of `azidentity`?  If so, perhaps a bit more descriptive message?

sdk/azidentity/cache/internal/aescbc/aescbc.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (a *AESCBCHMACSHA2) Encrypt(iv, plaintext, additionalData []byte) (EncryptResult, error) {
+	result := EncryptResult{}
+	block, err := aes.NewCipher(a.encKey)
+	if err != nil {
+		return result, err
+	}
+	plaintext = pad(plaintext)
+	if err != nil {
```
jhendrixMSFT, line 74: Superfluous error check.

sdk/azidentity/cache/internal/aescbc/aescbc.go
jhendrixMSFT: Comment applies to a few other places.

sdk/azidentity/cache/linux.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+var (
+	cacheDir   = os.UserHomeDir
+	tryKeyring = func() error {
+		k, err := newKeyring("azidentity-test-cache")
+		if err != nil {
+			return err
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
jhendrixMSFT, line 40: If the key ring isn't available, could this actually hang for one minute?

sdk/azidentity/cache/linux.go
```diff
@@ -0,0 +1,242 @@
...
+		msg := fmt.Sprintf("cache encryption is impossible because the key retention facility isn't usable: %s", err)
+		if o.AllowUnencryptedStorage {
+			f := ""
+			f, err = internal.CacheFilePath(name)
+			if err == nil {
+				log.Write(azidentity.EventAuthentication, msg+". Falling back to unencrypted storage")
+				return file.New(f)
+			}
+		} else {
+			err = errors.New(msg + ". Set AllowUnencryptedStorage on TokenCachePersistenceOptions to store the cache in plaintext instead of returning this error")
```
jhendrixMSFT, line 70: Nit, maybe lower-case `Set` to be consistent with lower-casing?

sdk/azidentity/cache/linux.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return err
+	}
+	content, err := j.Serialize()
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(k.file, []byte(content), 0600)
+	if errors.Is(err, os.ErrNotExist) {
+		err = os.MkdirAll(filepath.Dir(k.file), 0700)
```
jhendrixMSFT, line 154: You could call this before `os.WriteFile()`, it's a no-op if the dir exists.

sdk/azidentity/cache/linux.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("couldn't store encryption key due to error %q", err)
+	}
+	k.key = key[:keySize]
+	k.keyID = id
+	return k.key, nil
+}
+
+func (k *keyring) decrypt(j jwe.JWE) ([]byte, error) {
+	for tries := 0; tries < 2; tries++ {
```
jhendrixMSFT, line 180: How did you decide on 2?

sdk/azidentity/cache/linux.go
chlowell: No, that would be very fast. This is pure theater because none of the methods use the context. Maybe it's clearer to just pass `context.Background` here?

sdk/azidentity/cache/internal/aescbc/aescbc.go
chlowell: This message is deliberately vague to prevent a padding oracle attack.

jhendrixMSFT: Makes sense.  Might be good to add a comment about this someplace in case somebody tries to improve the error messages in the future.

sdk/azidentity/cache/linux.go
jhendrixMSFT: Yeah, I think `context.Background()` would be better along with comment explaining that this can't really hang forever.

sdk/azidentity/cache/linux.go
jhendrixMSFT: I see.  Mostly I was curious if the value was selected from some targeted testing/experimentation.  But one try plus one retry makes sense.

In the event that the retry fails, what error is surfaced to the caller, and how does one recover?

sdk/azidentity/cache/linux.go
chlowell: It doesn't change the filesystem in that case but it still makes a syscall, which I imagine is more expensive than `errors.Is` (haven't measured it though)

sdk/azidentity/cache/linux.go
chlowell: I dunno, this follows a period closing another sentence, shouldn't it be capitalized?

jhendrixMSFT: True, and I waffled on commenting due to this.  I just think about how error messages in general aren't proper sentences, so the `S` looks weird to me.  Either way is fine though.

sdk/azidentity/cache/internal/aescbc/aescbc.go
chlowell: You know you've truly become a Go developer when error checks turn into punctuation you don't notice anymore 😩

sdk/azidentity/cache/go.mod
chlowell: I don't know yet, but I won't ship this feature before that release.

sdk/azidentity/cache/linux.go
chlowell: If the retry fails, it means the data can't be decrypted because its encryption key has been replaced. There's no recovery from that so `Read` returns no data and a `nil` error as if the cache doesn't exist, which is effectively the case.

sdk/azidentity/cache/internal/aescbc/aescbc_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAESCBCHMAC(t *testing.T) {
+	for _, test := range []struct {
+		aad, key, plaintext, iv, ciphertext, tag []byte
+		name                                     string
+	}{
+		{
```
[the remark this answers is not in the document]
chlowell: The end to end tests in `linux_test.go` cover this because `keyring` doesn't use aad

sdk/azidentity/cache/linux.go
jhendrixMSFT: So all of this is transparent to the caller then.  And since there's effectively no cache, a new one is created then?

chlowell: Yes. The next `Write` will encrypt with the new key and overwrite the file.

jhendrixMSFT: Got it, thanks for walking me through it.

jhendrixMSFT (approved):

sdk/azidentity/cache/linux.go
heaths: > I just think about how error messages in general aren't proper sentences

A subject of much chagrin, IMO, especially when the fallback case seems to be "write the error to stderr (or stdout) and terminate". We could probably be more dev-friendly with actionable errors.

PR #21656, opened 2023-09-29
Add pipeline for azidentity/cache

chlowell:


jhendrixMSFT (approved):
Does the pipeline for `azidentity` need to exclude this module or do you want both to run when updating `azidentity`?

chlowell:
Both should run because a change in `azidentity` could break `azidentity/cache`

PR #21681, opened 2023-10-03
Fix azidentity/cache CI

chlowell:
This skips testing the Linux persistent cache implementation when it isn't useable, as in our CI where the user who executes tests apparently isn't allowed to write files. Skipping those tests requires setting a low coverage bar. I also added a changelog, readme and license file to get the analyze job passing.

### DIFF
--- a/sdk/azidentity/cache/README.md
+++ b/sdk/azidentity/cache/README.md
@@ -1,6 +1,6 @@
 # Azure Identity Cache Module for Go
 
-This module implements a cross-platform persistent token cache for [azidentity][https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity] credentials. See that module's [examples](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#pkg-examples) for sample code showing how to configure persistent caching for a credential.
+This module implements a cross-platform persistent token cache for [azidentity](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) credentials. See that module's [examples](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#pkg-examples) for sample code showing how to configure persistent caching for a credential.
 
 ## Provide Feedback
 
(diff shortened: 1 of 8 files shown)


sdk/azidentity/cache/README.md
```diff
@@ -0,0 +1,29 @@
+# Azure Identity Cache Module for Go
+
+This module implements a cross-platform persistent token cache for [azidentity][https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity] credentials. See that module's [examples](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#pkg-examples) for sample code showing how to configure persistent caching for a credential.
```
jhendrixMSFT, line 3: Is the link to `azidentity` supposed to be in parenthesis?

sdk/azidentity/cache/linux_test.go
```diff
@@ -18,9 +18,21 @@ import (
...
 
-var ctx = context.Background()
+var (
+	ctx              = context.Background()
+	storageAvailable = false
+)
+
+func TestMain(m *testing.M) {
+	err := tryKeyring()
+	storageAvailable = err == nil
```
jhendrixMSFT, line 28: What kind of error is happening in CI?  I wonder if there's a way we can fix this so we don't have to skip coverage.

sdk/azidentity/cache/README.md
chlowell: Yes! Thanks for spotting this

sdk/azidentity/cache/linux_test.go
```diff
@@ -18,21 +18,19 @@ import (
...
-	ctx              = context.Background()
-	storageAvailable = false
-)
+var ctx = context.Background()
 
 func TestMain(m *testing.M) {
-	err := tryKeyring()
-	storageAvailable = err == nil
+	// ensure the user running tests in CI is able to write cache files
+	if tmpdir, ok := os.LookupEnv("Agent.TempDirectory"); ok {
```
benbp, line 25: Sorry @chlowell I misled you a bit. In the yaml you can use `$(Agent.TempDirectory)` to set new env variables or pass as a parameter. However in the environment it gets converted to `AGENT_TEMPDIRECTORY`.

sdk/azidentity/cache/linux_test.go
chlowell: Turns out there were two issues: a test bug used the wrong path, and the user isn't allowed to search its persistent keyring. I fixed the test bug but the permission issue requires more investigation. I think we should skip the tests in CI for now, to unblock an initial beta release while I figure out whether there's a better way to handle or work around this permission issue.

jhendrixMSFT: Glad you found the root cause.

PR #21692, opened 2023-10-05
Update immutability Mode default comment

KastenMike:
It says the Mode defaults to unlocked but it defaults to the current version of the blob. If the blob's current mode is locked & you extend it, the mode remains locked.

I had added a comment about GMT/UTC but noticed it was resolved in https://github.com/Azure/azure-sdk-for-go/pull/20495  so removed it


KastenMike:
> This field is optional, and if not specified, default value is Unlocked. 

If the default value is Unlocked then it would fail when calling `SetImmutabilityPolicyOptions` on an existing blob that is set to Locked (because Locked cannot be changed to Unlocked). Instead it continues using the Locked by default

PR #21698, opened 2023-10-05
azcore v1.9.0-beta.1 release

jhendrixMSFT:
Updated changelog with v1.8.0 release notes.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -1,15 +1,28 @@
 # Release History
 
-## 1.8.0-beta.4 (Unreleased)
+## 1.9.0-beta.1 (2023-10-05)
+
+### Other Changes
+
+* The beta features for tracing and fakes have been reinstated.
+
+## 1.8.0 (2023-10-05)
 
 ### Features Added
 
+* This includes the following features from `v1.8.0-beta.N` releases.
+  * Claims and CAE for authentication.
+  * New `messaging` package.
+  * Various helpers in the `runtime` package.
+  * Deprecation of `runtime.With*` funcs and their replacements in the `policy` package.
 * Added types `KeyCredential` and `SASCredential` to the `azcore` package.
   * Includes their respective constructor functions.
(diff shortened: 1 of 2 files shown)


sdk/azcore/CHANGELOG.md
```diff
@@ -1,15 +1,28 @@
...
+
+### Other Changes
+
+* The beta features for tracing and fakes have been reinstated.
+
+## 1.8.0 (2023-10-05)
 
 ### Features Added
 
+* This includes the following features from `v1.8.0-beta.N` releases.
```
chlowell, line 13: Doesn't it go without saying that 1.9.0 is a superset of 1.8.0?

chlowell (approved):

sdk/azcore/CHANGELOG.md
jhendrixMSFT: This is for the `v1.8.0` release.  I was condensing the release notes from the `release/azcore-v1.8.0` branch.  I thought about omitting this entirely but thought this would make it easier to follow given we've jumped around a bit.  Maybe it would be better to say something like "includes all previous v1.8.0-beta.N features except fakes and tracing"?

chlowell: That works but this makes sense too, I just overlooked the version line 🤦 

I probably would copy the text directly from v1.8.0 on the rationale that the changelog is a single, immutable document

jhendrixMSFT: That was my original strategy, but you end up with some duplicated content as the `v1.8.0` changelog doesn't contain the entries for the `v1.8.0-beta.N` series.  I'm probably overthinking this and can just paste it verbatim from `v1.8.0` if you think that's best.

chlowell: Ah, I missed that the 1.8.0 changelog doesn't mention the betas. That makes it malformed in my mind so I'm okay with fixing it in the next release--what you have here looks good to me

jhendrixMSFT: That's a good point, maybe the `v1.8.0` changelog should have included the beta release notes.  We can revisit during the next release cycle.

Issue #21704, opened 2023-10-06
Add policy.WithQueryStringParameter to azcore/policy


serbrech:
### Feature Request

Similar to [policy.WithHTTPHeader](https://github.com/Azure/azure-sdk-for-go/blob/54b44cfe6513178432767a7f63a054ca27771e57/sdk/azcore/policy/policy.go#L175C1-L180C2) some apis need query string parameters that aren't necessarily declared in the openapi spec.

haitch:
I have one already, might just open a PR next week.

RickWinter:
This is not a generalized scenario that we want to support. AKS can update TypeSpec to declare the query parameters or add a custom policy within the AKS client sdks.

PR #21705, opened 2023-10-06
Add WithQueryParameters to policy package

jhendrixMSFT:
Enables custom query params for API calls.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21704

### DIFF
--- a/sdk/azcore/runtime/policy_query_params.go
+++ b/sdk/azcore/runtime/policy_query_params.go
@@ -1,0 +1,48 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package runtime
+
+import (
+	"net/http"
+	"net/url"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+// queryParamsPolicy is a policy that adds custom query parameters to a request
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
@haitch FYI

JeffreyRichter:
The code looks fine but why are we adding this? Is there some customer scenario that requires this?

jhendrixMSFT:
AKS folks asked for this.

JeffreyRichter:
OK but I'd still like to know more. If the query params are in swagger/TypeSpec, then we would have them in the SDK. So, do they have undocumented query parameters that they need this for? And, if so, this is not ideal.

jhendrixMSFT:
Good question.  @serbrech can you elaborate on your scenario?

sdk/azcore/runtime/policy_query_params.go
```diff
@@ -0,0 +1,48 @@
...
+		var err error
+		allQP, err = url.ParseQuery(req.Raw().URL.RawQuery)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	newQP := rawQP.(url.Values)
+	if len(allQP) > 0 {
+		// merge the existing query params with the ones in the context.
```
haitch, line 37: this merge might have a issue, it's always replace behavior.

http allows queryparameter to be repeated, like ?name=bob&name=alice. 

let's say I have name=bob in my original request, I want to append name=alice, this policy won't work, it always replace.

haitch (commented):
overall, it will work for us. I am not going to block it, but maybe there could be case where this is a issue.

jhendrixMSFT:
Closing for now until we work out a few more details.

PR #21731, opened 2023-10-10
[azeventgrid] Use CE instead of EG (otherwise we're generating a SAS based on a separate topic)

richardpark-msft:
Copy/paste error from another PR.


jhendrixMSFT (approved):

richardpark-msft:
Overriding check-enforcer. The dep check isn't related to my PR, it's a publishing issue with azidentity that's being worked on.

PR #21743, opened 2023-10-11
[azopenai] Update to latest .tsp

richardpark-msft:
They did a refactor and it removed an unused type.


RickWinter (approved):
We should rev the Beta version with this change since its breaking.

PR #21789, opened 2023-10-17
Remove dead code from azcore

jhendrixMSFT:
The functionality to detect R/O fields has been removed from generated code for quite some time now, making this code obsolete.


sdk/azcore/runtime/request.go
```diff
@@ -105,9 +102,6 @@ func MarshalAsByteArray(req *policy.Request, v []byte, format Base64Encoding) er
 
 // MarshalAsJSON calls json.Marshal() to get the JSON encoding of v then calls SetBody.
 func MarshalAsJSON(req *policy.Request, v interface{}) error {
-	if omit := os.Getenv("AZURE_SDK_GO_OMIT_READONLY"); omit == "true" {
```
richardpark-msft, line 108: Is there any way to ensure that nobody is using the "escape hatch"? I have no idea how widely it was published but it's hard to Bing for so I assume it was mostly an internal detail.

richardpark-msft (approved):
One comment only - just wondering if we need to add something to the CHANGELOG or update some documentation if it was mentioned anywhere.

sdk/azcore/runtime/request.go
jhendrixMSFT: It was never publicly documented. We just had it in our back pocket to prescribe on an as-needed basis.  At any rate, the generated code no longer emits the `azure:"ro"` tag on model fields so even if a customer were to set this it would have no effect.

PR #21790, opened 2023-10-17
[azopenai] Increasing test time to account for longer/slower API calls.

richardpark-msft:
Raising the test timeout. We work across two APIs, both of which can be on the slower side and as I write more tests, push us over the default 15m limit.

### DIFF
--- a/sdk/ai/azopenai/ci.yml
+++ b/sdk/ai/azopenai/ci.yml
@@ -25,15 +25,18 @@ pr:
 stages:
   - template: /eng/pipelines/templates/jobs/archetype-sdk-client.yml
     parameters:
+      # We need to allow for longer retry times with tests that run against the public endpoint
+      # which throttles under load.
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (approved):

richardpark-msft:
Looking a bit at my log it looks like test times have been creeping up a bit on some APIs, but the test runs failing is making true measurement difficult. I'm upping the times a bit.

My latest runs appear to be showing that whisper is more contentious on the new resource I have moved to (creds/config change so it doesn't show up here).

Still investigating.

benbp (approved):

richardpark-msft:
Might need some more adjustments, but I wanted to try removing some of the retries and see where our timeouts are still coming from.

richardpark-msft:
Looks like I've gotten it all tuned the way it needs to be. I'm going to check this in and see if everything works reasonably.

PR #21793, opened 2023-10-18
Update internal test package usage to v1.4.0

benbp:


RickWinter (approved):

benbp:
@RickWinter FYI there are some updates we're going to need to make en-masse to the test packages. Some just need a slight refactor to use proxy properly, others appear to be using it incorrectly, and a final set appear to not be structured using the TestMain pattern.

RickWinter:
> @RickWinter FYI there are some updates we're going to need to make en-masse to the test packages. Some just need a slight refactor to use proxy properly, others appear to be using it incorrectly, and a final set appear to not be structured using the TestMain pattern.

Are you making these updates or do you need assistance from the eng team?

benbp:
@RickWinter I'm going to do a quick pass to see what I can get done, but I anticipate there may be a couple that I'll need help with. I will follow up.

benbp:
Closing in favor of https://github.com/Azure/azure-sdk-for-go/pull/21874 for data plane

PR #21802, opened 2023-10-18
Plumb TestRunTime parameter through job generation template

benbp:
We missed a spot extending the `-timeout` argument to `go test` so overridden values aren't getting plumbed through.


richardpark-msft:
(this is needed to extend the go test runtime)

richardpark-msft (approved):

PR #21810, opened 2023-10-19
fix: handle nil request case

### LINKED ISSUE #21809: azidentity: panic on AuthenticationFailedError.Error when using custom tranporter
### Bug Report

- import path of package in question, e.g. `.../services/compute/mgmt/2018-06-01/compute`

```
github.com/Azure/azure-sdk-for-go/sdk/azidentity
```

- SDK version e.g. `master`, `latest`, `18.1.0`

`master`

  - Specify the exact commit if possible; one way to get this is the REVISION
    column output by `go list -m <module>`, for example `go list -m github.com/Azure/azure-sdk-for-go/sdk/azcore`.
- output of `go version`

```
go version go1.21.0 darwin/arm64
```

- What happened?

When using custom http transport to return failure response to `GetToken` calls, the returned `Error` calls on the returned `err` object could cause panic. This is a snippet of reproduction:

```
type unitTestTransport struct {
	h http.Handler
}

var _ http.RoundTripper = (*unitTestTransport)(nil)

func (t *unitTestTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	rec := httptest.NewRecorder()
	t.h.ServeHTTP(rec, req)

	// httptest doesn't bind Request to Response by default
	return rec.Result(), nil
}

func TestCustomTransporter(t *testing.T) {
	data, err := os.ReadFile(certPath)
	handleError(err)

	certs, key, err := azidentity.ParseCertificates(data, nil)
	handleError(err)

	testTenantID := "72f988bf-86f1-41af-91ab-2d7cd011db47"
	unitTestTransport := &unitTestTransport{
		h: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			if strings.HasSuffix(r.URL.Path, "openid-configuration") {
                                 // lazy man's proxy to the first oidc discovery step. Failure from this step won't go to `AuthenticationFailedError`
				oidcConfigResp, err := http.DefaultClient.Get(r.URL.String())
				if err != nil {
					t.Errorf("unexpected error: %s", err)
				}
				defer oidcConfigResp.Body.Close()
				w.WriteHeader(oidcConfigResp.StatusCode)
				_, err = io.Copy(w, oidcConfigResp.Body)
				if err != nil {
					t.Errorf("unexpected error: %s", err)
				}
			}

			w.WriteHeader(http.StatusBadRequest) // simulate a bad request
		}),
	}
	unitTestClient := &http.Client{Transport: unitTestTransport}

	opts := &azidentity.ClientCertificateCredentialOptions{
		DisableInstanceDiscovery: true,
			Transport: unitTestClient,
		},
	}
	cred, err = azidentity.NewClientCertificateCredential(testTenantID, clientID, certs, key, opts)
	handleError(err)

	_, err = cred.GetToken(context.TODO(), policy.TokenRequestOptions{
		Scopes: []string{"test-scope"},
	})
	if err == nil {
		t.Error("unexpected nil error")
	}
	fmt.Println(err) // %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)
	_ = err.Error()  // <- panic
}
```

No panic and print out the error object.

- How can we reproduce it?

See above snippet.

This usage is very common on unit / integration test scenarios, where we want to simulate the real world flows. Our team usually use `httptest` to create a transport object. However, according to httptest's implementation, `http.Response`'s `Request` doesn't guarantee to be presented, hence a panic will be raised when goes to this code path:

 https://github.com/Azure/azure-sdk-for-go/blob/f52645651a7e4a4ae3dacc7bca501319deeab882/sdk/azidentity/errors.go#L60

bcho:
Fixes #21809

### DIFF
--- a/sdk/azidentity/errors.go
+++ b/sdk/azidentity/errors.go
@@ -57,7 +57,13 @@ func (e *AuthenticationFailedError) Error() string {
 	}
 	msg := &bytes.Buffer{}
 	fmt.Fprintf(msg, e.credType+" authentication failed\n")
-	fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
+	if e.RawResponse.Request != nil {
+		fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
+	} else {
+		// this happens when the response is created from a custom HTTP transporter,
+		// which doesn't guarantee to bind the original request to the response
+		fmt.Fprint(msg, "Request information not available\n")
+	}
 	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
 	fmt.Fprintf(msg, "RESPONSE %s\n", e.RawResponse.Status)
 	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
--- a/sdk/azidentity/errors.go
+++ b/sdk/azidentity/errors.go
@@ -62,7 +62,7 @@ func (e *AuthenticationFailedError) Error() string {
 	} else {
 		// this happens when the response is created from a custom HTTP transporter,
 		// which doesn't guarantee to bind the original request to the response
-		fmt.Fprint(msg, "Request information not available\n")
+		fmt.Fprintln(msg, "Request information not available")
 	}
 	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
 	fmt.Fprintf(msg, "RESPONSE %s\n", e.RawResponse.Status)
--- a/sdk/azidentity/errors_test.go
+++ b/sdk/azidentity/errors_test.go
@@ -59,3 +59,48 @@ func TestAuthenticationFailedErrorInterface(t *testing.T) {
 		t.Fatalf("expected troubleshooting link: %s", resBodyString)
 	}
 }
+
+func TestAuthenticationFailedErrorWithoutResponse(t *testing.T) {
+	err := newAuthenticationFailedError(credNameAzureCLI, "error message", nil, nil)
+	if _, ok := err.(*AuthenticationFailedError); !ok {
+		t.Fatalf("expected AuthenticationFailedError, received %T", err)
+	}
+	errMsg := err.Error()
+	if errMsg != "AzureCLICredential: error message" {
+		t.Fatalf("unexpected error message returned: %s", errMsg)
+	}
+}
+
+func TestAuthenticationFailedErrorWithoutOriginalRequest(t *testing.T) {
+	resBodyString := "Test Error Response Body"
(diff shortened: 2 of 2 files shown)


sdk/azidentity/errors_test.go
```diff
@@ -59,3 +59,48 @@ func TestAuthenticationFailedErrorInterface(t *testing.T) {
...
+	}
+	err := newAuthenticationFailedError(credNameAzureCLI, "error message", res, nil)
+	if e, ok := err.(*AuthenticationFailedError); ok {
+		if e.RawResponse == nil {
+			t.Fatal("expected a non-nil RawResponse")
+		}
+	} else {
+		t.Fatalf("expected AuthenticationFailedError, received %T", err)
+	}
+	errMsg := err.Error()
```
bcho, line 90: FYI, it looks like this:

```
AzureCLICredential authentication failed
Request information not available
RESPONSE 400 Bad Request
Test Error Response Body
To troubleshoot, visit https://aka.ms/azsdk/go/identity/troubleshoot#azure-cli
```

sdk/azidentity/errors_test.go
```diff
@@ -59,3 +59,48 @@ func TestAuthenticationFailedErrorInterface(t *testing.T) {
...
+	if _, ok := err.(*AuthenticationFailedError); !ok {
+		t.Fatalf("expected AuthenticationFailedError, received %T", err)
+	}
+	errMsg := err.Error()
+	if errMsg != "AzureCLICredential: error message" {
+		t.Fatalf("unexpected error message returned: %s", errMsg)
+	}
+}
+
+func TestAuthenticationFailedErrorWithoutOriginalRequest(t *testing.T) {
```
chlowell, line 74: Nitpicking: "original" implies we might handle differently some other category of request
```suggestion
func TestAuthenticationFailedErrorWithoutRequest(t *testing.T) {
```

sdk/azidentity/errors.go
```diff
@@ -57,7 +57,13 @@ func (e *AuthenticationFailedError) Error() string {
 	}
 	msg := &bytes.Buffer{}
 	fmt.Fprintf(msg, e.credType+" authentication failed\n")
-	fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
+	if e.RawResponse.Request != nil {
+		fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
+	} else {
+		// this happens when the response is created from a custom HTTP transporter,
+		// which doesn't guarantee to bind the original request to the response
+		fmt.Fprint(msg, "Request information not available\n")
```
chlowell, line 65:
```suggestion
		fmt.Fprintln(msg, "Request information not available")
```

chlowell (approved):
Thanks for doing this!

bcho:
@chlowell I have merged with your suggestions, PTAL!

chlowell (approved):

Issue #21674, opened 2023-10-02
The bearer token authentication policy should reject URLs that are not HTTPS.


ahsonkhan:
Similar to C++, .NET, and other language SDKs, consider adding validation that the scheme of the request URL within the bearer token authentication policy is HTTPS. Otherwise, fail.

https://github.com/Azure/azure-sdk-for-go/blob/790fea5020e687e1d6122c55fbd8c09d1d7e9fc4/sdk/azcore/runtime/policy_bearer_token.go#L72

For example, see these:
https://github.com/Azure/azure-sdk-for-cpp/blob/35e42b99cfcc5b2797aa278a39694030732da2bc/sdk/core/azure-core/src/http/bearer_token_authentication_policy.cpp#L24-L28
https://github.com/Azure/azure-sdk-for-net/blob/e902ef5a2ed5db3d6104ccb2db29b97a02258634/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L115-L118

PR #21832, opened 2023-10-25
Require TLS protected endpoints for key and SAS authentication

jhendrixMSFT:
Return consistent error message for all failed HTTPS checks. Fix misspelled file name.

### DIFF
--- a/sdk/azcore/runtime/policy_bearer_token.go
+++ b/sdk/azcore/runtime/policy_bearer_token.go
@@ -72,8 +72,8 @@ func (b *BearerTokenPolicy) authenticateAndAuthorize(req *policy.Request) func(p
 
 // Do authorizes a request with a bearer token
 func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
-	if strings.ToLower(req.Raw().URL.Scheme) != "https" {
-		return nil, shared.NonRetriableError(errors.New("bearer token authentication is not permitted for non TLS protected (https) endpoints"))
+	if err := checkHTTPSForAuth(req); err != nil {
+		return nil, err
 	}
 	var err error
 	if b.authzHandler.OnRequest != nil {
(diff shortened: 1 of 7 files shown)


jhendrixMSFT:
In response to https://github.com/Azure/azure-sdk-for-go/issues/21674

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -72,8 +72,8 @@ func (b *BearerTokenPolicy) authenticateAndAuthorize(req *policy.Request) func(p
 
 // Do authorizes a request with a bearer token
 func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
-	if strings.ToLower(req.Raw().URL.Scheme) != "https" {
-		return nil, shared.NonRetriableError(errors.New("bearer token authentication is not permitted for non TLS protected (https) endpoints"))
+	if err := checkHTTPSForAuth(req); err != nil {
+		return nil, err
```
RickWinter, line 76: Could we have a single policy to check for SSL rather than requiring each token policy to add this check?

jhendrixMSFT: The requirement for the check is tightly coupled to authentication.  E.g. `azcore.NewClient()` doesn't add any kind of authentication policy to the pipeline, the caller must add it. I do see the value though in having the check be centralized. I will think on this a bit but, at present, I don't see an easy way to solve this.

jhendrixMSFT: I suppose we could expose a new pipeline policy that performs the `https` check. It would have to be publicly exposed so that SDKs can take advantage of it. This approach would mean that _all_ SDKs that include an authentication policy in their pipelines would need to include this new policy. Put another way, the current approach lets us centralize the checks. A new pipeline policy would require updating all SDKs as required.

jhendrixMSFT: Having a separate policy for `https` validation also means that the authentication policies are insecure by default.

chlowell (approved):

Issue #21838, opened 2023-10-26
(&azcore.ResponseError{}).Error() panics


jim-minter:
### Bug Report

- import path of package in question: "github.com/Azure/azure-sdk-for-go/sdk/azcore"
- SDK version: v1.8.0
- output of `go version`: go version go1.21.1 linux/amd64

The expression `(&azcore.ResponseError{}).Error()` panics because the implementation of `Error()` requires multiple internal fields to be set.

The minimum instantiation of `azcore.ResponseError` that doesn't panic when `Error()` is called is something like:
```
(&azcore.ResponseError{
	RawResponse: &http.Response{
		Request: &http.Request{
			URL: &url.URL{},
		},
		Body: io.NopCloser(bytes.NewReader(nil)),
	},
}).Error()
```

This isn't something that we see is causing a production problem (although perhaps there is a minor robustness issue in the highly improbable production case that any of these fields is not set).

But, it's something that trips us up from time to time in a unit test context - when unit testing some other behavior, it's easy to inadvertently instantiate a ResponseError that will then cause a panic when Error() is called.

If only to simplify unit testing, it would be neat if `(*azcore.ResponseError) Error()` did not panic on a zero value `&azcore.ResponseError{}`.

jhendrixMSFT:
Fix will go out in `azcore@v1.9.0` in November.

PR #21839, opened 2023-10-26
Prevent panic in ResponseError.Error()

jhendrixMSFT:
If RawResponse is nil (unit test scenarios), print "Missing RawResponse" instead of panicing.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21838

### DIFF
--- a/sdk/azcore/internal/exported/response_error.go
+++ b/sdk/azcore/internal/exported/response_error.go
@@ -113,33 +113,41 @@ type ResponseError struct {
 // Error implements the error interface for type ResponseError.
 // Note that the message contents are not contractual and can change over time.
 func (e *ResponseError) Error() string {
+	const separator = "--------------------------------------------------------------------------------"
 	// write the request method and URL with response status code
 	msg := &bytes.Buffer{}
-	fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
-	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
-	fmt.Fprintf(msg, "RESPONSE %d: %s\n", e.RawResponse.StatusCode, e.RawResponse.Status)
+	if e.RawResponse != nil {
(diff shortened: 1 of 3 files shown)


RickWinter (approved):
The change seems reasonable.  Would guidance to users be to log and stop?  We should give guidance on how to handle this case since now if it happened in production the customer needs to action

jhendrixMSFT:
Customers should _never_ be making programmatic choices based on the error message string as they are subject to change. Or did I misunderstand the question?

RickWinter:
> Customers should _never_ be making programmatic choices based on the error message string as they are subject to change. Or did I misunderstand the question?

Before this change the app terminated with a panic.  now it will return an error.  The customer will need to do something... We should guide them on how to handle this newfound freedom.

jhendrixMSFT:
SDKs only return an `*azcore.ResponseError` when the service returns a 400 (and possibly other) HTTP status code, which means we have a `*http.Response`. If a SDK method were ever to return an `*azcore.ResponseError` with a `nil` `RawResponse` field it means we have a bug either in the SDK or `azcore`.

sdk/azcore/internal/exported/response_error.go
```diff
@@ -113,33 +113,41 @@ type ResponseError struct {
...
 // Note that the message contents are not contractual and can change over time.
 func (e *ResponseError) Error() string {
+	const separator = "--------------------------------------------------------------------------------"
 	// write the request method and URL with response status code
 	msg := &bytes.Buffer{}
-	fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
-	fmt.Fprintln(msg, "--------------------------------------------------------------------------------")
-	fmt.Fprintf(msg, "RESPONSE %d: %s\n", e.RawResponse.StatusCode, e.RawResponse.Status)
+	if e.RawResponse != nil {
+		fmt.Fprintf(msg, "%s %s://%s%s\n", e.RawResponse.Request.Method, e.RawResponse.Request.URL.Scheme, e.RawResponse.Request.URL.Host, e.RawResponse.Request.URL.Path)
```
chlowell, line 120: `e.RawResponse.Request` could also be nil, so I suppose we ought to consider that as well a la #21810

chlowell (approved):

sdk/azcore/internal/exported/response_error.go
jhendrixMSFT: Good catch, fixed.

Issue #21784, opened 2023-10-17
azcore: traces for credential operations are showing up when not explicitly configured


jhendrixMSFT:
This is because `runtime.StartSpan` places the tracer in the context which is then propagated to `GetToken()`.  And if the pipeline for fetching tokens contains the HTTP trace policy (which it will), it's then fetched from the context and HTTP traces are created.

PR #21842, opened 2023-10-27
Add tracing.Tracer to requests instead of context

jhendrixMSFT:
This prevents traces from showing up for clients that didn't explicitly enable tracing via their ClientOptions. This required moving the tracer into the pipeline and out of the client.
Added TracingOptions type for configuring tracing via pipeline options. Creating a raw pipeline with a tracing provider will now create traces. Don't create a HTTP span if a SDK span hasn't been created. This prevents HTTP spans from showing up when tracing is enabled but a SDK hasn't yet incorporated spans.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21784

### DIFF
--- a/sdk/azcore/core.go
+++ b/sdk/azcore/core.go
@@ -92,7 +92,6 @@ type ClientOptions = policy.ClientOptions
 // Client is a basic HTTP client.  It consists of a pipeline and tracing provider.
 type Client struct {
 	pl runtime.Pipeline
-	tr tracing.Tracer
 
 	// cached on the client to support shallow copying with new values
 	tp        tracing.Provider
@@ -103,7 +102,7 @@ type Client struct {
 // NewClient creates a new Client instance with the provided values.
 //   - clientName - the fully qualified name of the client ("module/package.Client"); this is used by the telemetry policy and tracing provider.
 //     if module and package are the same value, the "module/" prefix can be omitted.
-//   - moduleVersion - the semantic version of the containing module; used by the telemetry policy
+//   - moduleVersion - the semantic version of the containing module; used by the telemetry policy and tracing provider.
 //   - plOpts - pipeline configuration options; can be the zero-value
 //   - options - optional client configurations; pass nil to accept the default values
 func NewClient(clientName, moduleVersion string, plOpts runtime.PipelineOptions, options *ClientOptions) (*Client, error) {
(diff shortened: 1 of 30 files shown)


jhendrixMSFT:
@JeffreyRichter this is different that the approach we originally discussed. Turns out that passing the `tracing.Tracer` via `runtime.NewRequestWithOptions` becomes complicated for pagers/pollers as they need to create `policy.Request` and while we can solve that, the code becomes quite ugly.

sdk/azcore/internal/exported/pipeline.go
```diff
@@ -57,21 +60,35 @@ func (tp transportPolicy) Do(req *Request) (*http.Response, error) {
...
+
 // Do is called for each and every HTTP request. It passes the request through all
 // the Policy objects (which can transform the Request's URL/query parameters/headers)
 // and ultimately sends the transformed HTTP request over the network.
 func (p Pipeline) Do(req *Request) (*http.Response, error) {
 	if req == nil {
 		return nil, errors.New("request cannot be nil")
 	}
+	// add the tracer to the in-flight request so any pipeline policies can access it
+	req.SetOperationValue(p.tracer)
```
jhendrixMSFT, line 86: This is the core of the change. Maybe a good way to frame this is that for tracing, we always want to send the `tracing.Tracer` with the associated `Request`. Since `tracing.Tracer` is now a pipeline-internal constructor parameter, we always have access to it and can easily send it down the pipeline.

sdk/azcore/internal/exported/pipeline.go
```diff
@@ -57,21 +60,35 @@ func (tp transportPolicy) Do(req *Request) (*http.Response, error) {
...
 	req.policies = p.policies
 	return req.Next()
 }
+
+// Tracer returns the [tracing.Tracer] associated with this pipeline.
+func (p Pipeline) Tracer() tracing.Tracer {
```
jhendrixMSFT, line 92: This provides access to the pipeline's `tracing.Tracer` and is new public surface area. Clients need access to this.

sdk/azcore/core.go
```diff
@@ -125,16 +124,18 @@ func NewClient(clientName, moduleVersion string, plOpts runtime.PipelineOptions,
...
 
 	tr := options.TracingProvider.NewTracer(client, moduleVersion)
-	if tr.Enabled() && plOpts.TracingNamespace != "" {
-		tr.SetAttributes(tracing.Attribute{Key: "az.namespace", Value: plOpts.TracingNamespace})
+	if tr.Enabled() && plOpts.Tracing.Namespace != "" {
+		tr.SetAttributes(tracing.Attribute{Key: shared.TracingNamespaceAttrName, Value: plOpts.Tracing.Namespace})
 	}
 
+	// replace the vanilla tracer with this client-specific one
+	exported.SetTracer(&pl, tr)
```
jhendrixMSFT, line 132: I admit that I don't totally love this part, but it seemed better than adding a new, exported, pipeline constructor that takes a `tracing.Tracer`.

sdk/azcore/internal/exported/pipeline.go
```diff
@@ -57,21 +60,35 @@ func (tp transportPolicy) Do(req *Request) (*http.Response, error) {
...
+		tracer:   tr,
 	}
 }
 
+// SetTracer sets the provided [tracing.Tracer] on [Pipeline].
+// Not directly exported, but used when creating/cloning clients.
+func SetTracer(pl *Pipeline, tracer tracing.Tracer) {
+	pl.tracer = tracer
+}
+
```
JeffreyRichter, line 77: Pipelines are supposed to be immutable and this makes them mutable. I see that you say "used when crating/cloning clients" so, if this is used correctly then this shouldn't be a problem, but it may not always be used correctly. Also (assuming we keep this) , why wouldn't this be a method with a *Pipeline receiver? 
In addition. this can fix the tracer problem but what about the context capturing the raw response or having custom headers or some other feature - this context should also not flow to the credential's pipeline. I think we need a fix for this problem still and ideally one solution could fix all these issues.

sdk/azcore/internal/shared/constants.go
```diff
@@ -31,6 +31,8 @@ const (
 
 const BearerTokenPrefix = "Bearer "
 
+const TracingNamespaceAttrName = "az.namespace"
```
JeffreyRichter, line 34: In anticipation of the unbranding, we may want to stop using "az" whenever we can

jhendrixMSFT:
Closing in favor of https://github.com/Azure/azure-sdk-for-go/pull/21863

Issue #19467, opened 2022-11-01
Add nonretriable error wrapper to sdk/internal


chlowell:
Such a wrapper is useful enough to deserve a single, shared implementation. See for example https://github.com/Azure/azure-sdk-for-go/pull/19448/files#diff-3d863849a5f2aa86242d7ce3874942d9cf755b57a4fc63b7bc9dcc58a5331656R109-R114

PR #21845, opened 2023-10-27
Add Non Retriable Error Wrapper for all SDK's

MBSolomon:
This PR adds a Non Retriable Error Wrapper to the azure-sdk-for-go repository. #19467 


sdk/internal/shared/shared.go
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License.
+
+package shared
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
+)
+
+// NonRetriableError marks the specified error as non-retriable.
+func NonRetriableError(err error) error {
```
jhendrixMSFT, line 14: Might simplify things to move this to the `errorinfo` package so this and `NonRetriable` are together.

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+package errorinfo
+
+import (
+	"testing"
+)
+
+func TestNonRetriableError(t *testing.T) {
+	//Create sample error.
+	err := NonRetriableError(NewSampleError("Do Not Retry"))
```
jhendrixMSFT, line 15: You could simplify this a bit by saying `errors.New("Do Not Retry")` and remove `NewSampleError` and friends.

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Check if the error is of type NonRetriableError
+	if _, ok := err.(*nonRetriableError); !ok {
```
richardpark-msft, line 18: The concrete type isn't necessarily the important part - it's just an implementation detail. You'll want to check that your type conforms to the "NonRetriable" shape by using an interface.

For example:

```go
var nonRetriableErr *interface { NonRetriable() }
require.ErrorAs(err, &nonRetriableErr)
```

sdk/internal/errorinfo/errorinfo.go
```diff
@@ -14,3 +14,24 @@ type NonRetriable interface {
 	error
 	NonRetriable()
 }
+
+// NonRetriableError marks the specified error as non-retriable.
+func NonRetriableError(err error) error {
+	return &nonRetriableError{err}
```
richardpark-msft, line 20: It doesn't seem like it needs to be a pointer, btw.

sdk/internal/errorinfo/errorinfo_test.go
jhendrixMSFT: Or use the predefined interface type.
```go
var nonRetriableErr *NonRetriable
```

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	// Check error message is correct
+	errMsg := "Do Not Retry"
+	if err.Error() != errMsg {
+		t.Fatalf("Expected error message to be '%q' but got '%q'.", errMsg, err.Error())
+	}
+}
+
+func NewSampleError(message string) error {
```
richardpark-msft, line 29: If you want there's already a convenient "create an error based on a string" with `errors.New`.

(For future stuff, if you want to format the error there's `fmt.Errorf`.)

sdk/internal/errorinfo/errorinfo.go
jhendrixMSFT: By convention we return errors as pointer-to-type. Plus, the receivers on `nonRetriableError` are by pointer.

sdk/internal/errorinfo/errorinfo_test.go
richardpark-msft: (if the intention here was to make sure you could "unwrap" the inner type then you'll want to do an `errors.As()` in your test too.)

sdk/internal/errorinfo/errorinfo.go
MBSolomon: Discussed with Richard offline. Resolving comment

sdk/internal/errorinfo/errorinfo.go
```diff
@@ -14,3 +14,24 @@ type NonRetriable interface {
...
+}
+
+type nonRetriableError struct {
```
RickWinter, line 23: Add doc comments for each of the methods.

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNonRetriableError(t *testing.T) {
+	//Create sample error.
+	err := NonRetriableError(errors.New("Do Not Retry"))
```
jhendrixMSFT, line 18: Nit. Make `Do Not Retry` a constant so you don't have to duplicate the value.

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// Check error message is correct
+	errMsg := "Do Not Retry"
+	if err.Error() != errMsg {
```
jhendrixMSFT, line 22: Can simplify to `require.EqualValues(t, errMsg, err.Error())`

sdk/internal/errorinfo/errorinfo_test.go
```diff
@@ -9,20 +9,21 @@ package errorinfo
...
-	errMsg := "Do Not Retry"
-	if err.Error() != errMsg {
-		t.Fatalf("Expected error message to be '%q' but got '%q'.", errMsg, err.Error())
+	if err.Error() != dnr {
+		t.Fatalf("Expected error message to be '%q' but got '%q'.", dnr, err.Error())
 	}
 
-	var nonRetriableErr interface{ NonRetriable() }
-	require.True(t, errors.As(&nonRetriableError{}, &nonRetriableErr))
+	var e *nonRetriableError
```
jhendrixMSFT, line 25: This should be.
```go
var e *NonRetriable
```
This ensures that `err` implements the `NonRetriable` interface.

RickWinter (approved):

jhendrixMSFT (approved):

Issue #21396, opened 2023-08-18
Managed identity authentication on AzureML Compute resources broken


maben-mw:
### Bug Report

- SDK version `master`, ([99c0a04](https://github.com/Azure/azure-sdk-for-go/commit/99c0a042b1a228db9cd1f0c54df095f24f13cf28)

- What happened?

On AzureML Compute instances, managed identity authentication is not working with current Go SDK version.

It appears that AzureML uses MSI "2017 style" managed identities with an `MSI_ENDPOINT` and `MSI_SECRET` and where the identity is specified using query parameter `clientid` (rather than `client_id`).

I.e. I can obtain the tokens I need successfully with:

```console
curl "${MSI_ENDPOINT}?resource=https://storage.azure.com&clientid=${DEFAULT_IDENTITY_CLIENT_ID}" --header "Secret: ${MSI_SECRET}"
```

In Java and Python SDK I can use these managed identities without issue.

In the Go SDK it appears this used to be supported in the past, but the option got removed (hence reporting this as a bug rather than an enhancement):

https://github.com/Azure/azure-sdk-for-go/commit/d67aff0d17e2cd6e1bf4c341b565bb88366f2dd4

chlowell:
Could you please describe your scenario i.e., how you use Go with Azure ML, and point me to some documentation describing how to deploy Go to Azure ML (I don't find any)? We'd need at least that much to test this feature.

You're correct about the history--`azidentity` coincidentally supported Azure ML's managed identity API until v0.13.2 because that API mimics a version of App Service's managed identity API. When App Service rolled out a new version, we stopped supporting the old one because we didn't see a reason to continue supporting it in `azidentity` v1.x.

maben-mw:
I don't directly use the Azure Go SDK, what I do is:

1. Create a Compute Instance on AzureML with an user assigned managed identity:

   https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-compute-instance?view=azureml-api-2&tabs=azure-studio

[...]

PR #21851, opened 2023-10-30
Support Azure ML managed identity

chlowell:
Though Azure ML doesn't directly support deploying Go, this support in the SDK will enable using Go tools in Azure ML compute environments, for example to move data (see #21396).

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -28,12 +28,14 @@ import (
 
 const (
 	arcIMDSEndpoint          = "IMDS_ENDPOINT"
+	defaultIdentityClientID  = "DEFAULT_IDENTITY_CLIENT_ID"
 	identityEndpoint         = "IDENTITY_ENDPOINT"
 	identityHeader           = "IDENTITY_HEADER"
 	identityServerThumbprint = "IDENTITY_SERVER_THUMBPRINT"
 	headerMetadata           = "Metadata"
 	imdsEndpoint             = "http://169.254.169.254/metadata/identity/oauth2/token"
 	msiEndpoint              = "MSI_ENDPOINT"
+	msiSecret                = "MSI_SECRET"
 	imdsAPIVersion           = "2018-02-01"
 	azureArcAPIVersion       = "2019-08-15"
 	serviceFabricAPIVersion  = "2019-07-01-preview"
@@ -47,6 +49,7 @@ type msiType int
 const (
 	msiTypeAppService msiType = iota
 	msiTypeAzureArc
+	msiTypeAzureML
 	msiTypeCloudShell
 	msiTypeIMDS
 	msiTypeServiceFabric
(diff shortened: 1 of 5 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -296,6 +306,29 @@ func (c *managedIdentityClient) createAppServiceAuthRequest(ctx context.Context,
...
+		return nil, err
+	}
+	request.Raw().Header.Set("secret", os.Getenv(msiSecret))
+	q := request.Raw().URL.Query()
+	q.Add("api-version", "2017-09-01")
+	q.Add("resource", strings.Join(scopes, " "))
+	q.Add("clientid", os.Getenv(defaultIdentityClientID))
+	if id != nil {
+		if id.idKind() == miResourceID {
+			log.Write(EventAuthentication, "WARNING: Azure ML doesn't support specifying a managed identity by resource ID")
```
jhendrixMSFT, line 321: Does this effectively become a no-op, or should this be a hard error?

chlowell: Personally I think it should be a hard error because we have no reason to believe we can realize the caller's intent and the result depends on how the server handles the undocumented and probably unexpected parameter. But I lost that argument and we took this approach across all languages, so I'm following along here for the sake of consistency.

chlowell: I'm hesitant to document this because the blurb would be complex (there are 3 other platforms on which we similarly set unspecified parameters), I believe it would generally be more distracting than helpful, and it could be rendered incorrect at any time by a service change we wouldn't observe. But if you think it's for the best, I'll take a stab at it.

jhendrixMSFT: Good point. I suppose we can leave it as-is for now and reassess based on customer feedback.

jhendrixMSFT (approved):

sdk/azidentity/managed_identity_client.go
```diff
@@ -28,12 +28,14 @@ import (
 
 const (
 	arcIMDSEndpoint          = "IMDS_ENDPOINT"
+	defaultIdentityClientID  = "DEFAULT_IDENTITY_CLIENT_ID"
```
RickWinter, line 31: Could get better grouping if we named it starting with IDENTITY
```suggestion
	identityDefaultClientID  = "IDENTITY_DEFAULT_CLIENT_ID"
```

RickWinter (approved):

sdk/azidentity/managed_identity_client.go
chlowell: This is Azure ML's variable--we have to use the name they chose

RickWinter: Can we push back on them to change that?

chlowell: I suppose we could try but because they've already shipped and documented this variable we'd be asking them to ship and document a new version of their managed identity endpoint. I imagine it would be very difficult to convince them changing this name is worth that effort.

PR #21874, opened 2023-11-02
Update data plane to use latest recording framework

benbp:
This PR updates data plane test packages to use the latest recording framework in sdk/internal v1.4.0. This includes proxy auto-install and start/stop functionality over `ResetProxy` for better parallel support (multiple proxy instances run on separate ports).


richardpark-msft (approved):
Changes look good for `azopenai` and `azeventgrid`.

sdk/azidentity/go.mod
```diff
@@ -4,20 +4,26 @@ go 1.18
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.8.0
-	github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity/cache v0.2.0
```
chlowell, line 7: `azidentity` mustn't require `azidentity/cache`
```suggestion
```

chlowell (changes requested):

souravgupta-msft (approved):
Looks good for storage SDK - `azblob`, `azdatalake`, `azfile` and `azqueue`.

sdk/azidentity/go.mod
jhendrixMSFT: This comes from `example_cache_test.go`. This _should_ be ok as starting with Go 1.17 modules are [lazily loaded](https://go.dev/ref/mod#lazy-loading). Other option is to move this example out of `azidentity` and into the `samples` directory.

jhendrixMSFT: Ok. I've opened https://github.com/Azure/azure-sdk-for-go/issues/21881 to track this as every time we `go mod tidy` the requirement will rematerialize (it's showing up in some pending work I'm doing for tracing).

jhendrixMSFT: I spoke with @chlowell about this, we can just leave the import so everything is current.

chlowell (approved):

PR #21885, opened 2023-11-03
Add tracing spans to azidentity credentials

jhendrixMSFT:
Replaced runtime.Pipelines with azcore.Clients.
Removed the pipeline adapter. Now, the confidential and public clients satisfy the MSAL HTTPClient interface.


sdk/azidentity/confidential_client.go
```diff
@@ -42,13 +41,22 @@ type confidentialClient struct {
 	name                     string
 	opts                     confidentialClientOptions
 	region                   string
+	azClient                 *azcore.Client
 }
 
-func newConfidentialClient(tenantID, clientID, name string, cred confidential.Credential, opts confidentialClientOptions) (*confidentialClient, error) {
+func newConfidentialClient(clientName, tenantID, clientID, name string, cred confidential.Credential, opts confidentialClientOptions, clientOpts azcore.ClientOptions) (*confidentialClient, error) {
```
jhendrixMSFT, line 47: We might not need this `clientName` param as I think we might be incorrectly setting the `otel.library.name` span attribute (I'm following up about this).

sdk/azidentity/workload_identity.go
```diff
@@ -83,6 +83,7 @@ func NewWorkloadIdentityCredential(options *WorkloadIdentityCredentialOptions) (
 		ClientOptions:              options.ClientOptions,
 		DisableInstanceDiscovery:   options.DisableInstanceDiscovery,
 	}
+	// TODO: this will incorrectly report ClientAssertionCredential in traces
```
jhendrixMSFT, line 86: If we end up not needing the `clientName` param cited earlier, this problem goes away.

jhendrixMSFT: Fixed.

sdk/azidentity/confidential_client.go
jhendrixMSFT: This has been removed per https://github.com/Azure/azure-sdk-for-go/pull/21893

sdk/azidentity/confidential_client.go
```diff
@@ -23,8 +24,6 @@ import (
 )
 
 type confidentialClientOptions struct {
-	azcore.ClientOptions
```
chlowell, line 26: What's the rationale for unembedding this? Is it a matter of style?

jhendrixMSFT: It seems like the `azcore.ClientOptions` are conceptually separate from the confidential and public client options which are MSAL specific, so I split them out. Purely cosmetic though, so I can revert this part if you prefer.

chlowell: I do prefer bundling all the options together because MSAL is an implementation detail of these wrapper types and the options are, at least in the abstract, all for the wrapper.

jhendrixMSFT: Good point, fixed.

chlowell (approved):

PR #21897, opened 2023-11-06
Fix keyvault test recording options. Update docs

benbp:
- Fix keyvault tests using recording options with correct process port - we were using the default options which don't include the newly assigned proxy port
- Fix tabs in developer setup docs


scbedd (approved):
👍

gracewilcox (approved):

PR #21903, opened 2023-11-07
Add app configuration internal support for snapshot

MBSolomon:
Adds App Configuration support internally for Snapshots


sdk/data/azappconfig/client.go
```diff
@@ -141,7 +141,9 @@ func (c *Client) GetSetting(ctx context.Context, key string, options *GetSetting
 
 	var lastModified *time.Time
 	if resp.LastModified != nil {
-		tt, err := time.Parse(http.TimeFormat, *resp.LastModified)
+		slm := *resp.LastModified
+		tflm := slm.Format(http.TimeFormat)
+		tt, err := time.Parse(http.TimeFormat, tflm)
```
jhendrixMSFT, line 146: I think these changes should be reverted?

MBSolomon: Yes we can revert this. It is displaying as an error locally with a failure to convert the `*resp.LastModified` (type time.time) to string. I reran it and this doesn't change behavior

MBSolomon: We do still need this change. Build is failing:
image

jhendrixMSFT: Looking closer, I see that the `LastModified` field in type `AzureAppConfigurationClientGetKeyValueResponse` has been removed. It appears to overlap with a field with the same name in `KeyValue` which is already in `time.Time` format. So we shouldn't need to perform any parsing/conversion.

sdk/data/azappconfig/client.go
```diff
@@ -139,19 +138,10 @@ func (c *Client) GetSetting(ctx context.Context, key string, options *GetSetting
...
-			return GetSettingResponse{}, err
-		}
-		lastModified = &tt
-	}
-
 	return GetSettingResponse{
 		Setting:      settingFromGenerated(resp.KeyValue),
 		SyncToken:    SyncToken(*resp.SyncToken),
-		LastModified: lastModified,
+		LastModified: resp.LastModified,
```
jhendrixMSFT, line 144: So this is interesting. Originally, `GetSettingResponse.LastModified` was populated from the `Last-Modified` header value which is no longer the case. Maybe it was duplicate data? If the value of `GetSettingResponse.KeyValue.LastModified` is the same, then your change is fine even though it's not technically accurate (i.e. it's not coming from the `Last-Modified` header). Assuming that's the case, this should be slightly changed.
```go
LastModified: resp.KeyValue.LastModified
```

MBSolomon: yes. This is right. Updating

jhendrixMSFT (dismissed):

PR #21909, opened 2023-11-07
Enable spans for azadmin

jhendrixMSFT:
Updated to latest version of the code generator.
Added spans to the custom poller used in the backup package.


gracewilcox (approved):

sdk/security/keyvault/azadmin/internal/version.go
```diff
@@ -7,6 +7,6 @@
 package internal
 
 const (
-	ModuleName = "azadmin"
+	ModuleName = "github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azadmin"
```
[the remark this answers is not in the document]
jhendrixMSFT: For spans we use the full module identity. However, we still use just the package name e.g. `azadmin` for telemetry like before this change.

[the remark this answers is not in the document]
jhendrixMSFT: For more context, the `moduleName` parameter in `azcore.NewClient()` is used to populate the `otel.library.name` attribute in spans. This used to be something like "packageName+clientName" however, that turned out to be incorrect (see [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) for docs on this).

sdk/security/keyvault/azadmin/backup/autorest.md
```diff
@@ -94,4 +96,9 @@ directive:
   - from: swagger-document
     where: $.definitions.SelectiveKeyRestoreOperationParameters
     transform: $["description"] = "Parameters for the selective restore operation"
+
+  # fix up span names
+  - from: client.go
+    where: $
+    transform: return $.replace(/StartSpan\(ctx, "Client/, "StartSpan(ctx, \"backup.Client");
```
MBSolomon, line 103: Do we need this here? Are we able to replace StartSpan(ctx, "client... with StartSpan(ctx, "backup.Client.... in client.go since this isn't modifying generated files? Or what is the benefit of adding this transform?

jhendrixMSFT: For the key vault SDKs, we do directly expose the generated code albeit with some build-time transforms to get the shape we want. So, this transform does impact the generated code.

That said, I did this in an effort to disambiguate the client names in the traces. For this SDK there are three clients across three subpackages, each with the name `Client`. So, I modified the span name to include the package name.

jhendrixMSFT: I will point out though that there are no common methods (at present) across the clients. So, one could make the argument that this is enough of a distinction that the package prefix isn't required. I prefer keeping it for being explicit but am open to options.

jhendrixMSFT: I will also look into seeing if we can automatically do this in the code generator. I opened https://github.com/Azure/autorest.go/issues/1071 to track that.

PR #21919, opened 2023-11-08
Add spans to azappconfig

jhendrixMSFT:

### DIFF
--- a/sdk/data/azappconfig/internal/generated/custom_client.go
+++ b/sdk/data/azappconfig/internal/generated/custom_client.go
@@ -8,6 +8,7 @@ package generated
 
 import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/tracing"
 )
 
 func NewAzureAppConfigurationClient(endpoint string, client *azcore.Client) *AzureAppConfigurationClient {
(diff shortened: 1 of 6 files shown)


MBSolomon (commented):
I see that in AzCore we don't have this Tracer in Internal : https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/core.go#L140C1-L143C2
Should we make these changes consistent?

Also for a different question why is this file in our generated?

sdk/data/azappconfig/internal/generated/custom_client.go
```diff
@@ -16,3 +17,7 @@ func NewAzureAppConfigurationClient(endpoint string, client *azcore.Client) *Azu
 		endpoint: endpoint,
 	}
 }
+
+func (a *AzureAppConfigurationClient) Tracer() tracing.Tracer {
+	return a.internal.Tracer()
+}
```
MBSolomon, line 23: I see that in AzCore we don't have this Tracer in Internal : https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/core.go#L140C1-L143C2
Should we make these changes consistent?

Also for a different question why is this file in our generated?

jhendrixMSFT: We don't generate client constructors for data-plane services as there's no consistent authentication across services. So, we hand-write them in a `custom_client.go` file next to the generated code.

In this instance, `internal` is the internal `*azcore.Client` used by the `azappconfig.Client`. So there's no need for an `internal` in the `azcore` version.

MBSolomon (approved):

PR #21920, opened 2023-11-08
Enable spans for aztables

jhendrixMSFT:

### DIFF
--- a/sdk/data/aztables/client.go
+++ b/sdk/data/aztables/client.go
@@ -125,6 +125,10 @@ func createTableResponseFromGen(g *generated.TableClientCreateResponse) CreateTa
 // NOTE: creating a table with the same name as a table that's in the process of being deleted will return an *azcore.ResponseError
 // with error code TableBeingDeleted and status code http.StatusConflict.
 func (t *Client) CreateTable(ctx context.Context, options *CreateTableOptions) (CreateTableResponse, error) {
+	var err error
+	ctx, endSpan := runtime.StartSpan(ctx, "Client.CreateTable", t.client.Tracer(), nil)
+	defer func() { endSpan(err) }()
+
 	if options == nil {
 		options = &CreateTableOptions{}
 	}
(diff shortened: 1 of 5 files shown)


sdk/data/aztables/client.go
```diff
@@ -125,6 +125,10 @@ func createTableResponseFromGen(g *generated.TableClientCreateResponse) CreateTa
 // NOTE: creating a table with the same name as a table that's in the process of being deleted will return an *azcore.ResponseError
 // with error code TableBeingDeleted and status code http.StatusConflict.
 func (t *Client) CreateTable(ctx context.Context, options *CreateTableOptions) (CreateTableResponse, error) {
+	var err error
```
richardpark-msft, line 128: I think we need to change the underlying line at 135, as well, to assign rather than shadow the 'err' declaration here.

richardpark-msft: (I'm guessing this pattern needs to be applied in multiple locations)

jhendrixMSFT: They're both in the same scope so there's no shadowing.

richardpark-msft: I see, in this case the func() {} wrapper to defer means we use the latest declared 'err'. Ie, it's purposeful shadowing.

richardpark-msft (commented):
Tests?

sdk/data/aztables/client.go
jhendrixMSFT: To be pedantic, there is no shadowing here as all declarations of `err` are within the same scope. See the section [declarations and scope](https://go.dev/ref/spec#Declarations_and_scope) for the details.

richardpark-msft: Pedantic is good! I see what you're saying now.

jhendrixMSFT:
Added tests to verify spans are created with the expected names and status.

sdk/data/aztables/tracing_test.go
jhendrixMSFT, line 1: If we think we want to perform this validation for all spans that are hand-written then we should refine this and move it to the `internal` module.

richardpark-msft (approved):

PR #21921, opened 2023-11-08
Fix regex to update all span names

jhendrixMSFT:
Regenered code removed JSON tag from previous bad merge.

### DIFF
--- a/sdk/security/keyvault/azadmin/backup/client.go
+++ b/sdk/security/keyvault/azadmin/backup/client.go
@@ -110,7 +110,7 @@ func (client *Client) BeginFullRestore(ctx context.Context, restoreBlobDetails R
 // Generated from API version 7.5-preview.1
 func (client *Client) fullRestore(ctx context.Context, restoreBlobDetails RestoreOperationParameters, options *BeginFullRestoreOptions) (*http.Response, error) {
 	var err error
-	ctx, endSpan := runtime.StartSpan(ctx, "Client.BeginFullRestore", client.internal.Tracer(), nil)
+	ctx, endSpan := runtime.StartSpan(ctx, "backup.Client.BeginFullRestore", client.internal.Tracer(), nil)
 	defer func() { endSpan(err) }()
 	req, err := client.fullRestoreCreateRequest(ctx, restoreBlobDetails, options)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


sdk/security/keyvault/azadmin/backup/models.go
```diff
@@ -70,12 +70,12 @@ type SASTokenParameters struct {
 	StorageResourceURI *string
 
 	// The SAS token pointing to an Azure Blob storage container
-	Token *string `json:"token,omitempty"`
```
gracewilcox, line 73: Strange this didn't get removed in the first PR

sdk/security/keyvault/azadmin/backup/client.go
```diff
@@ -110,7 +110,7 @@ func (client *Client) BeginFullRestore(ctx context.Context, restoreBlobDetails R
 // Generated from API version 7.5-preview.1
 func (client *Client) fullRestore(ctx context.Context, restoreBlobDetails RestoreOperationParameters, options *BeginFullRestoreOptions) (*http.Response, error) {
```
gracewilcox, line 111: Also, is this code still up to date with the new code generator? Because of the custom poller, I moved this code to custom_client.go. It might not follow the format of the new generated code

gracewilcox (approved):

sdk/security/keyvault/azadmin/backup/client.go
gracewilcox: I can regenerate and update it if need be

sdk/security/keyvault/azadmin/backup/models.go
jhendrixMSFT: When I rebased my changes, I didn't regenerate the code, I just accepted what was in HEAD which was built with the older code generator.

gracewilcox: Makes sense!

sdk/security/keyvault/azadmin/backup/client.go
jhendrixMSFT: Looking...

jhendrixMSFT: Good catch! We just need to propagate the client's `tracing.Tracer` to the `runtime.Poller[T]`, no need for the custom polling handler to have it.

Issue #21887, opened 2023-11-04
[azblob] Blob metadata doesn't load metadata keys with nil values


richardpark-msft:
### Bug Report

**Package and version**: `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.2.0` (I don't think this is a newer/older package issue)

**Details**

Blobs let you store metadata with a blob, independent of the content. These are just key/value pairs, where the values are `*string`s. It is possible to set the value of a blob property to null, and then retrieve that value as well. In the BlobItem type it looks like this:

```go
// BlobItem - An Azure Storage blob
type BlobItem struct {
  // Dictionary of
  Metadata map[string]*string `xml:"Metadata"`
}
```

However, our blob package doesn't seem to be able to deserialize this. Instead of getting back a map with an `ownerid` key and a `nil` value it returns an empty map altogether. 

You can use code like this to see the bug:

```go
pager := cc.NewListBlobsFlatPager(&container.ListBlobsFlatOptions{
  Include: container.ListBlobsInclude{
    Metadata: true,
  },
})

for pager.More() {
  page, err := pager.NextPage(context.Background())

  for _, blob := range page.Segment.BlobItems {
    // blob.Metadata will just be a nil - it won't contain an `ownerid` key.
    fmt.Printf("%s, %#v\n", *blob.Name, blob.Metadata)
  }
}
```

In the XML response from Azure Storage you can see they do send us back enough information to construct the map:

```xml
// a whole bunch of other XML elided
  <Metadata>
      <ownerid />
  </Metadata>
```

jhendrixMSFT:
Is creating metadata without a value a supported scenario? Per the [REST docs](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-metadata?tabs=microsoft-entra-id#request-headers) it doesn't imply that it is although it's a big vague. In addition, the portal won't let you create metadata without a value.

Could it be that we have a bug that's allowing creation of metadata with a `nil` value? But then you'd think the service would reject it.

jhendrixMSFT:
Metadata with `nil` values will be skipped, see the example [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/generated/zz_blob_client.go#L2429-L2435).

jhendrixMSFT:
Ok this isn't about `nil` values but empty values.
```go
_, err = client.SetMetadata(context.Background(), map[string]*string{
[...]

PR #21967, opened 2023-11-13
Include metadata keys with empty values

jhendrixMSFT:
Requires updating to the latest code generator.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/21887


sdk/storage/azblob/container/client_test.go
```diff
@@ -1415,6 +1415,68 @@ func (s *ContainerRecordedTestsSuite) TestListBlobIncludeMetadata() {
...
+	})
+
+	for pager.More() {
+		resp, err := pager.NextPage(context.Background())
+		_require.NoError(err)
+
+		_require.Len(resp.ListBlobsFlatSegmentResponse.Segment.BlobItems, 6)
+		for _, blob := range resp.ListBlobsFlatSegmentResponse.Segment.BlobItems {
+			_require.NotNil(blob.Metadata)
+			_require.Len(blob.Metadata, len(metadata))
```
souravgupta-msft, line 1453: Should we also add a case where we compare the metadata returned from GetProperties call?

jhendrixMSFT: For `GetProperties` the metadata is returned via HTTP headers which wasn't impacted by this change.

PR #21986, opened 2023-11-15
Add tracing spans to azquery

jhendrixMSFT:

### DIFF
--- a/sdk/monitor/azquery/autorest.md
+++ b/sdk/monitor/azquery/autorest.md
@@ -13,6 +13,7 @@ openapi-type: "data-plane"
 output-folder: ../azquery
 security: "AADToken"
 use: "@autorest/go@4.0.0-preview.60"
+inject-spans: true
 version: "^3.0.0"
 
 directive:
--- a/sdk/monitor/azquery/autorest.md
+++ b/sdk/monitor/azquery/autorest.md
@@ -3,7 +3,6 @@
 ``` yaml
 title: Logs Query Client
 clear-output-folder: false
-export-clients: true
 go: true
 input-file: 
     - https://github.com/Azure/azure-rest-api-specs/blob/72427ef3ff5875bd8409ef112ef5e6f3cf2b8795/specification/operationalinsights/data-plane/Microsoft.OperationalInsights/stable/2022-10-27/OperationalInsights.json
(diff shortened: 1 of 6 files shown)


sdk/monitor/azquery/autorest.md
```diff
@@ -13,6 +13,7 @@ openapi-type: "data-plane"
 output-folder: ../azquery
 security: "AADToken"
 use: "@autorest/go@4.0.0-preview.60"
+inject-spans: true
```
gracewilcox, line 16: Is this a new autorest command for the newest code generator version?

sdk/monitor/azquery/custom_client.go
```diff
@@ -46,7 +46,7 @@ func NewLogsClient(credential azcore.TokenCredential, options *LogsClientOptions
 	}
 
 	authPolicy := runtime.NewBearerTokenPolicy(credential, []string{c.Audience + "/.default"}, nil)
-	azcoreClient, err := azcore.NewClient("azquery.LogsClient", version, runtime.PipelineOptions{PerRetry: []policy.Policy{authPolicy}}, &options.ClientOptions)
+	azcoreClient, err := azcore.NewClient(moduleName, version, runtime.PipelineOptions{PerRetry: []policy.Policy{authPolicy}}, &options.ClientOptions)
```
gracewilcox, line 49: Is there a way to differentiate between the two clients when tracing? Know the documentation you sent earlier said it should be by package name, but I'm still curious.

gracewilcox (approved):

sdk/monitor/azquery/autorest.md
gracewilcox: Noticed we didn't add this when adding spans to keyvault.

jhendrixMSFT: Yeah, it will inject span creation in the generated code. It's on by default for ARM, and we should also enable it for data-plane packages that directly expose the generated code.

Same setting is also enabled for the key vault packages (for `azkeys` it was enabled in the previous beta PR).

sdk/monitor/azquery/custom_client.go
jhendrixMSFT: The spans will have names like `Client.Method`, with the `otel.library.name` set to the value of `moduleName` so one can see exactly where the call originated.

sdk/monitor/azquery/autorest.md
gracewilcox: Thanks! Makes sense, see that now.

sdk/monitor/azquery/custom_client.go
gracewilcox: That makes sense now! Thank you!

PR #21987, opened 2023-11-15
Add tracing spans to azingest

jhendrixMSFT:


sdk/monitor/azingest/autorest.md
```diff
@@ -6,7 +6,6 @@ description: Azure Monitor Ingestion Go Client
 generated-metadata: false
 
 clear-output-folder: false
-export-clients: true
```
gracewilcox, line 9: Any particular reason to remove this line? Noticed it was left in the other modules.

sdk/monitor/azingest/autorest.md
```diff
@@ -57,4 +58,10 @@ directive:
   - from: swagger-document
     where: $.paths./dataCollectionRules/{ruleId}/streams/{stream}.post
     transform: $["description"] = "Ingestion API used to directly ingest data using Data Collection Rules. Maximum size of of API call is 1 MB."
-```
\ No newline at end of file
+
+  # fix up body param to keep back-compat
+  - from: swagger-document
```
gracewilcox, line 63: What does this transform do? Noticed that without it the parameter will generate as type `[][]byte` instead of `[]byte`. Noticed this also happening in `azquery Table.Rows` with the newer code gen version (had to add an autorest transform). Is this supposed to be happening?

sdk/monitor/azingest/build.go
```diff
@@ -3,6 +3,7 @@
 
 //go:generate autorest ./autorest.md
 //go:generate rm ./models_serde.go
+//go:generate rm ./models.go
 //go:generate rm ./constants.go
```
gracewilcox, line 7:
```suggestion
```
Need to remove this line. Tried generating the code locally and received this error:
`rm: cannot remove './constants.go': No such file or directory`

sdk/monitor/azingest/autorest.md
jhendrixMSFT: It's a vestigial setting from older versions of the code generator. Doesn't hurt anything but doesn't do anything either. We should remove it everywhere at some point.

jhendrixMSFT: I updated the PR for `azquery` to remove it there too.

sdk/monitor/azingest/build.go
jhendrixMSFT: Removed.

sdk/monitor/azingest/autorest.md
jhendrixMSFT: The older code generator was accidentally treating arrays of JSON objects when `rawjson-as-bytes:true` is set as `[]byte` instead of `[][]byte`. It's a bug or a feature depending on your perspective. I assume that we wanted to keep this as `[]byte` so the transform removes the "arrayness" of the body param.

gracewilcox (approved):
Thanks for adding this to Monitor!

Issue #21988, opened 2023-11-15
Investigate issuing new certs for testing SNI authentication


chlowell:
Ours have expired, causing live tests to fail with an error like `"AADSTS1000502: The provided certificate is not within its specified validity window`. Issuing new SNI certs is nontrivial. In the meantime we can just forego live testing SNI auth and eSTS-R (which requires it) and continue using the recordings to maintain coverage.

PR #21989, opened 2023-11-15
Skip live SNI authentication tests

chlowell:
Our SNI cert has expired and we may not be able to get a new one. Pending a final determination on that point (#21988), let's skip these live tests. We don't really need them and can continue using the recordings for coverage.

### DIFF
--- a/sdk/azidentity/client_certificate_credential_test.go
+++ b/sdk/azidentity/client_certificate_credential_test.go
@@ -192,6 +192,9 @@ func TestClientCertificateCredential_Live(t *testing.T) {
 			if test.path == "" {
 				t.Skip("no certificate file specified")
 			}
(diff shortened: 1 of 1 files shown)


sdk/azidentity/client_certificate_credential_test.go
```diff
@@ -90,27 +90,15 @@ func TestClientCertificateCredential_GetTokenSuccess(t *testing.T) {
 	}
 }
 
-func TestClientCertificateCredential_GetTokenSuccess_withCertificateChain(t *testing.T) {
```
jhendrixMSFT, line 93: Was this a duplicate test?

chlowell: Yes. The one below it tests the same path and, with the update in this PR, is more comprehensive

jhendrixMSFT (approved):

PR #21990, opened 2023-11-16
Adding cross partition query implementation

afrancoc2000:
This PR allows the execution of cross partition queries using the CosmosDB REST API, allowing developers using the go SDK to run simple projections and filters, even though more complex queries are restricted by Gateway like TOP, ORDER  BY, OFFSET LIMIT, Aggregates, DISTINCT, and GROUP BY.

### DIFF
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -457,6 +457,67 @@ func (c *ContainerClient) NewQueryItemsPager(query string, partitionKey Partitio
 	})
 }
 
+// NewCrossPartitionQueryItemsPager executes a cross partition query in a Cosmos container.
+// This method uses the Azure Cosmos DB REST API, and is subject to the API limitations:
+// Any query that requires state across continuations cannot be served by the gateway. This includes:
+// TOP, ORDER BY, OFFSET LIMIT, Aggregates (Min, Max, Count, etc.), DISTINCT, GROUP BY.
+// Queries that can be served by the gateway include: Simple projections, Filters
+// More information can be found in the next link:
+// https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway
+// query - The SQL query to execute.
+// o - Options for the operation.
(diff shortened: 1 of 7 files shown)


afrancoc2000:
@microsoft-github-policy-service agree company="Microsoft"

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -457,6 +457,67 @@ func (c *ContainerClient) NewQueryItemsPager(query string, partitionKey Partitio
...
+// NewCrossPartitionQueryItemsPager executes a cross partition query in a Cosmos container.
+// This method uses the Azure Cosmos DB REST API, and is subject to the API limitations:
+// Any query that requires state across continuations cannot be served by the gateway. This includes:
+// TOP, ORDER BY, OFFSET LIMIT, Aggregates (Min, Max, Count, etc.), DISTINCT, GROUP BY.
+// Queries that can be served by the gateway include: Simple projections, Filters
+// More information can be found in the next link:
+// https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway
+// query - The SQL query to execute.
+// o - Options for the operation.
+func (c *ContainerClient) NewCrossPartitionQueryItemsPager(query string, o *QueryOptions) *runtime.Pager[QueryItemsResponse] {
```
ealsur, line 469: Thanks for the PR! However this is not really adding cross partition query support, it is just adding a method that sets the cross-partition header, which will not work in a lot of cases (as your comment states).

The problem is there is no way to enforce this and customers will not read this comment, and attempt to use this method in the same way they use the other SDKs that fully support cross-partition and end up with invalid results (worst case scenario) or an error they didn't expect. Which leads to Issue reporting.

@jcocchi what are your thoughts?

jcocchi: @afrancoc2000 thank you for your contribution! As Matias mentioned there are many aspects to cross partition query and we want to wait until we have a more complete story before adding this support directly into the SDK. For now, please use the workaround, and we can reconsider adding cross partition query support directly after this package reaches GA.

afrancoc2000:
Thank you @jcocchi and @ealsur for your feedback, I used the workaround and it worked, I'll close the PR then.

PR #21998, opened 2023-11-17
[azeventgrid] Updating to the latest swagger. Changes doc'd in the CHANGELOG

richardpark-msft:


sdk/messaging/azeventgrid/autorest.md
```diff
@@ -20,34 +18,110 @@ use: "@autorest/go@4.0.0-preview.52"
...
-  - from: client.go
+  - from: 
+    - client.go
     where: $
-    transform: return $.replace(/PublishCloudEvents\(/g, "internalPublishCloudEvents(");
-  - from: swagger-document
+    transform: | 
+      return $.replace(
+        /(func \(client \*Client\) publishCloudEventsCreateRequest.+?)return req, nil/s, 
+        '$1\nreq.Raw().Header.Set("Content-type", "application/cloudevents-batch+json; charset=utf-8")\nreturn req, nil');
```
jhendrixMSFT, line 33: Do you know if this will be added to the swagger at some point?

sdk/messaging/azeventgrid/autorest.md
```diff
@@ -20,34 +18,110 @@ use: "@autorest/go@4.0.0-preview.52"
...
     where: $.definitions.CloudEvent.properties.specversion
-    transform: $["x-ms-client-name"] = "SpecVersion"
-  - from: swagger-document
-    where: $.definitions.CloudEvent.properties.datacontenttype
-    transform: $["x-ms-client-name"] = "DataContentType"
-  - from: swagger-document
-    where: $.definitions.CloudEvent.properties.dataschema
-    transform: $["x-ms-client-name"] = "DataSchema"
-  # mark models as external so they're just omitted
+    transform: $.replace(/\s+\/\/ Anything\s+Interface any/sg, "$1");
```
jhendrixMSFT, line 86: Does this actually work? The `where` clause is for swagger but the `from` is Go source code.

sdk/messaging/azeventgrid/autorest.md
richardpark-msft: It's on my TODO list. I need to file some issues - there were a few more that seemed like "do I need to patch this long term?".

sdk/messaging/azeventgrid/autorest.md
richardpark-msft: It does but apparently `where` is just ignored for source transforms. Or...it's used but doesn't affect the transform itself.

I cleaned it up :)

sdk/messaging/azeventgrid/autorest.md
richardpark-msft: Filed as https://github.com/Azure/azure-sdk-for-go/issues/22007

sdk/messaging/azeventgrid/client.go
```diff
@@ -324,3 +383,65 @@ func (client *Client) releaseCloudEventsHandleResponse(resp *http.Response) (Rel
...
+// is successfully accepted. The response body will include the set of successfully renewed
+// lockTokens, along with other failed lockTokens with their corresponding error information.
+// If the operation fails it returns an *azcore.ResponseError type.
+//
+// Generated from API version 2023-10-01-preview
+//   - topicName - Topic Name.
+//   - eventSubscriptionName - Event Subscription Name.
+//   - renewLockOptions - RenewLockOptions
+//   - options - RenewCloudEventLocksOptions contains the optional parameters for the Client.RenewCloudEventLocks method.
+func (client *Client) RenewCloudEventLocks(ctx context.Context, topicName string, eventSubscriptionName string, renewLockOptions RenewLockOptions, options *RenewCloudEventLocksOptions) (RenewCloudEventLocksResponse, error) {
```
jhendrixMSFT, line 397: This looks a little weird, the two options types. Is `RenewLockOptions` actually optional?

richardpark-msft: This is all over their API at the moment, which is unfortunate.

What is our naming convention for "group of non-optional parameters"?

richardpark-msft: Body?

jhendrixMSFT: We don't have a standard prefix/suffix for such types as we should try to come up with clear, domain-specific names, thus will be on a case-by-case basis. Maybe something like `LockRenewal`, `LockRenewalValues`?

CC @JeffreyRichter

jhendrixMSFT: One could also make the argument that the slice of lock tokens should be a discrete parameter.

richardpark-msft: > One could also make the argument that the slice of lock tokens should be a discrete parameter.

I'd be a bit worried that expansion of that 'args' bag will mean I'll need to start creating multiple functions. So we should probably keep it in a bag for now, even if we need to come up with a name.

richardpark-msft: I'm trying to get this out so we can properly deserialize all the previous data since that recently broke. 

I'll create an APIView, we can hash it out there and I'll fix it up before the next beta goes out.

jhendrixMSFT: > creating multiple functions

Can you expand on that? Will the bag contain a mix of optional and/or mutually exclusive values?

richardpark-msft: https://apiview.dev/Assemblies/Review/05f3306f2ec84f1bbdd581445bf015d6?diffRevisionId=56aa37b5b45a48e8a8b6c0dc3f0596a4&diffOnly=False&revisionId=c2c10b9a596d46e39a6df83dcde04346&doc=False

richardpark-msft: > > creating multiple functions
> 
> Can you expand on that? Will the bag contain a mix of optional and/or mutually exclusive values?

I don't think it'll be a mix - ie, optional should be in Options, required args in "args/body/params". 

What I mean is that if they add even one more required parameter then we'd have to create multiple functions in the future. I'd rather not head down the path of assuming it's always going to be just the one required arg.

richardpark-msft: Thinking about this some more...they can't add another required parameter. I'll expand this out for now.

jhendrixMSFT: Well, if the body ends up requiring additional content, then that would be a breaking change. However, conceptually, I think that's desired as we're now saying that the contract for this method has changed in a meaningful way.

richardpark-msft: (talked offline)

We'll talk about this some more. In this case (and the others) we can just make the one field in the bag a required parameter, which makes the entire thing nicer anyways. 

But it'd be good to just head these kinds of conversations off at the pass, pick a consistent suffix (so customers can just expect it, like we have with our normal `Options` bags for functions). We can have that discussion in the APIView.

sdk/messaging/azeventgrid/models.go
```diff
@@ -64,34 +88,50 @@ type ReceiveResult struct {
...
+	// REQUIRED; Array of FailedLockToken for failed cloud events. Each FailedLockToken includes the lock token along with the
+	// related error information (namely, the error code and description).
+	FailedLockTokens []FailedLockToken
+
+	// REQUIRED; Array of lock tokens for the successfully renewed locks.
+	SucceededLockTokens []string
+}
+
+// RenewLockOptions - Array of lock tokens for the corresponding received Cloud Events to be renewed.
+type RenewLockOptions struct {
```
jhendrixMSFT, line 134: Will this one also get the flattening?

richardpark-msft: Yes, done.

jhendrixMSFT (approved):

PR #22024, opened 2023-11-21
Add `ContentDisposition` and `ContentType` to blob.GetSASURLOptions

yp05327:
- Add `ContentDisposition` and `ContentType` to blob.GetSASURLOptions
In old version, users can set `Content-Disposition` header in `GetSASURLOptions` for customizing the filename.
https://github.com/Azure/azure-sdk-for-go/blob/2187aca40bbf6931ffac666977e5d29128d82585/storage/blob.go#L86-L109
But this is not supported in the latest version now.

- About the test
In #19945, all tests for `GetSASURLOptions` were removed.
I added a test for `blob.GetSASURLOptions`, maybe this is unnecessary? If it is unnecessary, I will remove it.

### DIFF
--- a/sdk/storage/azblob/CHANGELOG.md
+++ b/sdk/storage/azblob/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Other Changes
 
 * Updated to the latest version of `azcore`.
+* Add `ContentDisposition` and `ContentType` to blob.GetSASURLOptions
 
 ## 1.2.0 (2023-10-11)
 
--- a/sdk/storage/azblob/blob/client.go
+++ b/sdk/storage/azblob/blob/client.go
@@ -304,13 +304,14 @@ func (b *Client) GetSASURL(permissions sas.BlobPermissions, expiry time.Time, o 
 	st := o.format()
 
 	qps, err := sas.BlobSignatureValues{
-		ContainerName: urlParts.ContainerName,
-		BlobName:      urlParts.BlobName,
-		SnapshotTime:  t,
-		Version:       sas.Version,
-		Permissions:   permissions.String(),
-		StartTime:     st,
-		ExpiryTime:    expiry.UTC(),
+		ContainerName:      urlParts.ContainerName,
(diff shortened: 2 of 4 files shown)


yp05327:

yp05327:
It seems that the CI fail is not related.
https://dev.azure.com/azure-sdk/public/_build/results?buildId=3278778&view=logs&j=3f04450f-291c-5eb3-bb7b-886aac67999b&t=df455e12-7dc7-50e0-bae4-4c8ec5aa19c5&l=33247

souravgupta-msft:
@yp05327, thanks for creating this PR. I want to mention that the `blob.GetSASURL()` method is for general use case. There are several other fields in [`BlobSignatureValues`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/sas/service.go#L22-L44) other than `ContentDisposition` and `ContentType`, that are not exposed in the [`blob.GetSASURLOptions`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blob/models.go#L500-L502).

You can refer [this](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/appendblob/client_test.go#L452-L461) code for creating blob SAS using these fields.

yp05327:
@souravgupta-msft, thanks for your comment. Do you mean that if users want to set values of fields in `BlobSignatureValues` other than `StartTime` and `ExpiryTime`, they need to implement a similar `GetSASURL` function by themselves?
And are there any reasons that not all fields in `BlobSignatureValues` can be exposed as `GetSASURLOptions`?

vibhansa-msft:
GetSASURL is a generic method built for most common use cases. If an expert user wishes to create a very specific request, it can be created by directly created using other methods that exposes all available options (which is internally used by GetSASUrL itself). Just exposing 2 options out of a bunch is not good.

yp05327:
Ok, I will close this PR.

PR #22081, opened 2023-12-01
[azopenai] Updating to the latest 2023-12-01 API

richardpark-msft:
Lots of changes, making areas that were somewhat weakly typed to be strongly typed:

- Chat message types for requests and responses are now different, with fields that are relevant to each more clearly called out. For instance, fields relevant to only assistant type messages (like tools) are separated properly.
- Strong types for OYD (ie, bring your own data) scenarios, instead of a `any` parameter.
- General updates to just expose new features for models like Dall-E

Also, some general fixes and new tests for things like gpt-4-vision.


sdk/ai/azopenai/custom_client.go
```diff
@@ -277,25 +271,66 @@ type clientData struct {
...
-	return body.AzureExtensionsOptions != nil && len(body.AzureExtensionsOptions.Extensions) > 0
+	return body.AzureExtensionsOptions != nil && len(body.AzureExtensionsOptions) > 0
+}
+
+// ChatRequestUserMessageContent contains the user prompt - either as a single string
+// or as a []ChatCompletionRequestMessageContentPart, enabling images and text as input.
+//
+// NOTE: This should be created using [azopenai.NewChatRequestUserMessageContent]
+type ChatRequestUserMessageContent struct {
+	str   *string
```
jhendrixMSFT, line 312: Since these are just inputs to `json.Marshal` couldn't they be combined `any` field? Type info doesn't seem to be needed at this point.

sdk/ai/azopenai/autorest.md
```diff
@@ -22,130 +21,219 @@ slice-elements-byval: true
...
   # This must be done in each operation as the parameter is not defined in the components section
-  - from: openapi-document
+  - from: swagger-document
     where: $.paths..parameters..[?(@.name=='deploymentId')]
     transform: $["x-ms-parameter-location"] = "client";
+```
+
+## Polymorphic adjustments
+
+The polymorphic _input_ models all expose the discriminator but it's ignored when serializing
```
jhendrixMSFT, line 42: What's the motivation for this, to just reduce the public surface area? At any rate, we have to be sure that for "root" and "sub-root" discriminated types that we export the discriminator. This is to handle cases where a type isn't yet modeled in the SDK but is supported by the service.

richardpark-msft: Ah, okay. I'll bring them back, didn't think about the "expandable future enum" part.

sdk/ai/azopenai/custom_client.go
jhendrixMSFT: Can be left as-is for now. If in the future more fields are added then this will need to be refactored.

richardpark-msft: No, no, your idea makes sense. I just misunderstood it.

sdk/ai/azopenai/autorest.md
richardpark-msft: It's back by popular demand!

jhendrixMSFT: You do have a good point though that for the concrete types, exposing the discriminator field seems unnecessary (for inputs) as we automatically populate the value and ignore any user-specified value. On the surface, it also seems unnecessary for output types as well. We could consider making the configurable in code generation, unless we can come up with a reason why it would be bad.

jhendrixMSFT: But we should not remove it for the "root" type(s) to support the "expandable future enum" thing mentioned earlier.

jhendrixMSFT (approved):

Issue #18983, opened 2022-08-30
Cosmos DB: Add GlobalEndpointManager


ealsur:
Add internal GlobalEndpointManager to handle:

Requires https://github.com/Azure/azure-sdk-for-go/issues/18982

* Periodic account information refresh (5M)
* Marking regions as unavailable

Reference: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Routing/GlobalEndpointManager.cs

jay-most:
3 - 4 weeks - > T-Shirt  L

ealsur:
Design content:



![Image](https://github.com/Azure/azure-sdk-for-go/assets/1633401/f2a734f1-1679-497f-b605-d5247fc514bd)


LocationCache is already present (done in https://github.com/Azure/azure-sdk-for-go/issues/18982) -> https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/cosmos_location_cache.go

GEM just needs to wrap the LocationCache Marking of unavailable and provide background refresh capabilities that wire the Account Properties (regions + MultiMaster) into the locationCache

PR #22100, opened 2023-12-05
[Cosmos] Add Global Endpoint Manager

simorenoh:
Adds the global endpoint manager to the SDK. For now it's just the basic functionality, exposing the location cache methods to the GEM and adding the API request to fetch the database account information. Issue: https://github.com/Azure/azure-sdk-for-go/issues/18983

Next steps would be to develop the policy that uses the global endpoint manager to refresh account information every 5 minutes, as well as making changes to the first call to fetch the database account information in order to include retries to any preferred regions set.

Closes https://github.com/Azure/azure-sdk-for-go/issues/18983

### DIFF
--- a/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
+++ b/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
 
 	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
@@ -20,23 +21,27 @@ type globalEndpointManager struct {
 	preferredLocations  []string
 	locationCache       *locationCache
 	refreshTimeInterval time.Duration
+	gemMutex            sync.Mutex
+	lastUpdateTime      time.Time
 }
 
-func newGlobalEndpointManager(client *Client, preferredLocations []string, refreshTimeInterval time.Duration) (globalEndpointManager, error) {
+func newGlobalEndpointManager(client *Client, preferredLocations []string, refreshTimeInterval time.Duration) (*globalEndpointManager, error) {
 	endpoint, err := url.Parse(client.endpoint)
 	if err != nil {
-		return globalEndpointManager{}, err
+		return &globalEndpointManager{}, err
 	}
 
 	if refreshTimeInterval == 0 {
 		refreshTimeInterval = defaultUnavailableLocationRefreshInterval
 	}
 
-	gem := globalEndpointManager{
+	gem := &globalEndpointManager{
 		client:              client,
 		preferredLocations:  preferredLocations,
 		locationCache:       newLocationCache(preferredLocations, *endpoint),
-		refreshTimeInterval: refreshTimeInterval}
+		refreshTimeInterval: refreshTimeInterval,
+		lastUpdateTime:      time.Time{},
+	}
 
 	return gem, nil
 }
@@ -73,12 +78,20 @@ func (gem *globalEndpointManager) RefreshStaleEndpoints() {
 	gem.locationCache.refreshStaleEndpoints()
 }
 
+func (gem *globalEndpointManager) ShouldRefresh() bool {
+	return time.Since(gem.lastUpdateTime) > gem.refreshTimeInterval
+}
+
 func (gem *globalEndpointManager) Update(ctx context.Context) error {
+	gem.gemMutex.Lock()
+	defer gem.gemMutex.Unlock()
+	if !gem.ShouldRefresh() {
+		return nil
+	}
 	accountProperties, err := gem.GetAccountProperties(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to retrieve account properties: %v", err)
 	}
-
 	err = gem.locationCache.update(
 		accountProperties.WriteRegions,
 		accountProperties.ReadRegions,
@@ -87,7 +100,7 @@ func (gem *globalEndpointManager) Update(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to update location cache: %v", err)
 	}
-
+	gem.lastUpdateTime = time.Now()
 	return nil
 }
 
--- a/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
+++ b/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
@@ -73,8 +73,8 @@ func (gem *globalEndpointManager) RefreshStaleEndpoints() {
 	gem.locationCache.refreshStaleEndpoints()
 }
 
-func (gem *globalEndpointManager) Update() error {
-	accountProperties, err := gem.GetAccountProperties()
+func (gem *globalEndpointManager) Update(ctx context.Context) error {
+	accountProperties, err := gem.GetAccountProperties(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to retrieve account properties: %v", err)
 	}
@@ -91,7 +91,7 @@ func (gem *globalEndpointManager) Update() error {
 	return nil
 }
 
-func (gem *globalEndpointManager) GetAccountProperties() (accountProperties, error) {
+func (gem *globalEndpointManager) GetAccountProperties(ctx context.Context) (accountProperties, error) {
 	operationContext := pipelineRequestOptions{
 		resourceType:    resourceTypeDatabaseAccount,
 		resourceAddress: "",
@@ -102,7 +102,7 @@ func (gem *globalEndpointManager) GetAccountProperties() (accountProperties, err
 		return accountProperties{}, fmt.Errorf("failed to generate path for name-based request: %v", err)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 60 * time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
 	azResponse, err := gem.client.sendGetRequest(path, ctx, operationContext, nil, nil)
 	cancel()
 	if err != nil {
--- a/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
+++ b/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
@@ -1,0 +1,128 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+const defaultUnavailableLocationRefreshInterval = 5 * time.Minute
+
+type globalEndpointManager struct {
+	client              *Client
+	preferredLocations  []string
+	locationCache       *locationCache
+	refreshTimeInterval time.Duration
+}
+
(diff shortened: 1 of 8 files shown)


sdk/data/azcosmos/shared_key_credential.go
```diff
@@ -69,22 +69,24 @@ func (c *KeyCredential) buildCanonicalizedAuthHeaderFromRequest(req *policy.Requ
 			resourceAddress = strings.ToLower(resourceAddress)
 		}
 
-		value = c.buildCanonicalizedAuthHeader(req.Raw().Method, resourceTypePath, resourceAddress, req.Raw().Header.Get(headerXmsDate), "master", "1.0")
+		var isDatabaseAccount bool = opValues.resourceType == resourceTypeDatabaseAccount
```
ealsur, line 72:
```suggestion
		isDatabaseAccount := opValues.resourceType == resourceTypeDatabaseAccount
```

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -68,6 +68,8 @@ func newLocationCache(prefLocations []string, defaultEndpoint url.URL) *location
 }
 
 func (lc *locationCache) update(writeLocations []accountRegion, readLocations []accountRegion, prefList []string, enableMultipleWriteLocations *bool) error {
+	lc.mapMutex.Lock()
```
ealsur, line 71: Why is the lock here needed?

ealsur: And not on the refreshStaleEndpoints?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,128 @@
...
+func (gem *globalEndpointManager) IsEndpointUnavailable(endpoint url.URL, ops requestedOperations) bool {
+	return gem.locationCache.isEndpointUnavailable(endpoint, ops)
+}
+
+func (gem *globalEndpointManager) RefreshStaleEndpoints() {
+	gem.locationCache.refreshStaleEndpoints()
+}
+
+func (gem *globalEndpointManager) Update() error {
+	accountProperties, err := gem.GetAccountProperties()
```
ealsur, line 77: The concurrency needs to be guarded here. Calling this method executes an HTTP request, right? If 2 calls to `Update` happen concurrently, then both will do the HTTP requests.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (gem *globalEndpointManager) IsEndpointUnavailable(endpoint url.URL, ops requestedOperations) bool {
+	return gem.locationCache.isEndpointUnavailable(endpoint, ops)
+}
+
+func (gem *globalEndpointManager) RefreshStaleEndpoints() {
+	gem.locationCache.refreshStaleEndpoints()
+}
+
+func (gem *globalEndpointManager) Update() error {
```
ealsur, line 76: Flow a context in:

```suggestion
func (gem *globalEndpointManager) Update(ctx context.Context) error {
```

Then pass it to the `GetAccountProperties()` instead of using `context.Background()` inside

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -0,0 +1,128 @@
...
+		resourceType:    resourceTypeDatabaseAccount,
+		resourceAddress: "",
+	}
+
+	path, err := generatePathForNameBased(resourceTypeDatabaseAccount, "", false)
+	if err != nil {
+		return accountProperties{}, fmt.Errorf("failed to generate path for name-based request: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
ealsur, line 105: Timeout should be 60 seconds for Gateway calls

```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 60 * time.Second)
```

In the future we can follow other SDKs and have a timeout policy that does multiple requests with different timeout values to handle Gateway upgrade scenarios

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,407 @@
...
+	assert.NoError(t, err)
+
+	// Assert the expected read endpoints
+	expectedReadEndpoints := []url.URL{
+		*serverEndpoint,
+	}
+	assert.Equal(t, expectedReadEndpoints, readEndpoints)
+}
+
+// Emulator Test
```
ealsur, line 346: This comment should be removed?

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,407 @@
...
+	// Multiple locations should be false for default GEM
+	canUseMultipleWriteLocs := gem.CanUseMultipleWriteLocations()
+	assert.False(t, canUseMultipleWriteLocs)
+
+	// Mock GEM with multiple write locations available should show true
+	canUseMultipleWriteLocs = mockGem.CanUseMultipleWriteLocations()
+	assert.True(t, canUseMultipleWriteLocs)
+}
+
+func TestGlobalEndpointManagerUpdate(t *testing.T) { //This test should be testing for the lock on the update/ refresh mechanism
```
ealsur, line 280: How is this test checking the lock? Are we issuing multiple concurrent Update calls and measuring how many requests the pipeline did?

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,407 @@
...
+	assert.Equal(t, expectedWriteEndpoints, writeEndpoints)
+}
+
+func TestGlobalEndpointManagerGetReadEndpoints(t *testing.T) {
+	headerPolicy := &headerPolicies{}
+	srv, close := mock.NewTLSServer()
+	defer close()
+	srv.SetResponse(mock.WithStatusCode(http.StatusOK))
+
+	verifier := headerPoliciesVerify{}
```
ealsur, line 60: This verifier is something used to verify that certain headers were set, it is not used in any of the tests in this file, can be removed?

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,407 @@
...
+		EnableMultipleWriteLocations: false,
+	}
+
+	jsonString, err := json.Marshal(properties)
+	if err != nil {
+		t.Fatal(err)
+	}
+	srv.SetResponse(mock.WithBody(jsonString))
+
+	verifier := headerPoliciesVerify{}
```
ealsur, line 304: one way to verify the number of HTTP calls is add a new policy that just counts the number of requests, and if you do multiple concurrent Update calls, the number of HTTP calls should be 1?

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
simorenoh: this was the reason why I had changed the lock lol - seems I put it in the wrong place

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
simorenoh: was just using it to mark the difference between the tests but I'll remove it

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
simorenoh: I left the note to change the test and forgot to do it 😢

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
simorenoh: will keep this unresolved as I work on the update tests

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
ealsur: Emulator tests should be in a separate file that starts with `emulator_......`

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,307 @@
...
+	assert.NoError(t, err)
+
+	req.SetOperationValue(pipelineRequestOptions{
+		isWriteOperation: true,
+	})
+
+	client := &Client{endpoint: srv.URL(), pipeline: pl}
+	fmt.Println(client.endpoint)
+
+	gem, err := newGlobalEndpointManager(client, []string{}, 5*time.Second)
```
ealsur, line 267: nit: Reduce this time, to like 10 seconds.

Do the concurrent work. Assert 1 call.

Wait 10 seconds.

Do a single call. Assert 2 calls.

That should cover that after the time passes, we actually allow another call to happen

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	jsonString, err := json.Marshal(properties)
+	if err != nil {
+		t.Fatal(err)
+	}
+	srv.SetResponse(mock.WithBody(jsonString))
+
+	pl := azruntime.NewPipeline("azcosmostest", "v1.0.0", azruntime.PipelineOptions{PerCall: []policy.Policy{countPolicy}}, &policy.ClientOptions{Transport: srv})
+	req, err := azruntime.NewRequest(context.Background(), http.MethodGet, srv.URL())
```
ealsur, line 258: this `req` variable is not used

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+	wg.Add(concurrency)
+
+	finishCh := make(chan struct{})
+
+	for i := 0; i < concurrency; i++ {
+		go func(wg *sync.WaitGroup) {
+			defer wg.Done()
+			// Call the function in each goroutine
+			err := gem.Update(context.Background())
+			if err != nil {
```
ealsur, line 282:
```suggestion
			assert.NoError(t, err)
```

ealsur: You can remove the other parts of the `if`

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	gem, err := newGlobalEndpointManager(client, []string{}, 5*time.Second)
+	assert.NoError(t, err)
+
+	// Call update concurrently and see how many times the policy gets called
+	concurrency := 5
+	wg := &sync.WaitGroup{}
+	wg.Add(concurrency)
+
+	finishCh := make(chan struct{})
```
ealsur, line 275: `finishCh` is not used for the concurrency, right?

ealsur (changes requested):
Waiting on the cleanup + test addition, great work!

sdk/data/azcosmos/cosmos_global_endpoint_manager_test.go
simorenoh: yeah it was leftover from some testing I was doing, will clean it up

Issue #22120, opened 2023-12-08
Retry policy should support retry-after-ms


jhendrixMSFT:
Provides sub-second retry granularity.

trrwilson:
Thanks, @jhendrixMSFT ! Parity link dump for easy reference:

- .NET's retry handling: [azure-sdk-for-net/sdk/core/Azure.Core/src/ResponseHeaders.cs at 7e7e9854d4c90fc25ee65d7729d69a8489f00d59 · Azure/azure-sdk-for-net (github.com)](https://github.com/Azure/azure-sdk-for-net/blob/7e7e9854d4c90fc25ee65d7729d69a8489f00d59/sdk/core/Azure.Core/src/ResponseHeaders.cs#L81)
- Java's retry handling: [azure-sdk-for-java/sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtils.java at 3594e48f1b837f2c9656337a5488e867a7aa465d · Azure/azure-sdk-for-java (github.com)](https://github.com/Azure/azure-sdk-for-java/blob/3594e48f1b837f2c9656337a5488e867a7aa465d/sdk/core/azure-core/src/main/java/com/azure/core/implementation/ImplUtils.java#L77)
- JS's retry handling: [azure-sdk-for-js/sdk/core/core-rest-pipeline/src/retryStrategies/throttlingRetryStrategy.ts at 17de1a2b7f3ad61f34ff62876eced7d077c10d4b · Azure/azure-sdk-for-js (github.com)](https://github.com/Azure/azure-sdk-for-js/blob/17de1a2b7f3ad61f34ff62876eced7d077c10d4b/sdk/core/core-rest-pipeline/src/retryStrategies/throttlingRetryStrategy.ts#L31)

Each of those, in the language-specific implementations, handles the three retry headers in precedence order, most succinctly in JS:

```js
const AllRetryAfterHeaders: string[] = ["retry-after-ms", "x-ms-retry-after-ms", RetryAfterHeader];
```

So our "spec" is effectively "if `retry-after-ms` is present, use that as units of milliseconds; otherwise, if `x-ms-retry-after-ms` is present, use that as units of milliseconds; otherwise, if `retry-after` is present, use that in units of whole seconds; otherwise, use standard backoff heuristics."

jhendrixMSFT:
Fix will be included in `azcore@v1.9.1` which should go out week of 12/11.

PR #22121, opened 2023-12-08
Add more retry after headers to check during retries

jhendrixMSFT:
Include the headers that provide sub-second granularity.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/22120

### DIFF
--- a/sdk/azcore/internal/shared/shared.go
+++ b/sdk/azcore/internal/shared/shared.go
@@ -44,22 +44,61 @@ func Delay(ctx context.Context, delay time.Duration) error {
 	}
 }
 
-// RetryAfter returns non-zero if the response contains a Retry-After header value.
+// RetryAfter returns non-zero if the response contains one of the headers with a "retry after" value.
+// Headers are checked in the following order: retry-after-ms, x-ms-retry-after-ms, retry-after
 func RetryAfter(resp *http.Response) time.Duration {
 	if resp == nil {
 		return 0
 	}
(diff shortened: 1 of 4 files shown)


sdk/azcore/internal/shared/shared.go
```diff
@@ -44,22 +44,61 @@ func Delay(ctx context.Context, delay time.Duration) error {
...
-		return time.Until(t)
+
+	for _, retry := range retries {
+		v := resp.Header.Get(retry.header)
+		if v == "" {
+			continue
+		}
+		if retryAfter, _ := strconv.Atoi(v); retryAfter > 0 {
+			return time.Duration(retryAfter) * retry.units
+		} else if d, err := retry.custom(v); err == nil {
```
richardpark-msft, line 97: Do we need to convert 'v' into time.<Second|Millisecond> here? Or communicate the header it came in on so they know the units?

richardpark-msft (commented):
Question about the .custom() function

sdk/azcore/internal/shared/shared.go
jhendrixMSFT: No. The `custom` implementation means "if my regular algorithm failed do this instead".

jhendrixMSFT: I added a comment explaining this.

richardpark-msft (approved):

PR #22133, opened 2023-12-13
Implement UnmarshalJSON method for custom types

zjy282:
A UnmarshalJSON method has been added to the ChatCompletionsToolChoice and ChatRequestUserMessageContent structs. These implementations allow the structs to satisfy the json.Unmarshaller interface, enabling them to decode JSON directly into their respective types.


zjy282:

richardpark-msft:
@zjy282, these types are only used as input types so we don't need unmarshal functions (generally). What is your use case?

zjy282:
> @zjy282, these types are only used as input types so we don't need unmarshal functions (generally). What is your use case?

@richardpark-msft Thanks for your reply.
My service provides our enterprise with a unified outlet for LLMs, managing all model calls and keys. 
Therefore, my service needs to parse the client's json and call it according to the Azure SDK after processing. 
In order to improve versatility, my service remains consistent with the input and output parameters of the Azure interface, so I need these structures to parse the corresponding input parameter Json.

jhendrixMSFT (approved):

richardpark-msft (approved):

richardpark-msft:
@zjy282, thank you for this. Your use case makes total sense and I appreciate you taking the time to contribute.

Issue #22131, opened 2023-12-13
Go: Improve generation tool to update the package import version path in readme while releasing major version


Alancere:
See this issue for one example.
https://github.com/Azure/azure-sdk-for-go/issues/22127

PR #22135, opened 2023-12-13
Added replaceReadmeModule function

Alancere:
https://github.com/Azure/azure-sdk-for-go/issues/22131

### DIFF
--- a/eng/tools/generator/cmd/v2/common/generation.go
+++ b/eng/tools/generator/cmd/v2/common/generation.go
@@ -328,6 +328,11 @@ func (ctx *GenerateContext) GenerateForSingleRPNamespace(generateParam *Generate
 			}
 		}
 
+		log.Printf("Replace README.md module...")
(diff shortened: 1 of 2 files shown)


eng/tools/generator/cmd/v2/common/generation.go
```diff
@@ -328,6 +328,11 @@ func (ctx *GenerateContext) GenerateForSingleRPNamespace(generateParam *Generate
 			}
 		}
 
+		log.Printf("Replace README.md module...")
+		if err = replaceReadmeModule(packagePath, generateParam.RPName, generateParam.NamespaceName, version.String()); err != nil {
```
raych1, line 332: Only run this module when major version > 1. Or exit directly if major version > 1 within the function.

Alancere: It was released during refresh mgmt. The current version is v2.0.0-beta-1, but a stable v1.2.0 is required

Alancere: In that case, the module in README.md will not correspond

raych1 (approved):

PR #22150, opened 2023-12-14
Add selective component governance step

benbp:
Selectively turning on component governance checks for scheduled CI. Turning CG off explicitly for PR builds to hopefully reduce scanning cost on the CG team side of things.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -151,11 +151,16 @@ stages:
           Codeql.Enabled: true
           Codeql.BuildIdentifier: ${{ parameters.ServiceDirectory }}
           Codeql.SkipTaskAutoInjection: false
+          skipComponentGovernanceDetection: true
       pool:
         name: azsdk-pool-mms-ubuntu-2004-general
 
       steps:
       - template: /eng/common/pipelines/templates/steps/check-spelling.yml
+
+      - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
+        displayName: "Component Detection"
+        condition: and(succeededOrFailed(), ne(variables['Build.Reason'],'PullRequest'))
 
       - task: GoTool@0
         inputs:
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -151,12 +151,17 @@ stages:
           Codeql.Enabled: true
           Codeql.BuildIdentifier: ${{ parameters.ServiceDirectory }}
           Codeql.SkipTaskAutoInjection: false
+          skipComponentGovernanceDetection: true
```
weshaggard, line 154: Why not import globals.yml here?

weshaggard (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
benbp: It looks like globals is specifically not imported because of the overridden codeql variables here. IIRC if we import globals first the overrides will still take effect?

weshaggard: Correct the last one wins.

benbp: Ok I'll switch then

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -148,15 +148,23 @@ stages:
     - job: Analyze
       displayName: Analyze
       variables:
-          Codeql.Enabled: true
-          Codeql.BuildIdentifier: ${{ parameters.ServiceDirectory }}
-          Codeql.SkipTaskAutoInjection: false
+        - template: /eng/pipelines/templates/variables/globals.yml
```
weshaggard, line 151: Actually given this is imported at the stage level above isn't it already included?

benbp: Oh you're right I was thinking it was imported at the job scope not the stage scope.

weshaggard (approved):

benbp:
FYI @jhendrixMSFT this should skip CG in PRs now.

PR #22154, opened 2023-12-16
[azservicebus] Updating batching to allow for a configurable wait time

### LINKED ISSUE #19172: [azservicebus] Make defaultTimeAfterFirstMsg configurable in Azure Service Bus Receiver
### Feature Request

https://github.com/Azure/azure-sdk-for-go/blob/31ef2b8995177ed199449960cfff6f73ef3ea762/sdk/messaging/azservicebus/receiver.go#L392

`defaultTimeAfterFirstMsg` is used to create a `context.WithTimeout`, to receive as many messages as possible within that time once the first message is received. Currently it's not configurable and the value is too low (20ms) for some use cases. 

This proposal is to make the setting configurable via `ReceiveMessagesOptions`.

richardpark-msft:
Updating batching to allow for a configurable wait time. Can lead to fuller batches for people that want to tune it.

Fixes #19172

### DIFF
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -47,7 +47,6 @@ type Receiver struct {
 	amqpLinks                internal.AMQPLinks
 	cancelReleaser           *atomic.Value
 	cleanupOnClose           func()
-	defaultTimeAfterFirstMsg time.Duration
 	entityPath               string
 	lastPeekedSequenceNumber int64
 	maxAllowedCredits        uint32
@@ -131,7 +130,6 @@ func newReceiver(args newReceiverArgs, options *ReceiverOptions) (*Receiver, err
 	receiver := &Receiver{
 		cancelReleaser:           &atomic.Value{},
 		cleanupOnClose:           args.cleanupOnClose,
-		defaultTimeAfterFirstMsg: 20 * time.Millisecond,
 		lastPeekedSequenceNumber: 0,
 		maxAllowedCredits:        defaultLinkRxBuffer,
 		retryOptions:             args.retryOptions,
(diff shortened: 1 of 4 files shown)


sdk/messaging/azservicebus/receiver.go
```diff
@@ -181,7 +172,13 @@ func (r *Receiver) newReceiverLink(ctx context.Context, session amqpwrap.AMQPSes
...
 // ReceiveMessagesOptions are options for the ReceiveMessages function.
 type ReceiveMessagesOptions struct {
-	// For future expansion
+	// TimeAfterFirstMessage controls how long, after a message has been received, before we return the
+	// accumulated batch of messages.
+	//
+	// Default value depends on the receive mode:
+	// - 20ms when the receiver is in ReceiveModePeekLock
+	// - 1s when the receiver is in ReceiveModeReceiveAndDelete
+	TimeAfterFirstMessage time.Duration
```
jhendrixMSFT, line 181: Is this something that other SDKs have and we're just catching up?

richardpark-msft: No, new feature we're exposing based on the idea that Go is more performance based.

There's a customer ask for it, and it's come up in an issue before about configuring the value (in that case, less rather than more).

jhendrixMSFT: I see. Why are the times different depending on the mode?

richardpark-msft: (this was awhile ago)

There was a noticeable different in the # of messages I could get, with the wait time of 20ms, between the two modes. I left it a bit longer since there's no danger of expiration so it's okay to get more messages at a time.

This "time after first message" timer is really there as a way to prevent messages from expiring, while still trying to give some perf gains by batching.

jhendrixMSFT (approved):

PR #22165, opened 2023-12-20
feat: add ReplaceReadmeNewClientName function

Alancere:
- add `ReplaceReadmeNewClientName` function, ensure that NewXxxClient exists
- public model.SortFuncItem function


raych1 (approved):
Looks good to me only display one example of the clients in readme.md and keep it updated with the new release.

PR #22205, opened 2024-01-04
fix generator tool

Alancere:
current version: v3.0.0-beta.1
to release: v2.7.0 (compare v2.6.0, not BreakingChange)

The above does not replace the import module in fake and live_test.
ref: https://github.com/Azure/azure-sdk-for-go/pull/22184/commits/b8368c704ce55b61adfc37894e6786d47b5aedac#diff-878839ecfc6b682d3adbdd34fdeff0178e7018301d85ea4dca7453563d9ca93c

### DIFF
--- a/eng/tools/generator/cmd/v2/common/fileProcessor.go
+++ b/eng/tools/generator/cmd/v2/common/fileProcessor.go
@@ -513,73 +513,62 @@ func GetTag(path string) (string, error) {
 	return "", nil
 }
 
-func replaceModuleImport(path, rpName, namespaceName, currentVersion, subPath string, suffixes ...string) error {
+func replaceModuleImport(path, rpName, namespaceName, previousVersion, currentVersion, subPath string, suffixes ...string) error {
+	previous, err := semver.NewVersion(previousVersion)
+	if err != nil {
+		return err
+	}
+
 	current, err := semver.NewVersion(currentVersion)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


eng/tools/generator/cmd/v2/common/fileProcessor.go
```diff
@@ -566,11 +565,25 @@ func replaceModuleImport(path, rpName, namespaceName, previousVersion, currentVe
...
 
+func getModuleVersion(autorestPath string) (*semver.Version, error) {
+	data, err := os.ReadFile(autorestPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
+		if strings.HasPrefix(line, autorest_md_module_version_prefix) {
+			return semver.NewVersion(strings.TrimSpace(line[len(autorest_md_module_version_prefix):]))
```
raych1, line 580: Will `NewVersion` return an error when the value is not a valid version?

Alancere: yes

raych1: What I meant is if NewVersion returns error on an invalid value then should we customize the error message before returning?

raych1 (approved):

eng/tools/generator/cmd/v2/common/fileProcessor.go
Alancere: There will be no invalid versions in autorest.md, and the program should be stopped if the module-version in autorest.md is invalid

PR #22206, opened 2024-01-04
[Cosmos] Add preferred regions to CosmosClientOptions

simorenoh:
These changes publicly expose PreferredRegions to the ClientOptions to be used when initializing the cosmos client. This will allow users to provide a list of regions to be used on initialization in case the default region fails.

Still need to finalize the global endpoint manager policy/ refresh mechanism in order to fully integrate this into the global endpoint manager.

Closes https://github.com/Azure/azure-sdk-for-go/issues/18984

Api view link: https://apiview.dev/Assemblies/Review/8950b757ef3544f59d3fed21774a7966?revisionId=11999647939a4a988f2f093111513806&diffOnly=False&doc=False&diffRevisionId=4b909ee2481540aa9854dd8656bb820a

PR #22223, opened 2024-01-09
[Cosmos] Adds global endpoint manager policy and links GEM to client

simorenoh:
These changes add the global endpoint manager policy to the SDK and client, which will periodically run checks for updates in the user's Database Account information and keep the SDK updated to any changes in the account regions. This PR also links the GEM to the client, a link that had been previously missing.

One last thing missing is the merging of the preferred regions PR (https://github.com/Azure/azure-sdk-for-go/pull/22206) since that will also propagate the preferred regions to the global endpoint manager as opposed to it being an empty list.

### DIFF
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -24,6 +24,7 @@ import (
 type Client struct {
 	endpoint string
 	pipeline azruntime.Pipeline
+	gem      *globalEndpointManager
 }
 
 // Endpoint used to create the client.
@@ -36,7 +37,14 @@ func (c *Client) Endpoint() string {
 // cred - The credential used to authenticate with the cosmos service.
 // options - Optional Cosmos client options.  Pass nil to accept default values.
 func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*Client, error) {
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newSharedKeyCredPolicy(cred), o)}, nil
+	internalClient := &Client{endpoint: endpoint, pipeline: newInternalPipeline(newSharedKeyCredPolicy(cred), o), gem: &globalEndpointManager{}}
+
+	//need to pass in preferredRegions from options here once those changes are merged
+	gem, err := newGlobalEndpointManager(internalClient, []string{}, 0)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{endpoint: endpoint, pipeline: newPipeline(newSharedKeyCredPolicy(cred), gem, o), gem: gem}, nil
 }
 
 // NewClient creates a new instance of Cosmos client with Azure AD access token authentication. It uses the default pipeline configuration.
@@ -48,7 +56,16 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o)}, nil
+
+	internalClient := &Client{endpoint: endpoint, pipeline: newInternalPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o), gem: &globalEndpointManager{}}
+
+	//need to pass in preferredRegions from options here once those changes are merged
+	gem, err := newGlobalEndpointManager(internalClient, []string{}, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), gem, o), gem: gem}, nil
 }
 
 // NewClientFromConnectionString creates a new instance of Cosmos client from connection string. It uses the default pipeline configuration.
@@ -87,7 +104,7 @@ func NewClientFromConnectionString(connectionString string, o *ClientOptions) (*
 	return NewClientWithKey(endpoint, cred, o)
 }
 
-func newPipeline(authPolicy policy.Policy, options *ClientOptions) azruntime.Pipeline {
+func newPipeline(authPolicy policy.Policy, gem *globalEndpointManager, options *ClientOptions) azruntime.Pipeline {
 	if options == nil {
 		options = &ClientOptions{}
 	}
@@ -98,7 +115,21 @@ func newPipeline(authPolicy policy.Policy, options *ClientOptions) azruntime.Pip
 				&headerPolicies{
 					enableContentResponseOnWrite: options.EnableContentResponseOnWrite,
 				},
+				&globalEndpointManagerPolicy{gem: gem},
 			},
+			PerRetry: []policy.Policy{
+				authPolicy,
+			},
+		},
+		&options.ClientOptions)
+}
+
+func newInternalPipeline(authPolicy policy.Policy, options *ClientOptions) azruntime.Pipeline {
+	if options == nil {
+		options = &ClientOptions{}
+	}
+	return azruntime.NewPipeline("azcosmos", serviceLibVersion,
+		azruntime.PipelineOptions{
 			PerRetry: []policy.Policy{
 				authPolicy,
 			},
@@ -176,6 +207,7 @@ func (c *Client) CreateDatabase(
 	if err != nil {
 		return DatabaseResponse{}, err
 	}
+	fmt.Printf("- db create succeeded with code %d", azResponse.StatusCode)
 
 	return newDatabaseResponse(azResponse)
 }
--- a/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go
+++ b/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+type globalEndpointManagerPolicy struct {
+	gem *globalEndpointManager
+}
+
+func (p *globalEndpointManagerPolicy) Do(req *policy.Request) (*http.Response, error) {
+	shouldRefresh := p.gem.ShouldRefresh()
+	if shouldRefresh {
+		fmt.Println("should refresh true go routine")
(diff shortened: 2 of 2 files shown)


sdk/data/azcosmos/cosmos_client.go
```diff
@@ -176,6 +207,7 @@ func (c *Client) CreateDatabase(
 	if err != nil {
 		return DatabaseResponse{}, err
 	}
+	fmt.Printf("- db create succeeded with code %d", azResponse.StatusCode)
```
ealsur, line 210: is this test code?

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -48,7 +56,16 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o)}, nil
+
+	internalClient := &Client{endpoint: endpoint, pipeline: newInternalPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o), gem: &globalEndpointManager{}}
```
ealsur, line 60: you don't need 2 clients, just 2 pipelines. You can have both pipelines in the same `Client` and then when you can pass the internal pipeline to `newGlobalEndpointManager` + the endpoint. Inside Global Endpoint Manager, instead of using the client to send the request, you can send it through the internal pipeline (which contains the auth too).

simorenoh: I was using an internal client as opposed to an internal pipeline here since regardless the call made within the Update() method are to the client.sendGetRequest() method, which also calls client.createRequest() to create the request we will send and client.executeAndEnsureSuccessResponse() to send said request - and only then in this method do we use the pipeline.

I figured the refactoring of all of this into whatever logic we would need to only use the internal pipeline would end up being overkill (separate logic for creating/sending the request), when in the end the way the request would be sent is through a pipeline anyway. I figured since it would be the same in terms of number of connections, doing it this way would be more straightforward. Maybe I'm missing something here though, happy to get a proper explanation if that's the case - not sure if there's some additional overhead I'm not seeing here.

simorenoh: I don't know why I thought the methods were a lot bigger - went ahead and pushed the updates, thanks for pointing this out 👍

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -115,19 +117,39 @@ func (gem *globalEndpointManager) GetAccountProperties(ctx context.Context) (acc
 		return accountProperties{}, fmt.Errorf("failed to generate path for name-based request: %v", err)
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
-	azResponse, err := gem.client.sendGetRequest(path, ctx, operationContext, nil, nil)
-	cancel()
+	finalURL := gem.clientEndpoint
+	if path != "" {
+		finalURL = azruntime.JoinPaths(gem.clientEndpoint, path)
+	}
```
simorenoh, line 123: we don't technically need this code at all (nor lines 113/115-116/118 above) since the request path just ends up being the database account link (which is the default endpoint) - I left it for uniformity in how we create the requests but wanted to point this out to see if we should just remove it

ealsur: I would just remove it

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -115,19 +117,39 @@ func (gem *globalEndpointManager) GetAccountProperties(ctx context.Context) (acc
...
+	defer cancel()
+	req, err := azruntime.NewRequest(ctxt, http.MethodGet, finalURL)
 	if err != nil {
-		return accountProperties{}, fmt.Errorf("failed to retrieve account properties: %v", err)
+		return accountProperties{}, err
 	}
 
-	properties, err := newAccountProperties(azResponse)
+	req.Raw().Header.Set(headerXmsDate, time.Now().UTC().Format(http.TimeFormat))
+	req.Raw().Header.Set(headerXmsVersion, "2020-11-05")
```
ealsur, line 133: nit: Do we have any const somewhere with this value?

ealsur: like supportedCapabilitiesHeaderValue

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -115,19 +117,39 @@ func (gem *globalEndpointManager) GetAccountProperties(ctx context.Context) (acc
...
+	req.SetOperationValue(operationContext)
+
+	azResponse, err := gem.pipeline.Do(req)
 	if err != nil {
-		return accountProperties{}, fmt.Errorf("failed to parse account properties: %v", err)
+		return accountProperties{}, err
 	}
 
-	return properties, nil
+	successResponse := (azResponse.StatusCode >= 200 && azResponse.StatusCode < 300) || azResponse.StatusCode == 304
```
ealsur, line 143: 304 for DatabaseAccount is not expected, you can just reduce this

```suggestion
	successResponse := (azResponse.StatusCode >= 200 && azResponse.StatusCode < 300) 
```

sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go
```diff
@@ -0,0 +1,25 @@
...
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+type globalEndpointManagerPolicy struct {
+	gem *globalEndpointManager
+}
+
+func (p *globalEndpointManagerPolicy) Do(req *policy.Request) (*http.Response, error) {
```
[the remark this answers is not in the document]
simorenoh: yep for sure, was holding off on the tests until the design was looking good to us - will work on adding them now

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
simorenoh: I don't believe we do since I grabbed this code as-is from our client code:
image
I could add it under the `version` file as a separate constant, or make a separate PR to add a constants file to the SDK which we can build on as it develops

ealsur: Add it to th cosmos-client.go file as a private const, and reuse it.

sdk/data/azcosmos/emulator_cosmos_global_endpoint_manager_test.go
```diff
@@ -73,3 +73,68 @@ func TestGlobalEndpointManagerEmulator(t *testing.T) {
 	assert.Equal(t, len(locationInfo.availReadEndpointsByLocation), len(availableEndpointsByLocation)+1)
 	assert.Equal(t, len(locationInfo.availWriteEndpointsByLocation), len(availableEndpointsByLocation)+1)
 }
+
+func TestGlobalEndpointManagerPolicyEmulator(t *testing.T) {
```
ealsur, line 77: How is this tests testing the policy? Based on the code, it is just accessing the GEM within the client and manually execting GetAccountProperties (which is not a bad test) but it does not verify that if a user does an operation (for example, ReadItem) the policy is calling GetAccountProperties in the back?

Maybe if you replace the GetAccountProperties line for a normal operation on the client, the other asserts can remain in place?

ealsur: Or inject a policy on client creation (like in some of the unit tests) to verify that there was an HTTP request made for the account policies?

ealsur: Alternatively, adding unit tests for the policy itself could also work (like we have for https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azcosmos/cosmos_headers_policy_test.go)

simorenoh: I am running an http request to populate the account information - the information is only populated after the create database call that is being made. If you'd like for me to remove some of the checks being done before the request is made I can, but what you mentioned in your first point here is exactly what I am testing.
image

ealsur: My bad, I got confused. It seems the piece of code with the gem.GetAccountProperties is really not part of the testing of the policy

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -48,7 +58,13 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o)}, nil
+	//need to pass in preferredRegions from options here once those changes are merged
+	gem, err := newGlobalEndpointManager(endpoint, newInternalPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o), []string{}, 0)
```
ealsur, line 62: Pass the preferred regions from the options?

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -36,7 +41,12 @@ func (c *Client) Endpoint() string {
 // cred - The credential used to authenticate with the cosmos service.
 // options - Optional Cosmos client options.  Pass nil to accept default values.
 func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*Client, error) {
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newSharedKeyCredPolicy(cred), o)}, nil
+	//need to pass in preferredRegions from options here once those changes are merged
```
ealsur, line 44: But: remove comment

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -48,7 +58,13 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), o)}, nil
+	//need to pass in preferredRegions from options here once those changes are merged
```
ealsur, line 61: Not: remove comment

ealsur (approved):

PR #22230, opened 2024-01-10
Remove v1.6.0-beta API

chlowell:
This removes persistent caching and the `AzureCLICredentialOptions.Subscription` field from the public surface so we can publish v1.5.0 without them. After that release I'll publish v1.6.0-beta.1 from main, which won't have this commit, giving us:

### v1.5.0
- AzureDeveloperCLICredential (the only new API)
- upgraded dependencies
- bug fixes
- preserve MSAL request headers
- tracing
- Azure ML managed identity support

### v1.6.0-beta.1
- persistent caching
- AzureCLICredentialOptions.Subscription


jhendrixMSFT (approved):
Looks like this was easier than branching for a `v1.5.0` release.

chlowell:
It was much easier than cherry-picking onto v1.4.0. But I could merge this to a release branch instead of main to avoid having to revert the commit later. I like that idea even though it has the downside of creating another immortal release branch 😄

PR #22245, opened 2024-01-12
[azdatalake][STG84] Append with flush

tanyasethi-msft:


sdk/storage/azdatalake/file/models.go
```diff
@@ -243,6 +243,8 @@ type AppendDataOptions struct {
 	LeaseAccessConditions *LeaseAccessConditions
 	// CPKInfo contains optional parameters to perform encryption using customer-provided key.
 	CPKInfo *CPKInfo
+	//Flush Optional. If true, the file will be flushed after append.
+	Flush *bool
```
jhendrixMSFT, line 247: If `Flush: nil` is semantically equivalent to `Flush: false` then it would be better to define this as `Flush: bool` as that's easier to work with.

tanyasethi-msft: We usually use pointers for optional boolean parameters, [like here](https://github.com/Azure/azure-sdk-for-go/blob/83f777f68033303b5629cc05e9b960357028ca8d/sdk/storage/azblob/blockblob/models.go#L47). Also, this Flush parameter is defined as a pointer in [generated code.](https://github.com/tanyasethi-msft/azure-sdk-for-go/blob/a807597a59ba65c82de5ccdba3af25a303dde874/sdk/storage/azdatalake/internal/generated/zz_models.go#L285)
Happy to discuss if a change is needed in this approach.

vibhansa-msft: Agree "true" looks better and more user friendly then "to.Ptr(true)". Are there other places where we send true/false in form of pointers ?

vibhansa-msft: I see that auto generated code itself take *bool and hence our wrappers have also taken dependency on similar types.

tanyasethi-msft: Yes @vibhansa-msft, everywhere we send true/false in form of pointers, eg - [Link
](https://github.com/Azure/azure-sdk-for-go/blob/83f777f68033303b5629cc05e9b960357028ca8d/sdk/storage/azblob/blockblob/models.go#L47)

souravgupta-msft: There is a check where if `flush` is nil, it doesn't add the query parameter in the request URL ([source](https://github.com/Azure/azure-sdk-for-go/blob/cef53c7521e50f982853cc71065c81f99904aec5/sdk/storage/azdatalake/internal/generated/zz_path_client.go#L70-L72)).
If we have this as pointer, it will prevent adding this query parameter. Otherwise, it will add `flush=false` every time if user has not set this parameter.

souravgupta-msft (approved):

sdk/storage/azdatalake/file/models.go
jhendrixMSFT: If omitting it is important, then go ahead and use `*bool`.

PR #22258, opened 2024-01-17
Fix/module name change

tanyasethi-msft:


vibhansa-msft (approved):

souravgupta-msft:
There should ideally be one commit showing here which should be the last one, right?

sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -45,12 +45,12 @@ const (
 const crc64Polynomial uint64 = 0x9A6C9329AC4BC9B5
 
 const (
-	AppendBlobClient = "azblob/appendblob.Client"
```
jhendrixMSFT, line 48: As of `azcore@v1.9.0`, the `moduleName` param to `azcore.NewClient()` takes the full module name instead of the `package.ClientName`. So, all of these constants can go away.

jhendrixMSFT: Fully qualified module name e.g. `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob`.

Originally, we wanted `package.ClientName` for use by the tracing policy. We've retooled how tracing works, so passing `package.ClientName` is no longer necessary.

Regarding the telemetry policy, we don't want/need the client name in the `User-Agent` string. See the [guidelines](https://azure.github.io/azure-sdk/general_azurecore.html#user-agent-format) for the format.

vibhansa-msft: In current implementation, "azsdk-go-service.Client/v1.2.0 (go1.19.3; Windows_NT)" this user-agent string is not able to identify whether request was generated by "blob", "datalake" or "files" client as all three can have service client. Hence we are removing the "/" to include "azblob" or "azdatalake" in the string as well.
For the other part "service.Client" does not make much sense for telemetry in any way. If we are interested in just knowing which module was used then "azblob.service" itself tell the whole story ".client" is not adding any value there.

vibhansa-msft: On FE side we are building telemetry to parase the User-Agent string and dump some of the keywords to a dgrep table. Longer the string or more vivid the string, higher resource consumption between FE and XArgus. "azsdk-go-blob" shall be good enough for us to understand what SDK was used. Does Azure-SDK team have any sort of telemetry already built to identify which customers are using what SDK or which clients?

vibhansa-msft:
Corrected the chageset in #22273

PR #22283, opened 2024-01-23
Data race while download file

tanyasethi-msft:
- [] The purpose of this PR is explained in this or a referenced issue.


sdk/storage/azblob/blob/client.go
```diff
@@ -525,7 +521,7 @@ func (b *Client) downloadFile(ctx context.Context, writer io.Writer, o downloadO
 		OperationName: "downloadBlobToWriterAt",
 		TransferSize:  count,
 		ChunkSize:     o.BlockSize,
-		NumChunks:     numChunks,
+		NumChunks:     uint16(numChunks),
 		Concurrency:   o.Concurrency,
 		Operation: func(ctx context.Context, chunkStart int64, count int64) error {
 			buff, err := acquireBuffer()
```
vibhansa-msft, line 527: We can directly get the buffer from channel here instead of making a func call.
```suggestion
			buff <- buffers.Acquire()
```

sdk/storage/azblob/internal/shared/shared.go
```diff
@@ -44,6 +44,15 @@ const (
 
 const crc64Polynomial uint64 = 0x9A6C9329AC4BC9B5
 
+const (
+	AppendBlobClient = "azblob/appendblob.Client"
```
vibhansa-msft, line 48: This shall be cleaned up.

Issue #22286, opened 2024-01-23
Azlog: Allow to only log Response Body on failures


ealsur:
### Feature Request

In Cosmos DB, customers send document information (might contain PII) in the body of the requests and the response on successful operations, contain the document information too.

For example, a call to `CreateItem` would, on a successful response, contain the created document.

When enabling logging through `AZURE_SDK_GO_LOGGING`, customers can log Request and Responses, and for the Responses, they include the Body.

From a logging perspective, the Body is useful on error scenarios (the service returned an error, which included details in the body), but for success responses, logging the body might contain PII information. 

The current PolicyOptions, allow to set to include the body:

https://github.com/Azure/azure-sdk-for-go/blob/6cdec79da53b0e16e3a47104da782a94ecf83244/sdk/azcore/runtime/policy_logging.go#L71-L75

but this affects all (success and failure) responses:

https://github.com/Azure/azure-sdk-for-go/blob/6cdec79da53b0e16e3a47104da782a94ecf83244/sdk/azcore/runtime/policy_logging.go#L141-L143

Would it be possible to define a behavior that only logs the body on failure responses (HTTP status code >= 400)?

Similar (but not equal) issue: https://github.com/Azure/azure-sdk-for-go/issues/18799

jhendrixMSFT:
Our current thinking is that we'll add a new logging event `ResponseError` that writes the contents of `azcore.ResponseError.Error()` to the log whenever one is created.

jhendrixMSFT:
The new functionality will go out in `azcore@v1.10.0` which should be first week of March. Note that in order to get the new functionality, you must update `newCosmosError()` to use `runtime.NewResponseErrorWithErrorCode()`.

simorenoh:
We have updated the azcore version to 1.10.0 in this PR and merged it: https://github.com/Azure/azure-sdk-for-go/pull/22288/ - thanks @jhendrixMSFT!

PR #22288, opened 2024-01-24
[Cosmos] Add logging changes and sample

simorenoh:
These changes extend the list of allowed headers in the default logging policy to take Cosmos headers into account and display them with their value, as opposed to "REDACTED". Also adds a small entry to the README showing users how they can use logging and how they can configure the listener to output the data to different sources.

Last pending item would be to remove the currently present `replace` directive in the go.mod file to utilize the new azcore version once the changes are merged. Sample error logs look something like this now, which is exactly what we were hoping for since they include the diagnostic string:
image

Closes https://github.com/Azure/azure-sdk-for-go/issues/22286


sdk/data/azcosmos/README.md
```diff
@@ -34,6 +34,27 @@ You can create an Azure Cosmos account using:
...
+
+Once that is done, the SDK will begin to collect diagnostics through `stdout` - printing directly to your console. If you'd like to configure a listener that acts differently, the small snippet below shows how you could do so. In this case, the listener is configured to write to a file instead of printing to the console.
+
+```go
+import (
+	"os"
+	azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
+)
+
+    f, err := os.Create("cosmos-log-file.txt")
```
ealsur, line 49: Please align padding correctly for the lines

sdk/data/azcosmos/README.md
```diff
@@ -34,6 +34,27 @@ You can create an Azure Cosmos account using:
...
+	handle(err)
+	defer f.Close()
+
+	azlog.SetListener(func(event azlog.Event, s string) {
+		f.WriteString(s + "\n")
+	})
```
ealsur, line 55: Might be worth adding the example on how to filter? Like https://github.com/Azure/azure-sdk-for-go/blob/750d8afc2cec865de4f1b18386654bb8acccec2b/sdk/ai/azopenai/README.md?plain=1#L70

And show requests, responses, and retry policies?

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -481,3 +482,79 @@ type pipelineRequestOptions struct {
...
+		cosmosHeaderIsBatchAtomic,
+		cosmosHeaderIsBatchOrdered,
+		cosmosHeaderSDKSupportedCapabilities,
+		headerXmsDate,
+		headerContentType,
+		headerIfMatch,
+		headerIfNoneMatch,
+		headerXmsVersion,
+		"Content-Location",
+		"X-Ms-Gatewayversion",
```
ealsur, line 524: Please use all lowercase for header names if possible.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -481,3 +482,79 @@ type pipelineRequestOptions struct {
...
+		cosmosHeaderIsUpsert,
+		cosmosHeaderOfferThroughput,
+		cosmosHeaderOfferAutoscale,
+		cosmosHeaderQuery,
+		cosmosHeaderOfferReplacePending,
+		cosmosHeaderOfferMinimumThroughput,
+		cosmosHeaderResponseContinuationTokenLimitInKb,
+		cosmosHeaderEnableScanInQuery,
+		cosmosHeaderMaxItemCount,
+		cosmosHeaderContinuationToken,
```
ealsur, line 508: Does ContinuationToken have any PII or critical information that shouldn't be logged?

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -481,3 +482,79 @@ type pipelineRequestOptions struct {
 	isRidBased            bool
 	isWriteOperation      bool
 }
+
+func getAllowedHeaders() []string {
+	return []string{
+		cosmosHeaderRequestCharge,
```
ealsur, line 488: Proposed list:

```
cosmosHeaderRequestCharge,
cosmosHeaderActivityId,
cosmosHeaderEtag,
cosmosHeaderSessionToken,
cosmosHeaderConsistencyLevel,
cosmosHeaderQuery,
cosmosHeaderMaxItemCount,
cosmosHeaderQueryMetrics,
cosmosHeaderCorrelatedActivityId,
cosmosHeaderIsBatchRequest,
cosmosHeaderIsBatchAtomic,
cosmosHeaderIsBatchOrdered,
headerXmsDate,
headerContentType,
headerXmsVersion,
"x-ms-cosmos-llsn",
"x-ms-cosmos-item-llsn",
"x-ms-item-lsn",
"x-ms-cosmos-quorum-acked-llsn",
"x-ms-global-committed-lsn",
"x-ms-last-state-change-utc",
"x-ms-quorum-acked-lsn",
"x-ms-request-duration-ms",
"x-ms-transport-request-id",
"x-ms-documentdb-partitionkeyrangeid",
"x-ms-cosmos-physical-partition-id",
"x-ms-item-count",
"x-ms-start-epk",
"x-ms-end-epk",
"a-im",
```

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -481,3 +482,79 @@ type pipelineRequestOptions struct {
...
+		headerXmsDate,
+		headerContentType,
+		headerIfMatch,
+		headerIfNoneMatch,
+		headerXmsVersion,
+		"Content-Location",
+		"X-Ms-Gatewayversion",
+		"lsn",
+		"X-Ms-Cosmos-Llsn",
+		"X-Ms-Cosmos-Item-Llsn",
```
kirankumarkolli, line 527: This list is now duplicated apart from main usage inside SDK right?

Thoughts on listing named header from single place and populate allowed list here from them. 
That way casing and consistency will be cover automatically right?

simorenoh: I agree with this suggestion, putting whatever additional headers we will use along with the ones already present in the SDK

sdk/data/azcosmos/cosmos_client.go
simorenoh: blacklisting model within the pipeline would depend on Core unless we make our own policy like we mentioned yesterday

sdk/data/azcosmos/cosmos_client.go
simorenoh: yep just left them as-is here while we decide on the final list

sdk/data/azcosmos/cosmos_client.go
simorenoh: continuation tokens afaik look like this:

'{"token":"-RID:~u3IiALb3QSEBAAAAAAAAAA==#RT:1#TRC:1#ISV:2#IEO:65567#QCF:8","range":{"min":"","max":"FF"}}'

so no? or am I missing something here?

sdk/data/azcosmos/cosmos_client.go
ealsur: @kirankumarkolli When we discussed this, the point here is ROI. In order to support blacklisting either the base core SDK needs to do it or we need to build our own logging policy (non-trivial).

Given the lack of any sort of feedback loop and the unknown on when or if the base core SDK will support it, shouldn't we move forward with this?

ealsur (approved):
Please get the ack from Kiran regarding his comment

sdk/data/azcosmos/cosmos_client.go
simorenoh: we discussed this offline and it's ok for the initial diagnostics work

sdk/data/azcosmos/cosmos_client_test.go
```diff
@@ -102,8 +100,8 @@ func TestEnsureErrorIsGeneratedOnResponse(t *testing.T) {
 	}
 
 	asError := err.(*azcore.ResponseError)
-	if asError.ErrorCode != someError.Code {
-		t.Errorf("Expected %v, but got %v", someError.Code, asError.ErrorCode)
+	if asError.ErrorCode != "404 Not Found" {
```
ealsur, line 103: With the new error, is the "SomeCode" body logged anywhere? This test should validate that the "SomeCode" is part of the error, maybe the body?

sdk/data/azcosmos/README.md
```diff
@@ -34,6 +34,33 @@ You can create an Azure Cosmos account using:
...
+
+// Configure the listener to write to a file rather than to the console
+azlog.SetListener(func(event azlog.Event, s string) {
+	f.WriteString(s + "\n")
+})
+
+// Filter the types of events you'd like to log by removing the ones you're not interested in (if any)
+// We recommend using the default logging with no filters - but if filtering we recommend *always* including 
+// `azlog.EventResponseError` since this is the event type that will help with debugging errors
+azlog.SetEvents(azlog.EventRequest, azlog.EventResponse, azlog.EventRetryPolicy, azlog.EventResponseError)
```
ealsur, line 61:
```suggestion
azlog.SetEvents(azlog.EventRequest, azlog.EventRetryPolicy, azlog.EventResponseError) 
```

Since `azlog.EventResponse` logs the full body, which might contain PII information, would it be better to not recommend it?

ealsur (approved):

sdk/data/azcosmos/README.md
ealsur: Nevermind, I read it wrong.

EventResponseError always returns the body, EventResponse only if PolicyOptions.IncludeBody is true

PR #22299, opened 2024-01-25
AzureCLICredential prefers expires_on value

### LINKED ISSUE #21886: [Identity] Evaluate expires_on field in AzureCLICredential
As of [Azure CLI v2.54.0](https://github.com/Azure/azure-cli/milestone/135), the `az account get-access-token` command returns a new `expires_on` field containing a POSIX timestamp. This behavior is a departure from previous versions, in which the local datetime was returned in an `expiresOn` field. Update `AzureCliCredential` to also consider the new `expires_on` field.

Related links:
- https://github.com/Azure/azure-sdk-for-cpp/pull/5076#issuecomment-1784680918
- https://github.com/Azure/azure-cli/pull/27476

chlowell:
When `az` specifies `expires_on`, we prefer that because it's a POSIX timestamp i.e. straightforward to parse. Otherwise, we fall back to parsing `expiresOn`, a local timestamp, as we do today. Closes #21886

### DIFF
--- a/sdk/azidentity/azure_cli_credential.go
+++ b/sdk/azidentity/azure_cli_credential.go
@@ -166,6 +166,7 @@ func (c *AzureCLICredential) createAccessToken(tk []byte) (azcore.AccessToken, e
 		AccessToken      string `json:"accessToken"`
 		Authority        string `json:"_authority"`
 		ClientID         string `json:"_clientId"`
+		Expires_On       int    `json:"expires_on"`
 		ExpiresOn        string `json:"expiresOn"`
 		IdentityProvider string `json:"identityProvider"`
 		IsMRRT           bool   `json:"isMRRT"`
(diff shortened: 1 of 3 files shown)


sdk/azidentity/azure_cli_credential.go
```diff
@@ -163,26 +163,21 @@ var defaultAzTokenProvider azTokenProvider = func(ctx context.Context, scopes []
...
-		ClientID         string `json:"_clientId"`
-		ExpiresOn        string `json:"expiresOn"`
-		IdentityProvider string `json:"identityProvider"`
-		IsMRRT           bool   `json:"isMRRT"`
-		RefreshToken     string `json:"refreshToken"`
-		Resource         string `json:"resource"`
-		TokenType        string `json:"tokenType"`
-		UserID           string `json:"userId"`
+		AccessToken string `json:"accessToken"`
+		Expires_On  int    `json:"expires_on"`
```
jhendrixMSFT, line 167: Should this be `int64`? On 64-bit systems it will be 64 bits wide, but on a 32-bit system it will be 32 bits wide. Will the value always fit into a 32-bit int?

chlowell: It will until 2038 i.e. the Epochalypse, because the value is a Unix timestamp which won't overflow 32 bits before then. But sure, this may as well be `int64`.

jhendrixMSFT (approved):

PR #22302, opened 2024-01-26
Ignore Mgmt README go.dev link

Alancere:


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -110,6 +110,14 @@ steps:
     failOnStderr: true
     workingDirectory: 'sdk/${{parameters.ServiceDirectory}}'
 
+  - task: PowerShell@2
+    displayName: 'Ignore Mgmt README.md Link'
```
raych1, line 114: Can you change the displayName to `Add module link in README.md to 'ignore-links.txt'`?

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,19 @@
+param (
+    [string]$serviePath,
+    [string]$outputFile
+)
+
+if ($serviePath -notlike "*sdk/resourcemanager*" -or $serviePath -like "*sdk/resourcemanager/internal*") {
+    return
+}
+
+$lines = Get-Content -Path $serviePath/README.md
```
raych1, line 10: Can you `Test-Path` the `README.md`?

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,19 @@
...
+
+$pattern = "(https://pkg\.go\.dev/github\.com/Azure/azure-sdk-for-go/$serviePath[a-zA-Z0-9_.\-/]*)"
+$matchPkg = $lines | Where-Object { $_ -match $pattern }
+
+foreach ($m in $matchPkg) {
+    $link = [regex]::Match($m, $pattern).Groups[1].Value
+    Write-Host "Ignore Mgmt README.md Link: $link"
+    $link | Out-File -FilePath $outputFile -Append
```
raych1, line 18: Same here to check if `outputFile` exist?

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,29 @@
+param (
+    [string]$serviePath,
```
raych1, line 2: Both of these two parameters appear to be required, can you add [ValidateNotNullOrEmpty()]?

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,29 @@
...
+    [string]$outputFile
+)
+
+if ($serviePath -notlike "*sdk/resourcemanager*" -or $serviePath -like "*sdk/resourcemanager/internal*") {
+    return
```
raych1, line 7: How about use `exit` either?

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,29 @@
...
+    Write-Host "$serviePath/README.md does not exist."
+    exit
+}
+
+if (-not (Test-Path $outputFile)) {
+    Write-Host "$outputFile does not exist."
+    exit
+}
+
+$lines = Get-Content -Path $serviePath/README.md
```
raych1, line 20: Can you wrap line#20 to line#28 within a `try-catch` block?

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -110,6 +110,14 @@ steps:
     failOnStderr: true
     workingDirectory: 'sdk/${{parameters.ServiceDirectory}}'
 
+  - task: PowerShell@2
```
weshaggard, line 113: @Alancere @raych1 what is the context for this change? I don't see any description in the PR so I'm not sure why we are doing this.

eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```diff
@@ -0,0 +1,37 @@
+param (
```
raych1, line 1: Can you add description for this script?
```
<#
.SYNOPSIS
  Add the Go management module link in readme.md to the 'ignore-links.txt'.

.DESCRIPTION
  This script is used to add the module link to 'ignore-links.txt' to skip it from the links verification. The reason to skip it from the validation is the module link is not available at this time during this pipeline run. It will be available after the release tag is created at Azure/azure-sdk-for-go repository.

.PARAMETER servicePath
  The path of the service.

.PARAMETER outputFile
  The path of the 'ignore-links.txt'.

.EXAMPLE
  Ignore-Mgmt-README-Link -servicePath "/home/azure-sdk-for-go/sdk/servicea" -outputFile "/home/azure-sdk-for-go/eng/ignore-links.txt"

#>
```

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -110,6 +110,14 @@ steps:
...
+    displayName: Add module link in README.md to 'ignore-links.txt'
+    inputs:
+      targetType: 'filePath'
+      filePath: ./eng/scripts/Ignore-Mgmt-REAMDE-Link.ps1
```
raych1, line 117: Can you fix the typo of 'README'?

eng/pipelines/templates/steps/analyze.yml
raych1: @weshaggard , I added more info to the description.

weshaggard:
@raych1 if this is just the code inside of the install link we shouldn't be trying to verify that link as that isn't a clickable hyperlink. Based on your example it looks like it is the badge that has the hyperlink.
```
[![PkgGoDev](https://pkg.go.dev/badge/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform/v2)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform/v2)
```
Do we know if that is interesting enough to keep? If so, then perhaps a solution like this could work. BTW I think we have a similar issue in the JS mgmt Readmes.

eng/pipelines/templates/steps/analyze.yml
weshaggard: Even if we go with this solution this is essentially always ignoring this link which we don't want to do. At what point would we ever validate it?

While this is an interesting idea I'm not convinced it is the best option here.

weshaggard (changes requested):
I would like us to think about some other options that don't aways ignore these package links otherwise they will end up broken at some point.

raych1:
> I would like us to think about some other options that don't aways ignore these package links otherwise they will end up broken at some point.

@weshaggard , I agree that we shall automate the process to verify a new version of a Go module availability in `pkg.go.dev`. At present, we install the new module locally after the release. The challenge of the automation is it may take a few minutes to make the module available in `pkg.go.dev` after the new tag and new release have been created in the Go SDK repo. I don't see JS release pipeline have the automation to verify the package availability on `npmjs.com`. 
Do you have any idea of the other languages? If not, then I will talk to the other language owners after my holidays.

raych1:
> @raych1 if this is just the code inside of the install link we shouldn't be trying to verify that link as that isn't a clickable hyperlink. Based on your example it looks like it is the badge that has the hyperlink.
> 
> ```
> [PkgGoDev](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform/v2)
> ```
> 
> Do we know if that is interesting enough to keep? If so, then perhaps a solution like this could work. BTW I think we have a similar issue in the JS mgmt Readmes.

We might ignore the badge link in the stage of verifying links then add a step after the git tag and release have been created.

weshaggard:
JS has been adding the link to ignore-links.txt https://github.com/Azure/azure-sdk-for-js/blob/main/eng/ignore-links.txt as part of the SDK generation step. Then later removing it from there after the package is published. I don't like that approach much either, but I do like it better then never checking the link.

raych1:
> JS has been adding the link to ignore-links.txt https://github.com/Azure/azure-sdk-for-js/blob/main/eng/ignore-links.txt as part of the SDK generation step. Then later removing it from there after the package is published. I don't like that approach much either, but I do like it better then never checking the link.

@weshaggard , I confirmed with JS language owner that the link added to `ignore-links.txt` for JS in generation step is the document link rather than the package link in `npmjs.com`.  Later on, it won't be removed manually on regular basis. 

I prefer to add the badge link check after the release tag is created. But the challenge is `pkg.go.dev` will take about 30 mins to 1 hour to make the new published module available base on our testing. 

JS release pipeline only verify the package can be installed locally and doesn't verify the package availability at `npmjs.com` after it's published to npmjs.com. This is [an pipeline run example](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3355126&view=logs&j=49259b2a-9f18-5945-e5de-00969e5597f4&t=0c39f749-0514-5b3a-97a8-63b59aa66884).

weshaggard:
I'd suggest we use a similar approach to what JS is doing with adding it to the ignore list and then removing it at some future point. Or we come up with some other mechanism like not having the badge link at all. The problem we have with providing links to something that doesn't exist yet is that it often can lead to broken links which is why we are trying to be strict about this link verification.

Alancere:
case: https://github.com/Azure/azure-sdk-for-go/pull/22428/checks?check_run_id=21967617298

raych1:
> I'd suggest we use a similar approach to what JS is doing with adding it to the ignore list and then removing it at some future point. Or we come up with some other mechanism like not having the badge link at all. The problem we have with providing links to something that doesn't exist yet is that it often can lead to broken links which is why we are trying to be strict about this link verification.

@weshaggard , the proposed change in this PR is to temporarily add the link to the `ignore-links.txt` file within the pipeline run. This will cause the subsequent verification step to bypass validation for this particular link. Importantly, changes to the ignore-links.txt file will not be committed to the repository, which I believe is a more preferable approach.

This issue is currently blocking release PRs that include major version upgrades for the SDK. Can we take this approach as short term solution and explore a long-term solution for badge link verification in the future? Otherwise, we will need to force merge the release PRs.

raych1:
This is an example PR that had to be forced merged due to a CI failure caused by 404 error with the badge link.
https://github.com/Azure/azure-sdk-for-go/pull/22428

weshaggard:
>@weshaggard , the proposed change in this PR is to temporarily add the link to the ignore-links.txt file within the pipeline run. This will cause the subsequent verification step to bypass validation for this particular link. Importantly, changes to the ignore-links.txt file will not be committed to the repository, which I believe is a more preferable approach.

The issue I have here is that the approach in this PR will always ignore this link and never verify it in the future either. At least by adding it manually to the ignore-links.txt file it will only ignore it until you remove it from that list after it releases.

Issue #22281, opened 2024-01-22
[AzCopy] Datalake SDK does not path escape path names when passed to NewDirectoryClient and NewFileClient methods. 


gapra-msft:
### Bug Report

- import path of package in question `.../storage/azdatalake`
- SDK version  `latest`,

- What happened?
Create a directory called #,%,? in a HNS enabled account.
Create a directory client from a filesystem client and perform operations that hit the blob endpoint and dfs endpoint on it.
fsClient.NewDirectoryClient("#,%,?")

Any operations hitting blob endpoint will succeed.
Any operations hitting dfs endpoint will fail with invalid URL escape "%,?"

For the fix, please ensure the same is tested for methods like NewFileClient too. 

- What did you expect or want to happen?
Hitting either the blob endpoint or the dfs endpoint should succeed.

- How can we reproduce it?
Instructions given above

- Anything we should know about your environment.
N/A

Related issue https://github.com/Azure/azure-storage-azcopy/issues/2549

gapra-msft:
@tanyasethi-msft We had another report of the issue, using the # symbol, so I want to make sure it is covered in the test scenarios. Linking the issue below. 

https://github.com/Azure/azure-storage-azcopy/issues/2559

PR #22308, opened 2024-01-30
[AzDatalake] Path escape names to file and directory client

tanyasethi-msft:


sdk/storage/azdatalake/filesystem/client.go
```diff
@@ -202,7 +203,15 @@ func (fs *Client) BlobURL() string {
 // The new directory.Client uses the same request policy pipeline as the Client.
 func (fs *Client) NewDirectoryClient(directoryPath string) *directory.Client {
 	directoryPath = strings.ReplaceAll(directoryPath, "\\", "/")
-	dirURL := runtime.JoinPaths(fs.generatedFSClientWithDFS().Endpoint(), directoryPath)
+	names := strings.Split(directoryPath, "/")
```
vibhansa-msft, line 206: As this code is being duplicated lets move this to another function and call that common method from all instances where we need to escape.

sdk/storage/azdatalake/filesystem/client.go
```diff
@@ -202,7 +203,15 @@ func (fs *Client) BlobURL() string {
...
+	path := ""
+	for i, name := range names {
```
vibhansa-msft, line 208: There is a sub-directory client also, just check if changes are needed there as well or not

sdk/storage/azdatalake/filesystem/client.go
```diff
@@ -202,7 +203,15 @@ func (fs *Client) BlobURL() string {
...
+		if i > 0 {
+			path += "/"
```
vibhansa-msft, line 210: In case of copy blob we fill the source path as well which goes as an query parameter, lets ensure encoding is correct there.

sdk/storage/azdatalake/filesystem/client.go
```diff
@@ -202,7 +203,15 @@ func (fs *Client) BlobURL() string {
...
+		}
+		path += url.PathEscape(name)
```
vibhansa-msft, line 212: Instead of doing string concat at each step, you can escape all parts and then join the string back with "/" as seperator. Code will look clearner with ath I beleive,

sdk/storage/azdatalake/filesystem/client_test.go
```diff
@@ -1318,6 +1319,59 @@ func (s *UnrecordedTestSuite) TestSASFileSystemClient() {
...
+
+	filesystemName := testcommon.GenerateFileSystemName(testName)
+	fsClient, err := testcommon.GetFileSystemClient(filesystemName, s.T(), testcommon.TestAccountDatalake, nil)
+	_require.NoError(err)
+	defer testcommon.DeleteFileSystem(context.Background(), _require, fsClient)
+
+	_, err = fsClient.Create(context.Background(), nil)
+	_require.NoError(err)
+
+	fileClient := fsClient.NewFileClient("#,%,?")
```
souravgupta-msft, line 1363: Can you please also add a test where the file path has / in it.

sdk/storage/azdatalake/filesystem/client_test.go
```diff
@@ -1318,6 +1319,59 @@ func (s *UnrecordedTestSuite) TestSASFileSystemClient() {
...
+	_require.NoError(err)
+
+	response, err := fileClient.Create(context.Background(), nil)
+	_require.NoError(err)
+	_require.NotNil(response)
+
+	owner := "4cf4e284-f6a8-4540-b53e-c3469af032dc"
+	_, err = fileClient.SetAccessControl(context.Background(), &file.SetAccessControlOptions{Owner: &owner})
```
souravgupta-msft, line 1371: I believe Create and SetAccessControl calls are going over dfs endpoint. Can you add a call for blob endpoint.

sdk/storage/azdatalake/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Escape paths for NewDirectoryClient and New FileClient in a file system
```
souravgupta-msft, line 10: You can mention the bug here, like
```suggestion
* Escape paths for NewDirectoryClient and New FileClient in a file system. Fixes [#22281](https://github.com/Azure/azure-sdk-for-go/issues/22281).
```

vibhansa-msft: *  [#22281](https://github.com/Azure/azure-sdk-for-go/issues/22281): Escape paths for NewDirectoryClient and New FileClient in a file system

sdk/storage/azdatalake/filesystem/client_test.go
tanyasethi-msft: added

sdk/storage/azdatalake/CHANGELOG.md
tanyasethi-msft: mentioned the bug in changelog, thanks!

sdk/storage/azdatalake/filesystem/client_test.go
```diff
@@ -1318,6 +1319,63 @@ func (s *UnrecordedTestSuite) TestSASFileSystemClient() {
...
+	owner := "4cf4e284-f6a8-4540-b53e-c3469af032dc"
+	_, err = dirClient.SetAccessControl(context.Background(), &file.SetAccessControlOptions{Owner: &owner})
+	_require.NoError(err)
+
+	// Perform an operation on blob endpoint
+	_, err = dirClient.SetMetadata(context.Background(), testcommon.BasicMetadata, nil)
+	_require.NoError(err)
+}
+
+func (s *RecordedTestSuite) TestCreateFileSystemFileClientWithSpecialFileName() {
```
gapra-msft, line 1351: can we also add a test getting a special file client from a directory client as well?

it might also be good to add encoding tests for all NewDirectoryClient/NewFileClient constructors

sdk/storage/azdatalake/filesystem/client_test.go
tanyasethi-msft: Added / in the same test

sdk/storage/azdatalake/filesystem/client.go
tanyasethi-msft: made the change

sdk/storage/azdatalake/filesystem/utils.go
```diff
@@ -0,0 +1,23 @@
...
+
+package filesystem
+
+import (
+	"net/url"
+	"strings"
+)
+
+// escapeSplitPaths is utility function to escape the individual strings by eliminating "/" in the path
+func escapeSplitPaths(filePath string) string {
```
souravgupta-msft, line 15: I think you should add this code in the `internal/shared/shared.go`.

sdk/storage/azdatalake/filesystem/client.go
tanyasethi-msft: Subdirectory client takes name of the directory instead of the path, so changes are not needed there

sdk/storage/azdatalake/filesystem/client.go
tanyasethi-msft: verified the above

sdk/storage/azdatalake/filesystem/client.go
tanyasethi-msft: moved changes to shared.go

sdk/storage/azdatalake/filesystem/client_test.go
tanyasethi-msft: Added tests for special file client from a directory client, encoding tests in all constructors.

souravgupta-msft (approved):

PR #22315, opened 2024-01-31
Update CODEOWNERS for ARM Advisor

jhendrixMSFT:
Updating per team's request.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -96,7 +96,7 @@
 #/<NotInRepo>/          @armleads-azure
 
 # ServiceLabel: %Advisor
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@jportugal0 please see the [instructions](https://github.com/Azure/azure-sdk-tools/blob/main/tools/codeowners-utils/README.md#user-is-not-a-public-member-of-azure) so we can get this cleaned up.

RickWinter:
@jhendrixMSFT Its been over a month and they haven't corrected the issue. Lets close this, they can make a PR for the change once they've fixed their account.

PR #22317, opened 2024-02-01
Replace ErrAuthenticationRequired with AuthenticationRequiredError

### LINKED ISSUE #22261: Include TokenRequestOptions with authentication required error
When `GetToken` returns an error because user interaction is required and the credential is configured not to prompt for it automatically, that error should include the `TokenRequestOptions` passed to `GetToken`. This would make `Authenticate` much easier to use by providing the application a simple way to request the necessary token. The calling pattern would look like this:
```go
cred, err := azidentity.NewInteractiveBrowserCredential(&azidentity.InteractiveBrowserCredentialOptions{
    DisableAutomaticAuthentication: true,
    TokenCachePersistenceOptions: &azidentity.TokenCachePersistenceOptions{},
})
client, err := NewServiceClient(cred)
err = client.Frobulate(context.TODO())
var authReqErr *azidentity.AuthenticationRequiredError
if errors.As(err, &authReqErr) {
    // User interaction required, app calls Authenticate when that's convenient.
    // Developer doesn't need to consider parameters because the given
    // TokenRequestOptions describes exactly the token the client requested.
    cred.Authenticate(context.TODO(), &authReqErr.TokenRequestOptions)
}
// TODO: retry Frobulate
```

chlowell:
This replaces the static error returned when a credential needs user interaction to acquire a token but is configured not to prompt for it automatically with a new error type that carries the argument to GetToken. This enables applications to call Authenticate with the same argument (see the new `ExampleAuthenticationRequiredError` for the pattern).

Closes #22261


sdk/azidentity/default_azure_credential_test.go
```diff
@@ -277,8 +277,8 @@ func TestDefaultAzureCredential_timeoutWrapper(t *testing.T) {
 	for i := 0; i < 2; i++ {
 		// expecting credentialUnavailableError because delay exceeds the wrapper's timeout
 		_, err = chain.GetToken(context.Background(), testTRO)
-		if _, ok := err.(*credentialUnavailableError); !ok {
-			t.Fatalf("expected credentialUnavailableError, got %T: %v", err, err)
+		if _, ok := err.(credentialUnavailable); !ok {
```
anlandu, line 280: Hi @chlowell! Is there a reason to leave the credentialUnavailable type unexported? Trying to implement a credentialErrorReporter wrapper for our custom ChainedTokenCredential but can't due to this line.

sdk/azidentity/default_azure_credential.go
```diff
@@ -158,7 +158,7 @@ type defaultCredentialErrorReporter struct {
 }
 
 func (d *defaultCredentialErrorReporter) GetToken(ctx context.Context, opts policy.TokenRequestOptions) (azcore.AccessToken, error) {
-	if _, ok := d.err.(*credentialUnavailableError); ok {
+	if _, ok := d.err.(credentialUnavailable); ok {
```
anlandu, line 161: Hi @chlowell! Is there a reason to leave the credentialUnavailable type unexported? Trying to implement a credentialErrorReporter wrapper for our custom ChainedTokenCredential but can't due to this line. Thank you!

chlowell: We didn't see a need to export the interface, and we generally prefer to keep public API surfaces as small as possible. What do you need `credentialUnavailable` for in your code? `defaultCredentialErrorReporter` doesn't need this line to function correctly.

anlandu: > defaultCredentialErrorReporter doesn't need this line to function correctly.
Got it, had figured I could just unconditionally new it but the presence of the additional type assertion check had me questioning myself. I guess it's in the defaultCredentialErrorReporter just to save unnecessary new-s. Thanks!

Issue #22156, opened 2023-12-18
DATA Race when downloading blob item from storage container


Todorov99:
### DATA Race when downloading blob item from storage container

- Imported package github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container
- SDK version: 1.2.0
- Go Version 1.21.5

Running one of our tests with the race detector flag (ginkgo --race ...) reported this data race issue:

```
==================
WARNING: DATA RACE
Read at 0x00c001c719c8 by goroutine 1206:
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared.(*mmbPool).Grow()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared/buffer_manager.go:49 +0x3c
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob.(*Client).downloadFile.func2()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob/client.go:474 +0x7c
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob.(*Client).downloadFile.func4()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob/client.go:524 +0xb0
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared.DoBatchTransfer.func2()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared/batch_transfer.go:65 +0x9c
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared.DoBatchTransfer.func1()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared/batch_transfer.go:50 +0x4b

Previous write at 0x00c001c719c8 by goroutine 1208:
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared.(*mmbPool).Grow()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/shared/buffer_manager.go:55 +0xfc
  github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob.(*Client).downloadFile.func2()
      vendor/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob/client.go:474 +0x7c
[...]

PR #22334, opened 2024-02-02
Fix azblob/download data race

tanyasethi-msft:
In Download File API, method aquireBuffer() in https://github.com/Azure/azure-sdk-for-go/commit/1dc804a987bde7cc943c339b0b374f98c99d30e1 is not thread safe. 
If 2 threads acquire the buffer and both try to increment the _count_ of buffers allocated at the same time, this leads to a Data Race.  

### DIFF
--- a/sdk/storage/azblob/blob/client.go
+++ b/sdk/storage/azblob/blob/client.go
@@ -9,6 +9,7 @@ package blob
 import (
 	"context"
 	"io"
+	"math"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


vibhansa-msft (approved):

souravgupta-msft (approved):

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/22156

sdk/storage/azblob/blob/client.go
```diff
@@ -525,7 +521,7 @@ func (b *Client) downloadFile(ctx context.Context, writer io.Writer, o downloadO
 		OperationName: "downloadBlobToWriterAt",
 		TransferSize:  count,
 		ChunkSize:     o.BlockSize,
-		NumChunks:     numChunks,
+		NumChunks:     uint16(numChunks),
```
vibhansa-msft, line 524: We are converting the type of this variable twice. In the first loop to float64 and then here to uint16. While assigning the value itself if we convert to uint16 then this can be avoided. Older code which is removed was doing the same I see.

tanyasethi-msft: Updated to reduce type conversion.

PR #22357, opened 2024-02-08
[azblob] Fix file offset post sequential write in downloadFile

tanyasethi-msft:

### DIFF
--- a/sdk/storage/azblob/blob/client.go
+++ b/sdk/storage/azblob/blob/client.go
@@ -596,6 +596,8 @@ func (b *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFil
 	}
 	do := (*downloadOptions)(o)
 
+	filePointer, err := file.Seek(0, io.SeekCurrent)
+
 	// 1. Calculate the size of the destination file
 	var size int64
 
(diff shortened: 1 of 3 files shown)


sdk/storage/azblob/blob/client.go
```diff
@@ -596,6 +596,8 @@ func (b *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFil
 	}
 	do := (*downloadOptions)(o)
 
+	filePointer, _ := file.Seek(0, io.SeekCurrent)
```
souravgupta-msft, line 599: We should check the error here

sdk/storage/azblob/blob/client.go
```diff
@@ -624,7 +626,9 @@ func (b *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFil
 	}
 
 	if size > 0 {
-		return b.downloadFile(ctx, file, *do)
+		writeSize, err := b.downloadFile(ctx, file, *do)
+		_, _ = file.Seek(filePointer, 0)
```
souravgupta-msft, line 630: Add error check here. We can also do something like,
```suggestion
		_, _ = file.Seek(filePointer, io.SeekStart)
```

souravgupta-msft (approved):

PR #22368, opened 2024-02-10
[azservicebus] Fixing a memory leak when doing cross receiver settlement.

### LINKED ISSUE #22318: [azservicebus] Memory leak when receiving messages
### Bug Report

- Package in question azure-sdk-for-go/sdk/messaging/azservicebus
- SDK version v1.6.0
- go version 1.21

### What happened?

This is a continuation of https://github.com/Azure/azure-sdk-for-go/issues/22157.  Trying to move out of https://github.com/Azure/azure-service-bus-go. Same usage of `azservicebus` but this time with higher load in production setup.  

We are noticing a memory leak as you can see on the screenshots below. 

Screenshot 2024-01-31 at 22 52 51

Service running for a bit more than a day, we can see there is a an increase of memory happening periodically.  Looking at profiling data looks like the culprit is `(*Session).mux`
Screenshot 2024-01-31 at 22 47 07

Our Service Bus instance was affected with a huge increase of memory consumption jumping from a steady `17%`  up to `86%` and would kept raising if we hadn't rolled back. 

Screenshot 2024-01-31 at 23 07 59

even though there was a noticeable improvement in reducing number of active connections

Screenshot 2024-01-31 at 23 20 58

### What did you expect or want to happen?

I would expect the new SDK to have a different memory footprint (either higher or lower memory consumption) but after some time i was expecting the memory usage of the service to be stable as before and not increasing time over time.

Also it was not expected that the Service Bus instance will be affected.

### How can we reproduce it?

Not sure how easy it is to be reproduced as we needed production load to notice this. But outlining the setup here so it might make troubleshooting easier. 

1. We have a **fixed** number of producer goroutines that are receiving messages and a number of consumer goroutines that process messages and communicate via a go channel. At the start of the service we create a pool of receivers client.NewReceiverForQueue for receiving messages in parallel. 
2. A message will never be received and settled by the same receiver (this was causing the issues before)
3. Senders/receivers are long lived
4. Typical message size is ~3Kib and max message can be up to 256Kib.

richardpark-msft:
go-amqp holds onto some tracking data that won't get cleared if we don't try to settle through the original Receiver. Our fix in here, combined with go-amqp's changes to route to the original receiver, should seal that up.

Benchmark added that also doubles as a stress test.

Fixes #22318

### DIFF
--- a/sdk/messaging/azservicebus/messageSettler.go
+++ b/sdk/messaging/azservicebus/messageSettler.go
@@ -5,6 +5,7 @@ package azservicebus
 
 import (
 	"context"
+	"errors"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal"
 	"github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus/internal/amqpwrap"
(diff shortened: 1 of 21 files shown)


sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -62,13 +53,22 @@ type CompleteMessageOptions struct {
...
+func (ms *messageSettler) CompleteMessage(ctx context.Context, message *ReceivedMessage, options *CompleteMessageOptions) error {
+	return ms.settleWithRetries(ctx, message, func(receiver amqpwrap.AMQPReceiver, rpcLink amqpwrap.RPCLink) error {
+		var err error
+
+		if shouldSettleOnReceiver(message) {
+			ms.notifySettleOnLink(message)
+			err = receiver.AcceptMessage(ctx, message.RawAMQPMessage.inner)
 		}
+
+		if shouldSettleOnMgmtLink(err, message) {
```
jhendrixMSFT, line 65: Does this intentionally fall through from the previous clause? If so, might be good to comment why.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -81,9 +81,29 @@ type AbandonMessageOptions struct {
...
+			}
+
+			err = receiver.ModifyMessage(ctx, message.RawAMQPMessage.inner, &amqp.ModifyMessageOptions{
+				DeliveryFailed:    false,
+				UndeliverableHere: false,
+				Annotations:       annotations,
+			})
+		}
+
+		if shouldSettleOnMgmtLink(err, message) {
```
jhendrixMSFT, line 104: Same fall-through question.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -119,9 +129,30 @@ type DeferMessageOptions struct {
...
+
+			err = receiver.ModifyMessage(ctx, message.RawAMQPMessage.inner,
+				&amqp.ModifyMessageOptions{
+					DeliveryFailed:    false,
+					UndeliverableHere: true,
+					Annotations:       annotations,
+				})
+		}
+
+		if shouldSettleOnMgmtLink(err, message) {
```
jhendrixMSFT, line 153: Same fall-through question.

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -181,7 +201,33 @@ func (s *messageSettler) DeadLetterMessage(ctx context.Context, message *Receive
...
+
+			amqpErr := amqp.Error{
+				Condition: "com.microsoft:dead-letter",
+				Info:      info,
+			}
+
+			err = receiver.RejectMessage(ctx, message.RawAMQPMessage.inner, &amqpErr)
+		}
+
+		if shouldSettleOnMgmtLink(err, message) {
```
jhendrixMSFT, line 228: Same fall-through question.

richardpark-msft: I added a comment in all four locations.

sdk/messaging/azservicebus/messageSettler.go
richardpark-msft: I added a comment in all four locations.

jhendrixMSFT (approved):

Issue #18987, opened 2022-08-30
Cosmos DB: SessionContainer


ealsur:
Add SessionContainer to help with accounts with Session Consistency.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/SessionContainer.cs
```[tasklist]
### Tasks
- [ ] Adding hashing support
- [ ] Add container properties cache (used to read the partition key path and apply hashing)
- [ ] Add partition key ranges cache (used to identify which partition the request goes to after applying hashing)
```

jay-most:
Notes: Time 1-2 weeks | T-shirt S | add session token to the header response. session cache. same session reads and writes. in memory cache.

PR #22391, opened 2024-02-14
[Cosmos DB] Add session container

zarenner:
Implements https://github.com/Azure/azure-sdk-for-go/issues/18987

/cc @ealsur @jay-most 

### DIFF
--- a/sdk/data/azcosmos/cosmos_session_policy.go
+++ b/sdk/data/azcosmos/cosmos_session_policy.go
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import (
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+type sessionPolicy struct {
+	sc interface {
+		GetSessionToken(resourceAddress string) string
+		SetSessionToken(resourceAddress string, containerRid string, sessionToken string)
+		ClearSessionToken(resourceAddress string)
+	}
+}
+
+func (p *sessionPolicy) Do(req *policy.Request) (*http.Response, error) {
+	o := pipelineRequestOptions{}
+	if !req.OperationValue(&o) || req.Raw().Header.Get(cosmosHeaderSessionToken) != "" {
+		return req.Next()
+	}
+
+	// Reads: use session token if available
+	if !o.isWriteOperation {
+		sessionToken := p.sc.GetSessionToken(o.resourceAddress)
+		if sessionToken != "" {
+			req.Raw().Header.Set(cosmosHeaderSessionToken, sessionToken)
+		}
+	}
+
+	response, err := req.Next()
+	if err != nil || !(response.StatusCode >= 200 && response.StatusCode < 300) {
+		// Allow potential retry attempt without session token
+		p.sc.ClearSessionToken(o.resourceAddress)
+		return response, err
+	}
+
+	// Successful Writes: cache session token
+	if o.isWriteOperation && response.StatusCode >= 200 && response.StatusCode < 300 {
+		sessionToken := response.Header.Get(cosmosHeaderSessionToken)
+		containerPath := response.Header.Get(cosmosHeaderAltContentPath)
+		containerRid := response.Header.Get(cosmosHeaderContentPath)
+
+		// We currently expect the container path and RID headers to be present
+		// (rather than falling back to o.resourceAddress and parsing _self)
+		if sessionToken != "" && containerPath != "" && containerRid != "" {
+			p.sc.SetSessionToken(containerPath, containerRid, sessionToken)
+		}
+	}
+
+	return response, nil
+}
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.3.7 (Unreleased)
 
 ### Features Added
+* Session consistency tokens are automatically managed when `EnableSessionConsistency` is set
 
 ### Breaking Changes
 
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -49,7 +49,8 @@ func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*C
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newSharedKeyCredPolicy(cred), gem, o), gem: gem}, nil
+
+	return &Client{endpoint: endpoint, pipeline: newPipeline(newSharedKeyCredPolicy(cred), gem, newSessionContainer(), o), gem: gem}, nil
 }
 
 // NewClient creates a new instance of Cosmos client with Azure AD access token authentication. It uses the default pipeline configuration.
@@ -70,7 +71,7 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 		return nil, err
 	}
 
-	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), gem, o), gem: gem}, nil
+	return &Client{endpoint: endpoint, pipeline: newPipeline(newCosmosBearerTokenPolicy(cred, scope, nil), gem, newSessionContainer(), o), gem: gem}, nil
 }
 
 // NewClientFromConnectionString creates a new instance of Cosmos client from connection string. It uses the default pipeline configuration.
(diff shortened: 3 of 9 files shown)


sdk/data/azcosmos/cosmos_session_policy.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	// Successful Writes: cache session token
+	if o.isWriteOperation && response.StatusCode >= 200 && response.StatusCode < 300 {
+		sessionToken := response.Header.Get(cosmosHeaderSessionToken)
+		containerPath := response.Header.Get(cosmosHeaderAltContentPath)
+		containerRid := response.Header.Get(cosmosHeaderContentPath)
+
+		// We currently expect the container path and RID headers to be present
+		// (rather than falling back to o.resourceAddress and parsing _self)
```
zarenner, line 48: I don't know if this is a safe assumption - other clients (Python, JS) have various behavior in terms of expecting both `X-Ms-Alt-Content-Path` and `X-Ms-Content-Path`. And the latter isn't documented at https://learn.microsoft.com/en-us/rest/api/cosmos-db/common-cosmosdb-rest-response-headers

It _seems_ like both are always set (at least for name-based requests) when a response includes a session token, but I wonder if this should be (falling back to?) getting the container rid by parsing `_self` and getting the container path by falling back to o.resourceAddress.

ealsur:
@zarenner Thank you for this contribution. Can you please clarify the proposed design? What are you basing the work on? 

There is a reason the Session Container as a feature is, in order, after other work: https://github.com/orgs/Azure/projects/103/views/13 as it requires things from the previous work (such as, detecting if the account has Session consistency or not). Session Container should not be an opt-in, it should be driven by the fact that the account is Session and the request is not downgrading consistency.

zarenner:
Hi @ealsur - generally, I based this off of the behaviors of the Python and JS clients.

w.r.t. enabling the session container:
* The Python library appears to enable session container automatically based on the account default consistency level, but also via [user-provided consistency level set on the client](https://github.com/Azure/azure-sdk-for-python/blob/f412402734d6535d6863eed4ed2482cbc3cefbf0/sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py#L250-L259) regardless of actual account level.
* The JS library appears to enable session container based on the [consistency level request header](https://github.com/Azure/azure-sdk-for-js/blob/c787e5a9d30c85d5369bfb0e73800de601351420/sdk/cosmosdb/cosmos/src/ClientContext.ts#L505-L514) (either set at client or individual request). I didn't see logic that automatically determines it based on the account default, but it's entirely possible I missed it.

> Session Container should not be an opt-in, it should be driven by the fact that the account is Session and the request is not downgrading consistency.

Totally agree, but given the above it seemed reasonable to me to (for now) allow enabling it as opt-in in some way. We have teams who need session consistency support, and have either unnecessarily used stronger consistency levels as a workaround or implemented custom logic. Given that and that the issue has been open for quite a long time, I figured some form of opt-in support was better than nothing for those needing the feature.

If there's a different (reasonably self contained) enablement mechanism that would be more appealing I'm happy to implement it. e.g. via a new `ConsistencyLevel` override at the client level like in the Python lib, or reading the account's level once for the lifetime of the client, etc.

Or if you'd prefer to hold off on this feature for now, that's of course fine too and I'll fork and/or implement custom logic around the client for our company's scenarios.

ealsur:
I'm not familiar with the Python/JS code exactly (I do know Java and .NET), the Session Container implementation proposed is missing some fundamental aspects:

* There is no logic to handle Container deletion/recreation: A container can be deleted and recreated with the same name, this changes the RID and invalidates all tokens.
* When sending requests, we cannot add all the Session Tokens, it will cause HTTP 400 responses from the service when the combined header size grows beyond the limit: Other SDKs can identify which partition the request goes to and only send that token.
* Overrides: There should be a `ConsistencyLevel` option on the `ClientOptions` to define the Client level. There is already one at the item level (https://github.com/Azure/azure-sdk-for-go/blob/b5a231a1404e4e72368ad20b63233db1cf6226fd/sdk/data/azcosmos/cosmos_item_request_options.go#L28)
* Clearing container: In the PR, the Session Container is cleaned on a non-successful response, but that would make a 429 (throttle for example) clean it. Because 429s are retried, then the retry after would have no token and thus reduced to Eventual consistency. 
* Write operations: Session Token should be sent on Write operations if the account is MultiMaster
* Partition Id: The Partition Id is part of the item response in a header, there is no need to parse the token to identify it. Parsing the token also avoids any future service changes if they decide to for some reason, alter the format. From the client perspective, it should be a string without complex attempts of interpretation.

> Totally agree, but given the above it seemed reasonable to me to (for now) allow enabling it as opt-in in some way. We have teams who need session consistency support, and have either unnecessarily used stronger consistency levels as a workaround or implemented custom logic. Given that and that the issue has been open for quite a long time, I figured some form of opt-in support was better than nothing for those needing the feature.

The SessionContainer is not required to provide session consistency. Customers can use Session consistency by obtaining the Session Token from the write operation and passing it to the Read operation (basically doing what the Container in a general sense does). If a customer wants to save the token in-memory or in other means to follow the user session (read their writes), that's up to the user implementation and is achievable today.

The problem with adding APIs is that once they are added and we GA the package, then we are stuck with them, we cannot add APIs/features that are stop-gap solutions to later do the correct way. I'm not against community contributions (we had PRs in the past with bugfixes or small features), it's just that these core components require a full understanding of the SDK.

zarenner:
> There is no logic to handle Container deletion/recreation: A container can be deleted and recreated with the same name, this changes the RID and invalidates all tokens.

That [logic](https://github.com/Azure/azure-sdk-for-go/pull/22391/files#diff-b5494ccb5e79ca4828b077aea434c10e94677db6e510f659aa00e463482cf15eR86) and [test](https://github.com/Azure/azure-sdk-for-go/pull/22391/files#diff-bbec07e3afc9d8206feee8520f4bacc9ea34eed6aa4ecd7f2c349a8fca5d4e31R56) is in this PR (at least for changes detected on writes).

> When sending requests, we cannot add all the Session Tokens, it will cause HTTP 400 responses from the service when the combined header size grows beyond the limit: Other SDKs can identify which partition the request goes to and only send that token.

That makes sense given large partition counts, although the [Python](https://github.com/Azure/azure-sdk-for-python/blob/c97bb56ca3e657ca58eb972fa7b0db1443f48c50/sdk/cosmos/azure-cosmos/azure/cosmos/_session.py#L66) and [JS](https://github.com/Azure/azure-sdk-for-js/blob/70865b9adcc7057b58c1a520174655804f327b60/sdk/cosmosdb/cosmos/src/session/sessionContainer.ts#L103C18-L103C47) SDKs do exclusively use combined tokens. Regardless, I considered implementing a partition key range -> id map, but the documentation on `x-ms-documentdb-partitionkeyrangeid` somewhat discourages its use and there were parts of the .NET library logic around maintaining this cache that I was unsure were required or optimizations so I followed Python/JS.

> Overrides: There should be a ConsistencyLevel option on the ClientOptions to define the Client level. 

Agree

> Clearing container: In the PR, the Session Container is cleaned on a non-successful response, but that would make a 429 (throttle for example) clean it. Because 429s are retried, then the retry after would have no token and thus reduced to Eventual consistency.

Yeah I considered this, but such logic wasn't obvious to me on the JS client. Looking again at the Python client, it does have that logic. Agree this should be fixed.

> Session Token should be sent on Write operations if the account is MultiMaster

I don't believe the Python library does so. Conversely the JS library AFAICT sends them on writes even if not multimaster.

> Partition Id: The Partition Id is part of the item response in a header, there is no need to parse the token to identify it. Parsing the token also avoids any future service changes if they decide to for some reason, alter the format. From the client perspective, it should be a string without complex attempts of interpretation.

Again was unsure on availability of `x-ms-documentdb-partitionkeyrangeid` based on docs, and so I took the approach of parsing out the partition exactly as the Python and JS libraries do.

> it's just that these core components require a full understanding of the SDK.

Agree, and that's unfortunate IMO. The Python and JS libraries are considered production ready AFAIK and yet still are apparently a poor reference given they have some of the flaws you describe, so even a good understanding of them is insufficient. Libraries also rely on undocumented response headers and internal concepts (e.g. rid format) and in general there's no public docs / specs I could find on how clients are supposed to behave. All that makes it difficult, as you say, for the community to contribute anything more than trivial fixes.

Again, I'm happy to make suggested improvements and help continue this work but it sounds like it may take enough of your time to not be worth the squeeze right now, and that's fine. For now I'll close out this PR and reimplement it outside the SDK.

ealsur:
Let me first say Thank You. Because you made a fantastic effort to reverse-engineer and understand what was done in other SDKs in order to create this PR. There are nuisances about features like this that are not publicly documented (or if they are, are poorly publicly documented).

> That [logic](https://github.com/Azure/azure-sdk-for-go/pull/22391/files#diff-b5494ccb5e79ca4828b077aea434c10e94677db6e510f659aa00e463482cf15eR86) and [test](https://github.com/Azure/azure-sdk-for-go/pull/22391/files#diff-bbec07e3afc9d8206feee8520f4bacc9ea34eed6aa4ecd7f2c349a8fca5d4e31R56) is in this PR (at least for changes detected on writes).

What I mean about clearing is, that there is a specific service response that signals that the request that was sent was not valid. The code you propose is after the response is received but the problem is on the request. If a request is sent with the Session Token of the "previous" Container Rid, it can fail the request forever or return a non-expected response (for example, not be able to read the write because the LSN in the token was less than the LSN in the new Container).

Reference on .NET: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/cba93d3fad6be84a59ddb4a920af990dff6a6fa4/Microsoft.Azure.Cosmos/src/RenameCollectionAwareClientRetryPolicy.cs#L84-L87

This is not obvious and I would not expect a public contribution to know this.

> All that makes it difficult, as you say, for the community to contribute anything more than trivial fixes.

I completely agree. My humble opinion is that there are some core functionalities that relies on service specifics that are poorly or not documented.

We do encourage community contributions but there are areas where the effort to onboard a community contribution means ramping up on deeper knowledge. I believe that it would be good to be clear which features / Issues are good for external contributors and which are better handled by the team. We were discussing this with @jcocchi, maybe we need to add proper

ealsur:
@zarenner We'll keep the PR for reference, there is a good chance of reusing part of it and I'd like to call out your contribution when we get to it

Issue #18985, opened 2022-08-30
Cosmos DB: Add ClientRetryPolicy for cross-region failover


ealsur:
Add ClientRetryPolicy that leverages GlobalEndpointManager to do cross region failover

Reference: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs or https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/ClientRetryPolicy.java (the same policy exists in other languages, can use Python or NodeJS also)

Requires https://github.com/Azure/azure-sdk-for-go/issues/18983

Requirements:

* Create a Pipeline Policy that executes `PerRetry` (see `newPipeline` on `cosmos_client.go`).
* The policy should cover the ClientRetryPolicy scenarios:
  * Before the request is sent, use the request information (is it a write or read requests? Any previous retries? See the RetryContext on other languages) to decide which Location to use from the GlobalEndpointManager (gem should be available on the client and can be passed to the new policy)
  * Verify existing retry count limits on existing ClientRetryPolicy implementations
  * Follow ClientRetryPolicy cases from references, which include :
    * On a 403.3 response, mark the current region as unavailable for writes, refresh the account information, and retry
    * On a 403.1008 response, mark the current region as unavailable for reads, refresh the account information, and retry
    * On HTTP 503 response, if it's a read request and preferredRegions > 1, retry on the next preferredRegion. If it's a write request and account is multi master and preferredRegions > 1, retry on the next preferredRegion.
    * On 404/1002 retry on the primary region (session consistency guarantee) once
    * On HTTP timeouts (Verify how HTTP timeouts surface in Golang), retry on read requests on the next preferredRegion if preferredRegions > 1

jay-most:
Notes: Time 2 weeks | T-Shirt  S  | -> failure cases, for the gem.  403 errors for example and mark unavailable.

fabistb:
Any update on this?
With the increasing popularity of dapr the number of indirect sdk users will increase.
We recently had a an outage in West Europe and since the fail over didn't work we ended up in an error state for multiple hours.

ealsur:
@fabistb The team is still working on it.

[...]

PR #22394, opened 2024-02-15
[Cosmos] Implements Client Retry policy

simorenoh:
Reference https://github.com/Azure/azure-sdk-for-go/issues/18985.

**Currently still finalizing the tests to ensure we are hitting all the scenarios we want to hit - pushing this PR pre-emptively since the implementation is pretty much done. Please ignore any comments/ print statements I am currently using for testing.**

This PR in it's entirety will include:

Client Retry policy implementation covering:
- [x] Session retry scenarios (404-1002)
- [x] Endpoint unavailable retry scenarios (403-1002/8)
- [x] Service unavailable retry scenarios (503)
- [x] DNS error scenarios
- [x] Tests covering all of the above scenarios
- [x] Tests covering external scenarios like connections dropping (DNS errors)
---------------------------------------------------
- [x] New function added to `location_cache` to resolve service endpoint to be used on cross-regional retries
- [x] Tests for the above new function
----------------------------------------------------
- [ ] New public surface area client option, `DisableCrossRegionalRetries` (known as `enable_endpoint_discovery` in most other SDKs - have to use a negative here for boolean default in Go).  -> this will be made into a separate PR
-------------------------------------------------------
- [x] Upgrade in `go.mod` file from go 1.18 to go 1.22 and update `ci.yml` - 1.21 is the earliest version containing a Go-native `min()` method and the ci pipelines test current version, N-1 and N-2


sdk/data/azcosmos/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos
 
-go 1.18
+go 1.21
```
jhendrixMSFT, line 3: This will fail CI as we run versions N, N-1 and 1.18 (N-2) as our baseline. We need to get 1.22 into CI which will resolve the first two parts. For N-2, you'll need to update your CI file to exclude it. See example [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/tracing/azotel/ci.yml#L28).

jhendrixMSFT: FYI CI has been updated. If you rebase/merge latest you'll see 1.22 and 1.21 as N and N-1 versions.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -42,10 +42,11 @@ func (c *Client) Endpoint() string {
 // options - Optional Cosmos client options.  Pass nil to accept default values.
 func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*Client, error) {
 	preferredRegions := []string{}
+	enableEndpointDiscovery := true
```
ealsur, line 45: nit: lets rename this, `enableEndpointDiscovery` melts the mind of anyone trying to understand

sdk/data/azcosmos/cosmos_client_options.go
```diff
@@ -15,4 +15,6 @@ type ClientOptions struct {
 	EnableContentResponseOnWrite bool
 	// PreferredRegions is a list of regions to be used when initializing the client in case the default region fails.
 	PreferredRegions []string
+	// When DisableCrossRegionalRetries is true the client will not apply the client retry policy, and will fail requests that would otherwise utilize cross-regional retries to attempt successful status.
+	DisableCrossRegionalRetries bool
```
ealsur, line 19: No public API changes

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -26,7 +27,7 @@ type globalEndpointManager struct {
 	lastUpdateTime      time.Time
 }
 
-func newGlobalEndpointManager(clientEndpoint string, pipeline azruntime.Pipeline, preferredLocations []string, refreshTimeInterval time.Duration) (*globalEndpointManager, error) {
+func newGlobalEndpointManager(clientEndpoint string, pipeline azruntime.Pipeline, preferredLocations []string, refreshTimeInterval time.Duration, enableEndpointDiscovery bool) (*globalEndpointManager, error) {
```
ealsur, line 30: nit: lets rename this, `enableEndpointDiscovery` melts the mind of anyone trying to understand

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -124,6 +126,7 @@ func newPipeline(authPolicy policy.Policy, gem *globalEndpointManager, options *
 			},
 			PerRetry: []policy.Policy{
 				authPolicy,
+				&clientRetryPolicy{gem: gem},
```
ealsur, line 129: Should the flag be wired here?

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+	preferredLocationIndex int
+}
+
+const maxRetryCount = 10 // this number needs to be higher, keeping low for testing purposes right now
+const defaultBackoff = 1
+
+func (p *clientRetryPolicy) Do(req *policy.Request) (*http.Response, error) {
+	response, err := req.Next() // err can happen in weird scenarios (connectivity, etc) - need to test
+	o := pipelineRequestOptions{}
+	req.OperationValue(&o)
```
ealsur, line 27: What is this call for?

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+
+	return response, err
+}
+
+func (p *clientRetryPolicy) shouldRetryStatus(status int, subStatus string) (shouldRetry bool) {
+	if (status == 403 && (subStatus == "3" || subStatus == "1008")) ||
+		(status == 404 && subStatus == "1002") ||
+		(status == 403 && subStatus == "3") ||
```
ealsur, line 68: This is repeated from the first line

sdk/data/azcosmos/cosmos_client_options.go
simorenoh: are we enabling this by default for everyone then?

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: I got this from the header policy, but it populates the previously empty pipelineRequestOptions with those that are present in the request

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: good catch, thanks!

sdk/data/azcosmos/cosmos_client.go
simorenoh: in other SDKs we generally pass the option forward to the GEM, but we technically only use it within the policy so we could

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+			p.sessionRetryCount = 0
+			break
+		}
+	}
+
+	return response, err
+}
+
+func (p *clientRetryPolicy) shouldRetryStatus(status int, subStatus string) (shouldRetry bool) {
+	if (status == 403 && (subStatus == "3" || subStatus == "1008")) ||
```
ealsur, line 66: nit: can we make the substatus be some const variables somewhere?

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+		return false
+	}
+	if isWriteOperation {
+		p.gem.MarkEndpointUnavailableForWrite(*req.Raw().URL)
+	} else {
+		p.gem.MarkEndpointUnavailableForRead(*req.Raw().URL)
+	}
+	p.gem.Update(req.Raw().Context())
+
+	resolvedEndpoint := p.gem.ResolveServiceEndpoint(p.retryCount, isWriteOperation)
```
ealsur, line 86: The ResolveServiceEndpoint call should be outside of the ifs to reduce duplicity

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+	req.Raw().Host = resolvedEndpoint.Host
+	req.Raw().URL.Host = resolvedEndpoint.Host
+	p.retryCount += 1
+	time.Sleep(defaultBackoff * time.Second)
```
ealsur, line 90: There is no backoff that I'm aware of for 403 handling

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+			endpoints = p.gem.locationCache.locationInfo.availReadLocations
+		}
+		if p.sessionRetryCount >= len(endpoints) {
+			return false
+		}
+		resolvedEndpoint := p.gem.ResolveServiceEndpoint(p.retryCount, isWriteOperation)
+		req.Raw().Host = resolvedEndpoint.Host
+		req.Raw().URL.Host = resolvedEndpoint.Host
+	} else {
+		if p.sessionRetryCount > 0 {
```
ealsur, line 109: For single master, how are we forcing the request to go to Write region?

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -50,20 +50,20 @@ type locationCache struct {
...
 }
 
-func newLocationCache(prefLocations []string, defaultEndpoint url.URL) *locationCache {
+func newLocationCache(prefLocations []string, defaultEndpoint url.URL, enableEndpointDiscovery bool) *locationCache {
 	return &locationCache{
 		defaultEndpoint:                   defaultEndpoint,
 		locationInfo:                      *newDatabaseAccountLocationsInfo(prefLocations, defaultEndpoint),
 		locationUnavailabilityInfoMap:     make(map[url.URL]locationUnavailabilityInfo),
 		unavailableLocationExpirationTime: defaultExpirationTime,
+		enableEndpointDiscovery:           enableEndpointDiscovery,
```
ealsur, line 66: Why is this flag here if this is something used in the client retry policy? Can we not pass it as the policy constructor?

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if isWriteOperation {
+		if p.gem.CanUseMultipleWriteLocations() {
+			locationalEndpoint := p.gem.GetPreferredLocationEndpoint(p.preferredLocationIndex, *req.Raw().URL)
+			req.Raw().URL = &locationalEndpoint
+		} else {
+			return false
+		}
+	} else {
+		locationalEndpoint := p.gem.GetPreferredLocationEndpoint(p.preferredLocationIndex, *req.Raw().URL)
```
ealsur, line 132: Why can't we use `p.gem.ResolveServiceEndpoint(p.retryCount, isWriteOperation)`?

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: I see it both in Java:
image

And in Python:
image

Although admittedly it is used differently in both - in Python we always do it, in Java it seems like only for read requests or if it's not our first write request retry. We can remove it if you'd like.

sdk/data/azcosmos/cosmos_location_cache.go
simorenoh: I was just aligning to other SDKs:
Java:
image

Python:
image


will make it a policy-only thing though

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: we can, but it is not the same - if a user doesn't supply preferred regions then it will not be populated. 

I think this might have been a gap on my end though, from reading the description on 503 retries I somehow figured we would be needing to use the actual user-supplied preferred regions, but I see in the different SDKs we just use a flag and then use resolveServiceEndpoint

sdk/data/azcosmos/cosmos_client_options.go
ealsur: The PR is enabling this by default, right? What I'm saying is that we can defer the public API name and public API review to another PR.

sdk/data/azcosmos/cosmos_client.go
ealsur: I see, because we use the flag within the GEM/LocationCache

sdk/data/azcosmos/cosmos_client_retry_policy.go
ealsur: https://github.com/Azure/azure-sdk-for-go/blob/10dd3c522da479beb9dc1330c6863b7cebcc6e18/sdk/data/azcosmos/cosmos_headers_policy.go#L24-L27

The header policy is using this to **extract** the values from the Request. You need to test if this returns `true` before using any values

sdk/data/azcosmos/cosmos_client_retry_policy.go
ealsur: You are right, let's leave it

sdk/data/azcosmos/cosmos_location_cache.go
ealsur: If we are using it in the location cache for something, then leave it. I'm just not sure if we area

sdk/data/azcosmos/cosmos_client_retry_policy.go
ealsur: ResolveServiceEndpoint uses the location cache, location cache Write/Read Endpoints are populated taking into account the preferred regions:

https://github.com/Azure/azure-sdk-for-go/blob/10dd3c522da479beb9dc1330c6863b7cebcc6e18/sdk/data/azcosmos/cosmos_location_cache.go#L97-L98

ealsur:
I just realized we are missing something critical. We are never "initializing" the SDK. The first operation that comes needs to "initialize" the SDK by performing a request to GEM to obtain the account information, that will initialize the location cache.

Otherwise, requests cannot be routed.

We can certainly leave that part for another PR, but it needs to happen only for the first request (like other SDKs)

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: ok makes sense - what should we do in case the returned boolean is false? if we default `isWriteOperation` to false then the current logic could stand no? or should we instead make the logic ourselves and determine if it is a write request based on the request method + parsing of the url?

simorenoh:
@ealsur On the initialization bit, I thought that was taken care of by the gemPolicy? when we initialize the gem that gets passed to the gem policy we set it's lastUpdateTime to time.Time{} (beginning of time), which ensures the first request in the pipeline will run the Update() method in the gemPolicy, populating the account informaiton in the gem/location cache

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -124,6 +125,7 @@ func newPipeline(authPolicy policy.Policy, gem *globalEndpointManager, options *
...
```
ealsur, line 128: Question: Should this policy be in the `PerCall` section instead of the `PerRetry`?

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,129 @@
...
+	retryCount             int
+	sessionRetryCount      int
+	preferredLocationIndex int
+}
+
+const maxRetryCount = 10 // this number needs to be higher, keeping low for testing purposes right now
+const defaultBackoff = 1
+
+func (p *clientRetryPolicy) Do(req *policy.Request) (*http.Response, error) {
+	response, err := req.Next() // err can happen in weird scenarios (connectivity, etc) - need to test
```
ealsur, line 25: Remember to include the `err` for the DNS error that Joel shared? https://pkg.go.dev/net#DNSError

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,129 @@
...
+	o := pipelineRequestOptions{}
+	req.OperationValue(&o)
+	subStatus := response.Header.Get(cosmosHeaderSubstatus)
+	if p.shouldRetryStatus(response.StatusCode, subStatus) {
+		p.retryCount = 0
+		p.sessionRetryCount = 0
+		for {
+			resolvedEndpoint := p.gem.ResolveServiceEndpoint(p.retryCount, o.isWriteOperation)
```
ealsur, line 33: This should be before the `response, err := req.Next()`.

1. Resolve endpoint
2. Execute 
3. if shouldRetry
4. Mark endpoints unavailable
5. Resolve endpoint again (this time the endpoints are marked)
6. Retry request

sdk/data/azcosmos/cosmos_client_retry_policy.go
simorenoh: yes I just haven't gotten to this, finishing the main test scenarios first

ealsur:
> @ealsur On the initialization bit, I thought that was taken care of by the gemPolicy? when we initialize the gem that gets passed to the gem policy we set it's lastUpdateTime to time.Time{} (beginning of time), which ensures the first request in the pipeline will run the Update() method in the gemPolicy, populating the account information in the gem + location cache

GEM policy runs on the background with a go routine, right? It does not wait.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -135,7 +135,7 @@ func newInternalPipeline(authPolicy policy.Policy, options *ClientOptions) azrun
 	if options == nil {
 		options = &ClientOptions{}
 	}
-	return azruntime.NewPipeline("azcosmos", serviceLibVersion,
+	return azruntime.NewPipeline("azcosmosgem", serviceLibVersion,
```
ealsur, line 138: What's the effect of the name change?

simorenoh: it changes the user agent that shows up - makes it so we know which request is coming from the main client and which one is coming from the gem. from the tests:
image

image

ealsur: Why would we want to? Remember that on the service side, the user agent is used to track telemetry and refreshing the account information and understanding if its the same client is a valid piece of information

simorenoh: during testing I found it made it a lot easier to understand where the requests where coming from and going to  since we do have two separate channels that are making requests, but I see your point as well - we can keep it as it was if this is not something we're concerned about

my only follow-up question to your point is - how would that difference in user agent affect our telemetry when we would still be looking for azsdk-go-azcosmos as startswith or azsdk-go-azcosmos and azsdk-go-azcosmosgem?

ealsur: 1. It's not something we do on any other SDK, so in terms of consistency, we shouldn't
2. Our telemetry categorizes SDKs by static parts of the user agent, having two potential values can make parsing more complicated than it should
3. These are requests the same client machine is doing, the user agent is supposed to represent the client. Altering the user agent within the same machine for the same process sounds incorrect?
4. It's just for one operation (getting account information) what are the chances of it being executed by the main pipeline? None, right? So we can assume all of these type of operations come from the gem policy
5. If the scenario is test purposes, we can alter the values on the tests

ealsur:
Once we have everything in place, these are the manual validation steps we should perform:

1. Create a real account with regions A and B. Assuming A is write, B is read.
2. Write a loop, performing 1 write and 1 read (the read do it of an existing doc, not the one created on the write), sleeping for a second inbetween
3. Configure the client with PreferredRegions B.
4. Enable logging

## Scenario 1 - Removing a region

1. Verify the Reads are going to Region B (the url endpoint of the request should be the regional of B)
2. Go to the Azure Portal, remove Region B from the account.
3. Client should not fail any request, read requests should now show to go to region A

## Scenario 2 - Adding a region

1. With the same loop running, add back region B.
2. The client should eventually start sending read requests back to region B. This might take some minutes.

## Scenario 3 - Failover

1. With the same loop running, go to the portal, and trigger a Manual Failover, converting Region B into the Write region.
2. Client should not fail any request, writes should eventually go to region B.

sdk/data/azcosmos/cosmos_database.go
```diff
@@ -209,8 +210,9 @@ func (db *DatabaseClient) Delete(
 	}
 
 	operationContext := pipelineRequestOptions{
-		resourceType:    resourceTypeDatabase,
-		resourceAddress: db.link,
+		resourceType:     resourceTypeDatabase,
+		resourceAddress:  db.link,
+		isWriteOperation: true,
```
ealsur, line 215: Don't forget Replace Container operations or Throughput changes

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -100,7 +91,7 @@ func (p *clientRetryPolicy) attemptRetryOnEndpointFailure(req *policy.Request, i
 	} else {
 		p.gem.MarkEndpointUnavailableForRead(*req.Raw().URL)
 	}
-	p.gem.Update(req.Raw().Context())
+	p.gem.Update(req.Raw().Context(), true)
```
ealsur, line 94: For 403/1008, we don't set it to true I think https://github.com/Azure/azure-cosmos-dotnet-v3/blob/deab135c9b8709abf3e62f7be08e6c022513e4bf/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs#L256

simorenoh: we seem to do it for both Java and Python for the entire scenario - but I can make it so we only do it on 403.3
image

image

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,188 @@
...
+		req.Raw().URL.Host = resolvedEndpoint.Host
+		response, err := req.Next() // err can happen in weird scenarios (connectivity, etc)
+		if err != nil {
+			if p.isNetworkConnectionError(err) {
+				shouldRetry, err := p.attemptRetryOnNetworkError(req)
+				if err != nil {
+					return nil, err
+				}
+				if !shouldRetry {
+					return nil, errorinfo.NonRetriableError(azruntime.NewResponseErrorWithErrorCode(response, response.Status))
```
ealsur, line 47: This is a bug I left, sorry!

```
shouldRetry, errRetry := p.attemptRetryOnNetworkError(req)
if errRetry != nil {
	return nil, errRetry
}
if !shouldRetry {
	return nil, err
```

If shouldRetry is false, we should return the original error

sdk/data/azcosmos/cosmos_client_retry_policy.go
```diff
@@ -0,0 +1,188 @@
...
+func (p *clientRetryPolicy) resetPolicyCounters() {
+	p.retryCount = 0
+	p.sessionRetryCount = 0
+	p.preferredLocationIndex = 0
+}
+
+// isNetworkConnectionError checks if the error is related to failure to connect / resolve DNS
+func (p *clientRetryPolicy) isNetworkConnectionError(err error) bool {
+	var dnserror *net.DNSError
+	return errors.As(err, &dnserror)
```
ealsur, line 187:
```suggestion
	return p.gem.locationCache.enableCrossRegionRetries && errors.As(err, &dnserror)
```

ealsur (approved):

PR #22397, opened 2024-02-15
Codeowner update

RickWinter:
Add missing AzureSdkOwners, ServiceLabel, and PRLabel metadata for the directories owned by the central azure sdk team

This data will be used by the engsys github bot to now auto-assign and tag owners in issue/pr comments (in addition to the auto-labeling it already does today)

Notes for code reviewers:

ServiceLabel is the label applied to issues, in nearly every single case this will be identical to PRLabel which is the label applied to pull requests.
AzureSdkOwners | ServiceOwners these fields control who the bot will auto-assign incoming issues to. Note that AzureSdkOwners has an overloaded significant meaning for PM team reporting purposes; it also means these owners work on the Central Azure SDK team (both dataplane and management plane) where as ServiceOwners means everyone else for their reporting. Don't shoot the messenger here.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,48 +12,80 @@
 # Samples
 /sdk/samples/                     @rickwinter @chlowell @jhendrixMSFT @richardpark-msft
 
+# AzureSDKOwners: @jhendrixMSFT
+# ServiceLabel: %Azure.Core
 # PRLabel: %Azure.Core
 /sdk/azcore/                  @rickwinter @chlowell @jhendrixMSFT @richardpark-msft
 
+# AzureSDKOwners: @chlowell
+# ServiceLabel: %Azure.Identity
 # PRLabel: %Azure.Identity
 /sdk/azidentity/              @chlowell @jhendrixMSFT @rickwinter @Azure/azure-sdk-write-identity
 
+# AzureSDKOwners: @richardpark-msft
+# ServiceLabel: %OpenAI
 # PRLable: %OpenAI
 /sdk/ai                       @richardpark-msft @jhendrixMSFT
 
+# AzureSDKOwners: @jhendrixMSFT
+# ServiceLabel: %Internal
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -12,48 +12,80 @@
...
+# ServiceLabel: %OpenAI
 # PRLable: %OpenAI
 /sdk/ai                       @richardpark-msft @jhendrixMSFT
 
+# AzureSDKOwners: @jhendrixMSFT
+# ServiceLabel: %Internal
 # PRLabel: %Internal
 /sdk/internal/                @chlowell @jhendrixMSFT @richardpark-msft @rickwinter
 
+# AzureSDKOwners: @antkmsft
```
jhendrixMSFT, line 35: Is this still accurate?

RickWinter: That should be you.
```suggestion
# AzureSDKOwners: @jhendrixMSFT 
```

.github/CODEOWNERS
```diff
@@ -12,48 +12,80 @@
...
 
+# AzureSDKOwners: @jhendrixMSFT
+# ServiceLabel: %Internal
 # PRLabel: %Internal
 /sdk/internal/                @chlowell @jhendrixMSFT @richardpark-msft @rickwinter
 
+# AzureSDKOwners: @jhendrixMSFT 
+# ServiceLabel: %App Configuration
 # PRLabel: %App Configuration
 /sdk/data/azappconfig/        @antkmsft @jhendrixMSFT @rickwinter
```
jhendrixMSFT, line 38: Do you want to remove Anton from here too while we're changing the file (separate is fine too)?

RickWinter, line 38:
```suggestion
/sdk/data/azappconfig/        @jhendrixMSFT @rickwinter
```

RickWinter: Updated

Issue #21841, opened 2023-10-27
Storage: authenticated requests must reject http endpoints by default


jhendrixMSFT:
This was done for Entra authentication in `azcore`, see https://github.com/Azure/azure-sdk-for-go/issues/21674. It should be expanded to all authentication schemes in blob storage.

For authentication types where http is allowed (SAS?), you can provide a mechanism to turn off the https requirement.

vibhansa-msft:
As these checks are being done in AzCore already, so we need to do any specific handling in storage moudles?

vibhansa-msft:
I see core is already handling this for Key, SAS and OAuth as part of this pr  :https://github.com/Azure/azure-sdk-for-go/pull/21832/files

jhendrixMSFT:
Does it also need to be applied to shared key authentication?

vibhansa-msft:
Other than SAS it shall be ok to reject http. SAS can be created only for http endpoint so not allowing http there may not work. For other modes it shall be ok. As part of the above PR, which I linked, I see key auth related changes are also done in core itself so is there any expectation or code change from the storage SDK side?

jhendrixMSFT:
`azblob` has its own `SharedKeyCredential` policy that needs to be updated https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/exported/shared_key_credential.go#L189-L218

souravgupta-msft:
Hi @roelarents. We have discussed this scenario internally and have decided to re-enable http endpoints for shared key based auth mode. You can currently either use SAS (where http is supported) or enable HTTPS in Azurite.
We'll do a release next week to fix this.

PR #22401, opened 2024-02-16
Storage: Re-enable HTTP endpoints for shared key credential

souravgupta-msft:
Revert the change made in #22183 and #21841.

Discussion with PM and some concerns raised on github for this reveals that Azure Stack and Azurite users either prefer or in some cases forced to use HTTP. Not allowing HTTP from SDK side block their setup. Similar issues were faced by AzCopy and Blobfuse in their pipelines to test against Azurite. Hence reverting back the changes and allowing HTTP with key credentials.

### DIFF
--- a/sdk/storage/azblob/CHANGELOG.md
+++ b/sdk/storage/azblob/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+* Re-enabled `SharedKeyCredential` authentication mode for non TLS protected endpoints.
+
 ### Other Changes
 
 ## 1.3.0 (2024-02-12)
--- a/sdk/storage/azblob/assets.json
+++ b/sdk/storage/azblob/assets.json
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "go",
   "TagPrefix": "go/storage/azblob",
-  "Tag": "go/storage/azblob_9f40a5a13d"
+  "Tag": "go/storage/azblob_71b0a04c12"
 }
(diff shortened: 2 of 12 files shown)


vibhansa-msft (approved):

jhendrixMSFT:
Is this just to address the storage emulator? If so, perhaps there's a different solution.

CC @JeffreyRichter

tasherif-msft (approved):

gapra-msft (approved):

gapra-msft:
> Is this just to address the storage emulator? If so, perhaps there's a different solution.
> 
> CC @JeffreyRichter

This also addresses environments like Azure Stack or private links where customers either prefer to use HTTP or are forced to use HTTP

gapra-msft:
> Is this just to address the storage emulator? If so, perhaps there's a different solution.
> 
> CC @JeffreyRichter

@jhendrixMSFT This also addresses some environments where customers are either forced to use HTTP or prefer to use HTTP. @seanmcc-msft and @schoag-msft to comment further.

tanyasethi-msft (approved):

jhendrixMSFT:
Yes but how common is that scenario? Disabling "secure by default" for something that I would think is the uncommon case seems a bit extreme, which is why I'm wondering is if there's another way to fix this.

JeffreyRichter:
Well, I was always of the opinion that HTTP should be supported and not stopped by the SDK. Just like it would not be stopped by a customer using raw HTTP without an SDK. But, I do appreciate that this flies in the face of secure by default but then you might say that HTTP is not secure by default too but it is how it works. 

Also, shared key does not mean that the key itself goes over the wire in clear text. The client SDK uses the key to sign a string and this signed string goes over the wire in clear text with HTTP. Since it works this, it is not really a problem.

jhendrixMSFT:
The clarification on how the string-to-sign works makes sense and was a missing piece of data.

jhendrixMSFT (approved):

PR #22402, opened 2024-02-16
Disable code coverage redesign in CI

jhendrixMSFT:
Go 1.22 changed some code coverage internals which is causing some cases to under-report coverage. Disabling for now until this is resolved.

See https://github.com/golang/go/issues/65570

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -80,7 +80,8 @@ steps:
       ${{ insert }}: ${{ parameters.EnvVars }}
       GOTRACEBACK: all
       AZURE_SDK_GO_LOGGING: all
-      GOEXPERIMENT: nocoverageredesign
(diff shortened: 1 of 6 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -80,6 +80,7 @@ steps:
       ${{ insert }}: ${{ parameters.EnvVars }}
       GOTRACEBACK: all
       AZURE_SDK_GO_LOGGING: all
+      GOEXPERIMENT: nocoverageredesign
```
jhendrixMSFT, line 83: Apparently we must set this only for Go 1.22.

jhendrixMSFT:
It also appears to affect the Go version string used in the `User-Agent` header which is causing some tests to fail.

`azsdk-go-azservicebus/v1.0.0 (go1.22.0 X:nocoverageredesign; Windows_NT)`

weshaggard (approved):

PR #22434, opened 2024-02-22
Update azcosmos

RickWinter:
* Use the latest version of azcore


sdk/data/azcosmos/go.mod
```diff
@@ -4,9 +4,9 @@ go 1.18
 
 require (
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.4.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.2
```
ealsur, line 7: Should we do it to https://github.com/Azure/azure-sdk-for-go/issues/22286#issuecomment-1958086028 when that is released?

RickWinter: Yes, when 1.10.0 ships, then this should be updated again.

ealsur (approved):

Issue #22456, opened 2024-02-27
[Aztables] Pagination is broken when using NewListEntitiesPager on some Cross Partition Queries


estebanreyl:
### Bug Report

- import path of package in question, e.g. `/sdk/data/aztables`
- SDK version: v1.1.0
  - Specify the exact commit if possible: 6c75210e1b4d41c68e3b23f68af3c7e97f3d8b79
- output of `go version` : go version go1.22.0 linux/amd64

- **What happened?** We encountered a situation where the default runtime pager fails to get all entities in a table when doing a cross partition scan. Instead, we will end up processing 0 entities as the More function returns false incorrectly.
- **What did you expect or want to happen?** When running through a cross partition query, it is possible for azure storage to return an empty list and a next partition only. In this case we need to use the partition as a continuation token to keep getting entries. Apparently, this is expected behavior  https://learn.microsoft.com/en-us/rest/api/storageservices/Query-Timeout-and-Pagination and we have verified this is happening on our storage account. 
- **How can we reproduce it?** I am still working on figuring this out, but my current guess is that if enough entities with different partitions are present in a single table this becomes more likely. 
- Anything we should know about your environment.
No

P.S. from looking at the code the fix amounts to changing the More function in sdk/dara/aztablesclient.go : NewListEntitiesPager to ignore the page.NextRowKey value. I still havent tested this and will try to follow up with a PR but want to leave this documented.

jhendrixMSFT:
Fix will be included in `aztables@v1.2.0` to be released March 5th.

estebanreyl:
Thanks for the quick fix :)

jhendrixMSFT:
Unfortunately there was a delay in the release, it's available now.

estebanreyl:
No worries, thanks for the update :)

PR #22469, opened 2024-02-28
NewListEntitiesPager fix

jhendrixMSFT:
Per the docs, if there are no continuation header values, there are no more entities to return.

Removed some superfluous if checks.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/22456

### DIFF
--- a/sdk/data/aztables/client.go
+++ b/sdk/data/aztables/client.go
@@ -166,10 +166,7 @@ func (t *Client) NewListEntitiesPager(listOptions *ListEntitiesOptions) *runtime
 		More: func(page ListEntitiesResponse) bool {
 			// if there are no continuation header values, there are no more pages
 			// https://learn.microsoft.com/rest/api/storageservices/Query-Timeout-and-Pagination
-			if (page.NextPartitionKey == nil || len(*page.NextPartitionKey) == 0) && (page.NextRowKey == nil || len(*page.NextRowKey) == 0) {
-				return false
-			}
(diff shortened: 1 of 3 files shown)


sdk/data/aztables/client.go
```diff
@@ -164,7 +164,9 @@ func (t *Client) NewListEntitiesPager(listOptions *ListEntitiesOptions) *runtime
 	}
 	return runtime.NewPager(runtime.PagingHandler[ListEntitiesResponse]{
 		More: func(page ListEntitiesResponse) bool {
-			if page.NextPartitionKey == nil || len(*page.NextPartitionKey) == 0 || page.NextRowKey == nil || len(*page.NextRowKey) == 0 {
+			// if there are no continuation header values, there are no more pages
+			// https://learn.microsoft.com/rest/api/storageservices/Query-Timeout-and-Pagination
+			if (page.NextPartitionKey == nil || len(*page.NextPartitionKey) == 0) && (page.NextRowKey == nil || len(*page.NextRowKey) == 0) {
```
chlowell, line 169:
```go
if (foo) {
    return false
}
return true
```
or `return !foo`?

chlowell (approved):

JeffreyRichter (approved):

sdk/data/aztables/client.go
jhendrixMSFT: Fixed.

PR #22509, opened 2024-03-02
[azservicebus, azeventhubs] Updating to take in the new AMQP version.

### LINKED ISSUE #21684: [azservicebus] Closing a session after it has idled out can take 30 seconds.
(We noticed this while troubleshooting a separate issue)

When a link is detached by the remotes service, because it was idle, it takes up to 30 seconds for us to actually close it locally. I haven't gotten a chance to dig in for root cause. Nothing appears to be broken afterwards - the client is still usable and new senders/receivers can be created. 

Note that this doesn't happen if you force the service to detach you - for instance, updating the remote entities metadata via ARM.

richardpark-msft:
Pulling in a new AMQP version with my fix for the "body size < frame size" that would cause to hang for 26 seconds while closing a session on a link that was idle.

Fixes #21684

### DIFF
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -1,14 +1,10 @@
 # Release History
 
-## 1.0.4 (Unreleased)
-
-### Features Added
-
-### Breaking Changes
+## 1.0.4 (2024-03-05)
 
 ### Bugs Fixed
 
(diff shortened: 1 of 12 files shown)


richardpark-msft:
NOTE to reviews: the go.mod's will be updated before merging once we release go-amqp including my latest fix. Everything else will remain the same.

jhendrixMSFT (approved):

Issue #21118, opened 2023-07-07
MQ: Automate Azure Identity SDK Manual/Vendor Tests


joshfree:
We should leverage @christothes 's work on https://github.com/Azure/azure-sdk-for-net/pull/33017/files to fully automate the list of manual/vendor test scenarios which are documented in the issues here: https://github.com/Azure/azure-sdk/issues?q=is%3Aopen+is%3Aissue+label%3AAzure.Identity

This includes:

- [ ] ~Automate Azure Arc testing~
- [ ] ~Automate Azure Identity multi tenant testing~
- [x] Automate Azure Container Instances testing #22652
- [ ] ~Automate Azure WebJobs testing~
- [ ] ~Automate Azure Service Fabric testing~
- [X] Automate Azure Functions testing #22562
- [x] Automate VM MSI testing #22517
- [x] Automate Workload Identity testing #22627
- [ ] ~Automate Pod Identity testing~ deprecated
- [ ] ~Automate Cloudshell testing~

PR #22517, opened 2024-03-05
Run live IMDS managed identity tests in weekly pipeline

chlowell:
We now have an agent pool whose VMs have a user-assigned identity, making it possible to test our IMDS managed identity code live. Here I've added that pool to the weekly pipeline's matrix and configured IMDS tests to run live within it. Part of #21118


sdk/azidentity/platform-matrix.json
```diff
@@ -0,0 +1,16 @@
...
+    "include": [
+        {
+            "Agent": {
+                "msi_image": {
+                    "OSVmImage": "ubuntu-22.04",
+                    "Pool": "azsdk-pool-mms-ubuntu-2204-identitymsi"
+                }
+            },
+            "GoVersion": [
+                "1.18"
```
jhendrixMSFT, line 11: Should this be expanded to include more versions?

jhendrixMSFT (approved):

sdk/azidentity/platform-matrix.json
RickWinter: We should run on n and n-1 at a minimum.

chlowell: I think one suffices because we test this code with other versions offline and I can't imagine a difference between two versions that only a live test would reveal 🤔

jhendrixMSFT: I tend to agree. Perhaps this should be the latest version though? We've already got 1.18 compat coverage elsewhere, so if we're going for single-version coverage it might be better to use the N version.

PR #22562, opened 2024-03-08
Live Azure Functions managed identity test

chlowell:
This adds a program that exercises our managed identity implementation by accessing storage accounts, scripts to deploy it to Azure Functions and create the storage accounts, and a test that triggers the Function. All this will run only in our weekly test pipeline, which I reconfigured to run on only one agent because the code under test runs remotely; the platform running the test suite is irrelevant and it isn't interesting to do so more than once. I chose to build the program in a container because the pipeline deploys resources before setting up a Go environment and we need a container to deploy to other platforms we want to test on anyway. Part of #21118


jhendrixMSFT (approved):

sdk/azidentity/test-resources.bicep
```diff
@@ -1 +1,138 @@
-param baseName string
+@minLength(6)
```
RickWinter, line 1: Should bicep files have a copyright?

sdk/azidentity/testdata/managed-id-test/go.mod
```diff
@@ -0,0 +1,24 @@
...
+
+go 1.18
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.10.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.5.1
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.3.1
+)
+
+replace github.com/Azure/azure-sdk-for-go/sdk/azidentity => ../../
```
RickWinter, line 11: Is this done to ensure we are testing the PR?  If so please add a comment

RickWinter (approved):

PR #22564, opened 2024-03-09
Refactored if statement tests to table tests to increase readability …

griffinbird:
…and maintainability.

- [ X] The purpose of this PR is explained in this or a referenced issue.
- [ X] The PR does not update generated files.
- [ X] Tests are included and/or updated for code changes.
- [ X] MIT license headers are included in each file.

### DIFF
--- a/sdk/data/azcosmos/cosmos_container_properties_test.go
+++ b/sdk/data/azcosmos/cosmos_container_properties_test.go
@@ -66,92 +66,41 @@ func TestContainerPropertiesSerialization(t *testing.T) {
 		t.Fatal(err, string(jsonString))
 	}
 
-	if properties.ID != otherProperties.ID {
-		t.Errorf("Expected Id to be %s, but got %s", properties.ID, otherProperties.ID)
+	tests := []struct {
+		name     string
+		got      interface{}
+		expected interface{}
+	}{
+		{"ID", otherProperties.ID, properties.ID},
+		{"ETag", *otherProperties.ETag, *properties.ETag},
+		{"SelfLink", otherProperties.SelfLink, properties.SelfLink},
+		{"ResourceID", otherProperties.ResourceID, properties.ResourceID},
+		{"LastModified", otherProperties.LastModified, properties.LastModified},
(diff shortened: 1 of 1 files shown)


sdk/data/azcosmos/cosmos_container_properties_test.go
```diff
@@ -66,92 +66,41 @@ func TestContainerPropertiesSerialization(t *testing.T) {
...
+		{"IndexingPolicyCompositeIndexesPath", otherProperties.IndexingPolicy.CompositeIndexes[0][0].Path, properties.IndexingPolicy.CompositeIndexes[0][0].Path},
+		{"UniqueKeyPolicy", otherProperties.UniqueKeyPolicy != nil, properties.UniqueKeyPolicy != nil},
+		{"UniqueKeyPolicyUniqueKeys", otherProperties.UniqueKeyPolicy.UniqueKeys[0].Paths[0], properties.UniqueKeyPolicy.UniqueKeys[0].Paths[0]},
+		{"ConflictResolutionPolicy", otherProperties.ConflictResolutionPolicy != nil, properties.ConflictResolutionPolicy !=nil},
+		{"ConflictResolutionPolicyMode", otherProperties.ConflictResolutionPolicy.Mode, properties.ConflictResolutionPolicy.Mode},
+		{"ConflictResolutionPolicyResolutionPath", otherProperties.ConflictResolutionPolicy.ResolutionPath, properties.ConflictResolutionPolicy.ResolutionPath},
+	}
+	
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
ealsur, line 99: I personally find this refactoring a bit less clear than the `if`s. Having to understand the usage of `t.Run` with the delegate seems a bit more complicated than understanding a list of `if` conditions.

But I would see value if these `if`s are unbounded/can keep growing, so it would reduce the volume of code required. But the reality is that these properties will not really change frequently so the chances of a developer having to keep adding checks here is very low to offset the increase in reading complexity.

Finally, the same pattern (`if`s) are used throughout the other files. If we want to think a common pattern, I guess we would need to refactor all the other files too?

lirenhe:
> @lirenhe can you analyze and determine if you want to take this change. Currently it is failing various pipelines so it needs work before it can be accepted

@RickWinter, could you help to provide more context about its impact? Is there any pipelines in mgmt. SDK that also need this change?

Issue #22448, opened 2024-02-25
Why should this be hardcoded to Https?


yanivsag:
https://github.com/Azure/azure-sdk-for-go/blob/bac2559f8c41323bc04ad06c356de22f50d97df5/sdk/storage/azblob/container/client.go#L351

I suggest basing this on the URL, since it makes the method unusable for users of the azurite storage emulator.

vibhansa-msft:
Hi @yanivsag 
GetSasURL is just a helper method to generate SAS from SharedKey or UserDelegationKey. We have used 'https' as default, but that does not restrict you in anyway. You can create your own BlobSignatureValue in context of your application and generate a SAS which allows both http and https.

vibhansa-msft:
0.3.0 was not a GA release. For the security reasons HTTP was not allowed in certain auth flows. By default its HTTPs now and for HTTP you need to create SAS by calling respective SDK apis and not through GetSASToken(). For now you can use this as a workaround, let me discuss this with team and see what shall be the correct way to handle this.

yanivsag:
got it, thanks!

vibhansa-msft:
This has been fixed as part of #22704

PR #22566, opened 2024-03-11
fix: add AZURE_GO_SDK_DISABLE_HTTPS_CHECK to provide workaround to disable https auth check

andyzhangx:
add AZURE_GO_SDK_DISABLE_HTTPS_CHECK to provide workaround to disable https auth check
background: during our AKS CSI driver migration to track2, we found track2 has enforced https, and would return error if it's http request. In our AKS production, we have set cloud.resourceManagerEndpoint as `http://localhost` to a proxy inside AKS pod(container), and finally the proxy would send ARM https request (this is safe since http request is handled inside container). We have a critical fix to rollout a new CSI driver version in AKS production while we found this https auth check has blocked our new version rollout, and we are trying to add https support inside our aks proxy, while that would take long time to rollout in production since this is not a trivial change. And I believe adding an env var to disable such https check is also a reasonable workaround since track1 sdk does not enforce such https auth check, this is a like a breaking change.

This PR provides a env var to disable https auth check in such case, and does not change the default behavior of go sdk behavior.

### DIFF
--- a/sdk/azcore/runtime/policy_bearer_token.go
+++ b/sdk/azcore/runtime/policy_bearer_token.go
@@ -6,6 +6,7 @@ package runtime
 import (
 	"errors"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


andyzhangx:
@hendrixMSFT could you take a look? I know https enforcement is a reasonable change in track2, while if there is a workaround, that would be great, we will try to support https in the near future even in our internal proxy container, thanks for understanding. thanks.

MartinForReal:
related #22448

jhendrixMSFT (changes requested):
After discussion, the current POR is to make this a `bool` in `azcore.ClientOptions`.

jhendrixMSFT:
Superseded by https://github.com/Azure/azure-sdk-for-go/pull/22573

Issue #22563, opened 2024-03-09
Resource provider not registered automatically for Microsoft.Security/pricings


felipebbc:
When enabling Defender for Cloud plans through the Terraform AzApi provider, the automatic registration of the Microsoft.Security provider does not occur.

Sample code:
```
resource "azapi_update_resource" "subscription_dfc_plans" {
  parent_id = "/subscriptions/${local.subscription_id}"
  type      = "Microsoft.Security/pricings@2024-01-01"
  name      = "CloudPosture"
  body = jsonencode({
    properties = {
      pricingTier = "Standard"
    }
  })
}
```

Error message:
```
 Error: checking for presence of existing Resource: (ResourceId "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx/providers/Microsoft.Security/pricings/CloudPosture" / Api Version "2024-01-01"): GET https://management.azure.com/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx/providers/Microsoft.Security/pricings/CloudPosture
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE: Subscription Not Registered
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "Subscription Not Registered",
│     "message": "Please register to Microsoft.Security in order to view your security status"
│   }
│ }
│ --------------------------------------------------------------------------------
```
This issue was partially addressed with #20848 by adding the non-standard error code `Subscription Not Registered`.

However, the sdk also checks if the response has a 409 error code:
https://github.com/Azure/azure-sdk-for-go/blob/6387cfc7d4421e4df032c64fc7fab4cf36f7c622/sdk/azcore/arm/runtime/policy_register_rp.go#L90-L92

Since the error message from the Microsoft.Security/pricings@2024-01-01 returns a **404 error code**, the provider registration doesn't move forward.

Also, the provider registration requires that the error message contain the provider name in the `error.details.target` field

https://github.com/Azure/azure-sdk-for-go/blob/6387cfc7d4421e4df032c64fc7fab4cf36f7c622/sdk/azcore/arm/runtime/policy_register_rp.go#L202-L207

The error message from the Microsoft.Security/pricings@2024-01-01 does not contain the `error.details.target` field, it only contains `error.code` and `error.message`:

[...]

PR #22569, opened 2024-03-11
Updated automatic provider registration to handle additional errors

felipebbc:
Fixes the issue with Resource provider not registered automatically for Microsoft.Security/pricings #22563. 

- Extends the check on http code 409 to include 404 and 400, the latter related to issue #20823. 
- Obtains the applicable resource provider (e.g., Microsoft.Security, Microsoft.Quota) from the `error.message` field in the http response if not available in the `error.details.target` field  
- Maintains the requirement to also match on error code/unregisteredRPCodes (e.g., "MissingSubscriptionRegistration", "Subscription Not Registered")

### DIFF
--- a/sdk/azcore/arm/runtime/policy_register_rp.go
+++ b/sdk/azcore/arm/runtime/policy_register_rp.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


RickWinter:
Expecting the service to return BadRequest is odd. Can you link to the ARM guidelines specifiying this is a valid design.

jhendrixMSFT (changes requested):
The service returning anything other than a 409 is a mistake and needs to be fixed on the service side (I'm following up on that). So, I'm hesitant to include other status codes by default as it will mask non-conformant behavior. That said, we realize there's a need here to get things working in the interim and are exploring some solutions.

Issue #22482, opened 2024-02-29
[azopenai]Error when connecting to privately deployed OpenAI API endpoint via HTTP protocol


wh7f:
### Bug Report

**Error Message**

`authenticated requests are not permitted for non TLS protected (https) endpoints`

**Environment**
- golang: 1.21
- SDK:
    - github.com/Azure/azure-sdk-for-go/sdk/ai/azopenai v0.4.1
    - github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.2

**Describe**

Example source code:
```go
func BuildClient(ctx context.Context) (*azopenai.Client, error) {
	baseURL := "http://xxx" // Privately deployed OpenAI API-compatible services
	apiKey := "kkk"
	keyCredential := azcore.NewKeyCredential(apiKey)

	openaiCompatibleClient, err := azopenai.NewClientForOpenAI(baseURL, keyCredential, nil)
	if err != nil {
		return nil, err
	}

	return openaiCompatibleClient, nil
}
```

- What happened?
    - An error occurred when requesting: `authenticated requests are not permitted for non TLS protected (https) endpoints`.
- What did you expect or want to happen?
    - Send request normally.
- How can we reproduce it?
    - Turn off the restriction that `azopenai.NewClientForOpenAI` cannot use keys on http endpoints.
    - Or provide a configuration that turns off the restriction.
- Anything we should know about your environment.
    - I need to use the same SDK to access the OpenAI service provided by Azure and the independently deployed OpenAI API-compatible service.
    - The independently deployed environment is isolated from the Internet, so there will be no security issues.

**Thanks!**

`azopenai` is the best golang OpenAI API client SDK. Thank you engineers for your hard work.

wh7f:
Yes, endpoints are provided by other teams, and they use authorization key to count calls and allocate costs.

jhendrixMSFT:
The ability to skip the HTTPS will ship in `azcore@v1.11.0` first week of April.

PR #22573, opened 2024-03-11
Add AllowInsecureAuth to ClientOptions

jhendrixMSFT:
When true, allows authenticated requests over HTTP.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/22482

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features Added
 
 * Added `StatusCodes` to `arm/policy.RegistrationOptions` to allow supporting non-standard HTTP status codes during registration.
+* Added field `AllowInsecureAuth` to `azcore.ClientOptions` and dependent authentication pipeline policies.
 
 ### Breaking Changes
 
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Features Added
 
 * Added `StatusCodes` to `arm/policy.RegistrationOptions` to allow supporting non-standard HTTP status codes during registration.
-* Added field `AllowInsecureAuth` to `azcore.ClientOptions` and dependent authentication pipeline policies.
+* Added field `InsecureAllowCredentialWithHTTP` to `azcore.ClientOptions` and dependent authentication pipeline policies.
 
 ### Breaking Changes
 
--- a/sdk/azcore/arm/policy/policy.go
+++ b/sdk/azcore/arm/policy/policy.go
@@ -20,12 +20,12 @@ type BearerTokenOptions struct {
 	// policy's credential must support multitenant authentication.
 	AuxiliaryTenants []string
 
+	// InsecureAllowCredentialWithHTTP enables authenticated requests over HTTP.
+	// WARNING: setting this to true will allow sending the authentication key in clear text. Use with caution.
+	InsecureAllowCredentialWithHTTP bool
+
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
-
-	// AllowInsecureAuth enables authenticated requests over HTTP.
-	// WARNING: setting this to true will allow sending the authentication key in clear text. Use with caution.
-	AllowInsecureAuth bool
 }
 
 // RegistrationOptions configures the registration policy's behavior.
--- a/sdk/azcore/arm/policy/policy.go
+++ b/sdk/azcore/arm/policy/policy.go
@@ -24,7 +24,7 @@ type BearerTokenOptions struct {
 	Scopes []string
 
 	// AllowInsecureAuth enables authenticated requests over HTTP.
-	// WARNING: setting this to true will send the authentication key in clear text. Use with caution.
+	// WARNING: setting this to true will allow sending the authentication key in clear text. Use with caution.
 	AllowInsecureAuth bool
 }
 
--- a/sdk/azcore/arm/policy/policy.go
+++ b/sdk/azcore/arm/policy/policy.go
@@ -21,6 +21,7 @@ type BearerTokenOptions struct {
 	AuxiliaryTenants []string
 
 	// InsecureAllowCredentialWithHTTP enables authenticated requests over HTTP.
+	// By default, authenticated requests to an HTTP endpoint are rejected by the client.
 	// WARNING: setting this to true will allow sending the authentication key in clear text. Use with caution.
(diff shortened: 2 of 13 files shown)


sdk/azcore/arm/runtime/policy_bearer_token_test.go
```diff
@@ -51,7 +51,7 @@ func newTestPipeline(opts *azpolicy.ClientOptions) runtime.Pipeline {
 	return runtime.NewPipeline("testmodule", "v0.1.0", runtime.PipelineOptions{}, opts)
 }
 
-func defaultTestPipeline(srv azpolicy.Transporter, scope string) (runtime.Pipeline, error) {
```
jhendrixMSFT, line 54: Unused param.

andyzhangx:
@jhendrixMSFT thanks for the fix! when this PR would be get merged, and the new release would be published? thanks.

jhendrixMSFT:
@andyzhangx our default release schedule for `azcore` is the first week of each month. I believe it would be 4/4. How does this impact AKS?

andyzhangx:
> @andyzhangx our default release schedule for `azcore` is the first week of each month. I believe it would be 4/4. How does this impact AKS?

@jhendrixMSFT it has no impact on AKS production now since I have reverted the track2 version change in AKS now, but if you could make this release faster, e.g. next week, that would be great since I could continue upgrade our CSI driver version with track2 change in AKS as early as possible, thanks.

sdk/azcore/arm/policy/policy.go
```diff
@@ -22,6 +22,10 @@ type BearerTokenOptions struct {
 
 	// Scopes contains the list of permission scopes required for the token.
 	Scopes []string
+
+	// AllowInsecureAuth enables authenticated requests over HTTP.
+	// WARNING: setting this to true will send the authentication key in clear text. Use with caution.
```
chlowell, line 27:
```suggestion
	// WARNING: setting this to true will allow sending the authentication key in clear text. Use with caution.
```

chlowell (approved):

sdk/azcore/arm/policy/policy.go
jhendrixMSFT: Good idea, fixed.

sdk/azcore/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # Release History
 
-## 1.10.1 (Unreleased)
+## 1.11.0 (Unreleased)
 
 ### Features Added
 
+* Added field `AllowInsecureAuth` to `azcore.ClientOptions` and dependent authentication pipeline policies.
```
jhendrixMSFT, line 7: I'm not super happy with the name `AllowInsecureAuth`. It's really more like "allow authorized requests over HTTP". I like having the word "Insecure" in the name as it's an indicator that enabling this has potentially dangerous repercussions. I can live with the name but open to suggestions.

richardpark-msft: AllowInsecureRequests seems a bit more general and better since that's our overall intent.

I'm curious if we want to mimic the Go setting's name for skipping certs, which is a similar idea.

So maybe `InsecureAuth` or `InsecureRequests`, similar to here: [tls.InsecureSkipVerify](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/crypto/tls/common.go;l=669)

richardpark-msft: Perhaps even InsecureSkipHTTPSCheck

jhendrixMSFT: I like the idea of having "Insecure" as the first word.

jhendrixMSFT: We'll go with `InsecureAllowCredentialWithHTTP`

ffeldmann:
HI @jhendrixMSFT, thanks for the fix. However it does not work when uploading blobs. Looks like in `sdk/storage/azblob/service/client.go` for the NewClient function the authPolicy function, responsible for the BearerTokenPolicy does not get the ClientOptions passed. Hence, the variable allowHTTP is always false.

https://github.com/Azure/azure-sdk-for-go/blob/18e40d09150fa9cafd442e42770288f5a51cd740/sdk/storage/azblob/service/client.go#L42-L45


I suggest passing the ÌnsecureAllowCredentialsWithHTTP variable directly to NewStorageChallangePolicy:

```Go
func NewClient(serviceURL string, cred azcore.TokenCredential, options *ClientOptions) (*Client, error) {
	audience := base.GetAudience((*base.ClientOptions)(options))
	authPolicy := shared.NewStorageChallengePolicy(cred, audience, options.InsecureAllowCredentialWithHTTP)
	conOptions := shared.GetClientOptions(options)
	plOpts := runtime.PipelineOptions{PerRetry: []policy.Policy{authPolicy}}
```

and in sdk/storage/azblob/service/client.go

``` golang
func NewStorageChallengePolicy(cred azcore.TokenCredential, audience string, allowHTTP bool) policy.Policy {
	s := storageAuthorizer{scopes: []string{audience}}
	return runtime.NewBearerTokenPolicy(cred, []string{audience}, &policy.BearerTokenOptions{
		InsecureAllowCredentialWithHTTP: allowHTTP,
		AuthorizationHandler: policy.AuthorizationHandler{
			OnRequest:   s.onRequest,
			OnChallenge: s.onChallenge,
		},
	})

```

tanyasethi-msft:
@ffeldmann - You can upgrade to azblob v1.3.2
The above fix has been rolled out in the latest release. 
Refer : https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fstorage%2Fazblob%2Fv1.3.2

andyzhangx:
> @ffeldmann - You can upgrade to azblob v1.3.2 The above fix has been rolled out in the latest release. Refer : https://github.com/Azure/azure-sdk-for-go/releases/tag/sdk%2Fstorage%2Fazblob%2Fv1.3.2

@tanyasethi-msft thanks a lot for the quick response, it's unblocking our AKS blob csi driver version upgrade now.

ukrocks007:
@ffeldmann You can enable this for custom_client as well by adding `InsecureAllowCredentialWithHTTP: options.InsecureAllowCredentialWithHTTP,` to `KeyCredentialPolicyOptions`
https://github.com/Azure/azure-sdk-for-go/blob/3a7fc1c5447b78311d02a20521c6e552c5323cb9/sdk/ai/azopenai/custom_client.go#L96-L98

PR #22577, opened 2024-03-11
Add StatusCodes option to ARM RP registration policy

jhendrixMSFT:
This lets customers work around services with non-conformant behavior. Replaced custom resource type parsing with in-box version.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/22563

I chose not to expose `StatusCodes` in `arm/policy.ClientOptions` as it should be very rare (not to mention temporal as the RP should fix their code). This does require one to disable the in-box registration policy and create their own version.

```go
regPol, err := armruntime.NewRPRegistrationPolicy(cred, &armpolicy.RegistrationOptions{
	StatusCodes: []int{http.StatusConflict, http.StatusNotFound},
})
if err != nil {
	// TODO: handle error
}

client, err := armmonitor.NewActivityLogsClient("subscription-id", cred, &arm.ClientOptions{
		PerCallPolicies: []policy.Policy{regPol},
	},
	DisableRPRegistration: true,
})
```

### DIFF
--- a/sdk/azcore/arm/policy/policy.go
+++ b/sdk/azcore/arm/policy/policy.go
@@ -44,6 +44,11 @@ type RegistrationOptions struct {
 	// The default valule is 5 minutes.
 	// NOTE: Setting this to a small value might cause the policy to prematurely fail.
 	PollingDuration time.Duration
+
+	// StatusCodes contains the slice of custom HTTP status codes to use instead
+	// of the default http.StatusConflict. This should only be set if a service
+	// returns a non-standard HTTP status code when unregistered.
+	StatusCodes []int
 }
 
 // ClientOptions contains configuration settings for a client's pipeline.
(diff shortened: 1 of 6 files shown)


chlowell (approved):

sdk/azcore/arm/policy/policy.go
```diff
@@ -44,6 +44,11 @@ type RegistrationOptions struct {
 	// The default valule is 5 minutes.
 	// NOTE: Setting this to a small value might cause the policy to prematurely fail.
 	PollingDuration time.Duration
+
+	// StatusCodes contains the slice of custom HTTP status codes to use instead
+	// of the default http.StatusConflict. This should only be set if a service
+	// returns a non-standard HTTP status code when unregistered.
+	StatusCodes []int
 }
```
jhendrixMSFT, line 52: We could also consider adding a slice for additional error codes as that has been problematic in the past.

JeffreyRichter (approved):

PR #22581, opened 2024-03-12
Improved support for multipart/form payloads

jhendrixMSFT:
Added type MultipartContent to allow specifying Content-Type and file name for a payload.
Fixed an issue in SetMultipartFormData that would incorrectly quote raw JSON data.

### DIFF
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -132,8 +132,6 @@ func MarshalAsXML(req *policy.Request, v any) error {
 	return req.SetBody(exported.NopCloser(bytes.NewReader(b)), shared.ContentTypeAppXML)
 }
 
-var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
-
 // SetMultipartFormData writes the specified keys/values as multi-part form
 // fields with the specified value.  File content must be specified as a ReadSeekCloser.
 // Byte slices will be treated as JSON. All other values are treated as string values.
@@ -165,6 +163,7 @@ func SetMultipartFormData(req *policy.Request, formData map[string]any) error {
 		}
 		// this is pretty much copied from multipart.Writer.CreateFormFile
 		// but lets us set the caller provided Content-Type and filename
+		quoteEscaper := strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
 		h := make(textproto.MIMEHeader)
 		h.Set("Content-Disposition",
 			fmt.Sprintf(`form-data; name="%s"; filename="%s"`,
--- a/sdk/azcore/runtime/request.go
+++ b/sdk/azcore/runtime/request.go
@@ -211,17 +211,19 @@ func SetMultipartFormData(req *policy.Request, formData map[string]any) error {
 			}
 			continue
 		}
+
+		var content string
 		if b, ok := v.([]byte); ok {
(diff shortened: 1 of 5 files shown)


sdk/azcore/runtime/request.go
```diff
@@ -161,13 +198,32 @@ func SetMultipartFormData(req *policy.Request, formData map[string]any) error {
...
+			}
+			continue
 		}
-		// ensure the value is in string format
-		s, ok := v.(string)
-		if !ok {
-			s = fmt.Sprintf("%v", v)
+
+		var content string
+		if b, ok := v.([]byte); ok {
```
chlowell, line 216: Just FYI in case you imagine we'll ever use `gocritic`, I think it wants `switch` instead of if-else if-else

sdk/azcore/runtime/request.go
```diff
@@ -148,6 +153,38 @@ func SetMultipartFormData(req *policy.Request, formData map[string]any) error {
 		return nil
 	}
 
+	writeMultipartContent := func(fieldname, filename string, mpc streaming.MultipartContent) error {
```
chlowell, line 156: Why is this local but not `quoteEscaper`?

sdk/azcore/runtime/request.go
```diff
@@ -148,6 +153,38 @@ func SetMultipartFormData(req *policy.Request, formData map[string]any) error {
...
+		if mpc.Body == nil {
+			return errors.New("streaming.MultipartContent.Body cannot be nil")
+		}
+		if mpc.ContentType == "" && mpc.Filename == "" {
+			return writeContent(fieldname, filename, mpc.Body)
+		}
+		if mpc.Filename != "" {
```
richardpark-msft, line 163: Why do we have a named parameter named 'filename' and _also_ an optional 'mpc.Filename'? Shouldn't it be one or the other?

sdk/azcore/runtime/request.go
jhendrixMSFT: Good point, moved `quoteEscaper`.

sdk/azcore/runtime/request.go
jhendrixMSFT: Converted to use a type switch.

sdk/azcore/runtime/request.go
jhendrixMSFT: [`multipart.Writer.CreateFormFile`](https://pkg.go.dev/mime/multipart#Writer.CreateFormFile) requires a file name. We default the file name to the field name. However, tsp has introduced a requirement that apps be able to specify the file name in case the endpoint cares about it.

I should point out that we will only generate `MultipartContent` for tsp. OpenAPI will continue to use `io.ReadSeekCloser` which doesn't allow specifying the file name or Content-Type. We _could_ change it if required (or if we think we should), but it would be a breaking change (I don't know that any real services would be impacted).

chlowell: This is probably just me being dense and not understanding this feature but I also don't get why there are two ways for a caller to specify `filename`. Why would a caller need another argument to override the value it passed as `mpc.Filename`?

jhendrixMSFT: I was the one being dense, for some reason what Richard pointed out went over my head. I've removed the redundant param.

chlowell (approved):

PR #22582, opened 2024-03-12
[aznamespaces] Adding in the ability to send events using BinaryMode.

### LINKED ISSUE #22304: [aznamespaces] Event Grid Namespaces needs support for a "binary cloud event"
There's a way to pass a cloud event so the body is the payload and the other attributes are encoded into HTTP headers. We just need to build support for this into the Event Grid Namespaces client.

richardpark-msft:
This is a more efficient way of sending CloudEvents as it avoids JSON serialization for the body. It has some restrictions/caveats, which are documented.

Fixes #22304

### DIFF
--- a/sdk/messaging/eventgrid/aznamespaces/client_custom.go
+++ b/sdk/messaging/eventgrid/aznamespaces/client_custom.go
@@ -231,8 +231,6 @@ func stringize(tmpV any) (string, error) {
 
 	case []byte:
 		return base64.StdEncoding.EncodeToString(v), nil
-	case *url.URL:
-		return v.String(), nil
 	case string:
 		return v, nil
 
--- a/sdk/messaging/eventgrid/aznamespaces/client_custom.go
+++ b/sdk/messaging/eventgrid/aznamespaces/client_custom.go
@@ -7,11 +7,22 @@
 package aznamespaces
 
 import (
+	"bytes"
 	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/messaging"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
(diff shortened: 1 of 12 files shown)


sdk/messaging/eventgrid/aznamespaces/client_custom.go
```diff
@@ -100,3 +111,140 @@ func (client *Client) ReleaseCloudEvents(ctx context.Context, topicName string,
...
+		return strconv.FormatUint(uint64(v), 10), nil
+	case uint:
+		return strconv.FormatUint(uint64(v), 10), nil
+	case uint32:
+		return strconv.FormatUint(uint64(v), 10), nil
+	case uint64:
+		return strconv.FormatUint(uint64(v), 10), nil
+
+	case []byte:
+		return base64.StdEncoding.EncodeToString(v), nil
```
jhendrixMSFT, line 233: This will include padding characters. Just checking this is ok.

sdk/messaging/eventgrid/aznamespaces/client_custom.go
```diff
@@ -100,3 +111,140 @@ func (client *Client) ReleaseCloudEvents(ctx context.Context, topicName string,
...
+		return v.String(), nil
+	case string:
+		return v, nil
+
+	case time.Time:
+		return v.Format(time.RFC3339), nil
+	case *time.Time:
+		return v.Format(time.RFC3339), nil
+
+	case fmt.Stringer:
```
jhendrixMSFT, line 244: I think this will check if the type of `v` is `fmt.Stringer`, not that `v` implements it? If so, you should pull this out of the `switch` and check if `v` implements `fmt.Stringer` first. It would also allow you to remove the case for `*url.URL`.

richardpark-msft: It is doing an "implements" check (just verified it since I can remove the url.URL case and it properly works).

Interestingly I had a hard time finding something saying that in the Go docs but there is this bit where they write the equivalent check in if/else statements that shows that 'case <interface>' does type assertions, which is what we were hoping for:

https://go.dev/ref/spec#:~:text=%7D%20else%20if%20i%2C%20isFloat64%20%3A%3D%20v.(float64)%3B%20isFloat64%20%7B

sdk/messaging/eventgrid/aznamespaces/client_custom.go
richardpark-msft: That's a good point, I'll check to see if that's called out in the spec or not.

richardpark-msft: It's not called out explicitly in the spec except to reference the Base64 RFC. With that I think we're good - it looks like other implementations write out padding as well.

jhendrixMSFT (approved):

PR #22583, opened 2024-03-12
[cleanup] Remove old cognitiveservices/azopenai module

richardpark-msft:
We moved this awhile ago into `ai/azopenai`. It's been published with our retraction so it should no longer be needed.


jhendrixMSFT (approved):

RickWinter (approved):

richardpark-msft:
Going to override check-enforcer here since my entire purpose was deleting this project, which means the pipeline no longer exists/works.

Issue #18986, opened 2022-08-30
Cosmos DB: Diagnostics for operations


ealsur:
Similarly to other Cosmos DB SDKs, expose Diagnostics for each operation that includes relevant troubleshooting information.

Initially it can be scoped to:

* Relevant client configuration (like which Preferred Regions)
* All requests involved in an operation, including retries handled by cross-region management

jay-most:
Notes: Time :)

ealsur:
https://github.com/Azure/azure-sdk-for-go/pull/22288 Added logging support and examples.

simorenoh:
https://github.com/Azure/azure-sdk-for-go/pull/22593 and https://github.com/Azure/azure-sdk-for-go/pull/22598 add logging of account and client settings

PR #22598, opened 2024-03-15
[Cosmos] add preferred regions logging

simorenoh:
Adds the missing client config logs since this flag was the last bit missing for that. Logs look like this: 
image

Progress towards closing https://github.com/Azure/azure-sdk-for-go/issues/18986.

### DIFF
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -14,11 +14,9 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-	azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
-	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
 const (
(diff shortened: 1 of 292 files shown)


sdk/data/azcosmos/cosmos_client_options.go
```diff
@@ -15,4 +17,11 @@ type ClientOptions struct {
 	EnableContentResponseOnWrite bool
 	// PreferredRegions is a list of regions to be used when initializing the client in case the default region fails.
 	PreferredRegions []string
+	// DisableCrossRegionalRetries is a boolean flag letting the SDK know whether or not it should apply cross-regional retries to requests.
```
ealsur, line 20: Add another `//` comment line below with `For more information, see https://learn.microsoft.com/azure/cosmos-db/nosql/troubleshoot-sdk-availability

simorenoh: as per our SDK conversation I'll be removing this entirely

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -70,7 +73,7 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-
+	log.Write(azlog.EventResponse, "\n===== Client configuration:\n"+o.String()+"\n=====\n")
```
ealsur, line 76: Our diagnostics in other SDKs print this on all operations, should we do the same here? Maybe before https://github.com/Azure/azure-sdk-for-go/blob/6289abf0dc0c0c6b34ba96ba385a135f34ab39a6/sdk/data/azcosmos/cosmos_client.go#L474 ?

simorenoh: does it make sense for us to though? can the client settings be changed by the user without restarting their client? I'd think these settings hold throughout the lifecycle of the user's application right?

I understand that parity is a big thing, but based on this morning's conversation I think we should be good to deviate from that no?

ealsur: Imagine someone going through the logs, finding an operation that went to the "wrong" region, then filing an Issue with that complaint and including only the log for the operation. It would not be possible to know why?

simorenoh:
with 1.21 in the `go.mod` file, these two runs complain of the lack of the min method:
image
image

ealsur (approved):

PR #22600, opened 2024-03-15
A few more yml updates

scbedd:
Swapping a few more `ci.yml` files to `extends` vs `stages` at their root.

Follow-up from #22590 and [the follow-up commit](https://github.com/Azure/azure-sdk-for-go/commit/88d8a7ebb55562f8b5fdc48644d31cc13d383c7d)

Not certain how it escaped the first ctrl-shift-f, but they did!


benbp (approved):

scbedd:
/check-enforcer override

Unrelated failure.

PR #22607, opened 2024-03-18
DataLake encryption context

tanyasethi-msft:
- [x]  Updates to module CHANGELOG.md are included.

This PR contains the following features - 
STG87 : Data Lake Encryption Context 
STG88 : Add Encryption Context to File Upload Options

### DIFF
--- a/sdk/storage/azdatalake/file/client.go
+++ b/sdk/storage/azdatalake/file/client.go
@@ -443,6 +443,13 @@ func (f *Client) uploadFromReader(ctx context.Context, reader io.ReaderAt, actua
 				urlParts.PathName, actualSize, o.ChunkSize, ((actualSize-1)/o.ChunkSize)+1)
 		}
 	}
+	// create a new file and set encryption context options to it if it is not nil
+	if o.EncryptionContext != nil {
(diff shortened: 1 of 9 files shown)


sdk/storage/azdatalake/file/client.go
```diff
@@ -483,6 +490,10 @@ func (f *Client) uploadFromReader(ctx context.Context, reader io.ReaderAt, actua
 	})
 
 	if err != nil {
+		_, err2 := f.Delete(ctx, nil)
```
souravgupta-msft, line 493: Should this code be moved after FlushData operation?

souravgupta-msft: Since, the file is created if `o.EncryptionContext` is not nil. Should we also add this check before deleting the file?

tanyasethi-msft: As discussed offline, 
For 1. Flush is called only when upload is successful, and we are checking upload errors itself. So, better if its not moved after FlushData operation.
For 2. Yes, added a check for the same before deleting the file.

tanyasethi-msft:
Changes in https://github.com/Azure/azure-sdk-for-go/pull/22617

PR #22610, opened 2024-03-18
Fix go identity matrix/pool selection

benbp:
The matrix generation, due to 1es pt limitations, was changed to filter against OS variables and run one generation pass for each OS type. This had a mismatch with the identity matrix, so this PR updates the config to work with the generation jobs filters.

### DIFF
--- a/sdk/azidentity/ci.yml
+++ b/sdk/azidentity/ci.yml
@@ -39,3 +39,6 @@ extends:
             GenerateVMJobs: true
             Path: sdk/azidentity/managed-identity-matrix.json
             Selection: sparse
+        MatrixReplace:
+          - Pool=.*LINUXPOOL.*/azsdk-pool-mms-ubuntu-2204-identitymsi
+          - OSVmImage=.*LINUXPOOL.*/azsdk-pool-mms-ubuntu-2204-1espt
--- a/sdk/azidentity/managed-identity-matrix.json
+++ b/sdk/azidentity/managed-identity-matrix.json
@@ -9,7 +9,7 @@
                 }
             },
             "GoVersion": [
-                "1.22"
+                "1.22.1"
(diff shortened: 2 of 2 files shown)


sdk/azidentity/ci.yml
```diff
@@ -39,3 +39,6 @@ extends:
             GenerateVMJobs: true
             Path: sdk/azidentity/managed-identity-matrix.json
```
weshaggard, line 40: I'm not sure I understand what this actually buys us. Isn't it equivalent?

benbp: @weshaggard the change to support hardcoded os types had us add filters that don't match the hardcoded pool name for identity (https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/pipelines/templates/jobs/generate-job-matrix.yml#L48-L57), as the filters expect the env var format instead. We could update the filters (for storage and identity pool names) or also consider adding another env var to image.yml that matches e.g. `LINUXAZIDENTITYPOOL`.

weshaggard: OK so it is the fact that those are looking for the env names in the matrix to determine. This seems like it might be a little order dependent and magical but I'm OK with it for now. We probably should document that somewhere to state what string matching we are dependent on in the matrix.

weshaggard (approved):

sdk/azidentity/ci.yml
benbp: We have some comments in the configs and the matrix documentation also specifies that we process filter/replace against the environment variable keys, not the resolved values.

benbp: I think going forward I will make an update to image.yml for the repo, as that keeps things all in one tidy place.

PR #22625, opened 2024-03-21
Paginated directory deletes

tanyasethi-msft:

### DIFF
--- a/sdk/storage/azdatalake/directory/client.go
+++ b/sdk/storage/azdatalake/directory/client.go
@@ -262,9 +262,14 @@ func (d *Client) Create(ctx context.Context, options *CreateOptions) (CreateResp
 // Delete deletes directory and any path under it.
 func (d *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResponse, error) {
 	lac, mac, deleteOpts := path.FormatDeleteOptions(options, true)
-	resp, err := d.generatedDirClientWithDFS().Delete(ctx, deleteOpts, lac, mac)
-	err = exported.ConvertToDFSError(err)
-	return resp, err
+	for {
+		resp, err := d.generatedDirClientWithDFS().Delete(ctx, deleteOpts, lac, mac)
+		if resp.Continuation == nil {
+			err = exported.ConvertToDFSError(err)
+			return resp, err
+		}
+		deleteOpts.Continuation = resp.Continuation
+	}
 }
 
 // GetProperties gets the properties of a directory.
--- a/sdk/storage/azdatalake/directory/client.go
+++ b/sdk/storage/azdatalake/directory/client.go
@@ -264,7 +264,7 @@ func (d *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResp
 	lac, mac, deleteOpts := path.FormatDeleteOptions(options, true)
 	for {
 		resp, err := d.generatedDirClientWithDFS().Delete(ctx, deleteOpts, lac, mac)
-		if resp.Continuation == nil {
+		if resp.Continuation == nil || err != nil {
 			err = exported.ConvertToDFSError(err)
 			return resp, err
 		}
--- a/sdk/storage/azdatalake/directory/client_test.go
+++ b/sdk/storage/azdatalake/directory/client_test.go
@@ -773,15 +773,13 @@ func (s *RecordedTestSuite) TestDeleteDirWithNilAccessConditions() {
 // and entityId needs to be set to the entity ID of the application.
 func (s *RecordedTestSuite) TestDeleteDirWithPaginatedDelete() {
 
-	if recording.GetRecordMode() == recording.LiveMode {
-		s.T().Skip("AAD app not configured for this test, this will be skipped")
-	}
+	s.T().Skip("AAD app not configured for this test, this will be skipped")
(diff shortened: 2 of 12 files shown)


sdk/storage/azdatalake/internal/generated/autorest.md
```diff
@@ -298,3 +298,17 @@ directive:
...
+```yaml
+directive:
+- from: 
+  - zz_service_client.go
+  - zz_filesystem_client.go 
+  - zz_path_client.go
+  where: $
+  transform: >-
+    return $.
+      replaceAll(`[]string{"2023-05-03"}`, `[]string{ServiceVersion}`).
```
vibhansa-msft, line 312: Why we need transformation for the service version here?

tanyasethi-msft: This Paginated delete feature is applicable for service version `2023-08-03` (STG90) and later. The swagger, though has not changed in STG89 and 90, so, the service version in latest swagger is `2023-05-03` (STG89).
To support STG90 features, this transformation is needed.

sdk/storage/azdatalake/internal/path/models.go
```diff
@@ -28,6 +29,9 @@ func FormatDeleteOptions(o *DeleteOptions, recursive bool) (*generated.LeaseAcce
 	if o == nil {
 		return nil, nil, deleteOpts
 	}
+	if o.Paginated != nil {
+		deleteOpts.Paginated = o.Paginated
+	}
```
souravgupta-msft, line 34: You can directly do, `deleteOpts.Paginated = o.Paginated` instead of checking if it is nil or not.

sdk/storage/azdatalake/directory/client.go
```diff
@@ -262,9 +262,14 @@ func (d *Client) Create(ctx context.Context, options *CreateOptions) (CreateResp
 // Delete deletes directory and any path under it.
 func (d *Client) Delete(ctx context.Context, options *DeleteOptions) (DeleteResponse, error) {
 	lac, mac, deleteOpts := path.FormatDeleteOptions(options, true)
-	resp, err := d.generatedDirClientWithDFS().Delete(ctx, deleteOpts, lac, mac)
-	err = exported.ConvertToDFSError(err)
-	return resp, err
+	for {
+		resp, err := d.generatedDirClientWithDFS().Delete(ctx, deleteOpts, lac, mac)
+		if resp.Continuation == nil {
```
souravgupta-msft, line 267: Add a check for err also.
```suggestion
		if resp.Continuation == nil || err != nil {
```

sdk/storage/azdatalake/directory/client_test.go
```diff
@@ -767,6 +769,72 @@ func (s *RecordedTestSuite) TestDeleteDirWithNilAccessConditions() {
...
+	_require.NoError(err)
+
+	dirClient, err := directory.NewClientWithSharedKeyCredential(dirURL, credential, nil)
+	_require.NoError(err)
+
+	resp, err := dirClient.Create(context.Background(), nil)
+	_require.NoError(err)
+	_require.NotNil(resp)
+
+	for i := 0; i < 5; i++ {
```
souravgupta-msft, line 803: Are we getting continuation token for only 5 files here?

sdk/storage/azdatalake/directory/client_test.go
```diff
@@ -767,6 +769,72 @@ func (s *RecordedTestSuite) TestDeleteDirWithNilAccessConditions() {
...
+		fileClient, err := dirClient.NewFileClient(testcommon.GenerateFileName(testName) + strconv.Itoa(i))
+		_require.NoError(err)
+		_require.NotNil(fileClient)
+
+		_, err = fileClient.Create(context.Background(), nil)
+		_require.NoError(err)
+	}
+
+	accessControlResp, err := dirClient.GetAccessControl(context.Background(), nil)
```
souravgupta-msft, line 812: I think this should be done for roorDirClient, since SetAccessControl is done at root level.

sdk/storage/azdatalake/directory/client_test.go
```diff
@@ -767,6 +769,72 @@ func (s *RecordedTestSuite) TestDeleteDirWithNilAccessConditions() {
 	_require.NotNil(resp)
 }
 
+func (s *UnrecordedTestSuite) TestDeleteDirWithPaginatedDelete() {
```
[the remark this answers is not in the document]
tanyasethi-msft: Test will be recorded, but it will be skipped by default, because it requires a customer AAD app that don't need to be configured in live test pipelines.

sdk/storage/azdatalake/directory/client_test.go
tanyasethi-msft: My bad, had reduced to 5 for testing. Reverted it back to > 5k files

sdk/storage/azdatalake/directory/client_test.go
tanyasethi-msft: Made the change, thanks

sdk/storage/azdatalake/internal/path/models.go
tanyasethi-msft: Made the change

souravgupta-msft (approved):

PR #22631, opened 2024-03-22
[aznamespaces] Fixing a brittle test that depended on the server-side error message

richardpark-msft:


sdk/messaging/eventgrid/aznamespaces/client_test.go
```diff
@@ -312,11 +312,17 @@ func TestSimpleErrors(t *testing.T) {
 	_, err := c.PublishCloudEvents(context.Background(), c.TestVars.Topic, []messaging.CloudEvent{
 		{},
 	}, nil)
-	var respErr *azcore.ResponseError
 
+	// this'll lead to an error message like this:
+	// "error": {
+	// 	"code":"BadRequest",
```
jhendrixMSFT, line 318: Is there a stable error code we can check for?

jhendrixMSFT (approved):

sdk/messaging/eventgrid/aznamespaces/client_test.go
jhendrixMSFT: But error code?

richardpark-msft: Added (and in another spot as well)

PR #22638, opened 2024-03-26
[azsystemevents] Regenerating based on c25006263b7c15ad5da755a5ef864e4301e56b54

richardpark-msft:
Added events ACSRouterWorkerUpdatedEventData and ACSAdvancedMessageDeliveryStatusUpdatedEventData.


sdk/messaging/eventgrid/azsystemevents/internal/version.go
```diff
@@ -9,5 +9,5 @@ const (
 	ModuleName = "azsystemevents"
 
 	// ModuleVersion is the semantic version (see http://semver.org) of this module.
-	ModuleVersion = "v0.2.0"
+	ModuleVersion = "v0.2.1"
```
jhendrixMSFT, line 12: I think this should be `v0.3.0` due to breaking changes.

sdk/messaging/eventgrid/azsystemevents/error.go
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azsystemevents
+
+// Error is an error that is included as part of a system event.
+type Error struct {
+	// Code is an error code from the service producing the system event.
+	Code string
+
+	message string
```
jhendrixMSFT, line 14: Debating if this should be exported for fakes purposes?

richardpark-msft: @JeffreyRichter and I had a bit of a back and forth on this. I totally don't mind it being exported. Jeff, what do you think?

JeffreyRichter: I don't know enough about how fakes work with AMQP.  To me, this is like how we only expose HTTP status code and code from ResponseError but it holds on to the raw JSON error response body. This has a way of working with fakes, right? If so, I'd expect this to work the same way.

richardpark-msft: Just to be clear, this isn't AMQP, it's protocol independent (the events can be received from anywhere). 

@jhendrixMSFT , thinking about this more this is the 'message' part, which is considered the cosmetic part. They can already change Code, which is really the only piece they should use.

So maybe we leave this one unexported.

jhendrixMSFT: That's fine, we can always export it later if required.

sdk/messaging/eventgrid/azsystemevents/custom_events_test.go
jhendrixMSFT, line 1: Missing copyright header.

jhendrixMSFT (approved):

PR #22643, opened 2024-03-27
[azeventgrid(old)] Officially deprecating this module

### LINKED ISSUE #22554: Remove modules for messaging/azeventgrid and cognitiveservices/azopenai that are either deprecated or retracted
We've published their replacements at this point. 

- [ ] messaging/azeventgrid needs one more publish to deprecate in favor of either eventgrid/azeventgrid or aznamespaces. 
- [x] cognitiveservices/azopenai should be deletable entirely as we've retracted the package and we tagged the retraction.

richardpark-msft:
Both packages that were hosted in this module have moved and been published in their new places:

* Event Grid Basic client has moved to github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/azeventgrid
* Event Grid Namespaces client has moved to github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/aznamespaces

We can deprecate this, publish it one more time and then scrub it from the repo.

Fixes #22554

### DIFF
--- a/sdk/messaging/azeventgrid/go.mod
+++ b/sdk/messaging/azeventgrid/go.mod
@@ -1,3 +1,5 @@
+// Deprecated: Event Grid Basic client has moved to github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/azeventgrid
+// Deprecated: Event Grid Namespaces client has moved to use github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/aznamespaces
 module github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventgrid
 
 go 1.18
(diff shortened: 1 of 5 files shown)


sdk/messaging/azeventgrid/go.mod
```diff
@@ -1,3 +1,5 @@
+// Deprecated: Event Grid Basic client has moved to github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/azeventgrid
+// Deprecated: Event Grid Namespaces client has moved to github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/aznamespaces
```
chlowell, line 2: Do we want the second `Deprecated:`? The [docs](https://go.dev/ref/mod#go-mod-file-module-deprecation) imply it's redundant and would appear as ordinary text in the message, but I dunno because I haven't seen a published case like this.

richardpark-msft: Hm.

> A module can be marked as deprecated in a block of comments containing the string Deprecated: (case-sensitive) at the beginning of a paragraph.

Now I'm wondering what that "a block of comments" actually means. We do need to list both modules, but I could list them in one long sentence to be safe. I will check.

chlowell (approved):

sdk/messaging/azeventgrid/go.mod
richardpark-msft: Updated to just avoid the discussion :)

PR #22646, opened 2024-03-27
The EventReader buffer size is now configurable

ChrisTrenkamp:
When streaming answers with BYOD, it's possible for a Server-side event to be sent with a larger size than the `EventReader` scanner's default buffer size, and the user has no way to configure this.

This MR adds a `Buffer` function to `EventReader` so you can set the buffer to a custom size.

### DIFF
--- a/sdk/ai/azopenai/CHANGELOG.md
+++ b/sdk/ai/azopenai/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features Added
 
 - Updating to the `2024-03-01-preview` API version. This adds support for using Dimensions with Embeddings as well as the ability to choose the embeddings format. PR(#22603)
+- `EventReader` now has a `Buffer` method that controls the scanner buffer size when streaming Server-sent event's.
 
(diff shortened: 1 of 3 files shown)


richardpark-msft:
Hi @ChrisTrenkamp, I'll take a look at this. The general rule is we want things to immutable so having a function configure it is probably not the way we'd go (I'll have to look if there are alternatives).

Just want you to know that I saw this, haven't forgotten it - it's on my list.

richardpark-msft:
I'm closing this in favor of https://github.com/Azure/azure-sdk-for-go/pull/22703, which uses bufio.Reader and should accomplish the same thing, but without requiring any guess work on your side as to which buffer size to use. 

Your contribution was much appreciated though - I wouldn't have noticed this issue. I've credited you in the [CHANGELOG](https://github.com/Azure/azure-sdk-for-go/pull/22703/files#diff-64ecdee0caadc0843c22418ca3cf971578ae86aad405919629b524f1a2c3f051R11) as well.

PR #22650, opened 2024-03-28
Correct IMDS resource ID query parameter

chlowell:
IMDS observes both `msi_res_id` and `mi_res_id` as the query parameter specifying an identity by its resource ID. We should prefer `msi_res_id` because it's the [documented parameter](https://github.com/Azure/azure-rest-api-specs/blob/dba6ed1f03bda88ac6884c0a883246446cc72495/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2018-10-01/imds.json#L233-L239) and because Azure Container Instances, which has a different implementation based on the IMDS API, observes only `msi_res_id`.

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -34,14 +34,14 @@ const (
 	identityServerThumbprint = "IDENTITY_SERVER_THUMBPRINT"
 	headerMetadata           = "Metadata"
 	imdsEndpoint             = "http://169.254.169.254/metadata/identity/oauth2/token"
+	miResID                  = "mi_res_id"
 	msiEndpoint              = "MSI_ENDPOINT"
+	msiResID                 = "msi_res_id"
 	msiSecret                = "MSI_SECRET"
 	imdsAPIVersion           = "2018-02-01"
 	azureArcAPIVersion       = "2019-08-15"
+	qpClientID               = "client_id"
 	serviceFabricAPIVersion  = "2019-07-01-preview"
-
-	qpClientID = "client_id"
-	qpResID    = "mi_res_id"
 )
 
 type msiType int
@@ -286,7 +286,7 @@ func (c *managedIdentityClient) createIMDSAuthRequest(ctx context.Context, id Ma
 	q.Add("resource", strings.Join(scopes, " "))
 	if id != nil {
 		if id.idKind() == miResourceID {
-			q.Add(qpResID, id.String())
+			q.Add(msiResID, id.String())
 		} else {
 			q.Add(qpClientID, id.String())
 		}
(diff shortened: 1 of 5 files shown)


jhendrixMSFT (approved):

jhendrixMSFT:
Does this need a changelog entry?

sdk/azidentity/managed_identity_client.go
```diff
@@ -34,14 +34,14 @@ const (
 	identityServerThumbprint = "IDENTITY_SERVER_THUMBPRINT"
 	headerMetadata           = "Metadata"
 	imdsEndpoint             = "http://169.254.169.254/metadata/identity/oauth2/token"
+	miResID                  = "mi_res_id"
```
ahsonkhan, line 37: How come this is still needed (in what scenarios)?

I can't easily tell from the logic below where we'd use this one vs the MSI one. 
Both seem to be wrapped around `if id.idKind() == miResourceID`

sdk/azidentity/managed_identity_client.go
```diff
@@ -437,7 +437,7 @@ func (c *managedIdentityClient) createCloudShellAuthRequest(ctx context.Context,
 		log.Write(EventAuthentication, "WARNING: Cloud Shell doesn't support user-assigned managed identities")
 		q := request.Raw().URL.Query()
 		if id.idKind() == miResourceID {
-			q.Add(qpResID, id.String())
+			q.Add(miResID, id.String())
```
ahsonkhan, line 440: Most places seem to keep using the mi string, instead of msi, but the PR description is implying we should convege and only use msi.

chlowell: We need both. We should use "msi_res_id" for IMDS (and implicitly, anything resembling IMDS). For other platforms such as App Service, "mi_res_id" is correct.

chlowell: Well, this was interesting. The App Service doc was changed after we implemented this feature because a reader saw that other APIs use `msi_res_id` and concluded `mi_res_id` was a typo. It isn't: `mi_res_id` is correct and App Service ignores `msi_res_id`. I opened https://github.com/MicrosoftDocs/azure-docs/pull/121176 to fix the doc.

sdk/azidentity/managed_identity_client.go
chlowell: We still need this for App Service. Each platform we support has a different API, so the query parameter we need to set depends on the platform. `id` in the line you pasted is the value of the parameter i.e. the resource ID.

tasnef:
We are the acknowledged author of the documents

Did you ask Azure? What documents does it have, and what work has it done to develop it؟

PR #22653, opened 2024-03-29
[azservicebus] Fixing a long-standing bug where we wouldn't recover (inline) on amqp.Receiver.Receive() failures

richardpark-msft:
Fixing a long-standing bug where the retry loop didn't cover the entire ReceiveMessage() function, making it so we wouldn't retry if a retryable failure came from amqp.Receiver.Receive().

### DIFF
--- a/sdk/messaging/azservicebus/CHANGELOG.md
+++ b/sdk/messaging/azservicebus/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Other Changes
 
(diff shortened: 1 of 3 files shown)


richardpark-msft:
(I've got another test coming in to make sure we don't launch multiple "releaser" goroutines)

richardpark-msft:
There's more (good) work to do here because eliminating this actually eliminates some redundant code paths. No rush as this is still handled, even if it's a little goofy today.

PR #22675, opened 2024-04-02
runtime.Poller.Result won't be done on non-terminal error

jhendrixMSFT:
If the underlying error when retrieving the final result is non terminal, don't mark the poller as done so callers can retry.

### DIFF
--- a/sdk/azcore/internal/pollers/util.go
+++ b/sdk/azcore/internal/pollers/util.go
@@ -189,5 +189,12 @@ func ResultHelper[T any](resp *http.Response, failed bool, out *T) error {
 // IsNonTerminalHTTPStatusCode returns true if the HTTP status code should be
 // considered non-terminal thus eligible for retry.
 func IsNonTerminalHTTPStatusCode(resp *http.Response) bool {
-	return resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
+	return exported.HasStatusCode(resp,
+		http.StatusRequestTimeout,      // 408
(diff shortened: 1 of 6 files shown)


sdk/azcore/internal/pollers/util.go
```diff
@@ -185,3 +185,9 @@ func ResultHelper[T any](resp *http.Response, failed bool, out *T) error {
 	}
 	return nil
 }
+
+// IsNonTerminalHTTPStatusCode returns true if the HTTP status code should be
+// considered non-terminal thus eligible for retry.
+func IsNonTerminalHTTPStatusCode(resp *http.Response) bool {
```
jhendrixMSFT, line 191: I wonder if this should be expanded to include all of the status codes [here](https://github.com/Azure/azure-sdk-for-go/pull/22675) which also happens to be our default list of HTTP status codes for retry.

jhendrixMSFT: Done.

chlowell (approved):

sdk/azcore/internal/pollers/util.go
```diff
@@ -185,3 +185,9 @@ func ResultHelper[T any](resp *http.Response, failed bool, out *T) error {
...
+	return resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
```
hong-db: Per Azure retry guidance, 429 and all 5xx errors can be retried. Can we expand the error code range?

hong-db (approved):
I see the expanded error codes.

PR #22676, opened 2024-04-02
DefaultAzureCredential probes IMDS before sending it a token request

### LINKED ISSUE #22265: Improve Probing Logic for Azure Instance Metadata Service (IMDS) for Virtual Machines
Tracking issue for addressing customer feedback on IMDS VM delays with DefaultAzureCredential due to IMDS probing retry logic in Azure Identity SDK.  @ahsonkhan is driving this investigation.

Currently our SDKs are inconsistent across languages where some languages have zero retries, some have a single 0.5 second retry, and others (.NET) have up to 8-seconds of retries.

We need to review all languages here and make them consistent across languages, as well as publish a Sample and Blog about IDMS on VMs with Azure Identity SDK and how customers can write 2-3 lines of cose to skip IMDS when needed.

chlowell:
Because requests sent to IMDS can take a very long time to time out when IMDS isn't available, DefaultAzureCredential imposes a 1 second timeout on managed identity token requests until one receives a response. This approach creates a (very small) chance of canceling a token request that would have succeeded given more time. This PR lowers that probability by adding a probing step in which the credential sends a malformed request to IMDS with the same 1 second timeout. Request validation is faster than acquiring an access token, so we assume IMDS will respond faster to this probe than to a token request. If the credential receives any response to a probe, it takes that to mean IMDS is available and thereafter sends normal token requests with no special timeout. This affects only DefaultAzureCredential. ManagedIdentityCredential doesn't probe or impose any timeout when used directly. Closes #22265

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -182,6 +182,8 @@ func (c *managedIdentityClient) provideToken(ctx context.Context, params confide
 
 // authenticate acquires an access token
 func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKind, scopes []string) (azcore.AccessToken, error) {
+	// no need to synchronize around this value because it's true only when DefaultAzureCredential constructed the client,
+	// and in that case ChainedTokenCredential.GetToken synchronizes goroutines that would execute this block
 	if c.probeIMDS {
 		cx, cancel := context.WithTimeout(ctx, imdsProbeTimeout)
 		defer cancel()
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -44,6 +44,8 @@ const (
 	serviceFabricAPIVersion  = "2019-07-01-preview"
 )
 
+var imdsProbeTimeout = time.Second
+
 type msiType int
 
 const (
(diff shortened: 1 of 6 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -180,6 +182,24 @@ func (c *managedIdentityClient) provideToken(ctx context.Context, params confide
 
 // authenticate acquires an access token
 func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKind, scopes []string) (azcore.AccessToken, error) {
+	if c.probeIMDS {
```
jhendrixMSFT, line 185: Is one instance of `*managedIdentityClient` shared? i.e. do we need to synchronize this?

chlowell: Good question. Goroutines can execute this method concurrently but this block executes only in `ChainedTokenCredential.GetToken`--[which is synchronized](https://github.com/Azure/azure-sdk-for-go/blob/aef767840f769b4a9d3925b5682f57fe888201e5/sdk/azidentity/chained_token_credential.go#L65-L68)--because otherwise the condition is false. I'll comment on that.

jhendrixMSFT (approved):

PR #22681, opened 2024-04-02
[(old)azeventgrid] delete old package. It's been released (deprecated) and should no longer be used

richardpark-msft:


jhendrixMSFT (approved):

richardpark-msft:
Fairly certain I have to override checks on this since I'm removing the folder.

PR #22683, opened 2024-04-02
[azsystemevents] Updating to release commit for April ca560e400087f54b306f3a21df582c8a5bbd0abc

richardpark-msft:
* Renames that came out of the arch board.
* 'string' became a fleshed out enumeration

### DIFF
--- a/sdk/messaging/eventgrid/azsystemevents/constants.go
+++ b/sdk/messaging/eventgrid/azsystemevents/constants.go
@@ -1049,17 +1049,17 @@ func PossibleMediaJobStateValues() []MediaJobState {
 	}
 }
 
-// RecordingChannelType - The recording channel type - Mixed, Unmixed
-type RecordingChannelType string
+// RecordingChannelKind - The recording channel type - Mixed, Unmixed
+type RecordingChannelKind string
 
 const (
-	RecordingChannelTypeMixed   RecordingChannelType = "Mixed"
-	RecordingChannelTypeUnmixed RecordingChannelType = "Unmixed"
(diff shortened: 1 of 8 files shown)


sdk/messaging/eventgrid/azsystemevents/constants.go
```diff
@@ -148,7 +172,6 @@ const (
 	ACSRouterUpdatedWorkerPropertyAvailableForOffers    ACSRouterUpdatedWorkerProperty = "AvailableForOffers"
 	ACSRouterUpdatedWorkerPropertyChannelConfigurations ACSRouterUpdatedWorkerProperty = "ChannelConfigurations"
 	ACSRouterUpdatedWorkerPropertyLabels                ACSRouterUpdatedWorkerProperty = "Labels"
-	ACSRouterUpdatedWorkerPropertyMaxConcurrentOffers   ACSRouterUpdatedWorkerProperty = "MaxConcurrentOffers"
```
chlowell, line 151: Should this, ChannelType -> ChannelKind, and the model field renames be in the changelog?

chlowell (approved):

sdk/messaging/eventgrid/azsystemevents/constants.go
richardpark-msft: Well, "should" was NOT true :) I've updated the changelog to incorporate all the renames. It wasn't too many. I also added in a rename for the RecordingChannelType (which was the only one I saw that hadn't gotten the rename treatment).

PR #22707, opened 2024-04-05
[azopenai] A rather large, but needed, cleanup of the tests. 

richardpark-msft:
It was getting difficult to tell what was and wasn't covered and also correlate tests to models, which can sometimes come with differing behavior.

So in this PR I:
- Moved things that don't need to be env vars (ie: non-secrets) to just be constants in the code - this includes model names and generic server identifiers for regions that we use to coordinate new feature development.
- Consolidated all of the setting of endpoint and models into one spot to make it simpler to double-check.
- Consolidated tests that tested the same thing into sub-tests with OpenAI or AzureOpenAI names.
  - If a function was only called by one test moved it into the test as an anonymous func

Also, I added in a test for logit_probs and logprobs/toplogprobs.


richardpark-msft (commented):
/azp run go - azopenai

richardpark-msft:
(There's an issue with the live pipelines right now that's unrelated to my change. Here's the manually queued pipeline, which should get around it: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3703444&view=results)

jhendrixMSFT (approved):

PR #22713, opened 2024-04-08
Upgrade keyvault/internal dependencies

chlowell:
This got a little interesting because the new azcore has BearerTokenPolicy ensure all the errors it returns are marked non retriable. That doesn't cause a functional regression for the challenge policy but it breaks a test that asserts error types without unwrapping. I updated the test to check whether the challenge policy's errors are non retriable--which is all we care about--and simplified the challenge policy to return plain old errors.


sdk/security/keyvault/internal/challenge_policy.go
```diff
@@ -105,29 +104,11 @@ func parseTenant(url string) string {
 	return tenant
 }
 
-type challengePolicyError struct {
```
jhendrixMSFT, line 108: So this was just redundant?

chlowell: Yes. I just edited an explanation into the top comment but the gist is, this type existed to make sure the policy returns non retriable errors. BearerTokenPolicy does that in the new azcore, so we don't need this type anymore.

jhendrixMSFT (approved):

gracewilcox (approved):

PR #22623, opened 2024-03-20
azcosmos - 1.0.0 review

ealsur:
This PR updates the version to the upcoming 1.0 release.

The goal is to also be used for final API review: https://apiview.dev/Assemblies/Review/8950b757ef3544f59d3fed21774a7966/44c2850f19944fcf84f512600078a1e7

The original API surface was [reviewed through [archboard](https://github.com/Azure/azure-sdk/issues/3379), after that, the subsequent added new feature APIs were also reviewed and approved:

* [Query pager](https://github.com/Azure/azure-sdk-for-go/pull/17657)
* [TransactionalBatch API](https://github.com/Azure/azure-sdk-for-go/pull/17795)
* [Connection String support](https://github.com/Azure/azure-sdk-for-go/pull/18549)
* [Parametrized queries](https://github.com/Azure/azure-sdk-for-go/pull/18577)
* [Patch API support](https://github.com/Azure/azure-sdk-for-go/pull/19628)
* [Database and Container Query pagers](https://github.com/Azure/azure-sdk-for-go/pull/19769)
* [PartitionKey.Null](https://github.com/Azure/azure-sdk-for-go/pull/20002)
* [ClientOptions.PreferredRegions](https://github.com/Azure/azure-sdk-for-go/pull/22206)

PR #22720, opened 2024-04-09
azcosmos - Fix continuations types

ealsur:
During https://github.com/Azure/azure-sdk-for-go/pull/22623 we fixed `ContinuationToken` to be `*string` but we missed some other places.

* QueryContainersOptions
* QueryDatabasesOptions
* QueryContainersResponse
* QueryDatabasesResponse


jhendrixMSFT (approved):
Does this need a changelog entry?

ealsur:
> Does this need a changelog entry?

Nice catch, will add

Issue #22744, opened 2024-04-12
Remove vcr based recording API


chlowell:
`internal/recording` has two parallel recording APIs, one built on our test proxy and another built on `go-vcr`. We should remove the latter because we stopped supporting it after migrating all our recordings to the test proxy.

PR #22745, opened 2024-04-12
Default test recording sanitizers

chlowell:
These will apply to all future recordings made through the proxy but don't affect existing recordings. Some of the new target values may not be secret, however we're scrubbing them all anyway to prevent false positives. We can revisit them later if sanitizing them causes any problems.

I marked `NewRecording` deprecated because it doesn't use the test proxy and isn't called anywhere in the repo. We should delete it and the rest of the vestigial `go-vcr` based code because we only support recording via the proxy (#22744).

Also, I reverted #21142 because it pinned our CI pipelines to an old version of the proxy that doesn't support adding sanitizers in bulk, apparently to work around an azdatalake issue. The azdatalake pipelines pass with this change so I guess the underlying problem has been solved? (@scbedd @benbp)

### DIFF
--- a/sdk/internal/recording/default_sanitizers.go
+++ b/sdk/internal/recording/default_sanitizers.go
@@ -193,4 +193,13 @@ var defaultSanitizers = []sanitizer{
 	newBodyRegexSanitizer("-----BEGIN PRIVATE KEY-----\\\\n(?<key>.+\\\\n)*-----END PRIVATE KEY-----\\\\n", "key"),
 	newBodyKeySanitizer("$..adminPassword.value"),
 	newBodyKeySanitizer("$..decryptionKey"),
+	newBodyRegexSanitizer("(?<=<UserDelegationKey>).*?(?:<Value>)(.*)(?:</Value>)", ""),
+	newBodyRegexSanitizer("(?<=<UserDelegationKey>).*?(?:<SignedTid>)(.*)(?:</SignedTid>)", ""),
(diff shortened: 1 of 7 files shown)


sdk/internal/recording/default_sanitizers.go
```diff
@@ -0,0 +1,205 @@
...
+		return err
+	}
+	req.Body = io.NopCloser(bytes.NewReader(b))
+	req.ContentLength = int64(len(b))
+	req.Header.Set("Content-Type", "application/json")
+	return handleProxyResponse(client.Do(req))
+}
+
+var defaultSanitizers = []sanitizer{
+	newGeneralRegexSanitizer(`("|;)Secret=(?<secret>[^;]+)`, "secret"),
```
RickWinter, line 100: Nit: Group the Sanitizers based on 
newGeneral
newBody
newHeader
etc...

sdk/internal/recording/recording.go
```diff
@@ -84,7 +85,12 @@ const (
 	Secret_Base64String VariableType = "secret_base64String"
 )
 
+// defaultSanitizersSet tracks whether default sanitizers have been added (this would be neater with Go 1.19's atomic.Bool)
```
RickWinter, line 88: Lets folow up and move to 1.19. Afterall 1.22.2 and 1.21.9 are the only in support versions on golang.org

RickWinter (approved):

sdk/internal/recording/recording.go
jhendrixMSFT: Will there be a ripple effect, requiring us to update everything? Might be best to do that separately.

PR #22752, opened 2024-04-16
Upgrade modules to sdk/internal v1.6.0

chlowell:
This ensures the default sanitizers added in `internal` v1.6.0 will apply the next time a test is recorded.


chlowell:
The azeventgrid failure is due to a broken link unrelated to this change. I don't want to fix it in this PR because that would mean running all the pipelines again 😅

jhendrixMSFT (approved):

richardpark-msft:
> The azeventgrid failure is due to a broken link unrelated to this change. I don't want to fix it in this PR because that would mean running all the pipelines again 😅

I'll post a PR for that one.

PR #22759, opened 2024-04-17
Deprecate go-vcr based recording API

chlowell:
...because we only support recording tests via the test proxy. I didn't find anything calling these methods on main, but because we don't want any module to use them for any reason, I've deleted their implementations and made them return errors or panic.


sdk/internal/recording/recording.go
```diff
@@ -85,340 +79,53 @@ const (
...
-
-	// set the recorder Matcher
-	recording.Matcher = defaultMatcher(c)
-	rec.SetMatcher(recording.matchRequest)
-
-	// wire up the sanitizer
-	recording.Sanitizer = defaultSanitizer(rec)
-
-	return recording, err
+	return nil, errUnsupportedAPI
```
RickWinter, line 91: This is the internal module we own and only we can bind to and call. We should remove the APIs entirely.

sdk/internal/recording/recording.go
```diff
@@ -85,340 +79,53 @@ const (
 	Secret_Base64String VariableType = "secret_base64String"
 )
 
-// defaultSanitizersSet tracks whether default sanitizers have been added (this would be neater with Go 1.19's atomic.Bool)
-var defaultSanitizersSet int32
+var (
+	// defaultSanitizersSet tracks whether default sanitizers have been added (this would be neater with Go 1.19's atomic.Bool)
+	defaultSanitizersSet int32
+
+	errUnsupportedAPI = errors.New("the vcr based recording API isn't supported. Use the test proxy instead")
```
jhendrixMSFT, line 86: Maybe add the word "test" in here just in case, i.e. `...based test recording...`

sdk/internal/CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Release History
 
-## 1.6.1 (Unreleased)
+## 1.7.0 (Unreleased)
 
 ### Features Added
 
 ### Breaking Changes
+* Deprecated the `go-vcr` based recording API. Its methods now return errors or panic.
```
jhendrixMSFT, line 8: `...based test recording...`

RickWinter (approved):
I would prefer we removed all the code. Afterall its in an internal module we own and control

jhendrixMSFT (approved):

sdk/internal/recording/recording.go
jhendrixMSFT: That's a breaking change, and I'm worried there will be some corner case that causes builds to break (@chlowell and I extensively discussed this).

jhendrixMSFT: I'll also add that we would never introduce breaking changes in production code outside of a major version change. We should hold our internal tooling to the same bar.

PR #22768, opened 2024-04-19
[Monitor Ingestion] update README

gracewilcox:
Update README

### DIFF
--- a/sdk/monitor/ingestion/azlogs/README.md
+++ b/sdk/monitor/ingestion/azlogs/README.md
@@ -1,16 +1,16 @@
 # Azure Monitor Ingestion client module for Go
 
-The Azure Monitor Ingestion client module is used to send custom logs to [Azure Monitor][azure_monitor_overview] using the [Logs Ingestion API][ingestion_overview].
+* Upload logs (this module): send custom logs to Azure Monitor using the Logs Ingestion API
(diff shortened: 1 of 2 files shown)


sdk/monitor/ingestion/azlogs/README.md
```diff
@@ -1,16 +1,16 @@
...
-
-Source code | Package (pkg.go.dev) | [Product documentation][azure_monitor_overview] | Samples
+[Source code][source_code] | [Package (pkg.go.dev)][pkg_go] | [Product documentation][azure_monitor_overview] | [Samples][examples]
 
 ## Getting started
 
 ### Prerequisites
 
-* A supported Go version (the Azure SDK supports the two most recent Go releases)
+* Go version 1.18 or higher
```
chlowell, line 13: I'd keep the current text here because although today most modules require 1.18, we only promise to support the latest two versions and it would be easy to miss updating this line when we raise the requirement.

chlowell (approved):

sdk/monitor/ingestion/azlogs/README.md
gracewilcox: fixed in [eb29085](https://github.com/Azure/azure-sdk-for-go/pull/22768/commits/eb29085d3bb2f85f5b2d9a414aaf7f00ff7a0344)

PR #22784, opened 2024-04-26
[azservicebus] Adding in the DeleteMessages feature. This is a new feature in preview.

### LINKED ISSUE #22800: [azservicebus] Implement the new DeleteMessages operation
DeleteMessages is a new feature, being added in preview, that lets you delete messages without having to transfer and Complete() them on the client side. 

- [ ] Build the code
- [ ] Add links back to documentation that show things like limits.

richardpark-msft:
Fixes #22800

### DIFF
--- a/sdk/messaging/azservicebus/receiver.go
+++ b/sdk/messaging/azservicebus/receiver.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -55,6 +56,11 @@ type Receiver struct {
 	receiving                bool
 	retryOptions             RetryOptions
 	settler                  *messageSettler
+
+	// sessionID is the actual session ID for this receiver - it's used in some management
+	// operations that are otherwise the same between session receivers and non-session
+	// receivers.
+	sessionID *string
 }
 
 // ReceiverOptions contains options for the `Client.NewReceiverForQueue` or `Client.NewReceiverForSubscription`
@@ -268,7 +274,7 @@ func (r *Receiver) PeekMessages(ctx context.Context, maxMessageCount int, option
 			updateInternalSequenceNumber = false
 		}
 
-		messages, err := internal.PeekMessages(ctx, links.RPC, links.Receiver.LinkName(), sequenceNumber, int32(maxMessageCount))
+		messages, err := internal.PeekMessages(ctx, links.RPC, links.Receiver.LinkName(), sequenceNumber, int32(maxMessageCount), r.sessionID)
 
 		if err != nil {
 			return err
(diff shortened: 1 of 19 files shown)


sdk/messaging/azservicebus/receiver.go
```diff
@@ -358,6 +364,83 @@ func (r *Receiver) DeadLetterMessage(ctx context.Context, message *ReceivedMessa
 	return r.settler.DeadLetterMessage(ctx, message, options)
 }
 
+// DeleteMessagesOptions contains the optional parameters for the [Client.DeleteMessages] method.
+type DeleteMessagesOptions struct {
+	// Count is the maximum number of messages to delete.
+	// Defaults to 4000.
+	Count int
```
jhendrixMSFT, line 371: Is there a max value for this that needs to be doc'ed?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -358,6 +364,83 @@ func (r *Receiver) DeadLetterMessage(ctx context.Context, message *ReceivedMessa
...
+	if options.Count == 0 {
+		options.Count = 4000
+	}
+
+	count := int64(0)
+
+	err := r.amqpLinks.Retry(ctx, EventReceiver, "DeleteMessages", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
+		value := map[string]any{
+			"enqueued-time-utc": options.BeforeEnqueueTime.UTC(),
+			"message-count":     int32(options.Count),
```
jhendrixMSFT, line 398: This makes me wonder if `Count` should be an `int32`. Will this always be limited to a 32-bit value?

richardpark-msft: Let me check on that. I'll also adjust the Count to be an int64/int32 depending on this as well.

richardpark-msft: So it is limited to an int32 (same value but larger type fails). I've changed Count's type to reflect that.

sdk/messaging/azservicebus/receiver.go
richardpark-msft: Currently it's 4000. However, do we normally document service-side limits on the client? It feels similar to client-side validation (ie, we should let the server control the messaging in it's error).

jhendrixMSFT: Good point. Maybe a link to the docs for this?

sdk/messaging/azservicebus/receiver.go
jhendrixMSFT: The cast is now redundant.

jhendrixMSFT (approved):

sdk/messaging/azservicebus/receiver.go
richardpark-msft: I added it as a task to the overall DeleteMessages issue (which I had no filed)
https://github.com/Azure/azure-sdk-for-go/issues/22800

PR #22789, opened 2024-04-26
Enable tests to disable default sanitizers

chlowell:
The new test proxy version which includes the default sanitizers also includes a new API for disabling them. We'll probably want this while fixing tests, so I'd like to get it out in the next release.

### DIFF
--- a/sdk/internal/recording/sanitizer.go
+++ b/sdk/internal/recording/sanitizer.go
@@ -389,6 +389,33 @@ func AddURISubscriptionIDSanitizer(value string, options *RecordingOptions) erro
 	return handleProxyResponse(client.Do(req))
 }
 
+// RemoveRegisteredSanitizers selectively disables sanitizers that are enabled by default such as "AZSDK1001"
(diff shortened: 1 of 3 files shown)


sdk/internal/recording/sanitizer.go
```diff
@@ -389,6 +389,33 @@ func AddURISubscriptionIDSanitizer(value string, options *RecordingOptions) erro
...
+	if err != nil {
+		return err
+	}
+	url := options.baseURL() + "/Admin/RemoveSanitizers"
+	req, err := http.NewRequest(http.MethodPost, url, io.NopCloser(bytes.NewReader(b)))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.ContentLength = int64(len(b))
```
richardpark-msft, line 414: ContentLength doesn't have to be manually set (NewRequest will handle it for you).

richardpark-msft (approved):

PR #22795, opened 2024-04-29
Eng gogetu all

jhendrixMSFT:

### DIFF
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -12,5 +12,8 @@ variables:
   # Set the default value so that 'Build Performance Tests' step can resolve it
   UseAzcoreFromMain: $[eq(variables['Use.AzcoreFromMain'], 'true')]
 
+  # Set to disable analyze step for requiring latest version of dependencies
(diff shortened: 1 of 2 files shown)


benbp (approved):

richardpark-msft:
@jhendrixMSFT, so basically a pre-PR step (for us) is to make sure we update all deps to the latest?

jhendrixMSFT:
That's correct.

richardpark-msft (approved):
I can't think of any time we've purposefully tried to use an older version of a library so this seems safe...

jhendrixMSFT:
Hmm ok I added support to skip this step in case it becomes a problem (which should be rare, if ever).

eng/pipelines/templates/variables/globals.yml
```diff
@@ -12,5 +12,8 @@ variables:
   # Set the default value so that 'Build Performance Tests' step can resolve it
   UseAzcoreFromMain: $[eq(variables['Use.AzcoreFromMain'], 'true')]
 
+  # Set to disable analyze step for requiring latest version of dependencies
+  SkipGoGetUAll: $[eq(variables['Skip.GoGetUAll'], 'true')]
```
jhendrixMSFT, line 16: @benbp does this logic and above usage look correct?

PR #22832, opened 2024-05-02
Do not create or block on api review for nested packages

benbp:
I do not think there are any pipelines where we need/want to be recursing on packages, as we keep to a one pipeline per package convention in go data plane. Currently nested packages, i.e. azidentity/cache are blocking parent packages from releasing if they don't have an api review approval.

### DIFF
--- a/eng/scripts/Create-ApiReview.ps1
+++ b/eng/scripts/Create-ApiReview.ps1
@@ -24,6 +24,10 @@ $createReviewScript = (Join-Path $PSScriptRoot .. common scripts Create-APIRevie
 $artifactList = @()
 foreach ($sdk in (Get-AllPackageInfoFromRepo $ServiceDirectory))
 {
+    if ($sdk.ServiceDirectory -ne $ServiceDirectory) {
+        Write-Host "Skipping nested package $($sdk.ServiceDirectory)"
+        continue
+    }
     Write-Host "Creating API review artifact for $($sdk.Name)"
     New-Item -ItemType Directory -Path $OutPath/$($sdk.Name) -force
     $fileName = Split-Path -Path $sdk.Name -Leaf


eng/scripts/Create-ApiReview.ps1
```diff
@@ -24,6 +24,10 @@ $createReviewScript = (Join-Path $PSScriptRoot .. common scripts Create-APIRevie
 $artifactList = @()
 foreach ($sdk in (Get-AllPackageInfoFromRepo $ServiceDirectory))
 {
+    if ($sdk.ServiceDirectory -ne $ServiceDirectory) {
```
weshaggard, line 27: I don't have enough context to know if there is a better option but this seems to have a smell to me. Should we be filtering at this level or at the Get-AllPackageInfoFromRepo level, or perhaps some other pipeline filter configuration?

benbp: This is a bit of a quick fix to unblock identity release. I'm going to be looking at a broader recurse/norecurse option at the higher level because it seems that both are used in different places.

chlowell (approved):

RickWinter (changes requested):
How are we verifying the sub-packages have APIView signoff?

chlowell:
It's perhaps confusing here that EngSys scripts like this one often reference "packages" where they mean modules. We create API reviews for modules, not packages, so e.g. `azcore/runtime` appears in the review for `azcore` and will continue to do so with this change because the pipeline will still run apiview automation in `skd/azcore`. What we don't want is to run that automation in a subdirectory of the module root that also contains a `go.mod` i.e., is a separate module. The script does that today because it recursively discovers ~~packages~~ modules.

PR #22833, opened 2024-05-03
[azeventhubs] Fixing broken behavior in Processor

richardpark-msft:
The Processor had incorrect behavior when Run had been stopped:
- Run() now returns an explicit error when called more than once on a single 
  Processor instance.
- NextProcessorClient now properly terminates (and returns nil) if called on a 
  stopped Processor.

Documentation has been clarified on this.


sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,19 @@
 # Release History
 
-## 1.1.1 (Unreleased)
-
-### Features Added
+## 1.2.0 (2024-05-07)
 
 ### Breaking Changes
 
+- Processor.Run() behavior has been clarified and returns errors and handles closing
```
jhendrixMSFT, line 7: Is this really a breaking change?

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,14 +1,19 @@
...
+  properly. See "Bugs Fixed" for more details about specific changes.
+
 ### Bugs Fixed
 
-### Other Changes
+The Processor had incorrect behavior when Run had been stopped:
+- Run() now returns an explicit error when called more than once on a single
```
jhendrixMSFT, line 13: The change looks slightly different than what's described here. It will prevent calling `Run()` on a stopped `Processor` but I don't see it preventing multiple calls to `Run()` on a non-stopped `Processor`. Maybe I missed it?

richardpark-msft: Ah, multiple, not concurrent (maybe `repeated` would be better). I can clarify.

sdk/messaging/azeventhubs/CHANGELOG.md
richardpark-msft: I suppose not. I was worried that people might have been "depending" on the buggy behavior without realizing it.

sdk/messaging/azeventhubs/CHANGELOG.md
richardpark-msft: Also, I should prevent the multiple calls thing too.

sdk/messaging/azeventhubs/CHANGELOG.md
richardpark-msft: (removed)

jhendrixMSFT (approved):

PR #22835, opened 2024-05-03
Upgrade all dependencies

chlowell:
`go get -u all` and `go mod tidy` all modules. This should get pipelines back to green, provided none of them fail for reasons unrelated to dependency versions 😅

PR #22836, opened 2024-05-03
Revert breaking dependency upgrades

chlowell:
This reverts the upgrade from an unstable to stable dependency version that broke these modules in #22835. The affected modules will build and their tests will pass, but their pipelines will still fail because of the upgrade check in the analyze job.


richardpark-msft (approved):
Approved for aznamespaces

PR #22844, opened 2024-05-06
Update CODEOWNERS

RickWinter:


RickWinter:
@weshaggard Made your same change against the latest sources

Issue #21934, opened 2023-11-09
[keyvault] enable weekly live managed HSM tests


gracewilcox:
Fix the azadmin and azkeys weekly pipelines to live test with managed HSM's.

PR #22848, opened 2024-05-06
[azadmin] live tests

gracewilcox:
Fixing tests to have live tests run weekly or when manually triggered.

Closes https://github.com/Azure/azure-sdk-for-go/issues/21934


sdk/security/keyvault/azadmin/ci.yml
```diff
@@ -26,16 +26,15 @@ extends:
...
-        GenerateVMJobs: true
 
-    ${{ if not(contains(variables['Build.DefinitionName'], 'tests-weekly')) }}:
-        # Due to the high cost of Managed HSMs, which keyvault-admin requires, we only want to run
-        # the live tests weekly.
-        MatrixFilters:
-          - ArmTemplateParameters=^(?!.*enableHsm.*true)
+    # Due to the high cost of Managed HSMs, which keyvault-admin requires, we only want to run
+    # the live tests weekly or for releases (manual builds).
+    ${{ if or(contains(variables['Build.DefinitionName'], 'weekly'), eq(variables['Build.Reason'], 'Manual')) }}:
```
chlowell, line 33: 🤔 I'm uncertain whether this would be better, but you could check only the build name. Then when you want to run these tests manually, you'd run `go - azadmin - weekly`

sdk/security/keyvault/azadmin/ci.yml
```diff
@@ -26,16 +26,15 @@ extends:
...
+      Location: australiaeast
+      RunLiveTests: true
+      MatrixConfigs:
+        - Name: keyvault_test_matrix_addons
+          Path: sdk/security/keyvault/azadmin/platform-matrix.json
+          Selection: sparse
+          GenerateVMJobs: true
```
chlowell, line 40:
```suggestion
          GenerateVMJobs: true

```

chlowell (approved):

sdk/security/keyvault/azadmin/ci.yml
gracewilcox: Was talking to Ben about this. Set it that way so that the live tests will run when the go - azadmin pipeline is triggered manually for releases. Don't want the release to go out without making sure the live tests work.

PR #22850, opened 2024-05-06
Document AZURE_CLIENT_SEND_CERTIFICATE_CHAIN env var used by EnvironmentCredential

scottaddie:
If present, `EnvironmentCredential` can use the `AZURE_CLIENT_SEND_CERTIFICATE_CHAIN` environment variable to configure a `ClientCertificateCredential` instance. Add supporting docs to indicate that behavior.


chlowell:
Can we put this in the 1P docs instead? It's a 1P-only feature.

RickWinter:
> Can we put this in the 1P docs instead? It's a 1P-only feature.

@scottaddie Can we move this and close this PR?

PR #22883, opened 2024-05-10
[azsystemevents] Changes to generate using typespec-go as the emitter, instead of autorest-go

richardpark-msft:
This isn't a 1:1 change since the TypeSpec and the Swagger were being worked on independently. 

I had to skip a couple of tests because the JSON sample given in the docs, the older JSON sample I copied over from .NET's tests, the swagger and now the TypeSpec don't agree.

Generally the teams appear to have taken a hands-off approach with the TypeSpec which means a lot of names have reverted back to their original names. All of those are documented in the CHANGELOG.

We're also using a beta emitter version, direct from GitHub, until we resolve which version we want to stamp as the first official NPM release.


richardpark-msft:
There's enough here to review and commit, with two events that need a bit more work. That's been filed: https://github.com/Azure/azure-sdk-for-go/issues/22903

sdk/messaging/eventgrid/azsystemevents/models.go
```diff
@@ -1,438 +1,14 @@
...
 // Licensed under the MIT License. See License.txt in the project root for license information.
-// Code generated by Microsoft (R) AutoRest Code Generator. DO NOT EDIT.
-// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+// Code generated by Microsoft (R) Go Code Generator. DO NOT EDIT.
 
 package azsystemevents
 
 import "time"
 
-// APICenterAPIDefinitionAddedEventData - Schema of the data property of an EventGridEvent for a Microsoft.ApiCenter.ApiDefinitionAdded
```
jhendrixMSFT, line 13: All of these breaking changes are expected?

richardpark-msft: TBH, I'm a bit conflicted. They're "expected" in that that's the way the models really are on the wire, but "unexpected" in that the arch board did decide on better names. I think I'm going to just apply the renames to bring them back to what was decided so we match what others are shipping.

richardpark-msft: I've reapplied the old fixes, so this is no longer a problem.

jhendrixMSFT (approved):

PR #22901, opened 2024-05-15
Support federated credentials in on-behalf-of flow

### LINKED ISSUE #22825: Allow the use of a client assertion in OnBehalfOfCredential
Confirm the MSAL APIs already exist, as they do in .NET. See https://github.com/Azure/azure-sdk-for-net/issues/43769 for more details.

chlowell:
Closes #22825 by adding an `OnBehalfOfCredential` constructor that takes a callback providing client assertions.

### DIFF
--- a/sdk/azidentity/example_test.go
+++ b/sdk/azidentity/example_test.go
@@ -64,7 +64,9 @@ func ExampleNewOnBehalfOfCredentialWithCertificate() {
 		// TODO: handle error
 	}
 
-	cred, err = azidentity.NewClientCertificateCredential(tenantID, clientID, certs, key, nil)
+	// userAssertion is the user's access token for the application. Typically it comes from a client request.
+	userAssertion := "TODO"
+	cred, err = azidentity.NewOnBehalfOfCredentialWithCertificate(tenantID, clientID, userAssertion, certs, key, nil)
 	if err != nil {
 		// TODO: handle error
 	}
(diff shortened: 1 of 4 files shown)


sdk/azidentity/example_test.go
```diff
@@ -64,7 +64,9 @@ func ExampleNewOnBehalfOfCredentialWithCertificate() {
 		// TODO: handle error
 	}
 
-	cred, err = azidentity.NewClientCertificateCredential(tenantID, clientID, certs, key, nil)
+	// userAssertion is the user's access token for the application. Typically it comes from a client request.
+	userAssertion := "TODO"
+	cred, err = azidentity.NewOnBehalfOfCredentialWithCertificate(tenantID, clientID, userAssertion, certs, key, nil)
```
jhendrixMSFT, line 69: Can the `func` argument legitimately be `nil`, or is this a placeholder for the example? If it's the former, would be good to include that in the doc comment. If the latter, maybe add a comment here explaining it needs to be populated.

chlowell: This example is intended to show certificate auth but was calling the wrong constructor for that. This change is just me fixing it, no relation to the new feature. But to your question, there's no scenario for a nil `getAssertion` func because all token requests must contain a client assertion. As I have it now, passing `nil` will provoke a panic when the credential tries to authenticate; I'll make that a constructor error instead

jhendrixMSFT: Derp, I misread that. Good idea on the `nil` check though given that the panic will manifest later, making it hard to diagnose.

jhendrixMSFT (approved):

PR #22931, opened 2024-05-21
Add @Pilchie to CODEOWNERS for Cosmos

Pilchie:


ealsur (approved):

RickWinter (approved):
The change is fine but the account needs to have public organization visibility. I sent you email on how to fix the linter problem that is being flagged.

PR #22938, opened 2024-05-22
Better handling cosmos regions

mberbero:
PR updates the preferred regions handling in the CosmosDB client options. It adds support for preferred regions in the CosmosDB client options and provides a method to convert the preferred regions to a string array. The preferred regions can now be specified using the new `ClientRegions` type, which includes a list of predefined region constants. This change improves the flexibility and usability of the CosmosDB client options.

### DIFF
--- a/sdk/data/azcosmos/cosmos_client_options.go
+++ b/sdk/data/azcosmos/cosmos_client_options.go
@@ -14,5 +14,101 @@ type ClientOptions struct {
 	// The default is false.
 	EnableContentResponseOnWrite bool
 	// PreferredRegions is a list of regions to be used when initializing the client in case the default region fails.
-	PreferredRegions []string
+	PreferredRegions ClientRegions
 }
+
+// PreferredRegionsAsStringArray returns the preferred regions as a string array.
+func (co *ClientOptions) PreferredRegionsAsStringArray() []string {
+	regions := make([]string, len(co.PreferredRegions))
+	for i, region := range co.PreferredRegions {
+		regions[i] = string(region)
+	}
+	return regions
+}
+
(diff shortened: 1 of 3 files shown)


mberbero:

sdk/data/azcosmos/cosmos_client_options.go
```diff
@@ -14,5 +14,104 @@ type ClientOptions struct {
...
+}
+
+// ClientRegions is a list of regions to be used when initializing the client in case the default region fails.
+type ClientRegions []ClientRegion
+
+// ClientRegion type defines the region to be used when initializing the client in case the default region fails.
+type ClientRegion string
+
+const (
+	ClientRegionEastUS              ClientRegion = "East US"
```
ealsur, line 36: We by design did not go this route in Go. Which is the same as other languages (java for example). The reason is, this couples the client side SDK with the existence of Azure regions, which can vary with time and would make users have to upgrade SDKs just to get new regions. It would also add overhead on the dev team having to maintain such a list.

And this is not Cosmos DB specific either, if such a list exists, it might be a better alternative to be part of the core package.

sdk/data/azcosmos/cosmos_client_options.go
```diff
@@ -14,5 +14,104 @@ type ClientOptions struct {
 	// The default is false.
 	EnableContentResponseOnWrite bool
 	// PreferredRegions is a list of regions to be used when initializing the client in case the default region fails.
-	PreferredRegions []string
+	PreferredRegions ClientRegions
```
ealsur, line 17: This is a breaking change, we already GAed the public API

PR #22953, opened 2024-05-28
Add NewUUID to runtime package

jhendrixMSFT:
For use by generated code to produce UUIDs.

Part of https://github.com/Azure/autorest.go/issues/770


RickWinter:
Why doesn't the generated code use the internal/uuid package?

jhendrixMSFT:
Because it won't have access to it outside of the repo (e.g. autorest.go).

JeffreyRichter (approved):

Issue #22961, opened 2024-05-29
[azeventgrid] Using a TokenCredential returns an InternalServerError


richardpark-msft:
Just tracking this bug - it looks like this is a service-side issue but we want to re-test this once the fix has been deployed.

richardpark-msft:
NOTE: the bug is apparently triggered if your identity is a member of many groups. So I was able to leave these tests enabled in CI and just use a switch so you can still run the tests locally.

PR #22962, opened 2024-05-29
[aznamespaces] Event Grid, now using full TypeSpec generation!

richardpark-msft:
The first generation of this client using pure TypeSpec, without swagger or autorest.

Some changes, in response to UX studies:
- Split the single client into two clients - SenderClient and ReceiverClient. This mirrors the structure we've used in other messaging libraries, and helps declutter each client.
- Remove 'CloudEvent' from all the method names, so ReceiveCloudEvents is now just Receive()
- topic and subscription are no longer method parameters, and are client level parameters instead.

TokenCredential auth is built-in, but testing is currently just using SAS keys due to #22961


sdk/messaging/eventgrid/aznamespaces/models.go
```diff
@@ -53,6 +49,10 @@ type FailedLockToken struct {
 	LockToken *string
 }
 
+// publishResult - The result of the Publish operation.
+type publishResult struct {
```
jhendrixMSFT, line 53: Is this something that's being generated?

richardpark-msft: Yes.

jhendrixMSFT: Mostly I was wondering if this was a base type that the emitter should have pruned. But that doesn't appear to be the case. It's harmless enough especially as it's not exported, but it does seem a bit weird. Might be worth following up with the event grid folks on why it's authored like this and if there's a better pattern.

richardpark-msft: So I think I can just not use it at all (it is embedded in public responses, which is a big no-no). 

This all hangs on what they said in the last arch-board meeting about publish operations _never_ returning a result. So my instinct is to remove this, and change the signature of Send and SendEvents to only return only an 'error'. What do you think?

jhendrixMSFT: I see. I think we should continue to return the response envelopes, just omit the `publishResult` type from them. This is consistent with our SDK patterns and is extensible in a non-breaking way.

richardpark-msft: Brought it all back, we'll leave it as generated (except publishResult will be exported)

sdk/messaging/eventgrid/aznamespaces/CHANGELOG.md
```diff
@@ -4,10 +4,15 @@
 
 ### Features Added
 
-- Added in "binary mode" support, which can improve efficiency when sending a single event (PR#22582)
+- Supports API version `2024-06-01`, including support for release with delay.
+- Added in binary mode support, which can improve efficiency when sending a single event (PR#22582)
```
jhendrixMSFT, line 8: Is binary mode still available or did that get removed in a previous commit?

richardpark-msft: It's gone now, didn't mean to leave the changelog entry there.

jhendrixMSFT (approved):

PR #23000, opened 2024-06-05
Change module search depth to 0

benbp:
It seems search depth 1 is going one directory deep when we want is to only search the current directory.

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -101,9 +101,8 @@ function Get-AllPackageInfoFromRepo($serviceDirectory)
   $searchPath = Join-Path $RepoRoot "sdk"
   $pkgFiles = @()
   if ($serviceDirectory) {
-    $searchPath = Join-Path $searchPath $serviceDirectory
-    # If we have a service directory only include the immediate match, no nested go.mod files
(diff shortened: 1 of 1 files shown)


chlowell (approved):
🤔 yeah I suppose zero-based depth makes sense

eng/scripts/Language-Settings.ps1
```diff
@@ -103,7 +103,7 @@ function Get-AllPackageInfoFromRepo($serviceDirectory)
   if ($serviceDirectory) {
     $searchPath = Join-Path $searchPath $serviceDirectory
     # If we have a service directory only include the immediate match, no nested go.mod files
-    [array]$pkgFiles = Get-ChildItem -Path $searchPath -Include "go.mod" -Depth 1
+    [array]$pkgFiles = Get-ChildItem -Path $searchPath -Include "go.mod" -Depth 0
```
weshaggard, line 106: Why is the depth even needed given it isn't using recursive.

benbp: I guess it's a holdover from using `-Include` which only works with `-Depth`. I should change this whole command to just `Get-ChildItem "$searchPath/go.mod"`.

weshaggard (approved):

PR #23011, opened 2024-06-07
Prep azappconfig v1.1.1-beta.1 for release

jhendrixMSFT:

### DIFF
--- a/sdk/data/azappconfig/CHANGELOG.md
+++ b/sdk/data/azappconfig/CHANGELOG.md
@@ -1,15 +1,11 @@
 # Release History
 
-## 1.1.1 (Unreleased)
+## 1.1.1-beta.1 (2024-06-11)
 
 ### Features Added
 * Support ETag-per-page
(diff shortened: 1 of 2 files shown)


sdk/data/azappconfig/CHANGELOG.md
```diff
@@ -1,16 +1,12 @@
 # Release History
 
-## 1.1.1 (Unreleased)
```
RickWinter, line 3: This should be 1.2.0-beta.1 since the release will be the 1.2.0 when it ships. 

Do we need to fix automation (seperate PR) for these versions to bump correctly?

jhendrixMSFT: Good catch, fixed.

Default bump of patch is fine IMO. Although, it might be good if we had some task in CI that would ensure we bump the minor version in post-GA packages when a new feature is added.

RickWinter: That's good idea. We could add a check that the minor must rev if the feature_added section has content. I made an issue to track this idea
https://github.com/Azure/azure-sdk-for-go/issues/23012

PR #23021, opened 2024-06-07
[aznamespaces] Prep for GA

richardpark-msft:


jhendrixMSFT (approved):

richardpark-msft:
Found a few more things that needed changing:
- The 'result' types didn't have 'Event' in them, like we do for methods
- Readme had another bit that was referencing 'publish' vs 'send'
- Removing the 'Send/PublishResult'.
- Cleaned up the tests so the temp override to force SAS auth in tests is more clear.

PR #23022, opened 2024-06-10
[STG93][AzFiles] Adding Rename Support to List Ranges API 

tanyasethi-msft:
`x-ms-file-support-rename: { Boolean }`
Supported in version 2024-05-04 and above. This header is allowed only when prevsharesnapshot query parameter is present. The Boolean value determines whether the changed ranges for a file should be listed when the file's location in the previous snapshot is different from the location in the Request URI, as a result of rename or move operations. If the value is true, the valid changed ranges for the file will be returned. If the value is false, the operation will result in a failure with 409 (Conflict) response. The default value is false.

### DIFF
--- a/sdk/storage/azfile/internal/generated/autorest.md
+++ b/sdk/storage/azfile/internal/generated/autorest.md
@@ -7,7 +7,7 @@ go: true
 clear-output-folder: false
 version: "^3.0.0"
 license-header: MICROSOFT_MIT_NO_VERSION
-input-file: "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/29ebfcffaddbbf8db7565a789ce19b5be3601748/specification/storage/data-plane/Microsoft.FileStorage/preview/2024-05-04/file.json"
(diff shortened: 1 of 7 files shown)


sdk/storage/azfile/internal/generated/autorest.md
```diff
@@ -7,7 +7,7 @@ go: true
 clear-output-folder: false
 version: "^3.0.0"
 license-header: MICROSOFT_MIT_NO_VERSION
-input-file: "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/29ebfcffaddbbf8db7565a789ce19b5be3601748/specification/storage/data-plane/Microsoft.FileStorage/preview/2024-05-04/file.json"
+input-file: "https://raw.githubusercontent.com/Azure/azure-rest-api-specs/70cfdba3d6acf2325fa5a9d264dfdf69c99fc0f9/specification/storage/data-plane/Microsoft.FileStorage/preview/2024-05-04/file.json"
```
souravgupta-msft, line 10: Is this commit from the main or some other branch?

souravgupta-msft (approved):

sdk/storage/azfile/internal/generated/autorest.md
tanyasethi-msft: This is from main

PR #23025, opened 2024-06-10
Add AzurePipelinesCredential parameters to its constructor

chlowell:
This is the planned final constructor API. I'll update markdown docs in another PR.

### DIFF
--- a/sdk/azidentity/azure_pipelines_credential.go
+++ b/sdk/azidentity/azure_pipelines_credential.go
@@ -6,7 +6,6 @@ package azidentity
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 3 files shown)


sdk/azidentity/azure_pipelines_credential.go
```diff
@@ -48,28 +46,36 @@ type AzurePipelinesCredentialOptions struct {
...
 	if u == "" {
 		return nil, fmt.Errorf("no value for environment variable %s. This should be set by Azure Pipelines", systemOIDCRequestURI)
 	}
-	sat := os.Getenv(systemAccessToken)
-	if sat == "" {
-		return nil, errors.New("no value for environment variable " + systemAccessToken)
+	if clientID == "" {
+		return nil, fmt.Errorf("no client ID specified")
+	}
+	if serviceConnectionID == "" {
```
jhendrixMSFT, line 69: Does `systemAccessToken` need similar validation?

jhendrixMSFT (approved):
I assume the changelog will be updated in the separate docs PR.

chlowell:
Yes, I'll update the changelog after publishing v1.6.0 (this PR will ship in v1.7.0-beta.1)

sdk/azidentity/azure_pipelines_credential.go
chlowell: 🤦 yep, that would be for the best

PR #23043, opened 2024-06-11
[azopenaiassistants] Regenerating to work with the new assistants v2 (beta) endpoint

richardpark-msft:
Regenerating based on the Assistants V2 update (it's beta, just v2 of their beta).

The non-autogenerated code is there to work around the way that `union` translates to swagger - it typically produces the leaf objects but the union itself is empty. model_custom.go are manual representations of the unions, themselves.
    
I've also addressed some linter errors, removing log.Fatal's from the examples.

NOTE: this is just to get some of the simpler-to-review changes out of the way. I will make another PR after this to address writing examples for some of the new features, and to introduce streaming.


sdk/ai/azopenaiassistants/models_custom.go
```diff
@@ -0,0 +1,197 @@
...
+	CodeInterpreterToolDefinition *CodeInterpreterToolDefinition
+
+	// FileSearchToolDefinition should be set if you attach files for the FileSearchToolDefinition tool.
+	FileSearchToolDefinition *FileSearchToolDefinition
+}
+
+// MarshalJSON implements the json.Marshaller interface for type MessageAttachmentToolDefinition.
+func (m *MessageAttachmentToolDefinition) MarshalJSON() ([]byte, error) {
+	switch {
+	case m.CodeInterpreterToolDefinition != nil:
```
jhendrixMSFT, line 27: This won't handle the case when sending a JSON `null` (unclear if that's important). See the `populate()` func that gets generated in the `models_serde.go` file.

sdk/ai/azopenaiassistants/models_custom.go
```diff
@@ -0,0 +1,197 @@
...
+		return json.Marshal(m.FileSearchToolDefinition)
+	default:
+		return nil, errors.New("neither CodeInterpreterToolDefinition or FileSearchToolDefinition were set")
+	}
+}
+
+// UnmarshalJSON implements the json.Unmarshaller interface for type MessageAttachmentToolDefinition.
+func (m *MessageAttachmentToolDefinition) UnmarshalJSON(data []byte) error {
+
+	var v *struct {
```
jhendrixMSFT, line 39: Is this intermediate type really necessary? It looks to be identical to `MessageAttachmentToolDefinition`.

jhendrixMSFT (approved):

sdk/ai/azopenaiassistants/models_custom.go
richardpark-msft: I deserialized this incorrectly - they're not sub-objects, it's a union. I'll fix it.

sdk/ai/azopenaiassistants/models_custom.go
richardpark-msft: Fixed with a little if/else so we still call json.Marshal(). The service does validation on the value, so there's no reason for me to.

sdk/ai/azopenaiassistants/models_custom.go
richardpark-msft: Fixed.

weshaggard (approved):

PR #23057, opened 2024-06-12
Remove plaintext fallback for persistent token caches

chlowell:
The problem with this feature is that it enables developers to choose weaker protection on behalf of users without informing them of the choice or its risks.


jhendrixMSFT (approved):
I presume the plain text variant can still be explicitly used?

chlowell:
It can't, this change removes plaintext storage entirely and there's no API for plugging in custom storage implementations. Applications will get an error when a credential can't encrypt its cache (currently it's an authentication error but I intend to make it a constructor error).

Issue #23052, opened 2024-06-12
CosmosDB `(*ContainerClient).ReadItem()` not thread safe (any longer)


calmh:
When upgrading from azcosmos 0.3.6 to 1.0.1 we see race detector warnings when concurrently doing `(*ContainerClient).ReadItem()`. We do this in several places for performance/concurrency, and so far it has worked well. Now, however it gives us this:

```
WARNING: DATA RACE
Write at 0x00c0002e59f0 by goroutine 143:
  github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos.(*clientRetryPolicy).resetPolicyCounters()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/data/azcosmos@v1.0.1/cosmos_client_retry_policy.go:178 +0x70
  github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos.(*clientRetryPolicy).Do()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/data/azcosmos@v1.0.1/cosmos_client_retry_policy.go:30 +0x68
  github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.(*Request).Next()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/azcore@v1.11.1/internal/exported/request.go:107 +0x218
  github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos.(*sharedKeyCredPolicy).Do()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/data/azcosmos@v1.0.1/shared_key_credential.go:122 +0x466
  github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.(*Request).Next()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/azcore@v1.11.1/internal/exported/request.go:107 +0x218
  github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime.(*retryPolicy).Do()
      /home/runner/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/azcore@v1.11.1/runtime/policy_retry.go:121 +0x739
  github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.(*Request).Next()
[...]

PR #23061, opened 2024-06-13
Thread-Safe clientRetryPolicy

ealsur:
`clientRetryPolicy` was storing state that was being shared across all requests, this caused race conditions and went against the Policy best practices.

This PR extracts the retry context into a property of the request itself.

Adding a concurrent test to enable race detection.

Closes https://github.com/Azure/azure-sdk-for-go/issues/23052


sdk/data/azcosmos/ci.yml
```diff
@@ -70,5 +70,6 @@ extends:
               ServiceDirectory: 'data/azcosmos'
               Image: $(vm.image)
               GoVersion: '1.22.0'
+              EnableRaceDetector: true
```
jhendrixMSFT, line 73: Doesn't look like this enabled the race detector. You can see the absence of `-race` in the `Run tests` stage. I _think_ this needs to be moved to the `parameters` block on line 25?

ealsur: It did, and actually failed the tests 

```
[Jun 13 21:53:06.734144] Retry: exit due to non-retriable status code
    testing.go:1398: race detected during execution of test
--- FAIL: TestItemConcurrent (0.63s)
```

I enabled it only on the Emulator run, because thats where the concurrent tests are running :)

ealsur: That made me discover that I had a bug on the test and I was concurrently writing on a shared variable

ealsur: Our ci.yml does 2 things, runs the normal tests through the arch template but that ignores the Emulator tests (can only run on Windows), then as a second stage, we run the other tests on Windows after installing the Emulator and here is where I added the flag because we want to catch probably the failures on real-ish scenarios.

ealsur: Reference to the build with `-race`: https://dev.azure.com/azure-sdk/public/_build/results?buildId=3872094&view=logs&j=07a8bf49-95b8-5c7c-de1a-77230f20b6cf&t=79c67422-3cb3-50fc-4ea3-00a141e27961&l=13

Pilchie (approved):

sdk/data/azcosmos/ci.yml
jhendrixMSFT: Ah ok got it, thanks for clarifying.

PR #23063, opened 2024-06-13
Allow opt-in for Go race detector per SDK

jhendrixMSFT:
This is a stop-gap until a more general solution is found. Some tests will cause an OOM when the race detector is enabled. For now, allow SDKs to opt in to the race detector if their tests don't cause an OOM.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -17,6 +17,9 @@ parameters:
   - name: TestRunTime
     type: string
     default: '600s'
+  - name: EnableRaceDetector
+    type: boolean
+    default: false
 
 steps:
   - task: Powershell@2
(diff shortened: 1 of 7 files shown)


benbp (approved):

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -70,7 +73,7 @@ steps:
     inputs:
       targetType: 'filePath'
       filePath: ./eng/scripts/run_tests.ps1
-      arguments: '${{ parameters.ServiceDirectory }} ${{ parameters.TestRunTime }}'
+      arguments: '${{ parameters.ServiceDirectory }} ${{ parameters.TestRunTime }} ${{ parameters.EnableRaceDetector }}'
```
jhendrixMSFT, line 76: @benbp is there some magic I can do here to pass a proper PS `$true`? It's passing `True` which the PS script doesn't like.

jhendrixMSFT: I tried adding a leading `$`, let's see if that works.

jhendrixMSFT: It worked.

jhendrixMSFT:
See https://github.com/Azure/azure-sdk-for-go/pull/20467 for history.

PR #23077, opened 2024-06-14
azcontainerregistry tests use internal credential factory

chlowell:


sdk/containers/azcontainerregistry/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.11.1
```
jhendrixMSFT, line 6: Since we're mucking about here, how about bumping this to `v1.12.0`?

jhendrixMSFT (approved):

richardpark-msft (approved):

PR #23093, opened 2024-06-19
[azopenaiassistants] Adding in streaming support for Assistants V2.

richardpark-msft:
Adds streaming support in, similar to how we did in azopenai, with the new feature that the events all have different types rather than the same type.

There's a very simple example, and a lot of tests to go through all the scenarios.

### DIFF
--- a/sdk/ai/azopenaiassistants/client_custom_streaming.go
+++ b/sdk/ai/azopenaiassistants/client_custom_streaming.go
@@ -8,9 +8,6 @@ package azopenaiassistants
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
-	"io"
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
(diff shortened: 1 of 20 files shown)


sdk/ai/azopenaiassistants/client_custom_streaming.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+)
+
+// API Reference: https://platform.openai.com/docs/api-reference/assistants-streaming
+
+// CreateThreadAndRunStreamResponse contains the response from [CreateThreadAndRunStream].
+type CreateThreadAndRunStreamResponse struct {
+	// Stream can be used to stream response events.
+	Stream *EventReader[StreamEvent]
```
jhendrixMSFT, line 22: Is the `StreamEvent` intermediate type really necessary? What I was envisioning was a pattern like this.
```go
event, _ := resp.Stream.Read()
switch e := event.(type) {
case *AssistantThread:
    // e is an *AssistantThread, do something with it....
}
```

richardpark-msft: I was thinking that too, but the issue is that `Kind` is not part of the different model objects like AssistantThread, so I'd have to add it. However, those same models are used in other non-streaming situations.

sdk/ai/azopenaiassistants/streaming_helpers_test.go
```diff
@@ -0,0 +1,237 @@
...
+		if errors.Is(err, io.EOF) {
+			t.Logf("(%s) Stream has ended normally", scenario)
+			break
+		}
+
+		require.NoError(t, err)
+
+		thread.EventCounts[event.Kind]++
+
+		switch event.Kind {
```
jhendrixMSFT, line 57: Can't we type switch on `event.Event`?

richardpark-msft: I do wish we could but you'd still have to switch, within each 'case *ThreadRun', for instance, to see what the actual event was (ie, was it the thread run being created, the thread being completed, in progress, etc...

So in case that you'd have to switch twice, instead of once.

jhendrixMSFT (approved):

Issue #22910, opened 2024-05-16
[query/azlogs] add logs batch


gracewilcox:
Generate logs batch method

PR #23109, opened 2024-06-25
[query/azlogs] add batch queries

gracewilcox:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/22910

Adding the batch functionality to azlogs


sdk/monitor/query/azlogs/client_test.go
```diff
@@ -205,6 +205,103 @@ func TestQueryResource_Advanced(t *testing.T) {
...
+	require.Len(t, res.Responses, 2)
+	for _, resp := range res.Responses {
+		require.Nil(t, resp.Body.Error)
+		require.NotNil(t, resp.Body.Tables)
+		if *resp.ID == "1" && len(resp.Body.Tables[0].Rows) != 100 {
+			t.Fatal("expected 100 rows from batch request 1")
+		}
+		if *resp.ID == "2" && len(resp.Body.Tables[0].Rows) != 2 {
+			t.Fatal("expected 2 rows from batch request 2")
+		}
```
chlowell, line 230: Where does this data come from? I'm wondering whether these tests depend on the state of resources they don't control.

chlowell (approved):

sdk/monitor/query/azlogs/client_test.go
gracewilcox: It all in the kusto query in line 18! The query provides the data, so it's always constant. No test resources required.

PR #23114, opened 2024-06-26
Redesign persistent token cache API

chlowell:
Removing the plaintext fallback option (#23057) changed error scenarios such that separating cache and credential construction provides a better experience for applications. This makes it obvious when persistent caching is unavailable, and easy for applications to notify users of the consequences and fall back to in-memory caching.


sdk/azidentity/internal/cache.go
```diff
@@ -0,0 +1,61 @@
...
+
+package internal
+
+import "github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
+
+// Cache represents a persistent token cache
+type Cache struct {
+	// impl is embedded as a pointer so a Cache
+	// can carry persistent state across copies
+	*impl
```
jhendrixMSFT, line 12: Can this be a field? If one were to add exported methods to this type, they'd show up in public surface area which might not be immediately obvious. I don't _think_ we'd lose the state across copies unless I'm missing some subtly here.

sdk/azidentity/internal/cache.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, nil
+	}
+	var (
+		err error
+		xr  cache.ExportReplace
+	)
+	if cae {
+		if i.cae == nil {
+			if xr, err = i.factory(cae); err == nil {
+				i.cae = xr
```
jhendrixMSFT, line 35: If I have multiple credentials sharing a cache, can `exportReplace` be called concurrently? Or will synchronization happen at a higher level? Wondering if this is going to race (also line 42).

sdk/azidentity/internal/cache.go
chlowell: Sure, that works too, good idea.

sdk/azidentity/internal/cache.go
chlowell: Good catch. A race is possible but probably not dangerous because a cache's name determines its storage and accessing that is synchronized elsewhere with a lock file. This looks bad and should trip the race detector though, I'll add synchronization.

sdk/azidentity/cache/cache.go
```diff
@@ -8,45 +8,99 @@
...
-		if o == nil {
-			return nil, nil
+var (
+	// once ensures New tests the storage implementation only once
+	once = &sync.Once{}
+	// storageError is the error from the storage test
+	storageError error
+	// tryStorage tests the storage implementation by round-tripping data
+	tryStorage = func() {
+		errFmt := "persistent storage isn't available due to error %q"
```
jhendrixMSFT, line 33: `const errFmt = ...`

sdk/azidentity/cache/cache.go
```diff
@@ -8,45 +8,99 @@
...
 
-func init() {
-	internal.NewCache = func(o *internal.TokenCachePersistenceOptions, enableCAE bool) (cache.ExportReplace, error) {
-		if o == nil {
-			return nil, nil
+var (
+	// once ensures New tests the storage implementation only once
+	once = &sync.Once{}
+	// storageError is the error from the storage test
+	storageError error
```
jhendrixMSFT, line 30: This effectively prevents retries if the first attempt fails. I suspect that this is ok as there shouldn't be any kind of transient error here (how would one even handle that?).

chlowell: That's my thinking as well. Some reasons for failure are:
1. can't write a file
2. can't unlock mac keychain
3. Linux kernel API unavailable/forbidden

So failures probably aren't transient. Even so, I guess a user could take some action outside the application that would make storage possible on retry, but that seems unlikely to actually happen and not worth the cost of `New` repeatedly testing the storage implementation in the success case (success is also probably not transient).

sdk/azidentity/example_user_auth_test.go
```diff
@@ -39,20 +37,27 @@ func storeRecord(record azidentity.AuthenticationRecord) error {
...
-//  3. Add the [AuthenticationRecord] to the credential's options
+//  4. Add the [AuthenticationRecord] to the credential's options
 func Example_persistentUserAuthentication() {
 	record, err := retrieveRecord()
 	if err != nil {
 		// TODO: handle error
 	}
+	c, err := cache.New(nil)
+	if err != nil {
+		// TODO: handle error. An error here means persistent
```
RickWinter, line 52: Why don't we simply error out. Continuing on seems bad.

sdk/azidentity/client_secret_credential.go
```diff
@@ -32,8 +32,10 @@ type ClientSecretCredentialOptions struct {
 	// the application responsible for ensuring the configured authority is valid and trustworthy.
 	DisableInstanceDiscovery bool
 
-	// TokenCachePersistenceOptions enables persistent token caching when not nil.
-	TokenCachePersistenceOptions *TokenCachePersistenceOptions
+	// Cache is a persistent cache the credential will use to store the tokens it acquires, making
```
RickWinter, line 35: Would this be better served to be on TokenCredentialOptions and the have each token credential include `azidentity.TokenCredentialOptions`.

Avoids needing the same line and comment in each credentialtype

sdk/azidentity/example_user_auth_test.go
chlowell: Whether to continue is up to developers. Continuing means the application can do what the user asked but will require the user to log in again the next time it runs. I show that in the example because I believe most developers would prefer it to the alternative, which is the app being unable to access Azure resources.

sdk/azidentity/client_secret_credential.go
chlowell: It's worth it 😄

Embedding makes literal syntax more awkward e.g.
```go
opts := azidentity.ClientSecretCredentialOptions{
    Cache: c,
}
```
becomes
```go
opts := azidentity.ClientSecretCredentialOptions{
    EmbeddedOptions: azidentity.EmbeddedOptions{
        Cache: c,
    },
}
```

Embedding also adds a layer of indirection in docs. See for example how the embedded `azcore.ClientOptions` appears in [package docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#AzurePipelinesCredentialOptions).

I'd also be afraid to add persistent caching to a consolidated options type because it doesn't fit with the other options shared by several credential types and we may in the future add more partly shared options. For example, `OnBehalfOfCredential` doesn't support persistent caching but shares `AdditionallyAllowedTenants` and `DisableInstanceDiscovery` options with several other types that do support persistent caching. So, consolidating and embedding would give us options types having the same field at different levels or options fragmented across a collection of small consolidating types, or some mix of both.

sdk/azidentity/internal/cache.go
```diff
@@ -0,0 +1,83 @@
...
+
+// NewCache is the constructor for Cache. It takes a factory instead of an instance
+// because it doesn't know whether the Cache will store both CAE and non-CAE tokens
+// (these must be stored separately).
+func NewCache(factory func(cae bool) (cache.ExportReplace, error)) Cache {
+	return Cache{&impl{factory: factory, l: &sync.RWMutex{}}}
+}
+
+// ExportReplace returns an implementation satisfying MSAL's ExportReplace interface.
+// It's an internal function instead of an exported method so packages in azidentity
```
richardpark-msft, line 78: So this function _is_ exported, but from internal. I'm curious if this is an older comment, or if exported is being used in a different context.

richardpark-msft (approved):
There's a lot here but it looks like Joel and Rick have been through it. I don't see anything particularly objectionable, and it makes sense how it works.

I added a couple of comments that it'd be nice to include more background info on some parts for visitors/observers of the code.

jhendrixMSFT (approved):

Issue #22475, opened 2024-02-28
[Key Vault] Add more samples


gracewilcox:
Increase the number of samples in the Key Vault modules, including a sample for `ImportKey` as requested in [this issue](https://github.com/Azure/azure-sdk-for-go/issues/22452) and `ImportCertificate` as requested in [this issue](https://github.com/Azure/azure-sdk-for-go/issues/21873).

An increase in samples will hopefully fill the documentation gap, especially helping new users unfamiliar with the Go SDK.

Roopan92:
@gracewilcox  - Any update here?. We need an sample for Importkey of BYOK

Roopan92:
@gracewilcox - Any update?

Roopan92:
@gracewilcox - Any examples added for Import BYOK In azure?. I don't see any still

PR #23123, opened 2024-06-28
[keyvault] add more samples

gracewilcox:
Adding requested samples 

Closes https://github.com/Azure/azure-sdk-for-go/issues/22475

### DIFF
--- a/sdk/security/keyvault/azcertificates/example_test.go
+++ b/sdk/security/keyvault/azcertificates/example_test.go
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
 	// In a soft-delete enabled vault, deleted resources can be recovered until they're purged (permanently deleted).
 	fmt.Printf("Certificate will be purged at %v", *resp.ScheduledPurgeDate)
 }
+
+func ExampleClient_ImportCertificate_pfx() {
+	// This example uses `ImportCertificate` to import a PFX certificate.
+
+	// Assuming you already have a PFX containing your key pair, you can import it into Key Vault.
+	// You can do this without setting a policy, but the policy is needed if you want the private key to be exportable
+	// or to configure actions when a certificate is close to expiration.
+	parameters := azcertificates.ImportCertificateParameters{
+		Base64EncodedCertificate: to.Ptr("<TODO: pfx cert value>"),
+	}
+
+	resp, err := client.ImportCertificate(context.TODO(), "pfxCertName", parameters, nil)
+	if err != nil {
+		// TODO: handle error
+	}
+
+	fmt.Printf("PFX certificate %s imported successfully.", resp.ID.Name())
+}
+
+func ExampleClient_ImportCertificate_pem() {
+	// This example uses `ImportCertificate` to import a PEM certificate.
+
+	// To import a PEM-formatted certificate, you must provide a CertificatePolicy that sets the ContentType to
+	// CertificateContentType.pem or the certificate will fail to import (the default content type is PFX).
+	parameters := azcertificates.ImportCertificateParameters{
+		Base64EncodedCertificate: to.Ptr("<TODO: pem cert value>"),
+		CertificatePolicy: &azcertificates.CertificatePolicy{
+
+			SecretProperties: &azcertificates.SecretProperties{
+
+				ContentType: to.Ptr("application/x-pem-file"),
+			},
+		},
+	}
+
+	resp, err := client.ImportCertificate(context.TODO(), "pemCertName", parameters, nil)
+	if err != nil {
+		// TODO: handle error
+	}
+
+	fmt.Printf("PFX certificate %s imported successfully.", resp.ID.Name())
+
+}
--- a/sdk/security/keyvault/azcertificates/example_test.go
+++ b/sdk/security/keyvault/azcertificates/example_test.go
@@ -106,9 +106,6 @@ func ExampleClient_ImportCertificate_pfx() {
 	}
 	encodedData := base64.StdEncoding.EncodeToString(data)
 
-	// Assuming you already have a PFX containing your key pair, you can import it into Key Vault.
-	// You can do this without setting a policy, but the policy is needed if you want the private key to be exportable
-	// or to configure actions when a certificate is close to expiration.
 	parameters := azcertificates.ImportCertificateParameters{
 		Base64EncodedCertificate: to.Ptr(encodedData),
 		CertificatePolicy: &azcertificates.CertificatePolicy{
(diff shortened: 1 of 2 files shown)


sdk/security/keyvault/azkeys/example_test.go
```diff
@@ -153,3 +154,40 @@ func ExampleClient_NewListKeyPropertiesPager() {
...
+		s = fmt.Sprintf("0%s", s)
+	}
+	ret, err := hex.DecodeString(s)
+	if err != nil {
+		// TODO: handle error
+	}
+	return ret
+}
+
+func ExampleClient_ImportKey() {
```
chlowell, line 169: I'm skeptical of this example. The hard parts of the task are
1. converting from whatever key encoding I have to bytes for Key Vault
1. setting the right `JSONWebKey` fields

Part 2 is already documented well on `JSONWebKey` but part 1 may be tricky. This example doesn't really address it because customers will typically get the key from external storage such as a PEM file (looks like [this](https://github.com/Azure/azure-sdk-for-go/blob/54cdd5a3aa7138994b89ddc2c8ee6beb8b890dc6/sdk/azidentity/testdata/certificate.pem#L1-L28)). They won't have hex strings, they'll have base64 or plain old bytes they must parse to extract the key components.

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
...
+
+	// To import a PEM-formatted certificate, you must provide a CertificatePolicy that sets the ContentType to
+	// CertificateContentType.pem or the certificate will fail to import (the default content type is PFX).
+	parameters := azcertificates.ImportCertificateParameters{
+		Base64EncodedCertificate: to.Ptr("<TODO: pem cert value>"),
+		CertificatePolicy: &azcertificates.CertificatePolicy{
+
+			SecretProperties: &azcertificates.SecretProperties{
+
+				ContentType: to.Ptr("application/x-pem-file"),
```
chlowell, line 127:
```suggestion
			SecretProperties: &azcertificates.SecretProperties{
				ContentType: to.Ptr("application/x-pem-file"),
```

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
...
 }
+
+func ExampleClient_ImportCertificate_pfx() {
+	// This example uses `ImportCertificate` to import a PFX certificate.
+
+	// Assuming you already have a PFX containing your key pair, you can import it into Key Vault.
+	// You can do this without setting a policy, but the policy is needed if you want the private key to be exportable
+	// or to configure actions when a certificate is close to expiration.
+	parameters := azcertificates.ImportCertificateParameters{
+		Base64EncodedCertificate: to.Ptr("<TODO: pfx cert value>"),
```
chlowell, line 105: Hmm I'd probably show reading bytes from a .pfx and encoding them. And I suppose the default `ContentType` is x-pkcs12 or similar, so it isn't strictly necessary to set it in this case? I'd set it anyway so readers know about it.

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
 	// In a soft-delete enabled vault, deleted resources can be recovered until they're purged (permanently deleted).
 	fmt.Printf("Certificate will be purged at %v", *resp.ScheduledPurgeDate)
 }
+
+func ExampleClient_ImportCertificate_pfx() {
+	// This example uses `ImportCertificate` to import a PFX certificate.
```
chlowell, line 99: FYI, if you put intro text in a comment on the `func`, it renders above the example code [like this](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity@v1.7.0-beta.1#example-AuthenticationRequiredError)
```suggestion
// This example uses `ImportCertificate` to import a PFX certificate.
func ExampleClient_ImportCertificate_pfx() {
```

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
...
+		},
+	}
+
+	resp, err := client.ImportCertificate(context.TODO(), "pemCertName", parameters, nil)
+	if err != nil {
+		// TODO: handle error
+	}
+
+	fmt.Printf("PFX certificate %s imported successfully.", resp.ID.Name())
+
```
chlowell, line 138:
```suggestion
```

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +94,46 @@ func ExampleClient_DeleteCertificate() {
...
+	fmt.Printf("PFX certificate %s imported successfully.", resp.ID.Name())
+}
+
+func ExampleClient_ImportCertificate_pem() {
+	// This example uses `ImportCertificate` to import a PEM certificate.
+
+	// To import a PEM-formatted certificate, you must provide a CertificatePolicy that sets the ContentType to
+	// CertificateContentType.pem or the certificate will fail to import (the default content type is PFX).
+	parameters := azcertificates.ImportCertificateParameters{
+		Base64EncodedCertificate: to.Ptr("<TODO: pem cert value>"),
```
chlowell, line 122: Can I just read a PEM file (containing only one cert) directly into this field?

sdk/security/keyvault/azkeys/example_test.go
gracewilcox: Fair point, won't add the sample

sdk/security/keyvault/azcertificates/example_test.go
gracewilcox: yes, added to sample in [ff9290f](https://github.com/Azure/azure-sdk-for-go/pull/23123/commits/ff9290f757c8c0b1189bef2823d87c3bc2b33275)

sdk/security/keyvault/azcertificates/example_test.go
gracewilcox: fixed in [ff9290f](https://github.com/Azure/azure-sdk-for-go/pull/23123/commits/ff9290f757c8c0b1189bef2823d87c3bc2b33275)

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +96,59 @@ func ExampleClient_DeleteCertificate() {
...
+	// Getting certificate data from locally stored PFX file then encoding to base64
+	data, err := os.ReadFile("PFX_CERT_PATH")
+	if err != nil {
+		// TODO: handle error
+	}
+	encodedData := base64.StdEncoding.EncodeToString(data)
+
+	// Assuming you already have a PFX containing your key pair, you can import it into Key Vault.
+	// You can do this without setting a policy, but the policy is needed if you want the private key to be exportable
+	// or to configure actions when a certificate is close to expiration.
```
chlowell, line 111: 🤔 the comment seems to raise more questions because the example doesn't show doing those things. I'd remove the comment instead of showing the advanced API, to keep the example to the point

sdk/security/keyvault/azcertificates/example_test.go
```diff
@@ -94,3 +96,59 @@ func ExampleClient_DeleteCertificate() {
...
+// This example uses `ImportCertificate` to import an existing PEM certificate.
+func ExampleClient_ImportCertificate_pem() {
+	// Getting certificate data from locally stored PEM file. Contents of .pem file are already base64 encoded
+	data, err := os.ReadFile("PEM_CERT_PATH")
+	if err != nil {
+		// TODO: handle error
+	}
+
+	// To import a PEM-formatted certificate, you must provide a CertificatePolicy that sets the ContentType to
+	// CertificateContentType.pem or the certificate will fail to import (the default content type is PFX).
```
chlowell, line 138: Do we have a `CertificateContentType` enum? If we do, this example should show it. In any case I'd delete this comment because I think the example clearly implies Key Vault requires a specific value for `ContentType`

chlowell (approved):

sdk/security/keyvault/azcertificates/example_test.go
gracewilcox: No CertificateContentType, you have to know the string

PR #23150, opened 2024-07-03
[aznamespaces] Test cleanup

### LINKED ISSUE #22961: [azeventgrid] Using a TokenCredential returns an InternalServerError

richardpark-msft:
Fixing a few things:
- I had some workaround code for TokenCredential that's no longer necessary now that the service fix has been deployed.
- Using Charles's new test credential
- Fixing recordings to work again.

Fixes #22961

### DIFF
--- a/sdk/messaging/eventgrid/aznamespaces/shared_test.go
+++ b/sdk/messaging/eventgrid/aznamespaces/shared_test.go
@@ -106,7 +106,7 @@ func newClientOptions(t *testing.T) azcore.ClientOptions {
 		}
 	} else {
 		options = azcore.ClientOptions{
-			Transport: newRecordingTransporter(t, tv),
+			Transport: newRecordingTransporter(t),
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


richardpark-msft:
(putting back into draft)

I've got some test recordings debt to clean up real quick...

sdk/messaging/eventgrid/aznamespaces/shared_test.go
```diff
@@ -181,72 +179,75 @@ func newReceiverClient(t *testing.T, useSASKey bool) *aznamespaces.ReceiverClien
 }
 
 func newRecordingTransporter(t *testing.T, testVars testVars) policy.Transporter {
-	transport, err := recording.NewRecordingHTTPClient(t, nil)
-	require.NoError(t, err)
+	recordingOpts := &recording.RecordingOptions{
+		ProxyPort:    os.Getpid()%10000 + 20000,
```
chlowell, line 183: This makes me nervous because it depends on the implementation of `recording.StartTestProxy`. The only non-default value in your `RecordingOptions` is `TestInstance`, and you only need to set that if you need to apply sanitizers to a single `testing.T`. It appears you don't need to do that i.e., you apply the same sanitizers to all recordings anyway, so I think you could just use the default `RecordingOptions`.

sdk/messaging/eventgrid/aznamespaces/shared_test.go
```diff
@@ -181,72 +179,75 @@ func newReceiverClient(t *testing.T, useSASKey bool) *aznamespaces.ReceiverClien
...
+	transport, err := recording.NewRecordingHTTPClient(t, recordingOpts)
 	require.NoError(t, err)
 
-	// err = recording.ResetProxy(nil)
-	// require.NoError(t, err)
+	err = recording.Start(t, recordingDirectory, recordingOpts)
+	require.NoError(t, err)
 
-	err = recording.AddURISanitizer(fakeTestVars.Endpoint, testVars.Endpoint, nil)
+	err = recording.AddURISanitizer(fakeTestVars.Endpoint, "https://[^/]+?/", recordingOpts)
```
chlowell, line 194: The proxy has a built-in URL sanitizer that replaces the first element of the host with `recording.SanitizedValue`. So if you make `fakeTestVars.Endpoint` something like `fmt.Sprintf("https://%s.eventgrid.azure.net", recording.SanitizedValue)` you wouldn't need this sanitizer.

sdk/messaging/eventgrid/aznamespaces/shared_test.go
```diff
@@ -256,6 +257,13 @@ func newRecordingTransporter(t *testing.T, testVars testVars) policy.Transporter
 func requireEqualCloudEvent(t *testing.T, expected messaging.CloudEvent, actual messaging.CloudEvent) {
 	t.Helper()
 
+	// the built-in sanitizers are now clearing out my CloudEvent's source attribute so we
+	// just have to assume it's 'Sanitized'.
+
+	if recording.GetRecordMode() == recording.PlaybackMode {
+		expected.Source = "Sanitized"
```
chlowell, line 264:
```suggestion
		expected.Source = recording.SanitizedValue
```

chlowell (approved):

sdk/messaging/eventgrid/aznamespaces/shared_test.go
richardpark-msft: I'm definitely okay with changing it. I was having issues with my sanitizers even firing off and this "fixed" it.

I wonder if, instead of setting up sanitizers over and over I can just do it once for the entire run instead - is that typically what people do? Or is it supposed to be safe to apply them over and over again?

chlowell: Adding the same sanitizers to every test individually isn't incorrect, it's just inefficient. If the sanitizers don't break a test, you can add them for all tests (zero `TestInstance`) immediately after starting the proxy. Every other test suite does this today because we don't have a sanitizer that's required for one test but that breaks another.

sdk/messaging/eventgrid/aznamespaces/shared_test.go
richardpark-msft: EventGrid has one extra bit with a region identifier. I could expand the default regex if that's something you think it should handle:

ex: <eventgrid-name>.eastus-1.eventgrid.azure.net

sdk/messaging/eventgrid/aznamespaces/shared_test.go
richardpark-msft: I think the primary problem was because I was adding in sanitizers after I called recording.Start(). I fixed that and now the options aren't needed anymore.

Also, I changed it so the sanitizers only initialize once. I'm not yet intentionally trying to create per-test sanitizers.

sdk/messaging/eventgrid/aznamespaces/shared_test.go
richardpark-msft: (leaving this in for now. We can talk about a more generic version at another point)

PR #23151, opened 2024-07-03
Upgrade eng/tools dependencies

chlowell:


chlowell:
@benbp these are internal standalone tools. We don't publish them and they have no tests, so the triggered pipelines are all irrelevant. Do we need to do something like add a basic `ci.yml` to prevent running arbitrary pipelines on `/eng/tools`?

benbp:
@chlowell the reason is because core/etc. specifically include the `eng` directory: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/ci.yml#L12

We could add an exclude to the ^^ config, though I think the directory was included because some parts of CI depend on things in `eng/tools`, for example: https://github.com/Azure/azure-sdk-for-go/blob/main/eng/scripts/run_tests.ps1#L50-L53

We could make the path includes a bit more fine grained as one option, e.g.

```
  paths:
    exclude:
    - eng/tools
    include:
    - sdk/azcore/
    - eng/tools/internal/coverage
```

tadelesh:
@chlowell we are doing the clean up, we could do the dep upgrade at the same time. if you are ok, close this pr and replace with this [one](https://github.com/Azure/azure-sdk-for-go/pull/23156).

chlowell:
Closing in favor of #23156

PR #23163, opened 2024-07-08
Add support for proxied IMDS services

noxoraAtWork:
Both AWS and GCP offer methods of altering their default IMDS endpoint in their clients without changing the type of client that is being created. This PR adds support to do the same for this client. 

My use case is that the default azure IMDS endpoint is not at the typical location, and as such, I'd like to be able to change the address. The current logic does not allow for this as setting either MSI_ENDPOINT or IDENTITY_ENDPOINT both cause changes to the type of client created. I have added a new environment variable rather than altering an existing one to prevent changes to existing applications

### DIFF
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Features Added
 * Restored `AzurePipelinesCredential` and persistent token caching API
+* Added support for a proxy IMDS client via the `IMDS_PROXY` environment variable
 
 ## Breaking Changes
 > These changes affect only code written against a beta version such as v1.6.0-beta.4
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -36,6 +36,7 @@ const (
 	identityServerThumbprint = "IDENTITY_SERVER_THUMBPRINT"
 	headerMetadata           = "Metadata"
 	imdsEndpoint             = "http://169.254.169.254/metadata/identity/oauth2/token"
+	imdsProxyEndpoint        = "IMDS_PROXY"
 	miResID                  = "mi_res_id"
 	msiEndpoint              = "MSI_ENDPOINT"
 	msiResID                 = "msi_res_id"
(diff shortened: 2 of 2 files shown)


noxoraAtWork:
microsoft-github-policy-service agree [company="{datadog}"]

noxoraAtWork:
microsoft-github-policy-service agree company="datadog"

noxoraAtWork:
@microsoft-github-policy-service agree company="datadog"

noxoraAtWork:
Thank you for that @chlowell ! That's probably going to be the best option for me. Do you know how this would work in the Azure SDKs for other languages?

chlowell:
The policy approach should work with other SDKs because they implement the same request pipeline concept, but I don't know their APIs for inserting a custom policy. Other Identity SDKs have implemented essentially this PR though, using the perhaps unfortunately named environment variable `AZURE_POD_IDENTITY_AUTHORITY_HOST`. So if you need the same behavior across SDK languages, you could set that variable in all cases and use a policy to implement it in Go. Something like this, updating my example above to consume a URL instead of a host because other SDKs expect `AZURE_POD_IDENTITY_AUTHORITY_HOST` to include a scheme:
```go
import (
	"net/http"
	"net/url"
	"os"

)

type hostReplacingPolicy struct {
	target      string
	replacement *url.URL
}

func (c *hostReplacingPolicy) Do(req *policy.Request) (*http.Response, error) {
	rawReq := req.Raw()
	if rawReq.Host == c.target {
		rawReq.URL.Scheme = c.replacement.Scheme
		rawReq.URL.Host = c.replacement.Host
	}
	return req.Next()
}

u, err := url.Parse(os.Getenv("AZURE_POD_IDENTITY_AUTHORITY_HOST"))
if err != nil {
	// TODO
}
cred, _ := azidentity.NewManagedIdentityCredential(&azidentity.ManagedIdentityCredentialOptions{
		PerRetryPolicies: []policy.Policy{
			&hostReplacingPolicy{
				target:      "169.254.169.254",
				replacement: u,
			},
		},
	}},
)
```

chlowell:
Thanks again for contributing this. I'm closing it because I take it the policy approach is working for you and if we do implement this environment variable approach, we (Azure SDK team) need to spend some design time on naming the variable. Please open an issue if you find the policy approach isn't working in some scenario and you need an environment variable instead.

PR #23168, opened 2024-07-09
[azservicebus] Adding in an example of how you could write your own lock renewer

### LINKED ISSUE #20857: [azservicebus] Show an example of how you could do repeated lock renewals
Showing how this can be done would be useful for customers and it'd also give us a chance to show a real-world usage of the azservicebus.Error type with it's Code field.

> Of course a built-in solution would be preferrable, and in the meantime a documented example that covers some of the possible pitfalls would be nice so
> Originally posted by @googol in https://github.com/Azure/azure-sdk-for-go/issues/15745#issuecomment-1549105520_

richardpark-msft:
Just a simple example for people that want to write their own renewer.

Also, for convenience, adding in a .ps1 file to generate the .env file if you already have a resource set up.

Fixes #20857

### DIFF
--- a/sdk/messaging/azservicebus/env.ps1
+++ b/sdk/messaging/azservicebus/env.ps1
@@ -1,0 +1,47 @@
+#!/usr/bin/env pwsh
+  
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# This script will bootstrap your .env file, given a resource group that's been deployed to.
+# We expect a premium and a standard Service Bus namespace for our tests.
+#
+# You might need to install the Azure Powershell module first and login:
+#
+# Install-Module -Name Az -Repository PSGallery -Force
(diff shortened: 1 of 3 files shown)


sdk/messaging/azservicebus/env.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+# TODO:
+$rg = "<your resource group should go here>"
+
+$contents = ""
+
+Get-AzServiceBusNamespace -ResourceGroup $rg
+| ForEach-Object {
+    $cs = (Get-AzServiceBusKey -ResourceGroup $rg -NamespaceName $_.Name -AuthorizationRuleName RootManageSharedAccessKey).PrimaryConnectionString
+    #$name = $_.Name
```
chlowell, line 23:
```suggestion
```

chlowell (approved):

PR #23186, opened 2024-07-13
[azcore] Adding in a function create a policy.Request using an existing *http.Request

richardpark-msft:
Adds a new function (NewRequestFromRequest) which lets you pass in an already created *http.Request and get back a policy.Request. 

This is used in situations where we want to run a pipeline outside of an Azure generated client.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-- Added policy.NewRequestFromRequest(), allowing for a policy.Request to be created from an existing *http.Request.
+- Added runtime.NewRequestFromRequest(), allowing for a policy.Request to be created from an existing *http.Request.
 
 ### Breaking Changes
 
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added policy.NewRequestFromRequest(), allowing for a policy.Request to be created from an existing *http.Request.
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/azcore/internal/exported/request.go
+++ b/sdk/azcore/internal/exported/request.go
@@ -74,23 +74,29 @@ func NewRequestFromRequest(req *http.Request) (*Request, error) {
 	policyReq := &Request{req: req}
 
 	if req.Body != nil {
-		// since this is an already populated http.Request we want to copy
-		// over it's body, if it has one.
-		bodyBytes, err := io.ReadAll(req.Body)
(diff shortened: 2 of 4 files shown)


sdk/azcore/internal/exported/request.go
```diff
@@ -67,6 +68,36 @@ func (ov opValues) get(value any) bool {
...
 
+// NewRequestFromRequest creates a new policy.Request with an existing *http.Request
+// Exported as runtime.NewRequestFromRequest().
+func NewRequestFromRequest(req *http.Request) (*Request, error) {
+	policyReq := &Request{req: req}
+
+	if req.Body != nil {
+		// since this is an already populated http.Request we want to copy
+		// over it's body, if it has one.
+		bodyBytes, err := io.ReadAll(req.Body)
```
jhendrixMSFT, line 79: If the existing body implements `io.ReadSeekCloser` then we shouldn't need to read/copy the body. We should check for that first, before making the copy.

richardpark-msft: Fixed, test added.

sdk/azcore/internal/exported/request.go
```diff
@@ -67,6 +68,42 @@ func (ov opValues) get(value any) bool {
...
 
+// NewRequestFromRequest creates a new policy.Request with an existing *http.Request
+// Exported as runtime.NewRequestFromRequest().
+func NewRequestFromRequest(req *http.Request) (*Request, error) {
+	policyReq := &Request{req: req}
+
+	if req.Body != nil {
+		// we can avoid a body copy here if the underlying stream is already a
+		// ReadSeekCloser.
+		readSeekCloser, isReadSeekCloser := req.Body.(io.ReadSeekCloser)
```
jhendrixMSFT, line 79: Unfortunately this doesn't have the effect I was hoping for. `io.NopCloser` erases the `io.Seeker` which is likely the common case for requests created outside our SDKs.

jhendrixMSFT: Doesn't hurt to leave it but we won't avoid copying as much as I'd hoped.

richardpark-msft: That's okay, it's still instructive and anything we can do to shave off obvious stuff is still nice.

sdk/azcore/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- Added policy.NewRequestFromRequest(), allowing for a policy.Request to be created from an existing *http.Request.
```
jhendrixMSFT, line 7: `runtime.NewRequestFromRequest`

richardpark-msft: Fixed.

chlowell (approved):

PR #23191, opened 2024-07-16
Update CODEOWNERS_baseline_errors.txt

RickWinter:
Exclude an error for a user who is OOF.


.github/CODEOWNERS_baseline_errors.txt
```diff
@@ -1,5 +1,3 @@
-chlowell is not a public member of Azure.
-lirenhe is not a public member of Azure.
 'Open Telemetry' is not a valid label for this repository.
```
RickWinter, line 1:
```suggestion
lirenhe is not a public member of Azure.
'Open Telemetry' is not a valid label for this repository.
```

jhendrixMSFT (approved):

Issue #22314, opened 2024-01-31
Move date-time parsing code to azcore


jhendrixMSFT:
At present, the date-time parsing code is purely on the codegen side, making it decentralized. So, any fix requires regenerating the world. We should move this to a `/runtime/datetime` package in `azcore`.

gracewilcox:
Fixed by [[azcore] move time parsing to core #25714](https://github.com/Azure/azure-sdk-for-go/pull/25714) and [move time helpers to azcore autorest.go#1787](https://github.com/Azure/autorest.go/pull/1787)

Issue #21584, opened 2023-09-19
Time should be correctly serialized to ISO 8601


giuliohome:
### Bug Report

- import path of package in question
For example ".../sdk/resourcemanager/storage/armstorage"
(e.g. compared to "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas")

- SDK 
Latest version
- output of `go version`
Latest version

- What happened?
The problem happens with package ".../sdk/resourcemanager/storage/armstorage"
When I use a time value for a field that accept ISO 8601 the api returns an error and I need to round the time value when running on Ubuntu or to convert to UTC on Windows.
- What did you expect or want to happen?

I would expect that, if the type-safe static checks pass, there should be no dynamic errors raised for something as format-specific as this. 
In other world I would expect that the time should be correctly serialized, e.g. with 7 decimal places, if that is required by the API specs.
In case of this code - using package "github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas" -

```go
blob_client, err := azblob.NewClientWithSharedKeyCredential(fmt.Sprintf("https://%s.blob.core.windows.net/", storageAccountName), scred, nil)
sas_url, err := blob_client.ServiceClient().GetSASURL(
	sas.AccountResourceTypes{ Container: true },
	sas.AccountPermissions{
		Create: true, Delete: true, List: true, Add: true,
	},
	time.Now().Add(24 * time.Hour),
	nil,	
)
```
I don't have to round the time, which is what I would expect also in case of my repro below

- How can we reproduce it?
For example:

```go
client := storageClientFactory.NewAccountsClient()
sasToken, err := client.ListAccountSAS(ctx, resourceGroupName, storageAccountName, armstorage.AccountSasParameters{
	KeyToSign:              to.Ptr("key1"),
	SharedAccessExpiryTime: to.Ptr(time.Now().Add(24 * time.Hour)),
	Permissions:            to.Ptr(armstorage.PermissionsR),
	Protocols:              to.Ptr(armstorage.HTTPProtocolHTTPSHTTP),
	ResourceTypes:          to.Ptr(armstorage.SignedResourceTypesS),
	Services:               to.Ptr(armstorage.ServicesB),
	SharedAccessStartTime:  to.Ptr(time.Now()),
}, nil)
```
In the above case of `ListAccountSAS`, I need to add something like `.UTC()` or` .Round(time.Second`) to make it work, while that is not needed for `GetSASURL`.

[...]

PR #23193, opened 2024-07-16
Add datetime package for date/time format helpers

jhendrixMSFT:
This replaces the date/time helpers that are generated per SDK. The Parse func is typically used for handling date/time values returned in respone headers. The date/time types are used in SerDe methods.

Part of the fix for https://github.com/Azure/azure-sdk-for-go/issues/21584
Fixes https://github.com/Azure/azure-sdk-for-go/issues/22314


jhendrixMSFT:
This replaces the time helpers that we generate, e.g. https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/time_rfc3339.go

The various `time_*.go` files will go away.

sdk/azcore/runtime/datetime/date_time.go
```diff
@@ -0,0 +1,289 @@
...
+	FormatUnix     Format = 5
+)
+
+// Parse parses a formatted date/time string and returns a [time.Time] or an error.
+//   - format is the expected format of the string value
+//   - value is the string value to parse
+func Parse(format Format, value string) (time.Time, error) {
+	switch format {
+	case FormatDateOnly:
+		return parse(time.DateOnly, value)
```
jhendrixMSFT, line 38: Apparently `time.DateOnly` was added in Go v1.20.0.

sdk/azcore/runtime/datetime/date_time.go
```diff
@@ -0,0 +1,289 @@
...
+func unmarshal[T TypeConstraint](format Format, data []byte, dest *T) error {
+	p, err := Parse(format, string(data))
+	if err != nil {
+		return err
+	}
+	*dest = T(p)
+	return nil
+}
+
+const timeOnly = "15:04:05.999999999"
```
jhendrixMSFT, line 255: Hmm, `time.TimeOnly` (Go 1.20) omits the sub-second precision. Maybe we don't need it here either?

sdk/azcore/runtime/datetime/date_time.go
jhendrixMSFT, line 1: Thinking about this more, maybe this isn't the best long-term solution. We really only need these types due to our reliance on `json.Marshal`. If we remove that, then this package would essentially be vestigial.

jhendrixMSFT: Well, XML is a bit more complicated and would still benefit from it (without completely redoing codegen for XML SerDe). Also, consolidating the RFC3339 parsing has value.

RickWinter: Can we hide them more in the internal package?

jhendrixMSFT: No, the types need to be publicly accessible for use outside the SDK repo.

jhendrixMSFT: Hmm, fakes require the target type for unmarshaling bodies. Here's an example.

https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/test/autorest/datetimerfc1123group/fake/zz_datetimerfc1123_server.go#L250

Not having these types would also require retooling some of the fakes implementation.

sdk/azcore/runtime/datetime/date_time.go
jhendrixMSFT: Fixed.

Issue #23142, opened 2024-07-02
DownloadFile without Range.Count incurs 2 calls to GetProperties


petenewcomb:
If `DownloadFileOptions.Range.Count` is zero (`CountToEnd`), then `DownloadFile()` calls `GetProperties()` to discover the size it needs to make the file before calling `downloadBuffer()`.  `downloadBuffer()` then also needs to know the size of the object to calculate the needed parallelism (among other things), so checks `Range.Count` and if zero calls `GetProperties()` again itself.  An examination of `GetProperties()` shows that these calls each go straight through to the generated protocol stub.  There appears to be no caching that would prevent the second request.

The lack of caching in `GetProperties()` is a good thing, as this would also interfere with polling for changes.  The application can implement its own caching if needed.  An internal caching mechanism could be considered that would be invalidated on each call to relevant exported functions (including `GetProperties()`), but even if desirable I believe it should be pursued as its own feature enhancement independent from this specific double-calling issue.

Today, the double call to `GetProperties()` can be avoided by the application by making sure that `Range.Count` is always populated before calling `DownloadFile()`.  The application can call `GetProperties()` itself if need be, and this is an acceptable workaround for the issue.  The problem is that unless a developer is curious enough to read the implementation (as I did) or monitor the actual calls made, these extra calls will impact applications' latency and cost.

To fix the issue, `DownloadFile()`, having called `GetProperties()` itself, should set `Range.Count` in the options structure it passes to `downloadBuffer()`.  Inserting the statement `do.Range.Count = size` just after `size` is calculated at the end of the conditional block that calls `GetProperties()` should do the trick.

Creating a test case that reproduces the issue and then verifies the fix is a more involved effort, as it's not obvious to me that the existing tests are capable of tracking and verifying the actual requests made.  Perhaps the recording mechanism could be leveraged, but I have yet to dive deep there.  The existing download performance test does not exercise `DownloadFile()` but only `DownloadStream()`.  A version that uses `DownloadFile()` with a zero `Range.Count` might also be a verification option.

Call to `GetProperties()` in `DownloadFile()`: https://github.com/Azure/azure-sdk-for-go/blob/a13b929bcec52de153155c9a543b7f38609b0ce1/sdk/storage/azblob/blob/client.go#L446

[...]

PR #23200, opened 2024-07-17
DownloadFile without Range.Count incurs 2 calls to GetProperties 

tanyasethi-msft:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/23142

There are no tests in this PR since the existing tests for _DownloadFile_ are very much capable of testing if the functionality is still working correctly or not. 
The tests check if the downloaded file is correct or not by verifying the checksums. 
This PR reduces the number of calls to _GetProperties_ by populating the blob/file size in the options just after it is fetched from the properties call, so that they needn't be fetched again. 

Also, changes are done in AzBlob and AzFile, and not in AzDatalake, since datalake uses the _DownloadFile_ implemented in AzBlob. 

### DIFF
--- a/sdk/storage/azfile/file/client.go
+++ b/sdk/storage/azfile/file/client.go
@@ -569,6 +569,7 @@ func (f *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFil
 			return 0, err
 		}
 		size = *props.ContentLength - do.Range.Offset
+		do.Range.Count = size
 	} else {
 		size = count
 	}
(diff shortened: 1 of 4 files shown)


sdk/storage/azfile/file/client.go
```diff
@@ -569,6 +569,7 @@ func (f *Client) DownloadFile(ctx context.Context, file *os.File, o *DownloadFil
 			return 0, err
 		}
 		size = *props.ContentLength - do.Range.Offset
+		do.Range.Count = size
```
souravgupta-msft, line 572: Do we also need to make this change in azdatalake?

souravgupta-msft (approved):

sdk/storage/azfile/file/client.go
tanyasethi-msft: azdatalake uses the DownloadFile implemented in AzBlob. So, change is not required there

PR #23206, opened 2024-07-18
Enable federated auth in Key Vault pipelines

chlowell:


chlowell:
Works for all live tests except azkeys (#23208)

gracewilcox (approved):

PR #23231, opened 2024-07-22
Upgrade azservicebus indirect yaml.v2 requirement

chlowell:
This is a quick fix to stop Component Governance flagging azservicebus. It isn't permanent because I believe the `replace` will block the release pipeline. `yaml.v2` arrives like this: 

```
github.com/microsoft/ApplicationInsights-Go@v0.4.4
└─── github.com/onsi/gomega@v1.5.0
     └─── gopkg.in/yaml.v2@v2.2.1
```
Interesting that it appears in `go.sum` but not `go.mod`.

### DIFF
--- a/sdk/messaging/azservicebus/go.mod
+++ b/sdk/messaging/azservicebus/go.mod
@@ -42,4 +42,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
+// yaml.v2 v2.2.1 is an indirect dependency of ApplicationInsights-Go and affected by a few CVEs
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
The `github.com/microsoft/ApplicationInsights-Go` module is no longer supported. @richardpark-msft can we remove this dependency?

richardpark-msft:
> The `github.com/microsoft/ApplicationInsights-Go` module is no longer supported. @richardpark-msft can we remove this dependency?

It's the way we report metrics and statistics for stress, so no, at least not til I find a viable replacement.

richardpark-msft (approved):

Issue #23232, opened 2024-07-22
feat(logs): Add request URL to `log.EventRetryPolicy` logs


erezrokah:
### Feature Request

The logs in `policy_retry.go` https://github.com/Azure/azure-sdk-for-go/blob/9844aeedce4fd0d3b4882e876fb146f5d3e6ac4f/sdk/azcore/runtime/policy_retry.go#L105 don't include any details on the request that's being retried, making it hard to identify which requests are being throttled.

It would be helpful to include at least the request URL in the logs

erezrokah:
Thanks for the quick fix @jhendrixMSFT 🚀

PR #23239, opened 2024-07-23
Include HTTP verb and URL in retry logs

jhendrixMSFT:
For each try, include the HTTP verb and URL in the log message so it's clear which operation is being retried.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/23232

### DIFF
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -102,7 +102,8 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 	try := int32(1)
 	for {
 		resp = nil // reset
-		log.Writef(log.EventRetryPolicy, "=====> Try=%d", try)
+		// unfortunately we don't have access to the custom allow-list of query params, so we'll redact everything but the default allowed QPs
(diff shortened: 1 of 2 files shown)


sdk/azcore/runtime/policy_retry.go
```diff
@@ -102,7 +102,8 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 	try := int32(1)
 	for {
 		resp = nil // reset
-		log.Writef(log.EventRetryPolicy, "=====> Try=%d", try)
+		// unfortunately we don't have access to the custom allow-list of query params, so we'll redact everything but the default allowed QPs
+		log.Writef(log.EventRetryPolicy, "=====> Try=%d for %s %s", try, req.Raw().Method, getSanitizedURL(*req.Raw().URL, getAllowedQueryParams(nil)))
```
richardpark-msft, line 106: Should we be passing the user's provided "allowed" query params? I note you passed 'nil' - do we not have them at this point?

jhendrixMSFT: Correct, we don't have them at this point, and it would require a new retry policy constructor to obtain them. I'd prefer to avoid that unless we get customer feedback stating that they're required.

richardpark-msft (approved):

RickWinter (approved):

PR #23240, opened 2024-07-23
Incorporate common sub config changes to live testing

benbp:
Relies on https://github.com/Azure/azure-sdk-tools/pull/8680

Updates the configs to support federated auth + sovereign cloud/multi subscription testing.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -45,25 +45,25 @@ parameters:
           - eng/common/TestResources/sub-config/AzurePublicMsft.json
       Preview:
         SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources-preview)
-        ServiceConnection: azure-sdk-tests
-        # TODO:
+        ServiceConnection: azure-sdk-tests-preview
         SubscriptionConfigurationFilePaths:
+          - eng/common/TestResources/sub-config/AzurePreviewMsft.json
       Canary:
-        SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
-        ServiceConnection: azure-sdk-tests
-        # TODO:
+        SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources-preview)
+        ServiceConnection: azure-sdk-tests-preview
         SubscriptionConfigurationFilePaths:
+          - eng/common/TestResources/sub-config/AzurePreviewMsft.json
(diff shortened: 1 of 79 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -45,25 +45,25 @@ parameters:
           - eng/common/TestResources/sub-config/AzurePublicMsft.json
       Preview:
         SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources-preview)
```
weshaggard, line 47: Can we remove the sub config as well or is there still stuff in there that isn't in the commited config?

benbp: I was thinking I can either be additive, since go hasn't gone to fed auth by default yet, or I can make all the changes here dependent on fed auth (which I think is required for looking at the git config path behavior?)

RickWinter: Would it make things easier to go to default on? @chlowell can comment if we are ready for that.

weshaggard: @benbp by having both configs you cannot work with both modes because the file sub config clears the client secret. So we will have to be in one world or the other.

benbp: > Would it make things easier to go to default on? @chlowell can comment if we are ready for that.

@RickWinter yes a bit simpler, I'm happy to roll those UseFederatedAuth changes into this PR set as well if you're ready.

chlowell: I believe we're ready for that. All our live tests should just work when switched to federated auth because they use the test credential factory returning `AzurePipelinesCredential` when running in an AzurePowerShell@5 task.

sdk/ai/azopenai/ci.yml
```diff
@@ -35,6 +35,9 @@ extends:
       UsePipelineProxy: false
       CloudConfig:
         Public:
+          ServiceConnection: azure-sdk-tests
+          SubscriptionConfigurationFilePaths:
+            - eng/common/TestResources/sub-config/AzurePublicMsft.json
           SubscriptionConfigurations:
```
weshaggard, line 41: If they are using a different sub with a secret this will likely break.

weshaggard (approved):

Issue #17384, opened 2022-03-25
go-redis/redis - Microsoft Entra Auth Sample


joshfree:
Tracking issue for dogfooding and creating code samples (or documentation) to show best practices for how to use AAD Auth with popular 3rd party OSS Redis clients when connecting to Azure.

Depending on the outcome of the dogfooding:

**Possible Outcome 1:** 0-1 days of extra work (minimal sample code is required to enable the experience):
- Check-in a thin / simple code sample project to our sdk repo.
- Share the sample with the Azure Redis team so that they can then share the sample with customers, publish updated docs, and also validate the service handles credential revocation and similar scenarios to test with the provided sample.

**Possible Outcome 2:** ~1 week of extra work (significant sample code is required to enable the experience):
- Discuss the investigation results with the Identity SDK and Redis teams before proceeding 
- Based on the discussion, decide on whether we should be building a sample or an extension library that is shipped instead.

**Possible Outcome 3:** (using AAD Auth with the client is blocked until code changes are first made to the OSS Redis client):
- Discuss the investigation results with the Identity SDK and Redis teams.
- File GitHub issue with 3rd party project.  Provide as much details as possible, including any code samples, gists, etc.

scottaddie:
go-redis v9.5.2 [added an Options.CredentialsProviderContext feature](https://github.com/redis/go-redis/pull/2998) that supports Microsoft Entra authN. Let's add a code sample to this repo. Once completed, add a new table row to [this Learn doc section](https://learn.microsoft.com/azure/azure-cache-for-redis/cache-azure-active-directory-for-authentication#client-library-support) with a link to the sample.

chlowell:
The example is live in the package docs for [v1.8.0-beta.2](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity@v1.8.0-beta.2#example-package-Redis). I'll leave this issue open to track updating the Learn doc after the stable v1.8.0 release.

chlowell:
Closed by https://github.com/MicrosoftDocs/azure-docs-pr/pull/290302

PR #23266, opened 2024-08-01
Add go-redis authentication sample

chlowell:
Part of #17384

### DIFF
--- a/sdk/azidentity/example_redis_test.go
+++ b/sdk/azidentity/example_redis_test.go
@@ -1,0 +1,78 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azidentity_test
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/redis/go-redis/v9"
+)
+
+// This example demonstrates how to use azidentity to authenticate a [go-redis] client
+// connecting to Azure Cache for Redis. See the [Azure Cache for Redis documentation]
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (approved):

scottaddie:
@chlowell Would it make sense to create a samples directory for this stuff?

chlowell:
What would you hope to improve by placing it somewhere else (not a rhetorical question 😄)? I wrote this to integrate with Go's built-in support for testable examples. Go will compile it as part of testing the library and place it in [this section of the package docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity@v1.8.0-beta.1#pkg-overview). Moving it outside `azidentity` would break that integration, so I don't see a clear benefit in doing that.

scottaddie:
My reasoning is it feels strange to mix source for the library with sample code. I wasn't aware of the testable examples thing you mentioned, so I suppose what you have does make complete sense.

Once this PR merges, do we just need to update the Learn doc I mentioned in the issue?

RickWinter (approved):
This looks good.

Why don't we put this function onto/into Azure.Identity library. Azure Redis is an Azure product, rather than have everyone copy the code we could put the callback logic into the library. Then customers could write

```
func Example_redis() {
	if err != nil {
		// TODO: handle error
	}
	client := redis.NewClient(&redis.Options{
		Addr:                       fmt.Sprintf("%s:6380", "TODO: redis host"),
		CredentialsProviderContext: azidentity.RedisCredentialProvider(credential),
		TLSConfig:                  &tls.Config{MinVersion: tls.VersionTLS12},
	})
	// TODO: use the client
	_ = client
}
```

chlowell:
> My reasoning is it feels strange to mix source code for the library with sample code

Ah, yes, I agree it looks odd, especially because `azidentity` now appears to require `go-redis`. There are a couple conventions/language subtleties that apply here:
- `*_test.go` files aren't part of a published library i.e., applications can't reference their content
- Go recognizes dependencies used only in `*_test.go` files and installs them only when running tests

chlowell:
> Why don't we put this function onto/into Azure.Identity library

I'm wary of doing that for a few reasons:
- we'd have to version this API with Redis's connection protocol
- I want to see evidence of widespread need before adding API
- different Redis clients have different interfaces; I imagine we'd always have some customers who must write adapter code

PR #23268, opened 2024-08-01
Cosmos DB: Add support for Span

### LINKED ISSUE #23185: Add tracing to the Cosmos Client
### Feature Request

I see that many of the packages such as aztables and azidentity have build in tracing support, but it appears that azcosmos does not yet have this.

Is this on the roadmap at all or would you be open to accepting a contribution adding this to the cosmos implementation?

ealsur:
This PR modifies `Client` to have the internal `azcore.Client` embedded to leverage the Pipeline and Trace.

All the methods (except Pagers because they internally already do it) that execute an async operation will now start a Span respectively.

Added a helper SpanValidator to be used in tests to validate that the expected Spans are generated.

## Properties

Reference: https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/database/cosmosdb.md

* The code defines the `TracingOptions` to set `az.namespace` to `Microsoft.DocumentDB`
* `db.system` = `cosmosdb`
* `db.cosmosdb.connection_mode` = `gateway`
* `db.namespace` (when applicable) = datbase id
* `db.collection.name` (when applicable) = container id
* `db.operation.name` = span name (for example `create_container`)
* `server.address` and `server.port` = based on the endpoint (for example `myaccount.documents.azure.com`, `server.port` is added only if it's different than 443)

## Span names

Covers the list in 

https://github.com/Azure/azure-sdk-for-go/blob/29ab5117d5e5a878f37d89ff27eb5b808a137dbc/sdk/data/azcosmos/otel_constants.go#L15-L35

Closes #23185

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-* Adds support for tracing. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
+* Added support for tracing. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
 
 ### Breaking Changes
 
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+* Adds support for tracing. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## 1.0.4 (Unreleased)
 
 ### Features Added
-
 * Added support for tracing. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
 
 ### Breaking Changes
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -27,9 +27,10 @@ const (
 
 // Client is used to interact with the Azure Cosmos DB database service.
 type Client struct {
-	endpoint string
-	internal *azcore.Client
-	gem      *globalEndpointManager
+	endpoint    string
+	internal    *azcore.Client
+	gem         *globalEndpointManager
+	endpointUrl *url.URL
 }
 
 // Endpoint used to create the client.
@@ -42,6 +43,10 @@ func (c *Client) Endpoint() string {
 // cred - The credential used to authenticate with the cosmos service.
 // options - Optional Cosmos client options.  Pass nil to accept default values.
 func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*Client, error) {
+	endpointUrl, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
 	preferredRegions := []string{}
 	enableCrossRegionRetries := true
 	if o != nil {
@@ -57,7 +62,7 @@ func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*C
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, internal: internalClient, gem: gem}, nil
+	return &Client{endpoint: endpoint, endpointUrl: endpointUrl, internal: internalClient, gem: gem}, nil
 }
 
 // NewClient creates a new instance of Cosmos client with Azure AD access token authentication. It uses the default pipeline configuration.
@@ -65,7 +70,11 @@ func NewClientWithKey(endpoint string, cred KeyCredential, o *ClientOptions) (*C
 // cred - The credential used to authenticate with the cosmos service.
 // options - Optional Cosmos client options.  Pass nil to accept default values.
 func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (*Client, error) {
-	scope, err := createScopeFromEndpoint(endpoint)
+	endpointUrl, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	scope, err := createScopeFromEndpoint(endpointUrl)
 	if err != nil {
 		return nil, err
 	}
@@ -83,7 +92,7 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	return &Client{endpoint: endpoint, internal: internalClient, gem: gem}, nil
+	return &Client{endpoint: endpoint, endpointUrl: endpointUrl, internal: internalClient, gem: gem}, nil
 }
 
 // NewClientFromConnectionString creates a new instance of Cosmos client from connection string. It uses the default pipeline configuration.
@@ -160,13 +169,8 @@ func newInternalPipeline(authPolicy policy.Policy, options *ClientOptions) azrun
 		&options.ClientOptions)
 }
 
-func createScopeFromEndpoint(endpoint string) ([]string, error) {
-	u, err := url.Parse(endpoint)
-	if err != nil {
-		return nil, err
-	}
-
-	return []string{fmt.Sprintf("%s://%s/.default", u.Scheme, u.Hostname())}, nil
+func createScopeFromEndpoint(endpoint *url.URL) ([]string, error) {
+	return []string{fmt.Sprintf("%s://%s/.default", endpoint.Scheme, endpoint.Hostname())}, nil
 }
 
 // NewDatabase returns a struct that represents a database and allows database level operations.
@@ -208,7 +212,7 @@ func (c *Client) CreateDatabase(
 	databaseProperties DatabaseProperties,
 	o *CreateDatabaseOptions) (DatabaseResponse, error) {
 	var err error
-	spanName, err := getSpanNameForDatabases(c.getAccountEndpoint(), operationTypeCreate, resourceTypeDatabase, databaseProperties.ID)
+	spanName, err := getSpanNameForDatabases(c.accountEndpointUrl(), operationTypeCreate, resourceTypeDatabase, databaseProperties.ID)
 	if err != nil {
 		return DatabaseResponse{}, err
 	}
@@ -274,7 +278,7 @@ func (c *Client) NewQueryDatabasesPager(query string, o *QueryDatabasesOptions) 
 		Fetcher: func(ctx context.Context, page *QueryDatabasesResponse) (QueryDatabasesResponse, error) {
 			var err error
 			// Move the span to the pager once https://github.com/Azure/azure-sdk-for-go/issues/23294 is fixed
-			spanName, err := getSpanNameForClient(c.getAccountEndpoint(), operationTypeQuery, resourceTypeDatabase, c.getAccountEndpoint().Hostname())
+			spanName, err := getSpanNameForClient(c.accountEndpointUrl(), operationTypeQuery, resourceTypeDatabase, c.accountEndpointUrl().Hostname())
 			if err != nil {
 				return QueryDatabasesResponse{}, err
 			}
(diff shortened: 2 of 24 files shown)


jhendrixMSFT (approved):

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -41,6 +41,9 @@ func (c *ContainerClient) ID() string {
 func (c *ContainerClient) Read(
 	ctx context.Context,
 	o *ReadContainerOptions) (ContainerResponse, error) {
+	var err error
+	ctx, endSpan := runtime.StartSpan(ctx, "ContainerClient.Read", c.database.client.internal.Tracer(), nil)
```
kushagraThapar, line 45: Though this is consistent with the other SDKs, I am curious if we should be more explicit here with the span name?
Also, in other Cosmos SDKs, we add additional information on the span names, like container id in this case.

From Java SDK - `this.readContainerSpanName = "readContainer." + this.id;`

I wonder if this has been considered in GO language? It is based on the feedback from customers, it really helps them to see which database / container these operations are for?

ealsur: In Go, it seems that's the pattern they've used:

https://github.com/Azure/azure-sdk-for-go/blob/f53cbcb8e060bc481a0893e003fbcf1dfddbf28e/sdk/data/aztables/service_client.go#L118

@jhendrixMSFT do you have any context?

kushagraThapar: Thanks @ealsur, tagging @lmolkova to provide her guidance here.

sdk/data/azcosmos/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.0.4 (Unreleased)
 
 ### Features Added
+* Added support for tracing. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
```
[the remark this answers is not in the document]
ealsur:
```suggestion
* Added support for OpenTelemetry trace spans. See [PR 23268](https://github.com/Azure/azure-sdk-for-go/pull/23268)
```

sdk/data/azcosmos/cosmos_container.go
lmolkova: CosmosDB span naming pattern is described here https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/database/database-spans.md. The format has slightly changed recently and it's slightly different from what Java and .NET populate.

We can stay consistent with Cosmos DB .NET/Java for now and update them all together later on or implement the latest version of OTel semantic conventions in Go and tackle Java and .NET later on. 

According to the latest version, the span name should be `readContainer {container name}` when performing operations within the container and `{operation} {database}` when doing some operations against the whole database.

The otel semconv take a priority over Azure SDK naming conventions, but otel semconv are defined for a (relatively) small subset of SDKs.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -471,6 +511,9 @@ func (c *ContainerClient) PatchItem(
 	itemId string,
 	ops PatchOperations,
 	o *ItemOptions) (ItemResponse, error) {
+	var err error
+	ctx, endSpan := runtime.StartSpan(ctx, "ContainerClient.PatchItem", c.database.client.internal.Tracer(), nil)
```
lmolkova, line 515: We should also populate attributes according to semantic convention on this and other spans, see  https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/database/cosmosdb.md. 

I think it could also be done as a follow up. /cc @jcocchi

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -75,6 +79,9 @@ func (c *ContainerClient) Replace(
 	ctx context.Context,
 	containerProperties ContainerProperties,
 	o *ReplaceContainerOptions) (ContainerResponse, error) {
+	var err error
+	ctx, endSpan := runtime.StartSpan(ctx, "ContainerClient.Replace", c.database.client.internal.Tracer(), nil)
```
lmolkova, line 83: @jhendrixMSFT do you happen to remember if we support named tracers in Go?
In other words, what would be the name of this tracer? 
Also, what would be the way to specify schema_url (it contains semantic convention version) for it? 

(excuse my Go ignorance 😳)

sdk/data/azcosmos/cosmos_container.go
ealsur: Added span names following https://github.com/ealsur/azure-sdk-for-go/blob/cd4154850235edd9165a4b72f5e21cbeacc97ef6/sdk/data/azcosmos/otel_constants.go#L9-L29

sdk/data/azcosmos/cosmos_container.go
ealsur: Properties are now wired

sdk/data/azcosmos/otel_constants.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func getSpanPropertiesForDatabase(endpoint *url.URL, operationName string, id string) runtime.StartSpanOptions {
+	return runtime.StartSpanOptions{Attributes: []tracing.Attribute{
+		{Key: "db.system", Value: "cosmosdb"},
+		{Key: "db.cosmosdb.connection_mode", Value: "gateway"},
+		{Key: "db.namespace", Value: id},
+		{Key: "db.operation.name", Value: operationName},
+		{Key: "server.address", Value: endpoint.Hostname()},
+		{Key: "server.port", Value: endpoint.Port()},
```
lmolkova, line 162: WDYT about only setting the port if it's not default (i.e. not 443) ? I assume in absolute majority of cases it is and it's just wasteful to set it.

I'm also proposing to change it here: https://github.com/open-telemetry/semantic-conventions/pull/1330

sdk/data/azcosmos/otel_constants.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+const (
+	otelSpanNameCreateDatabase              = "create_database"
+	otelSpanNameReadDatabase                = "read_database"
+	otelSpanNameDeleteDatabase              = "delete_database"
+	otelSpanNameQueryDatabases              = "query_databases"
+	otelSpanNameReadThroughputDatabase      = "read_database_throughput"
+	otelSpanNameReplaceThroughputDatabase   = "replace_database_throughput"
+	otelSpanNameCreateContainer             = "create_container"
```
lmolkova, line 21: check out discussion on whether we also need `create_container|database_if_exists` in the chat, also fine to do as a follow-up

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -249,6 +276,14 @@ func (c *Client) NewQueryDatabasesPager(query string, o *QueryDatabasesOptions)
...
 		},
 		Fetcher: func(ctx context.Context, page *QueryDatabasesResponse) (QueryDatabasesResponse, error) {
+			var err error
+			// Move the span to the pager once https://github.com/Azure/azure-sdk-for-go/issues/23294 is fixed
+			spanName, err := getSpanNameForClient(c.accountEndpointUrl(), operationTypeQuery, resourceTypeDatabase, c.accountEndpointUrl().Hostname())
+			if err != nil {
+				return QueryDatabasesResponse{}, err
+			}
+			ctx, endSpan := azruntime.StartSpan(ctx, spanName.name, c.internal.Tracer(), &spanName.options)
+			defer func() { endSpan(err) }()
```
lmolkova, line 286: we also need to report a few more attributes based on the response:
- db.cosmosdb.request_charge
- db.cosmosdb.status_code
- `db.cosmosdb.sub_status_code`

sdk/data/azcosmos/otel_constants.go
ealsur: These operations don't exist in this SDK.

sdk/data/azcosmos/otel_constants.go
ealsur: I can do that change, sure

sdk/data/azcosmos/cosmos_client.go
ealsur: These attributes only are available on the response (if succeeded) but Go SDK atm does not provide a way to add Attributes after span creation. They are already adding some common properties (like status code).

https://github.com/Azure/azure-sdk-for-go/blob/e8defc77367f50a8e61ba3cf9fbe221d81d4f39d/sdk/azcore/runtime/policy_http_trace.go#L23-L32

sdk/data/azcosmos/cosmos_container.go
lmolkova: > For named tracers, are you talking about this or something else?

I mean here

https://github.com/Azure/azure-sdk-for-go/blob/969cfcd7a36018cab9461bc768bdd0c8f1ff4e00/sdk/tracing/azotel/otel.go#L31

when we create a tracer, what would be the value of the namespace? 

> schema_url is set by [azotel](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/tracing/azotel/otel.go#L31) when the adapter is created. Do we need SDKs to set it?

Yes, I think so - cosmos and servicebus may generate different schema versions and should be able to express it - I'll create an issue to follow up - it's not blocking this PR.

sdk/data/azcosmos/cosmos_client.go
lmolkova: > Go SDK atm does not provide a way to add Attributes after span creation

@jhendrixMSFT is there some trick? we need to be able to add attributes about the response for non-HTTP cases. I can create an issue and we can follow up  on the Cosmos SDK once it's possible.

sdk/data/azcosmos/cosmos_container.go
lmolkova: created https://github.com/Azure/azure-sdk-for-go/issues/23307

sdk/data/azcosmos/cosmos_client.go
lmolkova: Created https://github.com/Azure/azure-sdk-for-go/issues/23308

lmolkova (approved):

sdk/data/azcosmos/cosmos_client.go
kjg: Hi! Would something like this work to set the response attributes?

```go
span := c.internal.Tracer().SpanFromContext(ctx)
span.SetAttributes([]tracing.Attribute{
	{Key: "db.cosmosdb.request_charge", Value: res.RequestCharge
})
```

ealsur: I don't know what's the expected contract for SpanFromContext if something creates children spans (or the current span has a parent?) If the answer to the linked issue is the above, we can certainly add it.

kjg: If I'm reading the code correctly `azruntime.StartSpan` returns a new context in which the span is the only possible span for that specific context, so I believe that should work.

I did just test this out, and for some reason the attributes passed into `runtime.StartSpan` `spanName.options` aren't being displayed for us. I'm still trying to figure out why.

kjg: >I did just test this PR out, and for some reason the attributes passed into runtime.StartSpan via spanName.options aren't being displayed for us. I'm still trying to figure out why.

Nevermind, this is working most of the time. It's just the span started by the `NewQueryItemsPager` `Fetcher` function that's not working for me. I see the `Pager[QueryItemsResponse].NextPage` span and the underlying HTTP POST span, but not the `query_items` span

ealsur: Yeah, the query's are still not working as expected (see the Issue I linked in that part in the code)

ealsur: Once that is fixed, then our span will appear (or we'll have to move the Span name and properties to a new input in the Pager)

kjg: Yeah, I see that "by policy" nested spans are rejected in the tracer code, that makes sense. Thanks!

https://github.com/Azure/azure-sdk-for-go/blob/b31a752c22d8046d0e5d28ef4df05d111f2a6cf1/sdk/azcore/runtime/policy_http_trace.go#L120-L125

ealsur: @kjg for reference, this is the feature request we have https://github.com/Azure/azure-sdk-for-go/issues/23294

kjg: Looks great, thanks for all of this, it is already quite valuable for us!

PR #23269, opened 2024-08-01
[azsystemevents] Regenerate based on latest update (adding in storage tiers to some system events)

richardpark-msft:
- Generate based on latest hash
- Update to latest emitter which culls out unreferenced models.

APIView here: https://apiview.dev/Assemblies/Review/db71641e728c4b7ba62abe3efde3cf7e/e0dacee4cb2f4d6d8bce12bcea9e9ba8?diffRevisionId=0ca000f8e5ed4f8a98275c18cfddcbfc&doc=False&diffOnly=True


richardpark-msft:
Hm. Just noticed the generator failed to generate one of the types. Investigating...

richardpark-msft:
This is broken at the moment - the TypeSpec needs to be adjusted so the enum + the model aren't orphaned, which is causing them to be removed before they're emitted.

richardpark-msft:
We're back to working again. Needed to update all of our packages and get a bug fixed for an accidental model cull. 

So there'll be some diffs, but the core of the change was just to get those two fields added in for the storage events.

jhendrixMSFT (approved):

PR #23270, opened 2024-08-02
azcontainerregistry: move method to function

stevekuznetsov:
The Go documentation strongly discourages structs from having methods with both pointer recievers and non-pointer recievers. For the auth policy, `getChallengeRequest` presumably did not have a pointer reciever since it did not make use of the struct at all. We can simply move this to a function.

### DIFF
--- a/sdk/containers/azcontainerregistry/authentication_policy.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"net/http"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


sdk/containers/azcontainerregistry/authentication_policy.go
```diff
@@ -11,12 +11,13 @@ import (
...
 	"errors"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"net/http"
 	"strings"
 	"sync/atomic"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+
```
chlowell, line 20:
```suggestion
```

chlowell (approved):
Thanks for contributing this!

stevekuznetsov:
@tadelesh sorry, looks like auto-merge didn't go through. Updated the formatting, pushed again.

Issue #20571, opened 2023-04-07
Expose azcontainerregistry.authenticationClient


PSanetra:
### Feature Request

To interact with the ACR api it is needed to get an ACR refresh token via the `/oauth2/exchange` endpoint and then using the refresh token to get an access token via the `/oauth2/endpoint`, like it is described in the [ACR docs](https://github.com/Azure/acr/blob/1484421ee09a745565b523eaf539dda0ed24a0ae/docs/AAD-OAuth.md#getting-credentials-programmatically).

Third party libraries which interact with container and artifact registry like `oras-go` already support using refresh tokens to get an access token, but currently the call to `/oauth2/exchange` needs to be implemented manually. It would be great if the `azcontainerregistry.authenticationClient` was exposed, so that [ExchangeAADAccessTokenForACRRefreshToken](https://github.com/Azure/azure-sdk-for-go/blob/0a423000ff47312c4ba1bbb5a8d9e1a0ad007b5c/sdk/containers/azcontainerregistry/authentication_client.go#L29-L50) can be used from the offical Azure go sdk.

Related: https://github.com/oras-project/oras-go/discussions/476

tadelesh:
@PSanetra Thank you for the feature request. When we reviewed our public surface of `azcontianerregistry`, we think the auth process should be transparent to our SDK users (when using any registry API, SDK will help to do the auth correctly), so we decided to hide the auth related API. Could you elaborate your scenario of using token exchange API that we could evaluate whether expose those API?

PSanetra:
@tadelesh yes, see for example the discussion in the oras go library: https://github.com/oras-project/oras-go/discussions/476

Due to missing support to call the token exchage via the SDK, it is needed to implement the token exchange manually. I think generally manual implementation is currently needed when a client should be implemented that should be able to communicate with generic container or artifact registries.

tadelesh:
If I understand correctly, you mean you need to use oras-go to communicate with generic registries, only need azure sdk to do the auth, right?

@JeffreyRichter could you help to evaluate this request?

PSanetra:
@tadelesh exactly :slightly_smiling_face:

JeffreyRichter:
Let me see if I understand...
[...]

PR #23272, opened 2024-08-02
azcontainerregistry: delegate token caching

stevekuznetsov:
azcontainerregistry: decouple token exchange from request specifics

I would like to propose a generally-useful token cache in a future
commit. Managing the refresh tokens and exchanging them for access
tokens does not need to rely specifically on the authentication policy
or the specific client request being made, so this commit decouples the
two concerns, thereby enabling the future refactor to be simpler.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

azcontainerregistry: delegate token caching

Mechanically refactor token caching, retrieval and exchange behavior
from the authentication policy into a dedicated object, allowing us to
separate token management concerns from Azure client policy concerns.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

azcontainerregistry: move code

This commit simply moves code from one location to another, broken out
from other commits for ease of review.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

Builds on top of #23270 

Proving out a separation of concerns that could solve https://github.com/Azure/azure-sdk-for-go/issues/20571

### DIFF
--- a/sdk/containers/azcontainerregistry/authentication_policy.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy.go
@@ -7,6 +7,7 @@
 package azcontainerregistry
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


chlowell:
I don't see anything wrong with this, however I also don't see a reason to make these changes, because they don't affect client behavior or the module's public API. Can you please describe the scenario motivating this? I take it you want to replace the ACR token cache of a `BlobClient` or `Client` instance with a custom implementation. Why is the current internal caching function not sufficient?

stevekuznetsov:
@chlowell were you aware of the conversation happening on https://github.com/Azure/azure-sdk-for-go/issues/20571? Sorry if I was not clear - I 100% agree with you that the changes here as they are will not be user-visible, but their value comes from the context of that issue.

stevekuznetsov:
Rebased after pre-requisite PR merged.

stevekuznetsov:
Would be great to do it but not sure if maintainers want this feature.

PR #23285, opened 2024-08-05
azcontainerregistry: expose authentication client, options

stevekuznetsov:
The ACR authentication API is a public-facing API. Consumers would like to use the well-tested code in the SDK to interface with this API instead of having to roll their own. This commit simply exposes all of the structs, functions and constants that are necessary to consume this API.

Part of https://github.com/Azure/azure-sdk-for-go/issues/20571

### DIFF
--- a/sdk/containers/azcontainerregistry/.gitignore
+++ b/sdk/containers/azcontainerregistry/.gitignore
@@ -1,0 +1,3 @@
+package-lock.json
+package.json
+node_modules
--- a/sdk/containers/azcontainerregistry/.gitignore
+++ b/sdk/containers/azcontainerregistry/.gitignore
@@ -1,3 +1,0 @@
-package-lock.json
-package.json
-node_modules
--- a/sdk/containers/azcontainerregistry/authentication_client_test.go
+++ b/sdk/containers/azcontainerregistry/authentication_client_test.go
@@ -8,22 +8,23 @@ package azcontainerregistry
 
 import (
 	"context"
+	"net/http"
+	"reflect"
+	"strings"
+	"testing"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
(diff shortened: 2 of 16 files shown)


sdk/containers/azcontainerregistry/authentication_client_test.go
```diff
@@ -8,22 +8,23 @@ package azcontainerregistry
 
 import (
 	"context"
+	"net/http"
```
stevekuznetsov, line 11: Didn't notice my editor do this, but it might be nice to turn on import sorting!

tadelesh:
it's better to change the config [here](https://github.com/stevekuznetsov/azure-sdk-for-go/blob/skuznets/expose-auth-api/sdk/containers/azcontainerregistry/autorest.md) and use `go generate` to regen the code instead of manually change the code.

stevekuznetsov:
@tadelesh thanks for the pointer, done!

sdk/containers/azcontainerregistry/.gitignore
chlowell, line 1: Do we need this? We shouldn't have any node artifacts to ignore

tadelesh:
> @tadelesh thanks for the pointer, done!

need to remove `.gitignore`, others lgtm, thanks for the contribution.

sdk/containers/azcontainerregistry/.gitignore
stevekuznetsov: Gotcha, sorry - how were you all installing `autorest`? Some README or similar would be helpful.

stevekuznetsov:
`.gitignore` removed.

sdk/containers/azcontainerregistry/.gitignore
chlowell: 👍 agree it's opaque--typically only SDK developers generate clients. Short answer is, we install `autorest` globally; more detailed docs are at https://aka.ms/autorest

chlowell (approved):
Thanks for doing this!

PR #23286, opened 2024-08-06
fix generator tool: ExecuteGoGenerate func

Alancere:
fix: https://github.com/Azure/azure-rest-api-specs/pull/30000/checks?check_run_id=28389745036

filter go: downloading log...

### DIFF
--- a/eng/tools/generator/cmd/v2/common/cmdProcessor.go
+++ b/eng/tools/generator/cmd/v2/common/cmdProcessor.go
@@ -60,8 +60,22 @@ func ExecuteGoGenerate(path string) error {
 
 	if err != nil || stderrBuffer.Len() > 0 {
 		if stderrBuffer.Len() > 0 {
-			fmt.Println(stderrBuffer.String())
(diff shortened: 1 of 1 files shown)


eng/tools/generator/cmd/v2/common/cmdProcessor.go
```diff
@@ -60,8 +60,22 @@ func ExecuteGoGenerate(path string) error {
...
+			lines := strings.Split(stderrBuffer.String(), "\n")
+			newLines := make([]string, 0, len(lines))
+			for _, line := range lines {
+				if !strings.HasPrefix(strings.TrimSpace(line), "go: downloading") {
+					newLines = append(newLines, line)
+				}
+			}
+
+			if len(newLines) > 0 {
+				fmt.Println(strings.Join(newLines, "\n"))
```
raych1, line 74: How about creating a variable to store the joined strings instead of calling it twice?

raych1 (approved):

PR #23305, opened 2024-08-09
[azservicebus,azeventhubs] Fixing security issue by forking the dep and updating it

### LINKED ISSUE #23283: Remove azservicebus ApplicationInsights-Go requirement
After #23231, Component Governance still flags `azservicebus` for requiring `yaml.v2@v2.2.1`, apparently because that version still appears in `go mod graph` output. The lowest-effort solution that comes to my mind is moving the `stress` package into a separate module under `testdata`, [like this](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity/testdata/managed-id-test).

richardpark-msft:
We use ApplicationInsights-Go for our stress tests, so they can report metrics, logging and events. However, that dependency is basically frozen, so no more updates are being made to it. 

I forking that repo, updated it and am now referencing that.

Fixes #23283

### DIFF
--- a/sdk/messaging/azeventhubs/go.mod
+++ b/sdk/messaging/azeventhubs/go.mod
@@ -34,3 +34,7 @@ require (
 
 // used in stress tests
 require github.com/microsoft/ApplicationInsights-Go v0.4.4
+
+// NOTE: ApplicationInsights-Go is a frozen project. We only use it for our stress tests to report metrics so we've just
+// done a small update to keep it's dependencies up to date.
+replace github.com/microsoft/ApplicationInsights-Go v0.4.4 => github.com/richardpark-msft/ApplicationInsights-Go v0.0.0-20240809214047-fa65343f2c29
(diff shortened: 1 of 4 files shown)


chlowell:
Release pipelines will block on `replace`, and is the Go compiler smart enough to shake your fork out of customer builds?

jhendrixMSFT:
Can we move the stress goop under `testdata` for an extra level of insurance?

Also, why is a `replace` necessary? If you've forked it, you should just directly reference your module.

richardpark-msft:
> Can we move the stress goop under `testdata` for an extra level of insurance?

They're just normal tests, I'm not sure why they're considered goop. If I move them under testdata I assume they'll no longer get compiled along with the module, correct?
 
> Also, why is a `replace` necessary? If you've forked it, you should just directly reference your module.

I forked it but it's still living under the github.com/Microsoft/Application-Insights-Go module path. I could go one step further and rename everything in it, but I was just doing a minimal thing.

richardpark-msft:
> Release pipelines will block on `replace`, and is the Go compiler smart enough to shake your fork out of customer builds?

You had a replace directive in there before so that's why I ended up using it.

And yes, they don't get compiled into customer builds as the package that's in there is internal and not referenced by the module code itself.

jhendrixMSFT:
Content under `testdata` is supposed to get pruned from a module (I'm having a hard time finding the docs on this). So, since the stress test really is a stand-alone executable with a bunch of other content to make it work (docker stuff, bicep files, etc), it would be ideal to omit this from the resultant module. We do this for the various perf tests ([example](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/data/aztables/testdata/perf)).

I forgot we had a block on release when there's a `replace` in the `go.mod` file. So I assume what Charles did was something temporary to make CG temporarily happy?

PR #23329, opened 2024-08-14
azcontainerregistry tests build images instead of importing them

### LINKED ISSUE #22988: azcontainerregistry live tests fail due to Docker Hub rate limit
These tests populate a newly created registry with a handful of images. This occasionally fails because Docker Hub doesn't allow the test code to pull all the images. The error message is `You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit`. I guess this rate limiting is based on the IP of the VM running the test suite, so the tests implicitly share a limit with whatever's previously run on that VM/NIC.

Off the top of my head, solutions are:
- authenticate to Docker Hub to access a higher limit
- pull fewer images
- don't pull from Docker Hub

Third option is probably the best--we could pull from [Microsoft Artifact Registry](https://mcr.microsoft.com/) instead.

chlowell:
Live tests are flaky and usually fail today because they import images from Docker Hub to the test registry, and Docker Hub throttles anonymous pulls. Furthermore, the test suite is hard to maintain because relationships and dependencies between imported images and test cases aren't clear. Some tests refer to images by name, some refer to images or their layers by hard-coded digest, tests can break each other by performing destructive operations on a shared image, etc. This PR addresses these problems by building images on demand via ACR scheduled tasks and organizing tests around the images they operate on so tests own images and don't need to hard-code any details.

Closes #22988

### DIFF
--- a/sdk/containers/azcontainerregistry/authentication_policy_test.go
+++ b/sdk/containers/azcontainerregistry/authentication_policy_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/temporal"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/test/credential"
@@ -201,20 +200,11 @@ func Test_authenticationPolicy_anonymousAccess(t *testing.T) {
 	endpoint, _, options := getEndpointCredAndClientOptions(t)
 	client, err := NewClient(endpoint, nil, &ClientOptions{ClientOptions: options})
 	require.NoError(t, err)
-	ctx := context.Background()
 	pager := client.NewListRepositoriesPager(nil)
-	repositoryName := ""
 	for pager.More() {
-		page, err := pager.NextPage(ctx)
+		_, err = pager.NextPage(ctx)
 		require.NoError(t, err)
-		require.NotEmpty(t, page.Repositories.Names)
-		if repositoryName == "" {
-			repositoryName = *page.Repositories.Names[0]
-		}
 	}
-	require.NotEmpty(t, repositoryName)
-	_, err = client.UpdateRepositoryProperties(ctx, repositoryName, &ClientUpdateRepositoryPropertiesOptions{Value: &RepositoryWriteableProperties{CanDelete: to.Ptr(true)}})
-	require.Error(t, err)
 }
 
 func Test_getChallengeRequest(t *testing.T) {
--- a/sdk/containers/azcontainerregistry/blob_client_test.go
+++ b/sdk/containers/azcontainerregistry/blob_client_test.go
@@ -125,12 +125,13 @@ func TestBlobClient(t *testing.T) {
 		res, err := client.GetBlob(ctx, repository, blobDigest, nil)
 		require.NoError(t, err)
 		require.NotEmpty(t, *res.ContentLength)
-		if recording.GetRecordMode() != recording.PlaybackMode {
-			reader, err := NewDigestValidationReader(blobDigest, res.BlobData)
-			require.NoError(t, err)
-			_, err = io.ReadAll(reader)
-			require.NoError(t, err)
+		reader, err := NewDigestValidationReader(blobDigest, res.BlobData)
+		require.NoError(t, err)
+		if recording.GetRecordMode() == recording.PlaybackMode {
+			reader.digestValidator = &sha256Validator{&fakeHash{}}
 		}
+		_, err = io.ReadAll(reader)
+		require.NoError(t, err)
 	})
 
 	t.Run("GetBlob_fail", func(t *testing.T) {
@@ -164,12 +165,13 @@ func TestBlobClient(t *testing.T) {
 			}
 			current += chunkSize
 		}
-		if recording.GetRecordMode() != recording.PlaybackMode {
-			reader, err := NewDigestValidationReader(blobDigest, blob)
-			require.NoError(t, err)
-			_, err = io.ReadAll(reader)
-			require.NoError(t, err)
+		reader, err := NewDigestValidationReader(blobDigest, blob)
+		require.NoError(t, err)
+		if recording.GetRecordMode() == recording.PlaybackMode {
+			reader.digestValidator = &sha256Validator{&fakeHash{}}
 		}
+		_, err = io.ReadAll(reader)
+		require.NoError(t, err)
 	})
 
 	t.Run("GetChunk_fail", func(t *testing.T) {
--- a/sdk/containers/azcontainerregistry/blob_client_test.go
+++ b/sdk/containers/azcontainerregistry/blob_client_test.go
@@ -8,27 +8,194 @@ package azcontainerregistry
 
 import (
 	"bytes"
-	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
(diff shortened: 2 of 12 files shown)


richardpark-msft:
Conceptually this makes sense (there's a lot of changes, and I'm not familiar with this test suite). 

Microsoft also has a registry, if you're just trying to get a base image, and we use this as part of our infrastructure for stress and (AFAIK) it doesn't throttle.

This is an example of a public image for mariner and there are quite a few others that are even smaller:
https://mcr.microsoft.com/en-us/product/cbl-mariner/distroless/minimal/tags

chlowell:
I suppose we could keep a list of smallish images and import them on demand, panicking if we need more images than the list contains, but the only advantage I see in that approach over ACR builds is that it doesn't require the Azure CLI. The images I'm building with this PR are tens of bytes and require no ongoing maintenance.

richardpark-msft:
> I suppose we could keep a list of smallish images and import them on demand, panicking if we need more images than the list contains, but the only advantage I see in that approach over ACR builds is that it doesn't require the Azure CLI. The images I'm building with this PR are tens of bytes and require no ongoing maintenance.

Yeah, I was just addressing this part from your commit message:

> Live tests are flaky and usually fail today because they import images from Docker Hub to the test registry, and Docker Hub throttles anonymous pulls. 

I took that to mean the primary issue was just connecting to Docker Hub in the first place.

Another option is you can just check-in the tarball that you built, once, and then just upload the tarball. No need to docker build anything at that point. Just FYI, `docker save` on your Dockerfile's built container is 13KB.

These are just ideas on how to reduce moving parts if you're interested, and it'd remove a requirement that the user have Docker on their machine to run these tests.

sdk/containers/azcontainerregistry/blob_client_test.go
```diff
@@ -8,99 +8,221 @@ package azcontainerregistry
...
+	res, err := client.GetManifest(ctx, image, imageDigest, &ClientGetManifestOptions{Accept: to.Ptr("application/vnd.docker.distribution.manifest.v2+json")})
 	require.NoError(t, err)
-	_, err = client.CancelUpload(ctx, *startRes.Location, nil)
+	reader, err := NewDigestValidationReader(*res.DockerContentDigest, res.ManifestData)
 	require.NoError(t, err)
+	manifest, err := io.ReadAll(reader)
+	require.NoError(t, err)
+	blobDigest := string(regexp.MustCompile("(sha256:[a-f0-9]{64})").Find(manifest))
+	require.NotEmpty(t, blobDigest)
+	_, hash, found := strings.Cut(blobDigest, ":")
```
richardpark-msft, line 44: 1.18 feature! :)

sdk/containers/azcontainerregistry/blob_client_test.go
```diff
@@ -8,99 +8,221 @@ package azcontainerregistry
...
+		getRes, err := client.GetBlob(ctx, repository, blobDigest, nil)
+		require.NoError(t, err)
+		blob, err := io.ReadAll(getRes.BlobData)
+		require.NoError(t, err)
+		startRes, err := client.StartUpload(ctx, "hello-world", nil)
+		require.NoError(t, err)
+		uploadResp, err := client.uploadChunk(ctx, *startRes.Location, streaming.NopCloser(bytes.NewReader(blob)), nil)
+		require.NoError(t, err)
+		_, err = client.completeUpload(ctx, "sha256:00000000", *uploadResp.Location, nil)
+		require.Error(t, err)
```
richardpark-msft, line 108: Is this error something we can test against (ie, one with an actual status code?). Or is it just text, and the user can't do much with it?

sdk/containers/azcontainerregistry/blob_client_test.go
```diff
@@ -8,99 +8,221 @@ package azcontainerregistry
...
+		endpoint, cred, options := getEndpointCredAndClientOptions(t)
+		client, err := NewBlobClient(endpoint, cred, &BlobClientOptions{ClientOptions: options})
+		require.NoError(t, err)
+		getRes, err := client.GetBlob(ctx, repository, blobDigest, nil)
+		require.NoError(t, err)
+		blob, err := io.ReadAll(getRes.BlobData)
+		require.NoError(t, err)
+		startRes, err := client.StartUpload(ctx, "hello-world", nil)
+		require.NoError(t, err)
+		uploadResp, err := client.uploadChunk(ctx, *startRes.Location, streaming.NopCloser(bytes.NewReader(blob)), nil)
```
richardpark-msft, line 105: Is there no corresponding public function for this? Ie, is this just us sneaking in so we can test something users don't normally look at?

sdk/containers/azcontainerregistry/blob_client_test.go
```diff
@@ -8,99 +8,221 @@ package azcontainerregistry
...
+	t.Run("GetBlob", func(t *testing.T) {
+		startRecording(t)
+		endpoint, cred, options := getEndpointCredAndClientOptions(t)
+		client, err := NewBlobClient(endpoint, cred, &BlobClientOptions{ClientOptions: options})
+		require.NoError(t, err)
+		res, err := client.GetBlob(ctx, repository, blobDigest, nil)
+		require.NoError(t, err)
+		require.NotEmpty(t, *res.ContentLength)
+		if recording.GetRecordMode() != recording.PlaybackMode {
+			reader, err := NewDigestValidationReader(blobDigest, res.BlobData)
```
richardpark-msft, line 129: Why doesn't this work in playback mode, out of curiosity? Or are we avoiding trying to save off a huge blob (understandable).

richardpark-msft (approved):
It looks okay to me. I did a cursory pass over the tests and had a few comments, but nothing that'd block the PR.

sdk/containers/azcontainerregistry/authentication_policy_test.go
```diff
@@ -201,20 +200,11 @@ func Test_authenticationPolicy_anonymousAccess(t *testing.T) {
...
+		_, err = pager.NextPage(ctx)
 		require.NoError(t, err)
-		require.NotEmpty(t, page.Repositories.Names)
-		if repositoryName == "" {
-			repositoryName = *page.Repositories.Names[0]
-		}
 	}
-	require.NotEmpty(t, repositoryName)
-	_, err = client.UpdateRepositoryProperties(ctx, repositoryName, &ClientUpdateRepositoryPropertiesOptions{Value: &RepositoryWriteableProperties{CanDelete: to.Ptr(true)}})
-	require.Error(t, err)
```
tadelesh, line 217: why we remove this? it is aim to ensure anonymous auth could only have the read premission

tadelesh (commented):
instead of build image when running test, is it better to have them uploaded to testdata? they should be not big i think.

sdk/containers/azcontainerregistry/authentication_policy_test.go
chlowell: Our client doesn't enforce access restrictions--that's the server's responsibility--so this part of the test was testing the server, not our client.

sdk/containers/azcontainerregistry/blob_client_test.go
chlowell: Its public face is [UploadChunk](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry#BlobClient.UploadChunk). I didn't write this test, however I would guess the motivation for calling the unexported method was to avoid faking the digest calculator.

sdk/containers/azcontainerregistry/blob_client_test.go
chlowell: In most cases (as in this test specifically) it's a `ResponseError`.

sdk/containers/azcontainerregistry/blob_client_test.go
chlowell: It was very exciting, truly Cutting edge

sdk/containers/azcontainerregistry/blob_client_test.go
chlowell: Good question. The validation reader compares a digest it computes to the server's. This breaks in playback because I recorded a static fake digest instead of the real thing, to prevent tests in general depending on values that could change. But instead of skipping digest validation In playback, these cases should use the `fakeHash` type I added in this PR, which returns a sum matching the recorded value; I'll make that change.

chlowell:
> instead of build image when running test, is it better to have them uploaded to testdata?

It could be better in that you might not need the Azure CLI to run these tests live (you may need Docker, a Storage account, and/or nontrivial management plane code). It would be worse in that we'd have to maintain the pre-built image collection alongside the test suite and take care not to reintroduce races.

tadelesh (approved):

PR #23331, opened 2024-08-14
Remove isWriteOperation on container delete

kjg:
Hi! For some reasons I'm getting back 500 errors when I attempt to delete a cosmos container. If I remove `isWriteOperation: true,` the Delete is successful.

I'm not sure if this is a server issue, or a client issue, but my hopes is that this PR can either spark the investigation or lead to a fix.

Interestingly I see that `(Database).Delete` also uses `isWriteOperation: true,` and that operation is successful for me (both with and without the option set)

I didn't see any tests added when this option was [first added](https://github.com/Azure/azure-sdk-for-go/pull/22394), so I'm not sure if there are any tests might be relevant for this change. I'd be happy to update the CHANGELOG if this is accepted, but wanted to start the discussion first since I think there is a chance that this is a server side issue (since Database deletion works).

### DIFF
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -138,9 +138,8 @@ func (c *ContainerClient) Delete(
 	}
 
 	operationContext := pipelineRequestOptions{
-		resourceType:     resourceTypeCollection,
-		resourceAddress:  c.link,
-		isWriteOperation: true,
+		resourceType:    resourceTypeCollection,
+		resourceAddress: c.link,
 	}
 
 	path, err := generatePathForNameBased(resourceTypeCollection, c.link, false)


sdk/data/azcosmos/cosmos_container.go
```diff
@@ -138,9 +138,8 @@ func (c *ContainerClient) Delete(
 	}
 
 	operationContext := pipelineRequestOptions{
-		resourceType:     resourceTypeCollection,
-		resourceAddress:  c.link,
-		isWriteOperation: true,
+		resourceType:    resourceTypeCollection,
+		resourceAddress: c.link,
```
ealsur, line 142: A delete is a Write operation. And we have tests that use this. The Write operation is used to define routing (write operations can only be sent to the primary region)

Please verify with the service team the origin of the HTTP 500.

ealsur (changes requested):

sdk/data/azcosmos/cosmos_container.go
ealsur: isWriteOperation does not modify the request that is sent, it is used to know which region to route it to.

https://github.com/Azure/azure-sdk-for-go/blob/2532a53cfb2fabda9cbb1f04d209d95e84d18cb8/sdk/data/azcosmos/cosmos_location_cache.go#L116

kjg:
Closing this because https://github.com/Azure/azure-sdk-for-go/pull/23335 is the correct fix for my issue

PR #23335, opened 2024-08-15
Cosmos DB: Fixes minimal response header being sent on metadata operations

ealsur:
Minimal content header by design should be used on Item operations. On non-Item operations it should ideally be ignored by the service but there's evidence that if sent during a Delete Container operation with Shared Throughput, it causes an HTTP 500 on the service.

This PR makes sure that the header is only sent on Item operations

### DIFF
--- a/sdk/data/azcosmos/cosmos_headers_policy.go
+++ b/sdk/data/azcosmos/cosmos_headers_policy.go
@@ -43,7 +43,7 @@ func (p *headerPolicies) Do(req *policy.Request) (*http.Response, error) {
 			}
 		}
 
-		if o.isWriteOperation && !enableContentResponseOnWrite {
+		if o.isWriteOperation && o.resourceType == resourceTypeDocument && !enableContentResponseOnWrite {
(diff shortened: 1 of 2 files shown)


FabianMeiswinkel (approved):
LGTM

sdk/data/azcosmos/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed sending `Prefer` header on metadata operations.
```
kushagraThapar, line 11: Shall we call out the full header name to make it clear? `PreferReturnMinimal`?
Or may be we can call out the API name so that we can track this change later? like `contentResponseOnWrite`?

kushagraThapar (approved):
LGTM, thanks @ealsur

kirankumarkolli (approved):

sdk/data/azcosmos/CHANGELOG.md
ealsur: Its for all APIs. The header name is `Prefer`, but I added the value for clarity.

kushagraThapar (approved):
Thank you @ealsur

PR #23339, opened 2024-08-15
Cosmos DB: Fixes metadata routing on multi-master

ealsur:
## Scenario

Metadata read requests (not item operations) can be routed to any available regions. Metadata write requests, should always be routed to the primary / hub region.

In cases where the account is Single Write Region, this rule applies correctly.

The SDK allows users to define a list of Preferred Regions to route requests preferably. This preference applies in the following way:

* Single Write Region: Affects read operations
* Multi Write Region: Affects read and write operations

## Problem

The problem is that the above rules should apply for Item operations. Metadata operations, even on a Multi Write Region account, cannot be routed to secondary regions.

Currently when routing requests, the Global Endpoint Manager is treating all requests the same way and for Metadata write operations and Multi Write Region accounts, it will apply the user preference.

## Solution

Take into consideration the Resource Type when deciding if it's possible to route write operations to preferred regions and only do it for Item operations.

Similar code path in .NET

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/2fbc27835fde0837d4b3efea0f722c5f5c64119e/Microsoft.Azure.Cosmos/src/Routing/LocationCache.cs#L530-L535

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Bugs Fixed
 * Fixed sending `Prefer` header with `return=minimal` value on metadata operations. See [PR 23335](https://github.com/Azure/azure-sdk-for-go/pull/23335)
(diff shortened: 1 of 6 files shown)


kushagraThapar (approved):
LGTM, thanks @ealsur

PR #23341, opened 2024-08-15
azidentity accepts object IDs for user-assigned managed identities

chlowell:
Fixes https://github.com/Azure/azure-sdk-for-go-pr/issues/44


sdk/azidentity/managed_identity_client.go
```diff
@@ -389,10 +395,13 @@ func (c *managedIdentityClient) createAzureMLAuthRequest(ctx context.Context, id
 	q.Add("resource", strings.Join(scopes, " "))
 	q.Add("clientid", os.Getenv(defaultIdentityClientID))
 	if id != nil {
-		if id.idKind() == miResourceID {
-			return nil, newAuthenticationFailedError(credNameManagedIdentity, "Azure ML doesn't support specifying a managed identity by resource ID", nil, nil)
-		} else {
+		switch id.idKind() {
+		case miClientID:
 			q.Set("clientid", id.String())
```
jhendrixMSFT, line 400: We use `Set()` here but `Add()` in the other places. Thinking it should be `Set()` everywhere, or do we do this for a reason?

jhendrixMSFT: If it's too risky to change, it's fine as-is.

jhendrixMSFT (approved):

sdk/azidentity/managed_identity_client.go
chlowell: It shouldn't matter functionally because we're constructing new requests. They have no params until we start calling Add/Set. I agree it's more correct conceptually to Set in all cases; I'll make that change. I guess whoever wrote the Add calls made an arbitrary choice because there's no scenario for multiples of any of these params.

PR #23359, opened 2024-08-21
NFS share snapshot

tanyasethi-msft:
This feature only applies to premium file accounts. 
Reason to change BuildCanonicalizedHeaders method - 
The 2 new headers added as part of this feature - 
```
x-ms-enable-snapshot-virtual-directory-access and
x-ms-enabled-protocols
```
are quite similar. 
Our shared key string to sign algorithm ordered it differently than what backend service does. Here, enable`d` is compared to a `-` while character to character comparison. The service expects lower case characters to preceed symbols like - 
image

which was not the case before. So, this new custom sorting algorithm is written and added to all modules - azblob, azfile, and azdatalake since this problem can come to any module at a later point of time. (as of now, this is the only header which exposed the issue - and this is in azfile)

Custom sort implemented in c++ - 
https://github.com/Azure/azure-sdk-for-cpp/blob/40fa516d87527020a9a26ba0874b973b3c7d16ef/sdk/storage/azure-storage-common/src/shared_key_policy.cpp#L51

Optimised in Python to not use 3 tables (since capitalized letter comparison is not needed in request headers) - https://github.com/Azure/azure-sdk-for-python/blob/513989e008612fa053dfa8ec1318a6c4bcdbf991/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py#L55 

Other changes in this PR are from the feature NFS share snapshot.


sdk/storage/azfile/internal/exported/shared_key_credential.go
```diff
@@ -109,6 +109,81 @@ func getHeader(key string, headers map[string][]string) string {
...
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x8012, 0x0, 0x0, 0x0, 0x0, 0x0, 0x8212, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+		0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+	}
+	tables := [][]int{tableLv0[:], tableLv2[:]}
```
jhendrixMSFT, line 136: This (the above really) is going to allocate for every invocation of `compareHeaders`. Ideally, lines 114-136 would be split into its own `func` that returns a `[][]int` which is then called before `sort.Slice` and the result passed to `compareHeaders`.

jhendrixMSFT (approved):

souravgupta-msft:
Run the pipelines of azblob and azdatalake also

sdk/storage/azfile/share/client_test.go
```diff
@@ -279,6 +279,26 @@ func (s *ShareRecordedTestsSuite) TestShareCreateNilMetadata() {
...
+func (s *ShareRecordedTestsSuite) TestShareCreateWithSnapshotVirtualDirectoryAccess() {
+	_require := require.New(s.T())
+	testName := s.T().Name()
+	svcClient, err := testcommon.GetServiceClient(s.T(), testcommon.TestAccountPremium, nil)
+	_require.NoError(err)
+
+	shareName := testcommon.GenerateShareName(testName)
+	shareClient := svcClient.NewShareClient(shareName)
+
+	_, err = shareClient.Create(context.Background(), &share.CreateOptions{EnabledProtocols: to.Ptr("NFS"), EnableSnapshotVirtualDirectoryAccess: to.Ptr(false)})
```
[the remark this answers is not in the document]
tanyasethi-msft: Added

sdk/storage/azfile/internal/exported/shared_key_credential.go
tanyasethi-msft: Agreed, thanks for catching this. Refactored.

souravgupta-msft: Update this in azblob and azdatalake also

tanyasethi-msft: Updated.

souravgupta-msft (approved):

PR #23374, opened 2024-08-27
fix generator tool

Alancere:
When filtering NonExported, NewXxx functions are also filtered out.


tadelesh (approved):
offline synced. please merge after refining the logic to ignore empty instead of hard code filtering.

PR #23381, opened 2024-08-28
[azcore] Add text [un]marshalling support to ResourceID

mbarnes:
It would be handy for [ResourceID](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/internal/resource#ResourceID) to support the [encoding.TextMarshaler](https://pkg.go.dev/encoding@go1.23.0#TextMarshaler) and [encoding.TextUnmarshaler](https://pkg.go.dev/encoding@go1.23.0#TextUnmarshaler) interfaces so that, for example, JSON data that includes a resource ID string can be unmarshaled directly to a [ResourceID](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/internal/resource#ResourceID) struct (and vice versa for marshaling).

### DIFF
--- a/sdk/azcore/arm/internal/resource/resource_identifier.go
+++ b/sdk/azcore/arm/internal/resource/resource_identifier.go
@@ -110,6 +110,21 @@ func (id *ResourceID) String() string {
 	return id.stringValue
 }
 
+// MarshalText returns a textual representation of the ResourceID
+func (id *ResourceID) MarshalText() ([]byte, error) {
+	return []byte(id.String()), nil
+}
+
+// UnmarshalText decodes the textual representation of a ResourceID
+func (id *ResourceID) UnmarshalText(text []byte) error {
+	newId, err := ParseResourceID(string(text))
+	if err != nil {
+		return err
+	}
+	*id = *newId
+	return nil
+}
+
 func newResourceID(parent *ResourceID, resourceTypeName string, resourceName string) *ResourceID {
 	id := &ResourceID{}
 	id.init(parent, chooseResourceType(resourceTypeName, parent), resourceName, true)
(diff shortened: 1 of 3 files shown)


mbarnes:
My suggestion is still valid.  Please leave open.

sdk/azcore/arm/internal/resource/resource_identifier.go
```diff
@@ -110,6 +110,21 @@ func (id *ResourceID) String() string {
 	return id.stringValue
 }
 
+// MarshalText returns a textual representation of the ResourceID
+func (id *ResourceID) MarshalText() ([]byte, error) {
```
jhendrixMSFT, line 114: Usually the receiver is by value for `Marshal*` methods. Is there a reason why it's a pointer here?

jhendrixMSFT:
Mind rebasing on `main`?

sdk/azcore/arm/internal/resource/resource_identifier.go
mbarnes: Mainly because `ParseResourceID` allocates a new `ResourceID`, and I was using [url.URL](https://pkg.go.dev/net/url#URL) as a guide.  Its [MarshalBinary](https://pkg.go.dev/net/url#URL.MarshalBinary) and [UnmarshalBinary](https://pkg.go.dev/net/url#URL.UnmarshalBinary) methods both take pointers.

It's less important for `MarshalText` but I think its the only way to get `UnmarshalText` to work without doing a field-by-field assignment.

jhendrixMSFT (approved):

venkateshsredhat (commented):
@mbarnes As i understand we added the wrapper to implement MarshalText and Unmarshall methods but since you have done the contribution upstream we dont need it anymore .
That makes sense , but i see we never used MarshalText or UnmarshallText , is that correct or we removed it sometime back?

PR #23400, opened 2024-09-04
[azservicebus] Fixing a bug where our management link would hang in certain circumstances

richardpark-msft:
Fixing a bug where we could end up getting our management link stuck.
    
The problem was basically a combination of errors:
1. We were missing the sender settlement mode, which made it so we were only 1/2 in the "receive and delete" mode that we intended to be.
2. Ignoring that, our AcceptMessage() call was actually load bearing and critical and _wasn't_ getting called in some cases, like if you cancelled the call.

### DIFF
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.2.3 (Unreleased)
+
+### Bugs Fixed
+
+- Fixed a bug where cancelling RenewMessageLock() calls could cause hangs in future RenewMessageLock calls. (PR#TBD)
+
 ## 1.2.2 (2024-08-15)
 
 ### Bugs Fixed
--- a/sdk/messaging/azeventhubs/CHANGELOG.md
+++ b/sdk/messaging/azeventhubs/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bugs Fixed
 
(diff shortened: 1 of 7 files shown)


richardpark-msft:
CC: @karenychen

sdk/messaging/azeventhubs/internal/rpc.go
```diff
@@ -125,6 +125,10 @@ func NewRPCLink(ctx context.Context, args RPCLinkArgs) (amqpwrap.RPCLink, error)
 	receiverOpts := &amqp.ReceiverOptions{
 		TargetAddress: link.clientAddress,
 		Credit:        defaultReceiverCredits,
+
+		// set our receiver link into the "receive and delete" mode - messages arrive pre-settled.
+		SettlementMode:            amqp.ReceiverSettleModeFirst.Ptr(),
+		RequestedSenderSettleMode: amqp.SenderSettleModeSettled.Ptr(),
```
jhendrixMSFT, line 131: Why do we want pre-settled messages for RPC?

jhendrixMSFT (approved):

sdk/messaging/azeventhubs/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.2.3 (Unreleased)
+
+### Bugs Fixed
+
+- Fixed a bug where cancelling RenewMessageLock() calls could cause hangs in future RenewMessageLock calls. (PR#TBD)
```
conniey, line 7: PR#TBD?

PR #23414, opened 2024-09-09
BearerTokenPolicy handles CAE claims challenges by default

chlowell:
This adds CAE support for all clients lacking a custom challenge handler i.e., everyone except Key Vault and Storage. The policy will request CAE tokens by default, however whether it receives them and claims challenges is up to RP and tenant configuration. With this change, CAE will just work by default when enabled by RP and tenant. To avoid opting clients like Key Vault and Storage in to receiving challenges they can't handle, I added a `SupportsCAE` option to `AuthorizationHandler` to control whether the policy requests CAE tokens when given a custom challenge handler.


jhendrixMSFT (approved):
Was there a specific trigger prompting this to be moved to the common bearer token policy, or was it just a matter of "it's baked enough"?

chlowell:
It's baked enough, there's a common implementation for RPs we can target. I see they now document how to handle multiple challenges, so let me update our logic to follow that guidance.

jhendrixMSFT (approved):

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -115,9 +124,80 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+							cae = true
+						}
+					}
+				}
+				if cae && claims != "" {
+					// remove the trailing quote and any base64 padding
+					if end := strings.IndexAny(claims, `="`); end > 0 {
+						claims = claims[:end]
+					}
+					if b, de := base64.RawURLEncoding.DecodeString(claims); de == nil {
```
christothes, line 194: Are we expected to decode the claims rather than just pass them along as is?

chlowell: That depends on the contract with MSAL. MSAL Go expects the decoded claims i.e. JSON.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -115,9 +124,80 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+					case "claims":
+						claims = v
+					case "error":
+						// v has an irrelevant trailing quote
+						if strings.HasPrefix(v, "insufficient_claims") {
+							cae = true
+						}
+					}
+				}
+				if cae && claims != "" {
```
christothes, line 189: Will claims ever be populated when the error value is something other than "insufficient_claims"? I'm wondering if it is necessary to gate parsing claims on this value or error.

chlowell: `claims` semantics aren't specified in e.g. an RFC, so it's up to the RP and clients need a way to distinguish the CAE case. I believe this is why [the docs](https://learn.microsoft.com/entra/identity-platform/claims-challenge?tabs=dotnet#claims-challenge-header-format) specify CAE claims challenges always have `error="insufficient_claims"` and recommend clients check for that ([here](https://learn.microsoft.com/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=dotnet#the-code)).

chlowell:
Moving this back to draft because I believe the core policy can handle CAE challenges even when the client has provided a custom challenge handler, at the cost of some more complexity. That could enable CAE support for e.g. Key Vault without any change to client code.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -121,3 +158,70 @@ func checkHTTPSForAuth(req *policy.Request, allowHTTP bool) error {
...
+		if c.scheme == "Bearer" {
+			if claims := c.params["claims"]; claims != "" && c.params["error"] == "insufficient_claims" {
+				b, de := base64.StdEncoding.DecodeString(claims)
+				if de != nil {
+					// don't include the decoding error because it's something
+					// unhelpful like "illegal base64 data at input byte 42"
+					err = errorinfo.NonRetriableError(errors.New("authentication challenge contains invalid claims: " + claims))
+					break
+				}
+				c.params["claims"] = string(b)
```
jhendrixMSFT, line 179: Since we're ranging over `parseChallenges` it seems like we could hit this case multiple times (in theory anyways), so the last one wins. Is the ok?

chlowell: Yes, "one response having multiple Bearer challenges" isn't a real scenario. There could in principle be another challenge, for a different scheme, so the loop could continue after finding a Bearer CAE challenge. However, this function wouldn't do anything with those challenges. I suppose it may as well `break` after finding the CAE challenge though.

jhendrixMSFT: I like the idea of adding a break as it makes it clear that we expect only one challenge and accept the first one we find.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -103,9 +112,37 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+				tro := policy.TokenRequestOptions{
+					Claims: caeChallenge.params["claims"],
+					Scopes: b.scopes,
+				}
+				if err = b.authenticateAndAuthorize(req)(tro); err == nil {
+					res, err = req.Next()
+				}
+			case b.authzHandler.OnChallenge != nil:
+				if err = b.authzHandler.OnChallenge(req, res, b.authenticateAndAuthorize(req)); err == nil {
+					if res, err = req.Next(); err == nil && res.StatusCode == http.StatusUnauthorized {
```
jhendrixMSFT, line 131: This looks to be duplicating earlier code. Can it be simplified to remove the duplication?

chlowell: I can reduce duplication but I'm uncertain the result will be simpler. At first glance I think it needs recursion. I'll give it a shot

chlowell: Well, it turned out recursive but not terrible. What do you think?

jhendrixMSFT (approved):

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -101,17 +110,44 @@ func (b *BearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
...
+				return res, parseErr
+			}
+			switch {
+			case caeChallenge != nil:
+				tro := policy.TokenRequestOptions{
+					Claims: caeChallenge.params["claims"],
+					Scopes: b.scopes,
+				}
+				if err = b.authenticateAndAuthorize(req)(tro); err == nil {
+					res, err = req.Next()
```
gracewilcox, line 135: Does the error returned here need to be marked as a non retriable error?

jhendrixMSFT: While I'll defer to @chlowell I think the answer is no. The reason being any policies after this one are responsible for returning non-retriable errors as such. We don't have enough context here to know if the error should be considered non-retriable.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -121,3 +157,70 @@ func checkHTTPSForAuth(req *policy.Request, allowHTTP bool) error {
...
+		// captures parameter names and values in a match of the above expression
+		challengeParams = regexp.MustCompile(`(\w+)="([^"]*)"`)
+	})
+	parsed := []authChallenge{}
+	// WWW-Authenticate can have multiple values, each containing multiple challenges
+	for _, h := range res.Header.Values(shared.HeaderWWWAuthenticate) {
+		for _, sm := range challenge.FindAllStringSubmatch(h, -1) {
+			// sm is [challenge, scheme, params]
+			// len checks aren't necessary but save you from wondering whether this function could panic
+			if len(sm) == 3 {
```
jhendrixMSFT, line 210: If the `len` check isn't necessary, then I'd prefer to remove it (but keep a comment explaining why). Else it looks like there should be an `else` clause returning an error if the `len` is not what we expect.

sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -121,3 +157,70 @@ func checkHTTPSForAuth(req *policy.Request, allowHTTP bool) error {
...
+	var (
+		caeChallenge *authChallenge
+		err          error
+	)
+	for _, c := range parseChallenges(res) {
+		if c.scheme == "Bearer" {
+			if claims := c.params["claims"]; claims != "" && c.params["error"] == "insufficient_claims" {
+				if b, de := base64.StdEncoding.DecodeString(claims); de == nil {
+					c.params["claims"] = string(b)
+					caeChallenge = &c
```
gracewilcox, line 173: The arm/runtime bearer token policy that's being removed checks for multiple claims in the header and throws an error if multiple claims are found. Do we need to handle that situation?

chlowell: No, the relevant RFC forbids a challenge to repeat parameters

sdk/azcore/runtime/policy_bearer_token.go
chlowell: That's the idea. This policy is responsible for making any error it originates--including those it propagates from OnChallenge--non-retriable but can't determine whether it would be correct to make another policy's error non-retriable.

sdk/azcore/CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
 # Release History
 
-## 1.14.1 (Unreleased)
+## 1.15.0 (2024-10-04)
```
jhendrixMSFT, line 3: We should move this to `10-7` as releasing on Friday seems risky.

sdk/azcore/runtime/policy_bearer_token.go
jhendrixMSFT: Looks good!

gracewilcox (approved):

jhendrixMSFT (approved):

PR #23441, opened 2024-09-16
fix: integer overflow on calcDelay

bcho:
The current `calcDelay` overflows when `try` >= 36 when using default retry options settings (800ms). Example: https://go.dev/play/p/-w0DcQsN1cD . This pull request fixed by adding bound checks on the values.

The new implementation does not increase the overhead of the function per the benchmark results:

```
# on commit: babb98a69995529ea3d2fd45995ef48a3a410849
$ go test -v -bench=. -run=BenchmarkCalcDelay -benchtime=20s
goos: linux
goarch: amd64
pkg: github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime
cpu: AMD EPYC 7763 64-Core Processor                
BenchmarkCalcDelay_defaultSettings
BenchmarkCalcDelay_defaultSettings-2    1000000000              17.03 ns/op
BenchmarkCalcDelay_overflow
BenchmarkCalcDelay_overflow-2           1000000000              16.64 ns/op
PASS
ok      github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime    37.016s

# on commit: d7effab1ad0c72aae14df58922f48d0a360bce02
$ go test -v -bench=. -run=BenchmarkCalcDelay -benchtime=20s
goos: linux
goarch: amd64
pkg: github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime
cpu: AMD EPYC 7763 64-Core Processor                
BenchmarkCalcDelay_defaultSettings
BenchmarkCalcDelay_defaultSettings-2    1000000000              15.53 ns/op
BenchmarkCalcDelay_overflow
BenchmarkCalcDelay_overflow-2           1000000000              15.91 ns/op
PASS
ok      github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime    34.586s
```

### DIFF
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -59,13 +59,36 @@ func setDefaults(o *policy.RetryOptions) {
 }
 
 func calcDelay(o policy.RetryOptions, try int32) time.Duration { // try is >=1; never 0
-	delay := time.Duration((1<<try)-1) * o.RetryDelay
+	const maxDelay = math.MaxInt64
 
-	// Introduce some jitter:  [0.0, 1.0) / 2 = [0.0, 0.5) + 0.8 = [0.8, 1.3)
-	delay = time.Duration(delay.Seconds() * (rand.Float64()/2 + 0.8) * float64(time.Second)) // NOTE: We want math/rand; not crypto/rand
+	var factor int64
+	if try >= 63 {
+		factor = maxDelay
+	} else {
+		factor = int64(1)<<try - 1
+	}
+
+	var delay time.Duration
+	if factor > maxDelay/int64(o.RetryDelay) {
+		delay = maxDelay
+	} else {
+		delay = time.Duration(factor) * o.RetryDelay
+	}
+
+	// Introduce jitter:  [0.0, 1.0) / 2 = [0.0, 0.5) + 0.8 = [0.8, 1.3)
+	jitterMultiplier := rand.Float64()/2 + 0.8 // NOTE: We want math/rand; not crypto/rand
+
+	delayFloat := float64(delay) * jitterMultiplier
+	if delayFloat > float64(maxDelay) {
+		delay = maxDelay
+	} else {
+		delay = time.Duration(delayFloat)
+	}
+
 	if delay > o.MaxRetryDelay {
 		delay = o.MaxRetryDelay
 	}
+
 	return delay
 }
 
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -65,7 +65,7 @@ func calcDelay(o policy.RetryOptions, try int32) time.Duration { // try is >=1; 
 		factor = time.Duration(int64(1)<<try - 1)
 	}
 
-	delay := time.Duration(factor) * o.RetryDelay
+	delay := factor * o.RetryDelay
 	if delay < factor {
 		// overflow has happend so set to max value
 		delay = time.Duration(math.MaxInt64)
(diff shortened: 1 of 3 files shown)


sdk/azcore/runtime/policy_retry.go
```diff
@@ -59,13 +59,33 @@ func setDefaults(o *policy.RetryOptions) {
 }
 
 func calcDelay(o policy.RetryOptions, try int32) time.Duration { // try is >=1; never 0
-	delay := time.Duration((1<<try)-1) * o.RetryDelay
+	// avoid overflow when shifting left
+	factor := time.Duration(math.MaxInt64)
+	if try < 63 {
+		factor = time.Duration(int64(1)<<try - 1)
```
chlowell, line 65: If we don't have a linting rule about redundant parentheses, I'd do this just so people don't have to check the order of evaluation like I did 😁 
```suggestion
		factor = time.Duration(int64(1<<try) - 1)
```

sdk/azcore/runtime/policy_retry.go
```diff
@@ -59,13 +59,33 @@ func setDefaults(o *policy.RetryOptions) {
...
+	}
+
+	delay := time.Duration(factor) * o.RetryDelay
+	if delay < factor {
+		// overflow has happend so set to max value
```
chlowell, line 70:
```suggestion
		// overflow has happened so set to max value
```

chlowell (approved):

bcho:
@chlowell / @jhendrixMSFT thanks for the reviews and approval. BTW, I saw there are few other retry implementations in the repo affected by this corner case. Ex: https://github.com/Azure/azure-sdk-for-go/blob/2312973b416942386b01143bb44ac1c7bdb537e7/sdk/messaging/azeventhubs/internal/utils/retrier.go#L117-L138 Do you want to update the implementations to use the same code like here?

jhendrixMSFT:
For the other cases can you please open issues? We'll let their owners clean them up.

jhendrixMSFT (approved):

sdk/azcore/runtime/policy_retry.go
```diff
@@ -59,13 +59,33 @@ func setDefaults(o *policy.RetryOptions) {
...
+	if delayFloat > float64(math.MaxInt64) {
+		// the jitter pushed us over MaxInt64, so just use MaxInt64
+		delay = time.Duration(math.MaxInt64)
+	} else {
+		delay = time.Duration(delayFloat)
+	}
 
-	// Introduce some jitter:  [0.0, 1.0) / 2 = [0.0, 0.5) + 0.8 = [0.8, 1.3)
-	delay = time.Duration(delay.Seconds() * (rand.Float64()/2 + 0.8) * float64(time.Second)) // NOTE: We want math/rand; not crypto/rand
 	if delay > o.MaxRetryDelay {
```
richardpark-msft, line 85: Apparently we've had a bug in here for awhile:
```suggestion
	if o.MaxRetryDelay > 0 && delay > o.MaxRetryDelay {
```

richardpark-msft (commented):
Looks good, caught a small bug in the previous code while porting this over.

sdk/azcore/runtime/policy_retry.go
bcho: I was expecting the MaxRetryDelay being back filled at here:

https://github.com/Azure/azure-sdk-for-go/blob/1fbf735512c6be7673c97c9f2843a27aede2dc42/sdk/azcore/runtime/policy_retry.go#L37-L42

jhendrixMSFT: Correct, `o.MaxRetryDelay` will never be zero or negative here (I had the same thought then realized this after making the same fix).

jhendrixMSFT: We should add a comment here explaining this.

bcho, line 85:
```suggestion
	if delay > o.MaxRetryDelay { // MaxRetryDelay is backfilled with non-negative value
```

bcho: I updated with a comment, PTAL

jhendrixMSFT:
Rebased on `main` to resolve merge conflict.

PR #23462, opened 2024-09-18
[azopenaiextensions] Initial package creation

richardpark-msft:


sdk/ai/azopenaiextensions/go.mod
```diff
@@ -0,0 +1,35 @@
+module github.com/Azure/azure-sdk-for-go/sdk/ai/azopenaiextensions
+
+go 1.21
```
jhendrixMSFT, line 3: This is newer than most baselines, is that expected?

richardpark-msft: Expected - our package is at the whims of the openi-go package:

https://github.com/openai/openai-go/blob/f9d9ecc0b31854d5bf2bd091cd75c3b0fd67bdf0/go.mod#L3

jhendrixMSFT (approved):

PR #23475, opened 2024-09-20
[azadmin] change frequency of poller in tests

gracewilcox:
Modifying the poller so that it polls every second instead of every 30 seconds in playback mode. Now in playback mode it'll take the tests 2 seconds to complete instead of 67.

### DIFF
--- a/sdk/security/keyvault/azadmin/backup/client_test.go
+++ b/sdk/security/keyvault/azadmin/backup/client_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
 	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azadmin/backup"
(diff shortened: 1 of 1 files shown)


sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -21,13 +22,23 @@ import (
...
 	testSerde(t, &sasToken)
 
+	// if in live mode, wait 30 seconds before polling again
+	// in playback mode, we don't need that delay, so shortening the retry frequency
+	var frequency time.Duration
+	if recording.GetRecordMode() != recording.PlaybackMode {
+		frequency = time.Second * 30
+	} else {
+		frequency = time.Second
+	}
```
chlowell, line 34: Can we simply pass a zero `Duration` when not in playback and thereby accept the default?
```suggestion
	var frequency time.Duration
	if recording.GetRecordMode() == recording.PlaybackMode {
		frequency = time.Second
	}
```

gracewilcox: Good idea, thanks!

sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -21,13 +22,21 @@ import (
 
 func TestBackupRestore(t *testing.T) {
 	client, sasToken := startBackupTest(t)
-
 	testSerde(t, &sasToken)
 
+	// if in live mode, wait 30 seconds before polling again
```
chlowell, line 27: Personally I think this code is too clear for a comment, but if you want something here, replace the "30 seconds" stuff with something about "default frequency"

chlowell (approved):

sdk/security/keyvault/azadmin/backup/client_test.go
gracewilcox: comment removed

PR #23485, opened 2024-09-23
Improve AzurePipelinesCredential diagnosability

### LINKED ISSUE #23444: Allow certain response headers coming from AzurePipelinesCredential to be logged
These two response headers are useful for the service team to debug and diagnose issues, and hence it is requested that we do log them.
- x-vss-e2eid
- x-msedge-ref

We decided, other than logging them, to also add these header values to the exception message that is thrown, so they are always visible, even with logging disabled.

Given our logged headers are based on an allow list, currently these two headers are REDACTED.

https://github.com/Azure/azure-sdk-for-go/blob/500650cf709907ff0dc0a11180547467407f1a06/sdk/azcore/runtime/policy_logging.go#L37-L63

### LINKED ISSUE #23482: Update the request header sent to the OIDC endpoint so it doesn't result in a redirect response when an invalid system access token is provided.
See https://github.com/Azure/azure-sdk-for-cpp/pull/6019 for reference

- Add X-TFS-FedAuthRedirect: Suppress in the request header to the request sent to the OIDC endpoint
- Update TSG and test

chlowell:
* enable logging for a couple DevOps response headers the service team uses when debugging and include them in errors. Closes #23444
* instruct OIDC endpoint to return 401 instead of 302 when given an invalid access token. Closes #23482

### DIFF
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -9,6 +9,8 @@
 ### Bugs Fixed
 
 ### Other Changes
+* `AzurePipelinesCredential` sets an additional OIDC request header so that it
+  receives a 401 instead of a 302 after presenting an invalid system access token
 
 ## 1.8.0-beta.3 (2024-09-17)
 
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -11,6 +11,8 @@
 ### Other Changes
 * `AzurePipelinesCredential` sets an additional OIDC request header so that it
   receives a 401 instead of a 302 after presenting an invalid system access token
+* Allow logging of debugging headers for `AzurePipelinesCredential` and include
+  them in error messages
 
 ## 1.8.0-beta.3 (2024-09-17)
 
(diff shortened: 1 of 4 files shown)


jhendrixMSFT (approved):

RickWinter (approved):

sdk/azidentity/azure_pipelines_credential.go
```diff
@@ -86,6 +88,8 @@ func NewAzurePipelinesCredential(tenantID, clientID, serviceConnectionID, system
 	if options == nil {
 		options = &AzurePipelinesCredentialOptions{}
 	}
+	// these headers are useful to the DevOps team when debugging OIDC error responses
+	options.ClientOptions.Logging.AllowedHeaders = append(options.ClientOptions.Logging.AllowedHeaders, xMsEdgeRef, xVssE2eId)
```
ahsonkhan, line 92: Do these options apply to the request that's sent to the OIDC endpoint?

It looks like they are being passed into `ClientAssertionCredentialOptions` (only), which means the second request will get the options, but if you have a separate pipeline and client for the OIDC request, it may not get the options (`AzurePipelinesCredential`).

sdk/azidentity/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 ### Bugs Fixed
 
 ### Other Changes
+* `AzurePipelinesCredential` sets an additional OIDC request header so that it
```
ahsonkhan, line 12: Consistency nit: Other languages place this particular change as a bug fix.

sdk/azidentity/azure_pipelines_credential.go
chlowell: Yes (a test covers this). OIDC and Entra requests transit the same pipeline

PR #23489, opened 2024-09-24
Give CODEOWNERS file an owner in CODEOWNERS

JimSuplizio:
Specifically give the CODEOWNERS file an owner. Also, moved the common owners closer to the bottom of the file since GitHub matches bottom up.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -89,10 +89,6 @@
 # PRLabel: %OpenTelemetry
 /sdk/tracing/azotel           @jhendrixMSFT @rickwinter
 
-# Add owners for go sdk tool
-/eng/tools/generator          @ArcturusZhang @lirenhe @tadelesh @raych1
-/eng/tools/internal           @ArcturusZhang @lirenhe @tadelesh @raych1
-
 # AzureSDKOwners: @benbp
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -89,18 +89,6 @@
...
-/.github/workflows/           @Azure/azure-sdk-eng
-/eng/config.json              @richardpark-msft @jhendrixMSFT @rickwinter @chlowell @gracewilcox
-/.config/1espt/               @benbp @weshaggard
-
-# Add owners for notifications for specific pipelines
-/eng/common/pipelines/codeowners-linter.yml       @rickwinter
-
 # Add owners for go sdk tool
 /eng/tools/generator          @ArcturusZhang @lirenhe @tadelesh @raych1
 /eng/tools/internal           @ArcturusZhang @lirenhe @tadelesh @raych1
```
RickWinter, line 94: Please keep these in the same area as the other eng/tools/.... entries. Makes it easier to find.

RickWinter (approved):

.github/CODEOWNERS
JimSuplizio: I'll move them down to just below the Eng Sys block.

Issue #23454, opened 2024-09-18
azcore.ResponseError fails to marshal


KStillings-Slalom:
### Bug Report

When trying to marshal the azcore.ResponseError into a data object, an unsupportedType occurs.
If we null out the azcore.ResponseError.RawResponse.Request object, the error marshals out as expected.

It appears this might be related to [21838](https://github.com/Azure/azure-sdk-for-go/issues/21838) and [22147](https://github.com/Azure/azure-sdk-for-go/issues/22147)

It would be nice to be able to marshal this out as well so we can return a full error context back to the client.

jhendrixMSFT:
The safest thing for us to do is omit the `RawResponse` field from marshaling. It's lossy though, so calling `Error()` on the unmarshaled instance won't contain full info. This shouldn't be a problem as that output is not contractual (only `ErrorCode` and `StatusCode` are), but I'd like to know a bit more about your scenario.

jhendrixMSFT:
Fix will be in upcoming `v1.14.1` release.

PR #23532, opened 2024-10-01
Omit RawResponse from JSON payload

jhendrixMSFT:
The *http.Request inside the *http.Response contains fields that aren't marshalable (func). Since the RawResponse is used to construct the error message which isn't contractual, just omit it from the payload.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/23454

### DIFF
--- a/sdk/azcore/internal/exported/response_error.go
+++ b/sdk/azcore/internal/exported/response_error.go
@@ -181,15 +181,10 @@ type responseError struct {
 }
 
 func (e ResponseError) MarshalJSON() ([]byte, error) {
-	// populate errMsg as needed
-	if e.errMsg == "" {
-		_ = e.Error()
-	}
-
 	return json.Marshal(responseError{
 		ErrorCode:    e.ErrorCode,
 		StatusCode:   e.StatusCode,
-		ErrorMessage: e.errMsg,
+		ErrorMessage: e.Error(),
 	})
 }
 
--- a/sdk/azcore/internal/exported/response_error.go
+++ b/sdk/azcore/internal/exported/response_error.go
@@ -118,11 +118,17 @@ type ResponseError struct {
 
 	// RawResponse is the underlying HTTP response.
 	RawResponse *http.Response `json:"-"`
+
+	errMsg string
 }
 
 // Error implements the error interface for type ResponseError.
 // Note that the message contents are not contractual and can change over time.
 func (e *ResponseError) Error() string {
+	if e.errMsg != "" {
+		return e.errMsg
+	}
+
 	const separator = "--------------------------------------------------------------------------------"
 	// write the request method and URL with response status code
 	msg := &bytes.Buffer{}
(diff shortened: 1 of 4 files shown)


sdk/azcore/internal/exported/response_error.go
```diff
@@ -117,7 +117,7 @@ type ResponseError struct {
 	StatusCode int
 
 	// RawResponse is the underlying HTTP response.
-	RawResponse *http.Response
+	RawResponse *http.Response `json:"-"`
```
RickWinter, line 120: Doesn't this hinder debugability?

Please add more to comment to explain this change and link back to the issue. In a year the context for why we do this will be vague.

jhendrixMSFT: Say I call some API and it fails, returning an `*azcore.ResponseError`. I then want to marshal that error so I can send it somewhere else.
```go
bytes, err := json.Marshal(someRespErr)
```
This will fail due to fields in `RawResponse.Request` that cannot be marshaled (func types).
This change tells the JSON marshaler (and unmarshaler) to skip the `RawResponse` field.

jhendrixMSFT: Oh, if you mean debuggability on the client side then the answer is yes as the error details will be lost. That said, we've never considered the contents returned by `Error()` to be contractual. Whether or not this falls into that bucket is debatable.

We can probably preserve the error message, but it will require computing the message, stashing it in an unexported field, and adding custom JSON un/marshaler methods that know how to handle it. We'd still need to omit the `RawResponse` though.

jhendrixMSFT: I've updated the marshaler to preserve the contents of the error message and added a doc comment describing the behavior.

sdk/azcore/internal/exported/response_error.go
```diff
@@ -163,5 +169,38 @@ func (e *ResponseError) Error() string {
...
+func (e ResponseError) MarshalJSON() ([]byte, error) {
+	// populate errMsg as needed
+	if e.errMsg == "" {
+		_ = e.Error()
+	}
+
+	return json.Marshal(responseError{
+		ErrorCode:    e.ErrorCode,
+		StatusCode:   e.StatusCode,
+		ErrorMessage: e.errMsg,
```
chlowell, line 192: I like `Error()` computing the message only once but just for the sake of clarity in my mind, do you really need to cache it on `ResponseError`? Couldn't you do
```suggestion
		ErrorMessage: e.Error(),
```

chlowell (approved):

sdk/azcore/internal/exported/response_error.go
jhendrixMSFT: The caching lets us preserve the output of `Error()` after the type has been unmarshaled.

jhendrixMSFT: Your are correct though that this can be simplified, I've made that change.

PR #23542, opened 2024-10-03
fix(pollers/util.go): Handle missing JSON path gracefully by using entire payload

yyoshiki41:
## What

`ResultHelper` attempts to extract the value at the specified JSON path.
If the specified JSON path is not found in the response, the entire payload is used as the result.

## Why

In #22980, JSON path extraction was introduced.
However, in Azure Document Intelligence that adopted Operation-Location header strategy, it is not included under the key payloadPath: "result". As a result, the correct output cannot be retrieved (always nil).

This fix updates the logic to use the entire response if the specified JSON path key does not exist.

## Ref
- [Azure Document Intelligence quickstart guide](https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/quickstarts/get-started-sdks-rest-api?view=doc-intel-3.0.0&preserve-view=true&pivots=programming-language-rest-api#examine-the-response).

---

### DIFF
--- a/sdk/azcore/internal/pollers/util.go
+++ b/sdk/azcore/internal/pollers/util.go
@@ -177,15 +177,17 @@ func ResultHelper[T any](resp *http.Response, failed bool, jsonPath string, out 
 		return err
 	}
 
+	// try to extract the payload from the response if JSON path is provided
 	if jsonPath != "" && len(payload) > 0 {
-		// extract the payload from the specified JSON path.
-		// do this before the zero-length check in case there
-		// is no payload.
 		jsonBody := map[string]json.RawMessage{}
 		if err = json.Unmarshal(payload, &jsonBody); err != nil {
 			return err
 		}
-		payload = jsonBody[jsonPath]
+		// if the JSON path is found, use that as the payload
+		// otherwise, use the entire payload
+		if v, ok := jsonBody[jsonPath]; ok {
+			payload = v
+		}
 	}
 
 	if len(payload) == 0 {
--- a/sdk/azcore/internal/pollers/util_test.go
+++ b/sdk/azcore/internal/pollers/util_test.go
@@ -159,10 +159,12 @@ func TestResultHelper(t *testing.T) {
 		Body:       http.NoBody,
 	}
 	var result string
+
+	// no content
 	err := ResultHelper(resp, false, "", &result)
 	require.NoError(t, err)
 	require.Empty(t, result)
-
+	// bad request
 	resp.StatusCode = http.StatusBadRequest
 	resp.Body = io.NopCloser(strings.NewReader(`{ "code": "failed", "message": "bad stuff" }`))
 	err = ResultHelper(resp, false, "", &result)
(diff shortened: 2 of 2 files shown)


yyoshiki41:

sdk/azcore/internal/pollers/util.go
```diff
@@ -177,15 +177,17 @@ func ResultHelper[T any](resp *http.Response, failed bool, jsonPath string, out
...
-		// do this before the zero-length check in case there
-		// is no payload.
 		jsonBody := map[string]json.RawMessage{}
 		if err = json.Unmarshal(payload, &jsonBody); err != nil {
 			return err
 		}
-		payload = jsonBody[jsonPath]
+		// if the JSON path is found, use that as the payload
+		// otherwise, use the entire payload
+		if v, ok := jsonBody[jsonPath]; ok {
```
jhendrixMSFT, line 188: Can we please get more info about what motivated this change? i.e. does this fix an issue you observed?

jhendrixMSFT: Sorry just saw the comment, reading it now.

jhendrixMSFT:
@JeffreyRichter shouldn't the payload for the terminal success case _always_ be under the `result` property in the resultant JSON? I thought that was the contract.

yyoshiki41:
## Reproduction Steps

After upgrading to github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.12.0, the result is always nil when executing the following code.

```go
{
	endpoint := "endpoint"
	req, err := runtime.NewRequest(ctx, http.MethodPost, endpoint)
	if err != nil {
		return nil, err
	}
	if err := req.SetBody(&readSeekCloser{bytes.NewReader(body)}, "application/json"); err != nil {
		return nil, err
	}
	// Start pipeline
	pipeline := c.azure.Pipeline()
	resp, err := pipeline.Do(req)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()

	if resp.StatusCode >= 400 {
		return nil, runtime.NewResponseError(resp)
	}

	// Poll until the operation is done
	options := &runtime.NewPollerOptions[AnalyzeResponse]{}
	poller, err := runtime.NewPoller(resp, pipeline, options)
	if err != nil {
		return nil, err
	}
	result, err := poller.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{})
}

// AnalyzeResponse represents a response from the Analyze operation
type AnalyzeResponse struct {
	Status              string        `json:"status"`
	CreatedDateTime     time.Time     `json:"createdDateTime"`
	LastUpdatedDateTime time.Time     `json:"lastUpdatedDateTime"`
	AnalyzeResult       AnalyzeResult `json:"analyzeResult"`
}
```

JeffreyRichter:
The guidelines are here: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#long-running-operations--jobs

We have 4 LRO patterns: PUT, DELETE, LRO Action on resource & LRO action not on a resource. Just the LRO action patterns have a result property and yes, this property should only be set/read when the status is "success".

yyoshiki41:
@JeffreyRichter @jhendrixMSFT 

> Just the LRO action patterns have a result property and yes, this property should only be set/read when the status is "success".

I understand that only the LRO action patterns have a `result` property, which should only be set or read when the status is “success.”
Does this mean that Azure Document Intelligence doesn’t fall under these LRO action patterns, and therefore, the result property wouldn’t apply in this case? If so, should we avoid using `runtime.NewPoller` in Document Indelligence?

```go
	poller, err := runtime.NewPoller(resp, pipeline, options)
```

Thanks.

yyoshiki41:
@JeffreyRichter @jhendrixMSFT

I wanted to kindly follow up on this pull request as it’s been pending for about a month now.
Could you please let me know if there’s anything further I can address to help move this forward?
I’d be happy to make any adjustments as needed to expedite the review process.

Thanks!

jhendrixMSFT:
Sorry for the delay, I'm looking into this some more.

jhendrixMSFT:
We had an internal discussion, and the conclusion is that the result path must be configurable instead of hard coded to "result". I've opened https://github.com/Azure/azure-sdk-for-go/pull/23733 to address this which should also fix your scenario.

yyoshiki41:
@jhendrixMSFT 

Thanks!

jhendrixMSFT:
Closing in favor of https://github.com/Azure/azure-sdk-for-go/pull/23733

PR #23543, opened 2024-10-03
[keyvault] add cae support

gracewilcox:
Closes https://github.com/Azure/azure-sdk-for-go-pr/issues/43

Adding CAE support to key vault libraries:
* Upgrade to the latest version of azcore with CAE support
* Add tests to the key vault challenge policy to ensure that it's compatible with CAE

### DIFF
--- a/sdk/security/keyvault/internal/challenge_policy_test.go
+++ b/sdk/security/keyvault/internal/challenge_policy_test.go
@@ -22,6 +22,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	challengedToken = "needs more claims"
+	claimsToken     = "all the claims"
+	kvChallenge     = `Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://vault.azure.net"`
+	caeChallenge1   = `Bearer realm="", authorization_uri="https://login.microsoftonline.com/common/oauth2/authorize", error="insufficient_claims", claims="dGVzdGluZzE="`
+	caeChallenge2   = `Bearer realm="", authorization_uri="https://login.microsoftonline.com/common/oauth2/authorize", error="insufficient_claims", claims="dGVzdGluZzI="`
+	// requireToken is a mock.Response predicate that checks a request for the expected token
+	requireToken = func(t *testing.T, want string) func(req *http.Request) bool {
+		return func(r *http.Request) bool {
+			_, actual, _ := strings.Cut(r.Header.Get("Authorization"), " ")
+			require.Equal(t, want, actual)
+			return true
+		}
+	}
+)
+
 type credentialFunc func(context.Context, policy.TokenRequestOptions) (azcore.AccessToken, error)
 
 func (cf credentialFunc) GetToken(ctx context.Context, options policy.TokenRequestOptions) (azcore.AccessToken, error) {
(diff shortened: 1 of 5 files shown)


sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -100,6 +100,150 @@ func TestChallengePolicy(t *testing.T) {
...
+}
+
+func TestChallengePolicy_TwoCAEChallenges(t *testing.T) {
+	srv, close := mock.NewServer(mock.WithTransformAllRequestsToTestServerUrl())
+	defer close()
+	srv.AppendResponse(
+		mock.WithHeader("WWW-Authenticate", kvChallenge),
+		mock.WithStatusCode(401),
+	)
+	srv.AppendResponse(
```
chlowell, line 210: It's not obvious from outside the policy, but this test is effectively the same as the "KVThenCAE" test above. To avoid that overlap, it should send a second request that receives the two CAE challenges
```suggestion
	srv.AppendResponse()
	srv.AppendResponse(
```

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -100,6 +100,161 @@ func TestChallengePolicy(t *testing.T) {
...
+		switch tkReqs {
+		case 1:
+			require.Empty(t, tro.Claims)
+		case 2:
+			// second call should include challenge claims
+			require.Equal(t, "testing1", tro.Claims)
+		default:
+			t.Fatal("unexpected token request")
+		}
+		return azcore.AccessToken{Token: accessTk, ExpiresOn: time.Now().Add(time.Hour)}, nil
```
chlowell, line 188: Can you please also check that the policy uses the correct access token after handling the CAE challenge? Something simple like this should do in the credential:
```suggestion
		return azcore.AccessToken{Token: fmt.Sprint(tkReqs), ExpiresOn: time.Now().Add(time.Hour)}, nil
```
...then a predicate in the `AppendResponse` calls can compare the actual value of the `Authorization` header

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -100,6 +100,161 @@ func TestChallengePolicy(t *testing.T) {
 	}
 }
 
+var (
+	accessTk      = "***"
+	kvChallenge   = `Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://vault.azure.net"`
```
chlowell, line 105: Consider sharing these constants with the test above

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -22,6 +22,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	challengedToken = "needs more claims"
+	claimsToken     = "all the claims"
+	kvChallenge     = `Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://vault.azure.net"`
+	caeChallenge1   = `Bearer realm="", authorization_uri="https://login.microsoftonline.com/common/oauth2/authorize", error="insufficient_claims", claims="dGVzdGluZzE="`
```
chlowell, line 29: Sorry to nitpick but most of these should be declared `const`. Also, a package-level block should contain only symbols used by multiple tests. `requireToken` for example is better declared in `TestChallengePolicy_CAE` because it's only used there

chlowell (approved):

sdk/security/keyvault/internal/challenge_policy_test.go
gracewilcox: fixed in [ab1b903](https://github.com/Azure/azure-sdk-for-go/pull/23543/commits/ab1b903cd656fdea1ed7f38884e432d8c5671095)

sdk/security/keyvault/internal/challenge_policy_test.go
gracewilcox: fixed in 8de3f62

PR #23567, opened 2024-10-11
Storage [STG95] Files binary Ace

tanyasethi-msft:
About the feature being introduced - 
In Azure Files, we currently support setting file permissions on files via both SMB and REST. However, there is not full parity between these two protocols: there are permissions that can be set via SMB that cannot be set via REST. This is because SMB sends and receives file permissions in Window’s native binary encoding, while REST uses the SDDL serialization format. However, SDDL cannot encode all the permissions that the native encoding can. Namely, permissions with custom ACE cannot be encoded as SDDL. (For more information on custom ACE and SDDL, see Technical primer on the parity gap.) 

The goal of this design spec is to bring feature parity between the two protocols by augmenting the set of permissions that can be set via REST to the same set that can be set via SMB. 


tanyasethi-msft:
TODO : Few Tests remaining, Changelog updates, recording updates

sdk/storage/azfile/internal/shared/shared.go
```diff
@@ -47,6 +47,8 @@ const (
 	// Inherit inherits permissions from the parent folder (default when creating files/folders)
 	DefaultFilePermissionString = "inherit"
 
+	DefaultFilePermissionFormat = "Sddl"
```
souravgupta-msft, line 50: Default value refers to "sddl" (complete lower case) in docs. Confirm if this is the case.

sdk/storage/azfile/internal/testcommon/clients_auth.go
```diff
@@ -52,7 +52,9 @@ const (
 )
 
 const (
-	ISO8601 = "2006-01-02T15:04:05.0000000Z07:00"
+	ISO8601                  = "2006-01-02T15:04:05.0000000Z07:00"
+	FilePermissionFormatSddl = "Sddl"
+	FilePermissionBinary     = "Binary"
```
souravgupta-msft, line 57: Check the case-sensitivity

sdk/storage/azfile/directory/client_test.go
```diff
@@ -269,6 +269,48 @@ func (d *DirectoryRecordedTestsSuite) TestDirCreateDeleteDefault() {
...
+		FilePermissions: &file.Permissions{
+			Permission: &testcommon.SampleBinary,
+		},
+	})
+	_require.NoError(err)
+	_require.NotNil(resp.ETag)
+	_require.NotNil(resp.RequestID)
+	_require.Equal(resp.LastModified.IsZero(), false)
+
+	_, err = dirClient.Rename(context.Background(), "testFile", &directory.RenameOptions{
```
souravgupta-msft, line 305: Also add a test for directory set properties with this new option.

souravgupta-msft (approved):
Looks good overall. Added few comments

vibhansa-msft (approved):

sdk/storage/azfile/directory/client_test.go
tanyasethi-msft: Added

Issue #21063, opened 2023-06-26
Azure Cosmos DB: support for hierarchical partition keys


shubham1172:
### Feature Request

This issue is to support hierarchical partition keys via the Go-SDK in Azure Cosmos DB. Here are the docs for the same https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys?tabs=net-v3%2Cbicep

Pietrrrek:
I've opened up a [PR ](https://github.com/Azure/azure-sdk-for-go/pull/23577)

zackrossman:
This is a blocker for us adopting CosmosDB, thanks @Pietrrrek

rs-malik:
Our usecase is blocked by this as well. Thank you @Pietrrrek for working on this.

meain:
https://github.com/Azure/azure-sdk-for-go/pull/23577 seems to have added support for hierarchical partition keys. Is there anything else missing for hierarchical partition keys? Just checking to ensure that hierarchical partition keys are usable.

PR #23577, opened 2024-10-14
Add API for creating hierarchical `PartitionKey`s

Pietrrrek:

This PR adds a builder for creating `PartitionKey` structs with multiple keys in order to support hierarchical partitioning keys.

Related issue: #21063


Pietrrrek:
@microsoft-github-policy-service agree company="Boyum IT"

Pilchie:
@simorenoh and/or @analogrelay - can you take a look at this when you get a chance?

sdk/data/azcosmos/partition_key_builder.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+// PartitionKeyBuilder is used to build a PartitionKey.
+type PartitionKeyBuilder struct {
```
analogrelay, line 7: My gut feeling is that we don't need a brand-new builder type here. The existing `PartitionKey` struct already holds a `values []interface{}` value, and these `Append` methods could just be added to that type, making it mutable and supporting syntax like this:

```go
pk := NewPartitionKeyString("key0").AppendString("key1")
```

Now, I'll confess to not being quite as up-to-speed on the latest Go idioms, so I'm not sure how strongly we feel about making a currently-immutable type mutable, but putting the `Append` methods on `PartitionKey` feels more compact to me. It also removes the ambiguity of having both a "Builder" API for partition keys and direct constructors.

Pietrrrek: The main reason behind this design is that the same pattern is found in [azure-cosmos-dotnet-v3](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5e6232d25f84074d5008e97670d3e0dc07891f0e/Microsoft.Azure.Cosmos/src/PartitionKeyBuilder.cs), but I'll gladly refactor the PR to this design instead.

An alternative could be `func NewPartitionKey(parts ...interface{})`.

analogrelay: > The main reason behind this design is that the same pattern is found in [azure-cosmos-dotnet-v3](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/5e6232d25f84074d5008e97670d3e0dc07891f0e/Microsoft.Azure.Cosmos/src/PartitionKeyBuilder.cs), but I'll gladly refactor the PR to this design instead.

Gotcha. I don't have direct context on why .NET does it this way, but the main thing it does is some validation, which we've generally tried to leave to the server where possible (so that the client doesn't end up restricting something that is later allowed.

Not having the full context, if I were designing the .NET API at this point I think I'd suggest a similar design to what I'm suggesting here in Go, so I'm not too worried if we diverge from how the .NET SDK does things here. We can see what other folks on the SDK team think about that too.

> An alternative could be func NewPartitionKey(parts ...interface{}).

Yeah, I considered that, but we lose out on type-checking, since you could put anything in there. We'd have to either return an error out of `NewPartitionKey` (which means consumers have to add a bunch of code), panic (which is usually something to avoid, though this is a "programming" error rather than a recoverable error), or report that error when the request is serialized (which feels "too late").

I like the `AppendString`/`AppendNumber`/etc. patterns here, so I think I'd like to keep those if possible.

Pilchie: @RickWinter - any thoughts on the API pattern and Go idioms here?

jhendrixMSFT: I too would prefer the APIs on the `PartitionKey` type as having a dedicated builder seems C#/Java-ish. There are a few warts though.
- The current constructors return `PartitionKey` by value so you can't chain the calls (e.g. `NewPartitionKeyString("foo").AppendNumber(123)` won't compile). While we can fix this with a `NewPartitionKey() *PartitionKey` constructor, the lack of uniformity is a little clunky (changing the existing constructors to return pointer-to-type is a breaking change which we should avoid).
- APIs take `PartitionKey` by value so one would have to dereference the key var before/at the call site which looks a little weird (e.g. `SomeAPI(*NewPartitionKey().AppendString("foo"))`).

Having the builder elides the above issues at the cost of introducing a new type and its constructor. So, while I could live with the second bullet, the first one is nudging me towards the builder.

jhendrixMSFT: Side-by-side comparison.
```go
	SomeAPI(*NewPartitionKey().AppendString("foo"))

	// vs

	SomeAPI(NewPartitionKeyBuilder().AppendString("foo").Build())
```
Thoughts?

analogrelay: Oof, yeah I hadn't considered that issue (I'll admit it's been a bit since I worked actively in Go so I forgot it doesn't support auto-referencing to call methods).

Given those restrictions, I think the builder might be the best approach here. I don't love that the builder requires an additional `Build()` call, but I agree that the inconsistent constructors would be strange.

Another option is for the `Append` APIs to take the receiver by value:

```go
func (pk PartitionKey) AppendString(value string) PartitionKey {
	pk.components = append(pkb.components, value)
	return pk
}
```

That would allow this code to continue working:

```go
SomeAPI(NewPartitionKeyString("parent").AppendString("child"))
```

The syntax for appending to a `PartitionKey` stored in a variable isn't as tidy: `pk = pk.AppendString("...")` (though this mirrors things like `append`, and there's possibly _some_ kind of impact on performance (though tbh it could go either way depending on things like copy elision for by-value and memory barriers for by-reference, so we really don't know without measuring).

jhendrixMSFT: Good idea, and I think I can live with that trade-off assuming that callers will either capture the final `PartitionKey` (`pk := NewPartitionKey().AppendString("foo").AppendString("bar")`) or make the call in-place like above.

Pietrrrek: I'll have a go at the suggestions tomorrow morning (EU time) :^)

Pietrrrek:
The coverage tests are [failing](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4238297&view=logs&j=07a8bf49-95b8-5c7c-de1a-77230f20b6cf&t=79c67422-3cb3-50fc-4ea3-00a141e27961&l=25439), but the error message doesn't seem quite right 

```text
FAIL
coverage: 84.4% of statements
FAIL	github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos	49.481s
FAIL
Searching for coverage files in D:\a\_work\1\s
Reading config data from D:\a\_work\1\s\/eng/config.json
(data/azcosmos) Failing if the coverage is below 0.40
D:\a\_work\1\s\coverage.xml
Status: Succeeded	Coverage file: D:\a\_work\1\s\coverage.xml	 Coverage Amount: 0.8439
```

sdk/data/azcosmos/partition_key_builder.go
Pietrrrek: That's perfect, I noticed the errors before but didn't know whether/how I should address it since it wasn't caused by anything in this PR.

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,211 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
 	verifyEncodingScenario(t, container, "ComputeGW-IdWithUnicodeCharacters", "WithUnicode鱀", http.StatusCreated, http.StatusOK, http.StatusOK, http.StatusNoContent)
 }
 
+func TestItemCRUDHierarchicalPartitionKey(t *testing.T) {
```
jhendrixMSFT, line 506: I dug through the [test logs](https://dev.azure.com/azure-sdk/29ec6040-b234-4e31-b139-33dc4287b756/_apis/build/builds/4238297/logs/180) and this test is failing under the emulator.
```
2024-10-16T14:21:15.4860400Z RESPONSE 400: 400 Bad Request
2024-10-16T14:21:15.4861044Z ERROR CODE: 400 Bad Request
2024-10-16T14:21:15.4862521Z --------------------------------------------------------------------------------
2024-10-16T14:21:15.4863133Z {
2024-10-16T14:21:15.4864052Z   "code": "BadRequest",
2024-10-16T14:21:15.4865785Z   "message": "Message: {\"Errors\":[\"Too many partition key paths (2) specified. A maximum of 1 is allowed.\"]}\r\nActivityId: 564a7bbc-66a2-4ad1-8da6-be65731282cb, Request URI: /apps/DocDbApp/services/DocDbMaster0/partitions/780e44f4-38c8-11e6-8106-8cdcd42c33be/replicas/1p/, RequestStats: , SDK: Microsoft.Azure.Documents.Common/2.14.0"
2024-10-16T14:21:15.4866735Z }
```

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,211 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+		ExpectedSpans: []string{},
+	}))
+
+	database := emulatorTests.createDatabase(t, context.TODO(), client, "itemCRUDHierarchicalPartitionKey")
+	defer emulatorTests.deleteDatabase(t, context.TODO(), database)
+	properties := ContainerProperties{
+		ID: "aContainer",
+		PartitionKeyDefinition: PartitionKeyDefinition{
+			Paths: []string{"/id", "/type"},
+		},
```
analogrelay, line 518: Ah, I think this is where the issue is arising. When creating a multi-partition-key container, you need to set the `kind` field to `MultiHash`. We don't currently support that in Go, so you'll need to add support for that. For example, you can see the [JS SDK has a `kind` property](https://github.com/Azure/azure-sdk-for-js/blob/497321f0400ba4a2adfbd8f59db62d7574b52824/sdk/cosmosdb/cosmos/src/documents/PartitionKeyDefinition.ts#L20-L22). We'd need to add this field in a way that doesn't break existing users. It might be worth a custom `MarshalJSON`/`UnmarshalJSON` implementation here, which automatically sets `kind` based on the number of partition key paths 🤔

Pietrrrek: Done! I've added custom (de)serialization where the value for `Kind` is inferred if its value is an empty string.

Pietrrrek: The tests are still failing, but thankfully in new places - progress!
The current issue seems to be that the type of the PartitionKey in the container doesn't match that which is supplied in the headers, I guess that `PartitionKey` also needs to be extended with `Kind`.

sdk/data/azcosmos/partition_key_definition.go
```diff
@@ -3,11 +3,84 @@
...
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type PartitionKeyKind string
+
+const (
+	Hash      PartitionKeyKind = "Hash"
+	MultiHash PartitionKeyKind = "MultiHash"
```
analogrelay, line 17: For other "enums", we usually prefix the name of the variant with the type name, so these should be:

```suggestion
	PartitionKeyKindHash      PartitionKeyKind = "Hash"
	PartitionKeyKindMultiHash PartitionKeyKind = "MultiHash"
```

It's more verbose, for sure, but it's consistent with what we've done in other places.

sdk/data/azcosmos/partition_key_definition.go
```diff
@@ -3,11 +3,84 @@
...
+				return fmt.Errorf("unmarshalling type %T: %v", pkd, err)
+			}
+		}
+	}
+
+	if pkd.Kind == "" && len(pkd.Paths) == 1 {
+		pkd.Kind = Hash
+	} else if pkd.Kind == "" && len(pkd.Paths) > 1 {
+		pkd.Kind = MultiHash
+	}
```
analogrelay, line 85: I don't mind doing this inference when *serializing* the struct, but I think that when we're deserializing, we should leave it as the server (since that's usually where a JSON payload to be deserialized is coming from) provided it.

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,213 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+	}
+
+	itemBeta := map[string]interface{}{
+		"id":    "1",
+		"type":  "beta",
+		"value": "0",
+	}
+
+	pkAlpha := NewPartitionKey().AppendString("1").AppendString("alpha")
+	pkBeta := NewPartitionKey().AppendString("1").AppendString("alpha")
```
analogrelay, line 546:
```suggestion
	pkBeta := NewPartitionKey().AppendString("1").AppendString("beta")
```

I think this is a typo

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,213 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+
+	if item0Res.SessionToken == nil {
+		t.Fatalf("Session token is empty")
+	}
+
+	if len(item0Res.Value) != 0 {
+		t.Fatalf("Expected empty response, got %v", item0Res.Value)
+	}
+
+	item1Res, err := container.CreateItem(context.TODO(), pkAlpha, marshalledBeta, nil)
```
analogrelay, line 571:
```suggestion
	item1Res, err := container.CreateItem(context.TODO(), pkBeta, marshalledBeta, nil)
```

And this too?

sdk/data/azcosmos/emulator_cosmos_item_test.go
analogrelay: I think those are some typos in your test code actually, mismatching some places where you're specifying the "alpha" partition key but sending the "alpha" document as content.

sdk/data/azcosmos/partition_key_definition.go
analogrelay: I think that would mean you don't need to implement `UnmarshalJSON` at all and could just use the default implementation and the struct tags.

analogrelay (changes requested):
Closing in on it! A few minor adjustments and test fixes, but I think you're close!

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,213 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+		"id":    "1",
+		"type":  "alpha",
+		"value": "0",
+	}
+
+	itemBeta := map[string]interface{}{
+		"id":    "1",
+		"type":  "beta",
+		"value": "0",
+	}
```
analogrelay, line 543: I don't think the documents can have the same ID, even if the partition key differs

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +503,213 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+	if item1ResBody["type"] != "beta" {
+		t.Fatalf("Expected type to be beta, got %v", item1ResBody["type"])
+	}
+
+	if item1ResBody["value"] != "0" {
+		t.Fatalf("Expected value to be 0, got %v", item1ResBody["value"])
+	}
+
+	pkAll := NewPartitionKey().AppendString("1")
+	pager := container.NewQueryItemsPager("SELECT * FROM c", pkAll, nil)
```
analogrelay, line 640: The Go SDK currently only supports single-partition queries. For hierarchical partition keys, I believe that means you'll have to provide all the partition keys. Any query that specifies fewer than the total number of partition keys is considered a cross-partition query and requires additional support from the SDK (which Go does not currently provide).

sdk/data/azcosmos/emulator_cosmos_item_test.go
Pilchie: They should be able to.  Cosmos DB ensures pk + id is unique, but id doesn't have to be unique across partitions.

analogrelay: Ah yeah, you're right. I was misreading some docs. I just verified and it does indeed work as long as the partition key is different. I expect the remaining test issue is the missing partition key below then.

sdk/data/azcosmos/emulator_cosmos_item_test.go
analogrelay: @Pietrrrek I suspect this is the cause of the remaining failure. If you can update this to be a single-partition query, then I think we can get this landed!

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +504,196 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+
+	if item1ResBody["type"] != "beta" {
+		t.Fatalf("Expected type to be beta, got %v", item1ResBody["type"])
+	}
+
+	if item1ResBody["value"] != "0" {
+		t.Fatalf("Expected value to be 0, got %v", item1ResBody["value"])
+	}
+
+	pager := container.NewQueryItemsPager("SELECT * FROM c WHERE", pkAlpha, nil)
```
analogrelay, line 640: You're missing the rest of the where clause here. You could safely remove the entire `WHERE` since the `pkAlpha` parameter already restricts the query to a single partition.

```suggestion
	pager := container.NewQueryItemsPager("SELECT * FROM c", pkAlpha, nil)
```

sdk/data/azcosmos/emulator_cosmos_item_test.go
```diff
@@ -503,6 +504,196 @@ func TestItemIdEncodingComputeGW(t *testing.T) {
...
+
+	if len(alphaItems) != 1 {
+		t.Fatalf("Expected 1 item, got %v", len(alphaItems))
+	}
+
+	if !reflect.DeepEqual(alphaItems[0], item0ResBody) {
+		t.Fatalf("Expected %v, got %v", item0ResBody, alphaItems[0])
+	}
+
+	pager = container.NewQueryItemsPager("SELECT * FROM c WHERE", pkBeta, nil)
```
analogrelay, line 668: Same thing here

```suggestion
	pager = container.NewQueryItemsPager("SELECT * FROM c", pkBeta, nil)
```

analogrelay (commented):
Looks like just a query syntax error left 🤞🏻 . If you're having some trouble running the tests (I believe they require the emulator which can be difficult to get running on some platforms), let me know. I'd be happy to help, or jump in and see if I can fix these tests up so we can land this!

analogrelay (approved):
I went ahead and made the change to remove the dangling `WHERE` clauses. This seemed to fix the tests on my machine. Assuming the tests do indeed pass, this is good to go!

PR #23590, opened 2024-10-15
Allow the span kind to be set via StartSpanOptions

kjg:
This PR adds a `Kind` field to azcore's StartSpanOptions so that consumers of `azcore.StartSpan` can customize the span kind.

I plan to follow up with a PR against azcosmos where it can classify its spans as Client spans so that consumers of azcosmos have its spans show up like db client spans.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+* Added field `Kind` to `runtime.StartSpanOptions` to allow a kind to be set when starting a span.
+
 ### Breaking Changes
 
 ### Bugs Fixed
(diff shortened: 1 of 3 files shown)


kjg:
@microsoft-github-policy-service agree company="GitHub"

lmolkova (approved):
LGTM from general tracing perspective. Thank you!

jhendrixMSFT (approved):

kjg:
Thanks for merging this! In order to use this within azcosmos, a new version of azcore will need to be published. What's the process there? Are new versions published on a specific cadence, or is there a process for requesting new releases be published?

jhendrixMSFT:
Typical process is to release during the first week of every month. Can this wait for the next release cycle?

kjg:
Yep, this can wait. Thanks for the info

PR #23608, opened 2024-10-21
Revert "Revert "added new tests with recordings""

tanyasethi-msft:
Reverts Azure/azure-sdk-for-go#23607


tanyasethi-msft:
Reverting the previous change.

Issue #22396, opened 2024-02-15
Allow setting Azure Key Vault API version in ClientOptions for Secrets, Keys and Certificates


nagyesta:
### Feature Request

When a new client version is released because support is added to a new Azure Key Vault API version (like in case of the recently added 7.5 API support), the default API version is always the latest.

Since setting API Version is not possible, I am not able to continue using the 7.4 API if I want to.

I have tried setting the APi version like this:

```go
httpClient := prepareClient()
cred, _ := azidentity.NewDefaultAzureCredential(nil) // Will use Managed Identity via the Assumed Identity container
client, _ := azcertificates.NewClient("https://localhost:8443",
	cred,
	&azcertificates.ClientOptions{ClientOptions: struct {
		APIVersion       string
		Cloud            cloud.Configuration
		Logging          policy.LogOptions
		Retry            policy.RetryOptions
		Telemetry        policy.TelemetryOptions
		TracingProvider  tracing.Provider
		Transport        policy.Transporter
		PerCallPolicies  []policy.Policy
		PerRetryPolicies []policy.Policy
	}{APIVersion: "7.4", Transport: &httpClient}, DisableChallengeResourceVerification: true})
```

Which generated this error:

```
failed to get the secret: this client doesn't support overriding its API version
```

Assuming, that this is the intended way to set the API version, could you please add support to API version overrides?

I can understand if it can be only set to certain values ("7.2", "7.3", "7.4", "7.5", etc.), but it would help me reduce the number of times my build breaks if I had a way to set the API version to not use latest.

Would this be possible?

Thank you in advance!

gracewilcox:
Hi @nagyesta! With each release we generate from a specific service version. Could you use an older version of the SDK? For example, if you need to use service version 7.4 you can use v1.0.0 of the SDK.

Is there a particular reason why you need this feature? There shouldn't be any breaking changes between minor version releases of the SDK.

nagyesta:
Hi @gracewilcox ,
thank you for looking into this and thank you for the additional details!

I realize, that my case might be fairly uncommon and I should have shared more context to make it more compelling. Sorry for not doing so in my initial post! 

I need this feature, because I am the developer of an Azure Key Vault test double. As part of this effort, I am providing example code, just like the snippet I have shared above, to let our users (i.e., people who are using both Azure Key Vault, your SDK and my test double) see how they need to configure their tests when the test double is used.
[...]

PR #23639, opened 2024-10-23
[key vault] add api version support

gracewilcox:
Closes https://github.com/Azure/azure-sdk-for-go/issues/22396

Adds API version support. Instead of only being able to use the default version, users can now set the API version themselves.

### DIFF
--- a/sdk/security/keyvault/azadmin/backup/client_test.go
+++ b/sdk/security/keyvault/azadmin/backup/client_test.go
@@ -8,13 +8,17 @@ package backup_test
 
 import (
 	"context"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/mock"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/recording"
+	azcred "github.com/Azure/azure-sdk-for-go/sdk/internal/test/credential"
 	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azadmin/backup"
 	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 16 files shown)


sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -177,3 +181,48 @@ func TestBeginSelectiveKeyRestoreOperation(t *testing.T) {
...
+
+func TestAPIVersion(t *testing.T) {
+	apiVersion := "7.3"
+	var requireVersion = func(t *testing.T) func(req *http.Request) bool {
+		return func(r *http.Request) bool {
+			version := r.URL.Query().Get("api-version")
+			require.Equal(t, version, apiVersion)
+			return true
+		}
+	}
```
chlowell, line 193: `t` is in scope and you don't need to capture it, so you can simplify this to
```suggestion
	requireVersion := func(r *http.Request) bool {
		version := r.URL.Query().Get("api-version")
		require.Equal(t, apiVersion, version)
		return true
	}
```

sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -177,3 +181,48 @@ func TestBeginSelectiveKeyRestoreOperation(t *testing.T) {
...
+	srv, close := mock.NewServer(mock.WithTransformAllRequestsToTestServerUrl())
+	defer close()
+	srv.AppendResponse(
+		mock.WithHeader("WWW-Authenticate", `Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://managedhsm.azure.net"`),
+		mock.WithStatusCode(401),
+		mock.WithPredicate(requireVersion(t)),
```
chlowell, line 199: It doesn't make a practical difference but I'd forego this one because the discovery request's query parameters are actually irrelevant. Whether we set them is an implementation detail

sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -177,3 +181,48 @@ func TestBeginSelectiveKeyRestoreOperation(t *testing.T) {
...
+			version := r.URL.Query().Get("api-version")
+			require.Equal(t, version, apiVersion)
+			return true
+		}
+	}
+	srv, close := mock.NewServer(mock.WithTransformAllRequestsToTestServerUrl())
+	defer close()
+	srv.AppendResponse(
+		mock.WithHeader("WWW-Authenticate", `Bearer authorization="https://login.microsoftonline.com/tenant", resource="https://managedhsm.azure.net"`),
+		mock.WithStatusCode(401),
```
chlowell, line 198:
```suggestion
		mock.WithStatusCode(http.StatusUnauthorized),
```

sdk/security/keyvault/azadmin/backup/client_test.go
```diff
@@ -177,3 +181,48 @@ func TestBeginSelectiveKeyRestoreOperation(t *testing.T) {
...
+		mock.WithPredicate(requireVersion(t)),
+	)
+	srv.AppendResponse() // when a response's predicate returns true, srv pops the following one
```
chlowell, line 201: Just occurred to me this is a better pattern because it makes the test fail if `srv` returns the response we expect it to skip
```suggestion
	srv.AppendResponse(mock.WithStatusCode(http.StatusInternalServerError))
```

chlowell (approved):

PR #23647, opened 2024-10-23
Prep for release of adls stg95 preview

tanyasethi-msft:
This PR also corrects the shared key algorithm for adls, and a few linter errors for the pipeline to pass


jhendrixMSFT:
https://github.com/Azure/azure-sdk-for-go/pull/23536 updated the latest two versions of Go and it appears that the code coverage numbers are calculated differently now. It appears that previously, the percentages in the internal content didn't contribute to the overall coverage percentage, and now they do.

tanyasethi-msft:
The azdatalake coverage numbers are changed temporarily until we have a fix for the issue stated in the above comment.

eng/config.json
```diff
@@ -38,7 +38,7 @@
         },
         {
             "Name": "azdatalake",
-            "CoverageGoal": 0.60
+            "CoverageGoal": 0.00
```
jhendrixMSFT, line 41: According to the test results, current CC is 26.17%. Maybe set this to something like 0.25?

tanyasethi-msft: Fair, changed to 0.25 for now

Issue #23308, opened 2024-08-11
[Tracing] We need to be able to add attributes after span creation


lmolkova:
See https://github.com/Azure/azure-sdk-for-go/pull/23268#discussion_r1712146494.

Some information is available in response (like response properties, error codes, etc) and should be provided after span creation.

kjg:
In https://github.com/Azure/azure-sdk-for-go/pull/23652, we added some attributes after span creation by using 

```go
span := c.internal.Tracer().SpanFromContext(ctx)
span.SetAttributes([]tracing.Attribute{
	{Key: "db.cosmosdb.request_charge", Value: res.RequestCharge
})
```

However there might still be the opportunity for a nicer way to get access to a span than via the context.

PR #23652, opened 2024-10-23
Add some response values as attributes to the azcosmos traces

kjg:
When https://github.com/Azure/azure-sdk-for-go/pull/23268 was being considered there was [a request](https://github.com/Azure/azure-sdk-for-go/pull/23268#discussion_r1712146494) to add attributes based on the response.

This PR adds `db.cosmosdb.request_charge` and `db.cosmosdb.status_code` to all requests (at least the ones that make use of `executeAndEnsureSuccessResponse` which appear to be all of them right now.

I could not figure out how to get the sub status code out of the HTTP response. I'd be happy to add it here or in a follow up if I can get some help as to where to find it in the response.

Refs: https://github.com/Azure/azure-sdk-for-go/issues/23308

### DIFF
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/tracing"
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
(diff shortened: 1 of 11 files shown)


sdk/data/azcosmos/cosmos_client.go
```diff
@@ -526,6 +529,14 @@ func (c *Client) accountEndpointUrl() *url.URL {
 	return c.endpointUrl
 }
 
+func (c *Client) addResponseValuesToSpan(ctx context.Context, resp *http.Response) {
+	span := c.internal.Tracer().SpanFromContext(ctx)
+	span.SetAttributes(
+		tracing.Attribute{Key: "db.cosmosdb.request_charge", Value: newResponse(resp).RequestCharge},
```
Pilchie, line 535: I assume we're not worried about the implications of parsing this out of the response headers.

Pilchie (approved):
Seems fine to me, but would like another set of :eyes:

sdk/data/azcosmos/cosmos_client.go
analogrelay: It should be fine. [The code handles errors appropriately](https://github.com/kjg/azure-sdk-for-go/blob/5a97c4385564bc1f21c988a8aebf39dac25583e6/sdk/data/azcosmos/cosmos_response.go#L35) and will just return `0` if the header is invalid.

There's some minor duplication here, since most of the calling APIs are going to parse this as well, but I'm not sure it's worth reworking all the layering to prevent that duplication right now.

analogrelay (approved):
Looks good to me too.

Issue #23294, opened 2024-08-07
Provide a way to customize the span created by Pager[T].NextPage


jhendrixMSFT:
The `NextPage` method creates a span with a name based on the response envelope type. At present, there's no way to customize this or add any additional attributes. CosmosDB tracing requirements include a different span name than what we calculate and additional attributes.

PR #23658, opened 2024-10-24
Azcosmos no tracer to pager

kjg:
- Refs https://github.com/Azure/azure-sdk-for-go/issues/23294

Originally the thought was that the azcore Pager should provide a way to better customize the spans it creates, and that the azcosmos Pager based spans could take advantage of that.

As I was digging it to try to make that work I did realize a few things.

1. The spans being set within the azcosmos Fetcher functions are currently being surpressed due to the StartSpan code policy to not allow nested spans most of the time.
2. Adding a way to customize the span creation in the azcore Pager would add complexity for what is currently just a simple wrapper around the Fetcher function
3. If a Tracer is not set on the PagingHandler a null Tracer is used which results in StartSpan not adding an active span to the context.

Considering all the above. I think the simplest way to allow azcosmos to customize the tracer used for Pager spans is to not set a tracer in PagingHandler, and allow the custom one within the Fetcher function to take precedence.

For testing, I've pushed a commit with just the tests to show that before this change the tests fail to have the `query_` spans. Then I've pushed the commit that removes setting the Tracer and you can see the tests all pass now.

[Example of a failing test](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4265647&view=logs&j=07a8bf49-95b8-5c7c-de1a-77230f20b6cf&t=79c67422-3cb3-50fc-4ea3-00a141e27961):
Messages: span query_items aContainer wasn't found

### DIFF
--- a/sdk/data/azcosmos/cosmos_client.go
+++ b/sdk/data/azcosmos/cosmos_client.go
@@ -278,7 +278,6 @@ func (c *Client) NewQueryDatabasesPager(query string, o *QueryDatabasesOptions) 
 		},
 		Fetcher: func(ctx context.Context, page *QueryDatabasesResponse) (QueryDatabasesResponse, error) {
 			var err error
-			// Move the span to the pager once https://github.com/Azure/azure-sdk-for-go/issues/23294 is fixed
 			spanName, err := getSpanNameForClient(c.accountEndpointUrl(), operationTypeQuery, resourceTypeDatabase, c.accountEndpointUrl().Hostname())
 			if err != nil {
 				return QueryDatabasesResponse{}, err
@@ -307,7 +306,6 @@ func (c *Client) NewQueryDatabasesPager(query string, o *QueryDatabasesOptions) 
 
 			return newDatabasesQueryResponse(azResponse)
 		},
-		Tracer: c.internal.Tracer(),
 	})
 }
 
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -3,12 +3,13 @@
 ## 1.1.1 (Unreleased)
 
 ### Features Added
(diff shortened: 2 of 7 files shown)


kjg:
Adding tests now.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -307,7 +306,6 @@ func (c *Client) NewQueryDatabasesPager(query string, o *QueryDatabasesOptions)
 
 			return newDatabasesQueryResponse(azResponse)
 		},
-		Tracer: c.internal.Tracer(),
```
analogrelay, line 310: Just so I understand what's going on here, since I'm still ramping up a bit on the Go SDK at large: 

* Previously, we tried to create two nested spans here, the default one in `azcore.Pager`, then one we created in the Fetcher.
* We wanted to be able to push custom attributes up to the one in `azcore.Pager`, hence the comment referencing #23294.
* This didn't actually create the nested span, since we disallow nested spans
* Refactoring `azcore.Pager` to support custom spans is possible, but at this point we're the only ones who need that
* Rather than doing that refactor just for us (which I agree is probably unnecessary), we're just disabling the default `azcore.Pager` spans (by passing no `Tracer`) and keeping our custom ones in the Fetcher.

All that sounds reasonable to me. I'm definitely a fan of not over-refactoring abstractions to handle cases which are, at least currently, fairly niche.

analogrelay (approved):
LGTM

sdk/data/azcosmos/cosmos_client.go
kjg: Yep, that's exactly right!

kjg:
@analogrelay Is there anything else that needs to be done before this can be merged?

analogrelay:
Nope, this is good to go. Thanks for the ping!

PR #23672, opened 2024-10-29
[azopenai] Inference changes for 2024-10-01-preview

glecaros:


sdk/ai/azopenai/models.go
```diff
@@ -1105,6 +1122,9 @@ type ChatCompletionsOptions struct {
 	// A collection of textual sequences that will end completions generation.
 	Stop []string
 
+	// Options for streaming response. Only set this when you set stream: true.
```
richardpark-msft, line 1125: StreamOptions should be moved to GetChatCompletionsStreamOptions - we provide two separate functions where one of them enables streaming and the other doesn't and this option doesn't apply to the non-streaming mode.

Underneath it does use these structures, but we don't expose that to the user.

sdk/ai/azopenai/models.go
```diff
@@ -1578,6 +1598,9 @@ type CompletionsOptions struct {
...
```
richardpark-msft, line 1601: StreamOptions should be moved to GetCompletionsStreamOptions - we provide two separate functions where one of them enables streaming and the other doesn't and this option doesn't apply to the non-streaming mode.

richardpark-msft (changes requested):

richardpark-msft:
@glecaros, the CI failures probably just mean the tests need to be re-recorded.

sdk/ai/azopenai/models_extra.go
```diff
@@ -0,0 +1,282 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// ChatCompletionsOptionsFunctionCall - Controls how the model responds to function calls. "none" means the model does not
+// call a function, and responds to the end-user. "auto" means the model can pick between an end-user or calling a
+// function. Specifying a particular function via {"name": "my_function"} forces the model to call that function. "none" is
+// the default when no functions are present. "auto" is the default if functions
+// are present.
+type ChatCompletionsOptionsFunctionCall struct {
```
richardpark-msft, line 18: Why were these moved from custom_models.go?

sdk/ai/azopenai/client_extra.go
```diff
@@ -0,0 +1,255 @@
...
+		return err
+	}
+
+	return &ContentFilterResponseError{
+		ResponseError:        *respErr,
+		ContentFilterResults: envelope.Error.InnerError.ContentFilterResults,
+	}
+}
+
+func (client *Client) formatURL(path string, deployment *string) string {
```
richardpark-msft, line 63: Similarly, we have custom_client.go. What's the distinction you're going for between client_extra and custom_client?

richardpark-msft (approved):
Just a comment and a changelog entry and I think it's good.

After this we'll want to submit a new APIView as well.

sdk/ai/azopenai/models_extra.go
richardpark-msft: (offline conversation - it's needed because of a chicken-and-egg situation with code gen)

sdk/ai/azopenai/client_extra.go
richardpark-msft: (offline conversation - it's needed because of a chicken-and-egg situation with code gen)

PR #23713, opened 2024-11-07
Refresh OIDC token in azadmin test resources post script

benbp:
Refresh OIDC token in azadmin test resources post script


gracewilcox (approved):

benbp:
Closing as this doesn't help, the token still expires by the time deployment completes. We'll need a creative solution.

Issue #23307, opened 2024-08-11
[Tracing] Allow Client libraries to configure schema URL


lmolkova:
See https://github.com/Azure/azure-sdk-for-go/pull/23268#discussion_r1702424904 - different SDKs may emit telemetry of slightly different versions and we can't keep them in sync (e.g. servicebus and cosmosdb SDKs are owned by different teams, follow different semantic conventions, and prioritize telemetry differently, i.e. they need to be able to set different schema urls.

We currently hardcode and use schema url here - one for everyone:

https://github.com/Azure/azure-sdk-for-go/blob/969cfcd7a36018cab9461bc768bdd0c8f1ff4e00/sdk/tracing/azotel/otel.go#L31

This is blocker to stabilize tracing plugin (and tracing for anything beyond http).

kjg:
@lmolkova It seems like current azcore is responsible for tracing the HTTP spans and Client libraries can add additional tracing on top of that. Would it make sense for the Client Libraries to use a separate Tracer than Azcore? That way Azcore could have one schema URL and client libraries could their own?

lmolkova:
@kjg absolutely! 
One of the important (from otel perspective) this is that each tracer has a proper instrumentation scope (library name) so that it's clear where the telemetry is coming from and also it's possible to configure it on per-scope level. I.e. we should create a tracer per client. 

That's [what we do in Java](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/util/tracing/TracerProvider.java#L54):

- each library creates its own tracer instance with library name, version and, since recently, the schema url it follows. E.g. [here](https://github.com/Azure/azure-sdk-for-java/blob/d8af197b6238149d4520458b2cee9d48831f58d3/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java#L301)  Schema URL was added recently - https://github.com/Azure/azure-sdk-for-java/pull/41789. All these are library options - users don't get to change them.
- It also includes user-facing (if any) tracing configuration provided by users on the client level.

kjg:
[...]

PR #23715, opened 2024-11-08
Azcore separate tracer

kjg:
- Fixes https://github.com/Azure/azure-sdk-for-go/issues/23307

- This PR allows a schemaURL to be passed to the NewTracer function so that consumers can create tracers that use a different schemaURL than azcore.
- It updates azcore to use the otel semconv package that sets a schemaURL and then uses the semconv Constances for the span keys that have defined semconv keys.

This PR also adds a `WithCoreTracerName` function to the azcore client so the consumers that do want to do custom tracing can make azcore using azcore's module name and version for the HTTP tracing, and their own module name and version for their portion of the tracing. Using `WithCoreTracerName` only changes the values used for tracing to azcore values, it still leaves the passed in module name and version for use with the HTTP User Agent header.

I chose this approach because it looks like many of the consumers of this client 

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -4,7 +4,11 @@
 
 ### Features Added
 
+* `Client` has a new method called `WithCoreTracerName` that can be used by consumers that wish to have the azcore http tracer's module name and version set to azcore's values instead of the consumer modules values
+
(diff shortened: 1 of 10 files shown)


kjg:
I'd love to keep working on this if I can get some guidance. Thanks!

PR #23729, opened 2024-11-12
Add zos support to storage/azblob

MacMalainey:
Add support for zos to storage/azblob. This is mainly to enable building modules that depend on the azure-sdk module in some capacity.

Due to hardware requirements tests are not updated.


vibhansa-msft (approved):

tanyasethi-msft:
Hi @MacMalainey , do we want to update only in blob or in ADLS as well?

MacMalainey:
@tanyasethi-msft only blob for now - we have an actual case that uses blob but not ADLS

PR #23733, opened 2024-11-12
Add OperationLocationResultPath to NewPollerOptions[T]

jhendrixMSFT:
The result path must be configurable instead of hard-coded to result.

### DIFF
--- a/sdk/azcore/runtime/poller.go
+++ b/sdk/azcore/runtime/poller.go
@@ -50,7 +50,13 @@ const (
 // NewPollerOptions contains the optional parameters for NewPoller.
 type NewPollerOptions[T any] struct {
 	// FinalStateVia contains the final-state-via value for the LRO.
+	// NOTE: used only for Azure-AsyncOperation and Operation-Location LROs.
 	FinalStateVia FinalStateVia
+
+	// OperationLocationResultPath contains the JSON path to the result's
+	// payload when it's included with the terminal success response.
+	// NOTE: only used for Operation-Location LROs.
+	OperationLocationResultPath string
 
 	// Response contains a preconstructed response type.
 	// The final payload will be unmarshaled into it and returned.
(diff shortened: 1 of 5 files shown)


sdk/azcore/runtime/poller.go
```diff
@@ -50,8 +50,14 @@ const (
 // NewPollerOptions contains the optional parameters for NewPoller.
 type NewPollerOptions[T any] struct {
 	// FinalStateVia contains the final-state-via value for the LRO.
+	// NOTE: used only for Azure-AsyncOperation and Operation-Location LROs.
 	FinalStateVia FinalStateVia
 
+	// OperationLocationResultPath contains the JSON path to the result's
+	// payload when it's included with the terminal success response.
+	// NOTE: only used for Operation-Location LROs.
+	OperationLocationResultPath string
```
jhendrixMSFT, line 59: The name is long and I'd like to shorten it to `ResultPath`. However, since it's only applicable to `Operation-Location` LROs having the prefix does clarify the usage. Maybe the doc comment is sufficient? We already have a similar problem with `FinalStateVia`.

JeffreyRichter (approved):

sdk/azcore/runtime/poller.go
jhendrixMSFT: Will leave with the long name as this field is specific to `Operation-Location`.

yyoshiki41:
When will this pull request be released? I have been eagerly waiting for it.

richardpark-msft (approved):

jhendrixMSFT:
@yyoshiki41 sorry for the delay, this is now merged. We have a release moratorium for December due to the US holidays and people being out. This will get officially released in early Jan.

yyoshiki41:
I'm looking forward to it 👍 

Thanks!

PR #23748, opened 2024-11-13
Update azopenai* to use hardcoded variable group config

benbp:
Update azopenai* to use hardcoded variable group config


benbp:
@weshaggard FYI I'm just getting go in line with python/js because the sub config approach with the custom service connection wasn't working for annoying reasons (env setting, auth, etc.). Since I'm going to move everything over to the splatting approach, I think we can just get this yaml checked in to unblock azopenai live testing before Ignite.

richardpark-msft:
Thanks for doing this!

weshaggard (approved):

PR #23815, opened 2024-12-03
Allow data plane module paths

richardpark-msft:
(NOTE: I'm not sure if this is the true home for this code, or just a copy - please advise!)

Data plane module paths don't include `resourcemanager` and `arm`, so they were failing the previous regex. I modified it to allow both types, and, if it's data plane, to allow 'az' as the prefix.

This was causing a build failure for me here: https://github.com/Azure/azure-rest-api-specs/pull/31719/checks?check_run_id=33822699487

### DIFF
--- a/eng/tools/generator/typespec/typespec-go.go
+++ b/eng/tools/generator/typespec/typespec-go.go
@@ -42,11 +42,16 @@ func NewGoEmitterOptions(emitOption any) (*GoEmitterOptions, error) {
 	return &option, err
 }
 
-const moduleRegex = `^github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/(\w+)/arm(\w+)$`
+const moduleRegex = `^github.com/Azure/azure-sdk-for-go/sdk/` +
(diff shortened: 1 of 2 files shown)


tadelesh (approved):
the generator tool previously only implemented for control plane package generation, thus cause the ci failure. i don't know whether we should enable sdk automation for typespec for go data plane. @lirenhe thoughts? but this pr lgtm, it could at least unblock the pr merge.

richardpark-msft:
> > the generator tool previously only implemented for control plane package generation, thus cause the ci failure. i don't know whether we should enable sdk automation for typespec for go data plane. @lirenhe thoughts? but this pr lgtm, it could at least unblock the pr merge.
> 
> Just for my own education - the file I modified obviously is in the generator tool. What is the rest of the process if we wanted to enable data plane? Am I just stepping onto a big treadmill or is this just a few minor items?

I've moved this discussion over to here: https://github.com/Azure/azure-sdk-for-go/issues/23820

PR #23816, opened 2024-12-03
Changing aznamespaces over to using tsp-client

richardpark-msft:
Mostly the changes were cosmetic since we seem to be pulling in some more recent commits (unrelated to the change, just a result of me getting changes into azure-rest-api-specs `main` branch).


sdk/messaging/eventgrid/aznamespaces/custom_models.go
```diff
@@ -6,6 +6,20 @@
 
 package aznamespaces
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// Error - The error object.
+type Error struct {
```
jhendrixMSFT, line 15: Which version of the Go emitter is being used? This looks vestigial and is due to a buggy version. The latest is `v0.3.4` which should have fixed this.

richardpark-msft: Two questions, so I must provide two answers ;)

> Which version of the Go emitter is being used? This looks vestigial and is due to a buggy version. The latest is v0.3.4 which should have fixed this.

Since we're using tsp-client it's configured here: https://github.com/Azure/azure-sdk-for-go/blob/8fef5b723be95b4de2d4108deacb17408b26269f/eng/emitter-package.json

So I'm using:

```json
    "@azure-tools/typespec-go": "0.3.3"
```

However...

> Oops just noticed this is in hand-written code. Why is it needed when the emitter removes it in the generated code?

It's referenced as part of the FailedLockToken structure, as an Azure.Core.Foundations.Error - in the past we've been creating an Error type, in the package, to handle the deserialization of it so that's what's going on here, as well.

jhendrixMSFT (approved):

PR #23830, opened 2024-12-05
[azsystemevents] switch over to use tsp-client

richardpark-msft:


richardpark-msft:
@jhendrixMSFT, this PR is done, btw - we don't have live tests for the system events package as it doesn't have a service endpoint, in case you're waiting for those to run.

jhendrixMSFT:
Pretty big deltas. This is expected?

richardpark-msft:
> Pretty big deltas. This is expected?

Yes, they're cleaning house on old events as we contact the service teams. The media events all got removed as part of that.

I'll update the changelog.

jhendrixMSFT (approved):

PR #23860, opened 2024-12-11
[azservicebus] Enable distributed tracing

karenychen:


sdk/messaging/azservicebus/internal/tracing/fake_tracing.go
karenychen, line 1: Currently a couple of other packages have something similar to validate the spans in unit tests. We should move this to `azcore/tracing` so more libraries can use this fake trace provider in their tests?

sdk/messaging/azservicebus/sender_unit_test.go
```diff
@@ -45,14 +46,45 @@ func TestSender_UserFacingError(t *testing.T) {
 
 	var asSBError *Error
 
-	err = sender.SendMessage(context.Background(), &Message{}, nil)
+	msgID := "testID"
```
richardpark-msft, line 49: We still want the original test (	err = sender.SendMessage(context.Background(), &Message{}, nil) here, but otherwise loving the addition of tracing to the mix.

sdk/messaging/azservicebus/internal/constants.go
```diff
@@ -3,5 +3,7 @@
 
 package internal
 
+const ModuleName = "azservicebus"
```
jhendrixMSFT, line 6: This should be the fully qualified module name.

sdk/messaging/azservicebus/internal/tracing/fake_tracing.go
jhendrixMSFT: It would probably go in the `sdk/internal` module so it can be internally shared but not publicly surfaced.

sdk/messaging/azservicebus/client_test.go
```diff
@@ -471,6 +472,48 @@ func TestNewClientUnitTests(t *testing.T) {
 		require.EqualValues(t, ns.FQDN, "mysb.windows.servicebus.net")
 	})
 
+	t.Run("TracerIsSetUp", func(t *testing.T) {
+		hostName := "fake.servicebus.windows.net"
+		// when tracing provider is not set, use a no-op tracer.
+		client, err := NewClient(hostName, struct{ azcore.TokenCredential }{}, nil)
```
richardpark-msft, line 478: Side note, love this construct: "struct{ azcore.TokenCredential }{}". Makes total sense, and I'd never thought of doing it before.

karenychen:
Hi @lmolkova ! I had a small question regarding diagnostic-id in https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-end-to-end-tracing?tabs=net-standard-sdk-2
 
The .NET SDK seems to be hooking them up to a ReceiveMessages trace when the users set the diagnostic-id and linking the list of diagnostic ids from the messages to the Receive trace ([code here](https://github.com/Azure/azure-sdk-for-net/blob/54e9026c2d9be6245cc78d9837e46891f066f783/sdk/servicebus/Microsoft.Azure.ServiceBus/src/ServiceBusDiagnosticsSource.cs#L694)). I might be misunderstanding how .NET is doing it, but I am wondering what we enable with the diagnostic-ids?

sdk/messaging/azservicebus/client.go
```diff
@@ -369,3 +384,18 @@ func (client *Client) getCleanupForCloseable() (uint64, func()) {
...
 }
+
+// getHostName returns fullyQualifiedNamespace if it is set, otherwise it returns the host name from the connection string.
+// If the connection string is not in the expected format, it returns an empty string.
+func getHostName(creds clientCreds) string {
+	if creds.fullyQualifiedNamespace != "" {
+		return creds.fullyQualifiedNamespace
+	}
+
+	parts := strings.Split(creds.connectionString, "/")
```
richardpark-msft, line 395: We have an actual parser for the connection string, we should definitely use that. Look in `internal/conn`.

sdk/messaging/azservicebus/internal/constants.go
karenychen: Could you elaborate on what the fully qualified module name should be? For context, I was following the module naming for eventgrid: https://github.com/Azure/azure-sdk-for-go/blob/679a930e6edd51f69f811bc1f8296a0dece799fa/sdk/messaging/eventgrid/azeventgrid/internal/version.go#L9

sdk/messaging/azservicebus/tracing.go
```diff
@@ -0,0 +1,42 @@
...
+		tracer.SetAttributes(
+			tracing.Attribute{Key: tracing.ServerAddress, Value: hostName},
+		)
+	}
+
+	return tracer
+}
+
+func getSpanAttributesForMessage(message *Message) []tracing.Attribute {
+	attrs := []tracing.Attribute{}
```
richardpark-msft, line 32:
```suggestion
	var attrs []tracing.Attribute
```

I swear there's some linter that complains if you pre-init the slice and it's not technically needed anyways.

sdk/messaging/azservicebus/sender.go
```diff
@@ -93,7 +106,13 @@ type SendMessageBatchOptions struct {
...
 func (s *Sender) SendMessageBatch(ctx context.Context, batch *MessageBatch, options *SendMessageBatchOptions) error {
-	err := s.links.Retry(ctx, EventSender, "SendMessageBatch", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
+	var err error
+	ctx, endSpan := s.startSpan(ctx, "SendMessageBatch", tracing.SendOperationName,
+		tracing.Attribute{Key: tracing.BatchMessageCount, Value: int64(batch.NumMessages())},
+	)
+	defer func() { endSpan(err) }()
+
+	err = s.links.Retry(ctx, EventSender, "SendMessageBatch", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
 		return lwid.Sender.Send(ctx, batch.toAMQPMessage(), nil)
```
richardpark-msft, line 116: @lmolkova, in cases like this where I have retries, should the span reporting be _in_ the retry loop, or outside, like we have here?

sdk/messaging/azservicebus/sender.go
```diff
@@ -123,7 +149,14 @@ type ScheduleAMQPAnnotatedMessagesOptions struct {
 // delivered can be cancelled using `Receiver.CancelScheduleMessage(s)`
 // If the operation fails it can return an [*azservicebus.Error] type if the failure is actionable.
 func (s *Sender) ScheduleAMQPAnnotatedMessages(ctx context.Context, messages []*AMQPAnnotatedMessage, scheduledEnqueueTime time.Time, options *ScheduleAMQPAnnotatedMessagesOptions) ([]int64, error) {
-	return scheduleMessages(ctx, s.links, s.retryOptions, messages, scheduledEnqueueTime)
+	var err error
+	ctx, endSpan := s.startSpan(ctx, "ScheduleAMQPAnnotatedMessages", tracing.ScheduleOperationName,
+		tracing.Attribute{Key: tracing.BatchMessageCount, Value: int64(len(messages))},
+	)
+	defer func() { endSpan(err) }()
```
richardpark-msft, line 156: @jhendrixMSFT, would it be worth building this pattern (via a callback, probably) into the tracing library? It can be internal, but it seems like everyone's going to do the "last error gets passed to endSpan before block ends" pattern.

richardpark-msft: If not, @karenychen, we can build a helper function - maybe we'd stick it right in the retry function to make things easier since we're passing very similar information to both.

sdk/messaging/azservicebus/internal/constants.go
jhendrixMSFT: It's the value of `module` from the `go.mod` file so `github.com/Azure/azure-sdk-for-go/sdk/messaging/azservicebus`.

Looks like `azeventgrid` is incorrect. :(

sdk/messaging/azservicebus/sender.go
```diff
@@ -199,7 +238,20 @@ func (sender *Sender) createSenderLink(ctx context.Context, session amqpwrap.AMQ
 	return amqpSender, nil, nil
 }
 
+func (s *Sender) startSpan(ctx context.Context, spanName string, operationName tracing.MessagingOperationName, attrs ...tracing.Attribute) (context.Context, func(error)) {
+	spanName = fmt.Sprintf("Sender.%s", spanName)
+	attributes := []tracing.Attribute{
```
richardpark-msft, line 243: Maybe we should define these all the operation constants as 'Sender.ScheduledMessages' (ie, preformatted) and avoid the formatting operation/concat that we have to do here. That's easily for a future PR, I'm sure there's other spots I do the same thing.

richardpark-msft (commented):
Looks great so far, got some questions for the other experts in the crowd, but from an SB perspective it looks great.

sdk/messaging/azservicebus/sender_unit_test.go
karenychen: I think that might be git trying to trick us -- the test is still there on line 60!

sdk/messaging/azservicebus/internal/constants.go
richardpark-msft: Looks like we used to generate/have it one way, and now it goes another. I guess the issue is I haven't regenerated azeventgrid recently enough.

richardpark-msft: Ah, maybe it's not even auto-generated. I note that aztables swapped from the short to the long format, so I probably just missed the memo.

sdk/messaging/azservicebus/sender.go
jhendrixMSFT: Maybe it goes in the `sdk/internal` module?

sdk/messaging/azservicebus/sender.go
karenychen: Just to clarify, did we want to define the span names when we are calling the span, or did we want to have a set of pre-defined span names (similar to the operation name/type below) where we can directly grab and use?

richardpark-msft: "When calling" is fine, but having predefined 'const''s could be nice from a documentation perspective (ie: these are all the spans that exist).

sdk/messaging/azservicebus/sender.go
karenychen: Sounds good, I noticed the .NET SDK gives a more general naming to their spans, we can also follow their naming convention: https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-end-to-end-tracing?tabs=net-standard-sdk-2#instrumented-operations

sdk/messaging/azservicebus/client_test.go
karenychen: copilot 🐐

richardpark-msft: 🤖

sdk/messaging/azservicebus/sender.go
jhendrixMSFT: The way it works in our HTTP SDKs is that the method span is above the retry layer, and its child HTTP span is below the retry layer. So you'd have spans like this.
```
Some method call span
  HTTP span retry 1
  HTTP span retry 2
```
I presume we'd want to do the same thing here.

karenychen: @jhendrixMSFT I did a bit of digging -- the semantic conventions for HTTP has examples for how the spans should look like for retries: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-authorization-retry-examples. However, for messaging systems there is not much information besides [this chunk](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/): 

image

I am not sure if this implies the messaging spans are 1 per operation?

sdk/messaging/azservicebus/internal/amqpLinks.go
```diff
@@ -66,6 +67,12 @@ type AMQPLinks interface {
 
 	// Prefix is the current logging prefix, usable for logging and continuity.
 	Prefix() string
+
+	// Tracer returns the tracer for the AMQPLinks instance.
+	Tracer() tracing.Tracer
+
+	// SetTracer sets the tracer for the AMQPLinks instance.
+	SetTracer(tracing.Tracer)
```
richardpark-msft, line 75: Does AMQPLinks needs to own a Tracer object or should it just be passed in as an argument to each function call?

sdk/messaging/azservicebus/internal/namespace.go
```diff
@@ -433,7 +434,7 @@ func (ns *Namespace) startNegotiateClaimRenewer(ctx context.Context,
 				return
 			case <-time.After(nextClaimAt):
 				for {
-					err := utils.Retry(refreshCtx, exported.EventAuth, "NegotiateClaimRefresh", func(ctx context.Context, args *utils.RetryFnArgs) error {
+					err := utils.Retry(refreshCtx, tracing.NewNoOpTracer(), exported.EventAuth, "NegotiateClaimRefresh", func(ctx context.Context, args *utils.RetryFnArgs) error {
```
richardpark-msft, line 437: Is the NoOpTracer here temporary or is there a reason we shouldn't trace this?

sdk/messaging/azservicebus/sender.go
richardpark-msft: Although I'd say that we'd do it the same way as what @jhendrixMSFT outlined above - start a span when the whole retry() function starts, and then a new sub-span within each retry operation.

sdk/messaging/azservicebus/liveTestHelpers_test.go
```diff
@@ -186,10 +187,11 @@ func deleteSubscription(t *testing.T, ac *admin.Client, topicName string, subscr
 // and fails tests otherwise.
 func peekSingleMessageForTest(t *testing.T, receiver *Receiver) *ReceivedMessage {
 	var msg *ReceivedMessage
+	// TODO
```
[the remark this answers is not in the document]
richardpark-msft: You can run the same suite of live tests. I need to update the sample.env to list them properly, but for now you can see the ones we use here: https://github.com/karenychen/azure-sdk-for-go/blob/bd50b2a1e4c72b1d22ba11314d315d939796c201/sdk/messaging/azservicebus/internal/test/test_helpers.go#L81

Just create a .env file and place it in the `azservicebus` folder, and run `go test`.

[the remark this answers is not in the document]
richardpark-msft: Also, you can add `/azp run go - azservicebus` as a comment on your PR to run it as part of your CI.

sdk/messaging/azservicebus/internal/tracing/tracing.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type SpanConfig struct {
+	name       SpanName
+	attributes []Attribute
+}
+
+type SetAttributesFn func([]Attribute) []Attribute
+
+func NewSpanConfig(name SpanName, options ...SetAttributesFn) *SpanConfig {
```
richardpark-msft, line 29: I think this indirection (for SetAttributesFn) is a bit odd - it looks like `attributes ...Attribute` would cover what we use it for, and helps a bit in showing that's the only thing we use it for.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -361,6 +369,12 @@ func (r *Receiver) DeadLetterMessage(ctx context.Context, message *ReceivedMessa
 }
 
 func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, options *ReceiveMessagesOptions) ([]*ReceivedMessage, error) {
+	var err error
```
richardpark-msft, line 372: If you move this span code into `ReceiveMessages` it'll be less tricky since you'll only have a single spot where you need to log the error, and we don't have to be wary of assigning `err` in all the code here.

richardpark-msft (commented):
It's coming along really well. I left some comments, but I'm really liking it.

sdk/messaging/azservicebus/internal/namespace.go
karenychen: The claim negotiation is not outlined in the Otel conventions for SB, so I left it to a no-op tracer https://opentelemetry.io/docs/specs/semconv/messaging/azure-messaging/

Definitely open to adding it if we want to support it though

richardpark-msft: Hm. I think we could consider it a request, just not user initiated. We can figure it out later, but maybe we could file a separate issue so we pick it up later.

It's definitely useful info.

sdk/messaging/azservicebus/internal/amqpLinks.go
karenychen: For this I am open to your preference :) Currently the tracer starts a span at the `Retry()` function level. So we can either have it in the amqpLink layer, or keep it in the Sender/Receiver layer and passing it as an argument to each function call all the way down to the `Retry()` function level. Which option do you think is more appropriate here?

richardpark-msft: Let me know if it gets too gnarly though.

sdk/messaging/azservicebus/internal/namespace.go
karenychen: Ah! Just noticed we can use `tracing.SpanKind` to denote that it is an internal span.

https://github.com/Azure/azure-sdk-for-go/blob/af2aacb0bf5b03231cba3fdc08e330469f297cd4/sdk/azcore/tracing/constants.go#L7C1-L27C2

I will make that change

sdk/messaging/azservicebus/receiver.go
karenychen: updated!

sdk/messaging/azservicebus/internal/tracing/tracing.go
karenychen: switched this one up a bit to directly return list of attributes, let me know what you think!

sdk/messaging/azservicebus/internal/amqpLinks.go
karenychen: Updated to move the tracers 1 level up. Now they live in `Sender`, `Receiver` and `Namespace`

sdk/messaging/azservicebus/sender.go
```diff
@@ -93,9 +95,18 @@ type SendMessageBatchOptions struct {
...
+			getMessageBatchSpanAttributes(int(batch.NumMessages()))...,
+		),
+	}
+
 	err := s.links.Retry(ctx, EventSender, "SendMessageBatch", func(ctx context.Context, lwid *internal.LinksWithID, args *utils.RetryFnArgs) error {
 		return lwid.Sender.Send(ctx, batch.toAMQPMessage(), nil)
-	}, RetryOptions(s.retryOptions))
+	}, RetryOptions(s.retryOptions), to)
 
 	return internal.TransformError(err)
```
lmolkova, line 111: we need to make sure we record error on the span.
See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#span-status for the details

sdk/messaging/azservicebus/sender.go
```diff
@@ -123,10 +135,19 @@ type ScheduleAMQPAnnotatedMessagesOptions struct {
...
-	return scheduleMessages(ctx, s.links, s.retryOptions, messages, scheduledEnqueueTime)
+	sequenceNumbers, err := scheduleMessages(ctx, s.tracer, s.links, s.retryOptions, messages, scheduledEnqueueTime)
+	return sequenceNumbers, err
 }
 
-func scheduleMessages[T amqpCompatibleMessage](ctx context.Context, links internal.AMQPLinks, retryOptions RetryOptions, messages []T, scheduledEnqueueTime time.Time) ([]int64, error) {
+func scheduleMessages[T amqpCompatibleMessage](ctx context.Context, tracer tracing.Tracer, links internal.AMQPLinks, retryOptions RetryOptions, messages []T, scheduledEnqueueTime time.Time) ([]int64, error) {
+	to := &tracing.TracerOptions{
+		Tracer:   tracer,
+		SpanName: tracing.ScheduleSpanName,
```
lmolkova, line 145: SB-specific operation names (language-agnostic) are documented here https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/azure-messaging.md

lmolkova (commented):
Thank! Left some comments on naming and following conventions more closely

sdk/messaging/azservicebus/sender.go
serbrech: I recall seeing this done with event entries within a span, instead of additional spans.

sdk/messaging/azservicebus/sender.go
```diff
@@ -93,9 +95,18 @@ type SendMessageBatchOptions struct {
 // Message batches can be created using [Sender.NewMessageBatch].
 // If the operation fails it can return an [*azservicebus.Error] type if the failure is actionable.
 func (s *Sender) SendMessageBatch(ctx context.Context, batch *MessageBatch, options *SendMessageBatchOptions) error {
+	to := &tracing.TracerOptions{
```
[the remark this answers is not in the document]
serbrech: https://github.com/pyohannes/semantic-conventions/blob/b836bbc45d8e49825ce3a2bc03e0be18b72d8fa5/docs/messaging/messaging-spans.md#batch-publishing

sdk/messaging/azservicebus/sender.go
karenychen: TODO: we should set the `errors.type` attribute in  the [azcore/runtime](https://github.com/Azure/azure-sdk-for-go/blob/799c8579d2d8f32d3f30457cd6da736be08d9c6b/sdk/azcore/runtime/policy_http_trace.go#L147) package

sdk/internal/test/tracing/validator.go
karenychen, line 1: This is a span validator used to validate spans in unit tests. I will make a separate PR on it but putting it here for reference

sdk/messaging/azservicebus/messageSettler.go
```diff
@@ -22,27 +24,32 @@ type messageSettler struct {
...
 	}
 }
 
-func (s *messageSettler) settleWithRetries(ctx context.Context, settleFn func(receiver amqpwrap.AMQPReceiver, rpcLink amqpwrap.RPCLink) error) error {
+func (s *messageSettler) settleWithRetries(ctx context.Context, settleFn func(receiver amqpwrap.AMQPReceiver, rpcLink amqpwrap.RPCLink) error, to *tracing.StartSpanOptions) error {
 	if s == nil {
 		return internal.NewErrNonRetriable("messages that are received in `ReceiveModeReceiveAndDelete` mode are not settleable")
 	}
 
+	if to != nil {
```
karenychen, line 42: We are setting the tracer here so we don't hit a nil-pointer dereference in cases where messageSettler is set to nil in a `receive-and-delete` sender

sdk/messaging/azservicebus/internal/tracing/tracing.go
```diff
@@ -0,0 +1,158 @@
...
+			Attributes: attrs,
+			Links:      options.Links,
+		})
+	return ctx, func(err error) {
+		// unwrap any errors from upstream
+		if unwrappedErr := errors.Unwrap(err); unwrappedErr != nil {
+			endSpan(unwrappedErr)
+			return
+		}
+		endSpan(err)
```
karenychen, line 110: needed this so we are not just logging `fmt.wrapError` as the error type

sdk/messaging/azservicebus/messageSettler.go
richardpark-msft: There might be a bug, but I wouldn't expect to call into this code _at all_ with a ReceiveAndDelete sender. 

I should just fix that bug instead of you adding workaround code here.

sdk/messaging/azservicebus/message_batch.go
```diff
@@ -150,6 +152,26 @@ func (mb *MessageBatch) addAMQPMessage(tempMsg amqpCompatibleMessage) error {
 	return nil
 }
 
+func (mb *MessageBatch) injectCreationCtx(ctx context.Context, tracer tracing.Tracer, sendSpan tracing.Span) error {
+	mb.mu.Lock()
+	defer mb.mu.Unlock()
+
+	for i, msg := range mb.marshaledMessages {
```
richardpark-msft, line 159: I'll look at this some more - I'd have for the act of adding telemetry to force us to deseralize and then reserialize every message again.

richardpark-msft: In fact, can we just add this instrumentation code into MessageBatch.AddMessage() instead, so each message gets instrumented as they are added?

sdk/messaging/azservicebus/receiver.go
```diff
@@ -205,6 +209,10 @@ func (r *Receiver) ReceiveMessages(ctx context.Context, maxMessages int, options
 		return nil, errors.New("receiver is already receiving messages. ReceiveMessages() cannot be called concurrently")
 	}
 
+	var err error
```
richardpark-msft, line 212: In this case we're actually returning internal.TransformErr(err), not 'err', directly to the client. Is the span intended to match that?

(if so, we can go with the named return values trick again)

sdk/messaging/azservicebus/receiver.go
```diff
@@ -438,21 +455,36 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
 	// at that time
 	if len(result.Messages) == 0 {
 		if internal.IsCancelError(result.Error) || rk == internal.RecoveryKindFatal {
-			return nil, result.Error
+			err = result.Error
```
richardpark-msft, line 458: What do you think about using named return values to (passes out)

richardpark-msft (commented):
Just dropping a few comments. Still reviewing.

sdk/messaging/azservicebus/tracing.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
karenychen: I am open to your preference for this one :)

sdk/messaging/azservicebus/internal/tracing/propagation.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// messageWrapper implements the TextMapCarrier interface for sender side
+type messageWrapper struct {
+	message *amqp.Message
+}
+
+// messageCarrierAdapter wraps a Message so that it implements the propagation.TextMapCarrier interface
+func messageCarrierAdapter(message *amqp.Message) tracing.Carrier {
+	if message == nil {
```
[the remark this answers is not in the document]
karenychen: is it possible for customers to pass in a `nil` message when they settle a message?

We do context extractions there, so I considered that case in the adapter.

sdk/messaging/azservicebus/internal/tracing/constants.go
```diff
@@ -0,0 +1,72 @@
...
+type SpanContext = tracing.SpanContext
+
+const (
+	SpanStatusUnset = tracing.SpanStatusUnset
+	SpanStatusError = tracing.SpanStatusError
+	SpanStatusOK    = tracing.SpanStatusOK
+)
+
+const (
+	AttrServerAddress     = "server.address"
```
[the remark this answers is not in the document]
karenychen: I'd be open to your preference on this piece as well. In the `azcore/tracing` package, we are defining the `Attribute` type as follows:

https://github.com/karenychen/azure-sdk-for-go/blob/9176e582a8dd8c0357a98216c2f38efbf97d3266/sdk/azcore/tracing/tracing.go#L237-L246

`Attribute.Key` is a `string` type. Meaning if we want to use a dedicated attribute type, there needs to be a place where we explicitly cast it back to a string. We can do that when we are adding the attributes, or we can do that in batch before we start the span. However I am not sure if making them into a dedicated type is worth the additional conversions

sdk/messaging/azservicebus/internal/tracing/tracing.go
```diff
@@ -0,0 +1,158 @@
...
+	default:
+		return ""
+	}
+}
+
+func getSpanKind(operationType MessagingOperationType, operationName MessagingOperationName, attrs []Attribute) SpanKind {
+	isBatch := isBatchOperation(attrs)
+	isSession := isSessionOperation(operationName)
+	switch {
+	case operationType == CreateOperationType, operationType == SendOperationType && !isBatch:
```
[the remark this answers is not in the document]
karenychen: I agree it's not very straightforward :( this part is based on the messaging conventions: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind

image

sdk/messaging/azservicebus/receiver.go
karenychen: Let me test this out -- I don't like what `internal.TransformErr` gives me because everything is just `exported.Error` and not very meaningful to the user. I do have a little workaround to unwrap the error in `endSpan` though, so maybe this will just work out

sdk/messaging/azservicebus/message_batch.go
karenychen: That would require us to take in a `context.Context` from the user at `AddMessage()` and therefore changing a public facing API.

But I agree the deseralize -> reserialize part is a bit ugly, if you are down to change the API signature there, we can definitely add it in `AddMessage()`

richardpark-msft: Good point. I think we're going to need to rethink this because you can't add stuff to a message _after_ it's been added to a batch, because it increases the size, which can lead to a valid batch being rejected when sent since now it's grown past it's initial size.

This one we might want to leave out, and I can add it back in when I come up with a solution. We don't need to hold the entire PR for it.

sdk/messaging/azservicebus/tracing.go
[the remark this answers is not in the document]
richardpark-msft: I'm okay with the repetitive code - it's small, and that area is "set in stone", so it won't change.

sdk/messaging/azservicebus/internal/tracing/propagation.go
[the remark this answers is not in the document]
karenychen: I lied, going to remove this chunk as we don't necessarily have to do it in this layer

sdk/messaging/azservicebus/internal/tracing/tracing_test.go
```diff
@@ -0,0 +1,90 @@
...
+package tracing
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/test/tracingvalidator"
+	"github.com/stretchr/testify/require"
+)
+
```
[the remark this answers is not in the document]
karenychen: Good catch! I updated the unit test to cover these

sdk/messaging/azservicebus/receiver.go
karenychen: I lied and will have to keep these in 2 places!

sdk/messaging/azservicebus/message_batch.go
karenychen: Currently I kept the tracing for it but removed the context propagation. So we'd still have a `send` trace, just the message doesn't have an injected context.

sdk/messaging/azservicebus/session_receiver.go
```diff
@@ -266,14 +275,24 @@ func (sr *SessionReceiver) RenewSessionLock(ctx context.Context, options *RenewS
...
+		Tracer:        sr.inner.tracer,
+		OperationName: tracing.RenewSessionLockOperationName,
+	})
 
 	return internal.TransformError(err)
 }
 
 // init ensures the link was created, guaranteeing that we get our expected session lock.
 func (sr *SessionReceiver) init(ctx context.Context) error {
+	var err error
```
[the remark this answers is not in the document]
karenychen: This will cause issues with the `internal.TransformError()` function, basically what we talked about here: https://github.com/Azure/azure-sdk-for-go/pull/23860#discussion_r1985961884

sdk/messaging/azservicebus/internal/utils/retrier_test.go
```diff
@@ -32,7 +32,7 @@ func TestRetrier(t *testing.T) {
 			return nil
 		}, func(err error) bool {
 			panic("won't get called")
-		}, exported.RetryOptions{})
+		}, exported.RetryOptions{}, nil)
```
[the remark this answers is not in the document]
karenychen: Added this to the UT. You should see the SpanStatus/Error to be set to the last attempt!

sdk/messaging/azservicebus/receiver.go
richardpark-msft: I think it is what we want:

1. It's the type the user will ultimately see
2. The error message of the Error actually includes the inner-most error, so we don't lose any information there and it includes the classification code, which user's can use, if they want to process the error in code.

For instance:

```go
err := NewError(CodeConnectionLost, nil)
require.Equal(t, err.Error(), "(connlost): unknown error")

err = NewError(CodeLockLost, &amqp.Error{Condition: amqp.ErrCond("com.microsoft:message-lock-lost")})
require.Equal(t, err.Error(), "(locklost): *Error{Condition: com.microsoft:message-lock-lost, Description: , Info: map[]}")
```

karenychen: update: resolved this in offline convo and got rid of all the bad `var err error`'s

PR #23862, opened 2024-12-11
Fixing ModuleName to be compliant with the standard

richardpark-msft:
Fixing all the outdated projects - ModuleName is intended to be the full module path, not just the folder name.

Also, updated the template project.

### DIFF
--- a/sdk/messaging/eventgrid/azeventgrid/internal/version.go
+++ b/sdk/messaging/eventgrid/azeventgrid/internal/version.go
@@ -6,7 +6,7 @@ package internal
 // Constants to identify the module
 const (
 	// ModuleName is the module name that shows in telemetry.
-	ModuleName = "azeventgrid"
+	ModuleName = "github.com/Azure/azure-sdk-for-go/sdk/messaging/eventgrid/azeventgrid"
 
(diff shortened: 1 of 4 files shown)


richardpark-msft:
@ronniegeraghty, this will affect what we report, as the package name, in our user agent. I'm not sure if we have any reports based on the string.

jhendrixMSFT (approved):

richardpark-msft:
(overrode approval on this - looks like @jhendrixMSFT and I still have some folders that aren't covered by our current CODEOWNERS permissions.

ronniegeraghty:
> @ronniegeraghty, this will affect what we report, as the package name, in our user agent. I'm not sure if we have any reports based on the string.

None that are still working and I don't believe we use the package name. Since our telemetry data will come from the different services data streams they are already divided by service. Event Grid might be interesting though. I don't know if they have separate data streams for event grid basic vs namespaces. The system events package doesn't actually make http requests right? it's just a catalog of the different system event types, right? So we'd never see telemetry from that library.

jhendrixMSFT:
It won't impact the name used in telemetry as we extract the package name (e.g. `azfoo`) from the fully qualified module name.

PR #23894, opened 2024-12-27
Fix DefaultAzureCredential skipping managed identity in Azure Container Instances

### LINKED ISSUE #23890: DefaultAzureCredential in azidentity v1.8.0 rejects Azure Container Instance's IMDS
### Bug Report

- What happened:
  When using `DefaultAzureCredential` in `github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.8.0` on **Azure Container Instance**, it throws error **ManagedIdentityCredential: unexpected response to IMDS probe**.
  Azure Container Instance implements same IMDS token request protocol, but its response to SDK's probe request is different with the authentic IMDS. Authentic IMDS returns `{"error":"invalid_request","error_description":"Required metadata header not specified"}`, but ACI's mock IMDS only returns plain string `Required metadata header not specified or not correct`.
  azidentity v1.7.0 works. Checking whether the response to probe request is a valid JSON is added in #23273.
  Azure SDK should work on Azure Container Instance, by accepting the probe response to be a string, or... coordinate with ACI team so ACI also returns a JSON for probe request.
- How can we reproduce it?
  Deploy this bicep template to get an ACI instance which simulates the probe request sent by azidentity SDK:
```bicep
resource aci 'Microsoft.ContainerInstance/containerGroups@2024-05-01-preview' = {
  name: 'test'
  location: resourceGroup().location
  identity: { type: 'SystemAssigned' }
  properties: {
    sku: 'Standard'
    containers: [
      {
        name: 'test'
        properties: {
          image: 'mcr.microsoft.com/azurelinux/base/core:3.0'
          command: ['curl', '-v', 'http://169.254.169.254/metadata/identity/oauth2/token']
          resources: { requests: { cpu: 1, memoryInGB: 1 } }
        }
      }
    ]
    osType: 'Linux'
    restartPolicy: 'Never'
  }
}
```
  Output: Image
  In comparison, this is the response of an authentic IMDS server, by running the same curl command on a VM: 
```
root [ / ]# curl -v http://169.254.169.254/metadata/identity/oauth2/token
*   Trying 169.254.169.254:80...
* Connected to 169.254.169.254 (169.254.169.254) port 80
> GET /metadata/identity/oauth2/token HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/8.8.0
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 400 Bad Request
< Content-Type: application/json; charset=utf-8
< Server: IMDS/150.870.65.1544
< x-ms-request-id: 3bca7686-9e4e-4911-8092-3571a9ce7ebb
< Date: Wed, 25 Dec 2024 06:21:59 GMT
< Content-Length: 88
<
* Connection #0 to host 169.254.169.254 left intact
{"error":"invalid_request","error_description":"Required metadata header not specified"}
```

chlowell:
Closes #23890 by fixing a regression in #23273, which had DefaultAzureCredential interpret a non-JSON response to its IMDS probe as indicating the server isn't IMDS or a supported imitator (I've filled the test gap that enabled the regression). This behavior is reasonable for IMDS, which always responds with JSON, but incorrect in general because ACI managed identity responds to the probe with a flat string. Sadly, there's no good way to identify ACI as the host, so the fix here is to check for its error message in the response 😞

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -230,16 +230,19 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKi
 			}
 			return azcore.AccessToken{}, newCredentialUnavailableError(credNameManagedIdentity, msg)
 		}
-		// because IMDS always responds with JSON, assume a non-JSON response is from something else, such
-		// as a proxy, and return credentialUnavailableError so DefaultAzureCredential continues iterating
+		// Determine whether the response is from IMDS, a service whose managed identity API imitates IMDS such as Azure
+		// Container Instances (ACI), or something other than a managed identity API. Assume a JSON response is from IMDS.
+		// If the response is not JSON, check for a known ACI error message. If the response is neither JSON nor an ACI
+		// error, assume it's from something like a proxy and return credentialUnavailableError so DefaultAzureCredential
+		// continues to its next credential.
 		b, err := azruntime.Payload(res)
 		if err != nil {
 			return azcore.AccessToken{}, newCredentialUnavailableError(credNameManagedIdentity, fmt.Sprintf("failed to read IMDS probe response: %s", err))
 		}
-		if !json.Valid(b) {
+		if !json.Valid(b) && !strings.HasPrefix(string(b), "Required metadata header") {
 			return azcore.AccessToken{}, newCredentialUnavailableError(credNameManagedIdentity, "unexpected response to IMDS probe")
 		}
-		// send normal token requests from now on because IMDS responded
+		// send normal token requests from now on because IMDS, or something imitating it, responded
 		c.probeIMDS = false
 	}
 
--- a/sdk/azidentity/CHANGELOG.md
+++ b/sdk/azidentity/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bugs Fixed
 * User credential types inconsistently log access token scopes
+* `DefaultAzureCredential` skips managed identity in Azure Container Instances
 
 ### Other Changes
 
--- a/sdk/azidentity/default_azure_credential_test.go
+++ b/sdk/azidentity/default_azure_credential_test.go
@@ -357,6 +357,27 @@ func TestDefaultAzureCredential_IMDS(t *testing.T) {
 				require.NoError(t, err, "DefaultAzureCredential should continue after receiving a non-JSON response from IMDS")
 			}
 		})
+
+		t.Run("Azure Container Instances", func(t *testing.T) {
+			// ensure GetToken returns an error if DefaultAzureCredential skips managed identity
+			before := defaultAzTokenProvider
+			defer func() { defaultAzTokenProvider = before }()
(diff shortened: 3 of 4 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -254,13 +248,21 @@ func (c *managedIdentityClient) authenticate(ctx context.Context, id ManagedIDKi
 	}
 
 	if azruntime.HasStatusCode(resp, http.StatusOK, http.StatusCreated) {
-		return c.createAccessToken(resp)
+		tk, err := c.createAccessToken(resp)
+		if err != nil && c.chained && c.msiType == msiTypeIMDS {
```
jhendrixMSFT, line 252: It's not clear to me why we only want to do this if the credential is part of a chain. If we return the unmarshaling error, is that not surfaced to the caller?

chlowell: The caller gets an error in any case. The type matters though. createAccessToken returns AuthenticationFailedError, and ChainedTokenCredential will stop iterating when it gets one of those. I want the chain to proceed to its next credential, so I return credentialUnavailableError instead when this credential's part of a chain. I'm afraid to do that unconditionally because someone may have code like this:
```go
cred, _ := azidentity.NewManagedIdentityCredential(nil)
client := azfoo.New(cred)
err := client.Frobulate()
if err != nil {
	var e *azidentity.AuthenticationFailedError
	if errors.As(err, &e) {
		// ...
	}
}
```

jhendrixMSFT: So if `authenticate()` returns an arbitrary error, and that error isn't a `credentialUnavailableError`, it's surfaced as an `AuthenticationFailedError`, is that right?

chlowell: Yes, that happens elsewhere:
https://github.com/Azure/azure-sdk-for-go/blob/bf1e31b5690cbd12f8bfdc5a7926782c4c7e84eb/sdk/azidentity/confidential_client.go#L114-L115

sdk/azidentity/managed_identity_client.go
```diff
@@ -60,7 +60,10 @@ const (
 )
 
 type managedIdentityClient struct {
-	azClient  *azcore.Client
+	azClient *azcore.Client
+	// chained indicates whether the client is part of a credential chain. If true, the client will return
+	// a credentialUnavailableError instead of an AuthenticationFailedError for an unexpected IMDS response.
+	chained   bool
```
jhendrixMSFT, line 66: I must say that changing the error type based on the calling context makes me a bit nervous, but maybe I'm overthinking it. If there isn't a cleaner way to do this, I can live with it though (or if we already have this pattern elsewhere).

This should be moved to after `probeIMDS` though in order to avoid some unnecessary padding in the struct layout (we need to lint for this at some point as I'm sure there are other cases in the codebase).

jhendrixMSFT (approved):

PR #23926, opened 2025-01-08
[azcosmos] Add LIMITED cross-partition query support

analogrelay:
This PR adds very limited support for cross-partition queries to the Cosmos Go SDK.

It works by _always_ adding the `enablecrosspartitionquery` header to the query request. We didn't want to make a separate option because in the future we're hoping to have richer support for cross partition queries that can be lit up automatically.

> [!IMPORTANT]
> The cross-partition query functionality added here is **limited** to cross-partition queries that can be successfully executed against the Gateway API. See https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway for more details. The doc comments describe this limitation.

I'm hoping the doc comments largely explain things, so I don't want to get too deep here, if you have questions, please ask them and I can clarify the doc comments! Essentially, the `partitionKey` parameter in `NewQueryItemsPager` is now optional and can be omitted by passing `azcosmos.NewPartitionKey()`, which creates an "empty" partition key (the interior array of values is non-nil, but empty). This was the easiest way to allow users to omit this parameter. We couldn't just accept `nil` here, because the `partitionKey` parameter is not a pointer, it's a concrete type.

I also reworked the `emulator_cosmos_query_test` tests to be a little more readable (IMO) by using `stretchr/testify`, which we were already using in some other test files. It provides the `assert` package which has more readable test assertions (again, IMO). I added some cross-partition tests there to validate behavior against the emulator.

### DIFF
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -498,7 +498,7 @@ func (c *ContainerClient) DeleteItem(
 // If the query does not filter down to a single partition (i.e. it does not filter on partition key at all, or filters on only some of the partition keys a container defines), the query will be executed as a cross partition query.
 // The Azure Cosmos DB Gateway API, used by the Go SDK, can only perform a LIMITED set of cross-partition queries.
 // Specifically, the gateway can only perform simple projections and filtering on cross partition queries.
-// See https://learn.microsoft.com/en-us/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway for more details.
+// See https://learn.microsoft.com/rest/api/cosmos-db/querying-cosmosdb-resources-using-the-rest-api#queries-that-cannot-be-served-by-gateway for more details.
 //
 // When performing a cross-partition query, the Gateway may return pages of inconsistent size, or even empty pages (while still having a non-nil continuation token).
 // Ensure you fully iterate the pager, even if you receive empty pages, to ensure you get all results.
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -490,7 +490,7 @@ func (c *ContainerClient) DeleteItem(
 // o - Options for the operation.
 //
 // Limited cross partition queries ARE possible with the Go SDK.
-// You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will indicate the partitions to query.
+// You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will specify which partitions to query.
 // If you specify partition keys in the `partitionKey` parameter, you must specify ALL partition keys that the container has (in the case of hierarchical partitioning).
 //
 // If the query itself contains WHERE clauses that filter down to a single partition, the query will be executed on that partition.
--- a/sdk/data/azcosmos/cosmos_container.go
+++ b/sdk/data/azcosmos/cosmos_container.go
@@ -489,8 +489,9 @@ func (c *ContainerClient) DeleteItem(
 // partitionKey - The partition key to scope the query on. See below for more information on cross partition queries.
 // o - Options for the operation.
 //
-// Limited cross partition queries ARE possible with the Go SDK.
 // You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will specify which partitions to query.
+//
+// Limited cross partition queries ARE possible with the Go SDK.
 // If you specify partition keys in the `partitionKey` parameter, you must specify ALL partition keys that the container has (in the case of hierarchical partitioning).
 //
 // If the query itself contains WHERE clauses that filter down to a single partition, the query will be executed on that partition.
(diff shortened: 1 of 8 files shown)


sdk/data/azcosmos/cosmos_container.go
```diff
@@ -486,8 +486,20 @@ func (c *ContainerClient) DeleteItem(
 
 // NewQueryItemsPager executes a single partition query in a Cosmos container.
 // query - The SQL query to execute.
-// partitionKey - The partition key to scope the query on.
+// partitionKey - The partition key to scope the query on. See below for more information on cross partition queries.
 // o - Options for the operation.
+//
+// Limited cross partition queries ARE possible with the Go SDK.
+// You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will indicate the partitions to query.
```
analogrelay, line 493:
```suggestion
// You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will specify which partitions to query.
```

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -486,8 +486,20 @@ func (c *ContainerClient) DeleteItem(
 
 // NewQueryItemsPager executes a single partition query in a Cosmos container.
 // query - The SQL query to execute.
-// partitionKey - The partition key to scope the query on.
+// partitionKey - The partition key to scope the query on. See below for more information on cross partition queries.
 // o - Options for the operation.
+//
+// Limited cross partition queries ARE possible with the Go SDK.
+// You can specify an empty list of partition keys by passing `NewPartitionKey()` to the `partitionKey` parameter, to indicate that the query WHERE clauses will specify which partitions to query.
```
kirankumarkolli, line 493: nit: How about swap below two lines, that way all in-side query context will be co-located.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -486,8 +486,20 @@ func (c *ContainerClient) DeleteItem(
...
+// If you specify partition keys in the `partitionKey` parameter, you must specify ALL partition keys that the container has (in the case of hierarchical partitioning).
+//
+// If the query itself contains WHERE clauses that filter down to a single partition, the query will be executed on that partition.
+// If the query does not filter down to a single partition (i.e. it does not filter on partition key at all, or filters on only some of the partition keys a container defines), the query will be executed as a cross partition query.
+// The Azure Cosmos DB Gateway API, used by the Go SDK, can only perform a LIMITED set of cross-partition queries.
+// Specifically, the gateway can only perform simple projections and filtering on cross partition queries.
```
kirankumarkolli, line 499: Cross-partition execution by gateway might yield with no-results but continuation, does this needs to be documented explicitly?

kirankumarkolli (approved):

sdk/data/azcosmos/cosmos_container.go
Pilchie: Good idea - we know this is a frequent source of confusion.

Pilchie (approved):

sdk/data/azcosmos/cosmos_container.go
analogrelay: I didn't know that! I've included information on that. I'm also assuming that a user should also expect somewhat inconsistent page sizes, since I assume the reason for the empty pages with continuation is because of how the gateway is querying each partition separately and merging results. Even if that's not entirely true, I think the advice to be prepared for inconsistent page sizes is still helpful.

Pilchie (approved):

analogrelay:
As a bit of a last-minute rethink, I added an option to `QueryOptions` to _disable_ the use of the cross-partition query header. In my testing, adding this header seems to be harmless and does not affect a single-partition query, but I'm a little nervous I've missed something and felt it would be good to give customers the option to disable this entirely if necessary.

With this PR, our default is **still** to **enable cross-partition queries** by default. The option *defaults* to `true` and users should generally not need to do anything. However, if they want to, they can disable the setting using code like this:

```go
enableCrossPartition := false
pager := container.NewQueryItemsPager("...", ..., &azcosmos.QueryOptions{EnableCrossPartitionQuery:&enableCrossPartition})
```

sdk/data/azcosmos/cosmos_query_request_options.go
```diff
@@ -39,6 +39,11 @@ type QueryOptions struct {
 	QueryParameters []QueryParameter
 	// Options for operations in the dedicated gateway.
 	DedicatedGatewayRequestOptions *DedicatedGatewayRequestOptions
+	// EnableCrossPartitionQuery configures the behavior of the query engine when executing queries.
+	// If set to true, the query engine will set the 'x-ms-documentdb-query-enablecrosspartition' header to true for cross-partition queries.
+	// If set to false, cross-partition queries will be rejected.
+	// The default value, if this is not set, is TRUE.
```
Pilchie, line 45: Nit: should this be `TRUE` or `true`?

Pilchie (approved):

sdk/data/azcosmos/cosmos_query_request_options.go
analogrelay: Probably just `true`, yeah.

rs-malik:
Thank you @analogrelay for this PR!

My expectation is that the queries which are not supported by the gateway will fail (400). Is this expectation correct?

zackrossman:
Thank you @analogrelay !

analogrelay:
> My expectation is that the queries which are not supported by the gateway will fail (400). Is this expectation correct?

@rs-malik Yes, this is correct. You'll only be able to perform simple projections and filtering with this API.

PR #23933, opened 2025-01-11
[azservicebus] Unexpose feature that's not yet public.

richardpark-msft:


jhendrixMSFT (approved):

richardpark-msft:
Overriding check-enforcer as I accidentally ran an incorrect pipeline for live testing that doesn't apply here.

PR #24011, opened 2025-01-25
Perf: Use WalkDir for optimized directory traversal

Gofastasf:
filepath.Walk calls os.Lstat for every file or directory to retrieve os.FileInfo, which can be slower. filepath.WalkDir avoids unnecessary system calls since it provides a fs.DirEntry, which includes file type information without requiring a stat call.

### DIFF
--- a/eng/tools/doccheck/main.go
+++ b/eng/tools/doccheck/main.go
@@ -21,7 +21,7 @@ func filter(f fs.FileInfo) bool {
 func findAllSubDirectories(root string) []string {
 	var ret []string
 
-	filepath.Walk(root, func(path string, info fs.FileInfo, err error) error {
+	filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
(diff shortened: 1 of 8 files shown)


Gofastasf:

Gofastasf:
There are some deprecated calls like ioutils.

And some general improvements that can be done like 
1. using filepath.Ext instead of strings.HasSuffix 
2. using filepath.ToSlash instead of strings.Replace
3. d.Name() instead of strings.Contains.

Should i send another PR or include on this one ?

Gofastasf:
@tadelesh Hey can you checkout this PR :)

benbp:
@jhendrixMSFT are you the right person to review?

jhendrixMSFT (approved):

benbp (approved):
Approving as `eng/` codeowner. @jhendrixMSFT I'll let you merge.

Thanks for the contribution @Gofastasf

PR #24039, opened 2025-02-01
Fix: make it OS compatible.

Gofastasf:
To make it cross platform

Use filepath.Ext instead of strings.HasSuffix for file extensions.
Use filepath.ToSlash instead of strings.Replace.
Use path/filepath methods for file system paths instead of path package

### DIFF
--- a/sdk/storage/azfile/directory/client.go
+++ b/sdk/storage/azfile/directory/client.go
@@ -11,7 +11,6 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
-	"path/filepath"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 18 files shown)


Gofastasf:
@jhendrixMSFT Hey there seems to be some CI issue. Can you check this out?

sdk/storage/azfile/directory/client.go
```diff
@@ -103,7 +104,7 @@ func NewClientFromConnectionString(connectionString string, shareName string, di
 		return nil, err
 	}
 
-	directoryPath = strings.ReplaceAll(directoryPath, "\\", "/")
+	directoryPath = filepath.ToSlash(directoryPath)
```
jhendrixMSFT, line 107: This has introduced a regression and is one of the reasons for failures in CI. In this case, [the test](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azfile/directory/client_test.go#L197-L200) intentionally uses `\` chars for directory separators expecting them to be replaced but they aren't on Linux. I suspect the other failures are something similar.

Can you please revert these changes from the storage packages?

jhendrixMSFT (changes requested):

jhendrixMSFT (approved):

Gofastasf:
Closing this as the other PR does this.

PR #24047, opened 2025-02-03
Conditionally allow and test connection strings (only in TME)

richardpark-msft:
Local resources can no longer contain any connection strings so we need to enable it conditionally. 

A change in New-TestResources.ps1 will pass the flag, so this is just making the change ahead of time and is okay to merge.


richardpark-msft (commented):
/azp run go - azeventhubs

jhendrixMSFT (approved):

PR #24057, opened 2025-02-05
[Storage] [AzFile] NFS over REST STG 97

tanyasethi-msft:

### DIFF
--- a/sdk/storage/azfile/directory/models.go
+++ b/sdk/storage/azfile/directory/models.go
@@ -18,9 +18,6 @@ import (
 // SharedKeyCredential contains an account's name and its primary or secondary key.
 type SharedKeyCredential = exported.SharedKeyCredential
 
-// NfsFileType specifies the type of the file or directory.
-type NfsFileType = generated.NfsFileType
-
 // NewSharedKeyCredential creates an immutable SharedKeyCredential containing the
 // storage account's name and either its primary or secondary key.
 func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredential, error) {
@@ -52,8 +49,6 @@ type CreateOptions struct {
 	Owner *string
 	// NFS only. The owning group of the file or directory.
 	Group *string
-	// NFS only. Type of the file or directory.
-	NfsFileType *NfsFileType
 }
 
 func (o *CreateOptions) format() *generated.DirectoryClientCreateOptions {
@@ -186,8 +181,8 @@ type SetPropertiesOptions struct {
 	Owner *string
 	// NFS only. The owning group of the file or directory.
 	Group *string
-	// NFS only. Type of the file or directory.
-	NfsFileType *NfsFileType
+	//// NFS only. Type of the file or directory.
+	//NFSFileType *NFSFileType
 }
 
 func (o *SetPropertiesOptions) format() *generated.DirectoryClientSetPropertiesOptions {
--- a/sdk/storage/azfile/directory/models.go
+++ b/sdk/storage/azfile/directory/models.go
@@ -181,8 +181,6 @@ type SetPropertiesOptions struct {
 	Owner *string
 	// NFS only. The owning group of the file or directory.
 	Group *string
-	//// NFS only. Type of the file or directory.
-	//NFSFileType *NFSFileType
 }
 
 func (o *SetPropertiesOptions) format() *generated.DirectoryClientSetPropertiesOptions {
--- a/sdk/storage/azfile/directory/models.go
+++ b/sdk/storage/azfile/directory/models.go
@@ -18,6 +18,9 @@ import (
 // SharedKeyCredential contains an account's name and its primary or secondary key.
 type SharedKeyCredential = exported.SharedKeyCredential
 
+// NfsFileType specifies the type of the file or directory.
+type NfsFileType = generated.NfsFileType
+
 // NewSharedKeyCredential creates an immutable SharedKeyCredential containing the
 // storage account's name and either its primary or secondary key.
 func NewSharedKeyCredential(accountName, accountKey string) (*SharedKeyCredential, error) {
@@ -43,6 +46,14 @@ type CreateOptions struct {
 	// binary, the permission is returned as a base64 string representing the binary
 	// encoding of the permission
 	FilePermissionFormat *FilePermissionFormat
+	// NFS only. The file mode of the file or directory
+	FileMode *string
+	// NFS only. The owner of the file or directory.
+	Owner *string
+	// NFS only. The owning group of the file or directory.
+	Group *string
+	// NFS only. Type of the file or directory.
+	NfsFileType *NfsFileType
 }
 
 func (o *CreateOptions) format() *generated.DirectoryClientCreateOptions {
@@ -60,6 +71,9 @@ func (o *CreateOptions) format() *generated.DirectoryClientCreateOptions {
 		FilePermission:    permission,
 		FilePermissionKey: permissionKey,
 		Metadata:          o.Metadata,
+		FileMode:          o.FileMode,
+		Owner:             o.Owner,
+		Group:             o.Group,
 	}
 
 	if permissionKey != nil && *permissionKey != shared.DefaultFilePermissionString {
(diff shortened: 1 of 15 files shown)


sdk/storage/azfile/directory/models.go
```diff
@@ -18,6 +18,9 @@ import (
 // SharedKeyCredential contains an account's name and its primary or secondary key.
 type SharedKeyCredential = exported.SharedKeyCredential
 
+// NfsFileType specifies the type of the file or directory.
+type NfsFileType = generated.NfsFileType
```
souravgupta-msft, line 22: Suggestion, should we rename this to `NFSFileType`?
Types having similar naming pattern for reference, https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azfile/file/models.go#L30-L34

sdk/storage/azfile/directory/models.go
```diff
@@ -43,22 +46,23 @@ type CreateOptions struct {
...
 	// encoding of the permission
 	FilePermissionFormat *FilePermissionFormat
+	// NFS only. The file mode of the file or directory
+	FileMode *string
+	// NFS only. The owner of the file or directory.
+	Owner *string
+	// NFS only. The owning group of the file or directory.
+	Group *string
+	// NFS only. Type of the file or directory.
+	NfsFileType *NfsFileType
```
souravgupta-msft, line 56: Same as previous. If changing, please change at all places

sdk/storage/azfile/directory/models.go
```diff
@@ -67,6 +71,9 @@ func (o *CreateOptions) format() *generated.DirectoryClientCreateOptions {
 		FilePermission:    permission,
 		FilePermissionKey: permissionKey,
 		Metadata:          o.Metadata,
+		FileMode:          o.FileMode,
+		Owner:             o.Owner,
+		Group:             o.Group,
```
souravgupta-msft, line 76: I don't see `NFSFileType` being set here

sdk/storage/azfile/directory/models.go
```diff
@@ -197,6 +205,9 @@ func (o *SetPropertiesOptions) format() *generated.DirectoryClientSetPropertiesO
 		FileLastWriteTime: fileLastWriteTime,
 		FilePermission:    permission,
 		FilePermissionKey: permissionKey,
+		FileMode:          o.FileMode,
+		Owner:             o.Owner,
+		Group:             o.Group,
```
souravgupta-msft, line 210: Same as previous, `NFSFileType` is not being set here, though it is part of the options

sdk/storage/azfile/file/models.go
```diff
@@ -786,6 +813,18 @@ func (o *ForceCloseHandlesOptions) format() *generated.FileClientForceCloseHandl
 	}
 }
 
+// CreateHardLinkOptions contains the optional parameters for the Client.CreateHardLink method.
+type CreateHardLinkOptions struct {
+	// NFS only. Required. Specifies the path of the target file to which the link will be created, up to 2 KiB in length.
+	TargetFile string
```
souravgupta-msft, line 819: If `TargetFile` is a required parameter, we should have this in the method argument and not in options bag

sdk/storage/azfile/file/client.go
```diff
@@ -326,6 +326,17 @@ func (f *Client) ListHandles(ctx context.Context, options *ListHandlesOptions) (
 	return resp, err
 }
 
+// CreateHardLink operation returns a list of open handles on a file.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/list-handles.
+func (f *Client) CreateHardLink(ctx context.Context, options *CreateHardLinkOptions) (CreateHardLinkResponse, error) {
+	if options == nil {
+		return CreateHardLinkResponse{}, errors.New("targetFile cannot be nil")
```
souravgupta-msft, line 333: If target file is mandatory parameter, it should be in the method argument. In this case, the check will fail if user passes non-nil options where target file is empty string.

sdk/storage/azfile/directory/models.go
tanyasethi-msft: Removed from few unwanted places, and Renamed others.

sdk/storage/azfile/directory/models.go
tanyasethi-msft: Removed from few unwanted places and renamed others.

sdk/storage/azfile/directory/models.go
tanyasethi-msft: This seems like a default argument - with value only being "Regular".  if not set, service will automatically set it. 
So removing from here for now.

sdk/storage/azfile/directory/models.go
tanyasethi-msft: This seems like a default argument - with value only being "Regular".  if not set, service will automatically set it. 
So removing this argument for now.

sdk/storage/azfile/file/models.go
tanyasethi-msft: Changed, thanks

sdk/storage/azfile/file/client.go
```diff
@@ -326,6 +326,17 @@ func (f *Client) ListHandles(ctx context.Context, options *ListHandlesOptions) (
 	return resp, err
 }
 
+// CreateHardLink operation returns a list of open handles on a file.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/list-handles.
```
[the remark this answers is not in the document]
tanyasethi-msft: This feature is not yet GA'ed - so the exact link is not available yet. I have changed it to the best of my knowledge as of now - Will change it before sdk GA if needed.

sdk/storage/azfile/file/models.go
```diff
@@ -786,6 +810,16 @@ func (o *ForceCloseHandlesOptions) format() *generated.FileClientForceCloseHandl
...
 }
 
+// CreateHardLinkOptions contains the optional parameters for the Client.CreateHardLink method.
+type CreateHardLinkOptions struct {
+	// LeaseAccessConditions contains optional parameters to access leased entity.
+	LeaseAccessConditions *LeaseAccessConditions
+}
+
+func (o *CreateHardLinkOptions) format() (*generated.FileClientCreateHardLinkOptions, *generated.LeaseAccessConditions) {
+	return &generated.FileClientCreateHardLinkOptions{}, o.LeaseAccessConditions
```
souravgupta-msft, line 820: Add a check if `o` is nil or not. Otherwise, it can result in panic in `o.LeaseAccessConditions`

sdk/storage/azfile/file/constants.go
```diff
@@ -41,6 +41,28 @@ func PossibleCopyStatusTypeValues() []CopyStatusType {
...
+type ModeCopyMode = generated.ModeCopyMode
+
+// OwnerCopyMode specifies the copy mode source or override.
+type OwnerCopyMode = generated.OwnerCopyMode
+
+const (
+	// NFSFileTypeRegular Default and only value for the parameter NFS File Type.
+	NFSFileTypeRegular   NFSFileType = generated.NfsFileTypeRegular
+	NfsFileTypeDirectory NFSFileType = generated.NfsFileTypeDirectory
+	NfsFileTypeSymlink   NFSFileType = generated.NfsFileTypeSymLink
```
souravgupta-msft, line 57: Just to be in sync with the above type
```suggestion
	NFSFileTypeDirectory NFSFileType = generated.NfsFileTypeDirectory
	NFSFileTypeSymlink   NFSFileType = generated.NfsFileTypeSymLink
```

tanyasethi-msft: Changed

sdk/storage/azfile/file/models.go
tanyasethi-msft: Added, missed this previously.

sdk/storage/azfile/file/client.go
```diff
@@ -326,6 +326,14 @@ func (f *Client) ListHandles(ctx context.Context, options *ListHandlesOptions) (
 	return resp, err
 }
 
+// CreateHardLink operation returns a list of open handles on a file.
+// For more information, see https://learn.microsoft.com/en-us/rest/api/storageservices/create-hard-link.
+func (f *Client) CreateHardLink(ctx context.Context, targetFile string, options *CreateHardLinkOptions) (CreateHardLinkResponse, error) {
```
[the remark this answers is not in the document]
tanyasethi-msft: I tried this, the service is sending a 4xx in response for this. So, not added an extra check for this - Added a test for the same.

sdk/storage/azfile/directory/models.go
```diff
@@ -174,21 +175,23 @@ type SetPropertiesOptions struct {
...
 	// FilePermissionFormat contains the format of the file permissions, Can be sddl (Default) or Binary.
 	FilePermissionFormat *FilePermissionFormat
+	// NFS only. The file mode of the file or directory
+	FileMode *string
+	// NFS only. The owner of the file or directory.
+	Owner *string
+	// NFS only. The owning group of the file or directory.
+	Group *string
+	//// NFS only. Type of the file or directory.
+	//NFSFileType *NFSFileType
```
souravgupta-msft, line 185: Remove this if not needed

tanyasethi-msft: removed, thanks

souravgupta-msft (approved):

PR #24069, opened 2025-02-07
[azcore/tracing] Add propagator and link wrapper

karenychen:
Adding Propagator and Links to `azcore/tracing`. This will allow context propagation and span links in the SDK.

### DIFF
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Release History
 
-## 1.18.0 (Unreleased)
+## 1.18.0-beta.1 (Unreleased)
 
 ### Features Added
 
+* Added field `NewPropagatorFn` to `tracing.ProviderOptions` to support custom propagators.
 * Added type `SpanContext` to the `tracing` package to support Span Context.
 * Added method `SpanContext()` to `tracing.Span` to support Span Context.
 * Added func `NewSpanContext()` to the `tracing` package to support Span Context.
@@ -11,12 +12,10 @@
 * Added method `LinkFromContext()` to `tracing.Tracer` to support Span Links.
 * Added method `AddLink()` to `tracing.Span` to support Span Links.
 * Added type `Propagator` to the `tracing` package to support context propagation.
-* Added interface `Carrier` to the `tracing` package to support context propagation.
+* Added type `Carrier` to the `tracing` package to support context propagation.
 * Added method `NewPropagator()` to type `tracing.Provider` to support context propagation.
 
 ### Breaking Changes
-
-* Method `tracing.NewProvider` now requires an argument of type `func() Propagator`.
 
 ### Bugs Fixed
 
--- a/sdk/azcore/CHANGELOG.md
+++ b/sdk/azcore/CHANGELOG.md
@@ -1,10 +1,22 @@
 # Release History
 
-## 1.17.1 (Unreleased)
+## 1.18.0 (Unreleased)
 
 ### Features Added
 
+* Added type `SpanContext` to the `tracing` package to support Span Context.
+* Added method `SpanContext()` to `tracing.Span` to support Span Context.
+* Added func `NewSpanContext()` to the `tracing` package to support Span Context.
+* Added type `Link` to the `tracing` package to support Span Links.
+* Added method `LinkFromContext()` to `tracing.Tracer` to support Span Links.
+* Added method `AddLink()` to `tracing.Span` to support Span Links.
+* Added type `Propagator` to the `tracing` package to support context propagation.
+* Added interface `Carrier` to the `tracing` package to support context propagation.
+* Added method `NewPropagator()` to type `tracing.Provider` to support context propagation.
+
 ### Breaking Changes
+
+* Method `tracing.NewProvider` now requires an argument of type `func() Propagator`.
 
 ### Bugs Fixed
 
--- a/sdk/azcore/core_test.go
+++ b/sdk/azcore/core_test.go
@@ -150,7 +150,7 @@ func TestNewClientTracingEnabled(t *testing.T) {
 				}
 				return ctx, tracing.Span{}
 			}, nil)
-		}, nil),
+		}, nil, nil),
 		Transport: srv,
 	})
 	require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


sdk/azcore/core_test.go
karenychen, line 1: @jhendrixMSFT Since we haven't added context propagation in the http requests, I left the `newPropagatorFn` here to nil. But please do let me know if you prefer it to be set :)

sdk/azcore/tracing/tracing.go
```diff
@@ -189,3 +237,159 @@ type Attribute struct {
...
+// TraceID is a unique identity of a trace.
+type TraceID [16]byte
+
+// TraceFlags contains flags that can be set on a SpanContext.
+type TraceFlags byte
+
+// TraceStateImpl contains the implementation for TraceState.
+type TraceStateImpl struct {
+	// String contains the implementation for the TraceState.String method.
+	String func() string
```
karenychen, line 266: There are more methods to the Otel implementation for TraceState, but `String()` is all we need at the moment to parse it to use the [`ParseTraceState()`](https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.33.0#ParseTraceState) function

sdk/azcore/tracing/tracing.go
```diff
@@ -19,16 +19,18 @@ type ProviderOptions struct {
 // NewProvider creates a new Provider with the specified values.
 //   - newTracerFn is the underlying implementation for creating Tracer instances
 //   - options contains optional values; pass nil to accept the default value
-func NewProvider(newTracerFn func(name, version string) Tracer, options *ProviderOptions) Provider {
+func NewProvider(newTracerFn func(name, version string) Tracer, newPropagatorFn func() Propagator, options *ProviderOptions) Provider {
```
jhendrixMSFT, line 22: This is a breaking change and is what CI is complaining about. You'll need to put `newPropagatorFn` in `ProviderOptions`.

sdk/azcore/tracing/tracing.go
```diff
@@ -189,3 +237,159 @@ type Attribute struct {
...
+	if p.impl.Fields != nil {
+		return p.impl.Fields()
+	}
+	return nil
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Carrier is the storage medium used by the Propagator.
+type Carrier interface {
```
jhendrixMSFT, line 388: Interfaces are brittle and don't version well (i.e. adding a method to an interface is a breaking change), so this makes me nervous. It's also why we use this struct/func pattern in this package as we can add new `func` fields in a non-breaking way.

karenychen: That's a good point, I will update to make this a struct

sdk/azcore/CHANGELOG.md
```diff
@@ -1,11 +1,23 @@
 # Release History
 
-## 1.17.1 (Unreleased)
+## 1.18.0 (Unreleased)
```
jhendrixMSFT, line 3: We should release this as `1.18.0-beta.1` so we have a chance to fix things before it GAs.

sdk/azcore/tracing/tracing.go
jhendrixMSFT: Makes sense. Let's just start with the methods we know we need (we can always add more later).

sdk/azcore/CHANGELOG.md
jhendrixMSFT: You'll also need to update the version `const` [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/shared/constants.go#L43).

karenychen: updated!

sdk/azcore/tracing/tracing.go
lmolkova: we don't  really need to modify or parse it - all we need is to propagate the string as is. I wonder it it's possible not to expose any of the TraceState-related features? We are able to propagate the context in Java and other languages without ever bothering with the tracestate.

jhendrixMSFT: @karenychen is it possible to remove tracestate per @lmolkova comment?

sdk/azcore/tracing/tracing.go
```diff
@@ -13,22 +13,32 @@ import (
...
 
 // NewProvider creates a new Provider with the specified values.
 //   - newTracerFn is the underlying implementation for creating Tracer instances
 //   - options contains optional values; pass nil to accept the default value
 func NewProvider(newTracerFn func(name, version string) Tracer, options *ProviderOptions) Provider {
+	if options == nil {
+		options = &ProviderOptions{}
+	}
+	if options.NewPropagatorFn == nil {
+		options.NewPropagatorFn = func() Propagator { return Propagator{} }
```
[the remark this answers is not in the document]
karenychen: updated!

sdk/azcore/tracing/tracing.go
karenychen: @lmolkova @jhendrixMSFT I changed `TraceState` into a simple string. However I don't think we can remove it from `SpanContextConfig` -- we will need it to convert between the wrapper `SpanContext` and the OTel `SpanContext`.

I cleaned up the azotel changes for us to see this better, but please do let me know if there is a better way we can do this: https://github.com/Azure/azure-sdk-for-go/blob/66f2c3bf70edf30fce250508b48a5a88570fca8a/sdk/tracing/azotel/otel.go#L141

lmolkova (approved):
LGTM from general tracing perspective. Thanks for putting it together!

sdk/azcore/tracing/tracing.go
```diff
@@ -106,13 +127,25 @@ func (t Tracer) SpanFromContext(ctx context.Context) Span {
 	return Span{}
 }
 
+// LinkFromContext returns a link encapsulating the SpanContext of the current context.
+// If the provided context has no Span, an empty Link is returned.
+func (t Tracer) LinkFromContext(ctx context.Context, attrs ...Attribute) Link {
```
jhendrixMSFT, line 132: What happens with `attrs` here, are they added to the returned `SpanContext`? Might be good to document how this arg is used.

jhendrixMSFT (approved):

sdk/azcore/tracing/tracing.go
karenychen: Added a short description to what attrs are used for

Issue #23650, opened 2024-10-23
Can we improve code coverage reporting in CI


jhendrixMSFT:
Changes in code coverage reporting in the latest two versions of Go impact total percentages. It's likely related to our usage of `./...` which causes all packages to be included. While this is useful for ensuring that all tests are executed, it has a direct impact on how CC numbers are generated.

For example, consider `azcore/internal/exported.HasStatusCode`. This is exported in `azcore/runtime` and there's a test for it there. However, since we use `./...` _both_ versions are included in CC numbers which means we need to have a test in _both_ packages (i.e. redundant tests) else we'll get a zero for the internal version, dragging down CC numbers.

I believe we can fix this with the `-coverpkg` flag however it's not clear what that looks like yet.

jhendrixMSFT:
See https://github.com/golang/go/issues/68627 which links various issues.

PR #24080, opened 2025-02-10
[Storage] stg97 features

tanyasethi-msft:

### DIFF
--- a/eng/config.json
+++ b/eng/config.json
@@ -42,7 +42,7 @@
         },
         {
             "Name": "azfile",
-            "CoverageGoal": 0.75
+            "CoverageGoal": 0.70
         },
         {
             "Name": "azopenai",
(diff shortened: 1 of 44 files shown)


eng/config.json
```diff
@@ -42,7 +42,7 @@
         },
         {
             "Name": "azfile",
-            "CoverageGoal": 0.75
+            "CoverageGoal": 0.70
```
vibhansa-msft, line 45: Any specific reason to lower the coverage goals?

tanyasethi-msft: Yes, the code coverage numbers are calculated differently after https://github.com/golang/go/issues/65570 was fixed in the latest two versions of Go. if you compare before/after updating the Go versions the numbers are quite different.
The percentages in the internal content are now being included. See https://github.com/Azure/azure-sdk-for-go/issues/23650 for more details. 
Till this is fixed, the coverage goal has been changed, we will come back to the original value.

tanyasethi-msft:
This PR contains changes from already approved PRs

jhendrixMSFT (approved):

PR #24081, opened 2025-02-10
[tracing/azotel] Enable span link and propagation support

karenychen:

### Description

Support span link and context propagation in `tracing/azotel`. 

The PR adds OTel implementation to:
* `Provider.NewPropagator()` to return a `propagation.TraceContext()` propagator for context propagation.
* `tracing.LinkFromContext()` to return a OTel Span Link with the passed in attributes attached to it.
* `Span.AddLink()` to add an Otel Span Link to the current span.
* `Span.SpanContext()` to return the OTel Span Context of the current span.

There is no public facing API change in this PR.

### DIFF
--- a/sdk/tracing/azotel/otel.go
+++ b/sdk/tracing/azotel/otel.go
@@ -21,7 +21,8 @@ import (
 
 // TracingProviderOptions contains the optional values for NewTracingProvider.
 type TracingProviderOptions struct {
-	// for future expansion
+	// propagator is the OpenTelemetry propagator to use. If nil, propagation.TraceContext is used.
+	propagator propagation.TextMapPropagator
 }
 
 // NewTracingProvider creates a new tracing.Provider that wraps the specified OpenTelemetry TracerProvider.
@@ -56,6 +57,9 @@ func NewTracingProvider(tracerProvider trace.TracerProvider, opts *TracingProvid
 		})
 	}, &tracing.ProviderOptions{
 		NewPropagatorFn: func() tracing.Propagator {
+			if opts != nil && opts.propagator != nil {
+				return convertPropagator(opts.propagator)
+			}
 			return convertPropagator(propagation.TraceContext{})
 		},
 	})
--- a/sdk/tracing/azotel/otel.go
+++ b/sdk/tracing/azotel/otel.go
@@ -21,9 +21,6 @@ import (
 
 // TracingProviderOptions contains the optional values for NewTracingProvider.
 type TracingProviderOptions struct {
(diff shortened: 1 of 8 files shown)


sdk/tracing/azotel/otel.go
```diff
@@ -14,13 +14,16 @@ import (
...
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
 // TracingProviderOptions contains the optional values for NewTracingProvider.
 type TracingProviderOptions struct {
-	// for future expansion
+	// propagator is the OpenTelemetry Propagator to use.
+	// If not set, a propagation.TraceContext Propagator is used.
+	propagator propagation.TextMapPropagator
```
jhendrixMSFT, line 26: This will need to be exported so that callers can set it.

What's the default behavior of `propagation.TraceContext`, is it a no-op propagator?

karenychen: Oops, thanks for catching that!

[`propagation.TraceContext`(https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/trace_context.go) is OTel's implementation of W3C trace context propagator. 
 
From our offline chat, @lmolkova mentioned OTel gives user an ability to customize propagator, but we (at least for now) ignore it in other languages and use the W3C trace context implementation. Therefore we are setting it as the default.

jhendrixMSFT: If we ignore it in other languages, then do we even need to expose this (we can always add it later)?

lmolkova: I agree that we don't need to expose it until there is a need to support custom propagators (which should then be done across all languages).

sdk/tracing/azotel/otel.go
```diff
@@ -14,13 +14,16 @@ import (
...
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
 // TracingProviderOptions contains the optional values for NewTracingProvider.
 type TracingProviderOptions struct {
-	// for future expansion
+	// Propagator is the OpenTelemetry propagator to use.
+	// If not set, a propagation.TraceContext propagator is used.
+	Propagator propagation.TextMapPropagator
```
jhendrixMSFT, line 26: Per our conversation, it sounds like this will be removed?

karenychen: Yes, I have removed it from the options

jhendrixMSFT:
Will also need a changelog entry.

Issue #23884, opened 2024-12-19
DownloadBuffer with HTTPRange return unexpected number of written bytes


tamir-laminar:
### Bug Report

- Bug in the following releases :
- - github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/   - 1.5.0
- - github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/     - 1.4.0

```golang
res := make([]byte, 100)
n, err := blobClient.DownloadBuffer(ctx, res, &blob.DownloadBufferOptions{
	Range: blob.HTTPRange{
		Offset: 0,
		Count:  100,
	},
})
fmt.Println(res[:n]) // n is always 100, even if the file has less them 100 bytes, so we get garbage.
```

If the file have less then requested `Count`, it should return the number of bytes that were actually written to the buffer (the variable `n` in that example).
Instead, it always return the `Count`, meaning that if we pass the `Count` in the range option, `n` is always equals to `Count`.

It's a problem since we don't now how to slice the response in order to get the actual data without garbage.

The above also true for azfile library.

### What I expected

```golang 
DownloadBuffer(ctx context.Context, buffer []byte, o *DownloadBufferOptions) (int64, error)
```

DownloadBuffer should return the number of bytes that were actually written to the buffer.

tanyasethi-msft:
Hi @tamir-laminar ,
Thanks for reporting this, we are looking into this issue and will get back to you soon

tanyasethi-msft:
Hi @tamir-laminar 
This will be fixed before our next release in March. This thread will be kept updated on the preogress.

PR #24111, opened 2025-02-12
Fix return value of DownloadBuffer when HTTPRange count is less than data length

ashruti-msft:
Fixes Bug [#23884](https://github.com/Azure/azure-sdk-for-go/issues/23884)


tamir-laminar:
@ashruti-msft Thanks for the fix! :)
I see you are fixing it for blob client, I just wanted to make sure that it is going to be fixed for fileshare client as well (`github.com/Azure/azure-sdk-for-go/sdk/storage/azfile/`), it's the same problem, might even worth doing it in the same PR

sdk/storage/azblob/blob/client.go
```diff
@@ -387,14 +387,17 @@ func (b *Client) downloadBuffer(ctx context.Context, writer io.WriterAt, o downl
...
+			progressLock.Unlock()
 			err = body.Close()
 			return err
 		},
 	})
 	if err != nil {
 		return 0, err
 	}
-	return count, nil
+	return min(dataLength, count), nil
```
tamir-laminar, line 400: Wondering, why do we need to take the `min` of dataLength & count and not just returning `dataLength`?

looking at this seems like it is possible to ask for `count` and still download much more bytes then requested - if that's the case it sound like another bug that is worth solving.

ashruti-msft: Datalength actually is the property of the blob which shows the full size its not related to what count was given by the user.

vibhansa-msft: Ideally, we shall be returning the number of bytes returned back by server as part of responses, as the data might be partial and not the full range we requested.

sdk/azidentity/go.work.sum
chlowell, line 1: Please revert this

chlowell (changes requested):

sdk/storage/azblob/blob/client.go
```diff
@@ -387,14 +387,17 @@ func (b *Client) downloadBuffer(ctx context.Context, writer io.WriterAt, o downl
 			if err != nil {
 				return err
 			}
+			progressLock.Lock()
```
vibhansa-msft, line 390: If we have computed the length of data using getProperties then we shall not be taking cost of locking and incrementing the variables here.

vibhansa-msft (approved):

sdk/storage/azblob/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fixed return value of DownloadBuffer when the HTTPRange count given is greater than the data length for azblob and azfile and added tests. Fixes [#23884](https://github.com/Azure/azure-sdk-for-go/issues/23884)
```
souravgupta-msft, line 10: Update this and add a changelog entry in azfile also
```suggestion
* Fixed return value of DownloadBuffer when the HTTPRange count given is greater than the data length. Fixes [#23884](https://github.com/Azure/azure-sdk-for-go/issues/23884)
```

sdk/storage/azblob/blob/client.go
```diff
@@ -387,14 +390,19 @@ func (b *Client) downloadBuffer(ctx context.Context, writer io.WriterAt, o downl
 			if err != nil {
 				return err
 			}
+			if computeReadLength {
+				progressLock.Lock()
+				dataDownloaded += *dr.ContentLength
+				progressLock.Unlock()
```
souravgupta-msft, line 396: `progressLock` is already used above when the user wants to track the progress of the download operation. Check the section, 376-388. So, we should not reuse the same lock.
We can create a new lock or use atomic variable. Atomic variable approach sounds good to me.

sdk/storage/azblob/client_test.go
```diff
@@ -606,7 +606,13 @@ func performUploadAndDownloadBufferTest(t *testing.T, _require *require.Assertio
 		if downloadCount == 0 {
 			_require.EqualValues(destBuffer, bytesToUpload[downloadOffset:])
 		} else {
-			_require.EqualValues(destBuffer, bytesToUpload[downloadOffset:downloadOffset+downloadCount])
+			_require.EqualValues(destBuffer[downloadOffset:downloadOffset+int(n)], bytesToUpload[downloadOffset:downloadOffset+int(n)])
+			if downloadCount > blobSize {
+				_require.EqualValues(blobSize, n)
```
souravgupta-msft, line 611: This can be incorrect when,
blobSize = 10, downloadOffset = 5, downloadCount = 100
n will be 5 here (download from offset 5 to 10)

souravgupta-msft: It should be `_require.EqualValues(blobSize-downloadOffset, n)`

sdk/storage/azfile/file/client.go
```diff
@@ -516,14 +520,19 @@ func (f *Client) download(ctx context.Context, writer io.WriterAt, o downloadOpt
 			if err != nil {
 				return err
 			}
+			if computeReadLength {
+				progressLock.Lock()
+				dataDownloaded += *dr.ContentLength
+				progressLock.Unlock()
```
souravgupta-msft, line 526: Same comment as azblob

sdk/storage/azblob/blob/client.go
tamir-laminar: @souravgupta-msft  Sorry for interrupting, I just have a quick question:
why do we need to get the `ContentLength` from the request?
two line above we have the `io.Copy` interface, it returns the exact number of bytes that were written to the writer. sound to me that this is exactly the number that we should return.

Doin so, we won't need to use locks or atomic variables

souravgupta-msft: We can use either `ContentLength` returned in the response or the number of bytes returned from `io.Copy`. However, we would still need to use locks or atomic variables as the `Operation` method in `DoBatchTransfer` is a concurrent method and `dataDownloaded` will return the total bytes downloaded by the concurrent download operations.

souravgupta-msft (approved):

chlowell (approved):

ashruti-msft:
@jhendrixMSFT can you please review this PR to suggest any modifications that may be required, thanks!

Issue #24031, opened 2025-01-31
KeyVaultChallengePolicy violates guarantees because it is not goroutine-safe


strager:
### Bug Report

- import path: `github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets`

- SDK version: unsure. Here is what go.mod says:
```
	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.11.1
	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.7.0
	github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets v0.12.0

	github.com/Azure/azure-sdk-for-go/sdk/internal v1.8.0 // indirect
	github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal v0.7.1 // indirect
```

- output of `go version`: `go version go1.23.2 darwin/arm64`

- What happened?

When I use an azsecret.Client from multiple goroutines, Go's race detector complains. For example:
```
WARNING: DATA RACE
Write at 0x00c000306b90 by goroutine 31:
  github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal.(*KeyVaultChallengePolicy).findScopeAndTenant()
      /Users/traduality/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/keyvault/internal@v0.7.1/challenge_policy.go:187 +0x330
  github.com/Azure/azure-sdk-for-go/sdk/keyvault/internal.(*KeyVaultChallengePolicy).Do()
      /Users/traduality/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/keyvault/internal@v0.7.1/challenge_policy.go:81 +0x180
  github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.(*Request).Next()
      /Users/traduality/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/azcore@v1.11.1/internal/exported/request.go:107 +0x1e8
  github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime.(*retryPolicy).Do()
      /Users/traduality/go/pkg/mod/github.com/!azure/azure-sdk-for-go/sdk/azcore@v1.11.1/runtime/policy_retry.go:121 +0x550
  github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.(*Request).Next()
[...]

PR #24120, opened 2025-02-13
[keyvault] make challenge policy thread safe

gracewilcox:
Closes https://github.com/Azure/azure-sdk-for-go/issues/24031

Same PR as https://github.com/Azure/azure-sdk-for-go/pull/24032

This PR fixes the race condition in the keyvault internal module's challenge policy.

Thank you @strager for your hard work!

### DIFF
--- a/sdk/security/keyvault/internal/challenge_policy_test.go
+++ b/sdk/security/keyvault/internal/challenge_policy_test.go
@@ -11,7 +11,10 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -355,3 +357,56 @@ func TestParseTenant(t *testing.T) {
...
+		// test.
+		DisableChallengeResourceVerification: true,
+	})
+	pl := runtime.NewPipeline("", "",
+		runtime.PipelineOptions{PerRetry: []policy.Policy{p}},
+		&policy.ClientOptions{Transport: srv.Client()},
+	)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < concurrentRequestCount; i += 1 {
```
chlowell, line 400:
```suggestion
	for i := 0; i < concurrentRequestCount; i++ {
```

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -355,3 +357,56 @@ func TestParseTenant(t *testing.T) {
...
+		go (func() {
+			defer wg.Done()
+			req, err := runtime.NewRequest(context.Background(), "GET", srv.URL)
+			require.NoError(t, err)
+			res, err := pl.Do(req)
+			require.NoError(t, err)
```
chlowell, line 406: This should be called only on the goroutine executing the test because when `err != nil` it terminates that goroutine. I'm not sure what happens if you (try to) do that from a different goroutine while the target is waiting but I expect  a mess. You should pass the error to the original goroutine and let it fail the test, [for example](https://github.com/Azure/azure-sdk-for-go/blob/9827fb4bf2a537010f3e451fad970ffe5ad8929a/sdk/azcore/runtime/pipeline_test.go#L140-L166)

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -355,3 +357,56 @@ func TestParseTenant(t *testing.T) {
...
+		DisableChallengeResourceVerification: true,
+	})
+	pl := runtime.NewPipeline("", "",
+		runtime.PipelineOptions{PerRetry: []policy.Policy{p}},
+		&policy.ClientOptions{Transport: srv.Client()},
+	)
+
+	wg := sync.WaitGroup{}
+	for i := 0; i < concurrentRequestCount; i += 1 {
+		go (func() {
```
chlowell, line 401: `go func() { ... } ()` is valid i.e., don't need the enclosing parens

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -355,3 +357,56 @@ func TestParseTenant(t *testing.T) {
 	actual = parseTenant(sampleURL)
 	require.Equal(t, expected, actual, "tenant was not properly parsed, got %s, expected %s", actual, expected)
 }
+
+func TestChallengePolicy_ConcurrentRequests(t *testing.T) {
+	concurrentRequestCount := 10
```
chlowell, line 362: this is read only once; I'd just inline it

sdk/security/keyvault/internal/challenge_policy_test.go
```diff
@@ -355,3 +357,56 @@ func TestParseTenant(t *testing.T) {
...
+
+	var srv *httptest.Server
+	srv = httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
chlowell, line 365:
```suggestion
	srv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```

jhendrixMSFT (approved):

PR #24142, opened 2025-02-19
[internal/test] Add Internal Span Validator for Unit Tests

karenychen:
This PR adds an internal Span Validator that can be shared across all packages. The implementation is iterated off of @jhendrixMSFT 's initial version at: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/tracing_test.go

Also included an usage example in `validator_test.go`.


sdk/internal/test/tracing/validator.go
```diff
@@ -0,0 +1,133 @@
...
+
+		t.Cleanup(func() {
+			require.NotNil(t, tt.match, "didn't find a span with name %s", tt.matcher.Name)
+			require.True(t, tt.match.ended, "span wasn't ended")
+			if tt.matcher.Kind != 0 {
+				require.EqualValues(t, tt.matcher.Kind, tt.match.kind, "span kind values don't match")
+			}
+			require.EqualValues(t, matcher.Status, tt.match.status, "span status values don't match")
+			require.ElementsMatch(t, matcher.Attributes, tt.match.attrs, "span attributes don't match")
+			require.ElementsMatch(t, matcher.Links, tt.match.links, "span links don't match")
```
karenychen, line 31: @jhendrixMSFT would you prefer this to be checked only if we specified it in the matcher (similar to the span kind above) or for this to be checked in all cases?

jhendrixMSFT: I think this is fine. IIRC there are legit cases where a span's `Kind` is the zero-value, so we only check it when it has an explicit value.

sdk/internal/go.mod
```diff
@@ -1,17 +1,20 @@
...
 
-go 1.18
+go 1.22.7
+
+toolchain go1.23.1
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.17.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.8.2
+	github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel v0.4.0
```
jhendrixMSFT, line 10: This will need to be removed as `azotel` requires a newer version of Go compared to the rest of our SDKs.

sdk/internal/go.mod
```diff
@@ -1,17 +1,20 @@
 module github.com/Azure/azure-sdk-for-go/sdk/internal
 
-go 1.18
+go 1.22.7
```
jhendrixMSFT, line 3: Revert this and remove the `toolchain` entry.

sdk/internal/go.mod
richardpark-msft: @karenychen, this complicates our story about testing this within the azservicebus module then as well. Looks like we'll have to go with in-memory testing only for now.

Side note: @jhendrixMSFT, I'm going to start the work to move the stress tests out of the azserviecbus module because we will want to use azotel for those. The stress tests would never be published, so it should be okay to move those to the latest version of Go.

karenychen: Hm, I think we need to remove the otel propagation in these validators then because that's what's requiring the higher version. I will make the validator take in provider options so they can decide which propagator to use, or if they need to test propagation at all

sdk/internal/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Release History
 
-## 1.10.1 (Unreleased)
+## 1.11.0-beta.1 (Unreleased)
```
karenychen, line 3: Is this supposed to be 1.10.1? Not sure why we are failing the pipeline :(

richardpark-msft: It's complaining because this file should match the version you have in the changelog: https://github.com/Azure/azure-sdk-for-go/blob/e52a629c15ce39c9d9de7e0bc707853c67bea1c8/sdk/internal/version.go

Just checking though, this is purely internal functionality for testing. Do we need to do a beta for this @jhendrixMSFT?

karenychen: Ah right.. forgot about that one 😅

karenychen: I checked and didn't see any beta versions in the change log, so updated both to 1.11.0

sdk/internal/test/tracing/validator.go
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tracing
```
jhendrixMSFT, line 4: Since this works with `azcore/tracing`, maybe we should rename this package to something like `tracevalidator` or something like that so it doesn't collide with `azcore/tracing`. What do you think?

sdk/internal/test/tracing/validator_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tracing
```
jhendrixMSFT, line 4: Would be good to use the `_test` suffix here so that you have to import `/internal/test/tracing` (or whatever we end up calling it).

sdk/internal/test/tracing/validator.go
karenychen: I think that makes sense, I also had the same doubt but not sure if the best practice for the SDK is to keep the package naming short. I will update it to `tracingvalidator`

karenychen: I personally like the word `validator` a bit more haha, I think it fits better with what it's doing. But I am open to your preferences :)

jhendrixMSFT (approved):

PR #24148, opened 2025-02-21
avoid clientFactory CTOR break change

JiaqiZhang-Dev:
related pr: https://github.com/Azure/autorest.go/pull/1496

modify build.go for avoiding break change in future

### DIFF
--- a/eng/scripts/build.ps1
+++ b/eng/scripts/build.ps1
@@ -12,7 +12,7 @@ param(
     [switch]$removeUnreferencedTypes,
     [switch]$factoryGatherAllParams,
     [string]$config = "autorest.md",
-    [string]$goExtension = "@autorest/go@4.0.0-preview.63",
+    [string]$goExtension = "@autorest/go@4.0.0-preview.70",
     [string]$filePrefix,
     [string]$outputFolder
 )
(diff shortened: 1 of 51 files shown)


tadelesh (approved):

JiaqiZhang-Dev:
@benbp could you please help review this pr? thanks~

eng/scripts/build.ps1
```diff
@@ -56,15 +57,21 @@ function Process-Sdk ()
             $removeUnreferencedTypesFlag = "true"
         }
         
+        $factoryGatherAllParamsFlag = "false"
+        if ($factoryGatherAllParams)
+        {
+            $factoryGatherAllParamsFlag = "true"
+        }
+
```
benbp, line 65:
```suggestion
        $factoryGatherAllParamsFlag = $factoryGatherAllParams.ToString().ToLower()
```

benbp (approved):
LGTM with one suggestion

eng/scripts/build.ps1
JiaqiZhang-Dev: It's a good suggestion! but `factoryGatherAllParams` is only a switch with no param, which means if it exists, the value is true, so there cannot extract `factoryGatherAllParams` value.

benbp: @JiaqiZhang-Dev yes the boolean value from the switch gets stringified, like so:

```
$ function foo([switch]$test) { echo $test.ToString().ToLower() }
$ foo
false
$ foo -test
true
```

Just a style nit, so not a big deal either way :)

PR #24156, opened 2025-02-25
Adopt MSAL managed identity implementation

### LINKED ISSUE #20409: Use MSAL for managed identity token requests
**Context:**
Managed identity support is being added to MSAL. Adapt the Azure Identity library to become a consumer of MSAL's managed identity token request logic.

chlowell:
Closes #20409. This is mostly straightforward refactoring, however the switch broke most of our tests because MSAL uses a per-process cache for managed identity tokens whereas we used a per-client cache.


sdk/azidentity/live_test.go
```diff
@@ -272,7 +284,16 @@ func (p *recordingPolicy) Do(req *policy.Request) (resp *http.Response, err erro
 		r.URL.Host = r.Host
 		r.URL.Scheme = "https"
 	}
-	return req.Next()
+	res, err := req.Next()
+	if err == nil {
+		// if the response is a recording mismatch, return it as a simple error that prints clearly in the test log
+		if e := res.Header.Get("x-request-mismatch-error"); e != "" {
```
jhendrixMSFT, line 290: This seems duplicated from your same change in `/sdk/internal`. I guess that hasn't been released yet though. Will this go away once that happens?

chlowell: Yes, I'll remove this before merging

Issue #23355, opened 2024-08-20
azfake support for azsecrets.Client


mjnovice:
### Feature Request

The current support for fake client seems to be missing for azsecrets.Client making it not so east to unit test it.

gracewilcox:
Hi @mjnovice! Thank you for the feature request, we'll work on adding it to our SDK.

gracewilcox:
waiting on support in tcgc

PR #24168, opened 2025-02-25
[azcertificates] add fakes support 

gracewilcox:
part of https://github.com/Azure/azure-sdk-for-go/issues/23355

add fake support to azcertificates


sdk/security/keyvault/azcertificates/build.go
```diff
@@ -1,4 +1,4 @@
-//go:generate tsp-client update
+//go:generate tsp-client update --local-spec-repo /home/grace/code/azure-rest-api-specs/specification/keyvault/Security.KeyVault.Certificates
```
richardpark-msft, line 1: I feel like this line shouldn't go in...

PR #24170, opened 2025-02-25
[azsecrets] add fakes support

gracewilcox:
part of https://github.com/Azure/azure-sdk-for-go/issues/23355

add fake support to azsecrets


sdk/security/keyvault/azsecrets/build.go
```diff
@@ -1,4 +1,4 @@
-//go:generate tsp-client update
+//go:generate tsp-client update --local-spec-repo /home/grace/code/azure-rest-api-specs/specification/keyvault/Security.KeyVault.Secrets
```
chlowell, line 1: 👀

sdk/security/keyvault/azsecrets/options.go
```diff
@@ -27,18 +27,21 @@ type GetSecretOptions struct {
 // ListDeletedSecretPropertiesOptions contains the optional parameters for the Client.NewListDeletedSecretPropertiesPager
 // method.
 type ListDeletedSecretPropertiesOptions struct {
-	// placeholder for future optional parameters
+	// Maximum number of results to return in a page. If not specified the service will return up to 25 results.
+	MaxResults *int32
```
chlowell, line 31: Can you please do this in a separate PR? It's related to fakes only by our process

gracewilcox: I wish, but I don't think that's logistically possible. The typespec needs to be correct for me to generate fakes. So I need to either repress max results in the typespec or keep it.

PR #24174, opened 2025-02-25
[azopenaiextensions] Doing some housekeeping and fixing some 'vet' errors

richardpark-msft:


jhendrixMSFT (approved):

richardpark-msft:
Just some fixes for vet errors, and a test assertion that appears to be using an outdated type. CC: @glecaros

PR #24175, opened 2025-02-26
First preview of azbatch module

chlowell:
Supporting the Azure Batch data plane. https://github.com/Azure/azure-rest-api-specs/pull/32768 is the corresponding spec PR.


chlowell:
This is ready to review but not to merge (I'll need to update the TSP commit after Azure/azure-rest-api-specs#32768 merges)

sdk/batch/azbatch/models.go
```diff
@@ -0,0 +1,4027 @@
...
+
+	// Details of a Job Manager Task to be launched when the Job is started. If the Job does not specify a Job Manager Task, the
+	// user must explicitly add Tasks to the Job. If the Job does specify a Job Manager Task, the Batch service creates the Job
+	// Manager Task when the Job is created, and will try to schedule the Job Manager Task before scheduling other Tasks in the
+	// Job. The Job Manager Task's typical purpose is to control and/or monitor Job execution, for example by deciding what additional
+	// Tasks to run, determining when the work is complete, etc. (However, a Job Manager Task is not restricted to these activities
+	// - it is a fully-fledged Task in the system and perform whatever actions are required for the Job.) For example, a Job Manager
+	// Task might download a file specified as a parameter, analyze the contents of that file and submit additional Tasks based
+	// on those contents.
+	JobManagerTask *JobManagerTask
```
richardpark-msft, line 381: I'm sure this came straight from the API, but there's an inconsistency with how these fields are named.

`JobManagerTask`, and friends exist (JobPreparationTask, JobReleaseTask, etc..) but then we have NetworkConfiguration (ie, not JobNetworkConfiguration, like the others).

TBH, I prefer _not_ having the `Job` prefix since we're already in a CreateJobContent.

chlowell: Thanks for pointing this out, I created #24508 to track making these names consistently good. I want to avoid revising the API in this PR because Batch devs are currently renaming things upstream for all languages (they're actually weeding out redundant prefixes like this) and I don't want to block our first beta on that work or to create churn.

richardpark-msft: Sounds good to me, it'd be ideal if they can rename it at the source!

richardpark-msft (approved):

PR #24206, opened 2025-03-05
Skip apireview creation on main scheduled+ci pipelines for non-sdk changes

benbp:
Skip apireview creation on main scheduled+ci pipelines for non-sdk changes

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -30,7 +30,7 @@ steps:
 
       Copy-Item "$(Build.ArtifactStagingDirectory)/PackageInfo" -Destination $directoryToPublish -Recurse
     displayName: 'Create Go APIView Artifact'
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -30,7 +30,7 @@ steps:
 
       Copy-Item "$(Build.ArtifactStagingDirectory)/PackageInfo" -Destination $directoryToPublish -Recurse
     displayName: 'Create Go APIView Artifact'
-    condition: succeeded()
+    condition: and(succeeded(), ${{ ne(true, startsWith(parameters.ServiceDirectory, 'eng')) }})  # exclude changes to eng/
```
weshaggard, line 33: Isn't IsSdkLibrary parameter already used for this? See condition on step before this.

benbp: Ah, good catch. Maybe I can use that instead.

weshaggard (approved):

PR #24208, opened 2025-03-05
Upgrade azotel dependencies

chlowell:
Transitive dependency github.com/golang/glog published a CVE


sdk/tracing/azotel/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel
 
-go 1.22.7
+go 1.23.0
 
-toolchain go1.23.1
+toolchain go1.24.0
```
jhendrixMSFT, line 5: Hmm I wonder if we should set the `toolchain` to the same version as the Go version. Thoughts?

jhendrixMSFT: Or I wonder if we even need to specify this? I've been reading https://go.dev/doc/toolchain but I haven't completely wrapped my head around it yet.

chlowell: Me neither 😆 

`go get` seems always to add it whereas `mod tidy` does so only for some states. I don't want it if it isn't necessary, so I suppose I'll push the "mod tidy doesn't want to add a toolchain directive" state

sdk/tracing/azotel/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel
 
-go 1.22.7
+go 1.23.0
```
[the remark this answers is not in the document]
chlowell: `mod tidy` insists on the patch revision

jhendrixMSFT (approved):

PR #24212, opened 2025-03-06
[Release] sdk/resourcemanager/applicationinsights/armapplicationinsights/2.0.0-beta.3

jliusan:
https://github.com/Azure/sdk-release-request/issues/5853#issuecomment-2702717702


JiaqiZhang-Dev:
@jliusan could you please add a `-factoryGatherAllParams` flag in build.go to avoid ClientFactory breaking change? fyi: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/advisor/armadvisor/build.go;
And the latest version tag has removed `-only` suffix, but it seems sill existed in generated code.

Issue #24022, opened 2025-01-29
[Key Vault] Re-generate libraries and release first beta for service version 7.6-preview.2


gracewilcox:
Use the service specification added in this commit. Changes should only include two new KV Administration LRO endpoints: /prebackup and /prerestore, much akin to /backup and /restore.

chlowell:
The 7.6-preview.2 commit is https://github.com/Azure/azure-rest-api-specs/commit/b8d26b0e4c1886458fa56c22aac09c3e3e9a5c9e

gracewilcox:
code merged, will go out in next release

PR #24220, opened 2025-03-07
[azadmin] generate 7.6-preview.2 

gracewilcox:
part of https://github.com/Azure/azure-sdk-for-go/issues/24022

Added `PreFullBackup` and `PreFullRestore`


sdk/security/keyvault/azadmin/internal/generate/transforms.go
```diff
@@ -46,6 +46,7 @@ func main() {
 	regexReplace("backup/models_serde.go", `(?:\/\/.*\s)+func \(\w \*?Error\).*\{\s(?:.+\s)+\}\s`, "")
 
 	//  modify Restore to use implementation with custom poller handler
+	regexReplace("backup/client.go", `\[PreFullRestoreResponse\], error\) \{\s(?:.+\s)+\}`, "[PreFullRestoreResponse], error) {return client.beginPreFullRestore(ctx, preRestoreOperationParameters, options)}")
```
chlowell, line 49: 🤔  for my education/memory, why does pre/restore and not pre/backup need a custom poller?

gracewilcox: Don't know why, but the service team isn't following any standard LRO policy for restore. So I need a custom poller to match the wacky things the service is doing.

chlowell (approved):

sdk/security/keyvault/azadmin/internal/generate/transforms.go
gracewilcox: Restore is an async PUT poller, so there should be a final GET on the original resource URL. The service is not doing the final GET, so need to use a custom poller that doesn't implement the final GET

PR #24230, opened 2025-03-10
[stress] Moving azservicebus's stress tests into messaging/stress

richardpark-msft:
This PR moves all of the stress tests for Service Bus into their own module (`messaging/stress`), which is NEVER intended to be referenced by any external or package code. It is never intended to be released

This approach allows us to avoid the project from being downloaded by users that are only intending to import `azservicebus` and also allows us to set a later version of Go to compile and run with.

There's been very few changes here - mostly moved the code over, adjusted package paths, and fixed a few spots because of the new directory structure.


chlowell (approved):

richardpark-msft:
> What's the plan for CI, does this need its own `ci.yaml` file to at least compile/vet/lint the code, or does that happen as part of the stress test proper?

There is a weekly pipeline that runs for stress tests that will build this, but having a normal CI pipeline makes sense too. I'll check this in, and start a new PR to come up with a sensible CI. It needs to skip some things that normal pipelines would do (like, building Go 1.18, etc..).

RickWinter (commented):
NIT: `sb` isn't the best name. I get no-one will use this package externally. We should still use reasonably descriptive package & module names.

richardpark-msft:
> NIT: `sb` isn't the best name. I get no-one will use this package externally. We should still use reasonably descriptive package & module names.

I can change it. I really just wanted to avoid seeing a `package servicebus` or anything like that that someone might find randomly, but it could be less cryptic :)

PR #24253, opened 2025-03-12
Adding an example of using a reasoning model

kingernupur:
- [x ] The purpose of this PR is explained in this or a referenced issue.
- [x ] The PR does not update generated files.

### DIFF
--- a/sdk/ai/azopenai/assets.json
+++ b/sdk/ai/azopenai/assets.json
@@ -2,5 +2,5 @@
   "AssetsRepo": "Azure/azure-sdk-assets",
   "AssetsRepoPrefixPath": "go",
   "TagPrefix": "go/ai/azopenai",
-  "Tag": "go/ai/azopenai_4a8ededa66"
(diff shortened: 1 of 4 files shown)


richardpark-msft:
@kingernupur, can you add this example to the https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/ai/azopenaiextensions project instead? We have a very small set of examples there, and I think it'd be good to have.

kingernupur:
> @kingernupur, can you add this example to the https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/ai/azopenaiextensions project instead? We have a very small set of examples there, and I think it'd be good to have.

@richardpark-msft That is a good suggestion, and I am working on writing a similar counterpart example in the extensions package. But I don't see any specific wrapper for reasoning params or models in there. 
What do you think about keeping the example and sample here also since we don't have anything related to reasoning here yet? Meanwhile, I will find something reasonable for the other repo; if not, I will add a similar counterpart.

Issue #22837, opened 2024-05-03
Entra access token authentication policies such as `BearerTokenAuthenticationPolicy` should respect `refresh_on` information


christothes:
Long lived credentials such as those received from managed identity authentication include additional metadata concerning when a token can/should be refreshed. Our authentication policies should take this information into account when refreshing access tokens.

This involves:
- Modifying relevant authentication policies
- Modifying the `AccessToken` type to include this optional information
- Modifying Azure.Identity credential implementations to populate the refresh_on information in the `AccessToken`

joshfree:
Related

https://github.com/Azure/azure-sdk-for-java/issues/40027

https://github.com/Azure/azure-sdk-for-python/issues/35473

https://github.com/Azure/azure-sdk-for-js/issues/29576

https://github.com/Azure/azure-sdk-for-go/issues/22837

https://github.com/Azure/azure-sdk-for-cpp/issues/5598

https://github.com/Azure/azure-sdk-for-rust/issues/1657

chlowell:
Blocked on https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/239

chlowell:
🤔 on second thought I believe the only change we need from MSAL is to expose any `refresh_in` value provided by the STS. Everything else should be feasible in azidentity. And I can imagine a hacky way to get `refresh_in` without MSAL's help.

chlowell:
On third thought, this is blocked because MSAL's token cache has a [hardcoded expiration time](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/ae2db6b72c7010958355f448e99209bd28e76e67/apps/internal/base/internal/storage/items.go#L128-L130) preventing us from acquiring a new token when a cached one has at least 5 minutes left to expiry.

andyzhangx:
is anyone working on this bug? that's a critical blocking bug preventing us to use track2. Recently we AKS team found that after migrating to track2 sdk, the managed identity token would expire after 24 hours which is not easy to be caught in e2e test, we have to revert to version using sdk track1, pls fix this issue ASAP, otherwise this track2 sdk is unusable for us, thanks!

andyzhangx:
here is an example fix in forked branch: https://github.com/hashicorp/go-azure-sdk/pull/362

chlowell:
Can you please explain how the lack of this feature makes track 2 unusable? What breaks when your application gets a token valid for 24 hours?

[...]

PR #24268, opened 2025-03-14
Make temporal.Resource refreshing configurable

chlowell:
This adds a callback that indicates whether the underlying resource should be refreshed despite its not having expired. Part of #22837

### DIFF
--- a/sdk/internal/temporal/resource.go
+++ b/sdk/internal/temporal/resource.go
@@ -89,7 +89,6 @@ func (er *Resource[TResource, TState]) Get(state TState) (TResource, error) {
 			if !er.acquiring {
 				// If another thread/goroutine is not acquiring/updating the resource, this thread/goroutine will do it
 				er.acquiring, acquire = true, true
-				er.lastAttempt = now
 				break
 			}
 			// Getting here means that this thread/goroutine will wait for the updated resource
--- a/sdk/internal/temporal/resource.go
+++ b/sdk/internal/temporal/resource.go
@@ -98,7 +98,6 @@ func (er *Resource[TResource, TState]) Get(state TState) (TResource, error) {
 				// If another thread/goroutine is not acquiring/renewing the resource, and none has attempted
 				// to do so within the last 30 seconds, this thread/goroutine will do it
 				er.acquiring, acquire = true, true
-				er.lastAttempt = now
 				break
 			}
 			// This thread/goroutine will use the existing resource value while another updates it
@@ -119,6 +118,7 @@ func (er *Resource[TResource, TState]) Get(state TState) (TResource, error) {
 		// This thread/goroutine has been selected to acquire/update the resource
 		var expiration time.Time
 		var newValue TResource
+		er.lastAttempt = now
 		newValue, expiration, err = er.acquireResource(state)
 
 		// Atomically, update the shared resource's new value & expiration.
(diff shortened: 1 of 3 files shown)


sdk/internal/temporal/resource.go
```diff
@@ -121,3 +148,8 @@ func (er *Resource[TResource, TState]) Expire() {
 	// Reset the expiration as if we never got this resource to begin with
 	er.expiration = time.Time{}
 }
+
+func (r *Resource[TResource, TState]) expiringSoon(TResource, TState) bool {
```
jhendrixMSFT, line 152: The receiver should be named `er` to match the other methods (worse name I know but then it's consistent; thought the linter would flag this).

sdk/internal/temporal/resource.go
```diff
@@ -91,7 +119,6 @@ func (er *Resource[TResource, TState]) Get(state TState) (TResource, error) {
 		// This thread/goroutine has been selected to acquire/update the resource
 		var expiration time.Time
 		var newValue TResource
-		er.lastAttempt = now
```
jhendrixMSFT, line 94: Is moving this necessary and safe? This will change one of the conditions when it's set.

chlowell: I believe it's safe, however it isn't necessary or worth the churn. I reverted the move.

sdk/internal/temporal/resource.go
```diff
@@ -62,12 +89,12 @@ func (er *Resource[TResource, TState]) Get(state TState) (TResource, error) {
 			if !er.acquiring {
 				// If another thread/goroutine is not acquiring/updating the resource, this thread/goroutine will do it
 				er.acquiring, acquire = true, true
+				er.lastAttempt = now
```
jhendrixMSFT, line 92: I think this should have also been removed in the reversion?

chlowell: You're quite right! My bad.

sdk/internal/temporal/resource.go
```diff
@@ -121,3 +147,7 @@ func (er *Resource[TResource, TState]) Expire() {
 	// Reset the expiration as if we never got this resource to begin with
 	er.expiration = time.Time{}
 }
+
+func (er *Resource[TResource, TState]) expiringSoon(TResource, TState) bool {
+	return er.expiration.Add(-5 * time.Minute).Before(time.Now())
```
jhendrixMSFT, line 152: Previously, we were using `now` which was computed at the beginning of `Get()`. Now, we're calculating a new `now` (locally) which is no longer constant in the `for` loop. Is this intentional, or should `expiringSoon` take a `time.Time` param which is the loop's `now`?

chlowell: It's intentional in that ShouldRefresh gets only the resource and the state and adding `now time.Time` there is too weird for me. I suppose recomputing now is harmful in that it arguably wastes some number of cycles, but do you see any possibility of incorrect behavior? I suppose it could trigger a refresh some nanoseconds earlier than expected, but I don't see harm in that because Get observes backoff (although it does shrink the backoff window by setting `er.lastAttempt` to the clock time before acquiring the lock and refreshing the resource)

jhendrixMSFT: Yeah I agree adding a `now time.Time` param would be weird.

Mostly I was worried that the `for` loop could somehow become live-locked due to drifting values of `now` and `local now` (I think there's a Space Balls joke in here somewhere). But since it doesn't sleep then I think that can't happen.

jhendrixMSFT: It would be good to add a comment though that `time.Now()` is intentional.

PR #24271, opened 2025-03-14
[azcore/runtime] Add ModuleName to StartSpanOptions

karenychen:
The error type is hardcoded as `azcore` in `runtime.StartSpan`. It will miscategorize all exported errors from modules external to `azcore` to `azcore.Error`.

This PR supports customization of the ModuleName in `runtime.StartSpanOptions`

### DIFF
--- a/sdk/azcore/runtime/policy_http_trace.go
+++ b/sdk/azcore/runtime/policy_http_trace.go
@@ -98,6 +98,9 @@ func (h *httpTracePolicy) Do(req *policy.Request) (resp *http.Response, err erro
 
 // StartSpanOptions contains the optional values for StartSpan.
 type StartSpanOptions struct {
+	// ModuleName is the name of the module that created the span.
+	// If not specified, ModuleName defaults to the azcore module name.
+	ModuleName string
 	// Kind indicates the kind of Span.
 	Kind tracing.SpanKind
 	// Attributes contains key-value pairs of attributes for the span.
@@ -136,6 +139,9 @@ func StartSpan(ctx context.Context, name string, tracer tracing.Tracer, options 
 	if options == nil {
 		options = &StartSpanOptions{}
 	}
+	if options.ModuleName == "" {
+		options.ModuleName = shared.Module
+	}
 	if options.Kind == 0 {
 		options.Kind = tracing.SpanKindInternal
 	}
@@ -148,7 +154,7 @@ func StartSpan(ctx context.Context, name string, tracer tracing.Tracer, options 
 	ctx = context.WithValue(ctx, ctxActiveSpan{}, options.Kind)
 	return ctx, func(err error) {
 		if err != nil {
-			errType := strings.Replace(fmt.Sprintf("%T", err), "*exported.", "*azcore.", 1)
+			errType := strings.Replace(fmt.Sprintf("%T", err), "*exported.", fmt.Sprintf("*%s.", options.ModuleName), 1)
 			span.SetAttributes(tracing.Attribute{Key: attrErrType, Value: errType})
 			span.SetStatus(tracing.SpanStatusError, fmt.Sprintf("%s:\n%s", errType, err.Error()))
 		}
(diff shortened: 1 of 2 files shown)


sdk/azcore/runtime/policy_http_trace.go
```diff
@@ -148,7 +154,7 @@ func StartSpan(ctx context.Context, name string, tracer tracing.Tracer, options
 	ctx = context.WithValue(ctx, ctxActiveSpan{}, options.Kind)
 	return ctx, func(err error) {
 		if err != nil {
-			errType := strings.Replace(fmt.Sprintf("%T", err), "*exported.", "*azcore.", 1)
+			errType := strings.Replace(fmt.Sprintf("%T", err), "*exported.", fmt.Sprintf("*%s.", options.ModuleName), 1)
```
jhendrixMSFT, line 157: Hmm I think it would be good if we can come up with a way that removes the need for this substitution. I also think it would be ideal to avoid the need to pass the type name as an optional param (we can save this idea just in case). I have some ideas below. Let me know what you think or if you have other ideas to add to the list.

### use reflection to build the complete type name ###
Since a package name like `exported` isn't very helpful, and can be ambiguous, perhaps we build the complete type name using reflection. Something like this.
```go
tt := reflect.TypeOf(err)
if tt.Kind() == reflect.Ptr {
	tt = tt.Elem()
}
var errType string
if pp := tt.PkgPath(); pp != "" {
	errType = fmt.Sprintf("%s.%s", pp, tt.Name())
} else {
	errType = fmt.Sprintf("%T", err)
}
```
The con I see here (besides the use of reflection which can be hard to get correct in all cases) is that you end up with really long type names.
```
github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.ResponseError
```

### add an interface that pretty-prints the error type name ###
We could type assert for some interface that pretty-prints the error type name. Using [`fmt.GoStringer`](https://pkg.go.dev/fmt#GoStringer) just for illustration purposes.
```go
var errType string
if gs, ok := err.(fmt.GoStringer); ok {
	errType = gs.GoString()
} else {
	errType = fmt.Sprintf("%#v", err)
}
```

###  just use `fmt.Sprintf("%T", err)` ###
While the simplest, the error names could end up being ambiguous/unhelpful. e.g. `*exported.ResponseError` doesn't clue in from where the error originated. Maybe this isn't as big of a deal in practice?

karenychen: Reading the OTel convention for the `error.types` attribute https://opentelemetry.io/docs/specs/semconv/attributes-registry/error/#error-type

> The error.type SHOULD be predictable, and SHOULD have low cardinality.

Based on this, I am not sure if the reflection way is the most ideal. Personally I would prefer the pretty-print method, it has a fallback to the `fmt.Sprintf("%T", err)` so the consuming modules will have a choice to either opt-in for a pretty print, or just use `*exported.Error` if that is not something they care about?

cc @lmolkova and @richardpark-msft for in case they have other preference/ideas 👀

lmolkova: for HTTP, the `error.type` should match the status code (if error) or fully qualified error type - 
https://github.com/open-telemetry/semantic-conventions/blob/301d6ee1955cf8c5e1def84d65edcdeb6be4f865/docs/http/http-spans.md?plain=1#L444

here's how golang otel sets fully qualified error name - https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a15738a6ca9b974a6fcd8345395136276a0df87c/instrumentation/net/http/otelhttp/internal/semconv/httpconv.go#L422. It  seems to align with one of the proposals above.

Having `github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/exported.ResponseError`  or `exported.ResponseError`  is almost equally unuseful - ideally we should record something more specific like "dns error" or "tls handshake failed" - e.g. .NET records one of these enum values - https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequesterror?view=net-9.0.

Is there something more specific like this in golang? If not, the length of `error.type` does not matter too much - HTTP error responses that end up in exception (and not in non-success status code) should be rare comparing to the volume of the rest of the telemetry and having a long type seems to align with general otel.

richardpark-msft: Feels like the "pretty print interface" for errors is the way to go then. Then we can just agree on some string constants, and you can classify things under "auth error", or "not found", which can be incredibly useful as a category.

I'm only saying this because it's clear everyone else agrees, so now it looks like I'm part of this amazing group.

PR #24276, opened 2025-03-15
export acr audience

mainred:
The intention of the PR is to export the preferable ACR audience to let other Azure client and for clouds like azure stack to refer this variable.

### DIFF
--- a/sdk/containers/azcontainerregistry/CHANGELOG.md
+++ b/sdk/containers/azcontainerregistry/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Other Changes
 * Default audience of Azure Container Registry of all clouds to https://containerregistry.azure.net
+* Export ACR audience variable
 
 ## 0.2.2 (2024-09-19)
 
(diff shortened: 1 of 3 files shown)


mainred:
cc @chlowell
Created another PR as a follow-up of https://github.com/Azure/azure-sdk-for-go/pull/24215

chlowell:
What's the use case for this? Naively, I don't expect this audience to apply to Stack or non-public clouds in general, and I doubt a developer using a non-Azure SDK client would want to take a dependency on the SDK to access this value

mainred:
> What's the use case for this?

AKS uses managed identity to get acr refresh token for acr image pull.

> I don't expect this audience to apply to Stack or non-public clouds in general

acr audience can be applied to all kinds of cloud per @northtyphoon 

I think it's a great chance for you ACR expert to review how AKS talks to ACR. Could you please take a look at this PR and overall implementation? That will be great help considering the future maintenance. https://github.com/kubernetes-sigs/cloud-provider-azure/pull/8593

@chlowell Thanks in advance.

chlowell:
Your kubernetes-sigs PR makes sense to me: an application that needs this value and doesn't use an azcontainerregistry client should simply copy the value.

I suppose that leaves private cloud developers who do use azcontainerregistry clients as the possible users of this constant. I'm still skeptical that this audience is correct for all private clouds, or even enough private clouds to make it worth exporting. Do you know any documentation covering that?

PR #24281, opened 2025-03-18
Upgrade azcore dependencies, go directive (to 1.23)

chlowell:
Targeting a release branch (v1.17.0 tag plus #24265) to avoid GAing new tracing API. Looks like we lost API approval for v1.17.0 though


benbp (approved):
+1 for eng/

sdk/azcore/testdata/perf/go.mod
```diff
@@ -1,15 +1,17 @@
 module github.com/Azure/azure-sdk-for-go/sdk/azcore/testdata/perf
 
-go 1.18
+go 1.23.0
+
+toolchain go1.24.0
```
jhendrixMSFT, line 5: If we want to run the perf tests against both versions in CI, then this should be removed.

chlowell: Thanks, I overlooked this one

jhendrixMSFT (approved):

PR #24296, opened 2025-03-20
Use commonly defined go N0, N-1 versions in test matrix configs

benbp:
Defining these in `globals.yml` and referencing them via matrix environment variable syntax allows us to have one single place to rev versions for testing.

### DIFF
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -20,3 +20,7 @@ variables:
 
   # disable code coverage redesign until it's been fixed (causes inadvertent deflation of CC numbers)
   GOEXPERIMENT: nocoverageredesign
+
+  # Supported versions for testing. These variables are referenced in test matrix files.
+  GO_VERSION_N0: 1.24.1
+  GO_VERSION_N1: 1.23.7
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -22,5 +22,5 @@ variables:
   GOEXPERIMENT: nocoverageredesign
 
   # Supported versions for testing. These variables are referenced in test matrix files.
-  GO_VERSION_N0: 1.24.1
-  GO_VERSION_N1: 1.23.7
+  GO_VERSION_LATEST: 1.24.1
+  GO_VERSION_PREVIOUS: 1.23.7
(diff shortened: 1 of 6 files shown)


chlowell (approved):
The purpose of these matrices is to reduce resource costs by using only one live test runner. Is there a way to accomplish that without ongoing maintenance like this?

benbp:
@chlowell probably the easiest way is to define a default `GoVersion` variable in https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/variables/globals.yml, and let the matrix configs optionally override it. This would only work for matrix files that define a single version though.

gracewilcox (approved):

benbp:
@chlowell Adding some complexity, we could also think of a more general solution, where the matrix instead specifies:

```
{
  "matrix": {
    ...
    "GoVersionSelector": [env:VERSION_N0, env:VERSION_N1]
  }
}
```

And then we can just define these variables in [globals.yml](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/pipelines/templates/variables/globals.yml). Then you'd have one central config file to update versions from, and custom matrix files could select version N-0, etc. without needing maintenance. Using the matrix env variable lookup for this would enable the generated pipeline job names to still make sense.

chlowell:
I like that idea. I see only 3 modules that would use those constants today but they're all trying to express "use only one runner". It would be nice to have a no-maintenance way to do that.

benbp:
@chlowell @jhendrixMSFT I updated this PR to centralize go version configs. Let me know what you think of the naming, etc.

weshaggard (approved):

eng/pipelines/templates/variables/globals.yml
```diff
@@ -20,3 +20,7 @@ variables:
 
   # disable code coverage redesign until it's been fixed (causes inadvertent deflation of CC numbers)
   GOEXPERIMENT: nocoverageredesign
+
+  # Supported versions for testing. These variables are referenced in test matrix files.
+  GO_VERSION_N0: 1.24.1
+  GO_VERSION_N1: 1.23.7
```
chlowell, line 26: Can we use these in the base matrix template or reference that template here so there's only one file to edit when a new Go ships?

benbp: This is the one file, and it will propagate to the matrix configs, unless I missed a spot? Or were you thinking we'd have these version settings elsewhere?

chlowell: I'm not sure what I was getting at; I suppose I misread this PR or misunderstood matrix generation. This looks good to me

jhendrixMSFT (approved):

PR #24320, opened 2025-03-26
Fix azadmin HSM weekly deployment tests/timeout

benbp:
Leverages the changes in https://github.com/Azure/azure-sdk-for-go/pull/24294 which split out the test resources post script to a separate step with a fresh az cli login when `PersistOidcToken` is set to true.

I don't think the pre script is necessary anymore. We only added it to make the login token extended by a minute or two by running it after some setup code, but it shouldn't be necessary anymore since we were timing out on that token with the HSM deployment anyway.

I'm updating the version in the matrix files here just so tests will pass, but this will be replaced soon after with https://github.com/Azure/azure-sdk-for-go/pull/24296


sdk/security/keyvault/azadmin/test-resources-post.ps1
```diff
@@ -90,6 +90,13 @@ if (!$DeploymentOutputs['AZURE_MANAGEDHSM_URL']) {
     exit
 }
 
+Log "Refreshing OIDC token"
+az cloud set -n $Environment
+az login --federated-token $env:ARM_OIDC_TOKEN --service-principal -t $TenantId -u $TestApplicationId
+if ($LASTEXITCODE) { exit $LASTEXITCODE }
+az account set --subscription $SubscriptionId
+if ($LASTEXITCODE) { exit $LASTEXITCODE }
+
```
gracewilcox, line 99: Should this also be added to azkeys? Know we also deploy managed hsm's on a weekly basis there

benbp: Oh! Yes I'll add it there.

chlowell (approved):

PR #24321, opened 2025-03-26
add build.go template

JiaqiZhang-Dev:


JiaqiZhang-Dev:
@tadelesh I'm not sure if we need to also change the generator logic to use `go generator` instead of `tsp-client init`, since there are some arguments (like emitter-options) needs to transfer from generator to tsp-client, if use `go generator`, we may need to dynamically modify the build.go file to add arguments.

tadelesh:
> @tadelesh I'm not sure if we need to also change the generator logic to use `go generator` instead of `tsp-client init`, since there are some arguments (like emitter-options) needs to transfer from generator to tsp-client, if use `go generator`, we may need to dynamically modify the build.go file to add arguments.

we should do that. for swagger time, you could use go generate to get the same code as we released. when migrate to typespec, i believe this should also work.

JiaqiZhang-Dev:
@tadelesh I change the `TypeSpecUpdateGeneator` to use `go generate`; For `TypeSpecOnboardGeneator`, it still uses `tsp-client init`, because the main logic of `go generate` is to execute `tsp-client update`, it relies on `tsp-location.yaml`.

eng/scripts/typespec_build.ps1
```diff
@@ -0,0 +1,113 @@
...
+    {
+        Write-Host "##[command]Executing TypeSpec generation in " $currentDirectory
+        
+        # Install tsp-client if not already installed
+        Write-Host "##[command]Ensuring TypeSpec tooling is installed..."
+        npm list -g @azure-tools/typespec-client-generator-cli || npm install -g @azure-tools/typespec-client-generator-cli@v0.10.0
+        if ($LASTEXITCODE) {
+            Write-Host "##[error]Error installing TypeSpec tooling"
+            exit $LASTEXITCODE
+        }
```
tadelesh, line 48: for env prerequisite, i don't believe we should ensure it here. for swagger case, we do not install `autorest` in the script, we just let cx to prepare the right env with doc.

eng/tools/generator/cmd/v2/common/cmdProcessor.go
```diff
@@ -294,6 +294,101 @@ func ExecuteTspClient(path string, args ...string) error {
...
+		for _, line := range strings.Split(stdoutBuffer.String(), "\n") {
+			if len(strings.TrimSpace(line)) == 0 {
+				continue
+			}
+			if strings.Contains(line, "generation complete") {
+				return nil
+			}
+		}
+	}
+	if cmdWaitErr != nil || stderrBuffer.Len() > 0 {
```
tadelesh, line 340: i'm wondering if we could rely on the ret code of `go generate` to judge if the generation is success or not? then we could use the same logic for both swagger and typespec.

JiaqiZhang-Dev: I tried it and it works, `cmd.Wait()` will return an error if command not exit with the zero code. It's a better way to judge whether the generation is success, instead of writing some hardcode for stderr.

eng/scripts/typespec_build.ps1
```diff
@@ -0,0 +1,110 @@
...
+
+function Process-TypeSpec ()
+{
+    $currentDirectory = Get-Location
+    
+    # Check for tsp-location.yaml first
+    if (-not (Test-Path "tsp-location.yaml")) {
+        Write-Host "##[error]This is not a TypeSpec-based SDK. No tsp-location.yaml file found in $currentDirectory"
+        return
+    }
```
tadelesh, line 25: shall we also update the `build.go` for swagger to prevent crash.

tadelesh (approved):
overall lgtm. please make sure run `typespec_build.ps1` separately could success update all sdks generated by typespec. and run `go generate` for each sdk generated by typespec could success with no code change.

eng/scripts/typespec_build.ps1
JiaqiZhang-Dev: Yes, it'll be good to enhance swagger script. I have done this change.

eng/tools/generator/cmd/v2/common/cmdProcessor.go
JiaqiZhang-Dev: Sorry for the incorrect information above. When I run tests, I found that even compiler has some error message, `tsp-client update` also exit with zero code and put the error message into stdout. (But generator will report error when get exports from new source code, because codegen do not generate code successfully). It seems to be a design issue of `tsp-client`. So, we may need to keep those hardcode logic to check whether generation is success or not🥲
image

weshaggard:
cc @lmazuel @raych1 

@JiaqiZhang-Dev @tadelesh can you help me understand this change a little more? We do want folks to be using tsp-client for generating their sdk for TypeSpec. That is the common way across all the languages we don't want to make folks learn a custom command for all languages. While it is fine if go generate works it isn't fine if it is the only way to work as tsp-client should work for all languages.

eng/scripts/typespec_build.ps1
```diff
@@ -0,0 +1,110 @@
...
+    [switch]$cleanGenerated,
+    [switch]$format,
+    [switch]$tidy,
+    [string]$emitterOptions = "module-version=0.1.0", # default options
+    [string]$localSpecRepo
+)
+
+. (Join-Path $PSScriptRoot .. common scripts common.ps1)
+
+function Process-TypeSpec ()
```
raych1, line 17: Will this work for data plane scenarios? 
@gracewilcox could you review if this script aligns with what commands you run for local developing purpose?

gracewilcox: For local development, I always run `go generate` which executes the commands in my [build.go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/security/keyvault/azkeys/build.go) file.

For example:
tsp-client update
go run ./internal/transforms.go
goimports -w .

eng/tools/generator/cmd/v2/common/generation.go
```diff
@@ -604,10 +604,25 @@ func (t *TypeSpecCommonGenerator) Generate(generateParam *GenerateParam) error {
 	if generateParam.TypeSpecEmitOption != "" {
 		emitOption = fmt.Sprintf("%s;%s", emitOption, generateParam.TypeSpecEmitOption)
 	}
-	err := ExecuteTypeSpecGenerate(ctx, emitOption, generateParam.TspClientOptions)
+	log.Printf("Start to run `tsp-client init` to initialize generation")
+	err := ExecuteTypeSpecInit(ctx, emitOption, generateParam.TspClientOptions)
```
raych1, line 608: Should 'init' command called only for the brand-new service case? It doesn't need to be called if the service folder and 'tsp-location.yaml' exist.

eng/scripts/typespec_build.ps1
raych1: @JiaqiZhang-Dev pls ensure data plane cases work with the new changes in SDK generation.

raych1:
@JiaqiZhang-Dev are automation and release aligned to use the same command to generate sdk?

eng/tools/generator/cmd/v2/common/generation.go
JiaqiZhang-Dev: For exist service, use `init` command can refresh `tsp-location.yaml` according to `generatedInput.json`, so that the brand-new service and exist service can both used `go generate` to generated code.

eng/tools/generator/cmd/v2/common/cmdProcessor.go
JiaqiZhang-Dev: I reported an issue about `tsp-client` exit code: https://github.com/Azure/azure-sdk-tools/issues/10234

lmazuel:
Why are we doing this in the first place? tsp-client should be the only way to generate things. I just got a discussion @richardpark-msft and he can provide some of the details (as I'm about to leave in vacation for a few days)

richardpark-msft:
> Why are we doing this in the first place? tsp-client should be the only way to generate things. I just got a discussion @richardpark-msft and he can provide some of the details (as I'm about to leave in vacation for a few days)

I think the key bits are:
- Other languages have accomplished what we're doing, in Go generate, by putting these commands into their actual emitters. I'm guessing most of these "post commands" are written in their own language, so they can take advantage of language features.

  So this is what I'm doing today in aznamespaces, for instance:

  - go generate, which runs:
    - go:generate tsp-client update
    - go:generate goimports -w .
    - go:generate go run ./internal/generate
    - go:generate goimports -w .

  So if we inverted this, and had our emitter do it I think it'd look more like this:
  - tsp-client update
    - Go language emitter
      - emit stuff happens
      - gofmt -w ./..
      - post `go generate` command runs

   @gracewilcox also has some things she does as well, so curious if that workflow would work for her as well.

With all that said, @jhendrixMSFT knows all of our code generation ins-and-outs, but this is my basic understand of how other people have structured their process so `tsp-client` can be the gateway command for everyone.

gracewilcox:
I'm pretty similar to Richard.

For `azkeys`, go generate runs:
* //go:generate tsp-client update
* //go:generate go run ./internal/transforms.go
* //go:generate goimports -w .

But something like `azadmin` is more complicated. For that, go generate runs:
* //go:generate tsp-client update --output-dir ./settings
* //go:generate tsp-client update --output-dir ./rbac
* //go:generate tsp-client update --output-dir ./backup
* //go:generate go run ./internal/generate/transforms.go
* //go:generate gofmt -w .
* //go:generate rm ./backup/constants.go

If we can figure out how to do all this in the emitter, I'll be impressed.

tadelesh:
so, go emitter should do:
1. run postprocess customization code
2. reorganize import: `goimports`
3. format code: `gofmt`

then we could use emitter e2e for go data plane generation scenario. @JiaqiZhang-Dev we could discuss it in this week's sync.

JiaqiZhang-Dev:
It’s better to have it in tsp-client which could be adopted across all languages. Catalina will track it: https://github.com/Azure/azure-sdk-tools/issues/10366

tadelesh:
close as we do not plan to do this anymore after discussion.

PR #24323, opened 2025-03-26
BearerTokenPolicy observes issuer's suggested token refresh time

chlowell:
Continuing on from #24268, this adds AccessToken.RefreshOn and updates BearerTokenPolicy to observe it when deciding whether to request a token. Part of #22837

### DIFF
--- a/sdk/azcore/runtime/policy_bearer_token.go
+++ b/sdk/azcore/runtime/policy_bearer_token.go
@@ -51,6 +51,14 @@ func acquire(state acquiringResourceState) (newResource exported.AccessToken, ne
 	return tk, tk.ExpiresOn, nil
 }
 
+// shouldRefresh determines whether the token should be refreshed. It's a variable so tests can replace it.
+var shouldRefresh = func(tk exported.AccessToken, _ acquiringResourceState) bool {
+	if tk.RefreshOn.IsZero() {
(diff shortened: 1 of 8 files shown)


sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -51,6 +51,14 @@ func acquire(state acquiringResourceState) (newResource exported.AccessToken, ne
 	return tk, tk.ExpiresOn, nil
 }
 
+// shouldRefresh determines whether the token should be refreshed. It's a variable so tests can replace it.
+var shouldRefresh = func(tk exported.AccessToken, _ acquiringResourceState) bool {
+	if tk.RefreshOn.IsZero() {
+		return tk.ExpiresOn.Add(-5 * time.Minute).Before(time.Now())
+	}
+	return tk.RefreshOn.Before(time.Now())
```
jhendrixMSFT, line 59: Do we not need a window here, or is that baked into the value in `RefreshOn`?

chlowell: I'll add a comment. We take the value of RefreshOn literally because it defines the authority's suggested refresh window

jhendrixMSFT (approved):

PR #24333, opened 2025-03-27
fix- get webapp keys as key value map

munishsuri:
Response from List Keys API is in the form of key value pair via API: https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}/functions/{functionName}/listkeys?api-version=2024-04-01

Screenshot 2025-03-27 at 3 42 06 PM

Current Parsing Logic takes this body in the **StringDictionary** struct, which will return in empty response always, hence changing it to **Properties** of **StringDictionary** , which will result in a key value map


tadelesh:
@munishsuri thanks for your contribution. since our sdk is auto generated from spec, we do not allow any manual code change. it looks like your problem is a wrong spec definition. could you open an issue in the spec repo, and let service team help to fix this [spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/web/resource-manager/Microsoft.Web/stable/2024-04-01/WebApps.json) first?

munishsuri:
Thanks @tadelesh for the comments, really appreciate it.
I will check with spec repo and proceed from there

PR #24345, opened 2025-04-01
[azservicebus] Removing the unneeded RenewSessionLock() call when accepting a session

### LINKED ISSUE #19745: [azservicebus] Eliminate unneeded RenewSessionLock call when accepting a session
As a workaround we're doing a RenewSessionLock call when accepting new session receivers. 

This adds some overhead (minimal) but we can eliminate it entirely when go-amqp starts exposing the Properties returned from the Attach frame (issue here: https://github.com/Azure/go-amqp/issues/220).

One complication for this is that we do rely on the DETACH frame error so we can return an appropriate error code: https://github.com/Azure/azure-sdk-for-go/blob/f8dc74a32154b6e3080de7d86658ac69e37a3eb2/sdk/messaging/azservicebus/internal/errors.go#L67

So we'll want to take that into account and make sure we maintain that contract.

richardpark-msft:
Removing the unneeded RenewSessionLock() call when accepting a session. We can do this since go-amqp now returns the link properties on ATTACH.

Fixes #19745


jhendrixMSFT (approved):

richardpark-msft (commented):
/azp run go - azservicebus

jhendrixMSFT (approved):

PR #24377, opened 2025-04-04
Resolve crashes in `analyze`

scbedd:
[Discovered in azidentity](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4721711&view=results)

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -135,16 +135,16 @@ steps:
           not(endsWith(variables['Build.Repository.Name'], '-pr'))
         )
 
-  - task: Powershell@2
-    displayName: 'Dependency Check'
+  - pwsh: |
+      Get-ChildItem -R $(Build.SourcesDirectory) | % { Write-Host $_.FullName }
+    displayName: "Show the visible files. Why can't we retrieve this?"
+
+  - pwsh: |
+      ./eng/scripts/Invoke-DependencyCheck.ps1 "$(Packages)"
+    displayName: Dependency Check
     condition: ne(variables['Skip.DependencyCheck'], 'true')
     env:
       GO111MODULE: 'on'
-    inputs:
-      targetType: filePath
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -311,13 +307,9 @@ steps:
...
 
-  - task: PowerShell@2
+  - pwsh: |
+      ./eng/scripts/Invoke-SmokeTests.ps1 "$(ChangedServices)"
     displayName: 'Run Nightly SmokeTests'
-    inputs:
-      targetType: 'filePath'
-      filePath: ./eng/scripts/Invoke-SmokeTests.ps1
-      pwsh: true
-      arguments: '$(ChangedServices)'
```
scbedd, line 320: This was converting "blah,blah2" to `[blah, blah2]` under the hood, which was messing with the filters. Just need it to pass as a string!

scbedd: Hmmm. Maybe it's because of the single quotes instead of the double, but I am confident this works so let's just go with this for now.

benbp (approved):

eng/scripts/Language-Settings.ps1
```diff
@@ -154,6 +154,7 @@ function Get-AllPackageInfoFromRepo($filterString)
     $searchPaths = ResolveSearchPaths $filterString
 
     foreach ($searchPath in $searchPaths) {
+      Write-Host "Searching for go modules in $searchPath"
```
weshaggard, line 157: How noisy does this make discovery? In general I've tried to keep all logging to verbose only in the package discovery because it can be noisy when doing things like prepare-release script runs for example.

scbedd: Ah you know, good point.

scbedd: I'm making a change for proxy and non-proxy in the pullrequest builds, will remove this then.

PR #24383, opened 2025-04-05
[azeventhubs] Optimize checkpoint listing to avoid wasted storage calls

### LINKED ISSUE #22783: [azeventhubs] Processor optimization to not list checkpoints when no ownership changes
### Feature Request

I noticed that in the `Processor#dispatch` function, that is called every 10s by default, [all the checkpoints are being fetched](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azeventhubs/processor.go#L305) but during most of the time (i.e. when no ownership is being re-shuffled) those checkpoints end up just not being used. There is an opportunity to save a lot of calls to the checkpoint store, and save $. Could it be optimized so checkpoints are only fetched when necessary?

richardpark-msft:
The Processor was listing all checkpoints on each dispatch round. We don't actually need to do that all the time, however, since partition ownership eventually reaches a steady state, and we no longer need to retrieve the checkpoints.

Now we grab them on-demand, and only when we actually establish new ownership on a partition.

Fixes #22783


sdk/messaging/azeventhubs/processor_test.go
```diff
@@ -529,8 +539,13 @@ func testWithLoadBalancer(t *testing.T, loadBalancerStrategy azeventhubs.Process
...
 	t.Logf("Checkpoint store created")
-	checkpointStore, err := checkpoints.NewBlobStore(cc, nil)
-	require.NoError(t, err)
+	var checkpointStore *countingCheckpointStore
+	{
+		inner, err := checkpoints.NewBlobStore(cc, nil)
+		require.NoError(t, err)
+
+		checkpointStore = &countingCheckpointStore{BlobStore: inner, listed: 0}
+	}
```
chlowell, line 548: TIL Go allows anonymous blocks 🤯

chlowell (approved):

sdk/messaging/azeventhubs/processor_test.go
richardpark-msft: Yeah, they don't work for defer's - those still go to the function level, but you can scope variables, etc..

PR #24395, opened 2025-04-07
Prepare changelog and version file for release of the internal package

scbedd:
We need to get the version from #24394 available so that `go - pullrequest` can stop crashing on every package that has `UsePipelineProxy` set to false.


richardpark-msft (approved):
Approving so we can get the ball moving.

chlowell (approved):

PR #24413, opened 2025-04-09
Resolve `linting` issues in eng/ PR

semick-dev:
I don't have access to an MS machine so I'm PRing this to see if I can get clean CI before I'm also away from a personal laptop.


semick-dev:
Ah because of branch names the triggers won't work out. I'll find a different avenue. Going to close this PR.

PR #24419, opened 2025-04-09
Fix azcontainerregistry live test authentication

chlowell:
It appears making post-deploy scripts "self-contained" (https://github.com/Azure/azure-sdk-tools/pull/10099) made this pipeline's script ineffective in that its `az login` no longer applies during the test run (this module is unique in using the CLI during live test runs). The quick fix here is to move the login step to a pre-deployment script; deployment is too fast for this to cause problems. I kept the no-op post script because `PersistOidcToken: true`, which we need to log in to the CLI in either a pre- or post-script, makes a post-deploy script mandatory.

@benbp I imagine there's a better way to address this but finding one isn't urgent because only this module needs a logged-in CLI to run live tests

### DIFF
--- a/sdk/containers/azcontainerregistry/CHANGELOG.md
+++ b/sdk/containers/azcontainerregistry/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.2.3 (2025-04-09)
+## 0.2.3 (2025-04-15)
 
 ### Other Changes
 * Default audience of Azure Container Registry of all clouds to https://containerregistry.azure.net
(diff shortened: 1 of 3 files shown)


benbp:
@chlowell Ah, good catch on this. I think the issue is that the azure cli task calls `az account clear` during cleanup. I think we originally worked around this issue with the approach you've added here. If we run az login in the pre script then it runs within an Azure Powershell task context which won't clear the az login. The problem with that approach was the timeouts for long deploys, hence the self-contained solution.

I think we can keep your fix as it's in line with our current approach elsewhere. The alternative for you is to update the test to first run `az login --federated-token` with the `ARM_OIDC_TOKEN` environment variable that we export, as it won't be at risk of expiry for this pipeline.

benbp (approved):

RickWinter (approved):

PR #24442, opened 2025-04-14
WorkloadIdentityCredential enhancement for AKS FIC limit

chlowell:
This is the client part of a workaround for the per-application federated credentials limit. The feature is conceptually simple: instead of sending token requests to Entra ID, the client sends them to an endpoint specified by AKS, which also provides TLS configuration for that endpoint. The implementation is a little complex though, because AKS may provide configuration in a file whose content it may change.

### DIFF
--- a/sdk/azidentity/workload_identity.go
+++ b/sdk/azidentity/workload_identity.go
@@ -8,8 +8,12 @@ package azidentity
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
+	"net/http"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -94,6 +98,24 @@ func NewWorkloadIdentityCredential(options *WorkloadIdentityCredentialOptions) (
 		ClientOptions:              options.ClientOptions,
 		DisableInstanceDiscovery:   options.DisableInstanceDiscovery,
 	}
+	sni := os.Getenv("AZURE_AURELIA_SNI_NAME")
+	host := os.Getenv("AZURE_AURELIA_TOKEN_ENDPOINT")
+	if sni != "" && host != "" {
+		co := caco.ClientOptions
+		co.Transport = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					RootCAs:    aksSNIPolicyCA,
+					ServerName: sni,
+				},
+			},
+		}
+		p := &aksSNIPolicy{
+			h:  host,
+			pl: runtime.NewPipeline(module, version, runtime.PipelineOptions{}, &co),
+		}
+		caco.ClientOptions.PerRetryPolicies = append(caco.ClientOptions.PerRetryPolicies, p)
+	}
 	cred, err := NewClientAssertionCredential(tenantID, clientID, w.getAssertion, &caco)
 	if err != nil {
 		return nil, err
@@ -138,4 +160,21 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
 		defer w.mtx.RUnlock()
 	}
 	return w.assertion, nil
+}
+
+// aksSNIPolicyCA enables tests to configure aksSNIPolicy to trust a test server's cert
+var aksSNIPolicyCA *x509.CertPool
+
+type aksSNIPolicy struct {
+	h  string
+	pl runtime.Pipeline
+}
+
+func (a *aksSNIPolicy) Do(req *policy.Request) (*http.Response, error) {
+	if r := req.Raw(); strings.HasSuffix(r.URL.Path, "/token") {
+		r.URL.Host = a.h
+		r.Host = ""
+		return a.pl.Do(req)
+	}
+	return req.Next()
 }
--- a/sdk/azidentity/workload_identity.go
+++ b/sdk/azidentity/workload_identity.go
@@ -7,10 +7,12 @@
 package azidentity
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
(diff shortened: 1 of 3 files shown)


sdk/azidentity/workload_identity.go
```diff
@@ -7,18 +7,31 @@
...
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/errorinfo"
 )
 
-const credNameWorkloadIdentity = "WorkloadIdentityCredential"
+const (
+	aksCAData                = "AZURE_KUBERNETES_CA_DATA"
+	aksCAFile                = "AZURE_KUBERNETES_CA_FILE"
+	aksSNIName               = "AZURE_KUBERNETES_SNI_NAME"
+	aksTokenEndpoint         = "AZURE_KUBERNETES_TOKEN_ENDPOINT"
```
enj, line 32: nit: the SDK changes are generic and are meant to be usable on any kube running on Azure, i.e. AKS, CAPZ, etc.

```suggestion
	azureCAData                = "AZURE_KUBERNETES_CA_DATA"
	azureCAFile                = "AZURE_KUBERNETES_CA_FILE"
	azureSNIName               = "AZURE_KUBERNETES_SNI_NAME"
	azureTokenEndpoint         = "AZURE_KUBERNETES_TOKEN_ENDPOINT"
```

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
 	}
 	return w.assertion, nil
 }
+
+// aksTokenRequestPolicy redirects token requests to the AKS token endpoint, sending them via its
+// own HTTP client. It sends all other requests unchanged, via the pipeline's configured transport.
+type aksTokenRequestPolicy struct {
+	// c is configured for the AKS token endpoint
+	c *http.Client
```
enj, line 167: Will the request seen by this client still contain the correct headers such as user agent, etc?

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
...
+	serverName := os.Getenv(aksSNIName)
+	if host == "" || serverName == "" {
+		// the AKS feature isn't enabled for this process
+		return nil, nil
+	}
+	b := []byte(os.Getenv(aksCAData))
+	f := os.Getenv(aksCAFile)
+	switch {
+	case len(b) == 0 && len(f) == 0:
+		return nil, fmt.Errorf("no value found for %s or %s", aksCAData, aksCAFile)
```
enj, line 184: These are optional, system roots should be used if these are not given.

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
...
+	// so we must read that file and maybe create a new client
+	b, err := os.ReadFile(a.caFile)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse %s: %s", aksCAFile, err)
+	}
+	if len(b) == 0 {
+		return nil, errors.New(aksCAFile + " specifies an empty file")
+	}
+	if !bytes.Equal(b, a.ca) {
+		a.ca = b
```
enj, line 258: This should only be set after `a.c` is successfully set.

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
...
+type aksTokenRequestPolicy struct {
+	// c is configured for the AKS token endpoint
+	c *http.Client
+	// ca trusted by c
+	ca                       []byte
+	caFile, host, serverName string
+}
+
+func newAKSTokenRequestPolicy() (*aksTokenRequestPolicy, error) {
+	host := os.Getenv(aksTokenEndpoint)
```
enj, line 174: This is not a host, it is a full URL that can contain a path.

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
...
+	return p, nil
+}
+
+func (a *aksTokenRequestPolicy) Do(req *policy.Request) (*http.Response, error) {
+	if r := req.Raw(); strings.HasSuffix(r.URL.Path, "/token") {
+		c, err := a.client()
+		if err != nil {
+			return nil, errorinfo.NonRetriableError(err)
+		}
+		r.URL.Host = a.host
```
enj, line 201: This is not correct.

sdk/azidentity/workload_identity.go
```diff
@@ -139,3 +159,115 @@ func (w *WorkloadIdentityCredential) getAssertion(context.Context) (string, erro
...
+	}
+	p := &aksTokenRequestPolicy{caFile: f, ca: b, host: host, serverName: serverName}
+	if _, err := p.client(); err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+func (a *aksTokenRequestPolicy) Do(req *policy.Request) (*http.Response, error) {
+	if r := req.Raw(); strings.HasSuffix(r.URL.Path, "/token") {
```
enj, line 196: I do not believe this assumption is correct.

enj (changes requested):
First pass.

PR #24465, opened 2025-04-18
[Cosmos] Pass request context through on GEM requests

### LINKED ISSUE #24351: [azcosmos] Request context lost in cosmos_global_endpoint_manager_policy.go
### Bug Report

Our RP relies on `azcosmos` from `azure-sdk-for-go` for cosmos DB operations. We are currently using `azcosmos` version `0.3.3`, but in an effort to resolve security related vulnerabilities, we are updating outdated packages. When bumping from `azcosmos` `0.3.3` to `1.3.0`, our cosmos DB operations fail with a nil pointer exception, when we try to extract apitracking information from the request context.

After closer debugging and tracing, it seems we are losing context due to `cosmos_global_endpoint_manager_policy.go` passing in `context.Background()` here instead of the context from the request: [azure-sdk-for-go/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go at 799c8579d2d8f32d3f30457cd6da736be08d9c6b · Azure/azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/799c8579d2d8f32d3f30457cd6da736be08d9c6b/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go)
 
The code path / tests succeed when I change this code locally and pass along the context from the request `req.Raw().Context()` like this:

Image

- package + version: `github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos v1.3.0`
- go version: `go version go1.24.1 linux/amd64`

Can repro / test with the following go code (succeeds with azcosmos 0.3.3 -> 0.3.6, fails with azcosmos 1.0.0 and later, due to [azure-sdk-for-go/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go at 799c8579d2d8f32d3f30457cd6da736be08d9c6b · Azure/azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/799c8579d2d8f32d3f30457cd6da736be08d9c6b/sdk/data/azcosmos/cosmos_global_endpoint_manager_policy.go):
```
package main

import (
	"context"
	"net/http"
	"time"

	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
	"github.com/sirupsen/logrus"
	apisubscriptions "goarc.io/arcplatform/core/api/subscriptions"
	"goarc.io/arcplatform/core/apitracking"
	"goarc.io/arcplatform/core/azclients/clientoptions"
	"goarc.io/arcplatform/core/azclients/databasev2"
	"goarc.io/arcplatform/core/log"
	"goarc.io/arcplatform/core/tracing"
	"goarc.io/arcplatform/core/tracing/noop"
)

func main() {
	tracing.Client = noop.New()
	ctx := context.Background()

	at := apitracking.APITracking{
		OperationID: "kunaltest",
	}
	ctx = apitracking.WithAPITracking(ctx, &at)
	logger := log.GetLogger()
	ctx = log.WithLogger(ctx, logger)

	cred, err := azidentity.NewAzureCLICredential(&azidentity.AzureCLICredentialOptions{})
	if err != nil {
		logger.Fatalf("Failed to create Azure CLI credential: %v", err)
	}

	cosmosDBInstanceURI := "https://e2e-tesla-dbvalidator.documents.azure.com:443/"
	client, err := azcosmos.NewClient(cosmosDBInstanceURI, cred, GetCosmosDBClientOptions(logger))
	if err != nil {
		logger.Fatalf("Failed to create CosmosDB client: %v", err)
	}

	cosmosDBClient := databasev2.CreateNewCosmosDBClient(client)
	subscriptions, err := databasev2.NewSubscriptions(logger, cosmosDBClient, "tesla-e2e", "tesla-db-atacamenan-Subscriptions")
	if err != nil {
		logger.Fatalf("Failed to create subscriptions client: %v", err)
	}

	sub, err := subscriptions.Create(ctx, &apisubscriptions.Subscription{
		ID: "foo3",
	})
	if err != nil {

		logger.Fatalf("Failed to create subscription: %v", err)

	}

	foo, err := subscriptions.Get(ctx, sub.ID)
	if err != nil {
		logger.Fatalf("Failed to get subscription: %v", err)
	}
	logger.Infof("Subscription: %+v", foo)
}

func GetCosmosDBClientOptions(log *logrus.Entry) *azcosmos.ClientOptions {
	log.Info("Creating new CosmosDB v2 client options")
	tr := clientoptions.GetHTTPTransport(log, true)
	tr.DialContext = clientoptions.NewFallbackDialer(clientoptions.DefaultFallbackDNSServerAddresses, log).DialContext
	opt := &azcosmos.ClientOptions{
		ClientOptions: policy.ClientOptions{
			Transport: &http.Client{
				Transport: tr,
			},
			// We call CosmosDB more frequently and its latency is much less than ARM,
			// so we use a different retry policy than ARM.
			Retry: policy.RetryOptions{
				// Create/Update/Delete operations might get ItemAlreadyExists error on a second retry
				StatusCodes: []int{
					http.StatusRequestTimeout,      // 408
					http.StatusTooManyRequests,     // 429
					http.StatusInternalServerError, // 500
					http.StatusBadGateway,          // 502
					http.StatusServiceUnavailable,  // 503
					http.StatusGatewayTimeout,      // 504
					// Adding 403 to work around random 403 from CosmosDB (see GitHub issue)
					http.StatusForbidden, // 403
				},
				MaxRetries:    3,
				TryTimeout:    5 * time.Second, // Maximum server-side latency in slow regions
				RetryDelay:    1 * time.Second, // Initial retry delay
				MaxRetryDelay: 2 * time.Second, // Exponential back-off max delay
			},
			PerRetryPolicies: []policy.Policy{
				&clientoptions.IdentifierHeadersPolicy{},
				&clientoptions.HttpTraceLoggerPolicy{},
			},
		},
	}
	log.Infof("CosmosDB client v2 options: %+v", opt)
	return opt
}
```

analogrelay:
This fixes #24351 by passing the Context associated with the request in to the GEM when calling `Update` _during_ the processing of a request. This ensures that any key/value pairs stored in that Context flow through the GEM requests as well, which ensures telemetry spans line up properly (among other things).

As discussed in the issue, I didn't flow the full context through. Specifically, the context provided when calling `Update` **excludes** the cancellation signal from the original request context. That means if the request is cancelled, the GEM update will **not** be cancelled and will complete fully.

The test I wrote failed as expected without this change. Initially, it fails because the expected context value is not present in the context when the GEM requests are made (the original issue reported by the bug). The test also validates that a cancellation signal is _ignored_ by the GEM update, but _honored_ by the rest of the pipeline.

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Release History
 
-## 1.3.1-beta.1 (Unreleased)
+## 1.4.0 (Unreleased)
 
 ### Features Added
+
 * Added an initial API for integrating an external client-side Query Engine with the Cosmos DB Go SDK. This API is unstable and not recommended for production use. See [PR 24273](https://github.com/Azure/azure-sdk-for-go/pull/24273) for more details.
 
 ### Breaking Changes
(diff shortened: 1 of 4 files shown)


Pilchie (approved):
Not an expert at the context stuff here, but this seems reasonable to me.

Pilchie (approved):

Issue #24460, opened 2025-04-17
Need to set API version on azquery or access non-regional ARM metrics from azmetrics.


PCaponetti:
### Feature Request

I need to query metrics for Microsoft.Resources/Subscription. This does not seem to be supported in the batch API, and it seems the [single metrics ARM API](https://learn.microsoft.com/en-us/rest/api/monitor/metrics/list?view=rest-monitor-2023-10-01&tabs=HTTP#code-try-0) only supports it on later API versions.

If I am right in assuming that I can only access the single metrics ARM API through the azquery package, could you add setting API version to that package as you did for [azmetrics](https://github.com/Azure/azure-sdk-for-go/issues/23819)?

CC: @gracewilcox

gracewilcox:
Thank you for bringing this to our attention! We'll work on generating from the [latest version](https://github.com/Azure/azure-sdk-for-go/issues/24469) and add API version override.

gracewilcox:
Also @PCaponetti , the complete set of monitor ARM API's can be found in our [`armmonitor`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/monitor/armmonitor) package. 

The method you're referencing would be [MetricsClient.List](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/monitor/armmonitor#MetricsClient.List)

Our monitor SDK's are actually in a [transition](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/monitor/query/MIGRATION.md), so this would be the recommended way of using our metrics ARM endpoints.

PCaponetti:
Any thoughts on when a new version will be released that I could install with this change?

gracewilcox:
Hi @PCaponetti! We are preparing to release `azquery` with the upcoming changes on our next scheduled ship day, which is set for Tuesday, May 6.

PR #24487, opened 2025-04-22
[azquery] allow API version override

gracewilcox:
closes https://github.com/Azure/azure-sdk-for-go/issues/24460

Allows customers to override the API version for metrics clients.

### DIFF
--- a/sdk/monitor/azquery/metrics_client_test.go
+++ b/sdk/monitor/azquery/metrics_client_test.go
@@ -8,11 +8,14 @@ package azquery_test
 
 import (
 	"context"
+	"net/http"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (approved):

sdk/monitor/azquery/metrics_client_test.go
```diff
@@ -156,3 +159,31 @@ func TestNewListNamespacesPager_Failure(t *testing.T) {
 	}
 
 }
+
+func TestMetricsAPIVersion(t *testing.T) {
+	apiVersion := "2023-10-01"
```
chlowell, line 164: consider using a value that obviously isn't a real version so it's clear this test doesn't redundantly specify the default

chlowell (approved):

PR #24554, opened 2025-05-01
Change azbatch live test region

chlowell:
Our test subscription has a higher core quota in brazilsouth


RickWinter (approved):
CoPilot had some decent suggestions

richardpark-msft (approved):

PR #24611, opened 2025-05-10
Updated with latest update for 2025-05 events.

richardpark-msft:


sdk/messaging/eventgrid/azsystemevents/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.6.2 (2025-05-06)
+## 0.6.2 (TBD)
```
jhendrixMSFT, line 3: Looks like there are breaking changes so this should be `0.7.0`.

richardpark-msft: Good catch. Looks like that field rename should NOT have happened, so I'm fixing the typespec.

richardpark-msft:
@jhendrixMSFT, should be good now. No rename needed for that details field.

sdk/messaging/eventgrid/azsystemevents/models.go
```diff
@@ -148,7 +148,7 @@ type ACSCallParticipantProperties struct {
 	CommunicationIdentifier *CommunicationIdentifierModel
 
 	// The role of the participant
-	Role *string
+	Role *ACSCallParticipantRoleKind
```
jhendrixMSFT, line 151: This is also a breaking change and will require a minor version bump.

sdk/messaging/eventgrid/azsystemevents/CHANGELOG.md
richardpark-msft: Updated. Also, added in an entry for the breaking change.

richardpark-msft:
@jhendrixMSFT, all feedback addressed, and breaking change doc'd.

jhendrixMSFT (approved):

PR #24668, opened 2025-05-22
Update all requirements for `armworkloadsapvirtualinstance`

scbedd:
Addressing this [failure.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4885827&view=logs&j=f8f2e80b-3520-5144-e95f-ffb4c3cddb31&t=070d95a1-4cc8-59c0-5578-bff4a021db8b)

I believe this should do it for the `go - pullrequest` failure. We probably need to update the generation bits someplace that I'm not aware of.


sdk/resourcemanager/workloadssapvirtualinstance/armworkloadssapvirtualinstance/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.23.0
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.17.1
```
jhendrixMSFT, line 6: Since we're already mucking about here, we should update all of these dependencies. Easiest way is to `go get -u all` followed by `go mod tidy`.

scbedd: Got it will execute.

jhendrixMSFT (approved):

tadelesh (approved):

tadelesh:
@JiaqiZhang-Dev will help to update the script to add `go get -u all` to our release script. then i think we need a refreshment for all arm lib.

JiaqiZhang-Dev (approved):

scbedd:
@tadelesh the changelog update that I had was correct, as I had updated the changelog to align with the new patch version of the package (no changes without some sort of version bump).

Is there a reason you removed it? I will be the first to confess I'm not aware of all the restrictions of the mgmt automation.

tadelesh:
@scbedd sorry, i shall ask you before commit the change. for mpg, we do not add changelog placeholder for next release, since they are all auto generated. the current ci error is i forgot to revert the version change in `constants.go`. usually, for dep upgrade, we will choose either:
1. bump version and release immediately if it is a critical upgrade.
2. just update the code. keep version and changelog not changed. the upgrade will be released with the next service version. this is the most common case.

benbp (approved):

Issue #24688, opened 2025-05-26
[bug] changelog generated by sdk automation shows wrong breaking change num


tadelesh:
we have a ci log [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4890158&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692).
```
info	[Changelog] ### Features Added
info	[Changelog] - Type of `ExpressionV2.Value` has been changed from `*string` to `any`
info	[Changelog] - New enum type `AmazonRdsForOracleAuthenticationType` with values `AmazonRdsForOracleAuthenticationTypeBasic`
info	[Changelog] - New enum type `ImpalaThriftTransportProtocol` with values `ImpalaThriftTransportProtocolBinary`, `ImpalaThriftTransportProtocolHTTP`
info	[Changelog] - New field `AuthenticationType`, `CryptoChecksumClient`, `CryptoChecksumTypesClient`, `EnableBulkLoad`, `EncryptionClient`, `EncryptionTypesClient`, `FetchSize`, `FetchTswtzAsTimestamp`, `InitialLobFetchSize`, `InitializationString`, `Server`, `StatementCacheSize`, `SupportV1DataTypes`, `Username` in struct `AmazonRdsForLinkedServiceTypeProperties`
info	[Changelog] - New field `DataSecurityMode` in struct `AzureDatabricksLinkedServiceTypeProperties`
info	[Changelog] - New field `EnableServerCertificateValidation` in struct `HiveLinkedServiceTypeProperties`
info	[Changelog] - New field `EnableServerCertificateValidation`, `ThriftTransportProtocol` in struct `ImpalaLinkedServiceTypeProperties`
info	[Changelog] - New field `EnableServerCertificateValidation` in struct `SparkLinkedServiceTypeProperties`
info	[Changelog] Total 1 breaking change(s), 10 additive change(s).
```
this changelog is generated by our sdk automation tool [here](https://github.com/Azure/azure-sdk-for-go/blob/778c4aa82f97557c632af97b65f43b1922cfd4de/eng/tools/generator/cmd/v2/common/generation.go#L274). we have a `FilterChangelog` logic in in. but it seems the filter logic has some issues for counting each kind of the changes.

PR #24691, opened 2025-05-27
Fix generator bug

JiaqiZhang-Dev:
resolve: https://github.com/Azure/azure-sdk-for-go/issues/24688


eng/tools/generator/cmd/v2/common/changelog.go
```diff
@@ -95,6 +95,7 @@ func writeChangelogForPackage(r *report.Package) string {
 
 	// write additional changes
 	toAny := typeToAny(r.BreakingChanges, true)
+	deleteTypeToAny(r.BreakingChanges)
```
tadelesh, line 98: if we only remove it from breaking change, will it impact the count of additive?

tadelesh: after rethink, i believe this logic should move to the changelog filter hock. the type extension should also be an additive change.

JiaqiZhang-Dev: done

eng/tools/internal/report/report.go
```diff
@@ -15,7 +15,7 @@ import (
```
tadelesh, line 15: could you double confirm if this is used by other tool?

eng/tools/generator/cmd/v2/common/changelog.go
```diff
@@ -475,6 +486,42 @@ func typeToAny(b *report.BreakingChanges, flag bool) []string {
 	return items
 }
 
+func deleteTypeToAny(b *report.BreakingChanges) {
```
tadelesh: is this func still needed?

eng/tools/generator/cmd/v2/common/changelog.go
```diff
@@ -275,7 +286,7 @@ func getSignatureChangeItems(b *report.BreakingChanges) []string {
 		}
 	}
 	// write struct changes
-	items = append(items, typeToAny(b, false)...)
+	items = append(items, typeToAny(b, countTypeToAny)...)
```
tadelesh: since you have already done this in filtering logic, do we still need this?

JiaqiZhang-Dev: I refactor the name of `typeToAny` to avoid ambiguity, and remove the `flag` parameter of this function.

eng/tools/generator/cmd/v2/common/changelog.go
JiaqiZhang-Dev: Don't need it, removed

eng/tools/internal/report/report.go
JiaqiZhang-Dev: I search for all code of Azure organization; It also used by `apidiff` tool (https://github.com/search?q=org%3AAzure+github.com%2FAzure%2Fazure-sdk-for-go%2Feng%2Ftools%2Finternal%2Freport&type=code). 

Because `internal` is exported as a go module, tools are used this under version control, so the code refactor not impact them immediately. It may only be affected when upgrading the internal package version.

tadelesh (approved):

PR #24820, opened 2025-06-19
Update pool to azsdk-pool and linux image to ubuntu 24

benbp:
Update pool to azsdk-pool and linux image to ubuntu 24

### DIFF
--- a/eng/pipelines/templates/variables/image.yml
+++ b/eng/pipelines/templates/variables/image.yml
@@ -2,18 +2,18 @@
 
 variables:
   - name: LINUXPOOL
-    value: azsdk-pool-mms-ubuntu-2004-general
+    value: azsdk-pool
   - name: LINUXNEXTPOOL
-    value: azsdk-pool-mms-ubuntu-2204-general
+    value: azsdk-pool
   - name: WINDOWSPOOL
-    value: azsdk-pool-mms-win-2022-general
+    value: azsdk-pool
 
   - name: LINUXVMIMAGE
-    value: azsdk-pool-mms-ubuntu-2004-1espt
+    value: ubuntu-24.04
   - name: LINUXNEXTVMIMAGE
-    value: azsdk-pool-mms-ubuntu-2204-1espt
+    value: ubuntu-24.04
   - name: WINDOWSVMIMAGE
-    value: azsdk-pool-mms-win-2022-1espt
+    value: windows-2022
 
   # Values required for pool.os field in 1es pipeline templates
   - name: LINUXOS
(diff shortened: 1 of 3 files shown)


mikeharder (approved):

eng/pipelines/templates/variables/image.yml
```diff
@@ -2,18 +2,18 @@
...
-    value: azsdk-pool-mms-ubuntu-2204-general
+    value: azsdk-pool
   - name: WINDOWSPOOL
-    value: azsdk-pool-mms-win-2022-general
+    value: azsdk-pool
 
   - name: LINUXVMIMAGE
-    value: azsdk-pool-mms-ubuntu-2004-1espt
+    value: ubuntu-24.04
   - name: LINUXNEXTVMIMAGE
```
weshaggard, line 13: nit: We can probably remove the NEXT image and pool if we aren't using it.

weshaggard (approved):

eng/pipelines/templates/variables/image.yml
benbp: I had assumed we wanted to push things forward to 24 fully as opposed to 22 and 24. My thinking was when 26 comes out, next pool becomes 26.

mikeharder: In the other PRs moving to azsdk-pool, I think we kept the LINUXNEXT vars, even if they just point to the same version as the standardard LINUX vars, or if they aren't used in the repo, etc.  For consistency, I think we should do the same for this PR.

However, I do agree the LINUXNEXT vars appeared like mostly noise / dead-code at this point.  In general, I think it's simpler (and sufficient) to only build and test on a single Ubuntu version, unless we specifically want coverage on multiple versions (e.g. C++).  I would support removing the LINUXNEXT vars from most repos, in a separate round of PRs.

benbp: @mikeharder Last time we tested out a new image version, we let it soak in a few pipelines for several weeks if I remember correctly. I would assume LINUXNEXT makes that process easier (vs. hardcoding the version)? Or would you take a different approach?

weshaggard: My care level is low, but I just have a tendency to remove things that aren't used, we can always add them back if we make usages of them.

mikeharder: In the recent updates to ubuntu 24, if the PR updating the OS version was green, we assumed this was sufficient and merged the PR without testing it live for a few weeks.  I believe any issues were detected immediately, and letting it soak a few weeks wouldn't have helped anything.

My approach to upgrade to ubuntu-26.  Soon after ubuntu-26 is supported:

1. Search/replace "ubuntu-24.04" with "ubuntu-26.04"
2. If PR is green, merge
3. Deal with remaining problems in prod, rollback to 24 if needed
4. If any problems found in prod, add PR test to prevent this next time

benbp: Sounds good then, I'll add a task to remove the next vm image.

PR #24823, opened 2025-06-20
Address CodeQL findings

chlowell:
...by ignoring some test code and removing pseudorandom input to a crypto algorithm (which wasn't dangerous but changing the input is easier than suppressing the warning)


CodeQL.yaml
jhendrixMSFT, line 1: How was the contents of this file determined, and is there a link that explains it in more detail?

jhendrixMSFT (approved):

CodeQL.yaml
jhendrixMSFT: Discussed offline.

PR #24833, opened 2025-06-23
cosmos: enable support for creating containers with vector and full-text policies

analogrelay:
This PR adds support for setting vector embedding policies in the Cosmos SDK. The PR **does not** bring any support for **executing** vector queries yet, but it does allow for _creating_ containers with the appropriate configuration.

I'll be merging this in to the `1.5.0.beta` branch as well, since I'm hoping to use this along with the query engine integration to start bringing vector query support. However, it's not something that needs to be in preview, so I'm putting it in `main` for now.


Pilchie (approved):
Next up, full-text policy and index support.

analogrelay:
That's a good point, and since I have to do one more cycle because I forgot to update the CHANGELOG, let me see if I can slide that in too and get it all done in one go.

Pilchie (approved):

PR #24835, opened 2025-06-23
cosmos: enable support for reading all partition key ranges from a given container

arespinosa:
Add a method to the go sdk as a container operation to retrieve all pkranges from a cosmos container. This will assist us with a future contribution to scale out change feed processing as we will want to enumerate over and enqueue jobs per physical partition. 


arespinosa:
@microsoft-github-policy-service agree [company="GitHub"]

arespinosa:

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	"net/http"
+
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+// PartitionKeyRange represents a single partition key range from GET /pkranges endpoint
+// Includes all the needed fields to represent a partition key range
+type PartitionKeyRange struct {
+	ID                      string   `json:"id"`
+	Rid                     string   `json:"_rid"`
```
analogrelay: We usually use the term `ResourceID` for this in public properties, rather than the abbreviation RID

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	ETag                    string   `json:"_etag"`
```
analogrelay: There is an `azcore.Etag` type we use for etags. Take a look at `ContainerProperties` for a sample of using that.

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	MinInclusive            string   `json:"minInclusive"`
+	MaxExclusive            string   `json:"maxExclusive"`
+	RidPrefix               int      `json:"ridPrefix"`
```
analogrelay: If we're using `ResourceID` elsewhere (for consistency with other types), I'd suggest making this `ResourceIDPrefix`

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	Self                    string   `json:"_self"`
```
analogrelay: We call this `SelfLink` in other types.

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	ThroughputFraction      float64  `json:"throughputFraction"`
+	Status                  string   `json:"status"`
+	Parents                 []string `json:"parents"`
+	OwnedArchivalPKRangeIds []string `json:"ownedArchivalPKRangeIds"`
+	Timestamp               int64    `json:"_ts"`
```
analogrelay: We name this `LastModified` in other types, and use a `time.Time` to store it. Between this and the Etag, you may need to write a custom `MarshalJSON` and `UnmarshalJSON` method rather than using the `json:` struct tags. You can see samples of this in `ContainerProperties` and other similar types.

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+	LSN                     int64    `json:"_lsn"`
```
analogrelay: Copilot pointed this out too, it's actually just called `lsn` in the JSON, not `_lsn`.

sdk/data/azcosmos/cosmos_partition_key_range_response.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// PartitionKeyRangeResponse represents the response from GET /pkranges endpoint
+// Contains the list of partition key ranges
+// Rid is for the high level resource id
+// count is for the number of partition key ranges returned
+type PartitionKeyRangeResponse struct {
+	Response
+	Rid                string              `json:"_rid"`
```
analogrelay: Again, for consistency, I'd use `ResourceID` here. The acronym `Rid` doesn't tend to appear in our public API very often.

sdk/data/azcosmos/cosmos_partition_key_range_response_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if parsedPkr1.MaxExclusive != "5A" {
+		t.Errorf("Expected MaxExclusive to be %s, but got %s", "5A", parsedPkr1.MaxExclusive)
+	}
+
+	if parsedPkr1.ThroughputFraction != 0.25 {
+		t.Errorf("Expected ThroughputFraction to be %f, but got %f", 0.25, parsedPkr1.ThroughputFraction)
+	}
+
+	if len(parsedPkr1.Parents) != 2 || parsedPkr1.Parents[0] != "parent1" {
```
analogrelay: Should check `Parents[1]` as well, and check `ownedArchivalPKRangeIds`. I'm fine if we only do that with one of the PKRanges, just want some coverage on all the properties if possible.

analogrelay (changes requested):
This is a great start, and something I've been thinking of adding recently so I'm glad to see it done! 

I had a couple comments about the formatting of the `PartitionKeyRange` struct, which should be fairly straightforward.

In addition, I'd like to see a test added to the emulator tests in `emulator_cosmos_container_test.go`. The test should just create a container, get the PK ranges, and validate a few key properties of it (The ID and min/max would be enough).

You can run those tests by launching the Cosmos Emulator (instructions for Windows/macOS/Linux here: https://learn.microsoft.com/azure/cosmos-db/how-to-develop-emulator) and then running these tests with the `EMULATOR=true` environment variable set. The Azure Pipelines build on this PR also runs them, so once you've added the tests, we should see them run there :).

adrianwclement:
Now revising the function to read all PartitionKeyRange objects to be internal, as advised by @FabianMeiswinkel

sdk/data/azcosmos/cosmos_partition_key_range.go
```diff
@@ -0,0 +1,201 @@
...
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// PartitionKeyRange represents the properties of a partition key range.
+type PartitionKeyRange struct {
```
analogrelay: If the APIs are internal, should the types (this one and `PartitionKeyRangeResponse`) be internal as well? Unless there's a specific reason they need to be public, I'd mark them both internal for now.

arespinosa: Hi @analogrelay! 😄 

We changed the PartitionKeyRange, PartitionKeyRangeResponse, and partitionKeyRangeOptions structs to be internal. We decided to keep the fields inside the struct uppercase so we can continue to use Go's automatic support for encoding/json and marshaling/unmarshaling.

analogrelay: That makes total sense. Exporting the fields when the struct is internal won't cause any public API issues, since the type can't be named from outside the package anyway.

analogrelay (approved):

PR #24843, opened 2025-06-24
cosmos: Cherry-Pick Vector/FT container policies to 1.5.0 branch

analogrelay:
This PR is a simple cherry-pick of the changes in #24833 from `main` over to the release branch for the preview release that includes query engine support. That seemed like the smoothest way to get something into this branch quickly to unblock vector scenarios there. I expect we'll release a `1.5.0` and move the query engine preview over to `1.6.0-beta` soon, and I'll reconcile things like CHANGELOG when that happens. Until then, I'd like to be able to make a preview release to allow folks to actually try Vector search with the query engine.

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Features Added
 
+* Added support for BypassIntegratedCache option See [PR 24772](https://github.com/Azure/azure-sdk-for-go/pull/24772)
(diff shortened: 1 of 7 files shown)


Pilchie (approved):
Yep, looks the same.  Curious about the choice of cherry-pick vs merge and preserve commit parent chains, but either way is fine with me.

analogrelay:
It's mostly because main has a lot of commits from unrelated SDKs and resulted in a very large changeset with lots of mandatory reviewers. An unfortunate consequence of having all the SDKs commingled like this.

PR #24860, opened 2025-06-26
Allow an additional retry for IMDS managed identity requests

### LINKED ISSUE #24842: ManagedIdentityCredential retry policy should handle 410 status response
DefaultAzureCredentialImdsRetryPolicy should delay for long enough so that the exponential retry total duration is at least 70 seconds in the event that the response code is 410.

[See the docs](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#errors-and-debugging) for more info on 410 status code

chlowell:
Closes #24842 with a small tweak to IMDS retry configuration. Our retry policy doesn't allow setting retry options based on a response, so this PR allows another retry in any case and shortens the `MaxRetryDelay` so the total retry period is about 70 seconds.

Here's a sample of the change's effect. Delays for retries 1-3 are unchanged (jitter causes the variation) but now there's a 5th retry, which like the 4th is limited to 25s.
| Retry | Delay before | Delay after |
|-------|----------------|---------------|
| 1     | 2.11s          | 1.71s         |
| 2     | 5.69s          | 7.46s         |
| 3     | 12.37s         | 13.38s        |
| 4     | 31.80s         | 25.00s        |
| 5     | —              | 25.00s        |
| **Total** | **51.97s**     | **72.55s**      |

### DIFF
--- a/sdk/azidentity/managed_identity_client.go
+++ b/sdk/azidentity/managed_identity_client.go
@@ -54,10 +54,10 @@ type managedIdentityClient struct {
 // setIMDSRetryOptionDefaults sets zero-valued fields to default values appropriate for IMDS
 func setIMDSRetryOptionDefaults(o *policy.RetryOptions) {
 	if o.MaxRetries == 0 {
-		o.MaxRetries = 5
+		o.MaxRetries = 6
 	}
 	if o.MaxRetryDelay == 0 {
-		o.MaxRetryDelay = 1 * time.Minute
+		o.MaxRetryDelay = 25 * time.Second
 	}
 	if o.RetryDelay == 0 {
 		o.RetryDelay = 2 * time.Second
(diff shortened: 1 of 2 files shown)


sdk/azidentity/managed_identity_client.go
```diff
@@ -82,9 +82,6 @@ func setIMDSRetryOptionDefaults(o *policy.RetryOptions) {
 			http.StatusNetworkAuthenticationRequired, // 511
 		}
 	}
-	if o.TryTimeout == 0 {
```
jhendrixMSFT: Was this ever documented, and do you know why it was set?

chlowell: Decided to revert this change after offline discussion concluded that although it has no effect in most cases, an application using a custom transport could see a longer wait before reaching an error if we remove this default

jhendrixMSFT (approved):

PR #24864, opened 2025-06-28
[azservicebus] Fixing bug where you could lose messages in ReceiveAndDelete

### LINKED ISSUE #24078: [azservicebus] Better ReceiveAndDelete behavior when receiving then closing the link
When doing this pattern, with ReceiveAndDelete links, you can lose messages even if you do everything right:

```go
messages, err := receiver.ReceiveMessages(ctx, n > 1)
// this limbo area is a problem
receiver.Close()
```

The problem here is if there are any excess credits after ReceiveMessages finishes (for instance, if you ask for 100, but only get 3) then those messages will be cached, and _lost_ when Close() is called.

I think we've got two solutions ahead of us, and I've already rejected one of them, making the rest of this process pretty easy:
- REJECTED: Allow the user to call into ReceiveMessges and _only_ receive cached messages. This has some precedent, as you can do this in some of the other clients, but you, as a consumer, wouldn't know how many messages are cached and you still have a race condition between messages arriving and Close(). 
- CHOSEN: Bring back the 'drain' approach we used, prior to our release callback method. This has the benefit of making sure that every ReceiveMessages() finishes clean - we'll never have any messages in our internal caches and can always make sure the user has everything.

So, we have a few work items for this:

- [ ] Bring back the DrainCredits() function, which has been removed in go-amqp
- [ ] Add in drain support into Service Bus
- [ ] Update our stress tests to validate that we're not dropping messages in ReceiveAndDelete mode. Probably just take the sample code above, and turn it up to 11.

richardpark-msft:
This PR makes it so you can call ReceiveMessages after a Receiver (only in ReceiveAndDelete mode) is closed, giving you access to any cached messages that might have arrived after your final ReceiveMessages() call, but before Close().

Fixes #24078

### DIFF
--- a/sdk/messaging/azservicebus/receiver_test.go
+++ b/sdk/messaging/azservicebus/receiver_test.go
@@ -1237,7 +1237,7 @@ func TestReceiverReceiveAndDeleteDoesNotLoseMessages(t *testing.T) {
 	sender, err := client.NewSender(queueName, nil)
 	require.NoError(t, err)
 
-	const totalSent = 2000
+	const totalSent = 3000
 
 	mustSendMessages(t, context.Background(), totalSent, 1000, sender)
 
@@ -1262,9 +1262,11 @@ func TestReceiverReceiveAndDeleteDoesNotLoseMessages(t *testing.T) {
 	// At this point the receiver is closed - we can continue to receive from it, even after closure. The messages
 	// are only what is cached. This lets users who are doing ReceiveAndDelete to make sure they get all messages that have
 	// arrived - previously, it was possible to have some data loss.
-	t.Logf("Now receiving after the receiver has been closed. We have %d/%d messages", len(receivedBeforeClose), totalSent)
+	t.Logf("Done with our first normal ReceiveMessages() call, when the Receiver is open. (%d/%d)", len(receivedBeforeClose), totalSent)
 
 	var after []*ReceivedMessage
+
+	totalReceived := func() int { return len(after) + len(receivedBeforeClose) }
 
 Loop:
 	for {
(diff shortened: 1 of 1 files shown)


sdk/messaging/azservicebus/receiver.go
```diff
@@ -453,6 +485,25 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
+func (r *Receiver) receiveFromCache(maxMessages int) []*ReceivedMessage {
+	if len(r.cachedMessages) == 0 {
+		return nil
+	}
+
+	n := min(len(r.cachedMessages), maxMessages)
+
+	receivedMessages := make([]*ReceivedMessage, n)
+
+	for i := range n {
```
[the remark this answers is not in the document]
richardpark-msft: This works, as of Go 1.22:

1.22https://go.dev/ref/spec#For_range:~:text=values%20received%20on%20a%20channel%2C%20integer%20values%20from%20zero%20to%20an%20upper%20limit%20%5BGo%201.22%5D%2C

sdk/messaging/azservicebus/receiver.go
```diff
@@ -453,6 +485,25 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
+		receivedMessages[i] = newReceivedMessage(r.cachedMessages[i], nil)
+	}
+
+	r.cachedMessages = r.cachedMessages[n:]
```
jhendrixMSFT: Are methods on `Receiver` allowed to be concurrently called? If so, I think this can race.

richardpark-msft: Gah, you're right. I actually thought I had that sealed up, but it's possible for receiver.Close() and receiver.ReceiveMessages() to intersect. Let me think...

richardpark-msft: I updated it to use an atomic.Pointer around the slice value. It's only ever _updated_ once, and never mutated afterwards so this should work just fine.

richardpark-msft:
(Just a note, the test failure was just a problem with my test, not the package code)

sdk/messaging/azservicebus/receiver.go
```diff
@@ -372,6 +400,10 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
 		return nil, internal.NewErrNonRetriable(fmt.Sprintf("maxMessages cannot exceed %d", r.maxAllowedCredits))
 	}
 
+	if msgs := r.receiveFromCache(maxMessages); len(msgs) > 0 {
```
jhendrixMSFT, line 403: Can this be predicated to run only when the receiver is closed? I would have thought that we only need to do this post closing.

richardpark-msft: It works out the same because the array of stored messages is only populated when the receiver is closed.

sdk/messaging/azservicebus/receiver.go
```diff
@@ -453,6 +485,26 @@ func (r *Receiver) receiveMessagesImpl(ctx context.Context, maxMessages int, opt
...
+
+	n := min(len(*cachedMessages), maxMessages)
+
+	receivedMessages := make([]*ReceivedMessage, n)
+
+	for i := range n {
+		receivedMessages[i] = newReceivedMessage((*cachedMessages)[i], nil)
+	}
+
+	(*cachedMessages) = (*cachedMessages)[n:]
```
jhendrixMSFT, line 504: Only one goroutine can ever execute this statement?

richardpark-msft: Yes - there's a mutex in ReceiveMessages(), which is driving all of this.

jhendrixMSFT (approved):

PR #24883, opened 2025-07-01
Copilot instruction to handle Go SDK generation

tadelesh:
Also, start to versioning generator tool.

### DIFF
--- a/.github/instructions/go-sdk-automation-analysis.instructions.md
+++ b/.github/instructions/go-sdk-automation-analysis.instructions.md
@@ -1,0 +1,47 @@
+# Analyze the Go SDK Automation Result
+
+Follow these steps to help users analyze Go SDK automation results from GitHub PRs. This guide focuses on interpreting results and providing resolution guidance.
+
+## Process Overview
+
+1. **Generate Go SDK locally** - Reproduce the automation result
+2. **Provide guidance** - Offer specific resolution steps and resources
+
+## Step 1: Generate Go SDK
+
+1. Follow [Generate Go SDK from API specification](./go-sdk-generation.instructions.md) to reproduce the automation result locally
+2. Collect all generation logs, output files, and error messages for analysis
+
+## Step 2: Analyze Results
+
+1. **Success Detection**: If the log contains `Finish processing typespec project` keyword and no `[ERROR]` keywords are found, then generation succeeded. Check the output and summarize the results.
+
+2. **Error Handling**: If any `[ERROR]` logs or stack traces are found, follow the [Azure Go SDK Automation Troubleshooting Guide](../../documentation/sdk-automation-tsg.md) to suggest how to resolve the errors.
+
+### Step 3: Resolution Guidance
+
+1. Provide specific actionable steps based on error category
+2. Include relevant documentation links
+3. Format issue templates for external reporting when needed
+
+## Output Format
+
+### Success Analysis
+
+```
+✅ **Status**: Automation Successful
+📦 **Package**: [package-name]
+📋 **Changelog**: [summary of changes]
+⚠️ **Breaking Changes**: [if any, with details]
+```
+
+### Failure Analysis
+
+```
+❌ **Status**: Automation Failed
+📦 **Package**: [package-name]
+🔍 **Error Category**: [category-name]
+💬 **Error Summary**: [concise description]
+🛠️ **Resolution**: [specific action items]
+📎 **Additional Info**: [links, references, issue templates]
+```
--- a/.github/instructions/go-sdk-generation.instructions.md
+++ b/.github/instructions/go-sdk-generation.instructions.md
@@ -1,0 +1,101 @@
+# Generate Go SDK from API specification
+
+Follow these steps to help users generate Go SDK from specific API specifications. Show the high-level description to users to help them understand the flow. Use the provided tools to perform actions and gather information as needed.
+
+## Go SDK Generation Process
+
+1. **Prerequisites**: Ensure user has the correct environment to generate the Go SDK.
+2. **Prepare generation config**: Get the information of API specification from user's input and prepare the generation config file.
+3. **Run generation command**: Execute command to perform the Go SDK generation.
+4. **Gather results and handle errors**: Retrieve the generation log and handle any problems.
+
+## Prerequisites
+
+1. Ensure Go version is 1.23 or later.
+2. Install the `generator` tool (v0.0.0-20250630212816-bb29ffdbc8c4) with the command:
+
+```bash
+go install github.com/Azure/azure-sdk-for-go/eng/tools/generator@v0.0.0-20250630212816-bb29ffdbc8c4
+```
+
+3. Ensure Node.js version is 20 or later.
+4. Install `tsp-client` tool (v0.21.0) with the command:
+
+```bash
+npm install -g @azure-tools/typespec-client-generator-cli@v0.21.0
+```
+
+5. Ensure GitHub CLI tool is installed and successfully logged in.
+6. Ensure Git is installed.
+
+## Step 1: Prepare Generation Config
+
+**Goal**: Generate the `generatedInput.json` file.
+
+**Actions**:
+
+1. Get API specification root path (`specPath`):
+
+- **From open editor files**: If `tspconfig.yaml` or `main.tsp` files are currently open in the editor, use their parent directory as the `specPath`.
+- **From local file path**: If the user provides a local `tspconfig.yaml` file path, use its parent directory as the `specPath`.
+- **From GitHub PR link**: If user provides a GitHub PR URL:
+  - Prompt user: "Do you have a local copy of the spec repository?"
+  - **If yes**:
+    - Use the local repository path
+  - **If no**:
+    - Clone the repository to system temp directory
+  - Use GitHub CLI to checkout the PR branch: `gh pr checkout <PR_NUMBER>`
+  - Analyze the PR's changed files to locate `tspconfig.yaml`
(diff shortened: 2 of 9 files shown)


richardpark-msft:
I've got some too!

https://github.com/Azure/azure-sdk-for-go/pull/24872

documentation/sdk-automation-tsg.md
```diff
@@ -0,0 +1,109 @@
+# Azure Go SDK Automation Trouble Shooting Guide
```
richardpark-msft:
```suggestion
# Azure Go SDK Automation Troubleshooting Guide
```

documentation/sdk-automation-tsg.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Overview
+
+The Azure Go SDK automation tool:
+
+- Generates Go SDK code from TypeSpec/OpenAPI specifications
+- Creates changelogs for API updates
```
richardpark-msft, line 8: This literally came up in another meeting, love to see this. Would it just use the commit messages from git to do it, or are you thinking of some other method?

richardpark-msft:
CC: @praveenkuttappan, who is also looking into this area, but for all languages.

.github/instructions/go-sdk-automation-analysis.instructions.md
```diff
@@ -0,0 +1,47 @@
+# Analyze the Go SDK Automation Result
+
+Follow these steps to help users analyze Go SDK automation results from GitHub PRs. This guide focuses on interpreting results and providing resolution guidance.
+
+## Process Overview
```
richardpark-msft: @timovv and I have had a hard time getting copilot to _truly_ follow a list of instructions. Is it better to just make this a single Powershell script, so it always does it deterministically? That'd also make it simpler for people, once they know, to just re-run it again without having to involve chat each time.

These instructions would then just be modified to say "run the script", which'll probably work every time and involve less pausing/eval from chat.

Another benefit is we can start standardizing the scripts a user is expected to run to generate SDKs, so if they're working across multiple repos, it's the same workflow in any case. (the script obviously would be smart enough to know which repo/language it's in).

documentation/sdk-breaking-changes-categories.md
```diff
@@ -0,0 +1,101 @@
+# Azure Go SDK Breaking Changes Categories
+
+## 1. Naming Changes
+
+### Model/Type Naming
+
+- **TypeSpec**: Model, enum, union name changes
```
richardpark-msft: I'm guessing these rules are going to be duplicated all over the place. Considering these are really TypeSpec-controlled, should we move it to azure-rest-api-specs?

Even more awesome would be if we could integrate with apiview to do this breaking change detection, since (I believe) we already have automation for this.

eng/tools/generator/README.md
```diff
@@ -1,19 +1,55 @@
...
+The `issue` command fetches release request issues from `github.com/Azure/sdk-release-request/issues` and parses them into configuration that other commands consume. The configuration outputs to stdout.
+
+**Usage:**
+```bash
+generator issue [flags]
+```
 
-### The `issue` command and the configuration file
+**Flags:**
+- `-t, --token`: Personal access token for GitHub authentication
```
richardpark-msft: Shelling out to using the `gh` CLI would be nice here - then I don't have to authenticate _twice_, you can just rely on the GitHub CLI to do the work.

You can get back JSON from the github cli as well, which I find useful sometimes:

```bash
gh issue list --json author,comments,number
```

It'll also simplify your tools docs and flags.

eng/tools/generator/README.md
```diff
@@ -1,19 +1,55 @@
 # Generator
 
-This is a command line tool for generating new releases for `github.com/Azure/azure-sdk-for-go`.
+This is a command line tool for generating new releases and managing automation for `github.com/Azure/azure-sdk-for-go`.
```
richardpark-msft: There's some duplication here with work that @praveenkuttappan is doing. His automation to drive this, via Copilot chat, is intended to work cross-repo, by my understanding. He can explain further, but I think there could be some nice sharing, and saving of work.

eng/tools/generator/main.go
```diff
@@ -4,13 +4,51 @@
...
+	// Configure version template to show just the version number
+	rootCmd.SetVersionTemplate("{{.Version}}\n")
+
+	rootCmd.AddCommand(
+		issue.Command(),
+		template.Command(),
+		automation_v2.Command(),
+		release_v2.Command(),
+		refresh_v2.Command(),
+		readme.Command(),
```
richardpark-msft: I'd be interested in the code for this command - I was thinking of doing something similar, and you might have beat me to the punch.

What I was hoping to do was take something like "what's the https://learn.microsoft.com URL for your service?" and then use that as the subject matter to generate a summary, key concepts, etc...

richardpark-msft:
@tadelesh, really looking forward to this - I love automation :)

.github/instructions/go-sdk-automation-analysis.instructions.md
```diff
@@ -0,0 +1,47 @@
+# Analyze the Go SDK Automation Result
```
richardpark-msft: These instructions files, if you want, can have wildcards so they only apply when you're in certain folders, or working with certain files.

I'm not sure "data plane vs ARM" would change things here, but if so we could modularize a bit and say things like:

```
---
applyTo: '/sdk/resourcemanager/**'
---
```

at the top, and have the instructions only apply if you're working with an ARM package, for instance.

I worry mostly that our overall instructions sent for each folder/file is going to become large, and start making our Copilot results more chaotic.

What do you think?

eng/tools/generator/main.go
richardpark-msft: Ah, I see now where you're getting the commands from. I'll take a look at 	"github.com/Azure/azure-sdk-for-go/eng/tools/generator/cmd/v2/readme", and see if my question still makes sense.

documentation/sdk-automation-tsg.md
tadelesh: For current Go SDK automation tool, we just use the diff of exported API between the new code and last version to generate the changelog, which is not so useful. We could think about enhance it with commit messsage.

.github/instructions/go-sdk-generation.instructions.md
```diff
@@ -0,0 +1,101 @@
+# Generate Go SDK from API specification
+
+Follow these steps to help users generate Go SDK from specific API specifications. Show the high-level description to users to help them understand the flow. Use the provided tools to perform actions and gather information as needed.
+
+## Go SDK Generation Process
+
+1. **Prerequisites**: Ensure user has the correct environment to generate the Go SDK.
```
richardpark-msft: I'd mentioned this in another spot, but I did a PR to show an approach I like to take for "lists of instructions":

https://github.com/Azure/azure-sdk-for-go/compare/instruction...richardpark-msft:azure-sdk-for-go:instructions-suggestions

Running each bit through Copilot means you have to wait for the command, hit accept, etc... a few times in a row. The script is faster, and we can just tell Copilot to execute that instead.

CC: @timovv, who I believe is doing similar things.

.github/instructions/go-sdk-automation-analysis.instructions.md
richardpark-msft: (I've added an example here: https://github.com/Azure/azure-sdk-for-go/pull/24883/files/4fa4da950149fbf9251a9185d0b3674b65dea7dc#r2193682642, where we're doing a check for prerequisites)

eng/tools/generator/README.md
tadelesh: This generator is an existed tool which we have used to handle ARM SDK release for a long time. I just update the legacy doc here. I heard that there is a big picture about the SDK agent, but in current PR, I want to focus on reducing our current daily work we are facing: handle local SDK generation to troubleshooting, review spec change's breaking changes, mitigate the breaking changes from the TypeSpec migration, etc.

.github/instructions/go-sdk-automation-analysis.instructions.md
tadelesh: I think it work for both ARM and data plane. This instruction aims to let copilot reproduce the SDK automation failure locally with the SDK repo, and help user to resolve the problem. Since the SDK automation tool is in our Go repo, I put it here.

tadelesh:
@richardpark-msft With current instructions, I could successfully generate the SDK and help with some of the ARM migration breaking change resolver works. Do you think we could merge this PR first? My next plan is to refactor the generator tool and make it exposed as an MCP server, but still keep the ability to execute with cmd line. Also, I will try to split the big instruction into some parts into the prompt in the MCP server, and try to leverage LLM more to deal with the process.

documentation/sdk-breaking-changes-guide.md
```diff
@@ -79,15 +122,38 @@ op restoreVolume is ArmResourceActionAsync<Volume, void, Volume>;
 ```
 
 **Resolution**:
+
 ```tsp
 @@clientLocation(Microsoft.ElasticSan.restoreVolume, "Management", "go");
 ```
 
+### User Standardard Operations List Operation
```
richardpark-msft:
```suggestion
### User Standard Operations List Operation
```

RickWinter:
@richardpark-msft @tadelesh, Are we moving this to MCP scripts? Should this be closed?

tadelesh:
It seems we are still discussing cross-languages mcp tool design. I used these instructions to help with some daily works. @richardpark-msft do you think we could merge it? After we have the mcp design, we could refine the impl.

tadelesh:
Close as we are discussing cross language design.

PR #24898, opened 2025-07-07
cosmos: enable support for reading the change feed for through feed ranges

adrianwclement:
This PR adds a function to the Go SDK as a Cosmos container operation to retrieve the Change Feed for a specific FeedRange.

The new functionality supports retrieving the Change Feed using various HTTP request headers such as `If-None-Match` and `If-Modified-Since`, providing fine-grained control over which changes are returned. This enables clients to efficiently process changes, implement custom polling intervals, and resume from specific points in the feed.

## Notes

- This PR builds on:
  - https://github.com/Azure/azure-sdk-for-go/pull/24835
  - https://github.com/Azure/azure-sdk-for-go/pull/24889

- **Please review after** the above PRs are merged, as this change set is rebased on top of them.
- Changes in this PR start from commit: [04ef4f8](https://github.com/arespinosa/azure-sdk-for-go/commit/04ef4f842d953ca45abac71783c5277a39b4682f) (first unique commit).
- Extension for ChangeFeed read support to PartitionKeys and an entire container is planned for the future, but this PR intentionally narrows scope to FeedRange only.

### DIFF
--- a/sdk/data/azcosmos/emulator_cosmos_container_test.go
+++ b/sdk/data/azcosmos/emulator_cosmos_container_test.go
@@ -795,54 +795,6 @@ func TestEmulatorContainerChangeFeed(t *testing.T) {
 		}
 	})
 
-	// Test change feed with simple with If-None-Match header
-	// Test change feed with FeedRange (gets composite continuation token, not simple ETag)
-	t.Run("IfNoneMatchHeader", func(t *testing.T) {
-		// Use GetChangeFeed with feed range (this will get composite continuation token)
-		options := &ChangeFeedOptions{
-			MaxItemCount: 10,
-		}
-		options.FeedRange = &feedRanges[0]
-
-		resp, err := container.GetChangeFeed(context.TODO(), options)
-		if err != nil {
-			t.Fatalf("Failed to get change feed: %v", err)
-		}
-
-		// Verify composite continuation token IS populated when using feed range
-		if resp.ContinuationToken == "" {
-			t.Error("Expected ContinuationToken to be populated for feed-range request")
-		}
-
-		// Verify ETag is present
-		if resp.ETag == "" {
-			t.Error("Expected ETag to be present")
-		}
-
-		// Test using the composite continuation token in next request
-		if resp.ContinuationToken != "" {
-			options2 := &ChangeFeedOptions{
-				MaxItemCount: 10,
-				Continuation: &resp.ContinuationToken,
-			}
-
-			resp2, err := container.GetChangeFeed(context.TODO(), options2)
-			if err != nil {
-				t.Fatalf("Failed to get change feed with composite token: %v", err)
-			}
-
-			t.Logf("Second request with composite token - Count: %d, StatusCode: %d", resp2.Count, resp2.RawResponse.StatusCode)
-			// With continuation, we might get 304 Not Modified if no changes
-			if resp2.RawResponse.StatusCode == 304 {
-				t.Log("Got 304 Not Modified - no new changes since last continuation token")
-				if resp2.Count != 0 {
-					t.Errorf("Expected Count to be 0 for 304 response, got %d", resp2.Count)
-				}
-			}
-		}
-	})
-
-	// Test change feed with If-Modified-Since header
 	// Test change feed with If-Modified-Since header
 	t.Run("IfModifiedSinceHeader", func(t *testing.T) {
 		// First, get all current changes to establish a baseline


sdk/data/azcosmos/cosmos_change_feed_request_options.go
```diff
@@ -0,0 +1,93 @@
...
+// ChangeFeedOptions defines the options for retrieving the change feed.
+// Incorporate Continuation
+type ChangeFeedOptions struct {
+	// MaxItemCount limits the number of items returned per page.
+	// Valid values are > 0. The service may return fewer items than requested.
+	MaxItemCount int32
+
+	// ChangeFeedStartFrom is a user-friendly way to specify the time for change feed
+	// Will be set to the IfModifiedSince header
+	ChangeFeedStartFrom *time.Time
```
analogrelay: Given that the type is `ChangeFeedOptions` I think this can just be called `StartFrom`

sdk/data/azcosmos/cosmos_change_feed_request_options.go
```diff
@@ -0,0 +1,93 @@
...
+func (options *ChangeFeedOptions) toHeaders(partitionKeyRanges []partitionKeyRange) *map[string]string {
+	headers := make(map[string]string)
+
+	headers[cosmosHeaderChangeFeed] = cosmosHeaderValuesChangeFeed
+
+	if options.MaxItemCount > 0 {
+		headers[cosmosHeaderMaxItemCount] = strconv.FormatInt(int64(options.MaxItemCount), cosmosBaseTen)
+	} else {
+		headers[cosmosHeaderMaxItemCount] = strconv.FormatInt(cosmosDefaultMaxItemCount, cosmosBaseTen)
+	}
```
analogrelay: If the value is `0`, we should just omit the header entirely and let the server control the default behavior. That's what we do for `PageSizeHint` in queries.

sdk/data/azcosmos/cosmos_change_feed_request_options.go
```diff
@@ -0,0 +1,93 @@
...
+			if options.FeedRange == nil {
+				options.FeedRange = &FeedRange{
+					MinInclusive: compositeToken.Continuation[0].MinInclusive,
+					MaxExclusive: compositeToken.Continuation[0].MaxExclusive,
+				}
+			}
+		} else {
+			headers[headerIfNoneMatch] = *options.Continuation
+		}
+	}
```
analogrelay, line 63: I think we'd need to do this before using the `FeedRange` to set the PKRangeID header, so that the `FeedRange` actually ends up being considered.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -707,6 +708,173 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
...
+				}
+			}
+		}
+	}
+
+	if options.FeedRange != nil {
+		return c.getChangeFeedForEPKRange(ctx, options.FeedRange, options)
+	} else {
+		// TODO: Implement logic to handle PartitionKey and Container reads in separate Helper functions
+		// Below is a placeholder for this logic—it should not be used in production
```
analogrelay: Is this TODO something you plan to address before merging?

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -707,6 +708,173 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
...
+// ) (ChangeFeedResponse, error) {
+// }
+
+// TODO: Use getChangeFeedForPartitionKey set in options to read the change feed for a specific partition.
+// func (c *ContainerClient) getChangeFeedForPartitionKey(
+// 	ctx context.Context,
+// 	partitionKey *PartitionKey,
+// 	options *ChangeFeedOptions,
+// ) (ChangeFeedResponse, error) {
+// }
```
analogrelay: We shouldn't leave commented-out code in `main`. If you're planning to add these functions later, that's fine, but we don't want to leave their signatures here commented-out.

sdk/data/azcosmos/cosmos_container.go
```diff
@@ -707,6 +708,173 @@ func (c *ContainerClient) ExecuteTransactionalBatch(ctx context.Context, b Trans
...
+			return response, err
+		}
+
+		if options.FeedRange != nil {
+			response.FeedRange = options.FeedRange
+			response.PopulateCompositeContinuationToken()
+		}
+
+		return response, nil
+		// return ChangeFeedResponse{}, fmt.Errorf("no FeedRange provided in options")
```
analogrelay: Left over from an earlier iteration?

sdk/data/azcosmos/cosmos_container_test.go
```diff
@@ -822,3 +822,255 @@ func TestContainerReadPartitionKeyRangesEmpty(t *testing.T) {
 		t.Fatalf("Expected 0 partition key ranges, got %d", len(resp.PartitionKeyRanges))
 	}
 }
+
+func TestContainerGetChangeFeedIfModifiedSince(t *testing.T) {
```
analogrelay: I'm not seeing where this test validates that the `if-modified-since` header is actually set on the outgoing request.

sdk/data/azcosmos/cosmos_container_test.go
```diff
@@ -822,3 +822,255 @@ func TestContainerReadPartitionKeyRangesEmpty(t *testing.T) {
...
+	}
+	if resp.Count != 2 {
+		t.Errorf("Expected Count 2, got %v", resp.Count)
+	}
+	if len(resp.Documents) != 2 {
+		t.Errorf("Expected 2 documents, got %v", len(resp.Documents))
+	}
+}
+
+func TestContainerGetChangeFeedPartitionKey(t *testing.T) {
```
analogrelay: Similarly here, I'm not seeing any assertion that the partition key is being used.

sdk/data/azcosmos/cosmos_container_test.go
```diff
@@ -822,3 +822,255 @@ func TestContainerReadPartitionKeyRangesEmpty(t *testing.T) {
...
+	}
+	if resp.Count != 1 {
+		t.Errorf("Expected Count 1, got %v", resp.Count)
+	}
+	if len(resp.Documents) != 1 {
+		t.Errorf("Expected 1 document, got %v", len(resp.Documents))
+	}
+}
+
+func TestContainerGetChangeFeed_FilteredByHeader(t *testing.T) {
```
analogrelay: This isn't really testing the filtering. You've configured the test server to return the unfiltered responses first, then the filtered ones. I'd expect this test to instead be validating that the `StartFrom` value results in the appropriate request headers being set.

sdk/data/azcosmos/cosmos_container_test.go
```diff
@@ -822,3 +822,255 @@ func TestContainerReadPartitionKeyRangesEmpty(t *testing.T) {
...
+	resp, err = container.GetChangeFeed(context.TODO(), options)
+	if err != nil {
+		t.Fatalf("GetChangeFeed (filtered) returned error: %v", err)
+	}
+	if resp.Count != 1 || len(resp.Documents) != 1 {
+		t.Errorf("Expected 1 document, got %d", len(resp.Documents))
+	}
+}
+
+func TestContainerGetChangeFeedForEPKRange(t *testing.T) {
```
analogrelay, line 1045: I don't see where the EPK Range is being verified here either.

sdk/data/azcosmos/cosmos_feed_range.go
```diff
@@ -19,3 +21,24 @@ func NewFeedRange(minInclusive, maxExclusive string) FeedRange {
...
+
+// TODO: Modify this function to use the partitionKeyRangeCache
+// findPartitionKeyRangeIDWithCache finds the partition key range ID for a FeedRange using a partitionKeyRangeCache.
+// func findPartitionKeyRangeIDWithCache(
+//     feedRange FeedRange,
+//     cache *partitionKeyRangeCache,
+// ) (string, error) {
+//     partitionKeyRanges := cache.getAll()
+//     return FindPartitionKeyRangeId(feedRange, partitionKeyRanges)
+// }
```
analogrelay: This TODO content can be removed for now too

sdk/data/azcosmos/cosmos_feed_range.go
```diff
@@ -19,3 +21,24 @@ func NewFeedRange(minInclusive, maxExclusive string) FeedRange {
...
+// findPartitionKeyRangeID finds the partition key range ID that matches the given FeedRange.
+// Returns the ID if found, or an error if no match exists.
+func findPartitionKeyRangeID(feedRange FeedRange, partitionKeyRanges []partitionKeyRange) (string, error) {
+	for _, pkr := range partitionKeyRanges {
+		if feedRange.MinInclusive == pkr.MinInclusive && feedRange.MaxExclusive == pkr.MaxExclusive {
+			return pkr.ID, nil
+		}
+	}
+	return "", fmt.Errorf("no matching partition key range found for feed range [%s, %s)", feedRange.MinInclusive, feedRange.MaxExclusive)
+}
```
analogrelay, line 34: Does this need to account for the fact that a `FeedRange` may not map to a single exact PKRange? For example, if there's a split or merge while a Change Feed Processor is running?

sdk/data/azcosmos/cosmos_change_feed_composite_continuation_token_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	compositeContinuationToken := newCompositeContinuationToken(resourceID, []changeFeedRange{changeFeedRangeInstance})
+	t.Logf("ResourceID: %s\nContinuation: %+v", compositeContinuationToken.ResourceID, compositeContinuationToken.Continuation)
+
+	data, err := json.Marshal(compositeContinuationToken)
+	if err != nil {
+		t.Fatalf("Failed to marshal composite token: %v", err)
+	}
+
+	expectedJSON := `{"resourceId":"testResource","continuation":[{"minInclusive":"","maxExclusive":"FF","continuationToken":"14"}]}`
```
analogrelay: From what I can tell, both the [Java ](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos-tests/src/test/java/com/azure/cosmos/implementation/feedranges/FeedRangeContinuationTest.java) and [Python](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/cosmos/azure-cosmos/azure/cosmos/_change_feed/feed_range_composite_continuation_token.py#L35) SDKs use the same JSON format (though the casing of property names is a little different). We should probably use the same format here, even though I don't think there's much reason to share continuation tokens between SDKs.

sdk/data/azcosmos/cosmos_change_feed_response.go
```diff
@@ -0,0 +1,117 @@
...
+	// Selected HTTP headers that we're retrieving from the response
+	ETag              string
+	ContinuationToken string
+	LSN               string
+
+	// Store the feed range if it was used in the request
+	FeedRange *FeedRange
+
+	// CompositeContinuationToken is automatically populated when using feed ranges
+	CompositeContinuationToken string
```
analogrelay: Does the user need both continuation tokens? I'd expect that if they're polling the change feed for a given FeedRange, what they _want_ to retrieve is the CompositeContinuationToken, since that token has everything they'd need to store in order to continue polling the change feed later. I'd expect us to just return the composite token as the `ContinuationToken` value.

sdk/data/azcosmos/cosmos_change_feed_response.go
```diff
@@ -0,0 +1,117 @@
...
+type ChangeFeedResponse struct {
+	// ResourceID is the unique identifier for the resource.
+	ResourceID string `json:"_rid"`
+	// Documents is a list of changed documents returned in the change feed.
+	Documents []json.RawMessage `json:"Documents"`
+	// Count is the number of documents returned in this page.
+	Count int `json:"_count"`
+
+	// Selected HTTP headers that we're retrieving from the response
+	ETag              string
```
analogrelay: The ETag is already exposed by the `azcosmos.Response` struct this type embeds.

sdk/data/azcosmos/cosmos_change_feed_response.go
```diff
@@ -0,0 +1,117 @@
...
+	ContinuationToken string
+	LSN               string
```
analogrelay: LSNs are generally an internal detail we don't expose to the user. What's the value of this property to the user?

analogrelay (changes requested):
Looks like a great start! I've got some questions and suggestions.

sdk/data/azcosmos/cosmos_container.go
adrianwclement: Not necessarily before merging. We have completed the large majority of these additions already in a separate [branch](https://github.com/arespinosa/azure-sdk-for-go/tree/feature/azcosmos/add-partition-key-continuation-support), but were intending to create a separate PR to help narrow scope for this large PR.

We'd be happy to add it here if its something you'd like us to include.

sdk/data/azcosmos/cosmos_feed_range.go
adrianwclement: I recall @FabianMeiswinkel mentioning that the initial iteration of ChangeFeedForEPKRange wouldn't handle partition splits or merges, since that requires more complex logic.

As a possible workaround, we could re-obtain the FeedRanges and PartitionKeyRanges for the container and re-call the function. I'm not entirely certain of the best implementation, but this might address the scenario.

sdk/data/azcosmos/cosmos_change_feed_response.go
arespinosa: I believe the user does need both continuation tokens. I think there's two scenarios, 

1. When the user uses a Feed Range as an option, it's because we want to either read the change feed for Epk ranges. Similarly, if a user wants to read the change feed for an entire container, this is when we'll return the Composite Continuation Token. It allows us to track progress across multiple ranges.

2. When the user uses any of the other header options(such as partition key-based queries) we will return the regular ContinuationToken as a string value representing the last position in the feed.

We got this concept from the js sdk, which has the [Change Feed Response](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/cosmosdb/cosmos/src/ChangeFeedResponse.ts) and the [Composite Continuation Token](https://github.com/Azure/azure-sdk-for-js/blob/6652c23d4b108a078f2e0f81c5caa8eef17b15d8/sdk/cosmosdb/cosmos/src/client/ChangeFeed/CompositeContinuationToken.ts) files.

sdk/data/azcosmos/cosmos_change_feed_response.go
arespinosa: We've removed the duplicate ETag field from ChangeFeedResponse. The ETag is now accessed through the azcosmos.Response. Thank you for that insight!

sdk/data/azcosmos/cosmos_change_feed_response.go
arespinosa: Initially, we included this value thinking it would be utilized. However, after reviewing our code further, we realized it wasn't needed and removed it from the file!

sdk/data/azcosmos/cosmos_feed_range.go
analogrelay: I'm OK with leaving this as-is for now. We should be able to update it later if we want to handle overlapping ranges.

sdk/data/azcosmos/cosmos_container.go
analogrelay: We'd prefer not having placeholder logic here at all, if possible. If you're working on a separate PR to add this, that's totally fine, it doesn't need to be in this PR. But I'd prefer that this branch return an error or panic rather than having production-unsafe placeholder logic in.

sdk/data/azcosmos/cosmos_change_feed_composite_continuation_token_test.go
kjg: Here's the [js version](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/cosmosdb/cosmos/src/client/ChangeFeed/CompositeContinuationToken.ts)

analogrelay:
Looks like a couple of the new change feed emulator tests are failing:

```
--- FAIL: TestEmulatorContainerChangeFeed (3.01s)
    --- PASS: TestEmulatorContainerChangeFeed/CompositeContinuationToken (0.08s)
    --- FAIL: TestEmulatorContainerChangeFeed/IfNoneMatchHeader (0.00s)
    --- FAIL: TestEmulatorContainerChangeFeed/IfModifiedSinceHeader (0.00s)
```

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5147746&view=logs&j=07a8bf49-95b8-5c7c-de1a-77230f20b6cf&t=54bb188b-9165-51d0-59ff-c33c3bef2fe5

analogrelay:
I'm not sure why the builds aren't automatically launching 🤔 , I'm seeing the same issue on one of my own PRs. Anyway, I'll kick it manually for now.

PR #24910, opened 2025-07-10
Release pipeline verifies major version suffix

jhendrixMSFT:
For v2+ releases, verify that the module's identity has a major version suffix that matches the const value's major version. If not then fail the build.

### DIFF
--- a/eng/scripts/validate_go_mod_major_version.ps1
+++ b/eng/scripts/validate_go_mod_major_version.ps1
@@ -22,23 +22,18 @@ if ($goModFile.Length -eq 0) {
 # fetch the major version suffix from go.mod if present
 $majorVersion = Get-GoModuleMajorVersion $goModFile
 
-$hasError = $false
 if ($majorVersion) {
     # go.mod has a major version
     # ensure it matches the major version defined in the constant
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:

eng/scripts/validate_go_mod_major_version.ps1
```diff
@@ -0,0 +1,48 @@
...
+} else {
+    # go.mod has no major version
+    # ensure the constant is v0 or v1
+    if ($modVersion -notmatch "^[01]\.") {
+        Write-Host "The module's identity is missing a major version suffix for const version $modVersion"
+        $hasError = $true
+    }
+}
+
+Pop-Location
```
weshaggard: If you have to push/pop I suggest using a try/finally to ensure you reset the working directory. However, if you can avoid changing the working directory that would be even better.

jhendrixMSFT: I'll take a look at that. I used `validate_go_mod.ps1` as a template which has the same pattern.

jhendrixMSFT: Fixed.

PR #24919, opened 2025-07-11
SDK MCP Server 

jeo02:
- Enable SDK MCP server so TypeSpec workflow can be started from language repo. This will also enable language team to use package readiness and release SDK tool.
- Also seems like you were excluding vscode twice. Not sure how the cspell was getting through I assume because it was there before the exclude was included.

### DIFF
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -1,0 +1,9 @@
+{  
+    "servers": {
+      "azure-sdk-mcp": {
+        "type": "stdio",
+        "command": "pwsh",        
+        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
+    },
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ _testmain.go
 *.swp
 *~
 .DS_Store
-.vscode
 .vs
 .idea
 
@@ -47,6 +46,7 @@ vendor/
 # vscode
 **/.vscode/*
 !.vscode/cspell.json
+!.vscode/mcp.json
 
 # api view file
 *.gosource


praveenkuttappan (approved):

.vscode/mcp.json
```diff
@@ -0,0 +1,9 @@
+{  
+    "servers": {
+      "azure-sdk-mcp": {
+        "type": "stdio",
+        "command": "pwsh",        
+        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
+    },
+  }
+}
```
RickWinter: Needs a new line
```suggestion
}

```

.github/copilot-instructions.md
```diff
@@ -3,3 +3,23 @@ You are an expert Go programmer that attempts to answer questions and provide co
...
 - To determine who owns a module, use the [CODEOWNERS file](https://github.com/Azure/azure-sdk-for-go/tree/main/.github/CODEOWNERS), and find the line that matches the module path. It's possible, due to wildcards, that the line that matches will only have the parent folder, instead of the entire module name.
+
+## SDK release
+
+There are two tools to help with SDK releases:
+- Check SDK release readiness
+- Release SDK
+
+### Check SDK Release Readiness
+Run `CheckPackageReleaseReadiness` to verify if the package is ready for release. This tool checks:
```
richardpark-msft, line 14: @praveenkuttappan, we should probably consider having an .md in the sdk-tools repo (or replicated here), and just referring to it here (to avoid duplicating the text all over the place).

richardpark-msft: Or maybe we can export these as prompts directly from the DevEx agent itself, since these tools wouldn't be available unless they actually have it running.

RickWinter (approved):

.github/copilot-instructions.md
jeo02: I'll open an issue for this

jeo02: > Or maybe we can export these as prompts directly from the DevEx agent itself, since these tools wouldn't be available unless they actually have it running.

But then it would be a mcp prompt which isn't discoverable unless user uses the slash command in the chat.

richardpark-msft: I can't imagine something _more_ discoverable than typing '/'!

(Let me amend this - '/' works in the Github interface for bringing up snippets, works in teams for activating actions, and it works in my heart for activating emotions. It feels as close to universal as we could get)

PR #25022, opened 2025-07-29
BearerTokenPolicy ensures authN errors are NonRetriable

chlowell:
They normally are in the one case I'm modifying here, but only because all `azidentity` errors are `NonRetriable`. Custom credential implementations may return non-NonRetriable errors and thereby cause RetryPolicy to send doomed retries.

### DIFF
--- a/sdk/azcore/runtime/policy_bearer_token.go
+++ b/sdk/azcore/runtime/policy_bearer_token.go
@@ -97,7 +97,9 @@ func (b *BearerTokenPolicy) authenticateAndAuthorize(req *policy.Request) func(p
 		as := acquiringResourceState{p: b, req: req, tro: tro}
 		tk, err := b.mainResource.Get(as)
 		if err != nil {
-			return err
+			// consider this error non-retriable because if it could be resolved by
+			// retrying authentication, the credential would have done so already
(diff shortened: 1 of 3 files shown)


sdk/azcore/runtime/policy_bearer_token.go
```diff
@@ -97,7 +97,9 @@ func (b *BearerTokenPolicy) authenticateAndAuthorize(req *policy.Request) func(p
 		as := acquiringResourceState{p: b, req: req, tro: tro}
 		tk, err := b.mainResource.Get(as)
 		if err != nil {
-			return err
+			// consider this error non-retriable because if it could be resolved by
+			// retrying authentication, the credential would have done so already
+			return errorinfo.NonRetriableError(err)
```
chlowell, line 102: #25021 makes this an almost no-op for the common case of errors returned by `azidentity`

richardpark-msft: Well...not a no-op, but at least it's now a single line for callers... :)

chlowell: Someone has to throw away that pointer 😁

PR #25064, opened 2025-08-08
Support module config with major version suffix

tadelesh:
This PR update generator tool to adopt the new emitter config changes.

1. Remove `version-number` config for `release` and `refresh` command because we have no scenario to hard code the module version. Accordingly, remove the `SpecificVersion` in `GenerateParam`.
2. Support `module` setting with major version suffix to extract the module path.


JiaqiZhang-Dev:
LGTM. It's better to update the changelog

JiaqiZhang-Dev (approved):

PR #25075, opened 2025-08-12
add parse link key

jliusan:
link in issue changes ,
 new example:https://github.com/Azure/sdk-release-request/issues/6478
 old example:https://github.com/Azure/sdk-release-request/issues/6473

### DIFF
--- a/eng/tools/generator/cmd/issue/request/request.go
+++ b/eng/tools/generator/cmd/issue/request/request.go
@@ -53,6 +53,7 @@ const (
 
 const (
 	linkKeyword        = "**Link**: "
+	linkKeywordNew     = "**API spec pull request link**: "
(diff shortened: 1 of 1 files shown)


eng/tools/generator/cmd/issue/request/request.go
```diff
@@ -53,6 +53,7 @@ const (
 
 const (
 	linkKeyword        = "**Link**: "
+	linkKeywordNew     = "**API spec pull request link**: "
```
tadelesh: Can we change to a regex comparison?

jliusan: It's not very convenient to achieve this using a regular expression matching variable.

jliusan: changed to regex comparison

PR #25077, opened 2025-08-12
cosmos: enable support for reading the change feed for through partition keys

adrianwclement:
This PR adds a function to the Go SDK as a Cosmos container operation to retrieve the Change Feed for a specific `PartitionKey`. It also configures the Change Feed to read from a single Partition when not provided a `FeedRange` or `PartitionKey`.

It adds a `ContinuationTokenForPartitionKey`—conceptually similar to a `CompositeContinuationToken`—to resume consumption without duplication when reading through the Change Feed. Additionally, this change honors conditional request headers (e.g., `If-None-Match` and `If-Modified-Since`), so callers can gain fine-grained control over incremental change processing, can reduce bandwidth on idle intervals, and can safely resume from previously persisted positions.

## Notes

- This PR builds on:
  - https://github.com/Azure/azure-sdk-for-go/pull/24835
  - https://github.com/Azure/azure-sdk-for-go/pull/24889
  - https://github.com/Azure/azure-sdk-for-go/pull/24898

### DIFF
--- a/sdk/data/azcosmos/cosmos_change_feed_request_options.go
+++ b/sdk/data/azcosmos/cosmos_change_feed_request_options.go
@@ -9,12 +9,6 @@ import (
 	"time"
 )
 
-const (
-	// cosmosDefaultMaxItemCount represents unlimited items in the response
-	cosmosDefaultMaxItemCount = -1
-	cosmosBaseTen             = 10
-)
-
 // ChangeFeedResourceType represents the resource type for change feed operations.
 type ChangeFeedResourceType int
 
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Added support for specifying Vector Search indexing policies when creating a container. See [PR 24833](https://github.com/Azure/azure-sdk-for-go/pull/24833)
 * Added support for reading Feed Ranges from a container. See [PR 24889](https://github.com/Azure/azure-sdk-for-go/pull/24889)
 * Added support for reading Change Feed through Feed Ranges from a container. See [PR 24898](https://github.com/Azure/azure-sdk-for-go/pull/24898)
-* Added support for reading Change Feed through Parition Keys from a container. Also configures Change Feed to read from a single partition when not provided a Feed Range or Partition Key. See [PR 25077] (https://github.com/Azure/azure-sdk-for-go/pull/25077)
+* Added support for reading Change Feed through Partition Keys from a container. Also configures Change Feed to read from a single partition when not provided a Feed Range or Partition Key. See [PR 25077](https://github.com/Azure/azure-sdk-for-go/pull/25077)
 
 ### Breaking Changes
 


sdk/data/azcosmos/cosmos_change_feed_continuation_token_for_partition_key.go
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License.
+
+package azcosmos
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+// Version 1 is the initial version of the composite continuation token.
+const cosmosContinuationTokenForPartitionKeyVersion = 1
+
+type continuationTokenForPartitionKey struct {
```
analogrelay, line 11: Rather than making this a separate type, could the `PartitionKey` property just be added to the current Composite Continuation Token struct as `json:omitempty`? Then, when consuming the token, the code can just check for the presence or absence of the `PartitionKey` property to determine if the token is for a specific partition.

sdk/data/azcosmos/cosmos_change_feed_request_options.go
```diff
@@ -57,18 +71,27 @@ func (options *ChangeFeedOptions) toHeaders(partitionKeyRanges []partitionKeyRan
...
+			if options.PartitionKey == nil && continuationTokenForPartitionKey.PartitionKey != nil {
+				pkBytes, err := json.Marshal(continuationTokenForPartitionKey.PartitionKey)
+				if err == nil {
+					var pk PartitionKey
+					err = json.Unmarshal(pkBytes, &pk)
+					if err == nil {
+						options.PartitionKey = &pk
+					}
+				}
+			}
```
analogrelay, line 88: If the continuation tokens are the same type (see comment above), this can basically just be moved into the other branch (the one that handles the previous type) since all it's really doing is setting the `PartitionKey` option from the token IF the token contains a partition key.

sdk/data/azcosmos/cosmos_change_feed_response.go
```diff
@@ -108,3 +111,44 @@ func (c ChangeFeedResponse) GetCompositeContinuationToken() (string, error) {
...
+// GetPartitionKey returns the partition key used in the request, if any.
+func (c ChangeFeedResponse) GetPartitionKey() (partitionKey *PartitionKey, ok bool) {
+	if c.PartitionKey != nil {
+		return c.PartitionKey, true
+	}
+	return nil, false
+}
+
+// GetContinuationTokenForPartitionKey creates a continuation token for a specific partition key.
+func (c ChangeFeedResponse) GetContinuationTokenForPartitionKey() (string, error) {
```
analogrelay, line 134: I'm a little unclear on why there would be a separate function to get a token for a specific partition. My understanding is that the `ChangeFeedResponse` will indicate the partition keys covered IF it was constrained to a specific partition key, and if it wasn't constrained to a partition that value would be nil. That means we'd be able to have a single `GetContinuationToken` that returns an appropriate token in both cases, depending on the presence or absence of the `PartitionKey` in the `ChangeFeedResponse`.

PR #25082, opened 2025-08-13
Standardized CODEOWNERS File Request

michiisai:
# Request for a standard CODEOWNERS file across all SDK Language Repositories

CODEOWNER changes:
* Renamed the `SDK (track2)` section to `Client Libraries` for clearer categorization and standardization across language repositories 
* Added a section header for `Management Libraries` to separate them from client libraries.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,9 +1,9 @@
 # Instructions for CODEOWNERS file format and automatic build failure notifications:
 # https://github.com/Azure/azure-sdk/blob/master/docs/policies/opensource.md#codeowners
 
-#############
-# SDK (track2)
-#############
+####################
+# Client Libraries
+####################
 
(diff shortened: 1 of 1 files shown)


RickWinter (approved):
The Client Libraries and Mgmt Libraries distinction feels unnecessary, you may wnat to consider not having that seperation. It's "shipping" the org chart.

ReilleyMilne:
> The Client Libraries and Mgmt Libraries distinction feels unnecessary, you may wnat to consider not having that seperation. It's "shipping" the org chart.

@RickWinter Just to clarify, do you mean just for the go repository? The only reason I ask is because we've already updated some of the categories in other language repositories.

RickWinter:
Its a NIT as the distinction seems unnecessary. It's nice to have all the repos be similar in structure. Thus I'd keep it if it's everywhere else.

Issue #25036, opened 2025-07-31
arm.ParseResourceType incorrectly parses resource type for nested CognitiveServices deletedAccounts


ms-henglu:
### Problem
The `arm.ParseResourceType` function is incorrectly parsing the resource type for CognitiveServices deletedAccounts resources.

### Expected Behavior
For the resource ID:
```
/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/providers/Microsoft.CognitiveServices/locations/eastus/resourceGroups/avm-res-cognitiveservices-account-rg-p26m/deletedAccounts/OpenAI-cog-p26m
```

The function should return:
```
Microsoft.CognitiveServices/locations/resourceGroups/deletedAccounts
```

### Actual Behavior
The function currently returns:
```
Microsoft.Resources/resourceGroups/deletedAccounts
```

### Analysis
The issue appears to be that the parser is incorrectly identifying `resourceGroups` as the primary resource provider namespace instead of recognizing it as part of the nested resource hierarchy under `Microsoft.CognitiveServices/locations`.

### Environment
- Package: `github.com/Azure/azure-sdk-for-go/sdk/azcore`
- Function: `arm.ParseResourceType`

### Steps to Reproduce
```go
resourceID := "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/providers/Microsoft.CognitiveServices/locations/eastus/resourceGroups/avm-res-cognitiveservices-account-rg-p26m/deletedAccounts/OpenAI-cog-p26m"
resourceType := arm.ParseResourceType(resourceID)
fmt.Println(resourceType) // Currently prints: Microsoft.Resources/resourceGroups/deletedAccounts
```

PR #25115, opened 2025-08-20
fix resource identifier parsing

tadelesh:

Fix: https://github.com/Azure/azure-sdk-for-go/issues/25036


sdk/azcore/arm/internal/resource/resource_identifier_test.go
jhendrixMSFT, line 1: I worry that our test coverage is incomplete. Are there any missing scenarios?

ArcturusZhang (commented):
maybe there are other cases that our current test cases do not cover (we could always have this), but I think it is impossible that we could exhaust all the possible cases.
from this issue, I could see the root cause is that one of the resource type segment happens to be the same as "resource group", we did not handle this correctly previously.
so if this is the root cause, I think it is reasonable that we add a case that the resource type segment has "subscriptions" (maybe we already have this case)

sdk/azcore/arm/internal/resource/resource_identifier_test.go
ArcturusZhang: It is possible but I think we could never know that now.

tadelesh: We could only cover the scenario we know for now. I refined the logic about `resourceGroup` resource and `subscription` resource and add two tests.

sdk/azcore/arm/internal/resource/resource_identifier.go
```diff
@@ -123,9 +123,9 @@ func newResourceIDWithProvider(parent *ResourceID, providerNamespace, resourceTy
 }
 
 func chooseResourceType(resourceTypeName string, parent *ResourceID) ResourceType {
-	if strings.EqualFold(resourceTypeName, resourceGroupsLowerKey) && parent != nil && parent.ResourceType.String() == SubscriptionResourceType.String() {
+	if strings.EqualFold(resourceTypeName, resourceGroupsLowerKey) && isSubscriptionResource(parent) {
```
jhendrixMSFT, line 126: The impl of `isSubscriptionResource()` has a slightly different condition that what's replaced (the addition of `isTenantResource(resourceID.Parent)`). Is this intentional?

Comment applies to other places that use the new helpers.

sdk/azcore/arm/internal/resource/resource_identifier.go
```diff
@@ -225,3 +225,15 @@ func splitStringAndOmitEmpty(v, sep string) []string {
 
 	return r
 }
+
+func isTenantResource(resourceID *ResourceID) bool {
+	return resourceID != nil && strings.EqualFold(resourceID.ResourceType.String(), TenantResourceType.String()) && resourceID.Parent == nil
```
jhendrixMSFT, line 231: Can we get some comments in all these helpers that explain how they come to their "conclusion"?

tadelesh: Added.

sdk/azcore/arm/internal/resource/resource_identifier.go
tadelesh: Yes. I fixed all the place that specialize handle the subscription resource and resource group resource, to prevent some resource with pattern of `resourceGroups` and `subscriptions`, but is not a real resource group resource or subscription resource.

PR #25146, opened 2025-08-26
[Release] sdk/resourcemanager/playwright/armplaywright/1.0.0 

jliusan:
https://github.com/Azure/sdk-release-request/issues/6488


tadelesh (approved):

sdk/resourcemanager/playwright/armplaywright/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Release History
 
+## 1.0.0 (2025-08-26)
+### Breaking Changes
+
+- Type of `Quota.Name` has been changed from `*QuotaName` to `*string`
```
mjmadhu, line 6: Is this intended or correct as the Typespec for the API still mentions the type as QuotaName - https://github.com/Azure/azure-rest-api-specs/blob/bd1d56e67e0e8d45fac093017d018118bc61d4c6/specification/loadtestservice/resource-manager/Microsoft.LoadTestService/playwright/typespec/quotas.tsp#L14 ?

mjmadhu: //cc: @tadelesh

tadelesh: It is intended. The breaking change is from our emitter bug fix. The `name` of the `Quota` resource should be `string`, and the query parameter for the `Quota` resource method should be `QuotaName`.

mjmadhu: Thanks for confirming.

JiaqiZhang-Dev: @mjmadhu Please hold on this SDK release since we need to confirm with architect side that if this breaking change could be accepted

Issue #25089, opened 2025-08-16
[Automation] Use the latest tsp-client even the version is locked in the init script


raych1:
### Bug Report

For the scripts used by the Go SDK generation pipeline, the init script lock the version of `tsp-client` to be installed. However, the latest version is used in the generate script.

It should use the locked version in the generate script.

Here is an [example pipeline run](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5226204&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=307) with this issue:

<img width="1731" height="466" alt="Image" src="https://github.com/user-attachments/assets/49f4065c-e280-4d27-a9a3-fb759430c063" />

CC: @tadelesh

tadelesh:
@raych1 I think the problem is not in SDK side. From the pipeline log, you could see the `PATH` prioritized the spec repo's node module folder than system's global node module folder , so any `tsp-client` cmd will use the version spec repo configured. 

<img width="1782" height="145" alt="Image" src="https://github.com/user-attachments/assets/9a34a1ab-4819-4065-b55e-42bf93a7e390" />

raych1:
@tadelesh , it should remove the `-g` flag while installing `tsp-client` at the ['init script'](https://github.com/Azure/azure-sdk-for-go/blob/51f157b6bca9ac20d0a3ab6604ee3c00cb2338e3/eng/scripts/automation_init.sh#L82). Additionally, it should call `npx tsp-client` in the 'generate script' to use the version installed to the SDK repo.

tadelesh:
I don't think we should remove the `-g`. We do not have `package.json` in our SDK repo and install a npm package locally in the SDK repo is weird. And Go release tool could be run in anywhere, rely on the SDK repo's node modules is not flexible. Why we add the local spec node modules in the gloabl `PATH`?

tadelesh:
Since Java and Python use the latest version of tsp-client, Go decided to follow them.

raych1:
> I don't think we should remove the `-g`. We do not have `package.json` in our SDK repo and install a npm package locally in the SDK repo is weird. And Go release tool could be run in anywhere, rely on the SDK repo's node modules is not flexible. Why we add the local spec node modules in the gloabl `PATH`?

[...]

PR #25156, opened 2025-08-27
Pin tsp-client version in SDK automation

tadelesh:
Fix: https://github.com/Azure/azure-sdk-for-go/issues/25089

### DIFF
--- a/eng/scripts/automation_init.sh
+++ b/eng/scripts/automation_init.sh
@@ -79,4 +79,4 @@ else
 fi
 
 echo Install tsp-client
-npm install -g @azure-tools/typespec-client-generator-cli@latest 2>&1
+npm install -g @azure-tools/typespec-client-generator-cli@0.28.1 2>&1
--- a/eng/tools/generator/cmd/v2/common/cmdProcessor.go
+++ b/eng/tools/generator/cmd/v2/common/cmdProcessor.go
@@ -200,7 +200,8 @@ func ExecuteGoFmt(dir string, args ...string) error {
 
 // execute tsp-client command
 func ExecuteTspClient(path string, args ...string) error {
-	cmd := exec.Command("tsp-client", args...)
+	args = append([]string{"tsp-client"}, args...)
+	cmd := exec.Command("npx", args...)
 	cmd.Dir = path
 
 	stdoutPipe, err := cmd.StdoutPipe()


eng/scripts/automation_init.sh
```diff
@@ -79,4 +79,4 @@ else
 fi
 
 echo Install tsp-client
-npm install -g @azure-tools/typespec-client-generator-cli@v0.21.0 2>&1
+npm install -g @azure-tools/typespec-client-generator-cli@latest 2>&1
```
weshaggard: Is there a reason you are picking latest instead of pinning? That may very well lead to reliability issues.

@catalinaperalta what are other languages doing here?

catalinaperalta: I discussed this with @raych1 about a year ago that we should pin tsp-client precisely to improve reliable performance for automation pipelines. Folks should make a conscious decision to upgrade to a new tsp-client version.

catalinaperalta: I believe all of the pipelines were updated to pin to a specific tsp-client version last year

tadelesh: Two problems here:
1. Java, Python does not pin to a specific version
2. The SDK automation pipeline has a global PATH setting to prioritize to use the spec repo's tsp-client version which makes this setting useless.

image

raych1: SDK repo shouldn't reply on the installed 'tsp-client' from the global cache, which is for spec repo purposes. I've suggested the right approach at [this comment](https://github.com/Azure/azure-sdk-for-go/issues/25089#issuecomment-3202482639).

There're separated issues created for [Python](https://github.com/Azure/azure-sdk-tools/issues/11221) and [Java](https://github.com/Azure/azure-sdk-tools/issues/11222) to pin the version of 'tsp-client'. Actually, @weidongxu-microsoft already has an [ongoing PR](https://github.com/Azure/azure-sdk-for-java/pull/46037) for this change request.
CC: @mikeharder

mikeharder: The only reliable way to install an NPM package with deps, is by installing from a package-lock.json which pins the versions of all packages including transitive deps.  Pinning just the top-level dependency is not enough (the previous code).  Installing `latest` of a top-level dependency is even riskier.

raych1: I agree that it is a best practice to introduce a 'package-lock.json' to lock dependencies and use 'npm ci' command for installation. 
On the other hand, most SDK language repositories don't have a root 'package.json' or lock file, and their automation tools are not NPM libraries, so they also lack a 'package.json'. Adding a 'package.json' and lock file solely for `tsp-client` could increase maintenance overhead. 

Would it make sense as a first step to pin the top-level dependency instead?

mikeharder: I would go all the way and use a lockfile.  Then if an SDK repo wants to add more JS-based tools, they're already setup.

Pinning just top-level deps is better than nothing, but be aware you can still break at any time if a transitive dep updates.

tadelesh: @raych1 As I mentioned, only pin the top-level dependency is not enough because the PATH has been overwritten, so I don't think weidong's PR could work. Then, the only solution is to add `package.json` and lock local dependency. I'll discuss the solution inside the Go team.

raych1: > only pin the top-level dependency is not enough because the PATH has been overwritten.

@tadelesh it will work by using NPX since this command prioritize the local 'node_modules' over the global cache.

tadelesh:
@raych1 Before eng add `package-lock.json`, is current PR meet your requirement? If so, could you help to approve and merge? Thanks.

eng/tools/generator/cmd/v2/common/cmdProcessor.go
```diff
@@ -200,7 +200,8 @@ func ExecuteGoFmt(dir string, args ...string) error {
 
 // execute tsp-client command
 func ExecuteTspClient(path string, args ...string) error {
-	cmd := exec.Command("tsp-client", args...)
+	args = append([]string{"tsp-client"}, args...)
+	cmd := exec.Command("npx", args...)
```
raych1, line 204: Just a reminder, it will need to use `tsp-client init --update-if-exists` for latest version. But, the change can be in a separate PR if automation needs more change to consume 'emitter-output-dir' and deprecate 'package-dir'.

raych1:
> @raych1 Before eng add `package-lock.json`, is current PR meet your requirement? If so, could you help to approve and merge? Thanks.

@tadelesh I approved it if you prefer to have 'tsp-client' command change in another separate PR.

eng/tools/generator/cmd/v2/common/cmdProcessor.go
raych1: I saw this change is already in the tool. Good to merge. @tadelesh

mikeharder: For automation, it's better to use `npm exec --no -- tsp-client` than `npx tsp-client`, for reasons like avoiding CLI argument confusion.

raych1: As discussed, we will use the lock file soon.

mikeharder: You should be able to replace `npx` with `npm exec --no --`, with or without package-lock.json.

PR #25172, opened 2025-08-31
Update Changelog After 1.4.1

tvaron3:
Updating changelog after 1.4.1 release.

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -20,7 +20,6 @@
 ## 1.4.1 (2025-08-27)
 
 ### Bugs Fixed
-
 * Fixed bug where the correct header was not being sent for writes on multiple write region accounts. See [PR 25127](https://github.com/Azure/azure-sdk-for-go/pull/25127)
 
 ## 1.5.0-beta.0 (2025-06-09)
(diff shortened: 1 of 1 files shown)


NaluTripician (approved):

sdk/data/azcosmos/CHANGELOG.md
```diff
@@ -17,6 +17,11 @@
 
 ### Other Changes
 
+## 1.4.1 (2025-08-27)
```
analogrelay: I'm not entirely sure the expected ordering for Changelog entries, is it always date ordered or is it version ordered? If it's date ordered, this is correct, but it looks odd to see `1.5.0-beta.1`, then `1.4.1`, then `1.5.0-beta.0`, and finally `1.4.0`. If that's the pattern we use elsewhere, that's just fine, but I wanted to double check.

analogrelay (approved):
Approving, assuming my comment isn't relevant, and the ordering is correct based on how we structure our Changelogs.

sdk/data/azcosmos/CHANGELOG.md
tvaron3: So in java, we have tried changing it to be sorted by version because have gotten feedback from customers that it is confusing by date but once the release script is called from azure core it will rewrite the changelog to be sorted by date (at least in java). I will change it to be sorted by version and next release we can see what happens :)

analogrelay (approved):

PR #25186, opened 2025-09-04
CLI credentials handle claims

chlowell:
...by returning errors explaining what must be done before they can acquire a token. For AzureCLICredential this means GetToken always returns an error when given claims because the user must run `az login`. AzureDeveloperCLICredential runs `azd` with the given claims and returns an error only when `azd` doesn't return a token. Most of this PR is from refactoring these credentials to share code for executing commands in a shell (doing so made the feature easier to implement and test).

### DIFF
--- a/sdk/azidentity/developer_credential_util.go
+++ b/sdk/azidentity/developer_credential_util.go
@@ -7,22 +7,73 @@
 package azidentity
 
 import (
+	"bytes"
+	"context"
 	"errors"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
 	"time"
 )
 
 // cliTimeout is the default timeout for authentication attempts via CLI tools
 const cliTimeout = 10 * time.Second
 
-// unavailableIfInChain returns err or, if the credential was invoked by DefaultAzureCredential, a
+// executor runs a command and returns its output or an error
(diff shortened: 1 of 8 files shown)


sdk/azidentity/developer_credential_util.go
```diff
@@ -7,22 +7,73 @@
...
 // credentialUnavailableError having the same message. This ensures DefaultAzureCredential will try
 // the next credential in its chain (another developer credential).
-func unavailableIfInChain(err error, inDefaultChain bool) error {
-	if err != nil && inDefaultChain {
-		var unavailableErr credentialUnavailable
-		if !errors.As(err, &unavailableErr) {
-			err = newCredentialUnavailableError(credNameAzureDeveloperCLI, err.Error())
-		}
+func unavailableIfInDAC(err error, inDefaultChain bool) error {
+	if err != nil && inDefaultChain && !errors.As(err, new(credentialUnavailable)) {
```
jhendrixMSFT, line 75: nit: I wonder if the previous impl didn't allocate (now we do).

chlowell: The only practical difference is that they call different constructors when they have to convert an error. The constructor called by the old impl allocates an additional string, so the new impl allocates less in that case

Issue #22303, opened 2024-01-26
[azappconfig] Support LabelFilter.Null


maffin-sa:
### Feature Request

I am using Dapr to connect to Azure AppConfig (https://github.com/dapr/components-contrib/tree/main/configuration/azure/appconfig) which uses this library behind the scenes and am trying to replicate the functionality described here https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-best-practices#key-value-compositions where you can request *only* config items that have a `null` label (I don't care about the `.Select(...` chaining in the example, I'm fine w/ making multiple queries myself). 
The linked page on learn.microsoft.com references the [LabelFilter.Null](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azureappconfiguration.labelfilter.null?view=azure-dotnet#microsoft-extensions-configuration-azureappconfiguration-labelfilter-null) value, but I can't tell what that value is converted to behind the scenes; I have tried passing in "Null", "null", "LabelFilter.Null", "" and "(No label)" as `label` values and none of those return any values, so I assume they are just looking for those specific labels as opposed to the `null` label.

If there is/could be a string value I could pass in as a label that would trigger the "only items with `null` label" filtering that would be sufficient for my purposes.

tlund101:
@maffin-sa I stumbled across your issue when looking to file one of my own.  I'm able to use this URL to send an HTTP request to Dapr and fetch only those configs without a label: `http://localhost:3500/v1.0/configuration/azureappconfig?metadata.label=%00`

Does that suffice for your problem?

RichardChen820:
@maffin-sa @tlund101 passing string `\x00` to labelFilter should work, it represents NULL char ASCII in Golang. It's indeed confusion leading, [LabelFilter.Null](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.azureappconfiguration.labelfilter.null?view=azure-dotnet#microsoft-extensions-configuration-azureappconfiguration-labelfilter-null) in .net SDK is also this value, we should have pre-defined variable in Go as well.

PR #25194, opened 2025-09-08
Tags filter support

RichardChen820:
**Tags filter support:**

* Added a `Tags` field to the `AddSettingOptions` and `SetSettingOptions` structs, enabling users to assign key-value metadata tags when creating or updating configuration settings. 
* Updated the `AddSetting` and `SetSetting` methods in `client.go` to include the new `Tags` field when constructing a `Setting`. 
* Modified the `SettingSelector` struct to support a `TagsFilter` field, allowing retrieval of settings filtered by tags. The filter is passed to the generated API calls.

**Documentation and examples:**

* Updated the README to describe how tags can be used for metadata and filtering, with examples of tag-based queries.
* Added a new example in `examples_test.go` demonstrating how to create a setting with tags and retrieve settings using a tag filter.

**REST API version update:**

* Bumped the Azure App Configuration REST API version from `2023-10-01` to `2023-11-01` in `autorest.md` and `appconfiguration_ext.json` to enable tag features.


sdk/data/azappconfig/options.go
```diff
@@ -19,6 +19,10 @@ type AddSettingOptions struct {
 
 	// Configuration setting label.
 	Label *string
+
+	// A dictionary of tags used to assign additional properties to a configuration setting.
+	// These can be used to indicate how a configuration setting may be applied.
+	Tags map[string]string
```
linglingye001: why using map[string]string in `AddSettingsOptions`? Is it by design?

RichardChen820: This should be a designed API, it should've tags support at the time of adding `AddSetting()` API, we need to confirm with the initial author about the design of it (asking them to review this PR is good enough I think).

Here. I just speculate the initial intention and propose my fix to this bug, two reasons:
1. It initializes the [`setting`](https://github.com/Azure/azure-sdk-for-go/blob/038f9dc38aaad217983ddbefe5314acfe888bc0b/sdk/data/azappconfig/client.go#L101C1-L101C101) and pass it to `PutKeyValue` function, and tags in `Setting` type is map[string]string , we just keep consistency is fine
2. It comply with other languages e.g., [dotnet](https://github.com/Azure/azure-sdk-for-net/blob/50485990713dd742ce3127ceff833bed2eda4c87/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L279C8-L308C10), `Tags` is also a dictionary property of `ConfigurationSetting` while adding the setting. dotnet has a slightly different API design becuase it has method overload, but `Go` doesn't, so I think that's why the caller needs to pass `Tags` in `AddSettingsOptions`. But its type is clear, it is supposed to be a dictionary in go as well.

sdk/data/azappconfig/setting.go
```diff
@@ -78,7 +78,8 @@ func toGeneratedETagString(etag *azcore.ETag) *string {
 func (cs Setting) toGenerated() generated.KeyValue {
 	tags := make(map[string]*string)
```
linglingye001: In generated.KeyValue, the tags is `map[string]*string`, and in `Setting` we have `map[string]string`, do you know why?

RichardChen820: Have no idea about the initial reason for doing so, may need input from @jhendrixMSFT @RickWinter

zhiyuanliang-ms (approved):

RickWinter:
@jhendrixMSFT Can you review. Also since the appconfig team now owns this SDK we should add the appropriate folks to the CODEOWNER file.

sdk/data/azappconfig/setting.go
jhendrixMSFT: The generator emits as `map[string]*type` (where `type` isn't implicitly nil-able) for a few reasons.
- this is for compat when switching to/from a dictionary with nullable values (we don't want such changes to be breaking)
- some services support sending JSON `null` for dictionary values to mean "delete this entry"

If you know your service will never need the map's value type to be nil-able, then exposing it as `map[string]string` is the right call as it has better ergonomics.

sdk/data/azappconfig/options.go
jhendrixMSFT: All correct. A method's options-type (i.e. the last param where the type's suffix is `Options`) is an extensibility point to add optional method params due to lack of overloads.

Note that for fields in these types, the zero-value _must not_ be ambiguous.  i.e. the zero-value means "don't send anything for this field" which is why a lot of fields are pointer-to-type.

Since maps are nil-able, what you have makes sense.

jhendrixMSFT:
@RichardChen820 can you please help us get the list of appropriate aliases for us to update the `CODEOWNERS` file?

sdk/data/azappconfig/README.md
jhendrixMSFT, line 1: Please also update the `CHANGELOG.md` file noting the new features.

In addition, the version must also be updated to `1.3.0` in the changelog and [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azappconfig/version.go#L11).

Do you think this should go out as an initial beta? If so, then keep the `-beta.1` suffix.

RichardChen820:
> @RichardChen820 can you please help us get the list of appropriate aliases for us to update the `CODEOWNERS` file?

Please add me and Lingling: @RichardChen820 @linglingye001

sdk/data/azappconfig/README.md
RichardChen820: I think we can just go with 1.3.0 stable. Is the 1.2.1-beta.1 planned but hasn't been released yet?

sdk/data/azappconfig/setting.go
RichardChen820: I checked our dataplane backend, and we don’t have any restriction on non-null tag values on the server side, null value is allowed. I also reviewed the[ .NET SDK implementation](https://github.com/Azure/azure-sdk-for-net/blob/c0e3f8082be4812792466d68e068587370fe891f/sdk/appconfiguration/Azure.Data.AppConfiguration/src/Models/ConfigurationSetting.cs#L26C1-L26C9), and it doesn’t prohibit null values either. So I’m leaning toward using `map[string]*string` for `Tags` in `Setting`. However, this would be a breaking change. Any concerns? @jhendrixMSFT

jhendrixMSFT: I don't think we should make such a change unless there's loss of functionality and a demand for it to be fixed. Does the exclusion of `nil` for a value preclude any scenarios from working?

sdk/data/azappconfig/README.md
jhendrixMSFT: No. After `v1.2.0` was released, the version is automatically bumped to the next patch-beta version as a placeholder.

RickWinter:
> > @RichardChen820 can you please help us get the list of appropriate aliases for us to update the `CODEOWNERS` file?
> 
> Please add me and Lingling: @RichardChen820 @linglingye001

You can edit the CODEOWNERS file and submit as part of this PR or you can make a separate PR. I'll approve

RichardChen820:
> You can edit the CODEOWNERS file and submit as part of this PR or you can make a separate PR. I'll approve

@RickWinter please approve this add owner PR https://github.com/Azure/azure-sdk-for-go/pull/25313

sdk/data/azappconfig/setting.go
RichardChen820: This does introduce some functionality issues. As I mentioned, since our dataplane backend allows users to assign a `null pointer` value to a tag, under the current implementation the tag would be dropped when the user retrieves the key-value. This is clearly problematic because an item that should exist gets lost.

If we instead treat `null` as an empty string, that leads to another issue: when the user writes the value back, its original value would be unintentionally altered (from `null` to empty string).

RichardChen820: So I’d prefer to make the change now, rather than waiting until users start complaining about it.

jhendrixMSFT: Got it, thanks for clarifying.

I still hesitate to fix this though as this has been like this for almost two years and nobody complained about it. Plus, introducing a new major version does come with some customer cost (and some cost for us, like updating docs/examples etc).

That said, if you think the cost is justified, I'll go along with it.

Ok after writing the above, I was going through the issues, and I found [this one](https://github.com/Azure/azure-sdk-for-go/issues/22303). Is this due to what we're discussing?

RichardChen820: That’s I think why we should make the breaking change for doing the right thing now, while there are still few users using Tags in Go,

RichardChen820: [This one ](https://github.com/Azure/azure-sdk-for-go/issues/22303) is another problem, I think we should add a variable to indicate the "Null" value of label to stop confusing users.

sdk/data/azappconfig/version.go
```diff
@@ -8,5 +8,5 @@ package azappconfig
 
 const (
 	moduleName    = "github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig"
-	moduleVersion = "v1.2.1-beta.1"
+	moduleVersion = "v1.3.0"
```
jhendrixMSFT: Breaking changes must increment the major version number, so this would become `v2.0.0`.  In addition, you'll need to perform the following changes.
- update the `go.mod` file to add a `/v2` major version suffix to the module's identity
- update all `import` statements to reference `v2`
- update all docs and examples to use `v2`

RichardChen820: @jhendrixMSFT Updated, please help take a look to see if I missed anything.

sdk/data/azappconfig/version.go
```diff
@@ -7,6 +7,6 @@
 package azappconfig
 
 const (
-	moduleName    = "github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig"
-	moduleVersion = "v1.2.1-beta.1"
+	moduleName    = "github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig/v2"
```
jhendrixMSFT: The name needs to omit the major version suffix.

sdk/data/azappconfig/version.go
jhendrixMSFT: The value for `moduleName` must not include the major version suffix. Everything else looks good.

sdk/data/azappconfig/version.go
RichardChen820: Ah, right, updated.

sdk/data/azappconfig/client_test.go
```diff
@@ -767,3 +767,478 @@ func CleanupSnapshot(client *azappconfig.Client, snapshotName string) error {
...
+	)
+
+	client := NewClientFromConnectionString(t)
+
+	// Test tags with special characters
+	tags := map[string]*string{
+		"app-name":    to.Ptr("my-app"),
+		"team.owner":  to.Ptr("backend-team"),
+		"cost_center": to.Ptr("engineering"),
+		"env_type":    to.Ptr("staging-test"),
```
avanigupta: Can we use some other special chars? `.`, `-` and `_` are not encoded in the URL if that's the special char handling you are trying to verify.

sdk/data/azappconfig/client_test.go
```diff
@@ -767,3 +767,478 @@ func CleanupSnapshot(client *azappconfig.Client, snapshotName string) error {
...
+	require.True(t, found, "Setting with special character tags not found")
+
+	// Clean up
+	_, err = client.DeleteSetting(context.Background(), key, &azappconfig.DeleteSettingOptions{
+		Label: to.Ptr("special-char-test"),
+	})
+	require.NoError(t, err)
+}
+
+func TestEmptyTagsHandling(t *testing.T) {
```
avanigupta, line 1027: Could you add a test for filtering by tags with empty or null value.
Eg: 
- `/kv?tags=EmptyTagVal=&api-version=...`
- `/kv?tags=NullTagVal=%00&api-version=...`

sdk/data/azappconfig/internal/generated/zz_azureappconfiguration_client.go
```diff
@@ -148,6 +148,11 @@ func (client *AzureAppConfigurationClient) checkKeyValuesCreateRequest(ctx conte
 	if options != nil && options.Snapshot != nil {
 		reqQP.Set("snapshot", *options.Snapshot)
 	}
+	if options != nil && options.Tags != nil {
+		for _, qv := range options.Tags {
+			reqQP.Add("tags", qv)
```
avanigupta, line 153: This looks like autogenerated code, but just to confirm - this will generate multiple occurrences of `tags` query parameter in the URL if there are multiple tag filters, right?
Eg: `/kv?tags={filter1}&tags={filter2}&tags={filter3}`

sdk/data/azappconfig/internal/generated/zz_options.go
```diff
@@ -49,17 +49,20 @@ type AzureAppConfigurationClientCheckKeyValuesOptions struct {
...
 	Key *string
 
-	// A filter used to match labels
+	// A filter used to match labels. Syntax reference: https://aka.ms/azconfig/docs/keyvaluefiltering
 	Label *string
 
 	// Used to select what fields are present in the returned resource(s).
 	Select []SettingFields
 
 	// A filter used get key-values for a snapshot. Not valid when used with 'key' and 'label' filters.
```
avanigupta:
```suggestion
	// A filter used get key-values for a snapshot. Not valid when used with 'key', 'label' or `tags` filters.
```

sdk/data/azappconfig/internal/generated/zz_options.go
```diff
@@ -196,6 +202,9 @@ type AzureAppConfigurationClientGetKeyValuesOptions struct {
 	// A filter used get key-values for a snapshot. The value should be the name of the snapshot. Not valid when used with 'key'
 	// and 'label' filters.
```
avanigupta:
```suggestion
	// A filter used get key-values for a snapshot. The value should be the name of the snapshot. Not valid when used with 'key',
	// 'label' or `tags` filters.
```

sdk/data/azappconfig/options.go
```diff
@@ -19,6 +19,10 @@ type AddSettingOptions struct {
 
 	// Configuration setting label.
 	Label *string
+
+	// A dictionary of tags used to assign additional properties to a configuration setting.
+	// These can be used to indicate how a configuration setting may be applied.
```
avanigupta: Not sure I follow the second line.
Can we just say "Configuration setting tags"? Or maybe keep only the first line of your summary?

sdk/data/azappconfig/setting_selector.go
```diff
@@ -30,6 +30,11 @@ type SettingSelector struct {
 	// Label filter that will be used to select a set of configuration setting entities.
 	LabelFilter *string
 
+	// Tags filter that will be used to select a set of configuration setting entities.
+	// This is a list of comma-separated tags. For more information about filtering by tags, see:
+	// https://learn.microsoft.com/azure/azure-app-configuration/rest-api-key-value?pivots=v23-11#filtering
```
avanigupta: Can use short link here

sdk/data/azappconfig/setting_selector.go
```diff
@@ -30,6 +30,11 @@ type SettingSelector struct {
 	// Label filter that will be used to select a set of configuration setting entities.
 	LabelFilter *string
 
+	// Tags filter that will be used to select a set of configuration setting entities.
+	// This is a list of comma-separated tags. For more information about filtering by tags, see:
```
avanigupta:
```suggestion
	// This is a list of tag filters in the format {tagName=tagValue}. For more information about filtering by tags, see:
```

sdk/data/azappconfig/options.go
RichardChen820: Make sense, let's only keep this first line.

sdk/data/azappconfig/internal/generated/zz_azureappconfiguration_client.go
RichardChen820: Right

sdk/data/azappconfig/client_test.go
RichardChen820: added

sdk/data/azappconfig/client_test.go
RichardChen820: updated

linglingye001 (approved):

sdk/data/azappconfig/client_test.go
```diff
@@ -1242,3 +1242,243 @@ func TestTagsFilteringEndToEnd(t *testing.T) {
...
+		require.Contains(t, foundKeys, "setting-with-normal-tags")
+		require.Contains(t, foundKeys, "setting-with-nil-tag-value")
+		require.Contains(t, foundKeys, "setting-with-empty-tag-value")
+		require.NotContains(t, foundKeys, "setting-mixed-tags") // has env=staging
+	})
+
+	// Test 2: Filter by tag key regardless of value (including nil/empty)
+	t.Run("TagKeyExistsFilter", func(t *testing.T) {
+		// Note: This test depends on how the backend handles tag key existence
+		// Some implementations might not return nil-valued tags
```
avanigupta: Backend does not support filtering by tag key only (or tag existence). The only supported scenario is filtering by exact match (case sensitive) of tag name and tag value both.

avanigupta: The name of the test and comment is misleading, but the test is correctly asserting that only "setting-with-normal-tags" should be returned.

sdk/data/azappconfig/client_test.go
```diff
@@ -1242,3 +1242,243 @@ func TestTagsFilteringEndToEnd(t *testing.T) {
...
+
+			for _, setting := range page.Settings {
+				if strings.HasPrefix(*setting.Key, "setting-") {
+					foundKeys = append(foundKeys, *setting.Key)
+				}
+			}
+		}
+
+		// Should find only the setting with service=api
+		require.Contains(t, foundKeys, "setting-with-normal-tags")
```
avanigupta, line 1377: Could you also do `require.Len(t, foundKeys, 1)` here?

sdk/data/azappconfig/setting_selector.go
```diff
@@ -31,8 +31,8 @@ type SettingSelector struct {
 	LabelFilter *string
 
 	// Tags filter that will be used to select a set of configuration setting entities.
-	// This is a list of comma-separated tags. For more information about filtering by tags, see:
-	// https://learn.microsoft.com/azure/azure-app-configuration/rest-api-key-value?pivots=v23-11#filtering
+	// This is a list of tag filters in the format {tagName=tagValue}. For more information about filtering by tags, see::
```
avanigupta: Typo - extra colon at the end

sdk/data/azappconfig/client_test.go
```diff
@@ -1242,3 +1242,243 @@ func TestTagsFilteringEndToEnd(t *testing.T) {
...
+		}
+
+		// Should find no settings with non-existent tag
+		require.Empty(t, foundKeys)
+	})
+
+	// Test 7: Filter by tag key that exists with nil values
+	t.Run("TagKeyExistsWithNilValuesFilter", func(t *testing.T) {
+		selector := azappconfig.SettingSelector{
+			TagsFilter: []string{"tag1=\x00"}, // tag1 exists with nil or empty values in some settings
```
avanigupta: What happens if you pass "tag1=\0"? I know it's interchangeable since both are URL encoded to "%00", but I'm curious if we prefer one over the other in Go SDK? We normally publicize "\0" as the filter for null value in our docs and sample code.

RichardChen820: It's language characteristic of GO, "\0" is not a valid escape sequence in GO.
Go expects either:
  - A recognized single-character escape (\n, \t)
  - A three-digit octal escape (\000)
  - A two-digit hex escape (\x00)
  - A Unicode escape (\u0000)

sdk/data/azappconfig/setting_selector.go
RichardChen820: fixed

sdk/data/azappconfig/client_test.go
RichardChen820: Good catch, I should've found it, I think it's AI hallucination.

sdk/data/azappconfig/client_test.go
RichardChen820: added

RichardChen820:
@jhendrixMSFT Does it looks good to you? I'm going to merge it.

sdk/data/azappconfig/version.go
```diff
@@ -8,5 +8,5 @@ package azappconfig
 
 const (
 	moduleName    = "github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig"
-	moduleVersion = "v1.2.1-beta.1"
+	moduleVersion = "v2.0.0"
```
RickWinter, line 11: Why is this a major version rev when the change is additive? Can this feature be added without a breaking change?

RichardChen820: Please refer to the discussion in this thread https://github.com/Azure/azure-sdk-for-go/pull/25194#discussion_r2342843983

RichardChen820: To summarize, it still comes down to maintaining feature completeness and consistency across languages. At first, I was hesitant about making a breaking change just to support nil pointer values—it felt a bit excessive. But since it does cause functionality issue, it’s better to fix it now rather than wait until problems arise and customers start complaining.

Moreover, since the SDK itself is quite fundamental, and we have two other projects (https://github.com/Azure/AppConfiguration-KubernetesProvider & https://github.com/Azure/AppConfiguration-GoProvider)  built on top of it, once tag filter support is added, those projects will also inherit the behavior. If we don’t fix this now, the issue will inevitably be exposed to users later on.

RickWinter:
I've cleared my blocking as the change seems reasonable.

NOTE: Per policy the v1 library must be maintained for 12 months after the date of the v2 release before v1 can be deprecated.

PR #25208, opened 2025-09-10
Upgrade azcore dependencies

chlowell:
This raises the go version because the latest `golang.org/x/net` requires 1.24


chlowell:
Closing this because the latest `golang.org/x` packages require Go 1.25 and we need to address #25217 before using that version in CI. We can continue using the N-1 version of those packages in the meantime (there's no CVE requiring us to upgrade)

PR #25228, opened 2025-09-11
Prep azcore v1.19.1 for release

jhendrixMSFT:


sdk/azcore/CHANGELOG.md
```diff
@@ -13,6 +9,7 @@
 ### Other Changes
 
 * Improved error fall-back for improperly authored long-running operations.
+* Upgraded dependencies to highest version that doesn't require Go 1.24.0 as a minimum version.
```
chlowell: I'd leave this out because I doubt it's interesting to users

jhendrixMSFT: I can. I only mentioned this in case somebody wonders why we didn't upgrade to the very latest versions.

jhendrixMSFT: fixed

PR #25239, opened 2025-09-15
[Release] sdk/resourcemanager/datamigration/armdatamigration/2.0.0  

jliusan:
https://github.com/Azure/sdk-release-request/issues/6501


tadelesh (commented):
@jliusan could you help to check the duplicate schema problem?

jliusan:
> > @jliusan could you help to check the duplicate schema problem?
> 
> `SystemData` defined in [datamigration](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/datamigration/resource-manager/Microsoft.DataMigration/stable/2025-06-30/sqlmigration.json#L3275) is consistent with the `SystemData` defined in [common-type](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v6/types.json#L536), and this problem has existed for several versions

PR #25252, opened 2025-09-16
[queyr/azlogs] remove `goimports`

gracewilcox:
* Removing `goimports` from build.go (generator now automatically runs gofmt as last step in the generation process)
* Regenerating

### DIFF
--- a/sdk/monitor/query/azlogs/build.go
+++ b/sdk/monitor/query/azlogs/build.go
@@ -1,5 +1,5 @@
 //go:generate go run ./testdata/generate/transforms.go
-//go:generate goimports -w .
+//go:generate gofmt -w .
 
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
--- a/sdk/monitor/query/azlogs/build.go
+++ b/sdk/monitor/query/azlogs/build.go
@@ -1,5 +1,4 @@
 //go:generate go run ./testdata/generate/transforms.go
-//go:generate gofmt -w .
 
 // Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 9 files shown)


sdk/monitor/query/azlogs/build.go
```diff
@@ -1,5 +1,4 @@
 //go:generate go run ./testdata/generate/transforms.go
-//go:generate goimports -w .
```
richardpark-msft, line 2: gofmt already being called?

sdk/monitor/query/azlogs/constants.go
```diff
@@ -54,8 +54,8 @@ const (
 	versionsV1 versions = "v1"
 )
 
-// possibleVersionsValues returns the possible values for the versions const type.
-func possibleVersionsValues() []versions {
+// possibleversionsValues returns the possible values for the versions const type.
+func possibleversionsValues() []versions {
```
richardpark-msft, line 58: Just checking, you lost your camel casing for 'versions'

richardpark-msft (approved):
Approving so we can get out of @jsquire's way, but just had some comments on some changes.

sdk/monitor/query/azlogs/constants.go
gracewilcox: This is correct. There was a bug in generator before where this private const was generating as private, except for this func. I had to add a custom transform to the transforms.go file, so that caused the old casing.

Joel was awesome and fixed the bug, and now the generator is generating this correctly.

sdk/monitor/query/azlogs/build.go
gracewilcox: Yes, the generator now automatically runs `gofmt` as the last step in the generation process. Having it here is redundant.

PR #25295, opened 2025-09-22
update docs

gracewilcox:
Updating some of the code generation docs

### DIFF
--- a/documentation/developer_setup.md
+++ b/documentation/developer_setup.md
@@ -35,33 +35,12 @@ All directory structures **MUST** be approved by the Go SDK team (contact azsdkg
 
 For more information, please consult the Azure Go SDK design guidelines on [directory structure][directory_structure].
 
-If your SDK won't be generated from OpenAPI (aka swagger) files, skip to the [next section](#create-module-skeleton).
-
-Once the directory structure has been created, you must decide if your SDK will directly export generated types (commonly referred to as a code generated client).
-The alternative is to make the generated content internal and export hand-written types, possibly along with generated types via type aliasing.
-
-### Code Generated Clients
-
-An SDK that uses code generated clients directly exposes the Autorest-generated code to consumers of the module and is the preferred approach.
-The [azkeys][azkeys_directory] module is an example of a code generated client (CGC).
-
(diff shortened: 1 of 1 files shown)


documentation/developer_setup.md
```diff
@@ -100,33 +79,41 @@ tsp-client bundles up the client generation process into a single program, makin
...
 
 	```yaml
 	# other YAML elided
 	options:
 	  # other emitters elided
 	  "@azure-tools/typespec-go":
-        module: "github.com/Azure/azure-sdk-for-go/sdk/<path to your 'az' package goes here>"
-        module-version: "0.0.1"
-        emitter-output-dir: "{project-root}"
+		module: "github.com/Azure/azure-sdk-for-go/{service-dir}/aznamespaces"
```
RickWinter, line 89: Rather than exact package names, make the {package name}

documentation/developer_setup.md
```diff
@@ -100,33 +79,41 @@ tsp-client bundles up the client generation process into a single program, makin
...
-    //go:generate goimports -w .
+	To generate using a local TypeSpec project,
+	```shell
+	azure-sdk-for-go/sdk/messaging/eventgrid/aznamespaces$ tsp-client update --local-spec-repo <path to TypeSpec project>
 	```
 
-4. In a terminal, `cd` into your package folder and type `go generate`. This should run without error and will create a client, along with code needed to serialize and deserialize models.
+Generated code **must not** be edited, as any edits would be lost on future regeneration of content. To make customizations, update the TypeSpec project's `client.tsp` file then regenerate.
 
 ### Using Autorest
```
RickWinter: Lets add ** DEPRECATED **

gracewilcox: added in [38c9ae3](https://github.com/Azure/azure-sdk-for-go/pull/25295/commits/38c9ae3229895758a7ba0462ef33aac79f573ed0)

documentation/developer_setup.md
gracewilcox: The rest of the documentation explicitly references namespaces as an example package so keeping it as-is for consistency.

PR #24609, opened 2025-05-09
Add sovereign cloud support to aztables

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/24396

PR #25305, opened 2025-09-23
Adds a fallback mechanism to AAD scope override.

aavasthy:
As part of this PR we are providing the ability to override AAD scope value.
If end-user provided a custom scope by overriding the environment variable() then we use that value for AAD authentication. In case of failures with the custom scope there is no retry/fallback.
If no custom scope is provided then we use account scope for AAD authentication. In case of failure with exception [AADSTS500011](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes), there is fallback implemented as part of this PR and then it retries with this scope "https://cosmos.azure.com/.default". This is a temporary change we are implementing till this "https://cosmos.azure.com/.default" is accepted as the scope across all environments and clouds.

### DIFF
--- a/sdk/data/azcosmos/emulator_cosmos_aad_test.go
+++ b/sdk/data/azcosmos/emulator_cosmos_aad_test.go
@@ -162,7 +162,6 @@ func (c *emulatorFailAccountScopeCredential) GetToken(ctx context.Context, tro p
 }
 
 func TestAAD_Fallback_E2E_WithEmulator(t *testing.T) {
-	t.Setenv("EMULATOR", "true")
 	t.Setenv(envCosmosScopeOverride, "") // ensure product code uses endpoint scope + fallback
 
 	em := newEmulatorTests(t)


sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,27 +84,64 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	scope, err := createScopeFromEndpoint(endpointUrl)
-	if err != nil {
-		return nil, err
+
+	var scopes []string
+	if v := strings.TrimSpace(os.Getenv(envCosmosScopeOverride)); v != "" {
```
analogrelay, line 89: I'm a little curious why we're using an environment variable rather than adding an option to the ClientOptions. It seems like that would be more flexible, since it would allow the end-user more control.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,27 +84,64 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
...
 	return &Client{endpoint: endpoint, endpointUrl: endpointUrl, internal: internalClient, gem: gem}, nil
 }
 
+func (c *aadFallbackCredential) GetToken(ctx context.Context, _ policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	tok, err := c.inner.GetToken(ctx, policy.TokenRequestOptions{Scopes: []string{c.primaryScope}})
+	if err == nil || c.fallbackScope == "" {
+		return tok, err
+	}
+	e := strings.ToUpper(err.Error())
+	if strings.Contains(e, "AADSTS500011") {
```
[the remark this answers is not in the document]
analogrelay: Copilot has a point here. If we have a way to extract the error code specifically, I'd prefer we try to use it. Of course, there's no guarantee that what it's suggesting here actually works 😭 . Still, if you could take a few mins to check if there's a clearer way to do this, I'd appreciate it.

But please feel free to timebox that investigation. If this works, I'm happy enough with it :)

sdk/data/azcosmos/emulator_cosmos_aad_test.go
```diff
@@ -142,3 +147,82 @@ func TestAAD(t *testing.T) {
...
+func (c *emulatorFailAccountScopeCredential) GetToken(ctx context.Context, tro policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	if len(tro.Scopes) == 1 && tro.Scopes[0] == c.accountScope {
+		// this exact string is what your client fallback looks for
+		return azcore.AccessToken{}, fmt.Errorf("AADSTS500011: simulated resource not found")
+	}
+	return c.delegate.GetToken(ctx, tro)
+}
+
+func TestAAD_Fallback_E2E_WithEmulator(t *testing.T) {
+	t.Setenv("EMULATOR", "true")
```
analogrelay: We don't set `EMULATOR` in the test itself. Instead, set it locally before running emulator tests on your local machine, and the CI will set it automatically. The env var is intended to signal to the tests that the emulator has been set up and is working. Setting it from within the test means that these tests will try to run even when the emulator isn't running and will likely fail during the unit testing phase of CI.

sdk/data/azcosmos/cosmos_client.go
aavasthy: This is a temporary fix till we have "https://cosmos.azure.com/.default" set as the global scope across all environments. We wanted to give this quick fix to clients to override the scope if account scope does not work with their environment.
Its implemented the same in java, python and .net sdks.

sdk/data/azcosmos/cosmos_client.go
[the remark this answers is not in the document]
aavasthy: I tried testing it locally and the only way to capture AADSTS500011 error code is through the error message.
RawResponse =
*net/http.Response nil
credType =
"InteractiveBrowserCredential"
message =
"AADSTS500011: The resource principal named was not found in the tenant named Microsoft. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You might have sent your authentication request to the wrong tenant. Trace ID: f93e243b-400e-4785-9226-8bec4bed1300 Correlation ID: ae90f0a5-1a60-40f7-88d6-bc00152a9f7d Timestamp: 2025-09-24 18:59:57Z"

sdk/data/azcosmos/emulator_cosmos_aad_test.go
aavasthy: removed this

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -23,7 +24,9 @@ import (
 )
 
 const (
-	apiVersion = "2020-11-05"
+	apiVersion             = "2020-11-05"
+	envCosmosScopeOverride = "AZURE_COSMOS_AAD_SCOPE_OVERRIDE"
+	defaultCosmosScope     = "https://cosmos.azure.com/.default"
```
jhendrixMSFT, line 29: Technically, this is audience + scope.

The correct way to fix this is to add cloud config support. This will allow `azcosmos` to work across sovereign clouds (and stack if Cosmos is supported there).

Please see https://github.com/Azure/azure-sdk-for-go/pull/24609 for how this was wired into `aztables`.

sdk/data/azcosmos/cosmos_client.go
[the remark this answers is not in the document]
analogrelay: Ok! I appreciate you looking into that. We can carry on with what you've got.

sdk/data/azcosmos/cosmos_client.go
analogrelay: Got it, sounds good!

sdk/data/azcosmos/cosmos_client.go
analogrelay: I believe this is how the other Cosmos SDKs are handling this. (@aavasthy can confirm)

It's possible there's larger work to do here around cloud config, which could be done in a separate PR.

analogrelay: Once the longer-term fix (https://github.com/Azure/azure-sdk-for-go/pull/25305#discussion_r2375168403) is ready, I do think we'll want to look at cloud config. Right now, _most_ accounts use the Account's own endpoint as the audience.

jhendrixMSFT: Fixing it like this does mean that `azcosmos` has a non-standard way of configuring sovereign cloud/stack support compared to the other Go SDKs.  It also adds some extra technical debt (i.e. not only do you need to add cloud config support, you'll also need to back out these changes).  And finally, it can be disruptive for customers that build a dependency on this behavior only to have it removed (in fact, we might not be able to remove it).

Maybe a better way to say it is, if we add this env var support, we'll likely be stuck with it.  How will that intersect when adding support for cloud config?  Now we have two ways to do the same thing.  Which one wins?

aavasthy: so we needed to come up with this design to use ["https://cosmos.azure.com/.default"]("https://cosmos.azure.com/.default") - this is guidance we got from Entra team. They are also currently making changes to cloud configs to ensure "https://cosmos.azure.com/.default" scope is used across all environments and clouds. 
We are providing customers who want to use this scope "https://cosmos.azure.com/.default" an option to override and not use account scope. Fabric team has already started using this this model and also all Fabric native Cosmos accounts are supporting only "https://cosmos.azure.com/.default" scope.

As part of this design, customers dont need to make any change if they want to continue using account scope.
This change is already implemented in .net, java and python sdks.

aavasthy: @kirankumarkolli fyi ^^

tvaron3 (approved):
LGTM

sdk/data/azcosmos/cosmos_client.go
jhendrixMSFT: Consistency across Go SDKs is always the high order bit (cross language where appropriate). I have a few more questions.
- What's preventing adding support for [`cloud.Configuration`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.19.1/cloud#Configuration)?
- If we take this, it will likely need to be supported in perpetuity. How will it interact with `cloud.Configuration` once that's added?

aavasthy: Adding the change to cloud config is work in progress with security team but we already have customers like Fabric who are using this scope "https://cosmos.azure.com/.default". So we are providing this option to override scope value if the app doesn't work with account scope.
Even once this change is being used both account scope and "https://cosmos.azure.com/.default" will be accepted audience scope for AAD token authentication.

analogrelay: I think @jhendrixMSFT 's point is certainly important here. I do want us to strive for solid consistency with other Go SDKs. I don't have the background on the scope of the cloud config work, but I'm open to prioritizing it

It's worth noting that the Go SDK is currently in beta for the next few releases because of the Client Engine integration work, which isn't ready for GA yet. @aavasthy, if there's a need to get _some_ kind of fix for Fabric-native in ASAP, could we land this and then migrate over to cloud config (and remove the env var while still in beta) before the next GA (I don't have a timeframe for that, but I'd like to be back in GA by around the end of the calendar year; that's just my preference though)?

analogrelay: > If we take this, it will likely need to be supported in perpetuity. How will it interact with cloud.Configuration once that's added?

I think that's fairly clear to me. When we use `cloud.Configuration`, that config will drive the scope we select by default,, and we could safely remove the fallback logic. The env var to override could still function and would replace whatever `cloud.Configuration` provides as the audience+scope. The env var would remain supported as the highest-level override, but wouldn't be necessary for most (if not all) users.

aavasthy:
Tracking the AAD scope change as part of this PR now https://github.com/Azure/azure-sdk-for-go/pull/25366

PR #25313, opened 2025-09-24
Add owner for App Configuration

RichardChen820:
Add owner for App Configuration

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -31,7 +31,7 @@
 # ServiceOwners:                                                   @miaojiang
 
 # PRLabel: %App Configuration
-/sdk/data/azappconfig/                                             @jhendrixMSFT @rickwinter @RichardChen820 @linglingye001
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -31,7 +31,7 @@
 # ServiceOwners:                                                   @miaojiang
 
 # PRLabel: %App Configuration
-/sdk/data/azappconfig/                                             @jhendrixMSFT @rickwinter
+/sdk/data/azappconfig/                                             @jhendrixMSFT @rickwinter @RichardChen820 @linglingye001
```
jhendrixMSFT: Can you please change the ordering here. The convention is that the left-most alias is the main point-of-contact, and the rest are backups. e.g. `[primary owner, backup1, backup2, ...backupN]`. You can put me last and remove Rick.

RickWinter (approved):

PR #25314, opened 2025-09-24
[azappconfig] Fix nil sync token cause nil ref panic bug

RichardChen820:
This pull request refactors how `SyncToken` values are handled throughout the `sdk/data/azappconfig/client.go` . This change improves the safety and robustness of the code by preventing possible nil pointer dereference panics.

Refactoring for safer `SyncToken` handling:

* Replaced all direct conversions of `*SyncToken` pointers to `SyncToken` with calls to the new `safeSyncToken` helper function in all response constructors (e.g., `AddSettingResponse`, `DeleteSettingResponse`, `GetSettingResponse`, `SetReadOnlyResponse`, `SetSettingResponse`, `ListRevisionsPageResponse`, `ListSettingsPageResponse`, `ListSnapshotsResponse`, `ListSettingsForSnapshotResponse`, `GetSnapshotResponse`, `updateSnapshotStatus` response). [[1]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL112-R112) [[2]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL135-R135) [[3]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL158-R158) [[4]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL181-R181) [[5]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL190-R190) [[6]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL222-R222) [[7]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL247-R247) [[8]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL277-R277) [[9]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL332-R332) [[10]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL378-R378) [[11]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL481-R481) [[12]](diffhunk://#diff-0f9bc02dede85b6146b31d5db9ec5f7eac9fff9471cfc079486b8838667ee9afL576-R589)

### DIFF
--- a/sdk/data/azappconfig/client.go
+++ b/sdk/data/azappconfig/client.go
@@ -109,7 +109,7 @@ func (c *Client) AddSetting(ctx context.Context, key string, value *string, opti
 
 	return AddSettingResponse{
 		Setting:   settingFromGenerated(resp.KeyValue),
-		SyncToken: SyncToken(*resp.SyncToken),
+		SyncToken: safeSyncToken(resp.SyncToken),
 	}, nil
 }
 
@@ -132,7 +132,7 @@ func (c *Client) DeleteSetting(ctx context.Context, key string, options *DeleteS
 
 	return DeleteSettingResponse{
 		Setting:   settingFromGenerated(resp.KeyValue),
-		SyncToken: SyncToken(*resp.SyncToken),
+		SyncToken: safeSyncToken(resp.SyncToken),
 	}, nil
 }
 
@@ -155,7 +155,7 @@ func (c *Client) GetSetting(ctx context.Context, key string, options *GetSetting
 
 	return GetSettingResponse{
 		Setting:      settingFromGenerated(resp.KeyValue),
-		SyncToken:    SyncToken(*resp.SyncToken),
+		SyncToken:    safeSyncToken(resp.SyncToken),
 		LastModified: resp.KeyValue.LastModified,
 	}, nil
 }
@@ -178,7 +178,7 @@ func (c *Client) SetReadOnly(ctx context.Context, key string, isReadOnly bool, o
 		if err == nil {
 			return SetReadOnlyResponse{
 				Setting:   settingFromGenerated(resp.KeyValue),
-				SyncToken: SyncToken(*resp.SyncToken),
+				SyncToken: safeSyncToken(resp.SyncToken),
 			}, nil
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


RichardChen820:
@jhendrixMSFT @RickWinter syncToken is not a required field in the response. Missing syncToken causes nil pointer panic

sdk/data/azappconfig/client.go
```diff
@@ -109,7 +109,7 @@ func (c *Client) AddSetting(ctx context.Context, key string, value *string, opti
 
 	return AddSettingResponse{
 		Setting:   settingFromGenerated(resp.KeyValue),
-		SyncToken: SyncToken(*resp.SyncToken),
+		SyncToken: safeSyncToken(resp.SyncToken),
```
jhendrixMSFT, line 112: When we initially added this, we were told by the AppConfig team that all responses include a sync token (this is why we designed it as such). Is that not the case?

In addition, I worry that this might cause other bugs as `ParseSyncToken()` will return an error if the `SyncToken` value is the empty string which is exactly what's happening here.

Can we please confirm the behavior of the service? I also noticed this was opened in response to the emulator. Is it the case that the emulator is omitting the sync token in some cases when it shouldn't be (to say it another way, we shouldn't change the SDK to work around a deficiency in the emulator)?

RichardChen820: I agree, let me confirm the data plane protocol first.

RichardChen820: I confirmed, the protocol is that success responses always have the syncToken in, so Go SDK does need this change, dropping this PR.

Issue #25324, opened 2025-09-25
[azcosmos]: add support for cloud config


jhendrixMSFT:
This allows setting the correct audience for sovereign clouds and Azure stack.

PR #25366, opened 2025-10-07
Adds a fallback mechanism to AAD scope override.

aavasthy:
As part of this PR we are providing the ability to add a custom AAD audience value by setting it in client options.
If end-user provided a custom audience then we use that value for AAD authentication. In case of failures with the custom scope there is no retry/fallback.
If no custom scope is provided then we use account scope which is create using hostname for AAD authentication (This is the existing setup for scope creation for AAD auth). In case of failures with the custom scope there is no retry/fallback.

### DIFF
--- a/sdk/data/azcosmos/cloud_config.go
+++ b/sdk/data/azcosmos/cloud_config.go
@@ -6,4 +6,4 @@ package azcosmos
 import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 
 // ServiceName is the [cloud.ServiceName] for Azure Cosmos DB, used to identify the respective [cloud.ServiceConfiguration].
-const ServiceName cloud.ServiceName = "data/cosmosdb"
+const ServiceName cloud.ServiceName = "data/azcosmos"


sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,10 +76,28 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
 	if err != nil {
 		return nil, err
 	}
-	scope, err := createScopeFromEndpoint(endpointUrl)
-	if err != nil {
-		return nil, err
+
+	var scope []string
+	builtInAudience := cloud.AzurePublic.Services["cosmosDB"].Audience
```
jhendrixMSFT: Note that `Audience` does _not_ contain a scope. The scope must be added to the audience like [this example](https://github.com/jhendrixMSFT/azure-sdk-for-go/blob/azcosmos-cloud-config/sdk/data/azcosmos/cosmos_client.go#L89).

jhendrixMSFT: Also, you must export a `const` for the `cloud.ServiceName` value like [this](https://github.com/jhendrixMSFT/azure-sdk-for-go/blob/azcosmos-cloud-config/sdk/data/azcosmos/cloud_config.go#L9).

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,10 +76,28 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
...
+
+	if o != nil && o.ClientOptions.Cloud.Services != nil {
+		if svcCfg, ok := o.ClientOptions.Cloud.Services[cloud.ServiceName("cosmosDB")]; ok && svcCfg.Audience != "" {
+			if svcCfg.Audience == builtInAudience {
+				scope = []string{builtInAudience}
+				log.Write(azlog.EventRequest, fmt.Sprintf("Using built-in Cosmos DB audience: %s", builtInAudience))
+			} else {
+				log.Write(azlog.EventRequest, fmt.Sprintf("WARNING: Ignoring custom audience %q, using account scope instead", svcCfg.Audience))
```
jhendrixMSFT: It's unclear why the custom value is being ignored. Isn't the idea to use the specified value?

sdk/data/azcosmos/cosmos_client.go
aavasthy: updated the code and now the end user provides the audience value and the SDK code create the scope from the given audience.

sdk/data/azcosmos/cosmos_client.go
aavasthy: Updated the logic and now we use the audience value provided by the end user as part of client options. If end user did not set this value then used account based scope.

sdk/data/azcosmos/cloud_config.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+// ServiceName is the [cloud.ServiceName] for Azure Cosmos DB, used to identify the respective [cloud.ServiceConfiguration].
+const ServiceName cloud.ServiceName = "cosmosDB"
```
analogrelay: Since this is going to be case sensitive, let's just have it be all lowercase.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,10 +75,29 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
...
+	var scope []string
+
+	if o != nil && o.ClientOptions.Cloud.Services != nil {
+		if svcCfg, ok := o.ClientOptions.Cloud.Services[ServiceName]; ok && svcCfg.Audience != "" {
+			audience := svcCfg.Audience
+			if !strings.HasSuffix(audience, "/.default") {
+				audience = strings.TrimSuffix(audience, "/") + "/.default"
+			}
+			scope = []string{audience}
+			log.Write(azlog.EventRequest, fmt.Sprintf("Using audience from client options: %s", audience))
```
analogrelay: This is actually the "scope" now, since you've appended `.default`. We should be accurate with the log message

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,10 +75,29 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
...
-	scope, err := createScopeFromEndpoint(endpointUrl)
-	if err != nil {
-		return nil, err
+
+	var scope []string
+
+	if o != nil && o.ClientOptions.Cloud.Services != nil {
+		if svcCfg, ok := o.ClientOptions.Cloud.Services[ServiceName]; ok && svcCfg.Audience != "" {
+			audience := svcCfg.Audience
+			if !strings.HasSuffix(audience, "/.default") {
```
analogrelay: I don't think we should support the scenario where the user put the scope in where the audience was expected. I'd rather not check this and let it produce an error with the server if the combined audience and scope end up wrong

sdk/data/azcosmos/cosmos_client_test.go
```diff
@@ -729,6 +732,81 @@ func TestSpanResponseAttributes(t *testing.T) {
 	}
 }
 
+func TestAADScope_UsesAudienceFromClientOptions(t *testing.T) {
+	srv, close := mock.NewTLSServer()
+	defer close()
+	srv.SetResponse(mock.WithStatusCode(200))
+
+	endpoint := srv.URL()
+	audience := "https://custom.audience.example.com"
```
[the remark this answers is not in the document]
analogrelay: Unnecessary. This is a unit test

sdk/data/azcosmos/cosmos_client_test.go
```diff
@@ -729,6 +732,81 @@ func TestSpanResponseAttributes(t *testing.T) {
...
+	expectedScope := audience + "/.default"
+
+	clientOptions := &ClientOptions{
+		ClientOptions: policy.ClientOptions{
+			Cloud: cloud.Configuration{
+				Services: map[cloud.ServiceName]cloud.ServiceConfiguration{
+					"cosmosDB": {Audience: audience},
```
analogrelay: We should use the constant for the key here

sdk/data/azcosmos/cloud_config.go
aavasthy: updated and using lower case

sdk/data/azcosmos/cosmos_client.go
aavasthy: updated the logic

sdk/data/azcosmos/cosmos_client_test.go
aavasthy: updated and using key now

sdk/data/azcosmos/cosmos_client.go
aavasthy: updated log

sdk/data/azcosmos/cloud_config.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+// ServiceName is the [cloud.ServiceName] for Azure Cosmos DB, used to identify the respective [cloud.ServiceConfiguration].
+const ServiceName cloud.ServiceName = "cosmosdb"
```
jhendrixMSFT: By convention, we use the module identity sans the `github.com/Azure/azure-sdk-for-go/sdk/` prefix. So, the value should be `data/azcosmos`.

sdk/data/azcosmos/cosmos_client.go
```diff
@@ -75,10 +75,26 @@ func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (
...
-	if err != nil {
-		return nil, err
+
+	var scope []string
+
+	if o != nil && o.ClientOptions.Cloud.Services != nil {
+		if svcCfg, ok := o.ClientOptions.Cloud.Services[ServiceName]; ok && svcCfg.Audience != "" {
+			audience := svcCfg.Audience
+			scope = []string{audience + "/.default"}
+			log.Write(azlog.EventRequest, fmt.Sprintf("Using custom scope  for authentication: %s", scope[0]))
```
jhendrixMSFT: nit, extra space between `scope` and `for`.

sdk/data/azcosmos/cloud_config.go
analogrelay: Ok, that works for me! As long as this doesn't have to match anything cross-language I'm good with that. We don't have the `data` prefix on our package/namespace names in every language.

sdk/data/azcosmos/cloud_config.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azcosmos
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+
+// ServiceName is the [cloud.ServiceName] for Azure Cosmos DB, used to identify the respective [cloud.ServiceConfiguration].
+const ServiceName cloud.ServiceName = "data/cosmosdb"
```
jhendrixMSFT: `"data/azcosmos"`

sdk/data/azcosmos/cloud_config.go
jhendrixMSFT: Correct, no cross-language requirements. It's purely to ensure unique entries in the cloud config map, and to a lesser extent, make it clear from where entries originate.

sdk/data/azcosmos/cosmos_client.go
aavasthy: updated

sdk/data/azcosmos/cloud_config.go
aavasthy: updated

sdk/data/azcosmos/cloud_config.go
aavasthy: updated

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/25324

PR #25377, opened 2025-10-07
gocov

gracewilcox:


eng/tools/internal/coverage/coverage_test.go
```diff
@@ -29,15 +28,15 @@ func Test_ParseCoverageMaximum(t *testing.T) {
 }
 
 func Test_ParseCoverageFloat(t *testing.T) {
-	coverageLine := "<coverage line-rate=\"0.23893805\" branch-rate=\"0\" lines-covered=\"216\" lines-valid=\"904\" branches-covered=\"0\" branches-valid=\"0\" complexity=\"0\" version=\"\" timestamp=\"1633570973824\">"
+	coverageLine := "total:                                                                                          (statements)                                      80.5%"
```
benbp, line 31: This seems like it will be different based on the test user's terminal width?

eng/scripts/test.ps1
```diff
@@ -33,14 +33,8 @@ if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simpl
...
 
-    gocov convert ./coverage.txt > ./coverage.json
-
-    # gocov converts rely on standard input
-    Get-Content ./coverage.json | gocov-xml > ./coverage.xml
-    Get-Content ./coverage.json | gocov-html > ./coverage.html
-
-    Move-Item -Force ./coverage.xml $repoRoot
-    Move-Item -Force ./coverage.html $repoRoot
+    go tool cover -func coverage.txt | Out-File -FilePath coveragefunc.txt
```
benbp, line 36:
```suggestion
    go tool cover -func coverage.txt | Out-File -FilePath coveragefunc.txt
    if ($LASTEXITCODE) { exit $LASTEXITCODE }
```

eng/tools/internal/coverage/coverage_test.go
gracewilcox: Yes, spacing can be different, so I made sure the regex in the coverage.go file could parse different lengths: 

```go
re := regexp.MustCompile(`total:.*?(\d+\.\d+)%`)
```

eng/scripts/test.ps1
gracewilcox: Is this different from $GOTESTEXITCODE? Notice that's the exit code referred to throughout the script

gracewilcox: fixed in 3607178

PR #25414, opened 2025-10-13
Go Workloads for Testing

tvaron3:
## Description
This is adding a way to startup a workload to use for testing in dr drills. The configurations can be set through environment variables. Initially, the script will create a database and container, and an item for the number of logical partitions specified in the configuration files. After this initial setup, concurrent operations should happen on the precreated items until the process is killed. Idea is for this to be the bare minimum and later we can add more things based on customer usage in go like aad, other operations, etc

### DIFF
--- a/sdk/data/azcosmos/workloads/workload_utils.go
+++ b/sdk/data/azcosmos/workloads/workload_utils.go
@@ -185,9 +185,9 @@ func vectorSearchQueries(ctx context.Context, container *azcosmos.ContainerClien
 
 		queryStr := ""
 		if pkField == "id" {
-			queryStr = "SELECT TOP 10 c.id FROM c WHERE c.id == @pkValue ORDER BY VectorDistance(c.embedding, @embedding)"
+			queryStr = "SELECT TOP 10 c.id FROM c WHERE c.id = @pkValue ORDER BY VectorDistance(c.embedding, @embedding)"
 		} else {
-			queryStr = "SELECT TOP 10 c.id FROM c WHERE c.pk == @pkValue ORDER BY VectorDistance(c.embedding, @embedding)"
+			queryStr = "SELECT TOP 10 c.id FROM c WHERE c.pk = @pkValue ORDER BY VectorDistance(c.embedding, @embedding)"
 		}
 		pkVal := GetPKValue(pkField, rng.Intn(count)+1)
 


sdk/data/azcosmos/workloads/workload_config.go
```diff
@@ -0,0 +1,102 @@
...
+		if v == "" {
+			return "", fmt.Errorf("missing env var %s", name)
+		}
+		return v, nil
+	}
+
+	var cfg workloadConfig
+	var err error
+
+	if cfg.Endpoint, err = get("COSMOS_URI"); err != nil {
```
simorenoh, line 43: how come we load everything in as env variables as opposed to the load in Python where we directly type in the values into string variables?

sdk/data/azcosmos/workloads/workload_utils.go
```diff
@@ -0,0 +1,172 @@
...
+		"type":      "testItem",
+		"createdAt": time.Now().UTC().Format(time.RFC3339Nano),
+		"seq":       i,
+		"value":     rand.Int63(), // pseudo-random payload
+	}
+}
+
+// runConcurrent executes count indexed jobs across at most workers goroutines.
+// jf should be idempotent per index; it receives a per-worker RNG (not safe to share across workers).
+func runConcurrent(ctx context.Context, count, workers int, jf func(ctx context.Context, index int, rng *rand.Rand) error) error {
```
simorenoh: nit: renaming jf to jobFunction or at least something more descriptive

sdk/data/azcosmos/workloads/r_w_q_workload.go
```diff
@@ -0,0 +1,43 @@
...
+		return fmt.Errorf("ensure container: %w", err)
+	}
+
+	var count = cfg.LogicalPartitions
+
+	log.Printf("Starting %d concurrent read/write/queries ...", count)
+
+	for {
+		if err := randomReadWriteQueries(ctx, container, count, cfg.PartitionKeyFieldName); err != nil {
+			return fmt.Errorf("read/write queries failed: %w", err)
```
simorenoh: does this return statement not break the loop execution?

sdk/data/azcosmos/workloads/workload_config.go
tvaron3: Yeah I was thinking this would be better because in python changing that file will frequently cause merge conflicts when adding new configs. Easy for us to fix but other people running the workloads could get confused.

sdk/data/azcosmos/workloads/r_w_q_workload.go
tvaron3: Yes you are right fixed it.

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -5,6 +5,7 @@ package azcosmos
 
 import (
 	"fmt"
+	"log"
```
analogrelay: This is fine for ad-hoc testing, but when we actually want to add logging, there's a facility built in to the Core SDK for that in the `"github.com/Azure/azure-sdk-for-go/sdk/internal/log"` package. I used it in a recent commit to the query engine code path: https://github.com/Azure/azure-sdk-for-go/commit/6d8a42734f87cd925e84d39cc942868315445037

sdk/data/azcosmos/workloads/workload_utils.go
```diff
@@ -0,0 +1,200 @@
...
+	wg.Wait()
+	close(errs)
+
+	var firstErr error
+	for e := range errs {
+		if firstErr == nil {
+			firstErr = e
+		}
+	}
+	return firstErr
```
analogrelay: You could use [`errors.Join`](https://pkg.go.dev/errors#example-Join) to make them into a single error if we want to be able to see all errors.

sdk/data/azcosmos/workloads/workload_config.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Comma-separated preferred locations (optional)
+	if pl := os.Getenv("PREFERRED_LOCATIONS"); pl != "" {
+		// Simple split on comma; whitespace trimming omitted for brevity
+		cfg.PreferredLocations = splitAndTrim(pl, ',')
+	}
+	return cfg, nil
+}
+
+func splitAndTrim(s string, sep rune) []string {
```
analogrelay: [`strings.Split`](https://pkg.go.dev/strings#Split) should be able to do this for us. It looks like this code is just splitting, not splitting and trimming. And if we do need splitting **and** trimming, we can just iterate through the slice created by `Split` and call [`strings.Trim`](https://pkg.go.dev/strings#example-Trim)

sdk/data/azcosmos/workloads/setup_env.sh
```diff
@@ -0,0 +1,14 @@
...
+# setup_env.sh - Automates Azure Cosmos SDK scale testing environment setup
+# Usage: bash setup_env.sh
+
+set -e
+
+# 1. System update and install dependencies
+echo "[Step 1] System update and install dependencies: started."
+sudo apt-get update
+sudo apt  install golang-go
+sudo apt install neovim
```
analogrelay: nit: `apt` and `apt-get` are actually separate tools (though for many commands they have essentially the same behavior. We should just use one OR the other and avoid mixing them.

sdk/data/azcosmos/workloads/README.md
analogrelay, line 1: nit: Go ahead and name this `README.md`. It makes sense as a "README" doc for these test workloads, and that means it'll appear in GitHub when someone is looking at this folder.

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -90,7 +89,7 @@ func (lc *locationCache) update(writeLocations []accountRegion, readLocations []
 	lc.refreshStaleEndpoints()
 	if readLocations != nil {
 		availReadEndpointsByLocation, availReadLocations, err := getEndpointsByLocation(readLocations)
-		log.Printf("Available read endpoints by location: %v", availReadEndpointsByLocation)
+		// log.Printf("Available read endpoints by location: %v", availReadEndpointsByLocation)
```
analogrelay, line 92: Let's either remove these altogether or switch them to use the core logging facility. (Basically, I just don't like committing commented-out code ;))

analogrelay (commented):
One small remaining thing about the commented-out log statements, but otherwise LGTM.

sdk/data/azcosmos/cosmos_location_cache.go
tvaron3: Yes I will remove  these. :)

sdk/data/azcosmos/workloads/workload
analogrelay, line 1: Is this a compiled application?

sdk/data/azcosmos/workloads/setup_env.sh
tvaron3: fixed

sdk/data/azcosmos/workloads/workload_config.go
tvaron3: fixed

sdk/data/azcosmos/workloads/workload_utils.go
tvaron3: switched to use errors.Join

sdk/data/azcosmos/workloads/workload
tvaron3: Currently was not thinking of adding it as an executable, but open to suggestions. :)

PR #25432, opened 2025-10-15
[Storage] returning err in appendblob.Client.AppendBlock 

Leinadium:
This PR fixes the return values of `AppendBlock()` of `appendblob.Client`.

`AppendBlock()` is supposed to return an error if

- `shared.ValidateSeekableStreamAt0AndGetCount` returns `err`;
- `o.TransactionalValidation.Apply` returns `err`.
 
But in both cases, a `nil` value is returned, instead of `err`:

```go
count, err := shared.ValidateSeekableStreamAt0AndGetCount(body)
if err != nil {
	return AppendBlockResponse{}, nil
}

// ...
body, err = o.TransactionalValidation.Apply(body, appendOptions)
if err != nil {
	return AppendBlockResponse{}, nil
}
```

I could not find any motives or explanations on why this behaviour exists, so I understand this is a bug.

### DIFF
--- a/sdk/storage/azblob/appendblob/client.go
+++ b/sdk/storage/azblob/appendblob/client.go
@@ -168,7 +168,7 @@ func (ab *Client) Create(ctx context.Context, o *CreateOptions) (CreateResponse,
 func (ab *Client) AppendBlock(ctx context.Context, body io.ReadSeekCloser, o *AppendBlockOptions) (AppendBlockResponse, error) {
 	count, err := shared.ValidateSeekableStreamAt0AndGetCount(body)
 	if err != nil {
-		return AppendBlockResponse{}, nil
+		return AppendBlockResponse{}, err
 	}
 
 	appendOptions, appendPositionAccessConditions, cpkInfo, cpkScope, modifiedAccessConditions, leaseAccessConditions := o.format()
@@ -176,7 +176,7 @@ func (ab *Client) AppendBlock(ctx context.Context, body io.ReadSeekCloser, o *Ap
 	if o != nil && o.TransactionalValidation != nil {
 		body, err = o.TransactionalValidation.Apply(body, appendOptions)
 		if err != nil {
-			return AppendBlockResponse{}, nil
+			return AppendBlockResponse{}, err
 		}
 	}
 


tanyasethi-msft:
Looks good. Needs a test for the same before merge

Leinadium:
I also noticed that the documentation states "This method panics if the stream is not at position 0.".

But the current implementation:

- calls  `..., err := shared.ValidateSeekableStreamAt0AndGetCount(body)`
- which calls `err := validateSeekableStreamAt0(body)`
- which checks if `pos := body.Seek(0, io.SeekCurrent) != 0`
  - and if is not 0, returns `errors.New("body stream must be set to position 0")`

(Actually the current implementation will return an empty response and a nil error if position is not 0)

tanyasethi-msft:
Working as expected.

Leinadium:
I don't think this is working as expected, because:

- even though an error is generated internally, it is being supressed and an empty response with nil error is being returned
- the function documentation states it should actually panic if seek is not 0

Here is an reproductible example:

```go
// settings for azure
connStr := "..."
container := "my-container"
filename := "foo.txt"

client, err := azblob.NewClientFromConnectionString(connStr, nil)
if err != nil {
	panic(err)
}
ab := client.ServiceClient().NewContainerClient(container).NewAppendBlobClient(filename)

// create file
file, err := os.CreateTemp("", "test")
if err != nil {
	panic(err)
}

// write some bytes
_, err = file.Write([]byte{1, 2, 3, 4, 5, 6, 7, 8})
if err != nil {
	panic(err)
}

// seek
_, err = file.Seek(2, io.SeekStart)
if err != nil {
	panic(err)
}

// according to docs, AppendBlock should panic if seek is not 0
// internally, it actually returns an error, which the function
// actually supress and returns empty response and nil error
_, err = ab.AppendBlock(context.Background(), file, nil)
fmt.Printf("returned %v\n", err)
```

jhendrixMSFT:
@tanyasethi-msft can you please clarify how this is working as expected? The fact that we eat the error is not intuitive. How does the caller know that the method call didn't do anything?

Leinadium:
@tanyasethi-msft could you please provide any context explaning why this behaviour is actually expected?

PR #25433, opened 2025-10-15
[Storage][Content Validation] Structured message reader and writer

tanyasethi-msft:


sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+func NewStructuredMessageResponseReader(body io.ReadCloser, structuredBodyType *string) (io.ReadCloser, error) {
+	if structuredBodyType == nil || *structuredBodyType == "" {
+		// No structured message, return body as-is
+		return body, nil
+	}
+
+	// Check if it's a structured message format we support
+	// Format: "XSM/1.0;CRC64" or "XSM/1.0;properties=crc64"
+	bodyType := strings.ToUpper(*structuredBodyType)
+	if !strings.HasPrefix(bodyType, "XSM/1.0") {
```
jhendrixMSFT: This check seems a bit lenient.  What's the possibility that another format is introduced that would cause this condition to be `true`?

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+func (r *StructuredMessageResponseReader) Read(p []byte) (int, error) {
+	if r.closed {
+		return 0, io.EOF
+	}
+
+	// Read header on first read
+	if !r.headerRead {
+		_, err := r.reader.ReadHeader()
+		if err != nil {
+			if err == io.EOF {
```
jhendrixMSFT: We should be using `errors.Is` instead of direct comparison.  You can also simplify some of the nesting.
```go
if errors.Is(err, io.EOF) {
    return 0, err
} else if err != nil {
    ...
}
```

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+				return 0, io.EOF
+			}
+			// Map structured message errors to blob errors
+			if errors.Is(err, structuredmsg.ErrInvalidVersion) {
+				return 0, fmt.Errorf("invalid structured message version: %w", err)
```
jhendrixMSFT, line 99: `%w` means that `err` is exposed via `errors.Is` and/or `errors.As`.  This is usually a mistake (depends on the context).  Is `structuredmsg.ErrInvalidVersion` publicly exported?  If so, then I think it would be clearer to drop the `fmt.Errorf()` and return structuredmsg.ErrInvalidVersion.

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+			return 0, err
+		}
+		r.headerRead = true
+	}
+
+	// Read data from current segment, or get next segment
+	totalRead := 0
+	for totalRead < len(p) {
+		// If we've exhausted current segment, read next one
+		if r.currentSegmentData == nil || r.currentSegmentIndex >= len(r.currentSegmentData) {
```
jhendrixMSFT, line 113: Can this be simplified by removing `r.currentSegmentData == nil`? It's ok to call `len()` on a `nil` slice (has a length of zero).

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+			segmentData, err := r.reader.ReadSegment()
+			if err != nil {
+				if err == io.EOF {
```
jhendrixMSFT, line 116: Same error comment as above (and the following lines).

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
jhendrixMSFT: On further investigation, I don't think so.

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+						return closeErr
+					}
+					if trailerErr == structuredmsg.ErrCRC64Mismatch {
+						return fmt.Errorf("CRC64 checksum mismatch in structured message trailer: %w", trailerErr)
+					}
+					return trailerErr
+				}
+				break
+			}
+			if err != nil {
```
jhendrixMSFT, line 202: If this is an `else if` you can remove the `break` statement on line 174 (took me a while to figure out what's going on here).

sdk/storage/azblob/internal/structuredmsg/example_test.go
```diff
@@ -0,0 +1,184 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package structuredmsg_test
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/internal/structuredmsg"
```
jhendrixMSFT, line 14: This import path won't be available outside the SDK repo.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
jhendrixMSFT, line 1: Is this package supposed to be publicly available for customers?  The presence of `example_test.go` makes me think it is.  If so, this shouldn't be under `internal`.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
```diff
@@ -0,0 +1,369 @@
...
+	ErrInvalidSegmentCount  = errors.New("invalid segment count")
+	ErrSegmentTooLarge      = errors.New("segment too large")
+	ErrCRC64Mismatch        = errors.New("CRC64 checksum mismatch")
+	ErrUnexpectedEOF        = errors.New("unexpected end of file")
+	ErrInvalidSegmentNumber = errors.New("invalid segment number")
+	ErrInvalidDataLength    = errors.New("invalid data length")
+)
+
+// StructuredMessageWriter encodes data into the binary message format
+type StructuredMessageWriter struct {
```
jhendrixMSFT, line 74: Since we're already in the `structuredmsg` package, we can drop the `StructuredMessage` prefix from these types.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
```diff
@@ -0,0 +1,369 @@
...
+}
+
+// Close marks the reader as closed
+func (r *StructuredMessageReader) Close() error {
+	r.closed = true
+	return nil
+}
+
+// IsClosed returns true if the reader is closed
+func (r *StructuredMessageReader) IsClosed() bool {
```
jhendrixMSFT, line 411: Is there a customer use-case for this?  I didn't see it in the examples.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
```diff
@@ -0,0 +1,369 @@
...
+	// Read segment data
+	data := make([]byte, dataLength)
+	n, err = io.ReadFull(r.reader, data)
+	if err != nil {
+		if err == io.EOF {
+			return nil, ErrUnexpectedEOF
+		}
+		return nil, err
+	}
+	if n != int(dataLength) {
```
jhendrixMSFT, line 304: I think this check can be removed.  [io.ReadFull](https://pkg.go.dev/io#ReadFull) will return an error if fewer than `dataLength` bytes were read.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
```diff
@@ -0,0 +1,369 @@
...
+	return expectedCRC64, nil
+}
+
+// GetTotalCRC64 returns the cumulative CRC64 of all segments read
+func (r *StructuredMessageReader) GetTotalCRC64() uint64 {
+	return r.totalCRC64
+}
+
+// Close marks the reader as closed
+func (r *StructuredMessageReader) Close() error {
```
jhendrixMSFT, line 405: Is this really required?  We should only add an `io.Closer` when `Close()` will release underlying resources.

sdk/storage/azblob/internal/structuredmsg/structured_message.go
```diff
@@ -0,0 +1,369 @@
...
+		reader:     reader,
+		crc64Table: shared.CRC64Table,
+		segmentNum: 0,
+		totalCRC64: 0,
+		closed:     false,
+	}
+}
+
+// ReadHeader reads and validates the message header
+func (r *StructuredMessageReader) ReadHeader() (*MessageHeader, error) {
```
jhendrixMSFT, line 216: Do customers ever need to call this (and `ReadTrailer`), or can these not be exported?

sdk/storage/azblob/internal/structuredmsg/structured_message.go
jhendrixMSFT: Ok this does look to be internal only.

sdk/storage/azblob/internal/structuredmsg/example_test.go
jhendrixMSFT, line 1: I think this file can be removed.  Examples are for public surface area, and since this is internal content, no examples will be rendered.

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+	for totalRead < len(p) {
+		// If we've exhausted current segment, read next one
+		if r.currentSegmentData == nil || r.currentSegmentIndex >= len(r.currentSegmentData) {
+			segmentData, err := r.reader.ReadSegment()
+			if err != nil {
+				if err == io.EOF {
+					// No more segments, validate trailer
+					_, trailerErr := r.reader.ReadTrailer()
+					if trailerErr != nil {
+						if trailerErr == structuredmsg.ErrCRC64Mismatch {
```
jhendrixMSFT, line 120: I think this (and other places) can be simplified.  If `ReadTrailer()` were to return the complete error message (e.g. "CRC64 checksum mismatch in structured message trailer"), there would be no need to export this error type and perform these comparisons.  You can simply return the error from `ReadTrailer()`.

sdk/storage/azblob/blob/constants.go
```diff
@@ -200,6 +200,12 @@ func TransferValidationTypeComputeCRC64() TransferValidationType {
 	return exported.TransferValidationTypeComputeCRC64()
 }
 
+// TransferValidationTypeStructuredMessage is a TransferValidationType that wraps data in structured message format
+// for large uploads (>4MB). For smaller uploads (<4MB), it uses simple CRC64 header validation.
+func TransferValidationTypeStructuredMessage() TransferValidationType {
```
jhendrixMSFT, line 205: Would be good to add some example usage (I assume this works for uploads and downloads?).

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil
+	}
+	r.closed = true
+
+	// If we haven't finished reading, try to read remaining segments and validate trailer
+	// This ensures we catch any CRC64 mismatches even if the user doesn't read all data
+	if r.headerRead {
+		// Read any remaining segments
+		for {
+			_, err := r.reader.ReadSegment()
```
jhendrixMSFT, line 185: This is going to create a lot of garbage.  If this is important to do on `Close()` then we need to change the design so that draining the reader doesn't allocate a bunch of `[]byte` that get discarded.

sdk/storage/azblob/internal/exported/structured_message_response_reader.go
```diff
@@ -0,0 +1,191 @@
...
+				return totalRead, err
+			}
+			r.currentSegmentData = segmentData
+			r.currentSegmentIndex = 0
+		}
+
+		// Copy data from current segment to output buffer
+		available := len(r.currentSegmentData) - r.currentSegmentIndex
+		needed := len(p) - totalRead
+		copySize := available
```
jhendrixMSFT, line 153:
```go
copySize := min(available, needed)
```

sdk/storage/azblob/blob/constants.go
tanyasethi-msft: ample examples will be added in the upcoming iteration, before this feature is fully merged.

tanyasethi-msft:
Will open another PR with small changes in each so that it is easier to review and do changes.

PR #25447, opened 2025-10-16
Exit test.ps1 specifically when coverage checks fail

benbp:
The current way the scripts are written do not actually bubble up errors from `go run coverage` but previous behavior that failed on stderr has been changed, exposing this bug.

### DIFF
--- a/eng/scripts/test.ps1
+++ b/eng/scripts/test.ps1
@@ -49,6 +49,10 @@ if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simpl
         -config $repoRoot/eng/config.json `
         -serviceDirectory $targetDirectory `
         -searchDirectory $repoRoot
+
+    if ($LASTEXITCODE -gt 0) {
+        exit $LASTEXITCODE
+    }
 }
 
 if ($GOTESTEXITCODE) {


eng/scripts/test.ps1
```diff
@@ -49,9 +49,13 @@ if (Select-String -path ./report.xml -pattern '<testsuites></testsuites>' -simpl
         -config $repoRoot/eng/config.json `
         -serviceDirectory $targetDirectory `
         -searchDirectory $repoRoot
+
+    if ($LASTEXITCODE -gt 0) {
+        exit $LASTEXITCODE
```
weshaggard, line 54: Do you want to exit here or set GOTESTEXITCODE?

benbp: That exit code is for capturing the result from the `go test` command, whereas this is for returning the result from `go run coverage`. The functionality would end up being the same if I set `$GOTESTEXITCODE` but I'd rather not overload the variable. Alternatively I could just change the `$GOTESTEXITCODE` to `$SCRIPTEXITCODE` or something and set that.

benbp: We could also have the behavior where we always return the go test exit code if it's `> 0`, otherwise return the coverage exit code if it's `> 0` otherwise return 0?

benbp:
FYI @RickWinter after merge, this may fail other internal pipelines that have newly insufficient coverage from the last few weeks since the underlying behavior change was introduced that caused us to miss coverage failures.

eng/scripts/test.ps1
weshaggard: I suspect in this case it probably doesn't matter too much because it is the last thing to run but what about the `gocov convert` command should we be checking that for failure as well?

benbp: Good catch, updated.

PR #25458, opened 2025-10-17
Read Many Client Engine

tvaron3:
This pull request introduces client engine support for the `ReadManyItems` operation in the Azure Cosmos DB Go SDK, allowing the SDK to leverage a query engine for more efficient and scalable multi-item reads. The changes refactor and parallelize the internal request execution logic, add new abstractions for item identity for query engine interface.

### ReadManyItems client engine integration

* Added support for using a query engine in `ContainerClient.ReadManyItems`, including a new code path that builds and executes a read-many pipeline when a query engine is provided. 

### Internal abstractions and concurrency improvements

* Implemented the `runEngineRequests` helper to parallelize per-partition requests using goroutines, improving throughput and resource utilization for both query and read-many pipelines.
* Added the `ItemIdentity` struct and extended the `QueryEngine` interface with `CreateReadManyPipeline` to support engine-driven multi-item reads.

### New Linting Rules

* Fixed linting errors from new rules added by eng sys that broke our pipeline. 

### Testing

* Added new tests for `ReadManyItems` with a query engine, including cases for empty input and successful item retrieval, using a mock engine to validate the SDK's integration logic.
* Implemented a mock read-many pipeline in the test infrastructure to simulate engine behavior for unit tests.

### Relevant Issue
https://github.com/Azure/azure-cosmos-client-engine/issues/38?issue=Azure%7Cazure-sdk-for-go%7C25338

### DIFF
--- a/sdk/data/azcosmos/cosmos_partition_key_range.go
+++ b/sdk/data/azcosmos/cosmos_partition_key_range.go
@@ -46,14 +46,14 @@ type partitionKeyRange struct {
 // MarshalJSON implements the json.Marshaler interface
 func (pkr partitionKeyRange) MarshalJSON() ([]byte, error) {
 	buffer := bytes.NewBufferString("{")
-	buffer.WriteString(fmt.Sprintf("\"id\":\"%s\"", pkr.ID))
+	fmt.Fprintf(buffer, "\"id\":\"%s\"", pkr.ID)
 
 	if pkr.ResourceID != "" {
-		buffer.WriteString(fmt.Sprintf(",\"_rid\":\"%s\"", pkr.ResourceID))
+		fmt.Fprintf(buffer, ",\"_rid\":\"%s\"", pkr.ResourceID)
 	}
 
 	if pkr.ETag != nil {
-		buffer.WriteString(",\"_etag\":")
+		fmt.Fprint(buffer, ",\"_etag\":")
 		etag, err := json.Marshal(pkr.ETag)
 		if err != nil {
 			return nil, err
@@ -62,23 +62,23 @@ func (pkr partitionKeyRange) MarshalJSON() ([]byte, error) {
 	}
 
 	if pkr.MinInclusive != "" {
-		buffer.WriteString(fmt.Sprintf(",\"minInclusive\":\"%s\"", pkr.MinInclusive))
+		fmt.Fprintf(buffer, ",\"minInclusive\":\"%s\"", pkr.MinInclusive)
 	}
 
 	if pkr.MaxExclusive != "" {
-		buffer.WriteString(fmt.Sprintf(",\"maxExclusive\":\"%s\"", pkr.MaxExclusive))
+		fmt.Fprintf(buffer, ",\"maxExclusive\":\"%s\"", pkr.MaxExclusive)
 	}
 
-	buffer.WriteString(fmt.Sprintf(",\"ridPrefix\":%d", pkr.ResourceIDPrefix))
+	fmt.Fprintf(buffer, ",\"ridPrefix\":%d", pkr.ResourceIDPrefix)
 
 	if pkr.SelfLink != "" {
-		buffer.WriteString(fmt.Sprintf(",\"_self\":\"%s\"", pkr.SelfLink))
+		fmt.Fprintf(buffer, ",\"_self\":\"%s\"", pkr.SelfLink)
 	}
 
-	buffer.WriteString(fmt.Sprintf(",\"throughputFraction\":%f", pkr.ThroughputFraction))
+	fmt.Fprintf(buffer, ",\"throughputFraction\":%f", pkr.ThroughputFraction)
 
 	if pkr.Status != "" {
-		buffer.WriteString(fmt.Sprintf(",\"status\":\"%s\"", pkr.Status))
+		fmt.Fprintf(buffer, ",\"status\":\"%s\"", pkr.Status)
 	}
 
 	if pkr.Parents != nil {
@@ -86,7 +86,7 @@ func (pkr partitionKeyRange) MarshalJSON() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		buffer.WriteString(",\"parents\":")
+		fmt.Fprint(buffer, ",\"parents\":")
 		buffer.Write(parents)
 	}
 
@@ -95,17 +95,17 @@ func (pkr partitionKeyRange) MarshalJSON() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		buffer.WriteString(",\"ownedArchivalPKRangeIds\":")
+		fmt.Fprint(buffer, ",\"ownedArchivalPKRangeIds\":")
 		buffer.Write(ids)
 	}
 
 	if !pkr.LastModified.IsZero() {
-		buffer.WriteString(fmt.Sprintf(",\"_ts\":%v", strconv.FormatInt(pkr.LastModified.Unix(), 10)))
+		fmt.Fprintf(buffer, ",\"_ts\":%v", strconv.FormatInt(pkr.LastModified.Unix(), 10))
 	}
 
-	buffer.WriteString(fmt.Sprintf(",\"lsn\":%d", pkr.LSN))
+	fmt.Fprintf(buffer, ",\"lsn\":%d", pkr.LSN)
 
-	buffer.WriteString("}")
+	fmt.Fprint(buffer, "}")
 	return buffer.Bytes(), nil
 }
 
--- a/sdk/data/azcosmos/cosmos_patch_operations.go
+++ b/sdk/data/azcosmos/cosmos_patch_operations.go
@@ -43,12 +43,12 @@ type PatchOperations struct {
 func (o PatchOperations) MarshalJSON() ([]byte, error) {
 	buffer := bytes.NewBufferString("{")
 	if o.condition != nil {
-		buffer.WriteString(fmt.Sprintf("\"condition\":\"%s\",", *o.condition))
+		fmt.Fprintf(buffer, "\"condition\":\"%s\",", *o.condition)
 	}
-	buffer.WriteString("\"operations\":[")
+	fmt.Fprint(buffer, "\"operations\":[")
 	for i, operation := range o.operations {
 		if i > 0 {
-			buffer.WriteString(",")
+			fmt.Fprint(buffer, ",")
 		}
 		operationBytes, err := json.Marshal(operation)
 		if err != nil {
@@ -56,7 +56,7 @@ func (o PatchOperations) MarshalJSON() ([]byte, error) {
 		}
 		buffer.Write(operationBytes)
 	}
-	buffer.WriteString("]}")
+	fmt.Fprint(buffer, "]}")
 	return buffer.Bytes(), nil
 }
 
--- a/sdk/data/azcosmos/cosmos_paths.go
+++ b/sdk/data/azcosmos/cosmos_paths.go
@@ -163,11 +163,11 @@ func createLink(parentPath string, pathSegment string, id string) string {
 	parentPathLength := len(parentPath)
 	completePath.Grow(parentPathLength + 2 + len(pathSegment) + len(id))
 	if parentPathLength > 0 {
-		completePath.WriteString(parentPath)
-		completePath.WriteString("/")
+		fmt.Fprint(&completePath, parentPath)
+		fmt.Fprint(&completePath, "/")
 	}
-	completePath.WriteString(pathSegment)
-	completePath.WriteString("/")
-	completePath.WriteString(url.PathEscape(id))
+	fmt.Fprint(&completePath, pathSegment)
+	fmt.Fprint(&completePath, "/")
+	fmt.Fprint(&completePath, url.PathEscape(id))
 	return completePath.String()
 }
--- a/sdk/data/azcosmos/cosmos_transactional_batch.go
+++ b/sdk/data/azcosmos/cosmos_transactional_batch.go
@@ -101,10 +101,10 @@ func (b batchOperationCreate) getOperationType() operationType {
 // MarshalJSON implements the json.Marshaler interface
 func (b batchOperationCreate) MarshalJSON() ([]byte, error) {
 	buffer := bytes.NewBufferString("{")
-	buffer.WriteString(fmt.Sprintf("\"operationType\":\"%s\"", b.operationType))
-	buffer.WriteString(",\"resourceBody\":")
+	fmt.Fprintf(buffer, "\"operationType\":\"%s\"", b.operationType)
+	fmt.Fprint(buffer, ",\"resourceBody\":")
 	buffer.Write(b.resourceBody)
-	buffer.WriteString("}")
+	fmt.Fprint(buffer, "}")
 	return buffer.Bytes(), nil
 }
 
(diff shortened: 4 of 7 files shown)


sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -63,3 +64,17 @@ type QueryPipeline interface {
 	// Close frees the resources associated with the pipeline.
 	Close()
 }
+
+// QueryPipeline is an interface that defines the methods for a query pipeline.
+type ReadManyPipeline interface {
```
analogrelay: Do we need a separate abstraction for a ReadMany pipeline? Could we just have these two `Create` methods, but they return the same `QueryPipeline` type?

sdk/data/azcosmos/cosmos_query_response.go
```diff
@@ -130,3 +130,14 @@ func newDatabasesQueryResponse(resp *http.Response) (QueryDatabasesResponse, err
 type queryDatabasesServiceResponse struct {
 	Databases []DatabaseProperties `json:"Databases,omitempty"`
 }
+
+// ReadManyItemsResponse contains response from the item query operation.
+type ReadManyItemsResponse struct {
+	Response
```
analogrelay: I'm hesitant to return a `Response` as part of this, because it sets the expectation that this API maps to a single HTTP request/response pair, even though it almost never will. For example, what headers do we include on the `Response` here?

There's a similar issue, IMO, with query responses today in the Go SDK, but since that's already released as public API, we can't do much about it.

My thinking is we just expose the items and the total RU charge for now, and add other data as needed later.

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -0,0 +1,45 @@
...
+// Executes a query using the provided query engine.
+func (c *ContainerClient) executeReadManyWithEngine(queryEngine queryengine.QueryEngine, items []ItemIdentity, readManyOptions *ReadManyOptions, operationContext pipelineRequestOptions) ([]ReadManyItemsResponse, error) {
+	// NOTE: The current interface for runtime.Pager means we're probably going to risk leaking the pipeline, if it's provided by a native query engine.
+	// There's no "Close" method, which means we can't call `queryengine.QueryPipeline.Close()` when we're done.
+	// We _do_ close the pipeline if the user iterates the entire pager, but if they don't we don't have a way to clean up.
+	// To mitigate that, we expect the queryengine.QueryPipeline to handle setting up a Go finalizer to clean up any native resources it holds.
+	// Finalizers aren't deterministic though, so we should consider making the pager "closable" in the future, so we have a clear signal to free the native resources.
+
+	// if empty list of items, return empty list
+	if len(items) == 0 {
```
analogrelay: We could probably promote this to the `ReadMany` API itself, just in case we do have alternate implementations of ReadMany. No matter which one we use, and empty list of items should always yield an empty result set.

sdk/data/azcosmos/cosmos_read_many_request_options.go
```diff
@@ -0,0 +1,83 @@
...
+	// If you wanted these nodes to participate in the same session (to be able read your own writes consistently across web tiers),
+	// you would have to send the SessionToken from the response of the write action on one node to the client tier, using a cookie or some other mechanism, and have that token flow back to the web tier for subsequent reads.
+	// If you are using a round-robin load balancer which does not maintain session affinity between requests, such as the Azure Load Balancer,the read could potentially land on a different node to the write request, where the session was created.
+	SessionToken *string
+	// ConsistencyLevel overrides the account defined consistency level for this operation.
+	// Consistency can only be relaxed.
+	ConsistencyLevel *ConsistencyLevel
+	// PopulateIndexMetrics is used to obtain the index metrics to understand how the query engine used existing indexes and how it could use potential new indexes.
+	// Please note that this options will incur overhead, so it should be enabled only when debugging slow queries and not in production.
+	PopulateIndexMetrics bool
```
analogrelay: Similarly to my above comment on the `Response` in the return value, I wonder if we should include query-specific options here, since there's no guarantee ReadMany is implemented using queries 🤔 .

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -6,9 +6,17 @@ package queryengine
 // QueryEngine is an interface that defines the methods for a query engine.
 type QueryEngine interface {
 	CreateQueryPipeline(query string, plan string, pkranges string) (QueryPipeline, error)
+	CreateReadManyPipeline(partitionKeyRangesData []byte, items []ItemIdentity, pkKind string, pkVersion int32) (QueryPipeline, error)
```
analogrelay: The `pkranges` parameter should be the same as for `CreateQueryPipeline`, since it should be expecting the same data.

```suggestion
	CreateReadManyPipeline(pkranges string, items []ItemIdentity, pkKind string, pkVersion int32) (QueryPipeline, error)
```

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -4,19 +4,252 @@
...
+	done := make(chan struct{})
+	providerDone := make(chan struct{})
+	var wg sync.WaitGroup
+	var chargeMu sync.Mutex
+
+	// Provider goroutine: serializes calls into the pipeline
+	go func() {
+		defer close(providerDone)
+		for res := range provideCh {
+			if err := readManyPipeline.ProvideData(res); err != nil {
```
analogrelay: :thinking: We should consider making it safe to call `ProvideData` simultaneously from multiple threads (it's _not_ safe to do so today, in the Rust engine code, so this code is correct).

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -4,19 +4,252 @@
...
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	azlog "github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
 	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos/queryengine"
+	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
-// executeReadManyWithEngine executes a query using the provided query engine.
+// Executes a query using the provided query engine.
 func (c *ContainerClient) executeReadManyWithEngine(queryEngine queryengine.QueryEngine, items []ItemIdentity, readManyOptions *ReadManyOptions, operationContext pipelineRequestOptions, ctx context.Context) (ReadManyItemsResponse, error) {
```
analogrelay, line 17: This is looking OK to me. I'd like to try and unify the pipeline "driving" code between read many and cross-partition query, since it has all the same basic patterns. The concurrency you're doing here should _also_ be possible with the query pipeline once we make some improvements. But I'm OK with doing that refactoring later, once we have an end-to-end working ReadMany implementation

tvaron3: Ill extend this to query. I will iterate first on this with testing and then will expand it to query.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: makes sense will update

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: I changed it but still debating over this because for read many some of these methods I don't believe are needed.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: Make sure pkVersion is not 0.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: Actually will leave it as query pipeline because it allows for the reused code between execute query and execute read many

sdk/data/azcosmos/cosmos_container_read_many.go
tvaron3: moved the check up

sdk/data/azcosmos/cosmos_read_many_request_options.go
tvaron3: Removed for now. Had them initially for parity with other SDKs.

sdk/data/azcosmos/cosmos_container_read_many.go
tvaron3: Expanded it to query as well.

sdk/data/azcosmos/cosmos_query_response.go
tvaron3: Makes sense followed suggestion

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -6,9 +6,18 @@ package queryengine
 // QueryEngine is an interface that defines the methods for a query engine.
 type QueryEngine interface {
 	CreateQueryPipeline(query string, plan string, pkranges string) (QueryPipeline, error)
+	CreateReadManyPipeline(pkranges string, items []ItemIdentity, pkKind string, pkVersion int32) (QueryPipeline, error)
```
simorenoh: Generally signatures follow more to least relevant pattern for the arguments being used, and item identities are the only mandatory user input. Just a comment since the ordering and naming have to be aligned with the client engine one as well.
```suggestion
	CreateReadManyPipeline(itemIdentities []ItemIdentity, pkranges string,  pkKind string, pkVersion int32) (QueryPipeline, error)
```

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -6,34 +6,101 @@ package azcosmos
...
+		}
+	}
+	var pkVersion int32
+	if containerRsp.ContainerProperties.PartitionKeyDefinition.Version == 0 {
+		pkVersion = int32(1)
+	} else {
+		pkVersion = int32(containerRsp.ContainerProperties.PartitionKeyDefinition.Version)
+	}
+
+	readManyPipeline, err := queryEngine.CreateReadManyPipeline(string(rawPartitionKeyRanges), newItemIdentities, string(containerRsp.ContainerProperties.PartitionKeyDefinition.Kind), pkVersion)
```
simorenoh: Since the client engine expects strings, we should serialize this into a JSON string before using it - also the ordering ideally changed
```suggestion
	newItemIdentitiesJSON, err := json.Marshal(newItemIdentities)
	if err != nil {
		return ReadManyItemsResponse{}, err
	}
	readManyPipeline, err := queryEngine.CreateReadManyPipeline(string(newItemIdentitiesJSON), string(rawPartitionKeyRanges), string(containerRsp.ContainerProperties.PartitionKeyDefinition.Kind), pkVersion)
```

simorenoh (changes requested):
Just two suggestions based on the client engine work - regardless of what gets done with them we should align to make sure these two look the same across the board.

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -6,34 +6,101 @@ package azcosmos
...
+			return ReadManyItemsResponse{}, err
+		}
+		newItemIdentities[i] = queryengine.ItemIdentity{
+			PartitionKeyValue: pkStr,
+			ID:                items[i].ID,
+		}
+	}
+	var pkVersion int32
+	if containerRsp.ContainerProperties.PartitionKeyDefinition.Version == 0 {
+		pkVersion = int32(1)
```
simorenoh: This should default to 2 (default across service) not 1 (legacy)

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: makes sense changed.

sdk/data/azcosmos/cosmos_container_read_many.go
Pilchie: Is that true?  I would think if it's omitted in the response, it's 1, but 2 is always specified in the response.  Can we verify?

tvaron3: Sorry, we discussed offline. Should have updated here. The default is 1. We verified with elasticity team and the defaults in others sdks.

simorenoh (approved):
Have been using these changes to test against the engine so looks good to me, thanks Tomas!

sdk/data/azcosmos/internal/mock_query_engine.go
```diff
@@ -79,6 +79,58 @@ func (m *MockQueryEngine) CreateQueryPipeline(query string, plan string, pkrange
...
+
+func (m *MockReadManyPipeline) Run() (*queryengine.PipelineResult, error) {
+	if m.IsComplete() {
+		return &queryengine.PipelineResult{IsCompleted: true, Items: m.resultingItems, Requests: nil}, nil
+	}
+	// first run return queries to execute
+	requests := make([]queryengine.QueryRequest, 0, len(m.items))
+	for i := range m.items {
+		pk := m.items[i].PartitionKeyValue
+		create_query := fmt.Sprintf("Select * from c where c.id = '%s' and c.pk = '%s'", m.items[i].ID, pk)
```
analogrelay, line 112: Super nit (no need to churn the PR just for this): This should be `createQuery`, to match Go style. Though I suspect `go fmt` will yell at you about it too, so maybe you'll have to update the PR anyway

sdk/data/azcosmos/cosmos_container_query_engine.go
```diff
@@ -151,71 +153,211 @@ func (c *ContainerClient) executeQueryWithEngine(queryEngine queryengine.QueryEn
...
+	errCh := make(chan error, 1)
+	done := make(chan struct{})
+	providerDone := make(chan struct{})
+	var wg sync.WaitGroup
+	var chargeMu sync.Mutex
+
+	// Provider goroutine ensures only one ProvideData executes at a time.
+	go func() {
+		defer close(providerDone)
+		for batch := range provideCh {
```
analogrelay, line 211: As a refactoring for later, we can actually just collect all the results and call `ProvideData` once with a single batch. That's what it's designed for. You're right that it shouldn't be called simultaneously, but calling it once with a batch of items will also reduce CGo overhead.

But, let's merge this as-is and address that later.

sdk/data/azcosmos/cosmos_container_query_engine.go
```diff
@@ -151,71 +153,211 @@ func (c *ContainerClient) executeQueryWithEngine(queryEngine queryengine.QueryEn
...
+						if err != nil {
+							select {
+							case errCh <- err:
+							default:
+							}
+							return
+						}
+						chargeMu.Lock()
+						totalCharge += qResp.RequestCharge
+						chargeMu.Unlock()
```
analogrelay, line 279: Slight performance suggestion for a separate PR: Store request charges in an array with `workerCount` slots. Each worker can update the request charge slot using it's worker index `w` without a lock. Sum up all the slots at the end, when all the workers are done.

Don't churn this PR just for that though.

analogrelay (approved):
Nothing blocking from my end

sdk/data/azcosmos/internal/mock_query_engine.go
tvaron3: fixed as had to fix a bunch of linting errors as well.

PR #25459, opened 2025-10-17
Fail on gocov convert

benbp:
Forgot to push this commit for #25447 but it's standalone.


weshaggard:
You need to account for this https://github.com/Azure/azure-sdk-for-go/pull/25377/files

benbp:
#25377 makes this irrelevant.

PR #25489, opened 2025-10-22
Fix coverage goal evaluation

chlowell:
[config.json](https://github.com/Azure/azure-sdk-for-go/blob/main/eng/config.json) defines coverage goals for packages identified by substrings of their module paths, for example `keyvault/azkeys`. CI scripts specify the module under test to the coverage goal tool with a disk path like `/mnt/vss/_work/1/s/sdk/security/keyvault/azkeys`. Today, the tool searches config.json for a package whose name exactly matches that path or, failing that, the first package whose (partial) name is a substring of it. In practice the coverage tool always returns that first substring match because no package name in config.json equals any disk path in the repo.

That behavior introduces a couple bugs that weren't obvious until this week because pipelines were evaluating but not enforcing coverage goals:
- applying the wrong goal to a package e.g. `azidentity/cache` gets the goal for `azidentity`
- incorrectly applying the default goal to a package e.g. config.json has a goal for `keyvault/azkeys` but the tool doesn't find it on Windows because there it's given a disk path like `D:\...\keyvault\azkeys`

Now that pipelines do enforce coverage goals, these bugs are causing CI runs to fail. The fix here is to have the coverage tool normalize slashes in the given path and search config.json for the longest i.e. most specific matching package name.

### DIFF
--- a/eng/tools/internal/coverage/coverage.go
+++ b/eng/tools/internal/coverage/coverage.go
@@ -72,20 +72,26 @@ func readConfigData(coverageConfig string) *codeCoverage {
 
 // This supports doing a single package at a time. If this needs to be expanded in the future
 // this method will have to return a []*float64 for each packages goal
+//
+// Packages are identified in configData (parsed from /eng/config.json) by segments of their
+// module paths, for example "keyvault/azkeys", whereas covFile may be a disk path like
+// "/vss/1/sdk/security/keyvault/azkeys". So, this function searches for the configData entry
+// whose Name is the longest substring of covFile.
 func findCoverageGoal(covFiles []string, configData *codeCoverage) float64 {
 	for _, covFile := range covFiles {
-
-		// check for an exact match _first_, then go to fuzzy matching
-		for _, p := range configData.Packages {
-			if covFile == p.Name {
-				return p.CoverageGoal
+		covFile = strings.ReplaceAll(covFile, `\`, "/")
+		var bestMatch *coveragePackage
+		for i := range configData.Packages {
+			p := &configData.Packages[i]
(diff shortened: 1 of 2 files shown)


eng/tools/internal/coverage/coverage.go
```diff
@@ -72,20 +72,26 @@ func readConfigData(coverageConfig string) *codeCoverage {
...
-
-		// check for an exact match _first_, then go to fuzzy matching
-		for _, p := range configData.Packages {
-			if covFile == p.Name {
-				return p.CoverageGoal
+		covFile = strings.ReplaceAll(covFile, `\`, "/")
+		var bestMatch *coveragePackage
+		for i := range configData.Packages {
+			p := &configData.Packages[i]
+			prx := regexp.MustCompile(`(^|/)` + regexp.QuoteMeta(p.Name) + `($|/)`)
```
jhendrixMSFT, line 86: `p.Name` here is something like `sdk/security/keyvault/azkeys` etc yes?

jhendrixMSFT:
At some point we should look at moving the coverage goals out of `/eng` and into their respective SDK directories (maybe in the `ci.yaml` file).

jhendrixMSFT:

eng/tools/internal/coverage/coverage.go
chlowell: No; package names in config.json are partial paths not necessarily rooted in any particular directory in the repo. I guess the intent was for names to be rooted in `/sdk` but that's not what we've ended up with. For example we have `keyvault/azkeys` and `security/keyvault/azadmin` (I suppose we didn't update the original `keyvault/...` entries after moving the modules under `sdk/security`)

jhendrixMSFT: Got it. So, I guess the value is enough to disambiguate SDKs.

PR #25503, opened 2025-10-24
cosmos: Allow the Query Pipeline to return an alternative query to execute in each request

tvaron3:
Porting https://github.com/Azure/azure-sdk-for-rust/pull/3166 as it is a pre-req for readmany.


sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -58,7 +79,14 @@ type QueryPipeline interface {
...
-	// ProvideData provides more data for a given partition key range ID, using data retrieved from the server in response to making a DataRequest.
+	// Data from multiple partition ranges may be provided at once.
+	// However, each page of data must be provided in order.
+	// So, for any given partition key range, page n's results must be earlier in the `data` vector than page n+1's results.
+	// Data from different partition key ranges may be interleaved, as long as each partition key range's pages are in order.
+	//
+	// The pipeline will use the QueryResult.RequestIndex field to validate this.
+	//
+	// When providing data from a draining request (i.e. a request with Drain set to true), all pages for that draining request can share the same QueryResult.RequestIndex.
 	ProvideData(data QueryResult) error
```
analogrelay: This should be a slice now, `[]QueryResult`

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -58,7 +79,14 @@ type QueryPipeline interface {
 	IsComplete() bool
 	// Run executes a single turn of the pipeline, yielding a PipelineResult containing the items and requests for more data.
 	Run() (*PipelineResult, error)
-	// ProvideData provides more data for a given partition key range ID, using data retrieved from the server in response to making a DataRequest.
+	// Data from multiple partition ranges may be provided at once.
+	// However, each page of data must be provided in order.
+	// So, for any given partition key range, page n's results must be earlier in the `data` vector than page n+1's results.
```
analogrelay: nit: since Go's terminology is different here.

```suggestion
	// So, for any given partition key range, page n's results must be earlier in the `data` slice than page n+1's results.
```

sdk/data/azcosmos/cosmos_container_query_engine.go
```diff
@@ -151,44 +151,57 @@ func (c *ContainerClient) executeQueryWithEngine(queryEngine queryengine.QueryEn
...
+
+						// Provide the data to the pipeline, make sure it's tagged with the partition key range ID so the pipeline can merge it into the correct partition.
+						result := queryengine.QueryResult{
+							PartitionKeyRangeID: request.PartitionKeyRangeID,
+							RequestIndex:        request.Index,
+							NextContinuation:    continuation,
+							Data:                data,
+						}
+						log.Writef(EventQueryEngine, "Received response for PKRange: %s. Continuation present: %v", request.PartitionKeyRangeID, continuation != "")
+						if err = queryPipeline.ProvideData(result); err != nil {
```
analogrelay: For now, to replicate what I did in Rust, you can just send a slice with one element in it. Eventually, the idea is to optimize things by allowing multiple results to be passed in, but this is just intended to put the APIs in place for us to do that later. To actually get a benefit from optimizations like `Drain` (which is designed to reduce calls between Go and Rust), we'll have to batch these requests and provide all the results at once, but that's a more complicated change.

sdk/data/azcosmos/cosmos_container_query_engine.go
```diff
@@ -151,44 +151,57 @@ func (c *ContainerClient) executeQueryWithEngine(queryEngine queryengine.QueryEn
...
-						return QueryItemsResponse{}, err
+
+					fetchMorePages := true
+					for fetchMorePages {
+
+						azResponse, err := c.database.client.sendQueryRequest(
+							path,
+							ctx,
+							query,
+							queryOptions.QueryParameters,
```
analogrelay: We should only do this if `qryRequest.Query == "" || qryRequest.IncludeParameters`. If there's an override query but `IncludeParameters` is false, we should provide just an empty set of parameters.

analogrelay: If it feels easier or just cleaner, I'm also happy to have the engine always set `IncludeParameters` if parameters should be included (even if there is no override query) to ensure the behavior of the SDK is consistent.

sdk/data/azcosmos/cosmos_container_query_engine_test.go
```diff
@@ -120,3 +133,137 @@ func TestQueryViaQueryEngine(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create test items: %v", err)
+	}
+
+	override := "SELECT * FROM c WHERE c.mergeOrder = @targetOrder"
+	cfg := &azcosmosinternal.QueryRequestConfig{Query: &override, IncludeParameters: true}
+	engine := azcosmosinternal.WithQueryRequestConfig(cfg)
+
+	// choose a target merge order present in the test data: use the first item's merge order (0)
+	target := items[0].MergeOrder
```
analogrelay, line 195: nit: There's a chance that this ends up being the same as the ID, which would invalidate the test. Can we do an assert that `items[0].ID != items[0].MergeOrder`?

tvaron3: good point! Changed the item generation and added an assertion to make sure they are different.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: changed

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: changed

sdk/data/azcosmos/cosmos_container_query_engine.go
tvaron3: followed this approach

sdk/data/azcosmos/cosmos_container_query_engine.go
tvaron3: Makes sense went with your first suggestion.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -13,15 +13,36 @@ type QueryEngine interface {
 type QueryRequest struct {
 	// PartitionKeyRangeID is the ID of the partition key range from which data is requested.
 	PartitionKeyRangeID string
+	// The index of this request, within the partition key range.
+	//
+	// This value will always increase for subsequent requests for the same partition key range.
+	// It must be provided back to the pipeline when providing data, so that the pipeline can ensure that data is provided in order.
+	Index uint
```
analogrelay: Can you change this to `Id` and update the doc comment so it no longer promises that the value will always increase? As I'm building out Hybrid search, I realized that it's quite difficult to use a sequential identifier here, and the SDK really doesn't need to know that it's sequential. All the SDK needs to do is that this value and pass it back with the `QueryResult`. Using `Id` makes it clearer this is just an opaque identifier. I'll be doing the same to the Rust SDK in a separate PR (since that's already merged).

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -13,15 +13,36 @@ type QueryEngine interface {
...
 }
 
 // QueryResult contains the result of a query for a specific partition key range.
 type QueryResult struct {
+	// The ID of the partition key range that was queried.
 	PartitionKeyRangeID string
-	NextContinuation    string
-	Data                []byte
+	// The index of the QueryRequest that generated this result.
+	RequestIndex uint
```
analogrelay: Similar to the above, this becomes `RequestId` (and the doc comment should be updated similarly.

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -13,15 +13,36 @@ type QueryEngine interface {
...
 	// Continuation is the continuation token to use in the request.
 	Continuation string
+	// The query to execute for this partition key range, if different from the original query.
+	Query string
+	// If a query is specified, this flag indicates if the query parameters should be included with that query.
+	//
+	// Sometimes, when an override query is specified, it differs in structure from the original query, and the original parameters are not valid.
+	IncludeParameters bool
+	// If specified, indicates that the SDK should IMMEDIATELY drain all remaining results from this partition key range, following continuation tokens, until no more results are available.
+	// All the data from this partition key range should be provided BEFORE any new items will be made available.
```
analogrelay, line 29: Just a small addition to clarify the interaction between RequestId and Drain.

```suggestion
	// All the data from this partition key range should be provided BEFORE any new items will be made available.
	// The data may be provided in multiple QueryResults, but every result correlated to this request should have the same RequestId value.
```

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -58,8 +79,15 @@ type QueryPipeline interface {
...
 	// Run executes a single turn of the pipeline, yielding a PipelineResult containing the items and requests for more data.
 	Run() (*PipelineResult, error)
-	// ProvideData provides more data for a given partition key range ID, using data retrieved from the server in response to making a DataRequest.
-	ProvideData(data QueryResult) error
+	// Data from multiple partition ranges may be provided at once.
+	// However, each page of data must be provided in order.
+	// So, for any given partition key range, page n's results must be earlier in the `data` slice than page n+1's results.
+	// Data from different partition key ranges may be interleaved, as long as each partition key range's pages are in order.
+	//
+	// The pipeline will use the QueryResult.RequestIndex field to validate this.
```
analogrelay: Update this after the field rename

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: added

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: renamed field

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: renamed field and updated comment

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: Changed to id and updated comment

analogrelay (approved):

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -13,15 +13,36 @@ type QueryEngine interface {
 type QueryRequest struct {
 	// PartitionKeyRangeID is the ID of the partition key range from which data is requested.
 	PartitionKeyRangeID string
+	// The ID of this request, within the partition key range.
+	//
+	// Opaque identifier that must be provided back to the pipeline when providing data.
+	Id uint
```
analogrelay: Sorry to force one more bit of churn but this should be a `uint64`, I think. I'll make a similar change on the Rust side. Because it's crossing the FFI barrier, it's important we be exact about sizes, and the pipeline may be doing bitwise operations on the ID (I'm doing a little of that for Hybrid).


```suggestion
	Id uint64
```

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
```diff
@@ -13,15 +13,36 @@ type QueryEngine interface {
...
 }
 
 // QueryResult contains the result of a query for a specific partition key range.
 type QueryResult struct {
+	// The ID of the partition key range that was queried.
 	PartitionKeyRangeID string
-	NextContinuation    string
-	Data                []byte
+	// The ID of the QueryRequest that generated this result.
+	RequestId uint
```
analogrelay:
```suggestion
	RequestId uint64
```

analogrelay (changes requested):
Sorry, one last change 😅

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: no worries, updated

sdk/data/azcosmos/queryengine/cosmos_query_engine.go
tvaron3: updated

PR #25514, opened 2025-10-28
fixed incorrect surrounding quotes check

whisk:
# Description

Payloads containing only `"` caused slice out of bounds panic.

### DIFF
--- a/sdk/azcore/internal/exported/exported.go
+++ b/sdk/azcore/internal/exported/exported.go
@@ -92,7 +92,7 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
 		return nil
 	}
 	payload := string(s)
-	if payload[0] == '"' {
+	if len(payload) >= 2 && payload[0] == '"' && payload[len(payload)-1] == '"' {
 		// remove surrounding quotes
 		payload = payload[1 : len(payload)-1]
 	}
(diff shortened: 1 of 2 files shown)


sdk/azcore/internal/exported/exported.go
```diff
@@ -92,7 +92,7 @@ func DecodeByteArray(s string, v *[]byte, format Base64Encoding) error {
 		return nil
 	}
 	payload := string(s)
-	if payload[0] == '"' {
+	if len(payload) >= 2 && payload[0] == '"' && payload[len(payload)-1] == '"' {
```
RickWinter, line 95: You're calculating the len(payload) of 3 times. Ideally a compiler is smart enough to optimize and hoist the checkt out, however we shouldn't rely on that optimization. Grab the len(payload) and reuse it in place of the three calls.

```
payload_len = len(payload)
```

whisk: Thank you for your attention!

As far as I know, `len(payload)` of the local string `payload` (it is made local on the line 94) is a direct memory read of the string header which contains the string length, and not a function call. The string header is on the stack for the local variable. This is very fast, that is why I decided to keep it simple.

However, I don't know if the compiler has preferences using register cache for explicit local int variables. This might be marginally more efficient.

jhendrixMSFT:
Thanks!

PR #25534, opened 2025-10-31
Fix the token audiences of aztables for sovereign clouds

xumou-ms:

### DIFF
--- a/sdk/data/aztables/cloud_config.go
+++ b/sdk/data/aztables/cloud_config.go
@@ -11,6 +11,8 @@ import "github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 const ServiceName cloud.ServiceName = "data/aztables"
 
 func init() {
+	// for Azure Table Storage endpoints, these values are ignored as the audience is always "https://storage.azure.com"
+	// for Cosmos endpoints, we will use the audiences as specified here
 	cloud.AzureChina.Services[ServiceName] = cloud.ServiceConfiguration{
 		Audience: "https://cosmos.azure.cn",
 	}
--- a/sdk/data/aztables/service_client.go
+++ b/sdk/data/aztables/service_client.go
@@ -39,6 +39,7 @@ func NewServiceClient(serviceURL string, cred azcore.TokenCredential, options *C
 		return nil, errors.New("cloud configuration is missing for Azure Tables")
 	}
 
+	// unlike Cosmos, Azure Table Storage uses the same audience for all clouds, public and sovereign.
 	audience := "https://storage.azure.com"
 	if isCosmosEndpoint(serviceURL) {
 		audience = cfg.Audience


sdk/data/aztables/cloud_config.go
```diff
@@ -12,12 +12,12 @@ const ServiceName cloud.ServiceName = "data/aztables"
 
 func init() {
 	cloud.AzureChina.Services[ServiceName] = cloud.ServiceConfiguration{
-		Audience: "https://storage.azure.cn",
+		Audience: "https://cosmos.azure.cn",
```
richardpark-msft, line 17: @jhendrixMSFT, @chlowell - @xumou-ms is trying to fix a bug with token audiences. 

Azure Storage uses storage.azure.com in all clouds (apparently). Cosmos does use cloud specific audiences.

My question to you is what our contract with CloudConfig is - is it considered a contract breakage if the scope changes from being the storage scope, to being a cosmos scope instead?

sdk/data/aztables/service_client.go
```diff
@@ -39,9 +39,9 @@ func NewServiceClient(serviceURL string, cred azcore.TokenCredential, options *C
 		return nil, errors.New("cloud configuration is missing for Azure Tables")
 	}
 
-	audience := cfg.Audience
+	audience := "https://storage.azure.com"
```
richardpark-msft, line 43:
```suggestion
  // unlike Cosmos, Azure Storage uses the same audience for all clouds, public and sovereign.
	audience := "https://storage.azure.com"
```

sdk/data/aztables/cloud_config.go
chlowell: The value isn't contractual, so I'd say the question is whether there's any operation that succeeds with the Storage audience and fails with the Cosmos audience. Do we have a Cosmos contact?

sdk/data/aztables/cloud_config.go
```diff
@@ -12,12 +12,12 @@ const ServiceName cloud.ServiceName = "data/aztables"
 
 func init() {
```
jhendrixMSFT, line 13: We should document that the default is for table storage which is `https://storage.azure.com`. We only consume these values if it's a Cosmos endpoint.

sdk/data/aztables/cloud_config.go
richardpark-msft: This just swaps out the audience for the token. The operations for the services are the same, apart from that, so I don't think there's anything for Cosmos to do here. This is purely a client-side thing.

jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/25542

sdk/data/aztables/cloud_config.go
richardpark-msft: Maybe something like this:

```suggestion
func init() {
   // unlike Cosmos, Azure Storage uses the same audience for all clouds (https://storage.azure.com), public and sovereign.
```

sdk/data/aztables/cloud_config.go
richardpark-msft: Talked with @jhendrixMSFT offline, we'll add some code comments for this but it should be fine to change cloud config.

richardpark-msft:
@xumou-ms, if you can merge in the two suggested comments then we should be good to go.

sdk/data/aztables/cloud_config.go
jhendrixMSFT:
```go
// for table storage endpoints, these values are ignored as the audience is always https://storage.azure.com
// for cosmos endpoints, we will use the audience as specified here
```

xumou-ms:
Thanks to @jhendrixMSFT @richardpark-msft for reviewing this PR! I merged in the two suggested comments

PR #25613, opened 2025-11-13
Fix wrong parsing logic for module name config for swagger with major version suffix.

jliusan:
https://github.com/Azure/azure-rest-api-specs/pull/38738


jliusan:
> Please add a changelog entry and bump the version.

done

JiaqiZhang-Dev (approved):

PR #25664, opened 2025-11-24
[azidentity] Allow Identity Binding mode to be set via configuration options

### LINKED ISSUE #25663: [azidentity] Allow Identity Binding mode to be set via configuration options instead of env vars
### Feature Request

Follow-up to: #25056

#### Background

Tody the identity binding mode in workload identity is controlled only via environment variables defined in:
https://github.com/Azure/azure-sdk-for-go/blob/4620071edb714a0087405b84a662696ec52e8ed0/sdk/azidentity/internal/customtokenproxy/transport.go#L20-L26

In addition, the Azure Federated Token for workload identity is currently only read from a file on disk:
https://github.com/Azure/azure-sdk-for-go/blob/4620071edb714a0087405b84a662696ec52e8ed0/sdk/azidentity/workload_identity.go#L70-L72

This makes the behavior hard to configure from code and hard to integrate into libiraries that already have their own configuration and/or use non-file token sources.

#### Request
1. Add a configuration option to set the identity binding proxy from code with explicit options taking precedence over env vars.
2. Keep env-var behavior as the default when the option is not set.
3. For workload identity, allow callers to provide a `GetFederatedToken` style override so that the token can come from non-file sources (e.g., token from request in csi scenario), while preserving the current file-based behavior as the default.

clnv:
fix #25663

### DIFF
--- a/sdk/azidentity/go.work.sum
+++ b/sdk/azidentity/go.work.sum
@@ -1,8 +1,0 @@
-github.com/golang-jwt/jwt/v4 v4.4.3/go.mod h1:m21LjoU+eqJr34lmDMbreY2eSTRJ1cv77w39/MY0Ch0=
-github.com/keybase/dbus v0.0.0-20220506165403-5aa21ea2c23a/go.mod h1:YPNKjjE7Ubp9dTbnWvsP3HT+hYnY6TfXzubYTBeUxc8=
-github.com/montanaflynn/stats v0.7.0/go.mod h1:etXPPgVO6n31NxCd9KQUMvCM+ve0ruNzt6R8Bnaayow=
-github.com/stretchr/objx v0.5.2/go.mod h1:FRsXN1f5AsAjCGJKqEizvkpNtU+EGNCLh3NxZ/8L+MA=
-golang.org/x/mod v0.26.0/go.mod h1:/j6NAhSk8iQ723BGAUyoAcn7SlD7s15Dp9Nd/SfeaFQ=
-golang.org/x/sync v0.16.0/go.mod h1:1dzgHSNfp02xaA81J2MS99Qcpr2w7fw1gpm99rleRqA=
-golang.org/x/term v0.34.0/go.mod h1:5jC53AEywhIVebHgPVeg0mj8OD3VO9OzclacVrqpaAw=
-golang.org/x/tools v0.35.0/go.mod h1:NKdj5HkL/73byiZSJjqJgKn3ep7KjFkBOkR/Hps3VPw=


sdk/azidentity/go.work.sum
enj, line 1: I think you can delete this file.

sdk/azidentity/internal/customtokenproxy/configuration.go
```diff
@@ -0,0 +1,139 @@
...
+		"only one of AzureKubernetesCAFile or AzureKubernetesCAData can be specified",
+	)
+)
+
+func noopConfigure(*policy.ClientOptions) {
+	// no-op
+}
+
+// Apply returns a function that configures the client options to use the custom token proxy.
+func Apply(opts *Options) (func(*policy.ClientOptions), error) {
```
enj:
```suggestion
func (opts *Options) Apply() (func(*policy.ClientOptions), error) {
```

sdk/azidentity/internal/customtokenproxy/configuration.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, fmt.Errorf("custom token endpoint URL %q must not contain a fragment", tokenProxy)
+	}
+	if tokenProxy.EscapedPath() == "" {
+		// if the path is empty, set it to "/" to avoid stripping the path from req.URL
+		tokenProxy.Path = "/"
+	}
+	return tokenProxy, nil
+}
+
+func (o *Options) defaults() {
```
enj: Only perform the defaulting when all fields are empty.  Do not attempt to merge the env config and any user set config.

sdk/azidentity/workload_identity.go
```diff
@@ -55,21 +59,31 @@ type WorkloadIdentityCredentialOptions struct {
...
 	// tokens directly from Entra ID.
 	//
 	// The proxy feature is designed for applications that deploy to many clusters and clusters that host many
 	// applications. See the Azure Kubernetes Service identity bindings documentation for more information on when
 	// to set this option: https://learn.microsoft.com/azure/aks/identity-bindings-concepts
 	EnableAzureProxy bool
 
+	// AzureProxy specifies the options for the Azure proxy.
+	// If EnableAzureProxy is false, this field is ignored.
+	AzureProxy WorkloadIdentityAzureProxyOptions
```
enj:
```suggestion
	AzureProxy customtokenproxy.Options
```

And delete the `WorkloadIdentityAzureProxyOptions` alias.

sdk/azidentity/workload_identity.go
```diff
@@ -55,21 +59,31 @@ type WorkloadIdentityCredentialOptions struct {
...
 	TenantID string
 
 	// TokenFilePath is the path of a file containing a Kubernetes service account token. Defaults to the value of the
 	// environment variable AZURE_FEDERATED_TOKEN_FILE.
+	// If GetFederatedToken is set, this field is ignored.
 	TokenFilePath string
+
+	// GetFederatedToken defines an optional func to get the Kubernetes service account token.
+	// If this function is set, it will be used to get the token instead of reading it from TokenFilePath.
+	GetFederatedToken func(ctx context.Context) (string, error)
```
enj: Make setting both of these an error.

sdk/azidentity/workload_identity.go
```diff
@@ -55,21 +59,31 @@ type WorkloadIdentityCredentialOptions struct {
...
+	// When this value is true and proxy configuration isn't present or this value is false, the credential will request
 	// tokens directly from Entra ID.
 	//
 	// The proxy feature is designed for applications that deploy to many clusters and clusters that host many
 	// applications. See the Azure Kubernetes Service identity bindings documentation for more information on when
 	// to set this option: https://learn.microsoft.com/azure/aks/identity-bindings-concepts
 	EnableAzureProxy bool
 
+	// AzureProxy specifies the options for the Azure proxy.
+	// If EnableAzureProxy is false, this field is ignored.
```
enj: Make setting this an error when `EnableAzureProxy` is `false`.

sdk/azidentity/workload_identity.go
jhendrixMSFT: The type `customtokenproxy.Options` is under the `/internal` directory which means callers outside of `azidentity` can't access the type.  The type alias provides a means for external access (see my other comment about this).

sdk/azidentity/workload_identity.go
```diff
@@ -26,12 +26,16 @@ const credNameWorkloadIdentity = "WorkloadIdentityCredential"
...
-	mtx             *sync.RWMutex
+	assertion, file      string
+	getAssertionOverride func(ctx context.Context) (string, error)
+	cred                 *ClientAssertionCredential
+	expires              time.Time
+	mtx                  *sync.RWMutex
 }
 
+// WorkloadIdentityAzureProxyOptions contains optional parameters for configuring WorkloadIdentity Azure Proxy.
+type WorkloadIdentityAzureProxyOptions = customtokenproxy.Options
```
jhendrixMSFT: Our convention when exporting internal types is to place the internal type in an `/internal/exported` package so it's clear that the internal type is actually part of public surface area (see [example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/exported/exported.go)).

sdk/azidentity/internal/customtokenproxy/configuration.go
jhendrixMSFT: If `Options` is to be exported, then the current implementation of `Apply()` is correct (i.e. it _must not_ be an exported method on `Options`).

jhendrixMSFT (changes requested):

sdk/azidentity/workload_identity.go
clnv: moved to `/internal/exported`.

sdk/azidentity/workload_identity.go
clnv: validation added.

sdk/azidentity/workload_identity.go
clnv: validation added

sdk/azidentity/internal/customtokenproxy/configuration.go
clnv: yea, that's why

sdk/azidentity/workload_identity.go
```diff
@@ -52,21 +57,32 @@ type WorkloadIdentityCredentialOptions struct {
...
-	// this value is true and proxy configuration isn't present or this value is false, the credential will request
+	// EnableAzureProxy determines whether the credential reads proxy configuration from environment variables or
+	// from the CustomTokenProxy field.
+	// When this value is true and proxy configuration isn't present or this value is false, the credential will request
 	// tokens directly from Entra ID.
 	//
 	// The proxy feature is designed for applications that deploy to many clusters and clusters that host many
 	// applications. See the Azure Kubernetes Service identity bindings documentation for more information on when
 	// to set this option: https://learn.microsoft.com/azure/aks/identity-bindings-concepts
 	EnableAzureProxy bool
```
clnv, line 68: @enj @chlowell do you think we should use `EnableCustomTokenProxy` instead?

clnv:
@microsoft-github-policy-service agree company="Microsoft"

PR #25683, opened 2025-11-28
[App Configuration] fix lint error and remove build tags for go 1.18

linglingye001:


eng/config.json
```diff
@@ -6,7 +6,7 @@
         },
         {
             "Name": "azappconfig",
-            "CoverageGoal": 0.19
+            "CoverageGoal": 0.15
```
RickWinter: Why did the coverage goal decrease? Rather than dropping the coverage, it would be better to add tests to increase the coverage.

linglingye001: Since last week the test failed in build pipeline because of insufficient coverage, not sure why this happened when there's no change in coverage configuration. Based on the log, the coverage is more than 80% in most folders except the `generated` which is `0`. Is there a way to exclude the `generated` folder when running tests?

jhendrixMSFT: Recent versions of Go have changed how the total percentage is calculated.  In short, the contents under `/internal` no longer get transitive coverage.

We need a general-purpose solution to how this is addressed.  For now, decreasing the coverage is ok.

RickWinter: Makes sense, Is there a plan to get coverage up for App Configuration? Even without the closure, 15% is really low.

jhendrixMSFT: Agreed more coverage would be good.

@linglingye001 @RichardChen820 are you currently tracking this?

linglingye001: Got it, added more tests in auth and revert the change on coverage config.

jhendrixMSFT (approved):

PR #25695, opened 2025-12-02
cosmos: Logging and Region ID canonicalization

analogrelay:
This PR makes a couple fixes to help with diagnosing issues:

* Region IDs are canonicalized before comparison, with spaces removed and all character made lowercase
* We now do a fair bit more logging around the GEM and regional failover logic
* Most headers are now unredacted, unless the user specifies their own list of AllowedHeaders. Authorization tokens remain redacted


sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -155,10 +160,19 @@ func (gem *globalEndpointManager) GetAccountProperties(ctx context.Context) (acc
...
 func newAccountProperties(azResponse *http.Response) (accountProperties, error) {
 	properties := accountProperties{}
-	err := azruntime.UnmarshalAsJSON(azResponse, &properties)
+	unmarshalErr := azruntime.UnmarshalAsJSON(azResponse, &properties)
+
+	// It's safe to call this multiple times (UnmarshalAsJSON does it internally) because the body is buffered and seeked back to start.
+	bodyBytes, err := azruntime.Payload(azResponse)
+
+	// Log the raw JSON
 	if err != nil {
```
pandeyabs, line 169: Should the condition check be reversed here? 
```suggestion
	if err == nil {
```

PR #25707, opened 2025-12-03
Add Float 16 Datatype Support 

tvaron3:
# Description 

Adds the float16 datatype for the vector embedding policy and adds a test creating a policy for each data type


Pilchie:
@tvaron3 / @TheovanKraay - don't we still need this?

tvaron3:
Yes there isn't emulator support for this feature or live tests in go to test this so I left this on stand by. I tested manually and it works. I can reopen and merge commenting out the test for now. @Pilchie

Pilchie:
@tvaron3 - yes, I think we should do that while we're waiting to bring up more testing for this sdk.

sdk/data/azcosmos/emulator_cosmos_container_test.go
```diff
@@ -354,6 +354,95 @@ func TestContainerVectorSearch(t *testing.T) {
...
+// 			// Clean up
+// 			_, err = container.Delete(context.TODO(), nil)
+// 			if err != nil {
+// 				t.Fatalf("Failed to delete container %s: %v", containerID, err)
+// 			}
+
+// 			_ = createdResp // Avoid unused variable warning
+// 		})
+// 	}
+// }
```
analogrelay: I get why we can't run these tests but I'd prefer we just delete them rather than commenting them out. They're not so complicated we couldn't just recreate them once the feature is in the emulator, and commented-out code gets broken fast.

tvaron3: Makes sense removed it

PR #25714, opened 2025-12-03
[azcore] move time parsing to core

gracewilcox:
Part of https://github.com/Azure/azure-sdk-for-go/issues/22314

Moving time to core. Will be consumed by generator.

### DIFF
--- a/sdk/azcore/runtime/datetime/RFC3339_test.go
+++ b/sdk/azcore/runtime/datetime/RFC3339_test.go
@@ -45,7 +45,6 @@ func TestRFC3339(t *testing.T) {
 	require.Equal(t, originalTime, time.Time(dt3))
 }
 
-<<<<<<< HEAD
 func TestRFC3339_WithSpace_JSON(t *testing.T) {
 	dst := timeInfoJSON{}
 	require.NoError(t, json.Unmarshal([]byte(`{"startTime":"2024-01-18 14:18:54Z"}`), &dst))
@@ -96,10 +95,10 @@ func TestEmptyAndNullTime(t *testing.T) {
 	require.Nil(t, dst.Expiry)
 	require.NoError(t, json.Unmarshal([]byte(`{"interval": null}`), &dst))
 	require.Nil(t, dst.Expiry)
-=======
+}
+
 func TestRFC3339_empty(t *testing.T) {
 	tt := RFC3339{}
 	require.NoError(t, xml.Unmarshal([]byte("<RFC3339/>"), &tt))
 	require.Zero(t, tt)
->>>>>>> refs/remotes/origin/grace-time-move
 }


sdk/azcore/runtime/datetime/RFC3339.go
```diff
@@ -0,0 +1,181 @@
...
+	utcDateTime        = "2006-01-02T15:04:05.999999999"
+	utcDateTimeJSON    = `"` + utcDateTime + `"`
+	utcDateTimeNoT     = "2006-01-02 15:04:05.999999999"
+	utcDateTimeJSONNoT = `"` + utcDateTimeNoT + `"`
+	dateTimeNoT        = `2006-01-02 15:04:05.999999999Z07:00`
+	dateTimeJSON       = `"` + time.RFC3339Nano + `"`
+	dateTimeJSONNoT    = `"` + dateTimeNoT + `"`
+)
+
+type DateTimeRFC3339 time.Time
```
jhendrixMSFT: Remove `DateTime` prefix.

jhendrixMSFT: Also will need docs.

sdk/azcore/datetime/date_type.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+const (
+	fullDateJSON = `2006-01-02`
+	jsonFormat   = `%04d-%02d-%02d`
+)
+
+type DateType time.Time
```
jhendrixMSFT: `type PlainDate time.Time`

gracewilcox: fixed

sdk/azcore/runtime/datetime/RFC3339.go
gracewilcox: fixed

sdk/azcore/runtime/datetime/plain_time_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlainTime(t *testing.T) {
+	originalTime := time.Date(2023, time.June, 15, 10, 30, 45, 0, time.UTC)
+	pt := PlainTime(originalTime)
+	result := pt.String()
+	require.Equal(t, "10:30:45Z", result)
+
+	// jsonBytes, err := pt.MarshalJSON()
```
gracewilcox: @jhendrixMSFT, would roundtrip on Marshal/UnmarshalJSON be expected to fail? Just copied over from the current implementation, and don't know if that's expected or if I need to adjust those methods

jhendrixMSFT: It should work.  My guess is that the marshaled value isn't quoted.  Is that the case?

sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+)
+
+// Time is a type constraint that represents the supported time types in the datetime package
+type Time interface {
+	PlainDate | PlainTime | RFC1123 | RFC3339 | Unix
+}
+
+const (
+	fullDateJSON = `2006-01-02`
```
jhendrixMSFT:
```go
plainDateJSON = `"2006-01-02"`
```

sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -0,0 +1,35 @@
...
+	jsonFormat   = `%04d-%02d-%02d`
+)
+
+// PlainDate represents a date value without time information in YYYY-MM-DD format.
+// It wraps time.Time and can be marshaled to and unmarshaled from JSON.
+type PlainDate time.Time
+
+// MarshalJSON marshals the PlainDate to a JSON byte slice.
+func (t PlainDate) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(jsonFormat, time.Time(t).Year(), time.Time(t).Month(), time.Time(t).Day())), nil
```
jhendrixMSFT: You can simplify this.
```go
return []byte(time.Time(t).Format(plainDateJSON)), nil
```

sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -0,0 +1,35 @@
...
+
+package datetime
+
+import (
+	"fmt"
+	"time"
+)
+
+// Time is a type constraint that represents the supported time types in the datetime package
+type Time interface {
```
jhendrixMSFT: How about `TypeConstraint`?

sdk/azcore/runtime/datetime/plain_date_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlainDate(t *testing.T) {
+	originalDate := time.Date(2023, time.January, 15, 0, 0, 0, 0, time.UTC)
+	pd := PlainDate(originalDate)
+
+	jsonBytes, err := pd.MarshalJSON()
+	require.NoError(t, err)
+	require.Equal(t, "2023-01-15", string(jsonBytes))
```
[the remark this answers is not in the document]
jhendrixMSFT:
```go
require.Equal(t, `"2023-01-15"`, string(jsonBytes))
```

sdk/azcore/runtime/datetime/plain_date.go
gracewilcox: Down for either name! Could be datetime.TypeConstraint or datetime.Time (or similiar). I just leaned towards Time because the type constraint examples I found were usual named after the object (example: Integer, Number, etc https://pkg.go.dev/golang.org/x/exp/constraints)

jhendrixMSFT: Ah interesting.  Maybe just `Constraints` then?

sdk/azcore/runtime/datetime/plain_time.go
```diff
@@ -0,0 +1,67 @@
...
+package datetime
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+const (
+	utcTimeJSON    = `"15:04:05.999999999"`
```
jhendrixMSFT: I think we should name these `plainTime`, `plainTimeJSON` etc...

jhendrixMSFT: We should also concatenate values instead of duplicating them (I got lazy).

sdk/azcore/runtime/datetime/plain_time.go
```diff
@@ -0,0 +1,67 @@
...
+	plainTimeJSON    = `"` + plainTime + `"`
+)
+
+// PlainTime represents a time value without date information. It supports HH:MM:SS format
+// with optional nanosecond precision and timezone information.
+type PlainTime time.Time
+
+// MarshalJSON marshals the PlainTime to a JSON byte slice.
+func (t PlainTime) MarshalJSON() ([]byte, error) {
+	s, _ := t.MarshalText()
```
jhendrixMSFT: We can simplify this.
```go
return []byte(time.Time(t).Format(plainTimeJSON)), nil
```

sdk/azcore/runtime/datetime/plain_time.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	layout := plainTimeUTC
+	if tzOffsetRegex.Match(data) {
+		layout = plainTime
+	}
+	return t.Parse(layout, string(data))
+}
+
+// Parse parses a time string using the specified layout
+func (t *PlainTime) Parse(layout, value string) error {
```
jhendrixMSFT: I don't recall where this is used in the generated code.  I only see it here and on `RFC3339`.  Can these be removed, or do we need them for _all_ time types?

sdk/azcore/runtime/datetime/plain_time.go
```diff
@@ -0,0 +1,67 @@
...
+// MarshalJSON marshals the PlainTime to a JSON byte slice.
+func (t PlainTime) MarshalJSON() ([]byte, error) {
+	s, _ := t.MarshalText()
+	return []byte(fmt.Sprintf("\"%s\"", s)), nil
+}
+
+// MarshalText returns a textual representation of PlainTime
+func (t PlainTime) MarshalText() ([]byte, error) {
+	tt := time.Time(t)
+	return []byte(tt.Format(plainTime)), nil
```
jhendrixMSFT: I think this should be `plainTimeUTC` (same in the `String()` method).

sdk/azcore/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # Release History
 
-## 1.20.1-beta.1 (Unreleased)
+## 1.30.0 (Unreleased)
```
jhendrixMSFT: Will also need to update the version `const`.

chlowell:
```suggestion
## 1.21.0 (Unreleased)
```

gracewilcox: Since I'm adding new public api, think it should be a minor version bump instead of a patch version

chlowell: Agreed, however the next minor version is 1.21.0, not 1.30.0. Current version is 1.20.0, next patch would be 1.20.1

sdk/azcore/runtime/datetime/plain_time.go
gracewilcox: don't think it's use in generated code. Think we can make internal

sdk/azcore/CHANGELOG.md
gracewilcox: you're right, I can't count

jhendrixMSFT:
We should port the time-specific tests from the emitter repo to here.
- https://github.com/Azure/autorest.go/pull/1129
- https://github.com/Azure/autorest.go/pull/1152

Both contain some corner-case tests for parsing time formats.

sdk/azcore/runtime/datetime/RFC3339.go
```diff
@@ -0,0 +1,88 @@
...
+package datetime
+
+import (
+	"regexp"
+	"strings"
+	"time"
+)
+
+// Azure reports time in UTC but it doesn't include the 'Z' time zone suffix in some cases.
+var tzOffsetRegex = regexp.MustCompile(`(?:Z|z|\+|-)(?:\d+:\d+)*"*$`)
```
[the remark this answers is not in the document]
jhendrixMSFT: The suggestion is wrong.  And while this might not be perfect, it's been working for years so I'd prefer to not change it.

PR #25738, opened 2025-12-08
cosmos: Fix issues with how read endpoint fallback is selected

analogrelay:
This pull request improves the logic for endpoint selection in the Cosmos DB location cache and enhances test coverage to ensure correct fallback behavior for read and write endpoints. The changes also clarify the representation of operation types using bitmasking and makes a small fix to logging behavior.

**Endpoint selection and fallback improvements:**

* Refactored the logic in `getPrefAvailableEndpoints` (in `cosmos_location_cache.go`) to ensure the fallback endpoint is only added once and always placed at the end if it wasn't already included, preventing duplicates and guaranteeing correct endpoint ordering.
* Updated the representation of `requestedOperations` to use bitmask values for `read` and `write`, allowing for more flexible and clear operation checks.

**Testing enhancements:**

* Added two new tests in `cosmos_location_cache_test.go` to verify that the default endpoint is used as a fallback for reads and that write endpoints are correctly used as fallback for reads when appropriate. These tests assert the correct ordering and inclusion of endpoints for both read and write scenarios.
* Imported the `strings` package to support endpoint order assertions in the new tests.

**Logging fix:**

* Adjusted the logging condition in `newAccountProperties` to only log the raw JSON if there was no error retrieving the payload, improving log accuracy.

### DIFF
--- a/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
+++ b/sdk/data/azcosmos/cosmos_global_endpoint_manager.go
@@ -171,7 +171,7 @@ func newAccountProperties(azResponse *http.Response) (accountProperties, error) 
 			string(bodyBytes)+"\n"+
 			"==================================================\n")
 	} else {
-		log.Write(azlog.EventResponse, "\n===== Database Account Properties =====\n<Failed to parse>\n==================================================\n")
+		log.Write(azlog.EventResponse, "failed to read Database Account Properties")
 	}
 
 	return properties, unmarshalErr


sdk/data/azcosmos/cosmos_global_endpoint_manager.go
```diff
@@ -166,7 +166,7 @@ func newAccountProperties(azResponse *http.Response) (accountProperties, error)
 	bodyBytes, err := azruntime.Payload(azResponse)
 
 	// Log the raw JSON
-	if err != nil {
+	if err == nil {
```
analogrelay, line 169: This was just a silly typo from getting the logging done quickly 🤦🏻‍♀️

sdk/data/azcosmos/cosmos_location_cache.go
```diff
@@ -16,10 +16,10 @@ import (
...
 
 const (
-	none requestedOperations = iota
-	read
-	write
-	all
+	none  requestedOperations = 0
+	read                      = 0b01
+	write                     = 0b10
+	all                       = read | write
```
analogrelay, line 22: No actual change here, the values for these constants are the same, but I'm paranoid and don't like relying on the ordering of these fields to act like a bitmask.

sdk/data/azcosmos/cosmos_location_cache_test.go
```diff
@@ -281,41 +290,31 @@ func TestReadEndpoints(t *testing.T) {
 	}
 
 	lc.lastUpdateTime = time.Now().Add(-1*defaultExpirationTime - 1*time.Second)
-	expectedReadEndpoints := []*url.URL{loc2Endpoint, loc4Endpoint, loc1Endpoint}
+	expectedReadEndpoints := endpointListToString([]url.URL{*loc1Endpoint, *loc2Endpoint, *loc4Endpoint})
```
analogrelay, line 293: There IS a behavior change here. In fact, this test was effectively asserting the buggy behavior. The only reason `loc1Endpoint` was previously expected to be at the end of the list was **because** of the bug.

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
FabianMeiswinkel: Shouldn't it still log the bodyBytes?

sdk/data/azcosmos/cosmos_location_cache.go
FabianMeiswinkel: Id there is any linter for RUST worth adding/enabling this rule there

sdk/data/azcosmos/cosmos_global_endpoint_manager.go
analogrelay: Not likely, this error indicates a problem in reading the actual bytes. I suppose that message isn't quite accurate, as it's not that it "Failed to parse".

sdk/data/azcosmos/cosmos_location_cache.go
analogrelay: Well, this is Go, so that's not super relevant 😅 . But no, there isn't really a linter we can use here. There's nothing necessarily wrong with defining something like an enum in the way we were defining it before. But it was relying on implicit Go syntax behavior in order to know that the values were being assigned appropriately for a bit field. That's not necessarily something a linter can catch and I'm not aware of one that does.

PR #25795, opened 2025-12-17
[App Configuration] add audience error handling policy

linglingye001:
Users running in clouds other than the public cloud must correctly configure ClientOptions.Cloud.Services["AzureAppConfiguration"].Audience when using the ConfigurationClient from the Azure SDK.

There are two main ways users can misconfigure
They do not specify audience when they are running in non-public cloud
They specify audience, and the audience they specify is the wrong one for the cloud is using

In both cases we will get an error when trying to get an Entra ID token that looks like:

"Microsoft.Identity.Client.MsalServiceException: AADSTS500011: The resource principal named https://appconfig.azure.com/ was not found in the tenant named msazurecloud. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant."

We should handle this error and surface up an improved error message

Audience not provided
If we get this error and audience is not provided we should surface an error message that says audience should be configured and link to our public doc that documents the appropriate audience for each cloud

Audience provided and incorrect
If we get this error and the audience is provided but is wrong, we should surface an error message that the configured audience is wrong and link to our public doc that documents the appropriate audience for each cloud.

### DIFF
--- a/sdk/data/azappconfig/cloud_test.go
+++ b/sdk/data/azappconfig/cloud_test.go
@@ -1,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package azappconfig
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/stretchr/testify/require"
+)
+
+type tokenCredFunc func(context.Context, policy.TokenRequestOptions) (azcore.AccessToken, error)
+
+func (t tokenCredFunc) GetToken(ctx context.Context, tro policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	if l := len(tro.Scopes); l != 1 {
+		return azcore.AccessToken{}, fmt.Errorf("unexpected scopes len %d", l)
+	}
+	return t(ctx, tro)
+}
+
+type fakeTransport struct{}
+
+func (fakeTransport) Do(req *http.Request) (*http.Response, error) {
+	return &http.Response{
+		Request:    req,
+		StatusCode: http.StatusNoContent,
+		Body:       http.NoBody,
+		Header:     http.Header{},
+	}, nil
+}
+
+func TestNewClient_SovereignClouds(t *testing.T) {
+	azureBleu := cloud.Configuration{
+		Services: map[cloud.ServiceName]cloud.ServiceConfiguration{
+			ServiceName: {
+				Audience: "https://appconfig.sovcloud-api.fr",
+			},
+		},
+	}
+
+	tests := []struct {
+		label    string
+		endpoint string
+		scope    string
+		cfg      cloud.Configuration
+	}{
+		{
+			label:    "AzureChina",
+			endpoint: "https://example.azconfig.azure.cn",
+			scope:    "https://appconfig.azure.cn/.default",
+			cfg:      cloud.AzureChina,
+		},
+		{
+			label:    "AzureGovernment",
+			endpoint: "https://example.azconfig.azure.us",
+			scope:    "https://appconfig.azure.us/.default",
+			cfg:      cloud.AzureGovernment,
+		},
+		{
+			label:    "AzurePublic",
+			endpoint: "https://example.azconfig.io",
+			scope:    "https://appconfig.azure.com/.default",
+			cfg:      cloud.AzurePublic,
+		},
+		{
+			label:    "AzureBleu",
+			endpoint: "https://example.azconfig.sovcloud-api.fr",
+			scope:    "https://appconfig.sovcloud-api.fr/.default",
+			cfg:      azureBleu,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.label, func(t *testing.T) {
+			client, err := NewClient(tt.endpoint, tokenCredFunc(func(_ context.Context, tro policy.TokenRequestOptions) (azcore.AccessToken, error) {
+				require.Equal(t, tt.scope, tro.Scopes[0])
+				return azcore.AccessToken{}, nil
+			}), &ClientOptions{
+				ClientOptions: policy.ClientOptions{
+					Cloud:     tt.cfg,
+					Transport: &fakeTransport{},
+				},
+			})
+			require.NoError(t, err)
+
+			// Call an API to trigger the pipeline which will call GetToken on our fake cred
+			_, _ = client.GetSetting(context.Background(), "fake-key", nil)
+		})
+	}
+}


sdk/data/azappconfig/client.go
```diff
@@ -59,7 +60,7 @@ func newClient(endpoint string, authPolicy policy.Policy, options *ClientOptions
 
 	cache := synctoken.NewCache()
 	client, err := azcore.NewClient(moduleName, moduleVersion, runtime.PipelineOptions{
-		PerRetry: []policy.Policy{authPolicy, synctoken.NewPolicy(cache)},
+		PerRetry: []policy.Policy{authPolicy, synctoken.NewPolicy(cache), audience.NewAudienceErrorHandlingPolicy(options.Cloud.Services != nil)},
```
juniwang: looks we are checking if Audience is present. why `options.Cloud.Services != nil` indicates the presence of Audience?

sdk/data/azappconfig/internal/audience/policy.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (p *AudienceErrorHandlingPolicy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	if err != nil {
+		if strings.Contains(err.Error(), aadAudienceErrorCode) {
+			if p.AudienceConfigured {
+				return nil, errors.New("unable to authenticate to Azure App Configuration. An incorrect token audience was provided. Please set ClientOptions.Cloud to the appropriate audience for the target cloud. For details on how to configure the authentication token audience visit https://aka.ms/appconfig/client-token-audience")
+			} else {
+				return nil, errors.New("unable to authenticate to Azure App Configuration. No authentication token audience was provided. Please set ClientOptions.Cloud to the appropriate audience for the target cloud. For details on how to configure the authentication token audience visit https://aka.ms/appconfig/client-token-audience")
```
juniwang: I'm confused by the error messages:
- the link `https://aka.ms/appconfig/client-token-audience` includes the correct audience, but I don't see how to set it.
- I'm assuming that `ClientOptions.Cloud` needs to be set to correct cloud. But I did some search, the well-known clouds like `cloud.AzureGovernment` does not include the app config audience, how does it work? 
- I still don't know how to set audience, and I even don't know if it's needed since I'm told to set the ClientOptions.Cloud

sdk/data/azappconfig/client.go
linglingye001: discussed offline. reached out to the Go sdk team for the service name

sdk/data/azappconfig/internal/audience/policy.go
linglingye001: Discussed offline

jhendrixMSFT: Cloud configuration must be specified per data-plane SDK.  See https://github.com/Azure/azure-sdk-for-go/pull/24609/changes for an example of how to implement it (`cloud_config.go` and `service_client.go` for its consumption).

sdk/data/azappconfig/client.go
jhendrixMSFT: You will want to check if the `Audience` field is set.  This will make more sense once you've implemented support for cloud config (see my other comment).

sdk/data/azappconfig/internal/audience/policy.go
```diff
@@ -0,0 +1,40 @@
...
+		AudienceConfigured: audienceConfigured,
+	}
+}
+
+func (p *AudienceErrorHandlingPolicy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	if err != nil {
+		if strings.Contains(err.Error(), aadAudienceErrorCode) {
+			if p.AudienceConfigured {
+				return nil, errors.New("unable to authenticate to Azure App Configuration. An incorrect token audience was provided. Please set ClientOptions.Cloud to the appropriate audience for the target cloud. For details on how to configure the authentication token audience visit https://aka.ms/appconfig/client-token-audience")
```
jhendrixMSFT: What's the error message currently being returned?

sdk/data/azappconfig/internal/audience/policy.go
linglingye001: Looks like the `Bleu` cloud is not supported by package `cloud`, will it be supported?

sdk/data/azappconfig/internal/audience/policy.go
linglingye001: we got the error in other language scenario which contains `aadAudienceErrorCode`. But in Golang we can create client successfully without setting the cloud audience in `AzureChina`

jhendrixMSFT: Some services (e.g. storage) use the same audience across all sovereign clouds.  If the same is true for Azure stack, then there's no need to add cloud config (i.e. you only need to add it as a mechanism to specify a unique audience per endpoint).

sdk/data/azappconfig/internal/audience/policy.go
jhendrixMSFT: What cloud is this, can we get more info?

sdk/data/azappconfig/internal/audience/policy.go
```diff
@@ -0,0 +1,40 @@
...
+func NewAudienceErrorHandlingPolicy(audienceConfigured bool) *AudienceErrorHandlingPolicy {
+	return &AudienceErrorHandlingPolicy{
+		AudienceConfigured: audienceConfigured,
+	}
+}
+
+func (p *AudienceErrorHandlingPolicy) Do(req *policy.Request) (*http.Response, error) {
+	resp, err := req.Next()
+	if err != nil {
+		if strings.Contains(err.Error(), aadAudienceErrorCode) {
```
jhendrixMSFT, line 29: I'm not convinced we need to add this.  Is the error message from MSAL not sufficient?

CC @chlowell

linglingye001: The error message returned from MSAL like this "Microsoft.Identity.Client.MsalServiceException: AADSTS500011: The resource principal named https://appconfig.azure.com/ was not found in the tenant named msazurecloud. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant.", users still don't know what the audience is for a specific cloud and how to set it. This policy provides an improved error message and links to our service public doc.

sdk/data/azappconfig/cloud_config.go
```diff
@@ -0,0 +1,40 @@
...
+//
+//	BleuCloudConfig.Services[ServiceName] = cloud.ServiceConfiguration{
+//		Audience: "https://appconfig.sovcloud-api.fr",
+//	}
+//
+//	clientOptions := azappconfig.ClientOptions{
+//		Cloud: BleuCloudConfig,
+//	}
+//
+//	client, err := azappconfig.NewClient("<endpoint>", credential, &clientOptions)
```
chlowell: This would be better as a testable example, say `Example_cloudConfig()`, so it's easier to discover and `go test` ensures it compiles

sdk/data/azappconfig/internal/audience/policy.go
```diff
@@ -0,0 +1,40 @@
...
+			if p.AudienceConfigured {
+				return nil, errors.New("unable to authenticate to Azure App Configuration. An incorrect token audience was provided. Please set ClientOptions.Cloud for the target cloud . For details on how to configure the authentication token audience visit https://aka.ms/appconfig/client-token-audience and examples in https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/azappconfig/cloud_config.go")
```
jhendrixMSFT: This will swallow the `azidentity.AuthenticationFailedError`.  For these cases we should wrap the error so it's accessible to callers.
```go
return nil, fmt.Errorf("... %w", err)
```
Comment also applies to line 33 below.

jhendrixMSFT: Also, instead of linking to `cloud_config.go`, it would be better to add an example and link to that (props to @chlowell for the idea).

linglingye001: Updated and removed `cloud_config` link, we will add examples in the linked MS learn doc.

sdk/data/azappconfig/cloud_config.go
```diff
@@ -0,0 +1,23 @@
...
+func init() {
+	cloud.AzureChina.Services[ServiceName] = cloud.ServiceConfiguration{
+		Audience: "https://appconfig.azure.cn",
+	}
+	cloud.AzureGovernment.Services[ServiceName] = cloud.ServiceConfiguration{
+		Audience: "https://appconfig.azure.us",
+	}
+	cloud.AzurePublic.Services[ServiceName] = cloud.ServiceConfiguration{
+		Audience: "https://appconfig.azure.com",
+	}
```
RichardChen820, line 22: Will we add Bleu after it's available to customers?

linglingye001: Yes, but it depends on the cloud package to export `Bleu`. We will add example for how to configure cloud/audience in the doc.

sdk/data/azappconfig/internal/audience/policy.go
linglingye001: https://learn.microsoft.com/en-us/industry/sovereign-cloud/national-partner-clouds/overview-national-partner-clouds

RichardChen820 (approved):

RichardChen820 (approved):

PR #25826, opened 2026-01-05
Add some missing methods to the types in datetime

jhendrixMSFT:
Added missing text (un)marshalers on some types so they're all uniform. ~Handle JSON nulls in JSON unmarshalers.~
Cleaned up some missing punctuation in doc comments.


sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -26,13 +28,32 @@ func (t PlainDate) MarshalJSON() ([]byte, error) {
...
+// MarshalText returns a textual representation of PlainDate.
+func (t PlainDate) MarshalText() ([]byte, error) {
+	tt := time.Time(t)
+	return []byte(tt.Format(plainDate)), nil
+}
+
 // UnmarshalJSON unmarshals a JSON byte slice into a PlainDate.
 func (d *PlainDate) UnmarshalJSON(data []byte) (err error) {
+	if string(data) == jsonNull {
+		return nil
```
chlowell: Is this new behavior? Naively, I expect `time.Parse` to return an error when given something unparseable like "null"

jhendrixMSFT: Correct, `time.Parse` will choke on that.  However, for unmarshaling JSON, we don't want to pass the JSON `null` to `time.Parse`.

chlowell: We would today though, no? How do we handle or avoid the error?

sdk/azcore/runtime/datetime/RFC3339_test.go
```diff
@@ -90,12 +86,10 @@ func TestRFC3339_WithSpace_XML(t *testing.T) {
 	require.WithinDuration(t, time.Date(2024, 1, 18, 14, 18, 54, 123000000, time.UTC), time.Time(*dst.Expiry), 0)
 }
 
-func TestEmptyAndNullTime(t *testing.T) {
-	dst := timeInfoSlice{}
-	require.NoError(t, json.Unmarshal([]byte("{}"), &dst))
-	require.Nil(t, dst.Expiry)
```
RickWinter, line 96: Aren't we expecting this to be an error now? Why did you get rid of this test?

jhendrixMSFT: The test didn't validate any of our code.  i.e. the empty object is exercising the JSON unmarshaler in the standard library, and the `"interval"` field doesn't exist in the `timeInfoSlice` type.

sdk/azcore/runtime/datetime/plain_time_test.go
```diff
@@ -36,6 +36,25 @@ func TestPlainTime(t *testing.T) {
...
+	require.Zero(t, pt)
+}
+
+func TestPlainTime_UnmarshalText_Empty(t *testing.T) {
+	var pt datetime.PlainTime
+	err := pt.UnmarshalText([]byte(""))
+	require.NoError(t, err)
+}
+
+func TestPlainTime_UnmarshalText_nil(t *testing.T) {
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/azcore/runtime/datetime/plain_date_test.go
```diff
@@ -65,12 +65,47 @@ func TestPlainDate_UnmarshalJSON_Empty(t *testing.T) {
...
+}
+
+func TestPlainDate_UnmarshalText(t *testing.T) {
+	var plainDate datetime.PlainDate
+	err := plainDate.UnmarshalText([]byte("2023-01-15"))
+	require.NoError(t, err)
+	require.Equal(t, time.Date(2023, time.January, 15, 0, 0, 0, 0, time.UTC), time.Time(plainDate))
+}
+
+func TestPlainDate_UnmarshalText_nil(t *testing.T) {
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/azcore/runtime/datetime/plain_time_test.go
```diff
@@ -36,6 +36,25 @@ func TestPlainTime(t *testing.T) {
...
+	var pt datetime.PlainTime
+	err := pt.UnmarshalText(nil)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

sdk/azcore/runtime/datetime/unix.go
```diff
@@ -17,8 +17,29 @@ func (t Unix) MarshalJSON() ([]byte, error) {
...
+
+// UnmarshalText decodes the textual representation of Unix.
+func (t *Unix) UnmarshalText(data []byte) error {
+	if len(data) == 0 {
+		return nil
+	}
+	return t.parse(data)
+}
+
+// parses a Unix timestamp from a byte slice.
```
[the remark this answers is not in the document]
jhendrixMSFT: method is no exported

sdk/azcore/runtime/datetime/plain_date.go
jhendrixMSFT: Did some digging, we handle the JSON `null` in the unmarshaling helper ([example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/time_rfc3339.go#L100-L110)).  This helper will move as part of the codegen work to ingest these changes (https://github.com/Azure/autorest.go/pull/1787).

I'm going to poke at this a bit more, but we can probably remove these checks since they're performed above us.

jhendrixMSFT: In fact, looks like JSON `null` is handled via the standard library.
```go
func TestRFC3339_Null2(t *testing.T) {
	dst := timeInfoJSON{}
	require.NoError(t, json.Unmarshal([]byte(`{"startTime":null}`), &dst))
	require.Zero(t, dst)
}
```
This doesn't invoke the unmarshaler on our type.

jhendrixMSFT: removed

sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -26,13 +26,29 @@ func (t PlainDate) MarshalJSON() ([]byte, error) {
 	return []byte(time.Time(t).Format(plainDateJSON)), nil
 }
 
+// MarshalText returns a textual representation of PlainDate.
+func (t PlainDate) MarshalText() ([]byte, error) {
```
chlowell: nitpicking
```suggestion
func (p PlainDate) MarshalText() ([]byte, error) {
```

sdk/azcore/runtime/datetime/unix_test.go
```diff
@@ -33,6 +33,31 @@ func TestUnix_Invalid(t *testing.T) {
...
+	tu := datetime.Unix(originalTime)
+	textBytes, err := tu.MarshalText()
+	require.NoError(t, err)
+	expected := fmt.Sprintf("%d", originalTime.Unix())
+	require.Equal(t, expected, string(textBytes))
+}
+
+func TestUnix_UnmarshalText(t *testing.T) {
+	originalTime := time.Date(2021, time.November, 5, 8, 15, 30, 0, time.Local)
+	text := fmt.Sprintf("%d", originalTime.Unix())
```
chlowell, line 47: please forgive more nitpicking
```suggestion
	text := fmt.Sprint(originalTime.Unix())
```

sdk/azcore/runtime/datetime/plain_date.go
```diff
@@ -26,13 +26,29 @@ func (t PlainDate) MarshalJSON() ([]byte, error) {
...
 func (d *PlainDate) UnmarshalJSON(data []byte) (err error) {
 	t, err := time.Parse(plainDateJSON, string(data))
 	*d = (PlainDate)(t)
 	return err
 }
 
+// UnmarshalText decodes the textual representation of PlainDate.
+func (t *PlainDate) UnmarshalText(data []byte) error {
+	if len(data) == 0 {
+		return nil
```
chlowell, line 40:
```go
p := datetime.PlainDate(time.Date(1999, time.December, 31, 0, 0, 0, 0, time.UTC))
newTimeText := []byte{}
if err := p.UnmarshalText(newTimeText); err != nil {
	// unreachable
}
if p.String() == "1999-12-31" {
	// TODO: am I surprised? 🤔
}
```
I'm thinking, probably better to return an error

sdk/azcore/runtime/datetime/plain_date.go
jhendrixMSFT: Fixed 'em all.

sdk/azcore/runtime/datetime/unix_test.go
jhendrixMSFT: Fixed

sdk/azcore/runtime/datetime/plain_date.go
jhendrixMSFT: This is to handle the case when unmarshaling an XML element that's empty (`<PlainDate />`).  This was a real bug we hit in the storage SDKs.  Unfortunately, removing this is non-trivial due to how we emit the XML unmarshalers.  We could in theory hoist it there like we do for JSON `null` checks, but it would be a lot of work (side note: this is why I added those JSON `null` checks as I thought we overlooked them, and forgot the check was done elsewhere).

sdk/azcore/runtime/datetime/datetime.go
```diff
@@ -0,0 +1,32 @@
...
+	"time"
+)
+
+// Constraints is a type constraint that represents the supported time types in the datetime package.
+type Constraints interface {
+	PlainDate | PlainTime | RFC1123 | RFC3339 | Unix
+}
+
+// Parse parses the provided string into one of the supported datetime types.
+func Parse[T Constraints](val string) (time.Time, error) {
```
gracewilcox: Is there a benefit to having this as a public function in azcore vs the current use of `time.Parse` by the generator?

sdk/azcore/runtime/datetime/datetime.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"time"
+)
+
+// Constraints is a type constraint that represents the supported time types in the datetime package.
+type Constraints interface {
+	PlainDate | PlainTime | RFC1123 | RFC3339 | Unix
+}
+
+// Parse parses the provided string into one of the supported datetime types.
```
chlowell:
```suggestion
// Parse parses the string representation of a supported datetime type into a time.Time
```

sdk/azcore/runtime/datetime/datetime.go
jhendrixMSFT: The benefit I see is that it keeps the parsing code in one place.

e.g.  at present we emit something like this for `PlainTime` (taken from [here](https://github.com/Azure/autorest.go/blob/29fed86b893d161794fc7e11c1bdc9b1ec9f3a1b/packages/autorest.go/test/maps/azalias/fake/zz_server.go#L364))
```go
headerTimeParam, err := time.Parse(time.TimeOnly, getHeaderValue(req.Header, "headerTime"))
```
This means that the emitter needs to know the format (`time.TimeOnly`) used by `PlainTime`.  And if it was to change for some reason (admittedly unlikely), then we have to update it in both places.

A more interesting case would be for `RFC3339` where we have handling for `T`, `t`, and other permutations (those are rare though).  The emitter doesn't account for those when parsing time values.

jhendrixMSFT: All that said, I could be convinced that the juice isn't worth the squeeze.

gracewilcox: Yeah, just worried the benefits don't outweigh having it as public API. 
Maybe there could be an internal solution in the emitter for coupling the formats used by datetime types. I know formats like `time.TimeOnly` are used in other places besides the time.Parse scenario.

sdk/azcore/runtime/datetime/plain_date.go
richardpark-msft: Seems worth adding in as a comment so the workaround is made clear.

Just curious, when they do specify <PlainDate/> does that just mean they don't care about the date, or is there a Storage specific default date they expect?

jhendrixMSFT: It's due to the element being required in the XML schema.  In JSON, when you want to say "no value" you can just omit the field from the payload.  That pattern doesn't work for XML though when the schema is enforced.  So, they just send an empty element to mean "no value".

jhendrixMSFT: Added comments explaining the issue.

jhendrixMSFT: For the gory details.
- https://github.com/Azure/autorest.go/pull/1152
- https://github.com/Azure/azure-sdk-for-go/pull/22485

sdk/azcore/runtime/datetime/datetime.go
jhendrixMSFT: I've removed this and the generic type constraint.  The POR is that we will emit both for internal use.

PR #25832, opened 2026-01-06
Remove the old syntax +build directive

RickWinter:


sdk/tracing/azotel/internal/version.go
```diff
@@ -1,5 +1,4 @@
 //go:build go1.18
```
jhendrixMSFT: This build constraint can be removed too.

PR #25890, opened 2026-01-16
Add retry loops to some test cases

jhendrixMSFT:
Tests sometimes fail, presumably due to a delay between updating the value and when the service returns the reflected update.

e.g. https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5760002&view=results


richardpark-msft (approved):

richardpark-msft (approved):
Looks good. RBAC propagation sometimes takes longer than expected.

PR #25912, opened 2026-01-21
Document code customization guidelines for Go SDKs

samvaity:


richardpark-msft:
Adding @gracewilcox, she's also done a lot with code customizations.

documentation/development/generate.md
```diff
@@ -90,6 +90,36 @@ Generated code **must not** be edited, as any edits would be lost on future rege
 That said, if there is a need to customize the generated code, you can add one or more [Autorest directives][autorest_directives] to your autorest.md file.
 This way, the changes are documented and preserved across regenerations.
 
+## Code Customizations
```
jhendrixMSFT: We also have the ability to invoke a post-generation action via [`go generate`](https://go.dev/blog/generate).  This is specified in an SDKs `tspconfig.yaml` file [like this](https://github.com/Azure/azure-rest-api-specs/blob/fd8fd88708d4a1e2bc3d7f064d295b17fb4fa3f7/specification/keyvault/Security.KeyVault.Certificates/tspconfig.yaml#L56).  This does require that a recent version of Go is installed on machine.

The hope is that the need for such customizations is rare, but it does allow a deep ability for customization when `client.tsp` isn't enough.

jhendrixMSFT: [emitter docs](https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/README.md#go-generate) for `go-generate`.

documentation/development/generate.md
gracewilcox, line 10: Add to table of contents

gracewilcox: You could even move post-generation customizations as a subheader under the 'UsingTypespec and tsp-client' section. It is TypeSpec specific and honestly shouldn't be a main header since we don't want people using it casually.

gracewilcox, line 59: could add a section on typespec customizations

documentation/development/generate.md
gracewilcox:
```suggestion
### Post-Generation Code Customizations
> **Note**: **Do not** use on new SDKs.  Post-generation code customizations should only be used to maintain back compatibility on older SDKs. Use TypeSpec customizations instead
```

Honestly, could use a stronger warning. No new SDK's should be using post-generation customizations. They should really only be used to maintain back-compatibility.

richardpark-msft: TBH, I was okay with leaving it out entirely, for this very reason.

jhendrixMSFT: That's a good point.  Do we document the procedure someplace though?

richardpark-msft: I do have some documentation, but not in public repos, for some of the customization that I had to do for Event Grid, for instance.

samvaity: I needed public documentation on code customizations for an inner-loop tooling process. Do you think adding this might lead users to navigate more toward code customizations unintentionally? I mainly wanted documentation to understand what’s available in Go.

richardpark-msft: I think the idea here is that users shouldn't need customizations of the kind that you'd use with `go generate`, so there's no point in documenting that it exists. If we do "recommend" it, then yeah, I expect that Copilot would attempt it, especially since these are the instructions we're feeding it.

We can always add it in later, if we find out there are massive holes in our generation story, but it'd be good to hold it back for now.

documentation/development/generate.md
```diff
@@ -56,6 +58,36 @@ Setting up your project involves a few steps:
 
 Generated code **must not** be edited, as any edits would be lost on future regeneration of content. To make customizations, update the TypeSpec project's `client.tsp` file, then regenerate.
 
+#### TypeSpec Customizations
+
+Before customizing generated code, consider whether your change should be made in TypeSpec (`client.tsp`) instead. TypeSpec customizations are cleaner and survive regeneration. See the [TypeSpec Client Customizations Reference](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/knowledge/customizing-client-tsp.md) for available decorators like `@@clientName`, `@@access`, etc.
```
gracewilcox, line 63: Could you expand this section more? For example, define what a TypeSpec customization is and how to use it. If someone just read this section, they might be confused as-is.

PR #26007, opened 2026-02-10
azcosmos: Go-native ReadMany

analogrelay:
*I had Copilot start by writing a design spec, which I've deleted from the branch now that the implementation is complete. See [here](https://github.com/Azure/azure-sdk-for-go/blob/dab2618558e3e8b2d7f6bcf3a4699e8a1f65ddcb/sdk/data/azcosmos/.specs/ReadItems.md) if you want to read it*

Replaces the per-item point-read strategy in `ReadManyItems` with a query-based approach that groups items by **physical partition range** using Effective Partition Key (EPK) hashing. This issues far fewer HTTP round-trips and reduces RU cost, especially when many items share the same physical partition.

### What changed

**New: EPK hashing (`internal/epk/epk.go`)**
- Implements MurmurHash3-based Effective Partition Key computation (V1, V2 Hash, V2 MultiHash) matching the Cosmos DB backend algorithm.
- Validated against official baseline test data (`PartitionKeyHashBaselineTest.*.xml`).

**New: Query builder (`cosmos_query_builder.go`)**
- Builds parameterized SQL queries for read-many operations using three shapes:
  - **ID-only IN** — when PK path is `/id` and every PK value equals the item ID.
  - **PK + ID IN** — when all items in a chunk share the same logical PK.
  - **OR-of-conjunctions** — general case with multiple logical PKs in the same physical range.

**Replaced: ReadMany strategy (`cosmos_container_read_many.go`)**
- Deleted `executeReadManyWithPointReads` (N individual point reads).
- Added `executeReadManyWithQueries` which:
  1. Fetches container properties and partition key ranges.
  2. Computes EPK for each item and groups by physical partition range (binary search over sorted ranges).
  3. Chunks each range's items at 1,000 per query.
  4. Executes queries concurrently using `x-ms-documentdb-partitionkeyrangeid` header routing.
- Added input validation: empty item IDs now return an error immediately.

**API contract note:** The order of items in the response is **unspecified** — callers must not rely on it matching the input order. Missing items (not found) are silently omitted.

### Testing

- **Unit tests** (`cosmos_query_builder_test.go`, `cosmos_container_read_many_test.go`): Full query text and parameter assertions for all query shapes, EPK range lookup, physical range grouping, and chunk splitting.
- **EPK baseline tests** (`internal/epk/epk_test.go`): Validated against official Cosmos DB hash baseline XML test data.
- **Emulator integration tests** (`emulator_cosmos_read_many_items_test.go`): Existing tests updated and passing.

### Files changed

| File | Change |
|---|---|
| `internal/epk/epk.go` | New — EPK hash computation |
| `internal/epk/epk_test.go` | New — EPK baseline tests |
| `internal/epk/testdata/*.xml` | New — Official baseline test data |
| `cosmos_query_builder.go` | New — Parameterized query builder |
| `cosmos_query_builder_test.go` | New — Query builder unit tests |
| `cosmos_container_read_many.go` | Replaced point-reads with EPK-grouped queries |
| `cosmos_container_read_many_test.go` | New — Grouping/chunking unit tests |
| `cosmos_container.go` | Switch default strategy + add ID validation |
| `partition_key.go` | Add `computeEffectivePartitionKey` method |
| `partition_key_test.go` | EPK computation tests |
| `emulator_cosmos_read_many_items_test.go` | Updated integration tests |

### DIFF
--- a/sdk/data/azcosmos/CHANGELOG.md
+++ b/sdk/data/azcosmos/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+* Improved the performance of the built-in ReadMany implementation (when no QueryEngine is present). See [PR 26007](https://github.com/Azure/azure-sdk-for-go/pull/26007)
+
 ### Breaking Changes
 
 ### Bugs Fixed


sdk/data/azcosmos/.specs/ReadItems.md
```diff
@@ -0,0 +1,335 @@
...
+
+### 3.3 EPK Hashing
+
+The Python SDK computes the EPK via `_get_epk_range_for_partition_key`, which uses the V2 MurmurHash-based effective partition key algorithm. The Go SDK must implement the same hashing or use the Gateway's partition key range resolution.
+
+**Recommended approach:** Use the existing `PartitionKey.toJsonString()` method and the Gateway-returned `partitionKeyRange.minInclusive`/`maxExclusive` boundaries. Each `PartitionKey` value maps to exactly one physical range. We can compute the effective partition key hash client-side (V2 hash) and binary-search the sorted range list, **or** we can simply send each query as a cross-partition query scoped to the known partition key value via the `x-ms-documentdb-partitionkey` header (letting the Gateway route it).
+
+**Simplest correct approach for V1:** Set the `x-ms-documentdb-partitionkey` header (PK value as JSON array) on each per-partition query. The Gateway resolves the physical range automatically. This avoids implementing EPK hashing in Go and is what the existing `NewQueryItemsPager` already does. Chunking is still done by logical PK grouping (items with equal PK value go together).
```
analogrelay: I'm going to have Copilot do this first, because I think it's right to have an initial "correct" implementation. But I don't expect to actually submit a final ready-to-review code PR until I've got EPK hashing in place.

sdk/data/azcosmos/.specs/ReadItems.md
```diff
@@ -0,0 +1,335 @@
...
+   - Match returned `id` fields back to original indices.
+   - Sort by original index.
+   - Sum `x-ms-request-charge`.
+8. **Return `ReadManyItemsResponse`.**
+
+### 3.3 EPK Hashing
+
+The Python SDK computes the EPK via `_get_epk_range_for_partition_key`, which uses the V2 MurmurHash-based effective partition key algorithm. The Go SDK must implement the same hashing or use the Gateway's partition key range resolution.
+
+**Recommended approach:** Use the existing `PartitionKey.toJsonString()` method and the Gateway-returned `partitionKeyRange.minInclusive`/`maxExclusive` boundaries. Each `PartitionKey` value maps to exactly one physical range. We can compute the effective partition key hash client-side (V2 hash) and binary-search the sorted range list, **or** we can simply send each query as a cross-partition query scoped to the known partition key value via the `x-ms-documentdb-partitionkey` header (letting the Gateway route it).
```
tvaron3: IMO epk hashing should be implemented. The benefit of readmany is client side concurrency by physical partitions. We shouldn't just convert it into a cross partititon query.

analogrelay: Yeah, I noted that in a separate comment :). I'm going to have it do the Logical Partition Key approach in a local iteration (on my machine) just to get everything else in place, but I'll add EPK hashing before submitting a full PR.

sdk/data/azcosmos/internal/epk/testdata/PartitionKeyHashBaselineTest.Lists.xml
analogrelay, line 1: These are slightly modified from the [versions in the .NET SDK](https://github.com/Azure/azure-cosmos-dotnet-v3/tree/master/Microsoft.Azure.Cosmos/tests/Microsoft.Azure.Cosmos.Tests/BaselineTest/TestBaseline). Those were not quite testing the "full" EPK algorithm, so their hashes for strings were a bit off (they were hashing _without_ the `0x00` suffix in v1, for some reason). So, I asked Copilot to take the same data and run it through the Python SDK (PR for those tests coming soon; fortunately, Python's implementation is just fine) to generate these updated XML baselines. This gives us a cross-SDK set of EPK test cases we can run.

sdk/data/azcosmos/internal/epk/epk_test.go
```diff
@@ -0,0 +1,137 @@
...
+		return []interface{}{val}
+	case float64:
+		return []interface{}{val}
+	case string:
+		switch val {
+		case "NaN":
+			// Use .NET's NaN bit pattern (0xFFF8000000000000) so the hash
+			// matches the baseline. Go's math.NaN() uses a different bit
+			// pattern (0x7FF8000000000001).
+			return []interface{}{math.Float64frombits(0xFFF8000000000000)}
```
analogrelay, line 63: This really makes me wonder if we should be raising an error if a user attempts to store NaN in a partition key. NaN has no single defined bit pattern, there are Quiet NaNs, Signaling NaNs, and _some_ architectures define a Canonical NaN pattern, but it's largely Processor Architecture-dependent. We should consider either:

1. Forbidding NaN entirely
2. Normalizing all NaN values to a single bit pattern across all SDKs.

(Using the Rust Driver may help us essentially achieve option 2 though, so probably no immediate action to take)

analogrelay: This stumped me and Copilot for a good 20 minutes while it tried to figure out why `math.NaN` wasn't hashing to the same value as `double.NaN` in .NET

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -181,18 +260,123 @@ func (c *ContainerClient) executeReadManyWithPointReads(items []ItemIdentity, re
...
+		azResponse, err := c.database.client.sendQueryRequest(
+			path,
+			ctx,
+			chunk.query,
+			chunk.params,
+			operationContext,
+			&localOpts,
+			func(req *policy.Request) {
+				req.Raw().Header.Set(cosmosHeaderPartitionKeyRangeId, rangeID)
+				req.Raw().Header.Set(cosmosHeaderEnableCrossPartitionQuery, "True")
```
analogrelay: Copilot put this in because we are theoretically doing a "Cross-Logical Partition" query (i.e. querying multiple logical partition key values, all contained in the same physical partition). I don't think it's actually needed though. I'll try removing it and run the tests tomorrow.

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -96,82 +97,160 @@ func (c *ContainerClient) executeReadManyWithEngine(queryEngine queryengine.Quer
...
+	}) - 1
+	if idx < 0 {
+		return "", false
 	}
+	// Verify epkValue < MaxExclusive (empty MaxExclusive means unbounded).
+	r := ranges[idx]
+	if r.MaxExclusive != "" && epkValue >= r.MaxExclusive {
+		return "", false
+	}
+	return r.ID, true
```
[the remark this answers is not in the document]
analogrelay: I don't _believe_ we have special handling for this in the Python SDK.  I had Copilot write a quick test and we indeed see `FFFFF` as greater than `FF`, thus out of the maximum range of EPKs. So I think this is fine and expected behavior. Because we mask the top bits to `3F`, the highest hash possible is `3F` and thus below `FF`.

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -96,82 +97,160 @@ func (c *ContainerClient) executeReadManyWithEngine(queryEngine queryengine.Quer
...
 					return
+				case <-ctx.Done():
+					return
 				default:
 				}
-				item := items[idx]
 
-				itemResponse, err := c.ReadItem(ctx, item.PartitionKey, item.ID, &itemOptions)
+				items, charge, err := c.executeOneChunk(ctx, chunks[idx], queryOpts, operationContext, path, done)
+				results[idx] = chunkResult{items: items, requestCharge: charge, err: err}
```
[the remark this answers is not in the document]
analogrelay: That's OK. The `chunkResult.err` will be set to the cancellation error propagated out of `executeOneChunk`. Later, `collectChunkResults` will see that a chunk failed and fault the entire operation, returning no items and a cancellation error. This is the idiomatic behavior for Go (cancelled operations are considered to have completely failed).

I've written a new test to validate that.

sdk/data/azcosmos/cosmos_container_read_many.go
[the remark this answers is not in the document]
FabianMeiswinkel: Correct - FF is absolute Max

sdk/data/azcosmos/cosmos_container_read_many.go
```diff
@@ -6,16 +6,21 @@ package azcosmos
...
 	"github.com/Azure/azure-sdk-for-go/sdk/internal/log"
 )
 
 // executeReadManyWithEngine executes a query using the provided query engine.
 func (c *ContainerClient) executeReadManyWithEngine(queryEngine queryengine.QueryEngine, items []ItemIdentity, readManyOptions *ReadManyOptions, operationContext pipelineRequestOptions, ctx context.Context) (ReadManyItemsResponse, error) {
-	path, _ := generatePathForNameBased(resourceTypeDocument, operationContext.resourceAddress, true)
+	path, err := generatePathForNameBased(resourceTypeDocument, operationContext.resourceAddress, true)
+	if err != nil {
+		// WE are specifying the resource type here, so this shouldn't fail.
+		panic("invalid resource address in operation context: " + operationContext.resourceAddress)
```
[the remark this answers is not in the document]
analogrelay: Yeah, I guess Copilot is right here

simorenoh:
should we add a Changelog entry?

PR #26032, opened 2026-02-18
Replace `azd config list` with `azd auth status` in TROUBLESHOOTING.md

scottaddie:
The `azd config list` command is no longer the recommended way to verify the logged-in Azure Developer CLI account. Update the troubleshooting guide to use `azd auth status` instead.

### DIFF
--- a/sdk/azidentity/TROUBLESHOOTING.md
+++ b/sdk/azidentity/TROUBLESHOOTING.md
@@ -193,17 +193,26 @@ az account get-access-token --output json --resource https://management.core.win
 
 #### Verify the Azure Developer CLI can obtain tokens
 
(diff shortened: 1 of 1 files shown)


sdk/azidentity/TROUBLESHOOTING.md
```diff
@@ -193,10 +193,10 @@ az account get-access-token --output json --resource https://management.core.win
 
 #### Verify the Azure Developer CLI can obtain tokens
 
-You can manually verify that the Azure Developer CLI is properly authenticated and can obtain tokens. First, use the `config` command to verify the account that is currently logged in to the Azure Developer CLI.
+You can manually verify that the Azure Developer CLI is properly authenticated and can obtain tokens. First, use the `auth status` command to verify the account that is currently logged in to the Azure Developer CLI.
```
chlowell: It might help to note the version that introduced `auth status` and/or suggest `config list` for older versions

PR #26051, opened 2026-02-24
azcore: join request and body close errors in retry policy

### LINKED ISSUE #25649: How to handle error from realClose in retry policy
In the retry policy we have the following bit of code.
```go
_ = rwbody.realClose()
```
We should propagate the error to the caller.  However, there could already be a pending error to return.  So, if `realClose()` also returns an error, what do we do?  Chaining might be the correct solution here, needs investigation.

kritibehl:
Fixes #25649

The retry policy previously ignored errors returned by retryableRequestBody.realClose().

This change ensures:

- realClose() errors are propagated to the caller.
- If a request error already exists, both errors are combined using errors.Join(reqErr, closeErr).
- Callers can inspect either error via errors.Is / errors.As.
- Adds a unit test covering the "both errors non-nil" case.
- Includes CHANGELOG entry.

### DIFF
--- a/sdk/azcore/runtime/policy_retry.go
+++ b/sdk/azcore/runtime/policy_retry.go
@@ -115,7 +115,8 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 		// do this outside the for loop so defers don't accumulate.
 		rwbody = &retryableRequestBody{body: req.Body()}
 		defer func() {
-			// TODO: https://github.com/Azure/azure-sdk-for-go/issues/25649
+			// Match net/http behavior: after a request has produced either a response
+			// or a transport error, request body close errors are not surfaced to callers.
 			_ = rwbody.realClose()
 		}()
 	}


kritibehl:

kritibehl:
Friendly ping for code-owner review when you have a moment.

This is a small fix for #25649 to ensure retryableRequestBody.realClose() errors are not dropped. 
Behavior change is limited to propagating the close error (joined with the request error when both occur). 
Includes a regression test and changelog entry.

Happy to adjust if you'd prefer different error composition semantics.

kritibehl:
Friendly ping on this PR when someone has a moment.

Small fix for #25649 so realClose() errors are not dropped. All checks are passing, and I added a regression test.

Happy to adjust anything if needed. Thanks!

sdk/azcore/CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ time values in various formats used by Azure services.
 
 * Fixed potential panic when decoding base64 strings.
 * Fixed an issue in resource identifier parsing which prevented it from returning an error for malformed resource IDs.
+* Retry policy now propagates errors returned by retryableRequestBody.realClose().
```
jhendrixMSFT: Sorry for the delay.

Can you please rebase this and move this to the `Bugs Fixed` section for the `1.22.0` release?

sdk/azcore/runtime/policy_retry.go
```diff
@@ -115,8 +115,13 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
 		// do this outside the for loop so defers don't accumulate.
 		rwbody = &retryableRequestBody{body: req.Body()}
 		defer func() {
-			// TODO: https://github.com/Azure/azure-sdk-for-go/issues/25649
-			_ = rwbody.realClose()
+			if cerr := rwbody.realClose(); cerr != nil {
```
jhendrixMSFT: From the [docs](https://pkg.go.dev/errors#Join), I think this can be simplified.
```go
cerr := rwbody.realClose()
err = errors.Join(err, cerr)
```

jhendrixMSFT: On second thought, let's leave this as is.

sdk/azcore/CHANGELOG.md
kritibehl: Thanks for the review! I rebased on main and moved the CHANGELOG entry under Bugs Fixed for the 1.22.0 release.

jhendrixMSFT: It looks like this is still under `1.21.0` which has already been released.

sdk/azcore/runtime/policy_retry.go
```diff
@@ -115,8 +115,13 @@ func (p *retryPolicy) Do(req *policy.Request) (resp *http.Response, err error) {
...
+				if err == nil {
+					err = cerr
+				} else {
+					err = errors.Join(err, cerr)
+				}
+			}
```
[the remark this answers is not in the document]
jhendrixMSFT: This is a really good point.  I dug into this more with Copilot, and in the standard library, if the request is successful (i.e. we have an `*http.Response`) then any error from `req.Body.Close()` is dropped.  It's also dropped when there's already a real error to return.

So, we should mimic this behavior to keep in line with how the standard library handles this.  It means we should update the comment with this info and continue to swallow the error.  I don't think joining it to an existing error is really worth it and might introduce other unforeseen side effects (i.e. I'd not want to do this unless we have evidence to the contrary).

Could you please update the comment, removing the `TODO` and explaining why we swallow the error?

sdk/azcore/CHANGELOG.md
jhendrixMSFT: This will go away as all we'll be doing is updating a comment.

sdk/azcore/runtime/policy_retry.go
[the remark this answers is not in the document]
kritibehl: Thanks for digging into this. I updated the comment to remove the TODO and explain that request-body close errors are intentionally swallowed here to stay consistent with net/http behavior.

sdk/azcore/runtime/policy_retry.go
```diff
@@ -5,7 +5,6 @@ package runtime
 
 import (
 	"context"
-	"errors"
```
jhendrixMSFT, line 8: This shouldn't have been removed and is causing the CI failures.

kritibehl: Thanks for catching that, I restored the errors import and pushed the fix.

jhendrixMSFT (approved):

jhendrixMSFT:
Thanks for your contribution!

PR #26058, opened 2026-02-25
Server-side encryption with Rekeying 101 

anu-avocado:
Introduced 3 new request headers for source CPK so that the service can decrypt the source data: 
1. x-ms-source-encryption-key 
2. x-ms-source-encryption-key-sha256 
3. x-ms-source-encryption-algorithm 

API Changes in :
BlockBlobClient.SyncUploadFromUri() 
BlockBlobClient.StageBlockFromUri() 
AppendBlobClient.AppendBlockFromUri() 
PageBlobClient.UploadPagesFromUri() 

Note: 
You must pass-in all 3 request headers or none at all. 

Also, added version parameter to the new client creations as generated code has changed the way it accesses service version.

### DIFF
--- a/sdk/storage/azblob/appendblob/models.go
+++ b/sdk/storage/azblob/appendblob/models.go
@@ -119,7 +119,7 @@ type AppendBlockFromURLOptions struct {
 	Range blob.HTTPRange
 
 	// Optional. Specifies the customer-provided encryption key to use to decrypt the source blob.
-	SourceCustomerProvidedKey *blob.CPKInfo
+	SourceCustomerProvidedKey *blob.SourceCPKInfo
 }
 
 func (o *AppendBlockFromURLOptions) format() (*generated.AppendBlobClientAppendBlockFromURLOptions, *generated.CPKInfo,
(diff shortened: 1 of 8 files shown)


sdk/storage/azblob/appendblob/models.go
```diff
@@ -176,5 +179,16 @@ type ExpiryTypeRelativeToCreation = exported.ExpiryTypeRelativeToCreation
...
+func toGeneratedSourceCPKInfo(o *blob.CPKInfo) *generated.SourceCPKInfo {
+	if o == nil {
+		return nil
+	}
+	return &generated.SourceCPKInfo{
+		SourceEncryptionAlgorithm: o.EncryptionAlgorithm,
+		SourceEncryptionKey:       o.EncryptionKey,
+		SourceEncryptionKeySHA256: o.EncryptionKeySHA256,
+	}
+}
```
souravgupta-msft: This method can be moved in a common package and can be called from all clients.

sdk/storage/azblob/blockblob/models.go
```diff
@@ -414,3 +420,14 @@ type ExpiryTypeNever = exported.ExpiryTypeNever
...
+func toGeneratedSourceCPKInfo(o *blob.CPKInfo) *generated.SourceCPKInfo {
+	if o == nil {
+		return nil
+	}
+	return &generated.SourceCPKInfo{
+		SourceEncryptionAlgorithm: o.EncryptionAlgorithm,
+		SourceEncryptionKey:       o.EncryptionKey,
+		SourceEncryptionKeySHA256: o.EncryptionKeySHA256,
+	}
+}
```
souravgupta-msft: Since, input and output params are same for all blob types, please move this method to a common package

souravgupta-msft:
Live tests for user delegation SAS is failing which will be fixed in later PR

PR #26095, opened 2026-03-02
Cross tenant

anu-avocado:
1. Add support for the DelegatedUserTenantId field in request body in Get User Delegation Key API as well as the SignedDelegatedUserTenantId field from the response body 

API Changes 
* UserDelegationKey.SignedDelegatedUserTenantId 
* GetUserDelegationKeyOptions.DelegatedUserTenantId (options bag for each GetUserDelegationKey() API) 
* Add support for a new SAS field skdutid in the User Delegation SAS, which should come from the SignedDelegatedUserTenantId returned from above 

### DIFF
--- a/sdk/storage/azblob/service/client_test.go
+++ b/sdk/storage/azblob/service/client_test.go
@@ -1921,7 +1921,7 @@ func (s *ServiceRecordedTestsSuite) TestServiceClientDefaultAudience() {
 	}
 }
 
-func (s *ServiceRecordedTestsSuite) TestGetUserDelegationCredentialError() {
+func (s *ServiceUnrecordedTestsSuite) TestGetUserDelegationCredentialError() {
 	_require := require.New(s.T())
 
 	accountName, _ := testcommon.GetGenericAccountInfo(testcommon.TestAccountDefault)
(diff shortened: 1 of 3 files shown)


sdk/storage/azblob/sas/query_params.go
```diff
@@ -293,6 +295,11 @@ func (p *QueryParameters) SignedDelegatedUserObjectID() string {
 	return p.signedDelegatedUserObjectID
 }
 
+// SignedDelegatedUserTenantId returns SignedDelegatedUserTenantId
+func (p *QueryParameters) SignedDelegatedUserTenantId() string {
```
jhendrixMSFT: `SignedDelegatedUserTenantID` (comment applies to other places where "Id" isn't capitalized).

sdk/storage/azblob/sas/service.go
```diff
@@ -194,6 +194,11 @@ func (v BlobSignatureValues) SignWithUserDelegation(userDelegationCredential *Us
 
 	udkStart, udkExpiry, _ := formatTimesForSigning(*udk.SignedStart, *udk.SignedExpiry, time.Time{})
 
+	var signedDelegatedUserTenantId string
+	if udk.SignedDelegatedUserTid != nil {
```
jhendrixMSFT: Is `SignedDelegatedUserTid` part of public surface area?  If so, then we should standardize on naming as I see this and `SignedDelegatedUserTenantId` (I assume "Tid" is short for Tenant ID).  If it's internal only then we should fix the casing so it's consistent with other names (`SignedDelegatedUserTID`).

sdk/storage/azblob/sas/query_params.go
souravgupta-msft: Done, updated at all places

sdk/storage/azblob/sas/service.go
souravgupta-msft: Done, updated *Tid to *TenantID

PR #26106, opened 2026-03-03
azcore/runtime: propagate W3C trace context headers in HTTP trace policy

### LINKED ISSUE #24008: Propagate distributed tracing context in HTTP tracing policy
Distributed tracing policy does not propagate context in HTTP headers

https://github.com/Azure/azure-sdk-for-go/blob/799c8579d2d8f32d3f30457cd6da736be08d9c6b/sdk/azcore/runtime/policy_http_trace.go#L67

(which should be done according to our [common policy](https://azure.github.io/azure-sdk/general_implementation.html#distributed-tracing))

> DO pass the context to the backend service through the appropriate headers (traceparent and tracestate per [W3C Trace-Context](https://www.w3.org/TR/trace-context/) standard)) to support [Azure Monitor](https://azure.microsoft.com/services/monitor/). This is generally done with the HTTP pipeline.

This is used by greenfield Azure services to correlate things.

Some notes/examples  on how it can be done

- OTel has implementation of W3C trace context propagator, here's the [golang one](https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/trace_context.go) 
- We just need to call propagator.inject similarly to this [example](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/7d7606c6523471ebd284ac3e37c07f23055cfda2/instrumentation/net/http/otelhttp/transport.go#L129) 
- 
Note: Otel gives user an ability to customize propagator, we should ignore one configured by the user and use W3C trace context implementation since that's the standard - Azure services are not going to support custom propagation formats.

kritibehl:
Fixes #24008

Inject W3C Trace Context headers (`traceparent` / `tracestate`) into outgoing
HTTP requests in `sdk/azcore/runtime/policy_http_trace.go` when tracing is enabled.

Implementation:
- Uses `propagation.TraceContext{}` explicitly instead of the global propagator.
- Injects headers after span creation so the active span context is propagated.

Test:
- Adds `TestHTTPTracePolicyInjectsW3CTraceContextHeaders`
- Verifies `traceparent` header injection when a valid OTel SpanContext exists.


kritibehl:
Friendly ping for code-owner review when you have a moment.

Quick summary of what this does:
- Adds 4 lines to `policy_http_trace.go` to inject `traceparent`/`tracestate` headers after span creation using `propagation.TraceContext{}` explicitly (not the global propagator, per the note in #24008)
- Adds `TestHTTPTracePolicyInjectsW3CTraceContextHeaders` validating header injection with a real OTel SpanContext

One thing I flagged proactively: `go.opentelemetry.io/otel` was already an indirect dependency in `sdk/azcore/go.mod` — this change makes it direct since the propagation call is now in production code. Happy to restructure if you’d prefer it stay indirect or be gated differently.

Happy to adjust anything else if you have concerns.

kritibehl:
Friendly ping on this PR when someone has a moment.
All checks are passing, and this is a small change to inject W3C trace context headers in the HTTP trace policy, with a focused test for traceparent propagation.
Happy to revise if needed. Thanks!

sdk/azcore/runtime/policy_http_trace.go
```diff
@@ -86,6 +88,8 @@ func (h *httpTracePolicy) Do(req *policy.Request) (resp *http.Response, err erro
 		}()
 
 		req = req.WithContext(ctx)
+		// Propagate W3C Trace Context headers (traceparent/tracestate) into the outgoing request.
+		propagation.TraceContext{}.Inject(ctx, propagation.HeaderCarrier(req.Raw().Header))
```
jhendrixMSFT: Sorry for the delay, and thanks for taking a look at this.

This will need to be refactored to avoid the direct dependency on any OTel modules.  For context, our intent in `azcore` is to not be tied to any one implementation and is why we have the `azcore/tracing` package to serve as an abstraction layer.

I'm not sure exactly how this would look though (having thought about tracing in some time).

jhendrixMSFT (changes requested):

sdk/azcore/runtime/policy_http_trace.go
kritibehl: Thanks for the review — that makes sense.

I agree `azcore/runtime` shouldn’t depend directly on OTel, and that this should stay behind the `azcore/tracing` abstraction layer.

I’ll rework the change so the HTTP trace policy uses an `azcore/tracing`-level abstraction for trace-context propagation rather than calling OTel propagation APIs from runtime.

I may sketch a small helper/API in `azcore/tracing` to handle header injection for the active span context, and then wire the policy through that. If there’s a specific shape you’d prefer for that abstraction, I’m happy to follow it; otherwise I’ll put together a first pass in that direction.

kritibehl: Done, the OTel call is out of azcore/runtime. The HTTP trace policy now calls tracer.InjectHeaders(ctx, req.Raw().Header), with the concrete W3C injection living in sdk/tracing/azotel. sdk/azcore/go.mod has no OTel entries. All checks passing.

The azotel wiring will need a follow-up PR once this merges and a new azcore version is cut.

kritibehl:
Rebased onto latest main and resolved the dependency diff. Still active.

kritibehl:
/reopen

Still interested in continuing this contribution. Happy to rebase and address any remaining feedback.

PR #26180, opened 2026-03-09
Revert CODEOWNERS entry for sdk/monitor/ingestion/azlogs

scottaddie:
Remove the entry for Logs Ingestion, per discussion with @RickWinter and @gracewilcox


.github/CODEOWNERS
```diff
@@ -81,9 +81,6 @@
 # PRLabel: %Monitor
 /sdk/monitor/                                                      @Azure/azure-sdk-write-monitor-data-plane @Azure/azure-sdk-write-monitor-query-logs @chlowell @gracewilcox @jhendrixMSFT
 
-# PRLabel: %Monitor
```
RickWinter, line 84: The crate still exists and until its deprecated and removed it needs to be supported. They still own it till that day arrives.

Is Monitor team aware they still need to support the package?

RickWinter (changes requested):

PR #26203, opened 2026-03-10
fix: add missing consumers.Delete() when checkpoint init fails (#24983)

### LINKED ISSUE #24983: [azeventhubs, v1+v2] Missing a consumer.Delete() call when we a checkpoint is invalid
As pointed out here (#24933), we were missing a call to `consumers.Delete` if we fail to init on a checkpoint (this is purely an in-memory operation). This leaves open a possibility that a consumer that can never work will block us from creating one that could, at a later time.

It's a small window of failure, but we should fix it.

This will need to be fixed in both v1 and v2.

j7nw4r:
## Summary

Fixes #24983

When `getStartPosition()` returns an error due to an invalid checkpoint, the partially initialized consumer was left in the `sync.Map`, blocking future consumer creation for that partition. This adds a `consumers.Delete()` call to clean up the map entry, matching the existing cleanup pattern already used for `openPartitionClient` errors.

### Changes

- Added `consumers.Delete(ownership.PartitionID)` in `processor.go` `addPartitionClient()` when `getStartPosition()` fails

---

### DIFF
--- a/sdk/messaging/azeventhubs/processor_unit_test.go
+++ b/sdk/messaging/azeventhubs/processor_unit_test.go
@@ -573,6 +573,31 @@ func TestUnit_Processor_addPartitionClient(t *testing.T) {
 		require.Nil(t, pc)
 	})
 
+	t.Run("getStartPositionFailure", func(t *testing.T) {
+		// A bad checkpoint (no Offset or SequenceNumber) causes getStartPosition()
+		// to fail. Without the fix from #24983, the stale consumer entry in the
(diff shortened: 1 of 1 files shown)


richardpark-msft:
When you bump up the version in the changelog you also need to change this file's version to match: https://github.com/Azure/azure-sdk-for-go/blob/20ebf405e6a07ee0d531485134ba1164067cb1cb/sdk/messaging/azeventhubs/internal/constants.go#L7

After each release there's an automatic PR that gets created to bump the version number, or you can just do it manually.

